PR #20723, opened 2023-08-31
[native] Enable cache TTL and export stats

gggrace14:
Enable TTL for AsyncDataCache and SsdCache, if AsyncDataCache is
enabled. It will call CacheTTLControl::applyTTL(ttlSecs).

Introduce three configs for TTL,
cache.velox.ttl-enabled
cache.velox.ttl-threshold
cache.velox.ttl-check-interval

Expose the following age counters for AsyncDataCache and SsdCache
presto_cpp.cache_max_age_secs
presto_cpp.memory_cache_num_cumulative_aged_out_entries
presto_cpp.ssd_cache_cumulative_aged_out_entries
presto_cpp.ssd_cache_cumulative_aged_out_regions

as well as
presto_cpp.ssd_cache_cached_regions
presto_cpp.ssd_cache_cached_entries
presto_cpp.ssd_cache_cached_bytes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
+++ b/presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
@@ -22,6 +22,7 @@
 #include "velox/common/base/StatsReporter.h"
 #include "velox/common/base/SuccinctPrinter.h"
 #include "velox/common/caching/AsyncDataCache.h"
+#include "velox/common/caching/CacheTTLController.h"
 #include "velox/common/caching/SsdFile.h"
 #include "velox/common/memory/MemoryAllocator.h"
 #include "velox/common/memory/MmapAllocator.h"
@@ -277,6 +278,12 @@ void PeriodicTaskManager::updateCacheStats() {
       kCounterMemoryCacheTotalPrefetchBytes, memoryCacheStats.prefetchBytes);
   RECORD_METRIC_VALUE(
       kCounterMemoryCacheSumEvictScore, memoryCacheStats.sumEvictScore);
+  RECORD_METRIC_VALUE(
+      kCounterSsdCacheCachedEntries, memoryCacheStats.ssdStats->entriesCached);
+  RECORD_METRIC_VALUE(
+      kCounterSsdCacheCachedRegions, memoryCacheStats.ssdStats->regionsCached);
+  RECORD_METRIC_VALUE(
+      kCounterSsdCacheCachedBytes, memoryCacheStats.ssdStats->bytesCached);
 
   // Interval cumulatives.
   RECORD_METRIC_VALUE(
@@ -307,7 +314,6 @@ void PeriodicTaskManager::updateCacheStats() {
   lastMemoryCacheEvictions_ = memoryCacheStats.numEvict;
   lastMemoryCacheEvictionChecks_ = memoryCacheStats.numEvictChecks;
   lastMemoryCacheStalls_ = memoryCacheStats.numWaitExclusive;
-  lastMemoryCacheAllocClocks_ = memoryCacheStats.allocClocks;
 
   // All time cumulatives.
   RECORD_METRIC_VALUE(
@@ -327,6 +333,7 @@ void PeriodicTaskManager::updateCacheStats() {
   RECORD_METRIC_VALUE(
       kCounterMemoryCacheNumCumulativeAllocClocks,
       memoryCacheStats.allocClocks);
+
   if (memoryCacheStats.ssdStats != nullptr) {
     RECORD_METRIC_VALUE(
         kCounterSsdCacheCumulativeReadEntries,
@@ -340,12 +347,6 @@ void PeriodicTaskManager::updateCacheStats() {
     RECORD_METRIC_VALUE(
         kCounterSsdCacheCumulativeWrittenBytes,
         memoryCacheStats.ssdStats->bytesWritten);
-    RECORD_METRIC_VALUE(
-        kCounterSsdCacheCumulativeCachedEntries,
-        memoryCacheStats.ssdStats->entriesCached);
-    RECORD_METRIC_VALUE(
-        kCounterSsdCacheCumulativeCachedBytes,
-        memoryCacheStats.ssdStats->bytesCached);
     RECORD_METRIC_VALUE(
         kCounterSsdCacheCumulativeOpenSsdErrors,
         memoryCacheStats.ssdStats->openFileErrors);
@@ -374,6 +375,31 @@ void PeriodicTaskManager::updateCacheStats() {
         kCounterSsdCacheCumulativeReadCheckpointErrors,
         memoryCacheStats.ssdStats->readCheckpointErrors);
   }
+
+  if (auto* cacheTTLController =
+          velox::cache::CacheTTLController::getInstance()) {
+    RECORD_METRIC_VALUE(
+        kCounterCacheMaxAgeSecs,
+        cacheTTLController->getCacheAgeStats().maxAgeSecs);
+
+    RECORD_METRIC_VALUE(
+        kCounterMemoryCacheNumAgedOutEntries,
+        memoryCacheStats.numAgedOut - lastMemoryCacheAgedOuts_);
+    lastMemoryCacheAgedOuts_ = memoryCacheStats.numAgedOut;
+    RECORD_METRIC_VALUE(
+        kCounterMemoryCacheNumCumulativeAgedOutEntries,
+        memoryCacheStats.numAgedOut);
+
+    if (memoryCacheStats.ssdStats != nullptr) {
+      RECORD_METRIC_VALUE(
+          kCounterSsdCacheCumulativeAgedOutEntries,
+          memoryCacheStats.ssdStats->entriesAgedOut)
+      RECORD_METRIC_VALUE(
+          kCounterSsdCacheCumulativeAgedOutRegions,
+          memoryCacheStats.ssdStats->regionsAgedOut);
+    }
+  }
+
   LOG(INFO) << "Cache stats:\n" << memoryCacheStats.toString();
 }
 
--- a/presto-native-execution/presto_cpp/main/PeriodicTaskManager.h
+++ b/presto-native-execution/presto_cpp/main/PeriodicTaskManager.h
@@ -134,6 +134,7 @@ class PeriodicTaskManager {
   int64_t lastMemoryCacheEvictionChecks_{0};
   int64_t lastMemoryCacheStalls_{0};
   int64_t lastMemoryCacheAllocClocks_{0};
+  int64_t lastMemoryCacheAgedOuts_{0};
 
   // Operating system related stats.
   int64_t lastUserCpuTimeUs_{0};
--- a/presto-native-execution/presto_cpp/main/PrestoServer.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoServer.cpp
@@ -43,6 +43,7 @@
 #include "presto_cpp/presto_protocol/presto_protocol.h"
 #include "velox/common/base/Counters.h"
 #include "velox/common/base/StatsReporter.h"
+#include "velox/common/caching/CacheTTLController.h"
 #include "velox/common/caching/SsdCache.h"
 #include "velox/common/file/FileSystems.h"
 #include "velox/common/memory/MmapAllocator.h"
@@ -97,6 +98,21 @@ std::string stringifyConnectorConfig(
     out << "  " << key << "=" << value << "\n";
   }
   return out.str();
+}
+
+bool isCacheTtlEnabled() {
+  const auto* systemConfig = SystemConfig::instance();
+  if (systemConfig->cacheVeloxTtlEnabled()) {
+    VELOX_USER_CHECK(
+        systemConfig->cacheVeloxTtlThreshold() > std::chrono::seconds::zero(),
+        "Config cache.velox.ttl-threshold must be positive.");
+    VELOX_USER_CHECK(
+        systemConfig->cacheVeloxTtlCheckInterval() >
+            std::chrono::seconds::zero(),
+        "Config cache.velox.ttl-check-interval must be positive.");
+    return true;
+  }
+  return false;
 }
 
 } // namespace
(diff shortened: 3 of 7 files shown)


gggrace14:
To match with Presto better, will change the configs and counters as below

Native configs
cache.velox.ttl-threshold
cache.velox.ttl-check-interval

Native counters
CachePagesAgesMax

Compared with Presto
Presto configs
cache.alluxio.ttl-threshold
cache.alluxio.ttl-check-interval

Presto counters
presto.com.facebook.alluxio:name=Client.CachePagesAges.presto:type=histograms:Max

gggrace14:
Add config cache.alluxio.ttl-enabled as TTL switch, instead of cache-ttl-frequency-secs, to match with Presto

spershin:
@gggrace14 
Chatted offline.
Decided to make config params close to Presto Java.
Presto Java has
```
               .put("cache.alluxio.ttl-check-interval", "60s")
                .put("cache.alluxio.ttl-enabled", "true")
                .put("cache.alluxio.ttl-threshold", "60m")

```We should have the equivalent
```
cache.velox.ttl-check-interval
cache.velox.ttl-threshold
cache.velox.ttl-enabled

```
And also support different units suffix.
Velox has `toDuration` function to help us with that.

File: presto-native-execution/presto_cpp/main/common/Counters.h [hunk 1]
```diff
@@ -310,10 +335,15 @@ constexpr folly::StringPiece kCounterSsdCacheCumulativeWrittenEntries{
...
     "presto_cpp.ssd_cache_cumulative_written_bytes"};
-constexpr folly::StringPiece kCounterSsdCacheCumulativeCachedEntries{
-    "presto_cpp.ssd_cache_cumulative_cached_entries"};
-constexpr folly::StringPiece kCounterSsdCacheCumulativeCachedBytes{
-    "presto_cpp.ssd_cache_cumulative_cached_bytes"};
+/// Cumulated number of SsdCache entries that are aged out given
+/// configured TTL.
+constexpr folly::StringPiece kCounterSsdCacheCumulativeAgedOutEntries{
+    "presto_cpp.ssd_cache_cumulative_aged_out_entries"};
+/// Cumulated number of SsdCache regions that are aged out given
```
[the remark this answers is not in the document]
spershin: Also, does 'aged out' means they were evicted as well?
Maybe add clarification.

presto-native-execution/presto_cpp/main/common/Counters.cpp
```diff
@@ -86,6 +86,19 @@ void registerPrestoMetrics() {
       0,
       62l * 1024 * 1024 * 1024, // max bucket value: 62GB
       100);
+
+  /// ================== Cache Counters ==================
+
+  DEFINE_HISTOGRAM_METRIC(
+      kCounterCacheMaxAgeSecs,
```
spershin, line 93: Do we need the histogram here?

presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
```diff
@@ -341,11 +361,11 @@ void PeriodicTaskManager::updateCacheStats() {
...
     RECORD_METRIC_VALUE(
-        kCounterSsdCacheCumulativeCachedEntries,
-        memoryCacheStats.ssdStats->entriesCached);
+        kCounterSsdCacheCumulativeAgedOutEntries,
+        memoryCacheStats.ssdStats->entriesAgedOut)
     RECORD_METRIC_VALUE(
-        kCounterSsdCacheCumulativeCachedBytes,
-        memoryCacheStats.ssdStats->bytesCached);
+        kCounterSsdCacheCumulativeAgedOutRegions,
+        memoryCacheStats.ssdStats->regionsAgedOut);
```
spershin, line 368: I'm a bit confused with all these cumulative counters.
What is the value of them?
I would assume we are interested in the instantaneous number rather than 'from the start of the server'...

spershin, line 368: IF this about aged out entries - do we only care about them when TTL is on or we always care about them?

spershin (commented):
Looks, good, series of questions.

presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
```diff
@@ -300,6 +314,9 @@ void PeriodicTaskManager::updateCacheStats() {
   RECORD_METRIC_VALUE(
       kCounterMemoryCacheNumAllocClocks,
       memoryCacheStats.allocClocks - lastMemoryCacheAllocClocks_);
+  RECORD_METRIC_VALUE(
+      kCounterMemoryCacheNumAgedOutEntries,
+      memoryCacheStats.numAgedOut - lastMemoryCacheAgedOuts_);
```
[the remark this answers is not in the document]
gggrace14: They'll be always 0 if we don't have TTL enabled. I can wrap them in condition. I'm okay with either case

presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
gggrace14: We only care about them when TTL is on

presto-native-execution/presto_cpp/main/PeriodicTaskManager.h
```diff
@@ -134,6 +134,7 @@ class PeriodicTaskManager {
   int64_t lastMemoryCacheEvictionChecks_{0};
   int64_t lastMemoryCacheStalls_{0};
   int64_t lastMemoryCacheAllocClocks_{0};
+  int64_t lastMemoryCacheAgedOuts_{0};
```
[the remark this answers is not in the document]
gggrace14: You're right. Added. I often only look at the cumulative version

presto-native-execution/presto_cpp/main/common/Counters.cpp
gggrace14: Histogram gives us MAX, while DEFINE_METRIC("", StatType::AVG) does not support MAX.

For a native worker that we collect data points from over 600 sec, each data point is the max age of all cache entries:
With histogram, we can get the MAX of all max age data points collected over the 600 sec.
With StatType::AVG, we can get the AVG of all max age data points collected over the 600 sec.

So histogram is precise, while StatType::AVG is approximate.

presto-native-execution/presto_cpp/main/common/Counters.h
```diff
@@ -214,6 +214,13 @@ constexpr folly::StringPiece kCounterSpillPeakMemoryBytes{
 
 /// ================== Cache Counters ==================
 
+/// Max possible age of AsyncDataCache and SsdCache entries since the raw file
+/// was opened to load the cache.
+constexpr folly::StringPiece kCounterCacheMaxAgeSecs{
+    "presto_cpp.cache_max_age_secs"};
+
+/// ------------- AsyncDataCache counters -------------
```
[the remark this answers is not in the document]
gggrace14: Revised.

presto-native-execution/presto_cpp/main/common/Counters.h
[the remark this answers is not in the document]
gggrace14: Revised

presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
gggrace14: Yes, the cumulative counters are "from the start of the server". And we can only collect the cumulative counters, primarily due to our current setup of counter reporting and collecting I think.

presto-native-execution/presto_cpp/main/common/Counters.cpp
spershin: Why 600 sec?
Default in Meta is 60 sec.
I've done some testing in the past with AVG and collecting max in the code and exporting it with another AVG.
They were roughly identical.

If you think, most of the counters we use are AVG and they are quite useful.

Imagine you exported 1000 sec at some point and in 30 seconds you exported 1030 sec.
So you would have 1015 sec data point for that minute (60 sec).

The difference is rather irrelevant: 1030 and 1015.

Of course, the can be cliffs:

1000 -> 30 in 30 seconds (when we evict). Then you have ~500, then ~60.
That will also be quite clear.

I don't believe we need second or even minute precision here.
We are talking about data TTL which would be months, weeks or days, unlikely even hours, although hours would work fine here too.

But I don't have a strong opinion here.
We can compare MAX with AVG and see how much delta we've got.

presto-native-execution/presto_cpp/main/common/Counters.cpp
```diff
@@ -86,6 +86,19 @@ void registerPrestoMetrics() {
...
+      10 * 60, // 10-min bucket,
+      0,
+      10 * 24 * 60 * 60, // max value 30 days
+      50,
+      100);
+
+  /// ------------- AsyncDataCache counters -------------
```
spershin, line 100: Still ----, not ====?

presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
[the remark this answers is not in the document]
spershin: It is better not to export what we don't need. Especially if we already have a condition.
Less load on the counters backend.

spershin (approved):
Thanks!

presto-native-execution/presto_cpp/main/common/Counters.cpp
gggrace14: Revised

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -61,6 +62,7 @@
 
 namespace facebook::presto {
 using namespace facebook::velox;
+using namespace std::chrono;
```
tanjialiang, line 65: Maybe let's not use this namespace to make it a bit more explicit as similar to other part of the codebase.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -99,6 +101,13 @@ std::string stringifyConnectorConfig(
   return out.str();
 }
 
+bool isCacheTtlEnabled() {
+  const auto* systemConfig = SystemConfig::instance();
+  return systemConfig->cacheVeloxTtlEnabled() &&
+      systemConfig->cacheVeloxTtlThreshold() > seconds::zero() &&
+      systemConfig->cacheVeloxTtlCheckInterval() > seconds::zero();
```
tanjialiang, line 108: Should we solely rely on cacheVeloxTtlEnabled flag for this method and add sanity check when setting other configs so that it throws if ttl is enabled but other settings are 0? This can cause silent failure and introduce troubles when debugging.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -630,6 +639,18 @@ void PrestoServer::initializeVeloxMemory() {
...
+    if (isCacheTtlEnabled()) {
+      cache::CacheTTLController::create(*cache_);
+      PRESTO_STARTUP_LOG(INFO) << fmt::format(
+          "Cache TTL is enabled, with TTL {} enforced every {}.",
+          succinctMillis(duration_cast<milliseconds>(
+                             systemConfig->cacheVeloxTtlThreshold())
+                             .count()),
+          succinctMillis(duration_cast<milliseconds>(
+                             systemConfig->cacheVeloxTtlCheckInterval())
+                             .count()));
```
tanjialiang, line 652: Yeah do sanity check inside here. And let the server crash by throwing if check fails

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -710,6 +731,26 @@ void PrestoServer::addServerPeriodicTasks() {
     periodicTaskManager_->addTask(
         [server = this]() { server->yieldTasks(); }, timeslice, "yield_tasks");
   }
+
+  if (isCacheTtlEnabled()) {
+    int64_t ttlThreshold =
```
tanjialiang, line 736: const auto ttlThresholdS

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -710,6 +731,26 @@ void PrestoServer::addServerPeriodicTasks() {
...
+        duration_cast<seconds>(
+            SystemConfig::instance()->cacheVeloxTtlThreshold())
+            .count();
+    int64_t ttlCheckInterval =
```
tanjialiang, line 740: const auto ttlCheckIntervalMs

presto-native-execution/presto_cpp/main/common/Configs.h
```diff
@@ -448,6 +448,17 @@ class SystemConfig : public ConfigBase {
   static constexpr std::string_view kDriverMaxPagePartitioningBufferSize{
       "driver.max-page-partitioning-buffer-size"};
 
+  /// Enable TTL for AsyncDataCache and SSD cache.
+  static constexpr std::string_view kCacheVeloxTtlEnabled{
+      "cache.velox.ttl-enabled"};
+  /// TTL duration for AsyncDataCache and SSD cache entries.
+  static constexpr std::string_view kCacheVeloxTtlThreshold{
```
tanjialiang, line 455: Let's remove velox and name it retention
kAsyncDataCacheTtlRetentionPeriod

And let's move these configs close to AsyncDataCache configs.

gggrace14: Presto (java) also has these configs, and, agreed with Sergey's suggestion, we want to match and in the above way.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
gggrace14: Revised

presto-native-execution/presto_cpp/main/PrestoServer.cpp
gggrace14: Revised

presto-native-execution/presto_cpp/main/common/Configs.h
tanjialiang: Yeah I'm a bit suspicious on the purpose of matching Java if these are not passed over by Java coordinator. The downside of matching Java here is clear though: We are making C++ configs more confusing. As someone who does not understand fully on C++ codebase, he might not relate directly these configs with async data cache while it is actually directly related IMO.

tanjialiang (approved):

presto-native-execution/presto_cpp/main/common/Configs.h
gggrace14: I tend to want to match Java, but don't have strong opinion. It happens to us quite often that we look for what value Presto Java or Trino are using and how they works. So matching the config name gives ppl confidence that the config is what they're looking for. btw, this TTL applies both to async data cache and ssd.

tanjialiang (approved):

presto-native-execution/presto_cpp/main/common/Configs.h
tanjialiang: AsyncData cache is a cache wrapper umbrella that takes ram as first level and ssd as second. So if ttl applies to async data cache then it should semantically be applied to both ram and ssd.

PR #20058, opened 2023-07-06
[native] Fail fast when dereferencing anonymous fields

mbasmanova:
Velox derefence expression (FieldAccessTypedExpr) takes field name. If field
name is not unique the query may fail or return incorrect results. Fail the
query during plan translation to avoid cryptic errors and hard-to-detect wrong
results.

See https://github.com/facebookincubator/velox/issues/5536

```
== NO RELEASE NOTE ==
```

PR #20727, opened 2023-08-31
[native] Map Presto DereferenceExpr to Velox DereferenceTypedExpr 

kevinwilfong:
Velox added a new DereferenceTypedExpr that takes a field index in place of a field name, aligning with the implementation 
in Presto.  This fixes an existing issue where non-unique field names may cause queries to fail or produce incorrect results.

See https://github.com/facebookincubator/velox/issues/5536 and https://github.com/prestodb/presto/pull/20058

### DIFF
--- a/presto-native-execution/src/test/java/com/facebook/presto/spark/TestPrestoSparkNativeSimpleQueries.java
+++ b/presto-native-execution/src/test/java/com/facebook/presto/spark/TestPrestoSparkNativeSimpleQueries.java
@@ -102,12 +102,16 @@ public class TestPrestoSparkNativeSimpleQueries
     }
 
     @Test
+    public void testDereference()
+    {
+        assertQuery("SELECT transform(array[row(orderkey, comment)], x -> x[2]) FROM orders");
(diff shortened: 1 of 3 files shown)


kevinwilfong:
I need to advance Velox

mbasmanova (commented):
Thank you for the fix. Please, add [native] prefix to commit title and add details to commit message if possible.

presto-native-execution/src/test/java/com/facebook/presto/spark/TestPrestoSparkNativeSimpleQueries.java
```diff
@@ -101,13 +101,17 @@ public void testJoins()
         assertQuery("SELECT * FROM orders o, lineitem l WHERE o.orderkey = l.orderkey AND o.orderkey % 2 = 1");
     }
 
+    @Test
+    public void testDerference()
```
mbasmanova, line 105: typo: testDerference -> testDereference

mbasmanova (approved):
Thanks. Looks good % a typo below.

PR #20735, opened 2023-09-01
[native] Add virtual destructor for PeriodicServiceInventoryManager

amitkdutta:
PeriodicServiceInventoryManager has virtual functions, hence it
requires virtual destructor. Also fixing a spell error.

Without virtual destructor, Meta internal build fails with `class has virtual functions but non-virtual destructor`


zacw7 (approved):
LGTM

mshang816 (approved):

xiaoxmeng (approved):
Can you briefly describe what's the fix? Thanks!

PR #20736, opened 2023-09-01
[native] Print velox user error log upon failure during start up.

amitkdutta:
If presto start up fails due to a user error (e.g. config file is missing), the exception is not printed and server exits without any trace of what happened. When this code was written, throwing VeloxUserError was printing the error. However, later this behavior is altered (https://github.com/facebookincubator/velox/pull/5610). Hence the error needs to be logged for better debugging of incidents where config files are missing.

### DIFF
--- a/presto-native-execution/presto_cpp/main/PrestoServer.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoServer.cpp
@@ -175,8 +175,7 @@ void PrestoServer::run() {
     }
     nodeLocation_ = nodeConfig->nodeLocation();
   } catch (const VeloxUserError& e) {
-    // VeloxUserError is always logged as an error.
-    // Avoid logging again.
+    PRESTO_STARTUP_LOG(ERROR) << "Failed to start server due to " << e.what();
     exit(EXIT_FAILURE);
   }
 


gggrace14 (approved):

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -175,8 +175,7 @@ void PrestoServer::run() {
     }
     nodeLocation_ = nodeConfig->nodeLocation();
   } catch (const VeloxUserError& e) {
-    // VeloxUserError is always logged as an error.
-    // Avoid logging again.
+    LOG(ERROR) << e.what();
```
xiaoxmeng, line 178: Can you give more context here on error?
```
LOG(ERROR) << "...: " << e.what();
```

laithsakka:
I think we shall change the exceptions to velox runtime errors. those will generate logs. 
Velox user error is more like an error in the SQL? shall a missing config be VELOX runtime error?
VeloxUserError is for errors that are suppressed by a try

presto-native-execution/presto_cpp/main/PrestoServer.cpp
majetideepak: May be "Failed to start server due to " << e.what()

amitkdutta: Velox exception prints a good amount of detail, hence I did not add any prefix. Below is the printout for this case:

```
I20230831 23:16:48.217065 328279 Configs.cpp:210] [PRESTO_STARTUP] Registered properties from '/Users/adutta/Documents/code/presto/presto-native-execution/etc/node.properties':
  node.internal-address=127.0.0.1
  node.location=testing-location
  node.id=e4901aae-a1c9-4ff7-97a9-5687835ad54c
  node.environment=testing
E20230831 23:16:48.217298 328279 PrestoServer.cpp:178] Exception: VeloxUserError
Error Source: USER
Error Code: INVALID_ARGUMENT
Reason: Couldn't open config file ~/presto/presto-native-execution/etc/velox.properties for reading.
Retriable: False
Function: readConfig
File:~/presto/presto-native-execution/presto_cpp/main/common/ConfigReader.cpp
Line: 28
Stack trace:
Stack trace has been disabled.Use --velox_exception_user_stacktrace=true to enable it.
```

amitkdutta:
> I think we shall change the exceptions to velox runtime errors. those will generate logs. Velox user error is more like an error in the SQL? shall a missing config be VELOX runtime error? VeloxUserError is for errors that are suppressed by a try

This is a good suggestion. Let me audit the user error in presto server and fix all together.

PR #20745, opened 2023-09-01
Fix bug in randomize null key in outer join optimizer for varchar type

feilong-liu:
## Description
Fix a bug in randomize null join key in outer join optimizer. It happens when the join key are of type varchar, and has value like 'r0', 'r11' etc in left side and NULL in the right side or value like 'l1', 'l12' on the right side and NULL in the left side. In this PR, an additional (l.key is NULL) = (r.key is NULL) is added so that the salted key for NULL value will not match non NULL value.

## Motivation and Context
Fix a bug in optimizer.

## Impact
Fix a bug in optimizer.

## Test Plan
Add unit tests and [verifier test](https://our.internmc.facebook.com/intern/presto/verifier/results/?test_id=137663)

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Fix a bug in `RandomizeNullKeyInOuterJoin` optimizer. 
    It produced incorrect results when the join key are of type varchar, and has value like 'r0', 'r11' etc in left side and NULL in the right side or value like 'l1', 'l12' on the right side and NULL in the left side.
```

### DIFF
--- a/presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java
+++ b/presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java
@@ -53,6 +53,7 @@ import java.util.stream.IntStream;
 import static com.facebook.presto.SystemSessionProperties.ADD_PARTIAL_NODE_FOR_ROW_NUMBER_WITH_LIMIT;
 import static com.facebook.presto.SystemSessionProperties.ENABLE_INTERMEDIATE_AGGREGATIONS;
 import static com.facebook.presto.SystemSessionProperties.FIELD_NAMES_IN_JSON_CAST_ENABLED;
+import static com.facebook.presto.SystemSessionProperties.HASH_PARTITION_COUNT;
 import static com.facebook.presto.SystemSessionProperties.KEY_BASED_SAMPLING_ENABLED;
 import static com.facebook.presto.SystemSessionProperties.KEY_BASED_SAMPLING_FUNCTION;
 import static com.facebook.presto.SystemSessionProperties.KEY_BASED_SAMPLING_PERCENTAGE;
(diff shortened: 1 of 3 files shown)


presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java
```diff
@@ -6282,6 +6283,13 @@ public void testRandomizeNullKeyOuterJoin()
                 .setSystemProperty(RANDOMIZE_OUTER_JOIN_NULL_KEY_STRATEGY, "key_from_outer_join")
                 .build();
         assertQuery(enableKeyFromOuterJoin, multipleJoin, getSession(), multipleJoin);
+
+        Session enableRandomizeFourPartition = Session.builder(getSession())
+                .setSystemProperty(RANDOMIZE_OUTER_JOIN_NULL_KEY, "true")
+                .setSystemProperty(HASH_PARTITION_COUNT, "4")
```
kaikalur, line 6289: keep it simple and set it to 1 for better chance of hitting the bug?

kaikalur (approved):

mlyublena:
Can you add more details about the bug? Was that a correctness issue? What is the significance of these values (r0, r11 etc)?

feilong-liu:
> Can you add more details about the bug? Was that a correctness issue? What is the significance of these values (r0, r11 etc)?

The randomize null key rule works by converting NULL join key to 'l'+random() for left side and 'r'+random() for right side. Here random() is a random integer ranging from 0 to hash partition count. For the example query added in this PR `select t.k, t2.k, t2.v from (values 'r0', 'r1', 'r2', 'r3') t(k) left join (values (null, 1), (null, 2), (null, 3), (null, 4)) t2(k, v) on t.k = t2.k"`, after applying this rule, we will have the join key on the right side to be `r0` (when setting hash partition count to be 1), and it will match the 'r0' on the left side, which should not match. Hence, we add a isNULL check for both sides. In this case, isNULL will be false for left side and true for right side and will not match.

pranjalssh (approved):

Issue #20355, opened 2023-07-20
Improve history based optimizer


feilong-liu:
Improve HBO (History Based Optimizer) in two aspects, which includes improve latency of HBO to make it work for both long running and short running queries, and use it to turn on more cost based optimizations.

- Reduce latency for HBO
HBO depends on data from two sources. One is from the storage which stores historical statistics. The other is metastore which is used to get data of input statistics. The access of these two data sources can be in the critical path in respect to latency.
The items involved in this part includes:
  - Add timeout for HBO optimizer, HBO will be skipped in case of timeout
  - Optimize access to the external data sources
- Increase the usage of HBO
Currently HBO is only used for join type/reorder and partial aggregation turn on/off. We can apply it to more optimizations. This includes:
  - Collecting more statistics. For example, skew of data
  - Enable for more optimizations. For example, RandomizeNullKeyInOuterJoin is used to null salt outer joins with many null keys. This is currently default to disabled and can be made cost based after we have more accurate statistics from HBO

pranjalssh:
Cc @tdcmeehan

PR #20337, opened 2023-07-18
Add number of NULL elements in hash table into operator stats

feilong-liu:
Add number of null elements in a hash table and size of a hash table into plan statistics.
It will be recorded in history base statistics, and used for mitigating nulls skew in outer joins.

### Test plan - (Please fill in how you tested your changes)

Tested in https://github.com/facebookexternal/presto-facebook/pull/2413

```
== NO RELEASE NOTE ==
```

PR #20746, opened 2023-09-01
Enable outer join null salt by cost model

feilong-liu:
## Description
Enable randomize null key for outer join optimization with HBO. Enable it by setting `randomize_outer_join_null_key_strategy` to be 'COST_BASED', and the trigger condition can be set by session properties `randomize_outer_join_null_key_null_count_threshold` and `randomize_outer_join_null_key_null_ratio_threshold`.

## Motivation and Context
Randomize null key for outer join was added to mitigate null value skew in outer joins. It's default to be disabled. With change in https://github.com/prestodb/presto/pull/20337, we have data about number of keys in join build side and the number of keys which are null in HBO. With these information we can enable the randomize null key for outer join with cost model.

## Impact
Will have randomize null key for outer join optimization enabled with cost based model.

## Test Plan
Tested end to end ([link](https://www.internalfb.com/intern/presto/query/?query_id=20230905_230840_00009_5th46#plan))

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Add an option to make `RandomizeNullKeyInOuterJoin` cost based. It can be enabled by setting `randomize_outer_join_null_key_strategy` to be `cost_based`. The trigger condition can be set by session properties `randomize_outer_join_null_key_null_count_threshold` and `randomize_outer_join_null_key_null_ratio_threshold`.

```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
+++ b/presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
@@ -258,6 +258,7 @@ public final class SystemSessionProperties
     public static final String OPTIMIZE_MULTIPLE_APPROX_PERCENTILE_ON_SAME_FIELD = "optimize_multiple_approx_percentile_on_same_field";
     public static final String RANDOMIZE_OUTER_JOIN_NULL_KEY = "randomize_outer_join_null_key";
     public static final String RANDOMIZE_OUTER_JOIN_NULL_KEY_STRATEGY = "randomize_outer_join_null_key_strategy";
+    public static final String RANDOMIZE_OUTER_JOIN_NULL_KEY_NULL_RATIO_THRESHOLD = "randomize_outer_join_null_key_null_ratio_threshold";
     public static final String IN_PREDICATES_AS_INNER_JOINS_ENABLED = "in_predicates_as_inner_joins_enabled";
     public static final String PUSH_AGGREGATION_BELOW_JOIN_BYTE_REDUCTION_THRESHOLD = "push_aggregation_below_join_byte_reduction_threshold";
     public static final String KEY_BASED_SAMPLING_ENABLED = "key_based_sampling_enabled";
@@ -1540,6 +1541,11 @@ public final class SystemSessionProperties
                         false,
                         value -> RandomizeOuterJoinNullKeyStrategy.valueOf(((String) value).toUpperCase()),
                         RandomizeOuterJoinNullKeyStrategy::name),
+                doubleProperty(
+                        RANDOMIZE_OUTER_JOIN_NULL_KEY_NULL_RATIO_THRESHOLD,
+                        "Enable randomizing null join key for outer join when ratio of null join keys exceed the threshold",
+                        0.02,
+                        false),
                 booleanProperty(
                         OPTIMIZE_CONDITIONAL_AGGREGATION_ENABLED,
                         "Enable rewriting IF(condition, AGG(x)) to AGG(x) with condition included in mask",
@@ -2704,6 +2710,11 @@ public final class SystemSessionProperties
             return RandomizeOuterJoinNullKeyStrategy.ALWAYS;
         }
         return session.getSystemProperty(RANDOMIZE_OUTER_JOIN_NULL_KEY_STRATEGY, RandomizeOuterJoinNullKeyStrategy.class);
+    }
+
+    public static double getRandomizeOuterJoinNullKeyNullRatioThreshold(Session session)
+    {
+        return session.getSystemProperty(RANDOMIZE_OUTER_JOIN_NULL_KEY_NULL_RATIO_THRESHOLD, Double.class);
     }
 
     public static boolean isOptimizeConditionalAggregationEnabled(Session session)
--- a/presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java
@@ -365,6 +365,7 @@ public class FeaturesConfig
     {
         DISABLED,
         KEY_FROM_OUTER_JOIN, // Enabled only when join keys are from output of outer joins
+        COST_BASED,
         ALWAYS
     }
 
--- a/presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/RandomizeNullKeyInOuterJoin.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/RandomizeNullKeyInOuterJoin.java
@@ -15,6 +15,10 @@ package com.facebook.presto.sql.planner.optimizations;
 
 import com.facebook.presto.Session;
 import com.facebook.presto.common.type.VarcharType;
+import com.facebook.presto.cost.CachingStatsProvider;
+import com.facebook.presto.cost.PlanNodeStatsEstimate;
+import com.facebook.presto.cost.StatsCalculator;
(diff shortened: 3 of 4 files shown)


kaikalur (approved):

presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java
```diff
@@ -364,6 +364,7 @@ public enum RandomizeOuterJoinNullKeyStrategy
     {
         DISABLED,
         KEY_FROM_OUTER_JOIN, // Enabled only when join keys are from output of outer joins
+        COST_BASED,
```
mlyublena, line 367: do we plan to make this the default value?

presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
```diff
@@ -1540,6 +1542,16 @@ public SystemSessionProperties(
                         false,
                         value -> RandomizeOuterJoinNullKeyStrategy.valueOf(((String) value).toUpperCase()),
                         RandomizeOuterJoinNullKeyStrategy::name),
+                integerProperty(
+                        RANDOMIZE_OUTER_JOIN_NULL_KEY_NULL_COUNT_THRESHOLD,
+                        "Enable randomizing null join key for outer join when number of null join keys exceed the threshold",
+                        100_000,
+                        false),
+                doubleProperty(
```
mlyublena, line 1550: can we simplify this and only have one parameter to control? we can keep the ratio parameter and choose not to apply the transformation if the overall cardinality of the input is small

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/RandomizeNullKeyInOuterJoin.java
```diff
@@ -259,9 +273,12 @@ public PlanNode visitJoin(JoinNode joinNode, RewriteContext<Set<VariableReferenc
             PlanNode rewrittenLeft = context.rewrite(joinNode.getLeft(), context.get());
             PlanNode rewrittenRight = context.rewrite(joinNode.getRight(), context.get());
 
+            PlanNodeStatsEstimate joinEstimate = statsProvider.getStats(joinNode);
+            boolean enabledByCostModel = strategy.equals(COST_BASED) && joinEstimate.getNullJoinBuildKeyCount() > getRandomizeOuterJoinNullKeyNullCountThreshold(session)
+                    && joinEstimate.getNullJoinBuildKeyCount() / joinEstimate.getJoinBuildKeyCount() > getRandomizeOuterJoinNullKeyNullRatioThreshold(session);
             List<JoinNode.EquiJoinClause> candidateEquiJoinClauses = joinNode.getCriteria().stream()
                     .filter(x -> isSupportedType(x.getLeft()) && isSupportedType(x.getRight()))
-                    .filter(x -> strategy.equals(ALWAYS) || (strategy.equals(KEY_FROM_OUTER_JOIN) && (context.get().contains(x.getLeft()) || context.get().contains(x.getRight()))))
+                    .filter(x -> enabledByCostModel || strategy.equals(ALWAYS) || (strategy.equals(KEY_FROM_OUTER_JOIN) && (context.get().contains(x.getLeft()) || context.get().contains(x.getRight()))))
```
mlyublena, line 281: nit: context.get() is used multiple times in this function, can we store it in a variable?

mlyublena, line 281: the strategy is fixed and doesn't need to get evaluated for each element, let's have a boolean variable outside the loop:
```
boolean isEnabled = enabledByCostModel || strategy.equals(ALWAYS) || (strategy.equals(KEY_FROM_OUTER_JOIN);
```

presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java
feilong-liu: Yes, will make this to be default after changes rolled out.

presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
feilong-liu: I assume that we will need some tuning for these parameters. Having more flexibility could help here.

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/RandomizeNullKeyInOuterJoin.java
feilong-liu: (strategy.equals(KEY_FROM_OUTER_JOIN) is combined with other condition check involving the lambda element. We can combine "enabledByCostModel || strategy.equals(ALWAYS)" but still leave the (strategy.equals(KEY_FROM_OUTER_JOIN) alone.

mlyublena: oh right!
yes, let's combine the first two. Can also introduce a variable for isEnabledForOuterJoinKeys and/or introduce a function for the outerjoin key check. I think shortening the check will make it a bit more readable

presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
mlyublena: I understand the flexibility argument, I'm just worried there are too many parameters creates an exponential number of combinations to tune.
What is the use case here that is not covered by just tuning the ratio?

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/RandomizeNullKeyInOuterJoin.java
feilong-liu: > yes, let's combine the first two

I think it's fine to have the three conditions in the filter? As the saving for not needed to evaluate for each element is negligible, and we currently have three settings to enable the optimization, having all three should be better for readability and not confusing people on wondering if two are special and need to be combined first.

> Can also introduce a variable for isEnabledForOuterJoinKeys and/or introduce a function for the outerjoin key check

Added a function for the outer join key case

mlyublena: it's more about readability than overhead: if I see something computed inside the loop, I think it is dependent on the loop

mlyublena: one solution would be to hide all of the checks inside a function that gets called for every element

feilong-liu: But having all three checks in one function and gets called for every element does not help readability imo? 

We have three cases where the optimization should be enabled 1) Users set the session to always enable, not related to join stats and not join key specific 2) Users set the session to be cost based, and cost model shows that it can be enabled, not join key specific 3) Users set the session to be enabled when join key is from the output of an outer join, is join key specific and need to check for every join key. These three checks takes different inputs to make decisions.

Puting all three checks within one function and gets called for every element will hide the fact that the first two decisions are join key independent and does not need to be called for every element. And also this function will mix the arguments needed for each case which is not a win?

mlyublena (approved):

pranjalssh (approved):

PR #20619, opened 2023-08-21
[native] Change json location for reuse with other third parties

czentgr:
Nlohmmann's json.hpp is included in the source tree rather than a third party install. The path used cannot be used in lieu of a system installation. Thus, changing the location of the json.hpp to match the path found when installed into the system.

This allows reuse of the already included library by other third party includes that expect it to be installed.

## Description
Refactoring the location of the json.hpp to allow it to be used by other third parties from the source tree rather than a system install.
The previous location did not use the directory structure necessary.

## Motivation and Context
The json.hpp should be able to be reused by other third parties without relying on a separate dependency installation. The dependency is already available in the source tree and it only needs to be resolved from the usual path.

## Impact
Internal build change only.

## Test Plan
Build succeeds.

## Contributor checklist

- [X] Please make sure your submission complies with our [development](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#development), [formatting](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#formatting), [commit message](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#commit-formatting-and-pull-requests), and [attribution guidelines](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#attribution).
- [X] PR description addresses the issue accurately and concisely.  If the change is non-trivial, a GitHub Issue is referenced.
- [X] Documented new properties (with its default value), SQL syntax, functions, or other functionality.
- [X] If release notes are required, they follow the [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines).
- [X] Adequate tests were added if applicable.
- [x] CI passed.

## Release Notes
```
== NO RELEASE NOTE ==
```

PR #20749, opened 2023-09-02
[native] Change nlohmann json location as done in PR 20619

amitkdutta:
This is a copy of https://github.com/prestodb/presto/pull/20619
PR 20619 got reverted due to a meta internal reason. The reason is 
since this has a file move, internal build code needs to be updated.


gggrace14 (approved):

czentgr (approved):
Thank you! I'm sure there were some discussions about this.
For reference, I suppose it is better to use the same naming convention when including a third party then how it would appear via install in the system (notably project path(s) and filename(s)). This would allow to find such dependencies in the source tree rather than the system (or other install path) and, therefore, allows for easy reuse.

PR #20752, opened 2023-09-03
Add support for array_least_frequent UDF

jainavi17:
## Description
UDF to return the least frequent element of an array

## Motivation and Context
Currently there is no easy way to get least frequent element of an array. This change add that support natively.

## Test Plan
`./mvnw clean install -Dtest=TestArraySqlFunctions -Dmaven.javadoc.skip=true -T1C -fn -pl presto-main`

== RELEASE NOTES ==

General Changes
- Add function :func:`array_least_frequent `

  - `SELECT ARRAY_LEAST_FREQUENT(ARRAY[1, 2, 2, 3, 3, 3]); -- 1`

### DIFF
--- a/presto-docs/src/main/sphinx/functions/array.rst
+++ b/presto-docs/src/main/sphinx/functions/array.rst
@@ -80,6 +80,15 @@ Array Functions
 
     Concatenates the elements of the given array using the delimiter and an optional string to replace nulls.
 
+.. function:: array_least_frequent(array(T)) -> array(T)
+
+    Returns the least frequent element of an array. If there are multiple elements with same frequency, the function returns the largest element.
+
+.. function:: array_least_frequent(array(T), n) -> array(T)
+
+    Returns n least frequent elements of an array. The elements are based on increasing order of their frequencies.
+    If two elements have same frequency then element with higher value will appear before lower value.
+
 .. function:: array_max(x) -> x
 
     Returns the maximum value of input array.
@@ -205,7 +214,7 @@ Array Functions
 .. function:: combinations(array(T), n) -> array(array(T))
 
     Returns n-element combinations of the input array.
-    If the input array has no duplicates, ``combinations`` returns n-element subsets. 
+    If the input array has no duplicates, ``combinations`` returns n-element subsets.
     Order of subgroup is deterministic but unspecified. Order of elements within
     a subgroup are deterministic but unspecified. ``n`` must not be greater than 5,
     and the total size of subgroups generated must be smaller than 100000::
--- a/presto-main/src/main/java/com/facebook/presto/operator/scalar/sql/ArraySqlFunctions.java
+++ b/presto-main/src/main/java/com/facebook/presto/operator/scalar/sql/ArraySqlFunctions.java
@@ -91,6 +91,26 @@ public class ArraySqlFunctions
         return "RETURN cardinality(array_duplicates(input)) > 0";
     }
 
+    @SqlInvokedScalarFunction(value = "array_least_frequent", deterministic = true, calledOnNullInput = true)
+    @Description("Determines the least frequent element in the array. If there are multiple elements, the function returns the smallest element")
+    @TypeParameter("T")
+    @SqlParameter(name = "input", type = "array(T)")
+    @SqlType("array<T>")
+    public static String array_least_frequent()
+    {
+        return "RETURN IF(COALESCE(CARDINALITY(REMOVE_NULLS(input)), 0) = 0, NULL, TRANSFORM(SLICE(ARRAY_SORT(TRANSFORM(MAP_ENTRIES(ARRAY_FREQUENCY(REMOVE_NULLS(input))), x -> ROW(x[2], x[1]))), 1, 1), x -> x[2]))";
+    }
+
+    @SqlInvokedScalarFunction(value = "array_least_frequent", deterministic = true, calledOnNullInput = true)
+    @Description("Determines the n least frequent element in the array in the ascending order of the elements.")
+    @TypeParameter("T")
+    @SqlParameters({@SqlParameter(name = "input", type = "array(T)"), @SqlParameter(name = "n", type = "bigint")})
+    @SqlType("array<T>")
+    public static String array_n_least_frequent()
+    {
+        return "RETURN IF(n < 0, fail('n must be greater than or equal to 0'), IF(COALESCE(CARDINALITY(REMOVE_NULLS(input)), 0) = 0, NULL, TRANSFORM(SLICE(ARRAY_SORT(TRANSFORM(MAP_ENTRIES(ARRAY_FREQUENCY(REMOVE_NULLS(input))), x -> ROW(x[2], x[1]))), 1, n), x -> x[2])))";
+    }
+
     @SqlInvokedScalarFunction(value = "array_max_by", deterministic = true, calledOnNullInput = true)
     @Description("Get the maximum value of array, by using a specific transformation function")
     @TypeParameter("T")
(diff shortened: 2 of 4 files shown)


presto-docs/src/main/sphinx/functions/array.rst
```diff
@@ -80,6 +80,10 @@ Array Functions
 
     Concatenates the elements of the given array using the delimiter and an optional string to replace nulls.
 
+.. function:: array_least_frequent(array(T)) -> T
+
+    Returns the least frequent element of an array. If there are multiple elements with same frequency, the function returns the biggest value.
```
steveburnett, line 85:
```suggestion
    Returns the least frequent element of an array. If there are multiple elements with the same frequency, the function returns the largest value.
```

A very small rephrasing suggestion.

steveburnett (commented):
LGTM (docs), with a very small suggestion for consideration.

presto-main/src/main/java/com/facebook/presto/operator/scalar/sql/ArraySqlFunctions.java
```diff
@@ -136,4 +136,14 @@ public static String removeNulls()
...
     }
+
+    @SqlInvokedScalarFunction(value = "array_least_frequent", deterministic = true, calledOnNullInput = true)
+    @Description("Determines the least frequent elemement in the array. If there are multiple elements, the function returns the biggest element")
+    @TypeParameter("T")
+    @SqlParameter(name = "input", type = "array(T)")
+    @SqlType("T")
+    public static String array_least_frequent()
+    {
+        return "RETURN IF(input IS NULL, NULL, IF(CARDINALITY(REMOVE_NULLS(input)) = 0, NULL, MAP_KEYS(MAP_TOP_N(TRANSFORM_VALUES(ARRAY_FREQUENCY(REMOVE_NULLS(input)), (k, v) -> -1 * v), 1))[1]))";
```
kaikalur, line 147: two things 
 * see if you want to N as a param - n least (and most) frequent
 * also see if you can use map_entries, transform them to (v,k) and sort/reverse

presto-main/src/test/java/com/facebook/presto/operator/scalar/sql/TestArraySqlFunctions.java
```diff
@@ -242,4 +242,25 @@ public void testArraySortDesc()
...
+        // Empty Case
+        assertFunction("ARRAY_LEAST_FREQUENT(ARRAY [])", UNKNOWN, null);
+        // Null Case
+        assertFunction("ARRAY_LEAST_FREQUENT(null)", UNKNOWN, null);
+        assertFunction("ARRAY_LEAST_FREQUENT(ARRAY [NULL])", UNKNOWN, null);
+        assertFunction("ARRAY_LEAST_FREQUENT(ARRAY [1, 2, 2, NULL])", INTEGER, 1);
+        assertFunction("ARRAY_LEAST_FREQUENT(ARRAY [NULL, NULL, NULL])", UNKNOWN, null);
+        // Error Case
+        assertInvalidFunction("ARRAY_LEAST_FREQUENT(ARRAY [1, 2, 'a'])", SemanticErrorCode.TYPE_MISMATCH);
+    }
```
kaikalur, line 265: we need real tests to test the logic

kaikalur (changes requested):

presto-main/src/main/java/com/facebook/presto/operator/scalar/sql/ArraySqlFunctions.java
jainavi17: Hmm...we can take N as param to cover more generalised cases. Let me take N as optional parameter default to 1. WDYT?

presto-main/src/test/java/com/facebook/presto/operator/scalar/sql/TestArraySqlFunctions.java
jainavi17: I can have it removed then.

kaikalur: Oh oops - nvm. this is fine

steveburnett (approved):
LGTM! (docs)

presto-main/src/main/java/com/facebook/presto/operator/scalar/sql/ArraySqlFunctions.java
```diff
@@ -91,6 +91,26 @@ public static String arrayHasDuplicatesVarchar()
...
+    {
+        return "RETURN IF(COALESCE(CARDINALITY(REMOVE_NULLS(input)), 0) = 0, NULL, TRANSFORM(SLICE(ARRAY_SORT(MAP_ENTRIES(ARRAY_FREQUENCY(REMOVE_NULLS(input))),(a, b) -> IF(a[2] > b[2], 1, IF(a[2] = b[2], IF(a[1] < b[1], 1, -1), -1))), 1, 1), x -> x[1]))";
+    }
+
+    @SqlInvokedScalarFunction(value = "array_least_frequent", deterministic = true, calledOnNullInput = true)
+    @Description("Determines the n least frequent elemement in the array in the descending order of the elements.")
+    @TypeParameter("T")
+    @SqlParameters({@SqlParameter(name = "input", type = "array(T)"), @SqlParameter(name = "n", type = "bigint")})
+    @SqlType("array<T>")
+    public static String array_n_least_frequent()
```
kaikalur, line 109: To be consistent with others like MAP_TOP_N etc, just make _N a suffix (we should also see if we can overload the function later).

presto-main/src/main/java/com/facebook/presto/operator/scalar/sql/ArraySqlFunctions.java
```diff
@@ -91,6 +91,26 @@ public static String arrayHasDuplicatesVarchar()
...
     }
 
+    @SqlInvokedScalarFunction(value = "array_least_frequent", deterministic = true, calledOnNullInput = true)
+    @Description("Determines the least frequent elemement in the array. If there are multiple elements, the function returns the biggest element")
+    @TypeParameter("T")
+    @SqlParameter(name = "input", type = "array(T)")
+    @SqlType("array<T>")
+    public static String array_least_frequent()
+    {
+        return "RETURN IF(COALESCE(CARDINALITY(REMOVE_NULLS(input)), 0) = 0, NULL, TRANSFORM(SLICE(ARRAY_SORT(MAP_ENTRIES(ARRAY_FREQUENCY(REMOVE_NULLS(input))),(a, b) -> IF(a[2] > b[2], 1, IF(a[2] = b[2], IF(a[1] < b[1], 1, -1), -1))), 1, 1), x -> x[1]))";
```
kaikalur, line 101: Did you try transposing and sorting?

array_sort(TRANSFORM(MAP_ENTRIES(m),(k,v)->(v,k)))

Then you don't need the lambda for the comparison

kaikalur (changes requested):

presto-main/src/main/java/com/facebook/presto/operator/scalar/sql/ArraySqlFunctions.java
kaikalur: yeah it will be good to do a sqlbenchmark to see which is best.

jainavi17: Ran the banchmark test. Turned out the non lambda version is 0.7 secs faster than compared to the one with lambda. 

Here is the result without lambda
Screenshot 2023-10-08 at 14 41 20

Here is the result with lambda
Screenshot 2023-10-08 at 14 30 29

kaikalur (approved):

NikhilCollooru (approved):

kaikalur:
[like]  Sreeni Viswanadha reacted to your message:

PR #20627, opened 2023-08-22
Add heartbeat manager

pranjalssh:
## Description
Add a hearbeat manager in C++ which sends regular heartbeats to presto server. This is needed when we use multi coordinator setup.

Extracted common changes from Announcer and HeartbeatManager into PeriodicServiceInventoryManager. Similar naming scheme is used in Java.

Also removed announcement-min-frequency-ms. Name of the config suggests we can make announcements with this frequency - but that is not the case. This is simply a minimum time we wait before retrying failed requests - which can be a in-code constant - as many other libraries do.

Added a config for heartbeat sending.

## Test Plan
Ran changes in a test cluster

## Contributor checklist

- [x] Please make sure your submission complies with our [development](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#development), [formatting](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#formatting), [commit message](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#commit-formatting-and-pull-requests), and [attribution guidelines](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#attribution).
- [x] PR description addresses the issue accurately and concisely.  If the change is non-trivial, a GitHub Issue is referenced.
- [x] Documented new properties (with its default value), SQL syntax, functions, or other functionality.
- [x] If release notes are required, they follow the [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines).
- [x] Adequate tests were added if applicable.
- [x] CI passed.

## Release Notes
Please follow [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines) and fill in the release notes below.

```
== NO RELEASE NOTE ==
```

PR #20754, opened 2023-09-04
[native] Fix announcment retry.

amitkdutta:
We only consider http status xxx as success, and http status kHttpNoContent should be considered as a failure. This is a regression introduced by recent [change PR](https://github.com/prestodb/presto/pull/20627). We are reverting it here.

### DIFF
--- a/presto-native-execution/presto_cpp/main/PeriodicServiceInventoryManager.cpp
+++ b/presto-native-execution/presto_cpp/main/PeriodicServiceInventoryManager.cpp
@@ -85,8 +85,7 @@ void PeriodicServiceInventoryManager::sendRequest() {
       .via(eventBaseThread_.getEventBase())
       .thenValue([this](auto response) {
         auto message = response->headers();
-        if (message->getStatusCode() != http::kHttpAccepted ||
-            message->getStatusCode() != http::kHttpNoContent) {
+        if (message->getStatusCode() != http::kHttpAccepted) {
           ++failedAttempts_;
           LOG(WARNING) << id_ << " failed: HTTP " << message->getStatusCode()
                        << " - " << response->dumpBodyChain();


tanjialiang:
Thanks for the fix. Could we have a class description on top of this PeriodicServiceInventoryManager as its public. Also can we add tests for this class? It seems we don't have any tests for this base class. It will be necessary to have tests around it to  guard regression like the fix itself.

presto-native-execution/presto_cpp/main/PeriodicServiceInventoryManager.cpp
```diff
@@ -85,8 +85,7 @@ void PeriodicServiceInventoryManager::sendRequest() {
       .via(eventBaseThread_.getEventBase())
       .thenValue([this](auto response) {
         auto message = response->headers();
-        if (message->getStatusCode() != http::kHttpAccepted ||
-            message->getStatusCode() != http::kHttpNoContent) {
+        if (message->getStatusCode() != http::kHttpAccepted) {
```
xiaoxmeng, line 88: Current code allows kHttpNoContent? This change makes kHttpNoContent as a failure case? Is it what we want @amitkdutta? thanks!

amitkdutta: When the statuscode is 202 (OK), response has no content. Then the logic `message->getStatusCode() != http::kHttpAccepted || message->getStatusCode() != http::kHttpNoContent` fails and announcement is treated as failure.

Check my comment here https://github.com/prestodb/presto/pull/20627#pullrequestreview-1609907289
Basically originally announcement was testing statuscode. I think content checking might be needed for Heartbeatmanagement for multi coordinator, but for general announcement (single coordinator case) its not needed. This PR revert the condition to old state to keep it working.

CC: @pranjalssh

xiaoxmeng: "message->getStatusCode() == http::kHttpNoContent" is treated as a failure case now so this is expected? What's the case in code for "When the statuscode is 202 (OK), response has no content."? The comment just makes me a bit confusing. It seems to me that we want to treat no content as a failure case?

xiaoxmeng (approved):
@amitkdutta accepted to unblock as this is a rollback to previous version. Consider to make the comment clear. Thanks!

presto-native-execution/presto_cpp/main/PeriodicServiceInventoryManager.cpp
amitkdutta: @xiaoxmeng Yea, so basically only status code OK should be considered as success. Anything else should be considered as failure. In current code, `kHttpNoContent` is considered as success also. We are reverting that change here.

PR #20755, opened 2023-09-04
[native] Remove spill console logging in periodic task manager.

amitkdutta:
Counters are exported for spilling, no need to log stats in
console periodically as the periodic task is very chatty and
many times server has empty stats. Keeping the logs clean by
removing redundant spill status printing.

### DIFF
--- a/presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
+++ b/presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
@@ -567,7 +567,6 @@ void PeriodicTaskManager::updateSpillStatsTask() {
       kCounterSpillFlushTimeUs, deltaSpillStats.spillFlushTimeUs);
   REPORT_IF_NOT_ZERO(
       kCounterSpillWriteTimeUs, deltaSpillStats.spillWriteTimeUs);
-  LOG(INFO) << "Spill Stats:\n" << deltaSpillStats.toString() << "\n";
   lastSpillStats_ = updatedSpillStats;
 }
 


presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
```diff
@@ -567,7 +567,6 @@ void PeriodicTaskManager::updateSpillStatsTask() {
       kCounterSpillFlushTimeUs, deltaSpillStats.spillFlushTimeUs);
   REPORT_IF_NOT_ZERO(
       kCounterSpillWriteTimeUs, deltaSpillStats.spillWriteTimeUs);
-  LOG(INFO) << "Spill Stats:\n" << deltaSpillStats.toString() << "\n";
```
xiaoxmeng, line 570: This is used for Sapphire as there is no monitoring in that case. The spill stats is updated every 60 seconds. You can add an empty method to skip logging if it is empty
```
if (!deltaSpillStats.empty()) {
   LOG(INFO) << "Spill Stats:\n" << deltaSpillStats.toString() << "\n";
}
```

xiaoxmeng (commented):
@amitkdutta thanks % comment.

presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
amitkdutta: @xiaoxmeng Can we consider `spilledBytes` to be zero as an indication of empty stats [here](https://github.com/facebookincubator/velox/blob/90c87d1310b9ae0ec36eb168a2a9ddcc059e467b/velox/exec/Spill.h#L145)?

xiaoxmeng: @amitkdutta yeah, that should work. I will make a change at Velox side to ease the handling here. Btw, shall we add a type or flag thing to tell if it is for Prestissimo or Spark? Then we could skip logging for Prestissimo next if you think logging is redundant for Prestissimo. But I guess if logging is not too chatty, it might be good for us have that log for Prestissimo? Thanks!

xiaoxmeng (approved):
@amitkdutta Thanks!

presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
tanjialiang: It is logged every 60 seconds. It should be good.

PR #20650, opened 2023-08-24
[native] Add node.internal-address to support hostname and FQDN

nmahadevuni:
## Description
Add node.internal-address to support hostname and FQDN.
This aligns with Presto Java config. 

## Motivation and Context
We can only provide IP address now currently in `node.ip` config. When we use secure TLS communication 
between internal nodes, if the certificate provided doesn't have the IP address in the Subject Alternative Names,
then native worker cannot establish communication with the coordinator. Hence, the need to support
 hostname/FQDN too and the need to change the config name also to `node.internal-address` which 
also matches the Presto Java config property name.

## Impact

No public API changes. `node.ip` is supported as legacy config and it will continue to work. 
No performance impacting changes.

## Test Plan
Added a test in SystemConfigTest.cpp to test the new config.

```
== RELEASE NOTES ==

General Changes
* Node config property `node.ip` is now a legacy config. It is replaced with `node.internal-address` to support hostname/FQDN too.
```

PR #20757, opened 2023-09-04
[native] Fix exchange client creation.

amitkdutta:
## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* ...
* ...

Hive Changes
* ...
* ...
```

```
== NO RELEASE NOTE ==
```


mshang816 (approved):

nmahadevuni:
@amitkdutta I had to make this change to have hostnames also in `node.internal-address` config. For certificate based authentication, in certificates, IP addresses won't be used, only hostnames and wildcards. So, we need this change. Also, the SocketAddress constructor documentation mentions so
`* @param host The IP address (or hostname, if allowNameLookup is true)`
So, if allowNameLookup is true, the host argument can be a hostname or IP address. If it fails, then it is probably a bug in SocketAddress implementation?

amitkdutta:
@nmahadevuni I had to revert to keep the master branch in working condition.  All queries were failing, hence the revert. I commented on the main PR also https://github.com/prestodb/presto/pull/20650.

Since passing IP address, can use the fast path, we should definitely keep it and don't use the resolution thing if IP address is already in hand.

If IP address is not there, then as you mentioned we should use the resolution and pass the host name. We can make Exchange client have two constructor, one taking IP and port and another host port. That will be very clean and we can change the client creation here https://github.com/prestodb/presto/blob/9c7c230453e86ea1f7a62d184c5d05c9fa49248a/presto-native-execution/presto_cpp/main/PrestoExchangeSource.cpp#L408 (similar to what was done for https). What do you think @nmahadevuni  @majetideepak ?

PR #20768, opened 2023-09-05
[native] Treat http 204 as announcement success

pranjalssh:
Followup of https://github.com/prestodb/presto/pull/20754

Previously we used OR instead of AND, which made all successful announcements be logged as failed. 
```
== NO RELEASE NOTE ==
```


amitkdutta (approved):
Thanks @pranjalssh

PR #20776, opened 2023-09-05
[Iceberg] Add support for TRUNCATE

ZacBlanco:
## Description

This commit adds support for deleting records from iceberg tables using the query `TRUNCATE TABLE <table>;`

## Motivation and Context

This allows removing data from Iceberg tables without having to drop the table from the catalog.

## Impact

Additional syntax support.

## Test Plan

Test cases:
- `TRUNCATE TABLE x` with a table that has no rows
- `TRUNCATE TABLE x` on a table with records
- Truncate and insert on the same table multiple times back to back

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* The Iceberg connector now supports `TRUNCATE TABLE <table>` statements
```

### DIFF
--- a/presto-docs/src/main/sphinx/connector/iceberg.rst
+++ b/presto-docs/src/main/sphinx/connector/iceberg.rst
@@ -479,6 +479,30 @@ Alter table operations are supported in the connector::
 
      ALTER TABLE iceberg.web.page_views DROP COLUMN location;
 
+TRUNCATE
+^^^^^^^^
+
+The iceberg connector can delete all of the data from tables without
+dropping the table from the metadata catalog using ``TRUNCATE TABLE``.
+
+.. code-block:: sql
+
+    TRUNCATE TABLE nation;
+
+.. code-block:: text
+
+    TRUNCATE TABLE;
+
+.. code-block:: sql
+
+    SELECT * FROM nation;
+
+.. code-block:: text
+
+     nationkey | name | regionkey | comment
+    -----------+------+-----------+---------
+    (0 rows)
+
 DROP TABLE
 ^^^^^^^^^^^
 
@@ -674,6 +698,5 @@ exists as we've rolled back to the previous state.
 Iceberg Connector Limitations
 -----------------------------
 
-* :doc:`/sql/delete` is only supported if the ``WHERE`` clause matches entire partitions.
 * The ``SELECT`` operations on Iceberg Tables with format version 2 do not read the delete files
   and remove the deleted rows as of now (:issue:`20492`).
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
@@ -46,15 +46,19 @@ import io.airlift.slice.Slice;
 import org.apache.iceberg.AppendFiles;
 import org.apache.iceberg.BaseTable;
 import org.apache.iceberg.DataFiles;
+import org.apache.iceberg.DeleteFiles;
+import org.apache.iceberg.FileScanTask;
 import org.apache.iceberg.PartitionSpecParser;
 import org.apache.iceberg.Schema;
 import org.apache.iceberg.SchemaParser;
 import org.apache.iceberg.Table;
 import org.apache.iceberg.Transaction;
+import org.apache.iceberg.io.CloseableIterable;
 import org.apache.iceberg.types.Type;
 import org.apache.iceberg.types.TypeUtil;
 import org.apache.iceberg.types.Types;
 
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
@@ -63,6 +67,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
 
 import static com.facebook.presto.common.type.BigintType.BIGINT;
 import static com.facebook.presto.iceberg.IcebergColumnHandle.primitiveIcebergColumnHandle;
@@ -79,6 +84,7 @@ import static com.facebook.presto.iceberg.PartitionFields.toPartitionFields;
 import static com.facebook.presto.iceberg.TableType.DATA;
 import static com.facebook.presto.iceberg.TypeConverter.toIcebergType;
 import static com.facebook.presto.iceberg.TypeConverter.toPrestoType;
+import static com.facebook.presto.spi.StandardErrorCode.GENERIC_INTERNAL_ERROR;
 import static com.facebook.presto.spi.StandardErrorCode.NOT_SUPPORTED;
 import static com.google.common.base.Verify.verify;
 import static com.google.common.collect.ImmutableList.toImmutableList;
@@ -277,12 +283,6 @@ public abstract class IcebergAbstractMetadata
         return primitiveIcebergColumnHandle(0, "$row_id", BIGINT, Optional.empty());
     }
 
-    @Override
-    public ConnectorTableHandle beginDelete(ConnectorSession session, ConnectorTableHandle tableHandle)
-    {
-        throw new PrestoException(NOT_SUPPORTED, "This connector only supports delete where one or more partitions are deleted entirely");
-    }
-
     protected List<ColumnMetadata> getColumnMetadatas(Table table)
     {
         return table.schema().columns().stream()
@@ -433,5 +433,37 @@ public abstract class IcebergAbstractMetadata
         }
 
         return getIcebergSystemTable(tableName, icebergTable);
+    }
+
+    /**
+     * Deletes all the files within a particular scan
+     *
+     * @return the number of rows deleted from all files
+     */
+    private long removeScanFiles(Table icebergTable, Iterable<FileScanTask> scan)
+    {
+        transaction = icebergTable.newTransaction();
+        DeleteFiles deletes = transaction.newDelete();
+        AtomicLong rowsDeleted = new AtomicLong(0L);
+        scan.forEach(t -> {
+            deletes.deleteFile(t.file());
+            rowsDeleted.addAndGet(t.estimatedRowsCount());
+        });
+        deletes.commit();
+        transaction.commitTransaction();
+        return rowsDeleted.get();
+    }
+
+    @Override
+    public void truncateTable(ConnectorSession session, ConnectorTableHandle tableHandle)
+    {
+        IcebergTableHandle handle = (IcebergTableHandle) tableHandle;
+        Table icebergTable = getIcebergTable(session, handle.getSchemaTableName());
+        try (CloseableIterable<FileScanTask> files = icebergTable.newScan().planFiles()) {
+            removeScanFiles(icebergTable, files);
+        }
+        catch (IOException e) {
+            throw new PrestoException(GENERIC_INTERNAL_ERROR, "failed to scan files for delete", e);
+        }
     }
 }
--- a/presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedTestBase.java
+++ b/presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedTestBase.java
@@ -17,11 +17,15 @@ import com.facebook.presto.testing.MaterializedResult;
 import com.facebook.presto.testing.QueryRunner;
 import com.facebook.presto.tests.AbstractTestDistributedQueries;
 import com.google.common.collect.ImmutableMap;
+import org.intellij.lang.annotations.Language;
 import org.testng.annotations.Test;
 
 import static com.facebook.presto.common.type.VarcharType.VARCHAR;
 import static com.facebook.presto.testing.MaterializedResult.resultBuilder;
+import static com.facebook.presto.testing.TestingAccessControlManager.TestingPrivilegeType.SELECT_COLUMN;
+import static com.facebook.presto.testing.TestingAccessControlManager.privilege;
 import static com.facebook.presto.testing.assertions.Assert.assertEquals;
+import static java.lang.String.format;
 import static java.util.Objects.requireNonNull;
 import static org.testng.Assert.assertTrue;
 
@@ -69,6 +73,38 @@ public class IcebergDistributedTestBase
     {
     }
 
+    @Test
+    public void testTruncate()
+    {
+        // Test truncate empty table
+        assertUpdate("CREATE TABLE test_truncate_empty (i int)");
+        assertQuerySucceeds("TRUNCATE TABLE test_truncate_empty");
+        assertEquals(getQueryRunner().execute("SELECT count(*) FROM test_truncate_empty").getOnlyValue(), 0L);
+        assertQuerySucceeds("DROP TABLE test_truncate_empty");
+
+        // Test truncate table with rows
+        assertUpdate("CREATE TABLE test_truncate AS SELECT * FROM orders", "SELECT count(*) FROM orders");
+        assertQuerySucceeds("TRUNCATE TABLE test_truncate");
+        MaterializedResult result = getQueryRunner().execute("SELECT count(*) FROM test_truncate");
+        assertEquals(result.getOnlyValue(), 0L);
+        assertUpdate("DROP TABLE test_truncate");
+
+        // test truncate -> insert -> truncate
+        assertUpdate("CREATE TABLE test_truncate_empty (i int)");
+        assertQuerySucceeds("TRUNCATE TABLE test_truncate_empty");
+        assertEquals(getQueryRunner().execute("SELECT count(*) FROM test_truncate_empty").getOnlyValue(), 0L);
+        assertUpdate("INSERT INTO test_truncate_empty VALUES 1", 1);
+        assertEquals(getQueryRunner().execute("SELECT count(*) FROM test_truncate_empty").getOnlyValue(), 1L);
+        assertQuerySucceeds("TRUNCATE TABLE test_truncate_empty");
+        assertEquals(getQueryRunner().execute("SELECT count(*) FROM test_truncate_empty").getOnlyValue(), 0L);
+        assertQuerySucceeds("DROP TABLE test_truncate_empty");
+
+        // Test truncate access control
+        assertUpdate("CREATE TABLE test_truncate AS SELECT * FROM orders", "SELECT count(*) FROM orders");
+        assertAccessAllowed("TRUNCATE TABLE test_truncate", privilege("orders", SELECT_COLUMN));
+        assertUpdate("DROP TABLE test_truncate");
+    }
+
     @Override
     public void testShowColumns()
     {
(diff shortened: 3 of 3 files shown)


ZacBlanco:
I will squash the commits once review is finished

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
```diff
@@ -264,9 +276,35 @@ public ColumnHandle getUpdateRowIdColumnHandle(ConnectorSession session, Connect
...
+        return tableLayoutHandle.map(x -> ((IcebergTableLayoutHandle) x).getTupleDomain().equals(TupleDomain.all())).orElse(true);
+    }
+
+    @Override
+    public OptionalLong metadataDelete(ConnectorSession session, ConnectorTableHandle tableHandle, ConnectorTableLayoutHandle tableLayoutHandle)
     {
-        throw new PrestoException(NOT_SUPPORTED, "This connector only supports delete where one or more partitions are deleted entirely");
+        IcebergTableLayoutHandle layoutHandle = (IcebergTableLayoutHandle) tableLayoutHandle;
+        if (!layoutHandle.getTupleDomain().equals(TupleDomain.all())) {
+            throw new PrestoException(INVALID_ARGUMENTS, "iceberg metadata delete doesn't support predicates");
```
ClarenceThreepwood, line 289: Is this reachable given the check in the previous function?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/Partition.java
```diff
@@ -235,7 +235,9 @@ public static Map<Integer, Object> toMap(Map<Integer, Type.PrimitiveType> idToTy
         ImmutableMap.Builder<Integer, Object> map = ImmutableMap.builder();
         idToMetricMap.forEach((id, value) -> {
             Type.PrimitiveType type = idToTypeMapping.get(id);
-            map.put(id, Conversions.fromByteBuffer(type, value));
+            if (type != null) {
```
ClarenceThreepwood, line 238: Just for my edification - When can this happen? Should it be an explicit error instead of ignoring the missing type?

ClarenceThreepwood (approved):

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
ZacBlanco: I couldn't generate a test case where this exception would be thrown. However while I assumed that `metadataDelete` would only be called when `supportsMetadataDelete` returns true, I don't actually know the contract between this method and `supportsMetadataDelete`. Is it ever possible it could change in the future. I added this conditional in as a defensive measure in case this was ever called. I suppose instead I could change this condition to just `if (!supportsMetadataDelete(session, tableHandle, Optional.of(layoutHandle))) {` instead to consolidate the logic in the case we want to keep this defensive check

If you think it is unnecessary, then I am not really against removing it

presto-iceberg/src/main/java/com/facebook/presto/iceberg/Partition.java
ZacBlanco: This occurs when the `idToMetricMap` contains an entry for a non-primitive type, since the `idToTypeMapping` only contains mappings for primitive types (e.g. we were using a lot of functions with `row` types). I'm not sure the cases when this actually occurs, but I had to put this check in due to some NPEs when running queries with some of the sample tables.

IIRC it may be that this change might not be explicitly required for this PR. I think it got chucked in to this commit by accident. If not included here I will probably include it in another PR down the road.

ClarenceThreepwood: I'd suggest removing for now. And then add it later with a test case that triggers this

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
ClarenceThreepwood: I think it is low overhead and ok to leave it. Yes - consolidating the two checks makes sense

presto-iceberg/src/main/java/com/facebook/presto/iceberg/Partition.java
ZacBlanco: fixed

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
ZacBlanco: fixed

agrawalreetika:
Hi @ZacBlanco
Wanted to confirm, does this also add support for DELETE if the WHERE clause matches entire partitions?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
```diff
@@ -152,6 +162,8 @@ public ConnectorTableMetadata getTableMetadata(ConnectorSession session, Connect
 
     protected abstract ConnectorTableMetadata getTableMetadata(ConnectorSession session, SchemaTableName table);
 
+    protected abstract Table getIcebergTable(ConnectorSession session, SchemaTableName schemaTableName);
```
agrawalreetika, line 165: I think if we are introducing  `getIcebergTable` then we can add this in the separate commit and make of use it even in the existing place where `getHiveIcebergTable` or `getNativeIcebergTable` is being called?

ZacBlanco:
> Hi @ZacBlanco Wanted to confirm, Does this also add support for DELETE if the WHERE clause matches entire partitions or do you know if that's already supported?

No, this does not support any kind of `WHERE` clause currently

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
ZacBlanco: Sure, I can remove this for now. I will open another PR for the refactor later.

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestDistributedQueries.java
```diff
@@ -361,7 +361,7 @@ public void testExplainAnalyzeDDL()
         computeActual("EXPLAIN ANALYZE DROP TABLE orders");
     }
 
-    private void assertExplainAnalyze(@Language("SQL") String query)
+    protected void assertExplainAnalyze(@Language("SQL") String query)
```
ChunxuTang, line 364: Why would you need to change it from `private` to `protected`? Could you run the tests without this change?

ChunxuTang (changes requested):
Hi @ZacBlanco, thanks for the contribution. I left a few comments below.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergNativeMetadata.java
```diff
@@ -332,4 +340,30 @@ public TableStatistics getTableStatistics(ConnectorSession session, ConnectorTab
...
+            });
+        }
+        catch (IOException e) {
+            throw new PrestoException(GENERIC_INTERNAL_ERROR, "failed to scan files for delete", e);
+        }
+        deletes.commit();
+        transaction.commitTransaction();
+        return OptionalLong.of(rowsDeleted.get());
+    }
 }
```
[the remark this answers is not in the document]
ZacBlanco: Hey @ChunxuTang, I would have combined these but for now I plan to leave them separate. I have another PR open which actually combines a lot of logic from both implementations. If this PR gets in first I will reduce the implementation into a single method in the other PR: https://github.com/prestodb/presto/pull/20856

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestDistributedQueries.java
ZacBlanco: I updated this so I could call this method in the test. e.g. here: https://github.com/prestodb/presto/pull/20776/files#diff-e86735ad8b4b0dd7856263d9696e731dd986026f5547b8f78265e1e389a308daR99

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergNativeMetadata.java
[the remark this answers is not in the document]
ZacBlanco: Originally, there was a single implementation, but changed it due to the review comment here: https://github.com/prestodb/presto/pull/20776#discussion_r1324004195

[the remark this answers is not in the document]
ChunxuTang: I noticed that https://github.com/prestodb/presto/pull/20856 has been merged. Would you adjust this PR accordingly?

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestDistributedQueries.java
ChunxuTang: I assume you changed the visibility because this PR only partially supports the base `testDelete` function. I don't suggest updating the visibility due to partial support of the deletion feature.
Would you happen to have a plan to support other deletion syntaxes?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergNativeMetadata.java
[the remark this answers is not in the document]
ZacBlanco: This PR has been updated to include a single implementation now 😄

agrawalreetika:
@ZacBlanco Can we also add the documentation for this along with few examples and limitations?

ZacBlanco:
> @ZacBlanco Can we also add the documentation for this along with a few examples and limitations? And also could you please squash the commit as well?

Fixed. I've also squashed all the commits

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -671,9 +671,38 @@ exists as we've rolled back to the previous state.
     -----------+------+-----------+---------
     (0 rows)
 
+DELETION
```
steveburnett, line 674: Looks great! 

One question: 
Would it be appropriate to consider naming this new content DELETE, and move it from the end of the page into the SQL Support heading as a subheading between ALTER TABLE and DROP TABLE?

ZacBlanco: Thanks for the quick review! I agree that your suggestion for the placement of the section is better. I've fixed it.

steveburnett (approved):
LGTM! (docs)

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestDistributedQueries.java
ChunxuTang: Because the function was designed to be `private`. If this PR supports the full `testDelete` function, we probably wouldn't need to update the visibility as we could just call the base `testDelete` function, right?

ChunxuTang: That's also why I asked about your plan to support other deletion syntaxes.

ZacBlanco: The plan for supporting the rest of the delete syntax is going to be up to the Presto's iceberg contributors. The next step is probably to support whole-partition deletes, but I don't have the intention of including that in this PR. 

I've changed the visibility back on this method

agrawalreetika (approved):
LGTM 👍🏻  Please review the test failures.

ChunxuTang:
@ZacBlanco Thanks for your work and the clarification on your plan to support deletion.

Hi @tdcmeehan and @beinan, would you mind sharing your thoughts on this PR? It represents an initial step towards adding support for deletion in the Iceberg connector. However, from the discussion with @ZacBlanco and from my knowledge, it appears that we currently don't have a concrete plan to support the complete set of deletion syntax, but rather a preliminary feature that covers "DELETE FROM <table>" statements without predicates. This means that customers may have access to an incomplete set of deletion syntax for an extended period.

I'm unsure whether we should proceed with delivering this partial deletion feature, allowing customers to use at least the basic functionality, or if it would be better to postpone the delivery until we have a clearer roadmap for full deletion syntax support. Any suggestions will be appreciated.

tdcmeehan:
Why not simply implement `TRUNCATE` rather than `DELETE`?

tdcmeehan:
I think it's worthwhile to add support for TRUNCATE, since that's semantically more clear.

@ChunxuTang : I think a logical series of steps would be first support truncation, then deletion by partition (works in v1 and v2), then row level deletion (works only with v2 tables).  We can add issues and track in our Iceberg project board the latter two issues.  WDYT?

ZacBlanco:
The commit, description, and PR title have all been updated to indicate truncate is implemented instead

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -674,6 +701,6 @@ exists as we've rolled back to the previous state.
 Iceberg Connector Limitations
 -----------------------------
 
-* :doc:`/sql/delete` is only supported if the ``WHERE`` clause matches entire partitions.
```
tdcmeehan, line 677: I suppose the old text was just wrong?

ZacBlanco: I just fixed this now -- but yes it was wrong

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -479,6 +479,30 @@ Alter table operations are supported in the connector::
...
+^^^^^^^^
+
+The iceberg connector can delete all of the data data from tables without
+dropping the table from the metadata catalog using ``TRUNCATE TABLE``.
+
+.. code-block:: sql
+
+    TRUNCATE TABLE nation;
+
+.. code-block:: text
```
tdcmeehan, line 492:
```suggestion
.. code-block:: sql
```

tdcmeehan (approved):

presto-docs/src/main/sphinx/connector/iceberg.rst
ZacBlanco: This is actually the output from the CLI after running `TRUNCATE TABLE nation`, so I think keeping `text` here is ok.

tdcmeehan: Ah, got it!

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -479,6 +479,30 @@ Alter table operations are supported in the connector::
 
      ALTER TABLE iceberg.web.page_views DROP COLUMN location;
 
+TRUNCATE
+^^^^^^^^
+
+The iceberg connector can delete all of the data data from tables without
```
steveburnett, line 485:
```suggestion
The iceberg connector can delete all of the data from tables without
```

steveburnett (changes requested):
Just one thing - "data data" seems unintentional. Yes?

presto-docs/src/main/sphinx/connector/iceberg.rst
ZacBlanco: Fixed, thanks for pointing out the typo!

steveburnett (approved):
LGTM! (docs)

New build of doc, typo mentioned in previous review verified as fixed. Thanks @ZacBlanco!

tdcmeehan:
@ChunxuTang / @beinan any further concerns?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
```diff
@@ -434,4 +434,36 @@ public Optional<SystemTable> getSystemTable(ConnectorSession session, SchemaTabl
 
         return getIcebergSystemTable(tableName, icebergTable);
     }
+
+    /**
+     * Deletes the files from a particular scan
+     *
+     * @return
+     */
```
ChunxuTang, line 442: The comment here looks unnecessary. Could you remove it?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
```diff
@@ -434,4 +434,36 @@ public Optional<SystemTable> getSystemTable(ConnectorSession session, SchemaTabl
...
+    private long deleteTable(Table icebergTable, Iterable<FileScanTask> scan)
```
ChunxuTang, line 443: With the new `truncateTable` function, the name of this function (`deleteTable`) looks a bit misleading. How about `deleteFiles`? Or any other ideas?

ChunxuTang (commented):
@tdcmeehan Thanks for your suggestion. Renaming the commit to support `TRUNCATE` looks nice to me.

@ZacBlanco Thanks for your consistent and nice contributions. I left a few comments. Would you mind taking a look?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
ZacBlanco: I kept the comment but updated it to properly reflect the functionality

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
ZacBlanco: I've updated the name to `removeScanFiles`

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
```diff
@@ -277,12 +283,6 @@ public ColumnHandle getUpdateRowIdColumnHandle(ConnectorSession session, Connect
         return primitiveIcebergColumnHandle(0, "$row_id", BIGINT, Optional.empty());
     }
 
-    @Override
-    public ConnectorTableHandle beginDelete(ConnectorSession session, ConnectorTableHandle tableHandle)
-    {
-        throw new PrestoException(NOT_SUPPORTED, "This connector only supports delete where one or more partitions are deleted entirely");
-    }
-
```
ChunxuTang, line 285: Quick question: As this PR turns to focus on supporting `TRUNCATE`, do we still need to remove this function?

ChunxuTang (approved):
@ZacBlanco Thanks for the work.

Generally lgtm. Just left a minor comment.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
ZacBlanco: This method is incorrectly implemented and misdirects users. If someone were to query the connector `DELETE from <table> WHERE <predicate>;` the error would say "This connector only supports delete where one or more partitions are deleted entirely".

However, that statement is incorrect and misdirects users as the connector never supported deleting partitions. This should be removed until the support is added.

PR #20779, opened 2023-09-06
[native] Skip redundant Ip object to string conversion.

amitkdutta:
SocketAddress constructor can take IPAddress object. Current
code converts a IP address string to IPAddress object and then
converts it back to string. Skipping the redundant conersion here.


xiaoxmeng (approved):
@amitkdutta LGTM. Thanks!

PR #20794, opened 2023-09-07
[native] Fix exchange client to use IP or hostname for destination

nmahadevuni:
## Description
When creating PrestoExchangeSource client, currently only IP address is expected to be present in the destination URI. Fixing it to use hostname too.

## Motivation and Context
Native worker's PrestoExchangeSource expects only IP address in the destination URI and fails if a hostname is present. Certificate based authentication generally has hostnames as Subject Alternative Names. When we use this, in Presto and native worker, we use `node.internal-address` config to set the hostname. When we set the hostname in this config, the PrestoExchangeSource fails since it won't be able to resolve hostname. 

## Impact
No public facing changes and performance impact

## Test Plan
Tested manually that with both hostname and IPAddress, the client creation works successfully.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/PrestoExchangeSource.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoExchangeSource.cpp
@@ -83,7 +83,13 @@ PrestoExchangeSource::PrestoExchangeSource(
       clientCertAndKeyPath_(clientCertAndKeyPath),
       ciphers_(ciphers),
       exchangeExecutor_(executor) {
-  folly::SocketAddress address(folly::IPAddress(host_), port_);
+  folly::SocketAddress address;
+  // If host_ is an IP address, set address from it, else resolve host_
+  if (folly::IPAddress::validate(host_)) {
+    address.setFromIpPort(host_, port_);
+  } else {
+    address.setFromHostPort(host_, port_);
+  }
   auto timeoutMs = std::chrono::duration_cast<std::chrono::milliseconds>(
       SystemConfig::instance()->exchangeRequestTimeout());
   VELOX_CHECK_NOT_NULL(exchangeExecutor_.get());


nmahadevuni:
@amitkdutta Can you review this? Enhances fix for #20757

presto-native-execution/presto_cpp/main/PrestoExchangeSource.cpp
```diff
@@ -83,7 +83,13 @@ PrestoExchangeSource::PrestoExchangeSource(
       clientCertAndKeyPath_(clientCertAndKeyPath),
       ciphers_(ciphers),
       exchangeExecutor_(executor) {
-  folly::SocketAddress address(folly::IPAddress(host_), port_);
+  folly::SocketAddress address;
```
amitkdutta, line 86: @nmahadevuni Thanks for looking in this. I tried this and still seeing from here:

https://github.com/facebook/folly/blob/main/folly/SocketAddress.cpp#L634-L639

We will need to keep the code path where an IP address object is directly specified. For hostname, lets create a different constructor. Basically when node.ip is provided, we don't need to verify if its a host or ip, it will use the constructor which takes IPAdress. When node.internal-address is provided, it can be host or IP. Then we can construct PrestoExchange like you did earlier.

nmahadevuni:
Closing this since @amitkdutta has fixed it in https://github.com/prestodb/presto/pull/20804

amitkdutta:
Thanks @nmahadevuni I tested https://github.com/prestodb/presto/pull/20804 with both configurations where hardware has multiple IP for a hostname. Keeping the fast path, and based on your code switching to host look up (instead of calling sockets internal method, which still fallbacks to look up even if IP address is given :( ).

PR #20797, opened 2023-09-07
[native] Move testDereference to a general native test

kevinwilfong:
testDereference was added as a Sapphire-specific test, but it's testing general native behavior, so it would be better placed in a class shared by Sapphire and Prestissimo.


mbasmanova (approved):
Thanks.

PR #20806, opened 2023-09-08
Enforce JWT authentication without bearer token

tdcmeehan:
## Description
This fixes a flaw in the JWT authentication for internal communication.

## Motivation and Context
Secure communication.

## Impact
This fixes a problem with secure internal communication with JWT--without it, it's possible to authenticate successfully by simply sending no bearer token when no other authenticators are configured.

## Test Plan
Unit tests were added. 

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Fix a security gap in the JWT authentication for internal communication.

```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/server/InternalAuthenticationManager.java
+++ b/presto-main/src/main/java/com/facebook/presto/server/InternalAuthenticationManager.java
@@ -41,7 +41,7 @@ public class InternalAuthenticationManager
 {
     private static final Logger log = Logger.get(InternalAuthenticationManager.class);
 
-    private static final String PRESTO_INTERNAL_BEARER = "X-Presto-Internal-Bearer";
+    public static final String PRESTO_INTERNAL_BEARER = "X-Presto-Internal-Bearer";
 
     private final boolean internalJwtEnabled;
     private final byte[] hmac;
--- a/presto-main/src/main/java/com/facebook/presto/server/security/AuthenticationFilter.java
+++ b/presto-main/src/main/java/com/facebook/presto/server/security/AuthenticationFilter.java
@@ -16,6 +16,7 @@ package com.facebook.presto.server.security;
 import com.facebook.airlift.http.server.AuthenticationException;
 import com.facebook.airlift.http.server.Authenticator;
 import com.facebook.presto.server.InternalAuthenticationManager;
+import com.facebook.presto.server.InternalCommunicationConfig;
 import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableList;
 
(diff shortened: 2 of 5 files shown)


majetideepak (approved):

czentgr (approved):

presto-main/src/main/java/com/facebook/presto/server/security/AuthenticationFilter.java
```diff
@@ -122,7 +137,7 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
 
     private boolean doesRequestSupportAuthentication(HttpServletRequest request)
     {
-        if (!authenticators.isEmpty() && request.isSecure()) {
+        if ((!authenticators.isEmpty() || internalJwtEnabled) && request.isSecure()) {
```
majetideepak, line 140: Can the request be `not isSecure` and JWT enabled?
In that case, we will still have the issue.

tdcmeehan: If the connection is not secure, then I would think we should just presume authentication can be spoofed.

czentgr: It still would take efforts to try a replay attack or try to get the secret to spoof a legit JWT.
In the C++ code we added a new option to define how long a JWT is valid. in Java it is hard coded to 5 mins which is likely too long. Either way, if JWT is enabled, I believe we should still give it a try and not let any request through.
Let's make it harder to exploit HTTP as well instead of opening the doors. Maybe I'm too optimistic and it is trivial from an attackers perspective.

tdcmeehan: As discussed, I don't think this PR should change the scope of this behavior for all authenticators, and it probably deserves its own dedicated debate.

presto-main/src/main/java/com/facebook/presto/server/InternalAuthenticationManager.java
```diff
@@ -41,7 +41,7 @@ public class InternalAuthenticationManager
 {
     private static final Logger log = Logger.get(InternalAuthenticationManager.class);
 
-    private static final String PRESTO_INTERNAL_BEARER = "X-Presto-Internal-Bearer";
+    public static final String PRESTO_INTERNAL_BEARER = "X-Presto-Internal-Bearer";
```
yingsu00, line 44: Can you please move this line to be the start of the class since it's public?

yingsu00 (approved):

skairali (approved):

tdcmeehan:
Superseded by #20852

PR #20810, opened 2023-09-08
Add noisy_sum_gaussian function

duykienvp:
## Description
This commit adds NOISY_SUM_GAUSSIAN aggregation and refactors some files related to noisy_count_gaussian function to support this new function. This noisy_sum_gaussian is to replace `SUM(col)` with `NOISY_SUM_GAUSSIAN(col, noiseScale[, lower, upper][, randomSeed])`. This is a continuation of the previous PR supporting noisy_count_gaussian.

`col` can be of numerical types: TINYINT, SMALLINT, INTEGER, BIGINT, REAL, DOUBLE, DECIMAL.

Because noise is of type Double, all values are converted to Double before being added to the sum, and the return type is Double.

When a bound [lower, upper] is provided, each value is clipped to this range before being added to the sum. The sum is then processed to make sure the sign of the final sum is consistent with the range. So, if `lower >= 0`, the sum is then equal to the max of the sum and 0; if `upper <= 0`, the sum is then equal to min of the sum and 0.

Optional randomSeed is used to get a fixed value of noise, often for reproducibility purposes. If randomSeed is omitted, SecureRandom is used. If randomSeed is provided, Random is used.

## Motivation and Context
This is one of aggregations in our effort to add Presto UDF for noisy aggregations, used as building block for differential privacy in Presto.

The purpose is to help build systems/tools/framework that provide differential privacy guarantees. Differential privacy has been used by multiple teams within Meta to develop privacy-preserving systems. Current implementation involves complicated SQL operation even for simplest aggregations, increasing development time, complexity, maintenance and sharing cost, and sometimes completely blocking development of new features.

While these functions on their own do not guarantee 100% differential privacy, they are the building blocks for other systems. That is also why we do not call these functions “differentially private aggregations” but only “noisy aggregations” to avoid a wrong impression of achieving differential privacy solely by using these functions.

## Impact
This commit adds `NOISY_SUM_GAUSSIAN(col, noiseScale[, lower, upper][, randomSeed])` aggregation which calculates the sum over the input values, and then adds random Gaussian noise with 0 mean and standard deviation of ``noise_scale`` to the true sum. This also provides options to clip values to a range `[lower, upper]` and a random seed for reproducibility

## Test Plan
- Unittest: Tested all of `NOISY_SUM_GAUSSIAN(col, noiseScale)`, `NOISY_SUM_GAUSSIAN(col, noiseScale, randomSeed)`, `NOISY_SUM_GAUSSIAN(col, noiseScale, lower, upper)`, `NOISY_SUM_GAUSSIAN(col, noiseScale, lower, upper randomSeed)`:
  - Available for all standard numerical data types: TINYINT, SMALLINT, INTEGER, BIGINT, REAL, DOUBLE, DECIMAL.
  - All following tests are for BIGINT, REAL, DOUBLE, DECIMAL types, while TINYINT, SMALLINT, INTEGER were only tested if such type can be processed because it is equivalent to BIGINT input
    - Tested with noiseScale < 0, noiseScale = 0, noiseScale = fixed value, with some noiseScale the output is within 50 x noiseScale
    - Tested with NULL rows
    - Tested with valid and invalid clipping bounds
    - Tested with random seed
    - Run local queries to test its output compared to SUM(col), and tested when input has 0 rows, with and without GROUP BY
  - Tested on tpch schema for `NOISY_SUM_GAUSSIAN(col, noiseScale)`, `NOISY_SUM_GAUSSIAN(col, noiseScale, randomSeed)` compared to normal SUM:
    - Clipping was not tested because normal SUM does not offer clipping
    - Only tested INTEGER, BIGINT, DOUBLE. Other types couldn't be found in the schema

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* `NOISY_SUM_GAUSSIAN(col, noiseScale[, lower, upper][, randomSeed])` aggregation which calculates the sum over the input values, and then adds random Gaussian noise with 0 mean and standard deviation of ``noise_scale`` to the true sum. 
  This also provides options to clip values to a range `[lower, upper]` and a random seed for reproducibility.
  All values are converted to `DOUBLE` before being added to the sum, and the return type is `DOUBLE`. 
  When there are no input rows, this function returns ``NULL``.
  When generating noise, if randomSeed is omitted, SecureRandom is used; otherwise, Random is used.

### CODE REVIEW

**Comment by duykienvp** [comment]
Most of line changes are kinda similar because 
- 1 file structure of `NoisySumGaussianDoubleAggregation` was copied for other types: Real, Long, Integer, SmallInt, TinyInt.
- 1 file structure of `NoisySumGaussianDecimalAggregation` was copied for other variations: NoisySumGaussianDecimalLowerUpperAggregation, NoisySumGaussianDecimalLowerUpperRandomSeedAggregation, NoisySumGaussianDecimalRandomSeedAggregation
- 1 file structure of unittest `TestNoisySumGaussianDoubleAggregation` was copied for testing other types

Not sure how to make those better but I think unittest is necessary

File: presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/NoisySumAggregationUtils.java [hunk 1]
```diff
@@ -0,0 +1,147 @@
...
+     */
+    public static double clip(double value, double lower, double upper)
+    {
+        return Math.max(lower, Math.min(upper, value));
+    }
+
+    /**
+     * Given a range [lower, upper], making sure that the sign of noisy sum is consistent with the sign of the bound.
+     */
+    public static double fixSignNoisySum(double noisySum, double lower, double upper)
```
jonhehir, line 112: I'm on the fence about whether these sorts of things should be always-on behavior. This kind of post-processing is probably often desirable but occasionally undesirable. In particular, this results in biased (not zero mean) noise addition for values close to zero. Depending on the situation, this could be surprising. For example, if you were to select `NOISY_SUM_GAUSSIAN(x, noiseScale, 0, 100) GROUP BY y`, then roll up (i.e., sum) the resulting values, the noisy rollup would overstate the true sum on average.

presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/NoisySumAggregationUtils.java
```diff
@@ -0,0 +1,147 @@
...
+import static com.facebook.presto.common.type.DoubleType.DOUBLE;
+import static com.facebook.presto.spi.StandardErrorCode.INVALID_FUNCTION_ARGUMENT;
+
+public class NoisySumAggregationUtils
+{
+    private NoisySumAggregationUtils()
+    {
+    }
+
+    public static void updateNoisySumState(NoisySumState state, double value, double noiseScale, Double lower, Double upper, Long randomSeed)
```
jonhehir, line 31: Should this be defined as an instance method in `NoisySumState` instead?

presto-main/src/test/java/com/facebook/presto/operator/aggregation/noisyaggregation/TestNoisySumGaussianDoubleAggregation.java
```diff
@@ -0,0 +1,419 @@
...
+    @Test
+    public void testNoisySumGaussianDoubleClippingZeroNoiseScale()
+    {
+        JavaAggregationFunctionImplementation noisySumGaussian = getFunction(DOUBLE, DOUBLE, DOUBLE, DOUBLE);
+
+        int numRows = 10;
+        List<Double> values = createTestValues(numRows, false, 1.0, false);
+        double lower = 2.0;
+        double upper = 8.0;
+        double expected = clippedSum(values, lower, upper);
```
jonhehir, line 200: I think I'd prefer to hardcode this, since `clippedSum` calls out to the very functionality being tested (`NoisySumAggregationUtils.clip`).

I think the answer is 53, but don't quote me on that. 😄

presto-main/src/test/java/com/facebook/presto/operator/aggregation/noisyaggregation/TestNoisySumGaussianDoubleAggregation.java
```diff
@@ -0,0 +1,419 @@
...
+    @Test(expectedExceptions = PrestoException.class)
+    public void testNoisySumGaussianDoubleClippingInvalidBound()
+    {
+        JavaAggregationFunctionImplementation noisySumGaussian = getFunction(DOUBLE, DOUBLE, DOUBLE, DOUBLE);
+
+        int numRows = 10;
+        List<Double> values = createTestValues(numRows, false, 1.0, false);
+        double lower = 2.0;
+        double upper = -8.0;
+        double expected = clippedSum(values, lower, upper);
```
jonhehir, line 222: same... (there are a bunch of these)

presto-main/src/test/java/com/facebook/presto/operator/aggregation/noisyaggregation/TestNoisySumGaussianDoubleAggregation.java
```diff
@@ -0,0 +1,419 @@
...
+        assertAggregation(
+                noisySumGaussian,
+                equalDoubleAssertion,
+                "Test noisy_sum_gaussian(double, noiseScale, lower, upper, randomSeed) that needs to fix sign",
+                new Page(
+                        createDoublesBlock(values),
+                        createRLEBlock(10000.0, numRows),
+                        createRLEBlock(lower, numRows),
+                        createRLEBlock(upper, numRows),
+                        createRLEBlock(23, numRows)), // this seed = 23 expected to create a negative random noise but lower > 0 => expect 0
```
jonhehir, line 328: A question for one of the Presto maintainers: is it generally considered safe to write unit tests that rely on a seeded PRNG returning a specific result? There are a few such tests in here.

jonhehir (commented):
I defer to @mlyublena and crew for the nitty gritty review of the Presto stuff, but I've left a few comments below, mostly relating to the docs, noise addition, and unit tests.

presto-docs/src/main/sphinx/functions/aggregate.rst
```diff
@@ -360,6 +360,74 @@ Approximate Aggregate Functions
         SELECT noisy_count_gaussian(orderkey, 20.0, 321) FROM tpch.tiny.lineitem WHERE false; -- NULL (1 row)
         SELECT noisy_count_gaussian(orderkey, 20.0, 321) FROM tpch.tiny.lineitem WHERE false  GROUP BY orderkey; --  (0 row)
 
+.. function:: noisy_sum_gaussian(x, noise_scale) -> double
+
+    Calculates the sum over the input values, and add random Gaussian noise
```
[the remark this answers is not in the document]
duykienvp: Will fix

presto-docs/src/main/sphinx/functions/aggregate.rst
```diff
@@ -360,6 +360,74 @@ Approximate Aggregate Functions
...
+    with 0 mean and standard deviation of ``noise_scale`` to the true count.
+    All values are converted to Double before being added to the sum, and the return type is Double.
+
+    When a bound ``[lower, upper]`` is provided, each value is clipped to this range before adding to the sum.
```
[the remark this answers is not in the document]
duykienvp: Will fix

presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/NoisySumAggregationUtils.java
duykienvp: I dont think it should be. `NoisySumState` is only an interface, not a class though.

presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/NoisySumAggregationUtils.java
duykienvp: I am on the fence too

presto-main/src/test/java/com/facebook/presto/operator/aggregation/noisyaggregation/TestNoisySumGaussianDoubleAggregation.java
duykienvp: Guess what. It is 45. The list is [0, ..., 9] though

jonhehir: LOL ❤️

presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/NoisySumAggregationUtils.java
jonhehir: Good point. My bad!

presto-main/src/test/java/com/facebook/presto/operator/aggregation/noisyaggregation/TestNoisySumGaussianDoubleAggregation.java
duykienvp: See. I was wrong as well. Other lines like this in this file are 45 because they have NULL instead of 0:
- When there is NULL, it is [1, ..., 9, NULL], then 1 is clipped to 2 and 9 is clipped to 8, so we still have 45
- When there is no NULL, it is [0, 1, ..., 9], then 0 -> 2,  1 -> 2, 9 -> 8, so we have 47 
:))

aditi-pandit:
@duykienvp : Please can you add an issue in https://github.com/orgs/prestodb/projects/3 to track implementing this function in Prestissimo (Presto native engine).

duykienvp:
> @duykienvp : Please can you add an issue in https://github.com/orgs/prestodb/projects/3 to track implementing this function in Prestissimo (Presto native engine).

@aditi-pandit I created 2 issues for this and the previous function noisy_count_gaussian() but cannot edit those issues to assign a project to them. Could you please help me add them to the project? 
https://github.com/prestodb/presto/issues/20826
https://github.com/prestodb/presto/issues/20827

presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/NoisySumAggregationUtils.java
duykienvp: Removed fixSign part

steveburnett (approved):
LGTM! (docs)

presto-main/src/test/java/com/facebook/presto/operator/aggregation/noisyaggregation/TestNoisySumGaussianDoubleAggregation.java
mlyublena: I think that's a good way to generate deterministic tests when checking for expected results. 
Would be good to also exercise the non-deterministic path, even if we don't check for exact result matches

presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/NoisySumState.java
```diff
@@ -0,0 +1,83 @@
...
+import com.facebook.presto.operator.aggregation.state.InitialBooleanValue;
+import com.facebook.presto.spi.function.AccumulatorStateMetadata;
+import io.airlift.slice.SliceInput;
+import io.airlift.slice.SliceOutput;
+
+import static io.airlift.slice.SizeOf.SIZE_OF_BYTE;
+import static io.airlift.slice.SizeOf.SIZE_OF_DOUBLE;
+
+@AccumulatorStateMetadata(stateSerializerClass = NoisySumStateSerializer.class)
+public interface NoisySumState
```
mlyublena, line 25: conceptually I find it a bit confusing that the sum state extends the count state though I don't have a much better proposal (apart from maybe renaming NoisyCountState to something more generic like NoisyAggState but then it might be confusing what the connection to count is)

mlyublena: perhaps add comments to what NoisyCountState represents in the corresponding file

mlyublena:
Please add documentation to the new classes.
Apart from that, looks good to me from Presto perspective, I trust domain experts to do a more thorough review of the algorithms

mlyublena (approved):

presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/NoisySumState.java
duykienvp: Thanks, I will add more docstring to the class.
In the implementation of normal SUM, there is a boolean flag of isNull which basically means there is no rows. If I understand correctly, it is equivalent to storing count explicitly and check if that count == 0. Maybe there is other context that I dont know

I decided to go with storing count explicitly to deal with 0 rows case because in the next implementation for Avg, we need to store that count anyway. So it is better to implement it here and then reuse it for Avg.

And the implementation of normal AVG is actually similar to what we have here: store count explicitly and check if count==0 then output NULL.

jonhehir (approved):
Looks good to me on the algo side!

presto-tests/src/test/java/com/facebook/presto/tests/TestNoisyAggregations.java
```diff
@@ -27,6 +32,13 @@ protected QueryRunner createQueryRunner()
         return TpchQueryRunnerBuilder.builder().build();
     }
 
+    @Override
+    protected QueryRunner createExpectedQueryRunner()
```
pranjalssh, line 36: Why is this needed?

duykienvp: Thanks. Short answer is to make sure that the actual and expected queries are run with the same runner.

queryRunner is a TpchQueryRunnerBuilder (which I assume would give us TpchQueryRunner)
expectedQueryRunner is by default a H2QueryRunner. 

I am not super what their difference is, but I think it is better for them to be the same.

In addition, in my WIP code to implement the next function in this series, noisy_avg_gaussian, **without** this override to make sure they are the same runner, I got an error where:
- `SELECT noisy_avg_gaussian(linenumber, 0) FROM lineitem` returns 3.004270876609888
- `SELECT avg(linenumber) FROM lineitem` returns 3.0

I needed this override to get that test working. So I think it is better for them to be the same anyway.

pranjalssh: Discussed offline, let's address this in the PR for noisy_avg_gaussian. H2QueryRunner is returning 3.0 for `SELECT avg(linenumber) FROM lineitem` which is weird and needs to be investigated. However, we can do that in parallel and not block this PR

pranjalssh: 3.0 is fishy, maybe its a type issue? can you try avg(linenumber+.0)?

pranjalssh: Looks like a type issue, avg(linenumber+.0) returns the correct value. So its ok to use TpchQueryRunner to compare restults, and in the comment we can add that H2QueryRunner is doing averages as ints instead of floats(not sure why).

duykienvp: yeah, I tried avg(linenumber + 0.0) or avg(cast(linenumber as double)) and both worked. 

Just for the next PR for avg, do you suggest overriding expectedQueryRunner or casting to double like this?

pranjalssh (approved):
LGTM

presto-tests/src/test/java/com/facebook/presto/tests/TestNoisyAggregations.java
pranjalssh: @duykienvp overriding is fine.

Issue #20815, opened 2023-09-09
[native] Plan conversion fails with not a scalar type! kind: ROW|ARRAY|MAP


mbasmanova:
```
SELECT * FROM orders WHERE (orderkey, custkey) IN ((1, 2), (3, 4))

VeloxRuntimeError:  not a scalar type! kind: ROW
	at Unknown.# 0  facebook::velox::VeloxException::VeloxException(char const*, unsigned long, char const*, std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >, bool, facebook::velox::VeloxException::Type, std::basic_string_view<char, std::char_traits<char> >)(Unknown Source)
	at Unknown.# 1  void facebook::velox::detail::veloxCheckFail<facebook::velox::VeloxRuntimeError, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&>(facebook::velox::detail::VeloxCheckFailArgs const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)(Unknown Source)
	at Unknown.# 2  facebook::presto::(anonymous namespace)::convertInExpr(std::vector<std::shared_ptr<facebook::velox::core::ITypedExpr const>, std::allocator<std::shared_ptr<facebook::velox::core::ITypedExpr const> > > const&, facebook::velox::memory::MemoryPool*)::$_0::operator()() const(Unknown Source)
	at Unknown.# 3  facebook::presto::VeloxExprConverter::toVeloxExpr(std::shared_ptr<facebook::presto::protocol::SpecialFormExpression>) const(Unknown Source)
	at Unknown.# 4  facebook::presto::VeloxExprConverter::toVeloxExpr(std::shared_ptr<facebook::presto::protocol::RowExpression>) const(Unknown Source)
[...]

PR #20816, opened 2023-09-09
[native] Fix plan conversion for IN (<complex type values>)

mbasmanova:
Add support for converting IN predicates with array, map and struct values.

After the fix, the query still fails because Velox doesn't support IN predicate with complex types.

Before the fix:

```
SELECT * FROM orders WHERE (orderkey, custkey) IN ((1, 2), (3, 4))

VeloxRuntimeError:  not a scalar type! kind: ROW
```

After the fix:

```
Scalar function in not registered with arguments: (ROW<"":BIGINT,"":BIGINT>, ARRAY<ROW<"":BIGINT,"":BIGINT>>).
```

See #20815

```
== NO RELEASE NOTE ==
```


amitkdutta (approved):
Thanks @mbasmanova

PR #20825, opened 2023-09-11
SetDigest jaccard_index implementation Issue

Akanksha-kedia:
SetDigest jaccard_index implementation Issue

https://en.wikipedia.org/wiki/Jaccard_index

 

For below query the jaccard_index should be equal to 0.3333333333333333 but the output is coming as 0.5 which is incorrect 

SELECT jaccard_index(make_set_digest(value), make_set_digest(value1)) FROM (VALUES ('abc', 'def'),('ee', 'abc')) T(value,value1);

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/type/setdigest/SetDigest.java
+++ b/presto-main/src/main/java/com/facebook/presto/type/setdigest/SetDigest.java
@@ -183,6 +183,34 @@ public class SetDigest
         return intersection / (double) sizeOfSmallerSet;
     }
 
+    public static double jaccardIndex_new(SetDigest a, SetDigest b)
+    {
+        SetDigest smaller;
+        SetDigest larger;
+        if (a.minhash.size() < b.minhash.size()) {
+            smaller = a;
+            larger = b;
+        }
+        else {
+            smaller = b;
+            larger = a;
+        }
+
+        LongSortedSet smallerSet = new LongRBTreeSet(smaller.minhash.keySet());
+        LongSortedSet largerSet = new LongRBTreeSet(larger.minhash.keySet());
+
+        int intersection = 0;
+        for (long key : smallerSet) {
+            if (largerSet.contains(key)) {
+                intersection++;
+            }
+        }
+        LongSortedSet union = new LongRBTreeSet(largerSet);
+        union.addAll(smallerSet);
+
+        return intersection / (double) union.size();
+    }
+
     public void add(long value)
     {
         addHash(Murmur3Hash128.hash64(value));
--- a/presto-main/src/main/java/com/facebook/presto/type/setdigest/SetDigestFunctions.java
+++ b/presto-main/src/main/java/com/facebook/presto/type/setdigest/SetDigestFunctions.java
@@ -55,6 +55,30 @@ public final class SetDigestFunctions
 
         long cardinality1 = digest1.cardinality();
         long cardinality2 = digest2.cardinality();
+        double jaccard = SetDigest.jaccardIndex_new(digest1, digest2);
+        digest1.mergeWith(digest2);
+        long result = Math.round(jaccard * digest1.cardinality());
+
+        // When one of the sets is much smaller than the other and approaches being a true
+        // subset of the other, the computed cardinality may exceed the cardinality estimate
+        // of the smaller set. When this happens the cardinality of the smaller set is obviously
+        // a better estimate of the one computed with the Jaccard Index.
+        return Math.min(result, Math.min(cardinality1, cardinality2));
+    }
+
+    @ScalarFunction
+    @SqlType(StandardTypes.BIGINT)
+    public static long intersectionCardinality_new(@SqlType(SetDigestType.NAME) Slice slice1, @SqlType(SetDigestType.NAME) Slice slice2)
+    {
+        SetDigest digest1 = SetDigest.newInstance(slice1);
+        SetDigest digest2 = SetDigest.newInstance(slice2);
+
+        if (digest1.isExact() && digest2.isExact()) {
+            return exactIntersectionCardinality(digest1, digest2);
+        }
+
+        long cardinality1 = digest1.cardinality();
+        long cardinality2 = digest2.cardinality();
         double jaccard = SetDigest.jaccardIndex(digest1, digest2);
         digest1.mergeWith(digest2);
         long result = Math.round(jaccard * digest1.cardinality());
--- a/presto-main/src/test/java/com/facebook/presto/sql/query/TestSetDigestFunctions.java
+++ b/presto-main/src/test/java/com/facebook/presto/sql/query/TestSetDigestFunctions.java
@@ -55,7 +55,20 @@ public class TestSetDigestFunctions
     {
         assertions.assertQuery(
                 "SELECT jaccard_index(make_set_digest(v1), make_set_digest(v2)) " +
-                        "FROM (VALUES (1, 1), (NULL,2), (2, 3), (NULL, 4)) T(v1, v2)", "VALUES CAST(0.5 AS DOUBLE)");
+                        "FROM (VALUES (2, NULL), (3, 4)) T(v1, v2)", "VALUES CAST(0.0 AS DOUBLE)");
+
+        assertions.assertQuery(
+                "SELECT jaccard_index(make_set_digest(v1), make_set_digest(v2)) " +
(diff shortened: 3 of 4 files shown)


Akanksha-kedia:
@tdcmeehan  please review

Akanksha-kedia:
@tdcmeehan please review and merge

tdcmeehan:
Could you please add some unit tests?

Akanksha-kedia:
@tdcmeehan HAVE ADDED TEST CASES

Akanksha-kedia:
The Jaccard Index is a measure used for comparing the similarity and diversity of sample sets. It is calculated as the intersection divided by the union of the sample sets.

However, in this case, you have provided pairs of values, not sets. The Jaccard Index is not typically used for pairs of values like this.

If we consider your pairs as sets, then:

For (1,1) and (NULL,2), there is no common element, so the Jaccard Index is 0.
For (1,1) and (2,3), there is no common element, so the Jaccard Index is 0.
For (1,1) and (NULL,4), there is no common element, so the Jaccard Index is 0.
For (NULL,2) and (2,3), there is one common element (2), so the Jaccard Index is 1/3 = 0.33.
For (NULL,2) and (NULL,4), there is one common element (NULL), so the Jaccard Index is 1/3 = 0.33.
For (2,3) and (NULL,4), there is no common element, so the Jaccard Index is 0.


If we treat these as two sets:

Set A = {1, null, 2, null} = {1, null, 2} (since sets only contain unique elements) Set B = {1, 2, 3, 4}

The intersection of A and B (elements common to both sets) is {1, 2} which has 2 elements.

The union of A and B (all unique elements in both sets) is {1, null, 2, 3, 4} which has 5 elements.

Therefore, the Jaccard Index, which is the size of the intersection divided by the size of the union, would be 2/5 = 0.4.

Akanksha-kedia:
@tdcmeehan please review

tdcmeehan:
I do not have enough knowledge about jaccard index to be confident in merging it.  Let's get some additional eyes--I'll find other reviewers.

presto-main/src/main/java/com/facebook/presto/type/setdigest/SetDigest.java
```diff
@@ -165,22 +165,16 @@ public static long exactIntersectionCardinality(SetDigest a, SetDigest b)
...
         LongSortedSet minUnion = new LongRBTreeSet(a.minhash.keySet());
         minUnion.addAll(b.minhash.keySet());
 
         int intersection = 0;
-        int i = 0;
         for (long key : minUnion) {
             if (a.minhash.containsKey(key) && b.minhash.containsKey(key)) {
                 intersection++;
             }
-            i++;
```
jaystarshot, line 178: why is this removed? this was a good optimization

presto-main/src/main/java/com/facebook/presto/type/setdigest/SetDigest.java
```diff
@@ -165,22 +165,16 @@ public static long exactIntersectionCardinality(SetDigest a, SetDigest b)
...
-            if (i >= sizeOfSmallerSet) {
-                break;
-            }
         }
-        return intersection / (double) sizeOfSmallerSet;
+        return intersection / (double) minUnion.size();
```
jaystarshot, line 177: I think the fix is only needed for this line

presto-main/src/main/java/com/facebook/presto/type/setdigest/SetDigest.java
jaystarshot, line 178: I was mistaken, small optimization - maybe we can break if (intersection >= sizeofSmallerSet) ?

jaystarshot:
Lgtm

presto-main/src/main/java/com/facebook/presto/type/setdigest/SetDigest.java
```diff
@@ -165,22 +165,16 @@ public static long exactIntersectionCardinality(SetDigest a, SetDigest b)
 
     public static double jaccardIndex(SetDigest a, SetDigest b)
     {
-        int sizeOfSmallerSet = Math.min(a.minhash.size(), b.minhash.size());
         LongSortedSet minUnion = new LongRBTreeSet(a.minhash.keySet());
         minUnion.addAll(b.minhash.keySet());
 
         int intersection = 0;
-        int i = 0;
         for (long key : minUnion) {
```
ClarenceThreepwood, line 172: This is iterating over all elements in minUnion and then checking for the element's membership in each input set.
How about just iterating over the elements in the smaller set among the two, and then checking if the element exists  in the other?

presto-main/src/main/java/com/facebook/presto/type/setdigest/SetDigest.java
ClarenceThreepwood: +1

ClarenceThreepwood (changes requested):

ClarenceThreepwood:
Also please fix checkstyle and failing tests

hmadison:
[Following up from Slack](https://prestodb.slack.com/archives/CLYMEEJ6S/p1695740671700649?thread_ts=1695737791.924169&cid=CLYMEEJ6S), while the implementation looks to be correct I'm concerned that this is a breaking change. Would making a three arity function for jaccard_index make more sense and leave the two arity function with the current implementation until there is some kind of deprecation cycle?

Akanksha-kedia:
Yes, it makes sense.But lets take more understanding from the community
people.

Akanksha-kedia:
@ClarenceThreepwood please review

presto-main/src/main/java/com/facebook/presto/type/setdigest/SetDigest.java
```diff
@@ -183,6 +183,25 @@ public static double jaccardIndex(SetDigest a, SetDigest b)
         return intersection / (double) sizeOfSmallerSet;
     }
 
+    public static double jaccardIndex(SetDigest a, SetDigest b)
+    {
+        int sizeOfSmallerSet = Math.min(a.minhash.size(), b.minhash.size());
```
ClarenceThreepwood, line 188: What do you think of something like the following instead

    public static double jaccardIndex(SetDigest a, SetDigest b)
    {
        SetDigest smaller;
        SetDigest larger;
        if (a.minhash.size() < b.minhash.size())
        {
            smaller = a;
            larger = b;
        }
        else {
            smaller = b;
            larger = a;
        }

        LongSortedSet smallerSet = new LongRBTreeSet(smaller.minhash.keySet());
        LongSortedSet largerSet = new LongRBTreeSet(larger.minhash.keySet());

        int intersection = 0;
        for (long key : smallerSet) {
            if (largerSet.contains(key)) {
                intersection++;
            }
        }
        LongSortedSet union = new LongRBTreeSet(largerSet);
        union.addAll(smallerSet);

        return intersection / (double) union.size();
    }

Akanksha-kedia:
-------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running com.facebook.presto.type.setdigest.TestSetDigest
[INFO] Tests run: 4, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 9.796 s - in com.facebook.presto.type.setdigest.TestSetDigest
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 4, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------


Tests run: 4, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 14.264 s - in com.facebook.presto.sql.query.TestSetDigestFunctions
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 4, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
 @tdcmeehan @ClarenceThreepwood

tdcmeehan:
@Akanksha-kedia please rebase to pick up test fixes.

Akanksha-kedia:
> test

i did not understand which tests i need to pick

tdcmeehan:
Please just register one function, and switch it with a config property.  Here's an example: https://github.com/prestodb/presto/blob/7a6f3768b0f5e2ca9c6052cd1e73f28420874aaa/presto-main/src/main/java/com/facebook/presto/metadata/BuiltInTypeAndFunctionNamespaceManager.java#L958

tdcmeehan:
Please reopen when you've updated the PR to address the feedback, thanks!

PR #20846, opened 2023-09-12
[native] Modify testScaleWriters() so it does not write DATE column.

spershin:
## Description
Modify testScaleWriters() so it does not write DATE column.
DWRF does not support DATE type and writes INTEGER instead.
After that INTEGER in the file schema and DATE in the table schema give mismatch.

```
== NO RELEASE NOTE ==
```


gggrace14 (approved):
So it means Prestissimo doesn't support writing column of Date type, does it?

PR #20849, opened 2023-09-13
[native] Allow updating spill directory in TaskManager.

amitkdutta:
It's possible that spill root directory is not available during server startup. For such cases, it's required to update spill directory (e.g. periodically). To support update, adding a lock to the spill directory path and exposing if it's empty method.

### DIFF
--- a/presto-native-execution/presto_cpp/main/TaskManager.h
+++ b/presto-native-execution/presto_cpp/main/TaskManager.h
@@ -42,6 +42,8 @@ class TaskManager {
 
   void setBaseSpillDirectory(const std::string& baseSpillDirectory);
 
+  bool emptyBaseSpillDirectory() const;
+
   /// Sets the time (ms) that a task is considered to be old for cleanup since
   /// its completion.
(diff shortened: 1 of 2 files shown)


presto-native-execution/presto_cpp/main/TaskManager.h
```diff
@@ -42,6 +42,8 @@ class TaskManager {
 
   void setBaseSpillDirectory(const std::string& baseSpillDirectory);
 
+  bool emptyBaseSpillDirectory() const;
```
xiaoxmeng, line 45: What's this used for? Can we remove? thanks!

xiaoxmeng (approved):
@amitkdutta LGTM. Thanks!

presto-native-execution/presto_cpp/main/TaskManager.h
amitkdutta: If a worker does not get spill path in start up, we want to schedule periodic task to update its spill path. Also we want to skip updating once it has a spill path. This method will tell the periodic task if a remote call is needed.

PR #20852, opened 2023-09-13
Enforce JWT authentication without bearer token for internal endpoints

nmahadevuni:
## Description
Enforces JWT authentication for all internal endpoints when enabled.

## Motivation and Context
When JWT authentication is enabled for internal communication, need to enforce that no internal endpoints can be accessed by external agents. 

## Impact
No public facing changes.

## Test Plan

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/server/security/InternalAuthenticationFilter.java
+++ b/presto-main/src/main/java/com/facebook/presto/server/security/InternalAuthenticationFilter.java
@@ -1,0 +1,96 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.server.security;
+
+import com.facebook.presto.server.InternalAuthenticationManager;
+
+import javax.annotation.security.RolesAllowed;
+import javax.inject.Inject;
+import javax.ws.rs.container.ContainerRequestContext;
+import javax.ws.rs.container.ContainerRequestFilter;
+import javax.ws.rs.container.ResourceInfo;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.Provider;
+
+import java.lang.reflect.Method;
+import java.security.Principal;
+import java.util.Arrays;
+import java.util.Optional;
+
+import static com.facebook.presto.server.security.RoleType.INTERNAL;
+import static java.util.Objects.requireNonNull;
+import static javax.servlet.http.HttpServletResponse.SC_UNAUTHORIZED;
+import static javax.ws.rs.core.Response.ResponseBuilder;
+
+@Provider
+public class InternalAuthenticationFilter
+        implements ContainerRequestFilter
+{
+    private final InternalAuthenticationManager internalAuthenticationManager;
+    private Optional<Principal> principal = Optional.empty();
+
+    @Context
+    ResourceInfo resourceInfo;
+
+    @Inject
+    public InternalAuthenticationFilter(InternalAuthenticationManager internalAuthenticationManager)
+    {
+        this.internalAuthenticationManager = requireNonNull(internalAuthenticationManager, "internalAuthenticationManager is null");
+    }
+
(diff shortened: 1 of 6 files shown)


nmahadevuni:
@tdcmeehan @czentgr @majetideepak Please review. If any of you can rerun just the failed tests, please restart them. They are all docker tests failing with "Temporary name resolution error"

presto-main/src/main/java/com/facebook/presto/server/security/InternalAuthenticationFilter.java
```diff
@@ -0,0 +1,95 @@
...
+                responseBuilder.status(SC_UNAUTHORIZED, "Unauthorized");
+                context.abortWith(responseBuilder.build());
+            }
+            else {
+                principal = Optional.of(authenticatedPrincipal);
+            }
+        }
+    }
+
+    boolean isAccessingInternalEndpoint()
```
tdcmeehan, line 80: Let's not use reflection to do this.  Reflection is very slow, and it's fairly wasteful to do this on a per-request basis.

nmahadevuni: We are not using any additional Reflection calls. Since this is using post-matching ContainerRequestFilter(By default, if we do not provide "@PreMatching" annotation, it is called post matching the URI to the ResourceMethod), the ResourceInfo is available at this point is just injected here.

presto-main/src/test/java/com/facebook/presto/server/security/TestInternalAuthenticationFilter.java
```diff
@@ -0,0 +1,171 @@
...
+        InternalAuthenticationManager internalAuthenticationManager = new InternalAuthenticationManager(Optional.of(sharedSecret), "nodeId", internalJwtEnabled);
+        InternalAuthenticationFilter internalAuthenticationFilter =
+                new InternalAuthenticationFilter(internalAuthenticationManager);
+        internalAuthenticationFilter.setResourceInfo(new ResourceInfoBuilder(TaskResource.class, null, null).build());
+
+        MockContainerRequestContext containerRequestContext = new MockContainerRequestContext(ImmutableListMultimap.of());
+
+        internalAuthenticationFilter.filter(containerRequestContext);
+
+        assertTrue(containerRequestContext.getResponse().getStatus() == SC_UNAUTHORIZED);
```
tdcmeehan, line 57: assertEquals here and elsewhere

presto-main/src/main/java/com/facebook/presto/server/security/InternalAuthenticationFilter.java
```diff
@@ -0,0 +1,95 @@
...
+    @Context
+    ResourceInfo resourceInfo;
+
+    @Inject
+    public InternalAuthenticationFilter(InternalAuthenticationManager internalAuthenticationManager)
+    {
+        this.internalAuthenticationManager = requireNonNull(internalAuthenticationManager, "internalAuthenticationManager is null");
+    }
+
+    Optional<Principal> getPrincipal()
```
tdcmeehan, line 53: public

presto-main/src/main/java/com/facebook/presto/server/security/InternalAuthenticationFilter.java
```diff
@@ -0,0 +1,95 @@
...
+import static java.util.Objects.requireNonNull;
+import static javax.servlet.http.HttpServletResponse.SC_UNAUTHORIZED;
+import static javax.ws.rs.core.Response.ResponseBuilder;
+
+@Provider
+public class InternalAuthenticationFilter
+        implements ContainerRequestFilter
+{
+    private final InternalAuthenticationManager internalAuthenticationManager;
+    private Optional<Principal> principal = Optional.empty();
```
tdcmeehan, line 42: private final

presto-main/src/main/java/com/facebook/presto/server/security/InternalAuthenticationFilter.java
```diff
@@ -0,0 +1,95 @@
...
+    {
+        this.internalAuthenticationManager = requireNonNull(internalAuthenticationManager, "internalAuthenticationManager is null");
+    }
+
+    Optional<Principal> getPrincipal()
+    {
+        return principal;
+    }
+
+    void setResourceInfo(ResourceInfo resourceInfo)
```
tdcmeehan, line 58: Please remove and place in constructor

presto-main/src/main/java/com/facebook/presto/server/security/InternalAuthenticationFilter.java
tdcmeehan: Please make it private.  Consider making it static.

Ideally, this would be done in airlift somewhere.  Consider creating an independent provider and injecting this provider.

tdcmeehan (changes requested):

nmahadevuni:
Should I move the InternalAuthenticationFilter class to Airlift project? RoleType class is in Presto has to be moved too.

czentgr (approved):

PR #20705, opened 2023-08-29
Rewrite CROSS JOIN with array not contains predicate into an anti join

mlyublena:
Rewrite queries of the form:

 ```
Filter not contains(r_array, l_key) 
  - Cross join 
        - scan l 
        - scan single_row_table r
```
into:

```
- Filter (l_array_elem IS NULL)
  - LOJ (l_key = l_array_elem) 
      - scan l 
      - Unnest l_array_elem <- unnest distinct_array 
          - project distinct_array := array_distinct(r_array) 
             - scan single_row_table
```

This will allow the optimizer to turn the CROSS JOIN into an anti semi join, which can be evaluated more efficiently.

PR #20861, opened 2023-09-13
Rewrite CROSS JOIN with array not contains predicate into an anti join

mlyublena:
This fixes a bug in the previous PR https://github.com/prestodb/presto/pull/20705 that crashed when the query had CONTAINS over an array expression rather than a simple array variable, for example:
```
presto:tpch> explain with t1 as (select * from (values (1, 'JAPAN'), (2, 'invalid_nation')) t(k, nation)) select t1.k, t1.nation from t1 where contains(transform((select array_agg(name) from nation), (x) ->lower(x)), lower(t1.nation));
Query 20230913_213945_00077_dm6ss failed: com.facebook.presto.spi.relation.CallExpression cannot be cast to com.facebook.presto.spi.relation.VariableReferenceExpression
java.lang.ClassCastException: com.facebook.presto.spi.relation.CallExpression cannot be cast to com.facebook.presto.spi.relation.VariableReferenceExpression
	at com.facebook.presto.sql.planner.iterative.rule.CrossJoinWithArrayContainsToInnerJoin.apply(CrossJoinWithArrayContainsToInnerJoin.java:129)
	at com.facebook.presto.sql.planner.iterative.rule.CrossJoinWithArrayContainsToInnerJoin.apply(CrossJoinWithArrayContainsToInnerJoin.java:73)
	at com.facebook.presto.sql.planner.iterative.IterativeOptimizer.transform(IterativeOptimizer.java:223)
	at com.facebook.presto.sql.planner.iterative.IterativeOptimizer.exploreNode(IterativeOptimizer.java:184)
	at com.facebook.presto.sql.planner.iterative.IterativeOptimizer.exploreGroup(IterativeOptimizer.java:146)
....
```

The reason for this is that the utility function that checks for applicability of the optimization was extended to handle the case with array expression, but the main optimization was not updated to handle this case properly.

Please review the second commit in this PR (8b97100db4b09d52b3d08623a70b5614861a0c5b) for the fix, will squash the commits before merging.

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* ...
* ...

Hive Changes
* ...
* ...
```

```
== NO RELEASE NOTE ==
```


presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/CrossJoinWithArrayContainsToInnerJoin.java
```diff
@@ -126,9 +127,9 @@ public Result apply(FilterNode node, Captures captures, Context context)
         List<RowExpression> andConjuncts = extractConjuncts(filterExpression);
         List<RowExpression> remainingConjuncts = andConjuncts.stream().filter(x -> !x.equals(arrayContainsExpression)).collect(toImmutableList());
 
-        VariableReferenceExpression array = (VariableReferenceExpression) ((CallExpression) arrayContainsExpression).getArguments().get(0);
+        RowExpression array = ((CallExpression) arrayContainsExpression).getArguments().get(0);
         VariableReferenceExpression element = (VariableReferenceExpression) ((CallExpression) arrayContainsExpression).getArguments().get(1);
```
feilong-liu, line 131: Is it possible that `element` can also be RowExpression rather than VariableReferenceExpression? I see that in candidate generation, we also assume that element can be a general RowExpression. If `element` cannot be a RowExpression, maybe we can change the candidate generation function to fix the gap.

mlyublena: Good point, I added this when I was developing the optimization for NOT CONTAINS but it seems all cases with an expression get pushed down in a projection. I'll make the function more restrictive and only allow variables for the element

mlyublena: Actually, for NOT CONTAINS I do hit this case, for example:
```
presto:tpch> explain with t1 as (select * from (values (1, 'JAPAN'), (2, 'invalid_nation')) t(k, nation)) select t1.k, t1.nation from t1 where not contains((select array_agg(name) from nation), t1.nation);
                                                                                                                            Query Plan                                                                                                          >
...
     - FilterProject[PlanNodeId 20,21][filterPredicate = not(contains(array_agg, CAST(field_0 AS varchar(25)))), projectLocality = LOCAL] => [field:integer, field_0:varchar(14)]                                                               >
         - CrossJoin[PlanNodeId 249] => [field:integer, field_0:varchar(14), array_agg:array(varchar(25))]
        ...
 ```

The element in the contains expression here is CAST(field_0 as varchar(25)). I handle this in the transformation and do the projection push down as part of the optimization.
The same query with contains has the cast pushed down before it gets to the optimization:
```
- FilterProject[PlanNodeId 325,324][filterPredicate = contains(array_agg, cast), projectLocality = LOCAL] => [field:integer, field_0:varchar(14)]
```

Comment (mlyublena) [reply presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/CrossJoinWithArrayContainsToInnerJoin.java]: I added some code to abort the CONTAINS optimization if the element is not a variable:
```
        if (!(element instanceof VariableReferenceExpression)) {
            // not currently supported
            return Result.empty();
        }
```
There are tests but I never hit this case so it's mostly a safe guard to not crash if we ever get there.

The same case for NOT CONTAINS is supported and there are tests to check this.

Comment (kaikalur) [reply presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/CrossJoinWithArrayContainsToInnerJoin.java]: Make it a checkstate(element instanceof VariableReferenceExpression) for readability

Comment (mlyublena) [reply presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/CrossJoinWithArrayContainsToInnerJoin.java]: checkstate will raise an error though - I want the behavior to be "don't optimize" and return original plan. Note that this case was not supported previously as the candidate computation function only looked at variables for both the array and the element arguments of CONTAINS

Comment (kaikalur) [reply presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/CrossJoinWithArrayContainsToInnerJoin.java]: Not here - but later where you explicitly just cast it as variable

File: presto-main/src/main/java/com/facebook/presto/sql/planner/PlannerUtils.java [hunk 2]
```diff
@@ -420,4 +440,29 @@ public static RowExpression coalesce(List<RowExpression> expressions)
...
+
+    public static boolean isSupportedArrayContainsFilter(FunctionResolution functionResolution, RowExpression filterExpression, List<VariableReferenceExpression> left, List<VariableReferenceExpression> right)
+    {
+        List<Type> supportedTypes = ImmutableList.of(BIGINT, INTEGER, VARCHAR, DATE);
+
+        if (filterExpression instanceof CallExpression) {
+            CallExpression callExpression = (CallExpression) filterExpression;
+            if (functionResolution.isArrayContainsFunction(callExpression.getFunctionHandle())) {
+                RowExpression array = callExpression.getArguments().get(0);
+                RowExpression element = callExpression.getArguments().get(1);
```
feilong-liu, line 452: How about also add a check here: `if (!(element instance of VariableReferenceExpression)) return false;` so that we filter out this case earlier.

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/CrossJoinWithArrayContainsToInnerJoin.java
feilong-liu: Maybe filter earlier in function `isSupportedArrayContainsFilter` and checkState here.

mlyublena: isSupportedArrayContainsFilter is shared between this optimization and the NOT CONTAINS optimization: for NOT CONTAINS we support this case by doing projection push down during optimization. For the CONTAINS case I couldn't find a use case that would trigger this, so I decided not to support it and rather just return early.
I'll add checkstate before casting as @kaikalur suggests

feilong-liu: Actually it also exists for the CFONTAINS case, for example a similar query like you have above 
```
presto:tpch> explain with t1 as (select * from (values (1, 'JAPAN'), (2, 'invalid_nation')) t(k, nation)) select t1.k, t1.nation from t1 where contains((select array_agg(name) from nation), t1.nation);
                                                                                                                            Query Plan                                                                             >
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------->
 - Output[PlanNodeId 25][k, nation] => [field:integer, field_0:varchar(14)]                                                                                                                                        >
         Estimates: {source: CostBasedSourceInfo, rows: ? (?), cpu: ?, memory: ?, network: ?}                                                                                                                      >
         k := field (1:101)                                                                                                                                                                                        >
         nation := field_0 (1:107)                                                                                                                                                                                 >
     - FilterProject[PlanNodeId 20,21][filterPredicate = contains(array_agg, CAST(field_0 AS varchar(25))), projectLocality = LOCAL] => [field:integer, field_0:varchar(14)]                                       >
             Estimates: {source: CostBasedSourceInfo, rows: ? (?), cpu: ?, memory: ?, network: ?}/{source: CostBasedSourceInfo, rows: ? (?), cpu: ?, memory: ?, network: ?}                                        >
         - CrossJoin[PlanNodeId 249] => [field:integer, field_0:varchar(14), array_agg:array(varchar(25))]                                                                                                         >
                 Estimates: {source: CostBasedSourceInfo, rows: 2 (120B), cpu: ?, memory: ?, network: ?}                                                                                                           >
                 Distribution: REPLICATED                                                                                                                                                                          >
             - LocalExchange[PlanNodeId 561][ROUND_ROBIN] () => [field:integer, field_0:varchar(14)]                                                                                                               >
                     Estimates: {source: CostBasedSourceInfo, rows: 2 (120B), cpu: 120.00, memory: 0.00, network: 0.00}                                                                                            >
                 - Values[PlanNodeId 0] => [field:integer, field_0:varchar(14)]                                                                                                                                    >
                         Estimates: {source: CostBasedSourceInfo, rows: 2 (120B), cpu: 0.00, memory: 0.00, network: 0.00}                                                                                          >
                         (INTEGER'1', VARCHAR'JAPAN')                                                                                                                                                              >
                         (INTEGER'2', VARCHAR'invalid_nation')                                                                                                                                                     >
             - EnforceSingleRow[PlanNodeId 17] => [array_agg:array(varchar(25))]                                                                                                                                   >
                     Estimates: {source: CostBasedSourceInfo, rows: 1 (60B), cpu: ?, memory: ?, network: ?}                                                                                                        >
                 - Aggregate(FINAL)[PlanNodeId 12] => [array_agg:array(varchar(25))]                                                                                                                               >
                         array_agg := "presto.default.array_agg"((array_agg_29)) (1:148)                                                                                                                           >
                     - LocalExchange[PlanNodeId 578][SINGLE] () => [array_agg_29:array(varchar(25))]                                                                                                               >
                         - RemoteStreamingExchange[PlanNodeId 584][GATHER] => [array_agg_29:array(varchar(25))]                                                                                                    >
                             - Aggregate(PARTIAL)[PlanNodeId 582] => [array_agg_29:array(varchar(25))]                                                                                                             >
                                     array_agg_29 := "presto.default.array_agg"((name)) (1:148)                                                                                                                    >
                                 - TableScan[PlanNodeId 9][TableHandle {connectorId='hive', connectorHandle='HiveTableHandle{schemaName=tpch, tableName=nation, analyzePartitionValues=Optional.empty}', layout='Op>
                                         Estimates: {source: CostBasedSourceInfo, rows: 25 (1.46kB), cpu: 302.00, memory: 0.00, network: 0.00}                                                                     >
                                         LAYOUT: tpch.nation{}                                                                                                                                                     >
                                         name := name:varchar(25):1:REGULAR (1:169)                                                                                                                                >
                                                                                                                                                                                                                   >
(1 row)
```

mlyublena: That's weird, I don't get that, in my case the cast is pushed down below the join:
```

presto:tpch> explain with t1 as (select * from (values (1, 'JAPAN'), (2, 'invalid_nation')) t(k, nation)) select t1.k, t1.nation from t1 where contains((select array_agg(name) from nation), t1.nation);
                                                                                                                            Query Plan
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 - Output[PlanNodeId 25][k, nation] => [field:integer, field_0:varchar(14)]
         Estimates: {source: CostBasedSourceInfo, rows: ? (?), cpu: ?, memory: ?, network: ?}
         k := field (1:101)
         nation := field_0 (1:107)
     - FilterProject[PlanNodeId 325,324][filterPredicate = contains(array_agg, cast), projectLocality = LOCAL] => [field:integer, field_0:varchar(14)]
             Estimates: {source: CostBasedSourceInfo, rows: ? (?), cpu: ?, memory: ?, network: ?}/{source: CostBasedSourceInfo, rows: ? (?), cpu: ?, memory: ?, network: ?}
         - CrossJoin[PlanNodeId 326] => [field:integer, field_0:varchar(14), cast:varchar(25), array_agg:array(varchar(25))]
                 Estimates: {source: CostBasedSourceInfo, rows: 2 (120B), cpu: ?, memory: ?, network: ?}
                 Distribution: REPLICATED
             - Project[PlanNodeId 323][projectLocality = LOCAL] => [field:integer, field_0:varchar(14), cast:varchar(25)]
                     Estimates: {source: CostBasedSourceInfo, rows: 2 (120B), cpu: 350.00, memory: 0.00, network: 0.00}
                     cast := CAST(field_0 AS varchar(25)) (1:22)
                 - LocalExchange[PlanNodeId 621][ROUND_ROBIN] () => [field:integer, field_0:varchar(14)]
                         Estimates: {source: CostBasedSourceInfo, rows: 2 (120B), cpu: 120.00, memory: 0.00, network: 0.00}
                     - Values[PlanNodeId 0] => [field:integer, field_0:varchar(14)]
                             Estimates: {source: CostBasedSourceInfo, rows: 2 (120B), cpu: 0.00, memory: 0.00, network: 0.00}
                             (INTEGER'1', VARCHAR'JAPAN')
                             (INTEGER'2', VARCHAR'invalid_nation')
             - EnforceSingleRow[PlanNodeId 17] => [array_agg:array(varchar(25))]
                     Estimates: {source: CostBasedSourceInfo, rows: 1 (60B), cpu: ?, memory: ?, network: ?}
                 - Aggregate(FINAL)[PlanNodeId 12] => [array_agg:array(varchar(25))]
                         array_agg := "presto.default.array_agg"((array_agg_29)) (1:148)
                     - LocalExchange[PlanNodeId 639][SINGLE] () => [array_agg_29:array(varchar(25))]
                         - RemoteStreamingExchange[PlanNodeId 645][GATHER] => [array_agg_29:array(varchar(25))]
                             - Aggregate(PARTIAL)[PlanNodeId 643] => [array_agg_29:array(varchar(25))]
                                     array_agg_29 := "presto.default.array_agg"((name)) (1:148)
                                 - TableScan[PlanNodeId 9][TableHandle {connectorId='hive', connectorHandle='HiveTableHandle{schemaName=tpch, tableName=nation, analyzePartitionValues=Optional.empty}', layout='Optional[tpch.nation{}]'}] => [name:varchar(25)]
                                         Estimates: {source: CostBasedSourceInfo, rows: 25 (1.46kB), cpu: 302.00, memory: 0.00, network: 0.00}
                                         LAYOUT: tpch.nation{}
                                         name := name:varchar(25):1:REGULAR (1:169)

(1 row)
```

feilong-liu: Oh, now I get all the context back.
The two existing cross join to inner join optimizations are used together with the optimization `PushDownFilterExpressionEvaluationThroughCrossJoin`. This `PushDownFilterExpressionEvaluationThroughCrossJoin` will create projections for the valid contains/or expression, and project the row expressions to VariableReferenceExpression. And that's why in existing `CrossJoinWithArrayContainsToInnerJoin` optimization, it always assume VariableReferenceExpression.

My test is in master branch, and your run is with change from this PR. As the `PushDownFilterExpressionEvaluationThroughCrossJoin` also changed in this PR, hence lead to different output here.

Let's discuss offline on what to do here, as it seems to need more thinking here.

mlyublena: I see, getCandidateArrayContainsExpression is now more permissive and allows for expressions, so those get picked up and pushed down in CrossJoinWithArrayContainsToInnerJoin. 
I think the old PushDownFIlter... had a bug: when looking whether there are candidate arrayContains expressions it only checked for variables, but the actual implementation is looking for expressions as well: addCandidateExpression is checking whether the used variables in the expression are present in the corresponding output columns, but with the old implementation of getCandidateArrayContainsExpression I don't think we would have selected such expressions as candidates at all.
Since we have tests to cover this case I think we should be fine

feilong-liu: The `CrossJoinWithArrayContainsToInnerJoin` relies on the `PushDownFilterExpressionEvaluationThroughCrossJoin` to map expressions to variables, and only consider the case where the input of contains function is variables.
I think the new not array contains optimizer should either follow the same pattern, i.e. rely on the `PushDownFilterExpressionEvaluationThroughCrossJoin` to do pushdown as existing ones, or act as an independent optimizer without touching the existing ones. The current PR leave the three cross join to inner join optimizers mixed together, but actually two of them rely on `PushDownFilterExpressionEvaluationThroughCrossJoin` while one does it by itself.

feilong-liu: > I think the old PushDownFIlter... had a bug: when looking whether there are candidate arrayContains expressions it only checked for variables, but the actual implementation is looking for expressions as well: addCandidateExpression is checking whether the used variables in the expression are present in the corresponding output columns, but with the old implementation of getCandidateArrayContainsExpression I don't think we would have selected such expressions as candidates at all.
Since we have tests to cover this case I think we should be fine


Would you mind describe it in more detail (or discuss it offline if it's hard to describe in comment here)? I get lost on this part.

mlyublena:
Based on online and offline discussions with @feilong-liu I changed the implementation of CrossJoinWithArrayNotContainsToAntiJoin and CrossJoinWithArrayContainsToInnerJoin to only expect CONTAINS on top of a variable element, pushing down of projections is delegated to PushDownFilterExpressionEvaluationThroughCrossJoin. I also extended the latter to push down expressions from NOT CONTAINS expressions. 
Please check the second commit for the fix

presto-main/src/main/java/com/facebook/presto/sql/planner/PlannerUtils.java
mlyublena: I changed the check below to ```right.contains(element)``` which would fail if element is not a VariableReferenceExpression

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/PushDownFilterExpressionEvaluationThroughCrossJoin.java
```diff
@@ -170,18 +171,21 @@ public Result apply(FilterNode filterNode, Captures captures, Context context)
     }
 
     // TODO: this function only works for filter in form of or condition and array contains function etc. make it generic to work for all RowExpressions
-    private List<Set<RowExpression>> getRowExpressions(RowExpression filterPredicate, List<VariableReferenceExpression> left, List<VariableReferenceExpression> right)
+    private List<Set<RowExpression>> getRowExpressions(FunctionResolution functionResolution, RowExpression filterPredicate, List<VariableReferenceExpression> left, List<VariableReferenceExpression> right)
     {
         List<Set<RowExpression>> candidateFromOrCondition = getRowExpressionsFromOrCondition(filterPredicate, left, right);
-        List<Set<RowExpression>> candidateFromArrayContains = getRowExpressionsFromArrayContains(filterPredicate, left, right);
+        List<Set<RowExpression>> candidateFromArrayContains = getRowExpressionsFromArrayContains(functionResolution, filterPredicate, left, right);
+        List<Set<RowExpression>> candidateFromArrayNotContains = getRowExpressionsFromArrayNotContains(functionResolution, filterPredicate, left, right);
```
mlyublena, line 178: added this so that we also push expressions from NOT CONTAINS

presto-main/src/main/java/com/facebook/presto/sql/planner/PlanOptimizers.java
```diff
@@ -505,6 +506,15 @@ public PlanOptimizers(
...
                                 new CrossJoinWithArrayContainsToInnerJoin(metadata.getFunctionAndTypeManager()))),
+                new IterativeOptimizer(
+                        metadata,
+                        ruleStats,
+                        statsCalculator,
+                        estimatedExchangesCostCalculator,
+                        ImmutableSet.of(
+                                new PushDownFilterExpressionEvaluationThroughCrossJoin(metadata.getFunctionAndTypeManager()),
+                                new CrossJoinWithOrFilterToInnerJoin(metadata.getFunctionAndTypeManager()),
+                                new CrossJoinWithArrayNotContainsToAntiJoin(metadata, metadata.getFunctionAndTypeManager()))),
```
feilong-liu, line 517: Why do we need to have a separate instance of IterativeOptimizer, instead of adding it to the instance above?

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/CrossJoinWithArrayContainsToInnerJoin.java
```diff
@@ -137,16 +119,21 @@ public Result apply(FilterNode node, Captures captures, Context context)
...
         List<RowExpression> andConjuncts = extractConjuncts(filterExpression);
         List<RowExpression> remainingConjuncts = andConjuncts.stream().filter(x -> !x.equals(arrayContainsExpression)).collect(toImmutableList());
 
-        VariableReferenceExpression array = (VariableReferenceExpression) ((CallExpression) arrayContainsExpression).getArguments().get(0);
-        VariableReferenceExpression element = (VariableReferenceExpression) ((CallExpression) arrayContainsExpression).getArguments().get(1);
-        boolean arrayAtLeftInput = leftInput.contains(array);
+        RowExpression array = ((CallExpression) arrayContainsExpression).getArguments().get(0);
+        RowExpression element = ((CallExpression) arrayContainsExpression).getArguments().get(1);
+
+        checkState(element instanceof VariableReferenceExpression, "Argument to CONTAINS is not a column");
```
feilong-liu, line 134: The error message is misleading, we can have argument of contains to be expression, just we do not allow here. Maybe we can leave this alone as this is only a safety check.

presto-main/src/main/java/com/facebook/presto/sql/planner/PlannerUtils.java
```diff
@@ -420,4 +440,28 @@ public static RowExpression coalesce(List<RowExpression> expressions)
...
     }
+
+    public static boolean isSupportedArrayContainsFilter(FunctionResolution functionResolution, RowExpression filterExpression, List<VariableReferenceExpression> left, List<VariableReferenceExpression> right)
+    {
+        List<Type> supportedTypes = ImmutableList.of(BIGINT, INTEGER, VARCHAR, DATE);
+
+        if (filterExpression instanceof CallExpression) {
+            CallExpression callExpression = (CallExpression) filterExpression;
+            if (functionResolution.isArrayContainsFunction(callExpression.getFunctionHandle())) {
+                RowExpression array = callExpression.getArguments().get(0);
```
feilong-liu, line 451: I think the PushDownFilterExpressionEvaluationThroughCrossJoin should already handle the pushdown of array expression, and array should be VariableReferenceExpression here?

mlyublena: fixed in the check below

presto-main/src/main/java/com/facebook/presto/sql/planner/PlanOptimizers.java
mlyublena: fixed

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/CrossJoinWithArrayContainsToInnerJoin.java
```diff
@@ -137,16 +119,21 @@ public Result apply(FilterNode node, Captures captures, Context context)
...
+        VariableReferenceExpression elementVar = (VariableReferenceExpression) element;
+        boolean arrayAtLeftInput = leftInput.containsAll(VariablesExtractor.extractAll(array));
```
feilong-liu, line 136: `array` is guaranteed to be VariableReference now, let's also change here?

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/CrossJoinWithArrayContainsToInnerJoin.java
feilong-liu: Maybe also add assert for `array`?

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/CrossJoinWithArrayNotContainsToAntiJoin.java
```diff
@@ -0,0 +1,213 @@
...
+        if (arrayNotContainsExpression == null) {
+            return Result.empty();
+        }
+        List<RowExpression> allConjuncts = extractConjuncts(filterExpression);
+        List<RowExpression> remainingConjuncts = allConjuncts.stream().filter(x -> !x.equals(arrayNotContainsExpression)).collect(Collectors.toList());
+
+        RowExpression arrayContainsExpression = arrayNotContainsExpression.getChildren().get(0);
+        RowExpression array = ((CallExpression) arrayContainsExpression).getArguments().get(0);
+        RowExpression element = ((CallExpression) arrayContainsExpression).getArguments().get(1);
+        boolean arrayAtLeftInput = leftColumns.containsAll(VariablesExtractor.extractAll(array));
```
feilong-liu, line 150: Similar, I think `array` and `element` are variable reference now, should we change here accordingly?

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/CrossJoinWithArrayContainsToInnerJoin.java
mlyublena: added

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/CrossJoinWithArrayContainsToInnerJoin.java
mlyublena: fixed and and added assert

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/CrossJoinWithArrayNotContainsToAntiJoin.java
mlyublena: fixed and added asserts

feilong-liu (approved):
Could you rerun queries which previously failed to make sure the problem is solved now?

pranjalssh (approved):

PR #20866, opened 2023-09-14
[native] Re-factor in PeriodicTaskManager.

amitkdutta:
`PeriodicTaskManager` has a public method `addTask` which is used to add periodic task to it's schedule. Since this is a public method, using it instead of calling `scheduler_` directly. This will keep the public function `addTask` in use inside OSS. This will prevent any accidental delete of this public function as it's not used.


xiaoxmeng (approved):
@amitkdutta LGTM. Thanks!

PR #20870, opened 2023-09-14
[native] Add a function to schedule task once in PeriodicTaskManager.

amitkdutta:
Exposing `addFunctionOnce` so that we can pass tasks that can run only once.

### DIFF
--- a/presto-native-execution/presto_cpp/main/PeriodicTaskManager.h
+++ b/presto-native-execution/presto_cpp/main/PeriodicTaskManager.h
@@ -68,6 +68,14 @@ class PeriodicTaskManager {
         func, std::chrono::microseconds{periodMicros}, taskName);
   }
 
+  /// Add a task to run once.
+  template <typename TFunc>
+  void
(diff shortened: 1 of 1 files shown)


presto-native-execution/presto_cpp/main/PeriodicTaskManager.h
```diff
@@ -68,6 +68,14 @@ class PeriodicTaskManager {
         func, std::chrono::microseconds{periodMicros}, taskName);
   }
 
+  /// Add a task to run once.
+  template <typename TFunc>
+  void
+  addTaskOnce(TFunc&& func, size_t periodMicros, const std::string& taskName) {
```
xiaoxmeng, line 74: Can you add a unit test with a test function to verify if it is only called once? Like verify by a counter incremented by the test function and wait for the test run a bit to verify the counter only incremented once? Thanks!

xiaoxmeng (approved):
@amitkdutta thanks for the change!

PR #20887, opened 2023-09-15
Support persistent CTE's

jaystarshot:
## Description

Support persistent query-level CTEs

The goal is to allow users to define persistent ctes and support them
Session property cte_materialization_strategy is added which when set to ALL will materialize all ctes to HDFS

This will be the first step for [19744](https://github.com/prestodb/presto/issues/19744).

## Motivation and Context
This change is based on the [design](https://docs.google.com/document/d/10J9_j08imqe6xEH9iepteG0DA5B-IpmYv4lSRvCgdJA/edit#heading=h.pz9gj1ifu7su). The eventual goal is to materialize or stream the CTEs (either in local memory or remote) using CBO rules and the architecture reflects that.
The temporary tables created for each cte will be partitioned on the first input column
In the current implementation, all CTEs will be evaluated **before** the query is run.

This change has 5 main commits
### [Add session property and ast changes](https://github.com/prestodb/presto/pull/20887/commits/25fc788d5f031443a47008c359d8d1591c4330a7)

This change adds the session property which allows CTE materialization. This information is also added to the CTEInformation collector.
The relational planner has the main change where a CTEReference Node is added if a CTE is persistent. This reference node will be used by the LogicalCteOptimizer to process.

### Planner changes 
Main changes 
**LogicalCteOptimizer** - Processes the CTE references into CteConsumers and CteProducers. It also creates a sequence node which is added as a child below the topmost node. The sequence node has a list of leftSources and a right source. The leftSources list has the ctes ordered in a topological order of execution. {a,b,c,d} --> order of execution {d->c->b->a}.
 
**PhysicalCteOptimizer** - Processes the CteConsumers and CteProducers into table writes and temporary table scan. The processing code is based on [basePlanFragmenter](https://github.com/prestodb/presto/blob/70ecc392450b9e4dc2cb08652c630733bea54adb/presto-main/src/main/java/com/facebook/presto/sql/planner/BasePlanFragmenter.java#L313). From logical to physical optimizers, the CteProducer, CteConsumers, and Sequence Node are present to enable further logical optimizations. Physical optimization is applied just before adding exchanges. Post this optimization stage, CteConsumers and CteProducers are no longer present.

Code change in other optimizers  and visitors
	
_AddExchanges.java_ - Handling Sequence Node
_HashGenerationOptimizer.java_ - Handling Sequence Node
_StreamPropertyDerivations.java_ - Handling Sequence Node
_PropertyDerivations.java_ - Handling Sequence Node
_PruneUnreferencedOutputs.java_ - Handling CteConsumer, CteProducer and Sequnce
_PushdownSubfields.java_ - handling CteProducer Node
_StreamPropertyDerivations.java_ - handling sequence
_UnaliasSymbolReferences.java_ - Handling CteConsumer, CteProducer and Sequnce

### Scheduling changes
_BasePlanFragmenter_ - Handle and transform sequence nodes. {a,b,c,d} --> order of execution {d->c->b->a}
We will create a chain of different sections of plans {d->c->b->a} and add it as a child of the main query section. 

[StreamingPlanSection](https://github.com/prestodb/presto/blob/70ecc392450b9e4dc2cb08652c630733bea54adb/presto-main/src/main/java/com/facebook/presto/execution/scheduler/StreamingPlanSection.java#L28) will make sure that the children are scheduled and executed before the parent

###  Tests
Integration test to test end to end and result correctness
Testing logical cte optimizer 

Current wips

- [x] Handle UnaliasSymbolReferences.java. 
- [x] Handle ValidateDependenciesChecker.java
- [x] Check if result correctness can be done at a record level
- [x] Add parser and planner unit tests

## Test Plan
Production tested with ALL CTEs materialized

## Contributor checklist

- [x] CI passed.

Todo in future PRs

1. Integrate with HBO and cost model
2. Local auto replacement of cte in the optimizer
3. Select the top five candidate strategies from the optimizer, execute the planning process for each from the logical to the physical stage within the optimizer, and determine the most optimal one.
4. Support non partitioned temporary tables 

Final Quality ToDos
- [x] Fix cases where the bucketing is done on unsupported buckets 
- [x] Production test again with changes

## Release Notes

```
== RELEASE NOTES ==

General Changes
* New Feature: Materialization of Common Table Expressions (CTEs) in Queries
* The relevant user connectors must support creating temporary tables 
* By materializing CTEs, users can take advantage of reusing common table expressions within their queries, which can lead to reductions in both storage operations and computational costs

## Recommended Configurations

1. CTE Materialization Strategy:
   - Enable by setting the session property ``cte_materialization_strategy`` or the configuration property ``cte-materialization-strategy`` to ``ALL``.
2. Temporary Table Storage Format:
   - For faster processing, set the configuration ``hive.temporary-table-storage-format`` or the session property ``temporary_table_storage_format`` to ``PAGEFILE``.
3. Temporary Table Schema:
   - It is recommended to set the configuration ``hive.temporary-table-schema`` or the session property ``temporary_table_schema`` to use the user's schema.
 4. Partitioning Provider Catalog
 - The config property ``query.partitioning-provider-catalog`` ``partitioning_provider_catalog`` session property must be set to enable cte materialization. The catalog must support providing a custom partitioning and ability to store temporary tables.

### DIFF
--- a/presto-hive/src/test/java/com/facebook/presto/hive/TestCteExecution.java
+++ b/presto-hive/src/test/java/com/facebook/presto/hive/TestCteExecution.java
@@ -1,0 +1,807 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.hive;
+
+import com.facebook.presto.Session;
+import com.facebook.presto.testing.MaterializedResult;
+import com.facebook.presto.testing.QueryRunner;
+import com.facebook.presto.tests.AbstractTestQueryFramework;
+import com.facebook.presto.tests.QueryAssertions;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import org.testng.annotations.Test;
+
+import java.util.Optional;
+
+import static com.facebook.presto.SystemSessionProperties.CTE_MATERIALIZATION_STRATEGY;
+import static com.facebook.presto.SystemSessionProperties.PARTITIONING_PROVIDER_CATALOG;
+import static com.facebook.presto.SystemSessionProperties.PUSHDOWN_SUBFIELDS_ENABLED;
+import static com.facebook.presto.testing.assertions.Assert.assertEquals;
+import static io.airlift.tpch.TpchTable.CUSTOMER;
+import static io.airlift.tpch.TpchTable.LINE_ITEM;
+import static io.airlift.tpch.TpchTable.NATION;
+import static io.airlift.tpch.TpchTable.ORDERS;
+import static io.airlift.tpch.TpchTable.PART;
+import static io.airlift.tpch.TpchTable.PART_SUPPLIER;
+import static io.airlift.tpch.TpchTable.REGION;
+import static io.airlift.tpch.TpchTable.SUPPLIER;
+
+@Test(singleThreaded = true)
+public class TestCteExecution
+        extends AbstractTestQueryFramework
+{
+    @Override
+    protected QueryRunner createQueryRunner()
+            throws Exception
+    {
+        return HiveQueryRunner.createQueryRunner(
+                ImmutableList.of(ORDERS, CUSTOMER, LINE_ITEM, PART_SUPPLIER, NATION, REGION, PART, SUPPLIER),
+                ImmutableMap.of(
+                        "query.partitioning-provider-catalog", "hive"),
+                "sql-standard",
+                ImmutableMap.of("hive.pushdown-filter-enabled", "true",
+                        "hive.enable-parquet-dereference-pushdown", "true"),
+                Optional.empty());
+    }
+
+    @Test
+    public void testSimplePersistentCte()
+    {
+        QueryRunner queryRunner = getQueryRunner();
+        compareResults(queryRunner.execute(getMaterializedSession(),
+                        "WITH  temp as (SELECT orderkey FROM ORDERS) " +
+                                "SELECT * FROM temp t1 "),
+                queryRunner.execute(getSession(),
+                        "WITH  temp as (SELECT orderkey FROM ORDERS) " +
+                                "SELECT * FROM temp t1 "));
+    }
+
+    @Test
+    public void testPersistentCteWithTimeStampWithTimeZoneType()
+    {
+        String testQuery = "WITH cte AS (" +
+                "  SELECT ts FROM (VALUES " +
+                "    (CAST('2023-01-01 00:00:00.000 UTC' AS TIMESTAMP WITH TIME ZONE)), " +
+                "    (CAST('2023-06-01 12:00:00.000 UTC' AS TIMESTAMP WITH TIME ZONE)), " +
+                "    (CAST('2023-12-31 23:59:59.999 UTC' AS TIMESTAMP WITH TIME ZONE))" +
+                "  ) AS t(ts)" +
+                ")" +
+                "SELECT ts FROM cte";
+        QueryRunner queryRunner = getQueryRunner();
+        compareResults(queryRunner.execute(getMaterializedSession(),
+                        testQuery),
+                queryRunner.execute(getSession(),
+                        testQuery));
+    }
+
+    @Test
+    public void testPersistentCteWithStructTypes()
+    {
+        String testQuery = "WITH temp AS (" +
+                "  SELECT * FROM (VALUES " +
+                "    (CAST(ROW('example_status', 100) AS ROW(status VARCHAR, amount INTEGER)), 1)," +
+                "    (CAST(ROW('another_status', 200) AS ROW(status VARCHAR, amount INTEGER)), 2)" +
+                "  ) AS t (order_details, orderkey)" +
+                ") SELECT * FROM temp";
+        QueryRunner queryRunner = getQueryRunner();
+        compareResults(queryRunner.execute(getMaterializedSession(),
+                        testQuery),
+                queryRunner.execute(getSession(),
+                        testQuery));
+    }
+
+    @Test
+    public void testCteWithZeroLengthVarchar()
+    {
+        String testQuery = "WITH temp AS (" +
+                "  SELECT * FROM (VALUES " +
+                "    (CAST('' AS VARCHAR(0)), 9)" +
+                "  ) AS t (text_column, number_column)" +
+                ") SELECT * FROM temp";
+        QueryRunner queryRunner = getQueryRunner();
+        compareResults(queryRunner.execute(getMaterializedSession(),
+                        testQuery),
+                queryRunner.execute(getSession(),
+                        testQuery));
+    }
+
+    @Test
+    public void testDependentPersistentCtes()
+    {
+        String testQuery = "WITH  cte1 AS (SELECT orderkey FROM ORDERS WHERE orderkey < 100), " +
+                "      cte2 AS (SELECT * FROM cte1 WHERE orderkey > 50) " +
+                "SELECT * FROM cte2";
+        QueryRunner queryRunner = getQueryRunner();
+        compareResults(
+                queryRunner.execute(getMaterializedSession(), testQuery),
+                queryRunner.execute(getSession(), testQuery));
+    }
+
+    @Test
+    public void testMultipleIndependentPersistentCtes()
+    {
+        String testQuery = "WITH  cte1 AS (SELECT orderkey FROM ORDERS WHERE orderkey < 100), " +
+                "      cte2 AS (SELECT custkey FROM CUSTOMER WHERE custkey < 50) " +
+                "SELECT * FROM cte1, cte2 WHERE cte1.orderkey = cte2.custkey";
+        QueryRunner queryRunner = getQueryRunner();
+        compareResults(
+                queryRunner.execute(getMaterializedSession(), testQuery),
+                queryRunner.execute(getSession(), testQuery));
+    }
+
+    @Test
+    public void testNestedPersistentCtes()
+    {
+        String testQuery = "WITH  cte1 AS (" +
+                "   SELECT orderkey FROM ORDERS WHERE orderkey IN " +
+                "       (WITH  cte2 AS (SELECT orderkey FROM ORDERS WHERE orderkey < 100) " +
+                "        SELECT orderkey FROM cte2 WHERE orderkey > 50)" +
+                ") SELECT * FROM cte1";
+        QueryRunner queryRunner = getQueryRunner();
+        compareResults(
+                queryRunner.execute(getMaterializedSession(), testQuery),
+                queryRunner.execute(getSession(), testQuery));
+    }
+
+    @Test
+    public void testRefinedCtesOutsideScope()
+    {
+        String testQuery = "WITH  cte1 AS ( WITH cte2 as (SELECT orderkey FROM ORDERS WHERE orderkey < 100)" +
+                "SELECT * FROM cte2), " +
+                " cte2 AS (SELECT * FROM customer WHERE custkey < 50) " +
+                "SELECT * FROM cte2  JOIN cte1 ON true";
+        QueryRunner queryRunner = getQueryRunner();
+        compareResults(
+                queryRunner.execute(getMaterializedSession(), testQuery),
+                queryRunner.execute(getSession(), testQuery));
+    }
+
+    @Test
+    public void testComplexRefinedCtesOutsideScope()
+    {
+        String testQuery = "WITH " +
+                "cte1 AS ( " +
+                "   SELECT orderkey, totalprice FROM ORDERS WHERE orderkey < 100 " +
+                "), " +
+                "cte2 AS ( " +
+                "   WITH cte3 AS ( WITH cte4 AS (SELECT orderkey, totalprice FROM cte1 WHERE totalprice > 1000) SELECT * FROM cte4) " +
+                "   SELECT cte3.orderkey FROM cte3 " +
+                "), " +
+                "cte3 AS ( " +
+                "   SELECT * FROM customer WHERE custkey < 50 " +
+                ") " +
+                "SELECT cte3.*, cte2.orderkey FROM cte3 JOIN cte2 ON cte3.custkey = cte2.orderkey";
+        QueryRunner queryRunner = getQueryRunner();
+        compareResults(
+                queryRunner.execute(getMaterializedSession(), testQuery),
+                queryRunner.execute(getSession(), testQuery));
+    }
+
+    @Test
+    public void testChainedPersistentCtes()
+    {
+        String testQuery = "WITH  cte1 AS (SELECT orderkey FROM ORDERS WHERE orderkey < 100), " +
+                "      cte2 AS (SELECT orderkey FROM cte1 WHERE orderkey > 50), " +
+                "     cte3 AS (SELECT orderkey FROM cte2 WHERE orderkey < 75) " +
+                "SELECT * FROM cte3";
+        QueryRunner queryRunner = getQueryRunner();
+        compareResults(
+                queryRunner.execute(getMaterializedSession(), testQuery),
+                queryRunner.execute(getSession(), testQuery));
+    }
+
+    @Test
+    public void testSimplePersistentCteWithJoinInCteDef()
+    {
+        String testQuery = "WITH  temp as " +
+                "(SELECT * FROM ORDERS o1 " +
+                "JOIN ORDERS o2 ON o1.orderkey = o2.orderkey) " +
+                "SELECT * FROM temp t1 ";
+        QueryRunner queryRunner = getQueryRunner();
+        compareResults(
+                queryRunner.execute(getMaterializedSession(), testQuery),
+                queryRunner.execute(getSession(), testQuery));
+    }
+
+    @Test
+    public void testSimplePersistentCteMultipleUses()
+    {
+        String testQuery = " WITH  temp as" +
+                " (SELECT * FROM ORDERS) " +
+                "SELECT * FROM temp t1 JOIN temp t2 on " +
+                "t1.orderkey = t2.orderkey WHERE t1.orderkey < 10";
+        QueryRunner queryRunner = getQueryRunner();
+        compareResults(
+                queryRunner.execute(getMaterializedSession(), testQuery),
+                queryRunner.execute(getSession(), testQuery));
+    }
+
+    @Test
+    public void testPersistentCteMultipleColumns()
+    {
+        String testQuery = " WITH  temp as (SELECT * FROM ORDERS) " +
+                "SELECT * FROM temp t1";
+        QueryRunner queryRunner = getQueryRunner();
+        compareResults(
+                queryRunner.execute(getMaterializedSession(), testQuery),
+                queryRunner.execute(getSession(), testQuery));
+    }
+
+    @Test
+    public void testJoinAndAggregationWithPersistentCtes()
+    {
+        String testQuery = "WITH  cte1 AS (" +
+                "   SELECT orderkey, COUNT(*) as item_count FROM lineitem" +
+                "   GROUP BY orderkey)," +
+                "    cte2 AS (" +
+                "   SELECT c.custkey, c.name FROM CUSTOMER c" +
+                "   WHERE c.mktsegment = 'BUILDING')" +
+                "   SELECT * FROM cte1" +
+                "   JOIN cte2 ON cte1.orderkey = cte2.custkey";
+        QueryRunner queryRunner = getQueryRunner();
+        compareResults(
+                queryRunner.execute(getMaterializedSession(), testQuery),
+                queryRunner.execute(getSession(), testQuery));
+    }
+
+    @Test
+    public void testNestedPersistentCtes2()
+    {
+        String testQuery = "WITH  cte1 AS (" +
+                "   WITH  cte2 AS (" +
+                "       SELECT nationkey FROM NATION" +
+                "       WHERE regionkey = 1)" +
+                "   SELECT * FROM cte2" +
+                "   WHERE nationkey < 5)" +
+                "SELECT * FROM cte1";
+        QueryRunner queryRunner = getQueryRunner();
+        compareResults(
+                queryRunner.execute(getMaterializedSession(), testQuery),
+                queryRunner.execute(getSession(), testQuery));
+    }
+
+    @Test
+    public void testPersistentCteWithUnion()
+    {
+        String testQuery = "WITH  cte AS (" +
+                "   SELECT orderkey FROM ORDERS WHERE orderkey < 100" +
+                "   UNION" +
+                "   SELECT orderkey FROM ORDERS WHERE orderkey > 500)" +
+                "SELECT * FROM cte";
+        QueryRunner queryRunner = getQueryRunner();
+        compareResults(
+                queryRunner.execute(getMaterializedSession(), testQuery),
+                queryRunner.execute(getSession(), testQuery));
+    }
+
+    @Test
+    public void testPersistentCteWithSelfJoin()
+    {
+        String testQuery = "WITH  cte AS (" +
+                "   SELECT * FROM ORDERS)" +
+                "SELECT * FROM cte c1" +
+                " JOIN cte c2 ON c1.orderkey = c2.orderkey WHERE c1.orderkey < 100";
+        QueryRunner queryRunner = getQueryRunner();
+        compareResults(
+                queryRunner.execute(getMaterializedSession(), testQuery),
+                queryRunner.execute(getSession(), testQuery));
+    }
+
+    @Test
+    public void testPersistentCteWithWindowFunction()
+    {
+        String testQuery = "WITH cte AS (" +
+                "   SELECT *, ROW_NUMBER() OVER(PARTITION BY orderstatus ORDER BY orderkey) as row" +
+                "   FROM ORDERS)" +
+                "SELECT * FROM cte WHERE row <= 5";
+        QueryRunner queryRunner = getQueryRunner();
+        compareResults(
+                queryRunner.execute(getMaterializedSession(), testQuery),
+                queryRunner.execute(getSession(), testQuery));
+    }
+
+    @Test
+    public void testPersistentCteWithMultipleDependentSubCtes()
+    {
+        String testQuery = "WITH  cte1 AS (" +
+                "   SELECT * FROM ORDERS)," +
+                "     cte2 AS (SELECT * FROM cte1 WHERE orderkey < 100)," +
+                "     cte3 AS (SELECT * FROM cte1 WHERE orderkey >= 100)" +
+                "SELECT * FROM cte2 UNION ALL SELECT * FROM cte3";
+        QueryRunner queryRunner = getQueryRunner();
+        compareResults(
+                queryRunner.execute(getMaterializedSession(), testQuery),
+                queryRunner.execute(getSession(), testQuery));
+    }
+
+    @Test
+    public void testTopCustomersByOrderValue()
+    {
+        String testQuery = "WITH  cte AS (" +
+                "   SELECT c.custkey, c.name, SUM(o.totalprice) as total_spent " +
+                "   FROM CUSTOMER c JOIN ORDERS o ON c.custkey = o.custkey " +
+                "   GROUP BY c.custkey, c.name)" +
+                "SELECT * FROM cte " +
+                "ORDER BY total_spent DESC " +
+                "LIMIT 5";
+        QueryRunner queryRunner = getQueryRunner();
+        compareResults(
+                queryRunner.execute(getMaterializedSession(), testQuery),
+                queryRunner.execute(getSession(), testQuery), true);
+    }
+
+    @Test
+    public void testSupplierDataAnalysis()
+    {
+        String testQuery = "WITH cte AS (" +
+                "   SELECT s.suppkey, s.name, n.name as nation, r.name as region, ROUND(SUM(ps.supplycost), 8)  as total_supply_cost " +
+                "   FROM partsupp ps JOIN SUPPLIER s ON ps.suppkey = s.suppkey " +
+                "                        JOIN NATION n ON s.nationkey = n.nationkey " +
+                "                        JOIN REGION r ON n.regionkey = r.regionkey " +
+                "   GROUP BY s.suppkey, s.name, n.name, r.name) " +
+                "SELECT * FROM cte " +
+                "WHERE total_supply_cost > 1000";
+        QueryRunner queryRunner = getQueryRunner();
+        compareResults(
+                queryRunner.execute(getMaterializedSession(), testQuery),
+                queryRunner.execute(getSession(), testQuery));
+    }
+
+    @Test
+    public void testCustomerOrderPatternAnalysis()
+    {
+        String testQuery = "WITH  cte AS (" +
+                "   SELECT c.name as customer_name, r.name as region_name, EXTRACT(year FROM o.orderdate) as order_year, COUNT(*) as order_count " +
+                "   FROM CUSTOMER c JOIN ORDERS o ON c.custkey = o.custkey " +
+                "                  JOIN NATION n ON c.nationkey = n.nationkey " +
+                "                  JOIN REGION r ON n.regionkey = r.regionkey " +
+                "   GROUP BY c.name, r.name, EXTRACT(year FROM o.orderdate)) " +
+                "SELECT * FROM cte " +
+                "ORDER BY customer_name, order_year";
+        QueryRunner queryRunner = getQueryRunner();
+        compareResults(
+                queryRunner.execute(getMaterializedSession(), testQuery),
+                queryRunner.execute(getSession(), testQuery));
+    }
+
+    @Test
+    public void testLowStockAnalysis()
+    {
+        String testQuery = "WITH cte AS (" +
+                "   SELECT p.partkey, p.name, p.type, SUM(ps.availqty) as total_qty " +
+                "   FROM PART p JOIN partsupp ps ON p.partkey = ps.partkey " +
+                "   GROUP BY p.partkey, p.name, p.type) " +
+                "SELECT * FROM cte " +
+                "WHERE total_qty < 100";
+        QueryRunner queryRunner = getQueryRunner();
+        compareResults(
+                queryRunner.execute(getMaterializedSession(), testQuery),
+                queryRunner.execute(getSession(), testQuery));
+    }
+
+    @Test
+    public void testComplexChainOfDependentAndNestedPersistentCtes()
+    {
+        String testQuery = "WITH  " +
+                "    cte1 AS (" +
+                "        SELECT * FROM ORDERS WHERE orderkey < 1000" +
+                "    )," +
+                "    cte2 AS (" +
+                "        SELECT * FROM cte1 WHERE custkey < 500" +
+                "    )," +
+                "     cte3 AS (" +
+                "        SELECT cte2.*, cte1.totalprice AS cte1_totalprice " +
+                "        FROM cte2 " +
+                "        JOIN cte1 ON cte2.orderkey = cte1.orderkey " +
+                "        WHERE cte1.totalprice < 150000" +
+                "    )," +
+                "     cte4 AS (" +
+                "        SELECT * FROM cte3 WHERE orderstatus = 'O'" +
+                "    )," +
+                "    cte5 AS (" +
+                "        SELECT orderkey FROM cte4 WHERE cte1_totalprice < 100000" +
+                "    )," +
+                "    cte6 AS (" +
+                "        SELECT * FROM cte5, LATERAL (" +
+                "            SELECT * FROM cte2 WHERE cte2.orderkey = cte5.orderkey" +
+                "        ) x" +
+                "    )" +
+                "SELECT * FROM cte6";
+        QueryRunner queryRunner = getQueryRunner();
+        compareResults(
+                queryRunner.execute(getMaterializedSession(), testQuery),
+                queryRunner.execute(getSession(), testQuery));
+    }
+
+    @Test
+    public void testComplexQuery1()
+    {
+        String testQuery = "WITH  customer_nation AS (" +
+                "   SELECT c.custkey, c.name, n.name AS nation_name, r.name AS region_name " +
+                "   FROM CUSTOMER c " +
+                "   JOIN NATION n ON c.nationkey = n.nationkey " +
+                "   JOIN REGION r ON n.regionkey = r.regionkey), " +
+                " customer_orders AS (" +
+                "   SELECT co.custkey, co.name, co.nation_name, co.region_name, o.orderkey, o.orderdate " +
+                "   FROM customer_nation co " +
+                "   JOIN ORDERS o ON co.custkey = o.custkey), " +
+                "order_lineitems AS (" +
+                "   SELECT co.*, l.partkey, l.quantity, l.extendedprice " +
+                "   FROM customer_orders co " +
+                "   JOIN lineitem l ON co.orderkey = l.orderkey), " +
+                " customer_part_analysis AS (" +
+                "   SELECT ol.*, p.name AS part_name, p.type AS part_type " +
+                "   FROM order_lineitems ol " +
+                "   JOIN PART p ON ol.partkey = p.partkey) " +
+                "SELECT * FROM customer_part_analysis " +
+                "WHERE region_name = 'AMERICA' " +
+                "ORDER BY nation_name, custkey, orderdate";
+        QueryRunner queryRunner = getQueryRunner();
+        compareResults(
+                queryRunner.execute(getMaterializedSession(), testQuery),
+                queryRunner.execute(getSession(), testQuery));
+    }
+
+    @Test
+    public void testComplexQuery2()
+    {
+        String testQuery = "WITH  supplier_region AS (" +
+                "   SELECT s.suppkey, s.name AS supplier_name, n.name AS nation_name, r.name AS region_name " +
+                "   FROM SUPPLIER s " +
+                "   JOIN NATION n ON s.nationkey = n.nationkey " +
+                "   JOIN REGION r ON n.regionkey = r.regionkey), " +
+                " supplier_parts AS (" +
+                "   SELECT sr.*, ps.partkey, ps.availqty, ps.supplycost " +
+                "   FROM supplier_region sr " +
+                "   JOIN partsupp ps ON sr.suppkey = ps.suppkey), " +
+                "parts_info AS (" +
+                "   SELECT sp.*, p.name AS part_name, p.type AS part_type, p.size AS part_size " +
+                "   FROM supplier_parts sp " +
+                "   JOIN PART p ON sp.partkey = p.partkey), " +
+                " full_supplier_part_info AS (" +
+                "   SELECT pi.*, n.comment AS nation_comment, r.comment AS region_comment " +
+                "   FROM parts_info pi " +
+                "   JOIN NATION n ON pi.nation_name = n.name " +
+                "   JOIN REGION r ON pi.region_name = r.name) " +
+                "SELECT * FROM full_supplier_part_info " +
+                "WHERE part_type LIKE '%BRASS' " +
+                "ORDER BY region_name, supplier_name";
+        QueryRunner queryRunner = getQueryRunner();
+        compareResults(
+                queryRunner.execute(getMaterializedSession(), testQuery),
+                queryRunner.execute(getSession(), testQuery));
+    }
+
+    @Test
+    public void testSimplePersistentCteForCtasQueries()
+    {
+        QueryRunner queryRunner = getQueryRunner();
+
+        // Create tables with Ctas
+        queryRunner.execute(getMaterializedSession(),
+                "CREATE TABLE persistent_table as (WITH  temp as (SELECT orderkey FROM ORDERS) " +
+                        "SELECT * FROM temp t1 )");
+        queryRunner.execute(getSession(),
+                "CREATE TABLE non_persistent_table as (WITH  temp as (SELECT orderkey FROM ORDERS) " +
+                        "SELECT * FROM temp t1) ");
+
+        // Compare contents with a select
+        compareResults(queryRunner.execute(getSession(),
+                        "SELECT * FROM persistent_table"),
+                queryRunner.execute(getSession(),
+                        "SELECT * FROM non_persistent_table"));
+
+        // drop tables
+        queryRunner.execute(getSession(),
+                "DROP TABLE persistent_table");
+        queryRunner.execute(getSession(),
+                "DROP TABLE non_persistent_table");
+    }
+
+    @Test
+    public void testComplexPersistentCteForCtasQueries()
+    {
+        QueryRunner queryRunner = getQueryRunner();
+        // Create tables with Ctas
+        queryRunner.execute(getMaterializedSession(),
+                "CREATE TABLE persistent_table as ( " +
+                        "WITH  supplier_region AS (" +
+                        "   SELECT s.suppkey, s.name AS supplier_name, n.name AS nation_name, r.name AS region_name " +
+                        "   FROM SUPPLIER s " +
+                        "   JOIN NATION n ON s.nationkey = n.nationkey " +
+                        "   JOIN REGION r ON n.regionkey = r.regionkey), " +
+                        " supplier_parts AS (" +
+                        "   SELECT sr.*, ps.partkey, ps.availqty, ps.supplycost " +
+                        "   FROM supplier_region sr " +
+                        "   JOIN partsupp ps ON sr.suppkey = ps.suppkey), " +
+                        "parts_info AS (" +
+                        "   SELECT sp.*, p.name AS part_name, p.type AS part_type, p.size AS part_size " +
+                        "   FROM supplier_parts sp " +
+                        "   JOIN PART p ON sp.partkey = p.partkey), " +
+                        " full_supplier_part_info AS (" +
+                        "   SELECT pi.*, n.comment AS nation_comment, r.comment AS region_comment " +
+                        "   FROM parts_info pi " +
+                        "   JOIN NATION n ON pi.nation_name = n.name " +
+                        "   JOIN REGION r ON pi.region_name = r.name) " +
+                        "SELECT * FROM full_supplier_part_info " +
+                        "WHERE part_type LIKE '%BRASS' " +
+                        "ORDER BY region_name, supplier_name)");
+        queryRunner.execute(getSession(),
+                "CREATE TABLE non_persistent_table as ( " +
+                        "WITH  supplier_region AS (" +
+                        "   SELECT s.suppkey, s.name AS supplier_name, n.name AS nation_name, r.name AS region_name " +
+                        "   FROM SUPPLIER s " +
+                        "   JOIN NATION n ON s.nationkey = n.nationkey " +
+                        "   JOIN REGION r ON n.regionkey = r.regionkey), " +
+                        " supplier_parts AS (" +
+                        "   SELECT sr.*, ps.partkey, ps.availqty, ps.supplycost " +
+                        "   FROM supplier_region sr " +
+                        "   JOIN partsupp ps ON sr.suppkey = ps.suppkey), " +
+                        "parts_info AS (" +
+                        "   SELECT sp.*, p.name AS part_name, p.type AS part_type, p.size AS part_size " +
+                        "   FROM supplier_parts sp " +
+                        "   JOIN PART p ON sp.partkey = p.partkey), " +
+                        " full_supplier_part_info AS (" +
+                        "   SELECT pi.*, n.comment AS nation_comment, r.comment AS region_comment " +
+                        "   FROM parts_info pi " +
+                        "   JOIN NATION n ON pi.nation_name = n.name " +
+                        "   JOIN REGION r ON pi.region_name = r.name) " +
+                        "SELECT * FROM full_supplier_part_info " +
+                        "WHERE part_type LIKE '%BRASS' " +
+                        "ORDER BY region_name, supplier_name)");
+
+        // Compare contents with a select
+        compareResults(queryRunner.execute(getSession(),
+                        "SELECT * FROM persistent_table"),
+                queryRunner.execute(getSession(),
+                        "SELECT * FROM non_persistent_table"));
+
+        // drop tables
+        queryRunner.execute(getSession(),
+                "DROP TABLE persistent_table");
+        queryRunner.execute(getSession(),
+                "DROP TABLE non_persistent_table");
+    }
+
+    @Test
+    public void testSimplePersistentCteForInsertQueries()
+    {
+        QueryRunner queryRunner = getQueryRunner();
+
+        // Create tables without data
+        queryRunner.execute(getSession(),
+                "CREATE TABLE persistent_table (orderkey BIGINT)");
+        queryRunner.execute(getSession(),
+                "CREATE TABLE non_persistent_table (orderkey BIGINT)");
+
+        // Insert data into tables using CTEs
+        queryRunner.execute(getMaterializedSession(),
+                "INSERT INTO persistent_table " +
+                        "WITH  temp AS (SELECT orderkey FROM ORDERS) " +
+                        "SELECT * FROM temp");
+        queryRunner.execute(getSession(),
+                "INSERT INTO non_persistent_table " +
+                        "WITH temp AS (SELECT orderkey FROM ORDERS) " +
+                        "SELECT * FROM temp");
+
+        // Compare contents with a select
+        compareResults(queryRunner.execute(getSession(),
+                        "SELECT * FROM persistent_table"),
+                queryRunner.execute(getSession(),
+                        "SELECT * FROM non_persistent_table"));
+
+        // drop tables
+        queryRunner.execute(getSession(),
+                "DROP TABLE persistent_table");
+        queryRunner.execute(getSession(),
+                "DROP TABLE non_persistent_table");
+    }
+
+    @Test
+    public void testComplexPersistentCteForInsertQueries()
+    {
+        QueryRunner queryRunner = getQueryRunner();
+        // Create tables without data
+        // Create tables
+        String createTableBase = " (suppkey BIGINT, supplier_name VARCHAR, nation_name VARCHAR, region_name VARCHAR, " +
+                "partkey BIGINT, availqty BIGINT, supplycost DOUBLE, " +
+                "part_name VARCHAR, part_type VARCHAR, part_size BIGINT, " +
+                "nation_comment VARCHAR, region_comment VARCHAR)";
+
+        queryRunner.execute(getSession(),
+                "CREATE TABLE persistent_table" + createTableBase);
+
+        queryRunner.execute(getSession(),
+                "CREATE TABLE non_persistent_table" + createTableBase);
+
+        queryRunner.execute(getMaterializedSession(),
+                "INSERT INTO persistent_table  " +
+                        "WITH  supplier_region AS (" +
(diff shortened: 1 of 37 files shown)

### CODE REVIEW

**Comment by mlyublena** [comment]
General comment: I wonder if it would have been possible to achieve the same without introducing new syntax (for example through a session parameter). I know if't not as flexible as allowing specification per CTE but also not as disruptive as new syntax

File: presto-spi/src/main/java/com/facebook/presto/spi/plan/SequenceNode.java [hunk 1]
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    @JsonProperty
+    public List<PlanNode> getLeftSources()
+    {
+        return this.leftSources;
+    }
+
+    @JsonProperty
+    public PlanNode getRight()
```
mlyublena, line 59: I think the names leftSource and right are a bit confusing.
Open to suggestion on how to name these better, but perhaps something like this:
- leftSources -> cteDefinitions or cteProducers
- right -> ???

jaystarshot:
> General comment: I wonder if it would have been possible to achieve the same without introducing new syntax (for example through a session parameter). I know if't not as flexible as allowing specification per CTE but also not as disruptive as new syntax

The temporary tables will be registered in the in-memory SemiTransactionalMetastore. The data will be removed at the end of transaction (end of a query execution). There could be multiple queries in the same session so session properties naively  won't work. 
Other than that, I feel like this is the easiest way for users to migrate their existing queries themselves and try out this feature. We could do something like the hive Create temporary table -[link](https://sparkbyexamples.com/apache-hive/hive-temporary-table-usage-and-how-to-create-2/) syntax but i feel like this syntax is easier to migrate and clear to use.

presto-spi/src/main/java/com/facebook/presto/spi/plan/SequenceNode.java
jaystarshot: Agreed

jaystarshot: > right -> source
This might be confusing since getSources needs to return leftsources + right. Maybe primarySource?

mlyublena: primarySource sounds good to me

steveburnett (approved):
LGTM! (docs) 

Local build of docs, formatting looks good.

mlyublena:
Is the cleanup of the materialized result happening if something crashes later on or the query is canceled after the CTE is materialized? Are there tests for that?

jaystarshot:
Yes i confirmed that the cleanup is indeed happening if the query is cancelled or fails. Also confirmed that the cleanup does not happen if the server crashes.
There are test cases added with [this](https://github.com/prestodb/presto/pull/12605/files#diff-85f3464057b643d587def1f5b64d640fa91693018b74935a8ae492adc3da546e) commit but i will check if they are robust enough/ or need more tests.

mlyublena:
> > General comment: I wonder if it would have been possible to achieve the same without introducing new syntax (for example through a session parameter). I know if't not as flexible as allowing specification per CTE but also not as disruptive as new syntax
> 
> The temporary tables will be registered in the in-memory SemiTransactionalMetastore. The data will be removed at the end of transaction (end of a query execution). There could be multiple queries in the same session so session properties naively won't work. Other than that, I feel like this is the easiest way for users to migrate their existing queries themselves and try out this feature. We could do something like the hive Create temporary table -[link](https://sparkbyexamples.com/apache-hive/hive-temporary-table-usage-and-how-to-create-2/) syntax but i feel like this syntax is easier to migrate and clear to use. Open to ideas

I meant instead of asking users to explicitly name persisted CTEs in their queries, just have a session flag to do that automatically, for example "set session materialize_ctes=true"
That way we don't introduce new syntax + users don't need to rewrite their queries. The difference to your approach is that all CTEs are either materialized or not materialized and users cannot apply this selectively for only some CTEs in their query.
We can have the auto-materializer be smart about not materializing trivial CTEs (for example a single table scan) which would probably cover most of the cases where we want to materialize some CTEs only.

mlyublena:
> Yes i confirmed that the cleanup is indeed happening if the query is cancelled or fails. Also confirmed that the cleanup does not happen if the server crashes. There are test cases added with [this](https://github.com/prestodb/presto/commit/e5b5403494956780dcfd3b9389130176f1303b94#diff-97caffd1bf83a7671054a8dbd37f2d7a7aebeba74d2cc02f976b4adbb2cce8ae) commit in the presto unlimited framework but i will check if they are robust enough/ or need more tests.

what is the plan for cleaning up if the server crashes?

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/LogicalCteOptimizer.java
```diff
@@ -0,0 +1,222 @@
...
+        {
+            this.cteInformationCollector = cteInformationCollector;
+            this.planNodeIdAllocator = planNodeIdAllocator;
+            this.variableAllocator = variableAllocator;
+        }
+
+        public PlanNode transformPersistentCtes(PlanNode root)
+        {
+            // ToDO: Proper fix is to add a new project of the child
+            checkArgument(root.getSources().size() == 1, "expected newChildren to contain 1node");
```
mlyublena, line 73: nit: space after 1

presto-hive/src/test/java/com/facebook/presto/hive/TestCteExecution.java
```diff
@@ -0,0 +1,560 @@
+/*
```
mlyublena, line 1: can you also add other types of queries like INSERT and create table, e.g.:
```
insert into r(a) with cte as (select 1) select * from cte;
 create table t as (with cte as (select 1 as a) select * from cte);
```

mlyublena: also create view:
```
create view v as with cte as (select 1 as a) select * from cte;
```
What is the semantics of persisted CTEs in views? We should avoid persisting them during view definition

jaystarshot:
> > > General comment: I wonder if it would have been possible to achieve the same without introducing new syntax (for example through a session parameter). I know if't not as flexible as allowing specification per CTE but also not as disruptive as new syntax
> > 
> > 
> > The temporary tables will be registered in the in-memory SemiTransactionalMetastore. The data will be removed at the end of transaction (end of a query execution). There could be multiple queries in the same session so session properties naively won't work. Other than that, I feel like this is the easiest way for users to migrate their existing queries themselves and try out this feature. We could do something like the hive Create temporary table -[link](https://sparkbyexamples.com/apache-hive/hive-temporary-table-usage-and-how-to-create-2/) syntax but i feel like this syntax is easier to migrate and clear to use. Open to ideas
> 
> I meant instead of asking users to explicitly name persisted CTEs in their queries, just have a session flag to do that automatically, for example "set session materialize_ctes=true" That way we don't introduce new syntax + users don't need to rewrite their queries. The difference to your approach is that all CTEs are either materialized or not materialized and users cannot apply this selectively for only some CTEs in their query. We can have the auto-materializer be smart about not materializing trivial CTEs (for example a single table scan) which would probably cover most of the cases where we want to materialize some CTEs only.

 if we could potentially implement both strategies. I am wondering if we Might we also introduce the syntax as an optional feature for those who prefer to make explicit choices?

jaystarshot:
> > Yes i confirmed that the cleanup is indeed happening if the query is cancelled or fails. Also confirmed that the cleanup does not happen if the server crashes. There are test cases added with [this](https://github.com/prestodb/presto/commit/e5b5403494956780dcfd3b9389130176f1303b94#diff-97caffd1bf83a7671054a8dbd37f2d7a7aebeba74d2cc02f976b4adbb2cce8ae) commit in the presto unlimited framework but i will check if they are robust enough/ or need more tests.
> 
> what is the plan for cleaning up if the server crashes?

The current plan is to have a ttl on the HDFS side and later a backup job that performs a cleanup. I am wondering if Meta already has that for presto unlimited

presto-hive/src/test/java/com/facebook/presto/hive/TestCteExecution.java
jaystarshot: Makes sense, I will add concrete test cases for insert, ctas and views. Will check for views, I think we only save view definition so mostly should be fine (will confirm).

jaystarshot: Confirmed that only definiations are stored with create views and added tests for all cases

tdcmeehan:
I agree with @mlyublena's comment.  Basically, whether or not the CTE is persistent is a detail that I think doesn't belong in the syntax, but as a hint to the engine.  I also agree with the example that some queries that currently may be "persistent" may in the future spool in memory and automatically choose to become persistent based on the size of the output.  To me, this suggests what's needed is more like a hint, which is well served by a session property.

AFAIA, it's also not a part of the spec, and so we need to think carefully if this is the **only** way to accomplish this feature.  I'm not convinced yet that this is the only way to go about the desired outcome we want, and so I'm in favor of proceeding conservatively especially when it comes to syntax changes.  Session properties can be deprecated, but syntax changes are much harder to migrate and deprecate.

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/LogicalCteOptimizer.java
```diff
@@ -0,0 +1,221 @@
...
+            // DFS will generate reverse order
+            Collections.reverse(topSortList);
+            List<PlanNode> topSortedCteProducerList = new ArrayList<>();
+            for (String cteName : topSortList) {
+                topSortedCteProducerList.add(cteProducerMap.get(cteName));
+            }
+            return topSortedCteProducerList;
+        }
+
+        private void dfs(String node, List<String> topSortList, Map<String, Integer> visitedMap)
```
tdcmeehan, line 204: You'll find examples in the codebase that we use the Guava Traverser utility to do a DFS, wondering if we can refactor this to use it?  May reduce the LOCs.

tdcmeehan (changes requested):
Just some surface review, will continue in depth.

presto-docs/src/main/sphinx/language/reserved.rst
```diff
@@ -78,4 +78,5 @@ Keyword                        SQL:2016      SQL-92
 ``WHEN``                       reserved      reserved
 ``WHERE``                      reserved      reserved
 ``WITH``                       reserved      reserved
+``PERSISTENT``
```
mlyublena, line 81: I'd vote for no new syntax (but if the syntax stays, this list is alphabetically ordered)

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/LogicalCteOptimizer.java
jaystarshot: I had initially used the guava implementation but detecting cycle there is nontrivial and unclean. I think LOC is an anti goal

jaystarshot:
Okay if that is the consensus,  I will remove the syntax and add a session property to materialize all CTEs in a query. The end goal anyway is to do this automatically

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/LogicalCteOptimizer.java
jaystarshot: Actually I should probably check recursive ctes and add test cases if possible

jaystarshot: Recursive ctes are not supported in Presto

jaystarshot: Actually I found a way to change the graph to use MutableGraph (marked unstable) and the complexity reduces significantly, please rereview

presto-main/src/main/java/com/facebook/presto/sql/planner/BasePlanFragmenter.java
```diff
@@ -383,240 +394,6 @@ private PlanNode createRemoteMaterializedExchange(ExchangeNode exchange, Rewrite
         return visitTableScan(scan, context);
     }
 
-    private PartitioningVariableAssignments assignPartitioningVariables(Partitioning partitioning)
```
jaystarshot, line 386: Moved to https://github.com/prestodb/presto/pull/20887/files#diff-6b68e5660ffcbb5b4436b88cf5c58be0a90a1117a0b42d1eec80cf1d076f08e6

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/LogicalCteOptimizer.java
jaystarshot: But we would need the cycle detection logic when we do the cost based replacement to filter out unwanted plans so i might use the old implementation there anyways

tdcmeehan:
CC: @feilong-liu, would you like to have a look as well?

presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
```diff
@@ -2280,6 +2286,11 @@ public static DataSize getFilterAndProjectMinOutputPageSize(Session session)
         return session.getSystemProperty(FILTER_AND_PROJECT_MIN_OUTPUT_PAGE_SIZE, DataSize.class);
     }
 
+    public static boolean isMaterializeAllCtes(Session session)
```
feilong-liu, line 2289: Instead of having a boolean session property, perhaps we can have a enum to offer more flexibility, like rename the property to CTE_MATERIALIZE_STRATEGY and offer ALL, COST_BASED, NONE etc options. So that in the end, we do not need to introduce another session property when introducing cost based materialization strategy.

presto-spi/pom.xml
```diff
@@ -88,7 +88,7 @@
         <dependency>
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
-            <scope>test</scope>
+            <scope>provided</scope>
```
feilong-liu, line 91: noob question, why this change?

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/DetermineJoinDistributionType.java
```diff
@@ -198,7 +200,7 @@ static double getSourceTablesSizeInBytes(PlanNode node, Lookup lookup, StatsProv
         }
 
         List<PlanNode> sourceNodes = PlanNodeSearcher.searchFrom(node, lookup)
-                .whereIsInstanceOfAny(ImmutableList.of(TableScanNode.class, ValuesNode.class, RemoteSourceNode.class))
+                .whereIsInstanceOfAny(ImmutableList.of(TableScanNode.class, ValuesNode.class, RemoteSourceNode.class, CteProducerNode.class, CteConsumerNode.class))
```
feilong-liu, line 203: Why CteProducerNode is also source nodes?

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/AddExchanges.java
```diff
@@ -616,6 +618,26 @@ public PlanWithProperties visitFilter(FilterNode node, PreferredProperties prefe
             return rebaseAndDeriveProperties(node, planChild(node, preferredProperties));
         }
 
+        @Override
+        public PlanWithProperties visitSequence(SequenceNode node, PreferredProperties preferredProperties)
+        {
+            List<PlanWithProperties> leftPlans = node.getCteProducers().stream()
+                    .map(source -> accept(source, PreferredProperties.any()))
+                    .collect(ImmutableList.toImmutableList());
```
feilong-liu, line 626: nit: ImmutableList.toImmutableList() -> toImmutableList(), similar for other places

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/AddExchanges.java
```diff
@@ -616,6 +618,26 @@ public PlanWithProperties visitFilter(FilterNode node, PreferredProperties prefe
...
+            PlanWithProperties rightPlan = accept(node.getPrimarySource(), PreferredProperties.any());
```
feilong-liu, line 627: Why right plan use `PreferredProperties.any()`, shouldn't it conform to what we are supposed to use for the primary source node?

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/AddExchanges.java
```diff
@@ -616,6 +618,26 @@ public PlanWithProperties visitFilter(FilterNode node, PreferredProperties prefe
...
+            List<PlanNode> childrenNodes = Stream.concat(
+                    leftPlans.stream().map(PlanWithProperties::getNode),
+                    Stream.of(rightPlan.getNode())
+            ).collect(ImmutableList.toImmutableList());
+            List<ActualProperties> actualPropertiesList = Stream.concat(
+                    leftPlans.stream().map(PlanWithProperties::getProperties),
+                    Stream.of(rightPlan.getProperties())
+            ).collect(ImmutableList.toImmutableList());
```
feilong-liu, line 635: The output property of the sequence node should be the property of the primary source (right plan) only?

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/HashGenerationOptimizer.java
```diff
@@ -162,6 +163,21 @@ public PlanWithProperties visitApply(ApplyNode node, HashComputationSet context)
             return new PlanWithProperties(node, ImmutableMap.of());
         }
 
+        public PlanWithProperties visitSequence(SequenceNode node, HashComputationSet context)
+        {
+            List<PlanNode> cteProducers = node.getCteProducers().stream()
+                    .map(c ->
+                            planAndEnforce(c, new HashComputationSet(), true, new HashComputationSet()).getNode())
+                    .collect(ImmutableList.toImmutableList());
+            PlanWithProperties primarySource = planAndEnforce(node.getPrimarySource(), new HashComputationSet(), true, new HashComputationSet());
```
feilong-liu, line 172: Shouldn't it respect what the primary source node expect?

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/ApplyConnectorOptimization.java
```diff
@@ -55,6 +58,9 @@ public class ApplyConnectorOptimization
         implements PlanOptimizer
 {
     static final Set<Class<? extends PlanNode>> CONNECTOR_ACCESSIBLE_PLAN_NODES = ImmutableSet.of(
+            CteProducerNode.class,
+            CteConsumerNode.class,
+            CteReferenceNode.class,
```
feilong-liu, line 63: why this change?

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/LogicalCteOptimizer.java
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    @Override
+    public PlanOptimizerResult optimize(PlanNode plan, Session session, TypeProvider types, VariableAllocator variableAllocator, PlanNodeIdAllocator idAllocator, WarningCollector warningCollector)
+    {
+        if (!isMaterializeAllCtes(session)) {
+            return PlanOptimizerResult.optimizerResult(plan, false);
+        }
+        PlanNode rewrittenPlan = new CteEnumerator(session.getCteInformationCollector(), idAllocator, variableAllocator).transformPersistentCtes(plan);
+        return PlanOptimizerResult.optimizerResult(rewrittenPlan, !rewrittenPlan.equals(plan));
```
feilong-liu, line 61: nit: simple rewrittenPlan.equals(plan) may introduce false positive

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/LogicalCteOptimizer.java
```diff
@@ -0,0 +1,197 @@
...
+        public CteConsumerTransformer(PlanNodeIdAllocator idAllocator, VariableAllocator variableAllocator)
+        {
+            this.idAllocator = idAllocator;
+            this.variableAllocator = variableAllocator;
+        }
+
+        @Override
+        public PlanNode visitCteReference(CteReferenceNode node, RewriteContext<CteTransformerContext> context)
+        {
+            context.get().addDependency(node.getCteName());
```
feilong-liu, line 109: Is the name returned by `node.getCteName()` unique to identify the CTE here? 
For example, will it be able to differentiate the two `t2` in the query here?
`with t1 as (with t2 as (select * from (values 1, 2)t(k)) select k+1 as k from t2), t2 as (select k+1 as k from t1) select k+1 from t2;`

jaystarshot: Hmm interesting, I am not sure if this is allowed in Sql Standard.

jaystarshot: I think nested ctes are not allowed in SQL standard, but for us it makes sense to add a test case/fix for this

presto-spi/pom.xml
jaystarshot: Used com.google.common.collect.ImmutableList etc in CteConsumer which is defined in presto-spi. Before this guava was not used in presto-spi

presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
jaystarshot: Makes sense

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/DetermineJoinDistributionType.java
jaystarshot: Nice catch, yes unintended

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/ApplyConnectorOptimization.java
jaystarshot: For connector optimizers to apply in cases where plan contains these nodes

presto-spi/pom.xml
tdcmeehan: Let's not add in Guava or presume it's needed in the SPI--let's figure out another way (unmodifiableCollection perhaps)

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/LogicalCteOptimizer.java
feilong-liu: Not sure about SQL standard (maybe @kaikalur know it), but it's a valid SQL which can run in Presto

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/LogicalCteOptimizer.java
```diff
@@ -0,0 +1,197 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static com.facebook.presto.SystemSessionProperties.isMaterializeAllCtes;
+import static com.facebook.presto.common.type.BigintType.BIGINT;
+import static com.google.common.base.Preconditions.checkArgument;
+
```
feilong-liu, line 44: Can you add some comment and including an example about before and after of query plan?

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/PhysicalCteOptimizer.java
```diff
@@ -0,0 +1,259 @@
...
+import static com.facebook.presto.SystemSessionProperties.getHashPartitionCount;
+import static com.facebook.presto.SystemSessionProperties.isMaterializeAllCtes;
+import static com.facebook.presto.spi.StandardErrorCode.NOT_SUPPORTED;
+import static com.facebook.presto.sql.TemporaryTableUtil.assignPartitioningVariables;
+import static com.facebook.presto.sql.TemporaryTableUtil.assignTemporaryTableColumnNames;
+import static com.facebook.presto.sql.TemporaryTableUtil.createTemporaryTableScan;
+import static com.facebook.presto.sql.TemporaryTableUtil.createTemporaryTableWriteWithoutExchanges;
+import static com.google.common.collect.ImmutableList.toImmutableList;
+import static java.lang.String.format;
+
```
feilong-liu, line 56: Can you add some comment and including an example about before and after of query plan?

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/PhysicalCteOptimizer.java
```diff
@@ -0,0 +1,259 @@
...
+    @Override
+    public PlanOptimizerResult optimize(PlanNode plan, Session session, TypeProvider types, VariableAllocator variableAllocator, PlanNodeIdAllocator idAllocator, WarningCollector warningCollector)
+    {
+        if (!isMaterializeAllCtes(session)) {
+            return PlanOptimizerResult.optimizerResult(plan, false);
+        }
+        PhysicalCteTransformerContext context = new PhysicalCteTransformerContext();
+        PlanNode producerReplaced = SimplePlanRewriter.rewriteWith(new CteProducerRewiter(session, idAllocator, variableAllocator), plan, context);
+        PlanNode rewrittenPlan = SimplePlanRewriter.rewriteWith(new CteConsumerRewrite(session, idAllocator, variableAllocator), producerReplaced, context);
+        return PlanOptimizerResult.optimizerResult(rewrittenPlan, !rewrittenPlan.equals(plan));
```
feilong-liu, line 76: `rewrittenPlan.equals(plan)` may not be a good way, which involves false positives

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/PhysicalCteOptimizer.java
```diff
@@ -0,0 +1,259 @@
...
+        {
+            // Create Table Metadata
+            PhysicalCteTransformerContext.TemporaryTableInfo tableInfo = context.get().getTableInfo(node.getCteName());
+            TableScanNode tempScan = tableInfo.getTableScanNode();
+
+            // Need to create new Variables for temp table scans to avoid duplicate references
+            List<VariableReferenceExpression> newOutputVariables = new ArrayList<>();
+            Map<VariableReferenceExpression, ColumnHandle> newColumnAssignmentsMap = new HashMap<>();
+
+            for (VariableReferenceExpression var : tempScan.getOutputVariables()) {
```
feilong-liu, line 187: var -> variable, we generally avoid abbreviation in code

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/LogicalCteOptimizer.java
```diff
@@ -0,0 +1,197 @@
...
+                    idAllocator.getNextId(),
+                    actualSource,
+                    node.getCteName(),
+                    variableAllocator.newVariable("rows", BIGINT), node.getOutputVariables());
+            context.get().addProducer(node.getCteName(), cteProducerSource);
+            return new CteConsumerNode(node.getSourceLocation(), idAllocator.getNextId(), actualSource.getOutputVariables(), node.getCteName());
+        }
+
+        @Override
+        public PlanNode visitApply(ApplyNode node, RewriteContext<CteTransformerContext> context)
```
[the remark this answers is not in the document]
jaystarshot: Because for apply we also need to rewrite subquery using the same context for correlated queries etc.
eg - [test](https://github.com/prestodb/presto/blob/20f0a66a2118eee1d55a12b369df395fd97ff8ab/presto-hive/src/test/java/com/facebook/presto/hive/TestCteExecution.java#L101) case

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/LogicalCteOptimizer.java
```diff
@@ -0,0 +1,197 @@
...
+
+        // a -> b indicates that b needs to be processed before a
+        MutableGraph<String> graph;
+
+        public String activeCte;
+
+        public CteTransformerContext()
+        {
+            cteProducerMap = new HashMap<>();
+            // The cte graph will never have cycles because sql won't allow it
```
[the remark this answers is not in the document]
jaystarshot: I need to revert to my original dfs implementation because its not trivial to add a cycle check, since mutable graph expects the graph to be directed. But during review it was observed that that implementation added unnecessary complexity

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/HashGenerationOptimizer.java
jaystarshot: It does respect the expectations set by the primary source node. Therefore, it does not provide any parentPreference when calling planAndEnforce. This approach is consistent with the logic found in visitJoin(), where it processes the left and right children without altering the inherited preferences

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/AddExchanges.java
jaystarshot: Yes I think this makes sense

presto-spi/pom.xml
jaystarshot: Any benefit in doing so?

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/LogicalCteOptimizer.java
```diff
@@ -0,0 +1,197 @@
...
+        }
+
+        public Map<String, CteProducerNode> getCteProducerMap()
+        {
+            return cteProducerMap;
+        }
+
+        public void addProducer(String cteName, CteProducerNode cteProducer)
+        {
+            cteProducerMap.putIfAbsent(cteName, cteProducer);
```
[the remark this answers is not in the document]
jaystarshot: You mean an assertion based on output variables being same or something? 
the output variables need not be the same (different projections) and I can only add a ouput variables.length() assertion so the value is limited

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/LogicalCteOptimizer.java
```diff
@@ -0,0 +1,197 @@
...
+                            context.get()),
+                    node.getSubqueryAssignments(),
+                    node.getCorrelation(),
+                    node.getOriginSubqueryError(),
+                    node.getMayParticipateInAntiJoin());
+        }}
+
+    public class CteTransformerContext
+    {
+        private static final String INACTIVE = "";
```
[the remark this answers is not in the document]
jaystarshot: Changed implementation to use stack and no longer need this

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/LogicalCteOptimizer.java
jaystarshot: the transform function returns the original plannode if untriggered

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/PhysicalCteOptimizer.java
jaystarshot: In this case, there are no rewrites if no cte producers or consumers are present and can be used

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/LogicalCteOptimizer.java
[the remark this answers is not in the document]
feilong-liu: I mean for the same cteName, can we assume there is only one cteProducer?  If yes, maybe assert that we will not have the same cteName to map to multiple different cteProducer nodes.

jaystarshot:
sidenote- Postgres supports MATERIALIZED syntax too - [link](https://www.postgresql.org/docs/current/queries-with.html#QUERIES-WITH-CTE-MATERIALIZATION)!

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/HashGenerationOptimizer.java
feilong-liu: I was thinking about something like:
`PlanWithProperties primarySource = plan(node.getPrimarySource(), context)`

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/PruneUnreferencedOutputs.java
```diff
@@ -502,6 +505,39 @@ public PlanNode visitFilter(FilterNode node, RewriteContext<Set<VariableReferenc
...
+        {
+            // Some output can be pruned but current implementation of PhysicalCteProducer does not allow cteconsumer pruning
+            return node;
+        }
+
+        @Override
+        public PlanNode visitCteProducer(CteProducerNode node, RewriteContext<Set<VariableReferenceExpression>> context)
+        {
+            Set<VariableReferenceExpression> expectedInputs = new HashSet<>();
+            expectedInputs.addAll(node.getOutputVariables());
```
feilong-liu, line 519: Can be combined into one line `Set<VariableReferenceExpression> expectedInputs = ImmutableSet.copyOf(node.getOutputVariables());`

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/PruneUnreferencedOutputs.java
```diff
@@ -502,6 +505,39 @@ public PlanNode visitFilter(FilterNode node, RewriteContext<Set<VariableReferenc
...
+        public PlanNode visitSequence(SequenceNode node, RewriteContext<Set<VariableReferenceExpression>> context)
+        {
+            ImmutableSet.Builder<VariableReferenceExpression> cteProducersBuilder = ImmutableSet.builder();
+            node.getCteProducers().forEach(leftSource -> cteProducersBuilder.addAll(leftSource.getOutputVariables()));
+
+            Set<VariableReferenceExpression> leftInputs = cteProducersBuilder.build();
+
+            ImmutableSet.Builder<VariableReferenceExpression> primarySourceBuilder = ImmutableSet.builder();
+            primarySourceBuilder.addAll(node.getPrimarySource().getOutputVariables());
+            Set<VariableReferenceExpression> rightInputs = primarySourceBuilder.build();
```
feilong-liu, line 534: `Set<VariableReferenceExpression> rightInputs = ImmutableSet.copyOf(node.getPrimarySource().getOutputVariables())`

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/PruneUnreferencedOutputs.java
```diff
@@ -502,6 +505,39 @@ public PlanNode visitFilter(FilterNode node, RewriteContext<Set<VariableReferenc
...
+
+            List<PlanNode> cteProducers = node.getCteProducers().stream().map(leftSource -> context.rewrite(leftSource, leftInputs)).collect(Collectors.toList());
```
feilong-liu, line 536: Consider moving up to immediately below the collection of leftInputs, so that the same logic code block stay closer.

feilong-liu: context.rewrite(leftSource, ImmutableSet.of()) is enough? leftInputs seems to be redundant

feilong-liu:
> sidenote- Postgres supports MATERIALIZED syntax too - [link](https://www.postgresql.org/docs/current/queries-with.html#QUERIES-WITH-CTE-MATERIALIZATION)!

I think we can leave this as a next step, for now we can go with the session property way.

feilong-liu (commented):
Still reviewing the second commit

presto-main/src/main/java/com/facebook/presto/sql/planner/BasePlanFragmenter.java
```diff
@@ -223,6 +201,31 @@ public PlanNode visitTableFinish(TableFinishNode node, RewriteContext<FragmentPr
         return context.defaultRewrite(node, context.get());
     }
 
+    @Override
+    public PlanNode visitSequence(SequenceNode node, RewriteContext<FragmentProperties> context)
+    {
+        //Since this is topologically sorted by the LogicalCtePlanner, need to make sure that execution order follows
```
mlyublena, line 207: nit: space after //

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/PruneUnreferencedOutputs.java
jaystarshot: Need to populate the rewrite context with the current anticipated variables that the visitor aims to extract from the children.

presto-main/src/main/java/com/facebook/presto/sql/planner/BasePlanFragmenter.java
feilong-liu: Why move it?

feilong-liu (commented):
Finish first round of review.

presto-main/src/main/java/com/facebook/presto/sql/planner/BasePlanFragmenter.java
jaystarshot: By utilizing similar code in PhysicalCteOptimizer, we can reduce redundancy through consolidation. Moreover, these utility functions are unrelated to the fragmenter

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/PushdownSubfields.java
```diff
@@ -284,6 +285,13 @@ public PlanNode visitOutput(OutputNode node, RewriteContext<Context> context)
             return context.defaultRewrite(node, context.get());
         }
 
+        @Override
+        public PlanNode visitCteProducer(CteProducerNode node, RewriteContext<Context> context)
```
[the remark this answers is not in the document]
jaystarshot: CteProducer functions as a data source, yet it requires the context of the output node to be inputted. I tried with a new visitor to see if it would work, but most of the tests failed

presto-main/src/main/java/com/facebook/presto/sql/planner/SqlPlannerContext.java
```diff
@@ -49,4 +63,58 @@ public void incrementLeafNodes(Session session)
             }
         }
     }
+
+    public class NestedCteStack
```
jaystarshot, line 67: NestedCteStack is responsible for managing the scope of the CTE
 cteStack keeps track of the current CTEs that are accessible outside this scope.
 The rawCtePathMap holds the localCtes along with their respective scopes, keeping them updated.

presto-main/src/main/java/com/facebook/presto/sql/planner/planPrinter/PlanPrinter.java
```diff
@@ -829,6 +833,34 @@ public Void visitTableScan(TableScanNode node, Void context)
...
+            addNode(node, "CteProducer");
+            return processChildren(node, context);
+        }
+
+        @Override
+        public Void visitCteReference(CteReferenceNode node, Void context)
+        {
+            addNode(node, "CteReference");
+            return processChildren(node, context);
+        }
```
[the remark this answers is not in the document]
jaystarshot: Added

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/HashGenerationOptimizer.java
jaystarshot: Yeah i think this works too and makes more sense

presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
feilong-liu: Better to return the strategy, and have check in optimizer, rather than do the boolean check here. You can refer to other similar session properties as a reference.

presto-main/src/main/java/com/facebook/presto/sql/planner/BasePlanFragmenter.java
```diff
@@ -223,6 +201,31 @@ public PlanNode visitTableFinish(TableFinishNode node, RewriteContext<FragmentPr
         return context.defaultRewrite(node, context.get());
     }
 
+    @Override
+    public PlanNode visitSequence(SequenceNode node, RewriteContext<FragmentProperties> context)
+    {
+        // Since this is topologically sorted by the LogicalCtePlanner, need to make sure that execution order follows
+        // Can be optimized further to avoid non dependents from getting blocked
+        int n = node.getCteProducers().size();
+        Preconditions.checkArgument(n >= 1, "Sequence Node has 0 cte producers");
```
feilong-liu, line 210: `Preconditions.checkArgument` -> `checkArgument`

presto-spi/pom.xml
feilong-liu: I think we need to add as little dependency as possible for SPI, let's remove the dependency here.

feilong-liu (commented):
Finished my review. My review mostly focused on optimizer and query plan change etc. and not as thorough for the temp table materialization and plan fragment generation part.

The only main suggestion left from my side is to remove guava dependency from SPI. It otherwise looks good to me except a few nits.

presto-spi/pom.xml
jaystarshot: I think I am not adding a new dependency here, I am just marking it provided, which means that its already available at runtime and just marking it available during compile time. What do you think? If not I can remove

tdcmeehan: The problem with marking it provided is people will need to supply this dependency independently.  Suppose someone is creating a custom plugin--if their provided dependency conflicts with the one expected by the compiler, then it will generate runtime errors.  It's best to minimize SPI dependencies as much as possible.

jaystarshot: Okay, i will remove it

feilong-liu (approved):
Please combine the "nit fix" commits into the original commit rather than being standalone commits before merging.

presto-main/src/main/java/com/facebook/presto/sql/TemporaryTableUtil.java
```diff
@@ -0,0 +1,390 @@
...
+        return result.build();
+    }
+
+    public static BasePlanFragmenter.PartitioningVariableAssignments assignPartitioningVariables(VariableAllocator variableAllocator,
+            Partitioning partitioning)
+    {
+        ImmutableList.Builder<VariableReferenceExpression> variables = ImmutableList.builder();
+        ImmutableMap.Builder<VariableReferenceExpression, RowExpression> constants = ImmutableMap.builder();
+        for (RowExpression argument : partitioning.getArguments()) {
+            checkArgument(argument instanceof ConstantExpression || argument instanceof VariableReferenceExpression, format("Expect argument to be ConstantExpression or VariableReferenceExpression, get %s (%s)", argument.getClass(), argument));
```
mlyublena, line 148: nit: grammar: get -> got

presto-main/src/main/java/com/facebook/presto/sql/TemporaryTableUtil.java
```diff
@@ -0,0 +1,390 @@
...
+            List<VariableReferenceExpression> outputs,
+            List<List<VariableReferenceExpression>> inputs,
+            List<PlanNode> sources,
+            Map<VariableReferenceExpression, RowExpression> constantExpressions,
+            PartitioningMetadata partitioningMetadata)
+    {
+        if (!constantExpressions.isEmpty()) {
+            List<VariableReferenceExpression> constantVariables = ImmutableList.copyOf(constantExpressions.keySet());
+
+            // update outputs
```
mlyublena, line 238: nit: comments are probably redundant here

presto-main/src/main/java/com/facebook/presto/sql/planner/BasePlanFragmenter.java
```diff
@@ -223,6 +201,31 @@ public PlanNode visitTableFinish(TableFinishNode node, RewriteContext<FragmentPr
         return context.defaultRewrite(node, context.get());
     }
 
+    @Override
+    public PlanNode visitSequence(SequenceNode node, RewriteContext<FragmentProperties> context)
+    {
+        // Since this is topologically sorted by the LogicalCtePlanner, need to make sure that execution order follows
+        // Can be optimized further to avoid non dependents from getting blocked
+        int n = node.getCteProducers().size();
+        checkArgument(n >= 1, "Sequence Node has 0 cte producers");
```
mlyublena, line 210: nit: since "CTE" is an abbreviation, can we use it consistently capitalized in error and warning messages displayed to the user

presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java
```diff
@@ -558,6 +566,19 @@ public FeaturesConfig setLegacyMapSubscript(boolean value)
         return this;
     }
 
+    public CteMaterializationStrategy getCteMaterializationStrategy()
+    {
+        return cteMaterializationStrategy;
+    }
+
+    @Config("cte-materialization-strategy")
+    @ConfigDescription("Set strategy used to determine whether to materialize ctes (ALL, NONE)")
```
mlyublena, line 575: nit: capitalize CTE in description

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/CteUtils.java
```diff
@@ -0,0 +1,76 @@
...
+        }
+        throw new PrestoException(GENERIC_INTERNAL_ERROR, "No Partitioning index found");
+    }
+
+    /*
+        Currently, Hive bucketing does not support the Presto type 'ROW'.
+    */
+    public static boolean isVariableMaterializable(VariableReferenceExpression var)
+    {
+        return !(var.getType() instanceof RowType);
```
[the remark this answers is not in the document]
mlyublena: also, can we add tests for these types?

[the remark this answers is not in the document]
jaystarshot: The recommended table format is **Pagefile** which doesn't require translation to hivetypes. Also, there special handling already that uses the format _pagefile_ for these types that are not supported by hive. 
https://github.com/prestodb/presto/blob/45bcbcf5e852e62339a859472ce80fc3a1e841d8/presto-hive/src/main/java/com/facebook/presto/hive/HiveMetadata.java#L1091

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/LogicalCteOptimizer.java
mlyublena: What @feilong-liu is referring to is that we had cases in other optimizers, where a part of the subplan was changed in an equivalent way (for example reordering of columns at intermediate nodes), and the equality check returned false even if the plans were equivalent. To avoid false negatives, we changed all PlanOptimizers to return PlanOptimizerResult which has a boolean flag whether the transform actually triggered or not.
Perhaps you can reuse that here to make the intention more obvious

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/LogicalCteOptimizer.java
mlyublena: please also check the case with view names: I believe a query can reference a view and a cte with the same name and there are some rules around how name resolution is done

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/LogicalCteOptimizer.java
```diff
@@ -0,0 +1,219 @@
...
+            CteTransformerContext context = new CteTransformerContext();
+            PlanNode transformedCte = SimplePlanRewriter.rewriteWith(new CteConsumerTransformer(planNodeIdAllocator, variableAllocator),
+                    root, context);
+            List<PlanNode> topologicalOrderedList = context.getTopologicalOrdering();
+            if (topologicalOrderedList.isEmpty()) {
+                return root;
+            }
+            SequenceNode sequenceNode = new SequenceNode(root.getSourceLocation(), planNodeIdAllocator.getNextId(), topologicalOrderedList,
+                    transformedCte.getSources().get(0));
+            return root.replaceChildren(Arrays.asList(sequenceNode));
```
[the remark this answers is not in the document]
jaystarshot: Done,thanks

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/CteUtils.java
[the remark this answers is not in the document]
jaystarshot: Added a test based on ttwithtz, thanks

jaystarshot:
It won't because the canonical hashes won't be generated and properly linked with the execution.
I have 3 follow up prs'
1. Add these new nodes in cost frameworks - [link](https://github.com/jaystarshot/presto/pull/2/commits/46ab4f7a6d21eea949a7cabdc0bf1dd0fe53d793) 
2. Support HBO with these new nodes - [link](https://github.com/jaystarshot/presto/pull/3/commits/cfcb76612b0b1e1979ae28bf097f46e4c3500e39)
3. Locally auto materialize the ctes - [link](https://github.com/jaystarshot/presto/pull/1/commits/ddd3e2672252cbc8fa28051429bda2fbb47bbdfe)

mlyublena:
> It won't because the canonical hashes won't be generated and properly linked with the execution. I have 3 follow up prs'
> 
> 1. Add these new nodes in cost frameworks - [link](https://github.com/jaystarshot/presto/pull/2/commits/46ab4f7a6d21eea949a7cabdc0bf1dd0fe53d793)
> 2. Support HBO with these new nodes - [link](https://github.com/jaystarshot/presto/pull/3/commits/cfcb76612b0b1e1979ae28bf097f46e4c3500e39)
> 3. Locally auto materialize the ctes - [link](https://github.com/jaystarshot/presto/pull/1/commits/ddd3e2672252cbc8fa28051429bda2fbb47bbdfe)

makes sense.
what about the opposite: if the CTE is not materialized, would HBO work as before?

jaystarshot:
Yes, the optimizer and entire flow remains unchanged unless the relational planner adds a cteReferenceNode, which only occurs if the specific flag is enabled.

jaystarshot:
Does this PR look good to merge?

jaystarshot:
As a follow up PR, I will also add a materialization strategy to only materialize ctes if references > threshold times and have a join or aggregate and maybe reduce data size

tdcmeehan:
> Does this PR look good to merge? Will need a committer approval

I'll have a look this week, other committers are free to get to it sooner!

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/PhysicalCteOptimizer.java
```diff
@@ -0,0 +1,301 @@
...
+                                        node.getSourceLocation(),
+                                        temporaryTableHandle,
+                                        actualSource.getOutputVariables(),
+                                        variableToColumnMap,
+                                        partitioningMetadata), node.getOutputVariables()));
+            }
+            catch (PrestoException e) {
+                if (e.getErrorCode().equals(NOT_SUPPORTED.toErrorCode())) {
+                    throw new PrestoException(
+                            NOT_SUPPORTED,
```
mlyublena, line 169: should we add this check to is isCteMaterializable?

jaystarshot: I think here its just a catch for the metadata.createTemporaryTable() which cannot be invoked in the parser since we do not have partition metadata and other info there

jaystarshot:
Adding a nit (using partitioningProviderCatalog) instead of hardcoded "hive" string in a use case.

presto-spi/src/main/java/com/facebook/presto/spi/plan/CteReferenceNode.java
```diff
@@ -0,0 +1,105 @@
...
+    @Override
+    public List<VariableReferenceExpression> getOutputVariables()
+    {
+        return source.getOutputVariables();
+    }
+
+    @Override
+    public PlanNode replaceChildren(List<PlanNode> newChildren)
+    {
+        checkArgument(newChildren.size() == 1, "expected newChildren to contain 1 node");
```
tdcmeehan, line 78: Please also validate that the input is not null with `Objects.requireNonNull`.

presto-main/src/main/java/com/facebook/presto/sql/TemporaryTableUtil.java
```diff
@@ -0,0 +1,387 @@
...
+import static java.util.function.Function.identity;
+
+// Planner Util for creating temporary tables
+public class TemporaryTableUtil
+{
+    private TemporaryTableUtil()
+    {
+    }
+
+    public static TableScanNode createTemporaryTableScan(
```
tdcmeehan, line 88: Here and elsewhere: Please validate inputs are not null

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/LogicalCteOptimizer.java
```diff
@@ -0,0 +1,229 @@
...
+        public void addDependency(String currentCte)
+        {
+            graph.addNode(currentCte);
+            Optional<String> parentCte = peekActiveCte();
+            parentCte.ifPresent(s -> graph.putEdge(currentCte, s));
+        }
+
+        public List<PlanNode> getTopologicalOrdering()
+        {
+            List<PlanNode> topSortedCteProducerList = new ArrayList<>();
```
tdcmeehan, line 223: Use immutable list

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/LogicalCteOptimizer.java
```diff
@@ -0,0 +1,229 @@
...
+        {
+            cteProducerMap = new HashMap<>();
+            // The cte graph will never have cycles because sql won't allow it
+            graph = GraphBuilder.directed().build();
+            activeCteStack = new Stack<>();
+        }
+
+        public Map<String, CteProducerNode> getCteProducerMap()
+        {
+            return cteProducerMap;
```
tdcmeehan, line 191: Can this return an immutable copy

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/PhysicalCteOptimizer.java
```diff
@@ -0,0 +1,301 @@
...
+{
+    private final Metadata metadata;
+
+    public PhysicalCteOptimizer(Metadata metadata)
+    {
+        this.metadata = metadata;
+    }
+
+    @Override
+    public PlanOptimizerResult optimize(PlanNode plan, Session session, TypeProvider types, VariableAllocator variableAllocator, PlanNodeIdAllocator idAllocator, WarningCollector warningCollector)
```
tdcmeehan, line 88: Here and elsewhere--please always validate the inputs are not null

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/PhysicalCteOptimizer.java
```diff
@@ -0,0 +1,301 @@
...
+
+        private final VariableAllocator variableAllocator;
+
+        private final Session session;
+
+        private boolean isPlanRewritten;
+
+        public CteProducerRewiter(Session session, PlanNodeIdAllocator idAllocator, VariableAllocator variableAllocator)
+        {
+            this.idAllocator = idAllocator;
```
tdcmeehan, line 116: Ditto

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/PhysicalCteOptimizer.java
```diff
@@ -0,0 +1,301 @@
...
+            this.variableAllocator = variableAllocator;
+            this.session = session;
+        }
+
+        @Override
+        public PlanNode visitCteProducer(CteProducerNode node, RewriteContext<PhysicalCteTransformerContext> context)
+        {
```
tdcmeehan, line 123: Ditto

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/PhysicalCteOptimizer.java
```diff
@@ -0,0 +1,301 @@
...
+            TableScanNode tableScanNode = new TableScanNode(
+                    Optional.empty(),
+                    idAllocator.getNextId(),
+                    tempScan.getTable(),
+                    newOutputVariables,
+                    newColumnAssignmentsMap,
+                    tempScan.getCurrentConstraint(),
+                    tempScan.getEnforcedConstraint());
+
+            /* The temporary table scan might have columns removed by the UnaliasSymbolReferences and other optimizers (its a plan tree after all),
```
tdcmeehan, line 239: In general we prefer `//` comments

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/PhysicalCteOptimizer.java
```diff
@@ -0,0 +1,301 @@
...
+
+        public static class TemporaryTableInfo
+        {
+            private final TableScanNode tableScanNode;
+            private final List<VariableReferenceExpression> originalOutputVariables;
+
+            public TemporaryTableInfo(TableScanNode tableScanNode, List<VariableReferenceExpression> originalOutputVariables)
+            {
+                this.tableScanNode = tableScanNode;
+                this.originalOutputVariables = originalOutputVariables;
```
tdcmeehan, line 287: Please check inputs are not null, and create an immutable copy of the original output variables (no cost if the input is immutable as well)

presto-main/src/main/java/com/facebook/presto/sql/planner/BasePlanFragmenter.java
```diff
@@ -223,6 +201,31 @@ public PlanNode visitTableFinish(TableFinishNode node, RewriteContext<FragmentPr
         return context.defaultRewrite(node, context.get());
     }
 
+    @Override
+    public PlanNode visitSequence(SequenceNode node, RewriteContext<FragmentProperties> context)
+    {
+        // Since this is topologically sorted by the LogicalCtePlanner, need to make sure that execution order follows
+        // Can be optimized further to avoid non dependents from getting blocked
+        int n = node.getCteProducers().size();
```
tdcmeehan, line 209: Please do not use single letter variables.  `cteProducerCount` or something similar

tdcmeehan (changes requested):
Please make sure to consistently validate inputs.  Please squash the `Tests` commit.

tdcmeehan:
Please look into the merge conflicts.

presto-main/src/main/java/com/facebook/presto/sql/TemporaryTableUtil.java
jaystarshot: Thanks added in other cases, not adding here because not really needed at this static function.

presto-spi/src/main/java/com/facebook/presto/spi/plan/CteReferenceNode.java
```diff
@@ -0,0 +1,107 @@
...
+    @Override
+    public List<VariableReferenceExpression> getOutputVariables()
+    {
+        return source.getOutputVariables();
+    }
+
+    @Override
+    public PlanNode replaceChildren(List<PlanNode> newChildren)
+    {
+        requireNonNull(newChildren);
```
tdcmeehan, line 79: Nit: please add the message too, so it's very obvious this is a validation failure and therefore very indicative of something that wasn't considered during development.
```suggestion
        requireNonNull(newChildren, "newChildren is null");
```

Please make sure the message is consistent, we use one convention which is `VARIABLE_NAME is null`.

tdcmeehan (approved):
Just one last nit, and I think good to merge.

tdcmeehan (approved):

PR #20888, opened 2023-09-16
[Native] Turn on support for bucketed sorted writer

kewang1024:
```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeGeneralQueries.java
+++ b/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeGeneralQueries.java
@@ -1121,18 +1121,20 @@ public abstract class AbstractTestNativeGeneralQueries
         // Generate temporary table name.
         String badBucketTableName = generateRandomTableName();
 
-        // TODO: update this test condition after bucket sort write is supported by native worker.
-        try {
-            this.assertQueryFails(session, String.format(
-                    "CREATE TABLE %s WITH (" +
-                            "partitioned_by = ARRAY[ 'orderstatus' ], " +
-                            "bucketed_by=array['orderkey'], " +
-                            "bucket_count=11, " +
-                            "sorted_by=array['orderkey']) " +
-                            "AS SELECT orderkey, orderstatus FROM orders", badBucketTableName), ".*Bucketed sorted table is not supported.*");
-        }
-        finally {
-            dropTableIfExists(badBucketTableName);
+        for (String tableFormat : TABLE_FORMATS) {
+            try {
+                getQueryRunner().execute(session, String.format(
+                        "CREATE TABLE %s WITH (format = '%s', " +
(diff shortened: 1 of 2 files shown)


presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeGeneralQueries.java
```diff
@@ -1121,18 +1121,20 @@ public void testCreateBucketSortedTableAsSelect()
...
+        for (String tableFormat : TABLE_FORMATS) {
+            try {
+                getQueryRunner().execute(session, String.format(
+                        "CREATE TABLE %s WITH (format = '%s', " +
+                                "partitioned_by = ARRAY[ 'orderstatus' ], " +
+                                "bucketed_by=array['custkey'], " +
+                                "bucket_count=1, " +
+                                "sorted_by=array['orderkey']) " +
+                                "AS SELECT custkey, orderkey, orderstatus FROM orders", badBucketTableName, tableFormat));
+                assertQueryOrdered(String.format("SELECT custkey, orderkey, orderstatus FROM %s where orderstatus = '0'", badBucketTableName), "SELECT custkey, orderkey, orderstatus FROM orders where orderstatus = '0'");
```
xiaoxmeng, line 1133: Is it possible to get the list files and use $PATH to read from each sorted bucket file and verify they are sorted? Like the result is the same with and without orderby as we do here. But we do this comparison on each individual file. Then we can test bucket count non-zero case. Thanks!

xiaoxmeng (approved):
@kewang1024 LGTM. Thanks!

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeGeneralQueries.java
kewang1024: The result is not the same without orderby here, because I used "assertQueryOrdered"

xiaoxmeng: I mean use $PATH to read from each single file written by bucket sort write, and the result will be same with and without order by? Thanks!

kewang1024: I know we can do "select $PATH from table_name",  I'm not sure if we can use SQL to read the data from the file itself?

Let me have a followup PR for it once we figure out a way to do it?

mbasmanova: @kewang1024 Yes, we can: "SELECT ... FROM t WHER \"$path\" = '..'"

kewang1024: New knowledge for me, it would be pretty useful, thanks @mbasmanova !

PR #20891, opened 2023-09-17
[native] Cancel function before adding once in PeriodicTaskManager.

amitkdutta:
Sometimes it's required to cancel a function that was registered in PeriodicTaskManager. Exposing cancel function through public method.

### DIFF
--- a/presto-native-execution/presto_cpp/main/PeriodicTaskManager.h
+++ b/presto-native-execution/presto_cpp/main/PeriodicTaskManager.h
@@ -68,10 +68,12 @@ class PeriodicTaskManager {
         func, std::chrono::microseconds{periodMicros}, taskName);
   }
 
-  /// Add a task to run once.
+  /// Add a task to run once. Before adding, cancels the any task that has same
+  /// name.
   template <typename TFunc>
   void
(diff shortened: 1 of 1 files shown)


presto-native-execution/presto_cpp/main/PeriodicTaskManager.h
```diff
@@ -68,10 +68,12 @@ class PeriodicTaskManager {
         func, std::chrono::microseconds{periodMicros}, taskName);
   }
 
-  /// Add a task to run once.
+  /// Add a task to run once. Before adding, cancels the any task that has same
```
xiaoxmeng, line 71:
```
/// Add a task to run once.
void addTaskOnce(TFunc&& func, size_t periodMicros, const std::string& taskName) {
  // NOTE: before adding the function, we need to cancel any function with the same name. It is because folly::FunctionScheduler only removes a run-once function after its execution internally.
  ...
```
The note is specific to folly::FunctionScheduler implementation and other executor might not has such limitation.

xiaoxmeng (approved):
@amitkdutta LGTM. Consider to add PeriodicTaskManagerTest.cpp next. Thanks!

PR #20894, opened 2023-09-18
Make config in FeaturesConfig be the same with that in FileFragmentRe…

JiamingMai:
## Description
There is a bug in fragment cache. We expect to enable fragment cache by setting `fragment-result-cache.enabled=true` in `config.properties`. But if we only set that in `config.properties`, it doesn't work. We still need to set the session property `fragment-result-cache.enabled=true` when using the `presto-cli`, since the default session value of `fragment-result-cache.enabled` is set by `experimental.fragment-result-caching-enabled`. That means we have to set both `fragment-result-cache.enabled` and `experimental.fragment-result-caching-enabled` to true if we want to enable fragment cache. This is intuitive and easily misunderstood.  Actually, one configuration `fragment-result-cache.enabled` is enough.

## Impact
We don't need to set both `fragment-result-cache.enabled` and `experimental.fragment-result-caching-enabled` to `true` if we want to enable fragment cache. Now only `fragment-result-cache.enabled=true` is required. 

## Test Plan
Remove the configuration `experimental.fragment-result-caching-enabled` and only set `fragment-result-cache.enabled=true` to test if fragment cache is enabled correctly. I tested it and it turns out that it works.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java
@@ -1564,7 +1564,8 @@ public class FeaturesConfig
         return fragmentResultCachingEnabled;
     }
 
-    @Config("experimental.fragment-result-caching-enabled")
+    @Config("fragment-result-cache.enabled")
+    @LegacyConfig("experimental.fragment-result-caching-enabled")
     @ConfigDescription("Enable fragment result caching and read/write leaf fragment result pages from/to cache when applicable")
     public FeaturesConfig setFragmentResultCachingEnabled(boolean fragmentResultCachingEnabled)
     {
--- a/presto-main/src/test/java/com/facebook/presto/sql/analyzer/TestFeaturesConfig.java
+++ b/presto-main/src/test/java/com/facebook/presto/sql/analyzer/TestFeaturesConfig.java
@@ -259,7 +259,7 @@ public class TestFeaturesConfig
                 .put("dynamic-filtering-max-per-driver-row-count", "256")
                 .put("dynamic-filtering-max-per-driver-size", "64kB")
                 .put("dynamic-filtering-range-row-limit-per-driver", "1000")
-                .put("experimental.fragment-result-caching-enabled", "true")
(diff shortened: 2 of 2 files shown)


presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java
```diff
@@ -1564,7 +1564,7 @@ public boolean isFragmentResultCachingEnabled()
         return fragmentResultCachingEnabled;
     }
 
-    @Config("experimental.fragment-result-caching-enabled")
+    @Config("fragment-result-cache.enabled")
```
tdcmeehan, line 1567: Can you leave the old config on by adding a `LegacyConfig`?  That way people don't need to change their configs when they upgrade.

tdcmeehan (changes requested):

presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java
JiamingMai: I just added a `LegacyConfig` for the old config.

beinan: Is it possible to keep the `experimental`, I don't think FRC is a production ready feature

beinan: But I'm ok, if we have to remove it to make it correct

beinan:
Synced with @JiamingMai offline, I'm good with the change now.  @JiamingMai could you squash the 3 commit into one? thanks for the contribution!

beinan:
Commit message is too long.  Is it possible to make it shorter and capitalized first letter? thanks!

JiamingMai:
> Commit message is too long. Is it possible to make it shorter and capitalized first letter? thanks!

Thanks beinan, I updated the commit message.

beinan (approved):
lgtm

tdcmeehan (approved):

PR #20905, opened 2023-09-19
[native] Use presto_native as prefix for spill path.

amitkdutta:
Instead of putting date as prefix of spill path, using presto-native as its easier to manager for remote storage.


xiaoxmeng (approved):
@amitkdutta LGTM. Thanks!

ajaygeorge (approved):

PR #20907, opened 2023-09-19
[native]Move proxygen session pool destruction to run at the background

xiaoxmeng:
In arbitration & spilling test on Prestissimo, we found driver thread is stuck
in http client destruction (from PrestoExchangeSource destruction) which 
tries to reset proxygen session pool in event base thread inline and wait for
its to completion. This might cause deadlock if there is task memory reclaim
operation waits for task to pause, and one of the task driver thread is inline
waiting for the session pool to be reset while the event thread is busy waiting
for memory arbitration. This forms a circular dependency. This PR fixes the
problem by scheduling a function to reset proxygen session pool in event
base thread. When http client destroys, there is no pending reference on
proxygen session pool so we don't have to inline execute and wait.
The shadowing test with arbitration runs a couple of hours without deadlock. 

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/http/HttpClient.cpp
+++ b/presto-native-execution/presto_cpp/main/http/HttpClient.cpp
@@ -46,9 +46,15 @@ HttpClient::HttpClient(
 
 HttpClient::~HttpClient() {
   if (sessionPool_) {
-    // make sure to destroy SessionPool on the EventBase thread
-    eventBase_->runImmediatelyOrRunInEventBaseThreadAndWait(
-        [this] { sessionPool_.reset(); });
+    // Make sure to destroy SessionPool on the EventBase thread.
+    //
+    // NOTE: we can't run the destruction callback inline in EventBase thread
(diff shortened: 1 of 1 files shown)


presto-native-execution/presto_cpp/main/http/HttpClient.cpp
```diff
@@ -47,8 +47,8 @@ HttpClient::HttpClient(
 HttpClient::~HttpClient() {
   if (sessionPool_) {
     // make sure to destroy SessionPool on the EventBase thread
-    eventBase_->runImmediatelyOrRunInEventBaseThreadAndWait(
-        [this] { sessionPool_.reset(); });
+    eventBase_->runInEventBaseThread(
```
mbasmanova, line 50: It would be nice to update the comment to explain why not run this inline.

mbasmanova (approved):
Thanks.

PR description has a typo:

` The shadowing test with arbitration runs a couple of hours with deadlock.`

Would you update commit message to add more details?

PR #20914, opened 2023-09-19
[Native] Add connector config

kewang1024:
Hive connector properties is a key-value structure read from config
file, however, it won't have the configs that are not present in config 
file.

This PR adds the infrastructural support for hive connector config to
include all.

### DIFF
--- a/presto-native-execution/presto_cpp/main/PrestoServer.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoServer.cpp
@@ -684,27 +684,25 @@ std::vector<std::string> PrestoServer::registerConnectors(
       auto catalogName =
           fileName.substr(0, fileName.size() - kPropertiesExtension.size());
 
-      auto connectorConf = util::readConfig(entry.path());
+      ConnectorConfig connectorConfig;
+      connectorConfig.initialize(entry.path());
+
       PRESTO_STARTUP_LOG(INFO)
           << "Registered properties from " << entry.path() << ":\n"
-          << stringifyConnectorConfig(connectorConf);
-
-      std::shared_ptr<const velox::Config> properties =
-          std::make_shared<const velox::core::MemConfig>(
-              std::move(connectorConf));
-
-      auto connectorName = util::requiredProperty(*properties, kConnectorName);
+          << stringifyConnectorConfig(connectorConfig.values());
 
       catalogNames.emplace_back(catalogName);
 
-      PRESTO_STARTUP_LOG(INFO) << "Registering catalog " << catalogName
-                               << " using connector " << connectorName;
+      PRESTO_STARTUP_LOG(INFO)
+          << "Registering catalog " << catalogName << " using connector "
+          << connectorConfig.connectorName();
 
       std::shared_ptr<velox::connector::Connector> connector =
-          facebook::velox::connector::getConnectorFactory(connectorName)
+          facebook::velox::connector::getConnectorFactory(
+              connectorConfig.connectorName())
               ->newConnector(
                   catalogName,
-                  std::move(properties),
+                  std::move(connectorConfig.config()),
                   connectorIoExecutor_.get());
       velox::connector::registerConnector(connector);
     }
--- a/presto-native-execution/presto_cpp/main/common/Configs.cpp
+++ b/presto-native-execution/presto_cpp/main/common/Configs.cpp
@@ -567,6 +567,22 @@ int32_t SystemConfig::oldTaskCleanUpMs() const {
   return optionalProperty<int32_t>(kOldTaskCleanUpMs).value();
 }
 
+ConnectorConfig::ConnectorConfig() {
+  registeredProps_ =
+      std::unordered_map<std::string, folly::Optional<std::string>>{
+          NONE_PROP(kConnectorName),
+          BOOL_PROP(kHiveImmutablePartitions, false),
+      };
+}
+
+std::string ConnectorConfig::connectorName() const {
+  return requiredProperty(kConnectorName);
+}
+
+bool ConnectorConfig::hiveImmutablePartitions() const {
+  return optionalProperty<bool>(kHiveImmutablePartitions).value();
+}
+
 NodeConfig::NodeConfig() {
   registeredProps_ =
       std::unordered_map<std::string, folly::Optional<std::string>>{
--- a/presto-native-execution/presto_cpp/main/common/Configs.cpp
+++ b/presto-native-execution/presto_cpp/main/common/Configs.cpp
@@ -142,21 +142,32 @@ ConfigBase::ConfigBase()
     : config_(std::make_unique<velox::core::MemConfig>()) {}
 
 void ConfigBase::initialize(const std::string& filePath) {
-  // See if we want to create a mutable config.
-  auto values = util::readConfig(fs::path(filePath));
+  auto propMapFromFile = util::readConfig(fs::path(filePath));
   filePath_ = filePath;
-  checkRegisteredProperties(values);
+  checkRegisteredProperties(propMapFromFile);
 
+  // See if we want to create a mutable config.
   bool mutableConfig{false};
-  auto it = values.find(std::string(kMutableConfig));
-  if (it != values.end()) {
+  auto it = propMapFromFile.find(std::string(kMutableConfig));
+  if (it != propMapFromFile.end()) {
     mutableConfig = folly::to<bool>(it->second);
   }
 
+  std::unordered_map<std::string, std::string> configMap;
+  for (const auto& prop : registeredProps_) {
+    if (prop.second.hasValue()) {
+      configMap[prop.first] = prop.second.value();
+    }
+  }
+  // Config from file would overwrite the value from registeredProps_.
+  for (const auto& prop : propMapFromFile) {
+    configMap[prop.first] = prop.second;
+  }
+
   if (mutableConfig) {
-    config_ = std::make_unique<velox::core::MemConfigMutable>(values);
+    config_ = std::make_unique<velox::core::MemConfigMutable>(configMap);
   } else {
-    config_ = std::make_unique<velox::core::MemConfig>(values);
+    config_ = std::make_unique<velox::core::MemConfig>(configMap);
   };
 }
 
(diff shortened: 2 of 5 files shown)


spershin:
@kewang1024 
registeredProps_ is used for two purposes:
* To store props we support (and report we have an unsupported prop or even fail to start later, like Presto does)
* To store props default values.

spershin:
@kewang1024 We need to add some description.

presto-native-execution/presto_cpp/main/common/Configs.h
```diff
@@ -461,6 +465,23 @@ class SystemConfig : public ConfigBase {
   int32_t oldTaskCleanUpMs() const;
 };
 
+/// Provides access to connector properties defined in
+/// catalog/$connector_name.properties file.
+class ConnectorConfig : public ConfigBase {
+ public:
+  static constexpr std::string_view kConnectorName{"connector.name"};
+  static constexpr std::string_view kHiveImmutablePartitions{
```
spershin, line 473: We have "cache.enabled=true" in the config file, should we add it now too?

presto-native-execution/presto_cpp/main/common/Configs.cpp
```diff
@@ -558,6 +571,27 @@ int32_t SystemConfig::oldTaskCleanUpMs() const {
   return optionalProperty<int32_t>(kOldTaskCleanUpMs).value();
 }
 
+ConnectorConfig::ConnectorConfig() {
+  registeredProps_ =
+      std::unordered_map<std::string, folly::Optional<std::string>>{
+          BOOL_PROP(kHiveImmutablePartitions, "false"),
```
spershin, line 577: kConnectorName is not registered?

presto-native-execution/presto_cpp/main/common/Configs.h
spershin, line 473: Btw, let's maybe add
  static constexpr const char* kOrcUseColumnNames = "hive.orc.use-column-names";
too?

presto-native-execution/presto_cpp/main/common/Configs.cpp
```diff
@@ -256,6 +271,9 @@ SystemConfig::SystemConfig() {
           STR_PROP(kUseMmapAllocator, "true"),
           STR_PROP(kMemoryArbitratorKind, ""),
           NUM_PROP(kQueryMemoryGb, 38),
+          NUM_PROP(kMemoryPoolInitCapacity, 128 << 20),
+          NUM_PROP(kMemoryPoolTransferCapacity, 32 << 20),
+          BOOL_PROP(kEnableSystemMemoryPoolUsageTracking, true),
```
spershin, line 276: Nice!

presto-native-execution/presto_cpp/main/common/Configs.cpp
```diff
@@ -142,21 +142,36 @@ ConfigBase::ConfigBase()
     : config_(std::make_unique<velox::core::MemConfig>()) {}
 
 void ConfigBase::initialize(const std::string& filePath) {
-  // See if we want to create a mutable config.
-  auto values = util::readConfig(fs::path(filePath));
+  std::unordered_map<std::string, std::string> configFromFile =
```
spershin, line 145: I think the guidelines of Velox is to use 'auto'.
Also, as we operate with config classes and this is a map, maybe call it 'propMapFromFile'?

presto-native-execution/presto_cpp/main/common/Configs.cpp
```diff
@@ -142,21 +142,36 @@ ConfigBase::ConfigBase()
...
+  // See if we want to create a mutable config.
   bool mutableConfig{false};
-  auto it = values.find(std::string(kMutableConfig));
-  if (it != values.end()) {
+  auto it = configFromFile.find(std::string(kMutableConfig));
+  if (it != configFromFile.end()) {
     mutableConfig = folly::to<bool>(it->second);
   }
 
+  std::unordered_map<std::string, std::string> config;
```
spershin, line 157: Same here - maybe not call map a 'config'?

presto-native-execution/presto_cpp/main/common/Configs.cpp
```diff
@@ -142,21 +142,36 @@ ConfigBase::ConfigBase()
...
+  for (auto property = registeredProps_.begin();
+       property != registeredProps_.end();
+       property++) {
+    if (property->second.hasValue()) {
+      config.insert({property->first, property->second.value()});
+    }
+  }
```
spershin, line 164: Use
```
for(const auto& prop : registeredProps_)
```

presto-native-execution/presto_cpp/main/common/Configs.cpp
```diff
@@ -142,21 +142,36 @@ ConfigBase::ConfigBase()
...
+  // Config from file would overwrite the value from registeredProps_.
+  for (auto property = configFromFile.begin(); property != configFromFile.end();
+       property++) {
+    config.insert({property->first, property->second});
```
spershin, line 168: "insert" inserts element(s) into the container, if the container doesn't already contain an element with an equivalent key.

Which means it won't overwrite the existing ones. Have you tested it?

spershin (commented):
Thanks!
Looks good, some comments.

presto-native-execution/presto_cpp/main/common/Configs.cpp
kewang1024: Nice catch!

presto-native-execution/presto_cpp/main/common/Configs.cpp
kewang1024: I just noticed that ConfigTest would fail with current version, after changing, it would pass.
Thanks for catching the issue!

presto-native-execution/presto_cpp/main/common/Configs.h
kewang1024: I plan to add the rest of the config all together later because we're missing a lot, do you think this one is especially important?

kewang1024: I don't think "cache.enabled=true" is used anymore, I think they're replaced by https://github.com/prestodb/presto/blob/5b604194c8908aeb0244771536a786351e81a991/presto-native-execution/presto_cpp/main/common/Configs.h#L196-L206

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -684,27 +684,25 @@ std::vector<std::string> PrestoServer::registerConnectors(
       auto catalogName =
           fileName.substr(0, fileName.size() - kPropertiesExtension.size());
 
-      auto connectorConf = util::readConfig(entry.path());
+      auto connectoConfig = ConnectorConfig::instance();
```
majetideepak, line 687: typo: `connectorConfig`

presto-native-execution/presto_cpp/main/common/Configs.h
spershin: Not really, just the one I know of. :)

spershin: Then maybe we should remove it from the config file.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -684,27 +684,25 @@ std::vector<std::string> PrestoServer::registerConnectors(
       auto catalogName =
           fileName.substr(0, fileName.size() - kPropertiesExtension.size());
 
-      auto connectorConf = util::readConfig(entry.path());
+      auto connectorConfig = ConnectorConfig::instance();
+      connectoConfig->initialize(entry.path());
```
spershin, line 688: We have multiple connectors and each of them has a config.
It seems we don't need a ConnectorConfig's instance to stay, since we just take prop maps from them and give it to the connectors.
So, no need to have ConnectorConfig::instance() method - just create one ConnectorConfig object (not a pointer) on the stack inside the loop.

presto-native-execution/presto_cpp/main/common/Configs.h
kewang1024: Yes, already did

presto-native-execution/presto_cpp/main/common/Configs.cpp
```diff
@@ -558,6 +567,28 @@ int32_t SystemConfig::oldTaskCleanUpMs() const {
...
+          NONE_PROP(kConnectorName),
+          BOOL_PROP(kHiveImmutablePartitions, false),
+      };
+}
+
+ConnectorConfig* ConnectorConfig::instance() {
+  static std::unique_ptr<ConnectorConfig> instance =
+      std::make_unique<ConnectorConfig>();
+  return instance.get();
+}
```
spershin, line 582: Let's remove this member function.

presto-native-execution/presto_cpp/main/common/Configs.h
```diff
@@ -461,6 +465,23 @@ class SystemConfig : public ConfigBase {
...
+      "hive.immutable-partitions"};
+
+  ConnectorConfig();
+
+  static ConnectorConfig* instance();
```
spershin, line 478: Let's remove this member function.

spershin (approved):
@kewang1024 
Thanks for this change!
Please, remove the unused member function before merging.

presto-native-execution/presto_cpp/main/common/Configs.h
```diff
@@ -461,6 +465,21 @@ class SystemConfig : public ConfigBase {
   int32_t oldTaskCleanUpMs() const;
 };
 
+/// Provides access to connector properties defined in
+/// catalog/$connector_name.properties file.
+class ConnectorConfig : public ConfigBase {
+ public:
+  static constexpr std::string_view kConnectorName{"connector.name"};
+  static constexpr std::string_view kHiveImmutablePartitions{
```
majetideepak, line 473: We today have the TPC-H connector and potentially other connectors in the future. Why are we adding Hive Connector specific fields here?

majetideepak (changes requested):

presto-native-execution/presto_cpp/main/common/Configs.cpp
```diff
@@ -558,6 +567,22 @@ int32_t SystemConfig::oldTaskCleanUpMs() const {
   return optionalProperty<int32_t>(kOldTaskCleanUpMs).value();
 }
 
+ConnectorConfig::ConnectorConfig() {
+  registeredProps_ =
+      std::unordered_map<std::string, folly::Optional<std::string>>{
+          NONE_PROP(kConnectorName),
+          BOOL_PROP(kHiveImmutablePartitions, false),
```
majetideepak, line 574: Same question here. ConnectorConfig seems to be generic but is tied to a Hive property.

presto-native-execution/presto_cpp/main/common/Configs.h
kewang1024: That's a good point, then I think the default value can be set in velox HiveConfig instead.

Will have this PR to solve the issue: https://github.com/facebookincubator/velox/pull/6658

presto-native-execution/presto_cpp/main/common/Configs.cpp
spershin: @majetideepak 
We might need to revisit this at some point. Worth looking how Java deals with different props for different connector configs and copy the behavior.
cc @kewang1024

PR #18095, opened 2022-07-25
Enable Presto Verifier to rewrite nondeterministic columns

gggrace14:
```
== NO RELEASE NOTE ==
```

PR #20923, opened 2023-09-20
Add FunctionCall expression rewriter to presto.verifier

gggrace14:
This change creates a FunctionCallRewriter that overrides the base class AstVisitor,
substituting the FunctionCall node and reconstructing the AST of the source query.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-parser/src/main/java/com/facebook/presto/sql/tree/CreateMaterializedView.java
+++ b/presto-parser/src/main/java/com/facebook/presto/sql/tree/CreateMaterializedView.java
@@ -31,6 +31,11 @@ public class CreateMaterializedView
     private final List<Property> properties;
     private final Optional<String> comment;
 
+    public CreateMaterializedView(QualifiedName name, Query query, boolean notExists, List<Property> properties, Optional<String> comment)
+    {
+        this(Optional.empty(), name, query, notExists, properties, comment);
+    }
+
     public CreateMaterializedView(Optional<NodeLocation> location, QualifiedName name, Query query, boolean notExists, List<Property> properties, Optional<String> comment)
     {
         super(location);
--- a/presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/DefaultTreeRewriter.java
+++ b/presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/DefaultTreeRewriter.java
@@ -1,0 +1,730 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.verifier.rewrite;
+
+import com.facebook.presto.sql.tree.AddColumn;
+import com.facebook.presto.sql.tree.AliasedRelation;
+import com.facebook.presto.sql.tree.Analyze;
+import com.facebook.presto.sql.tree.AstVisitor;
+import com.facebook.presto.sql.tree.Call;
+import com.facebook.presto.sql.tree.CallArgument;
+import com.facebook.presto.sql.tree.ColumnDefinition;
+import com.facebook.presto.sql.tree.CreateMaterializedView;
+import com.facebook.presto.sql.tree.CreateSchema;
+import com.facebook.presto.sql.tree.CreateTable;
+import com.facebook.presto.sql.tree.CreateTableAsSelect;
+import com.facebook.presto.sql.tree.CreateView;
+import com.facebook.presto.sql.tree.Cube;
+import com.facebook.presto.sql.tree.Deallocate;
+import com.facebook.presto.sql.tree.Delete;
+import com.facebook.presto.sql.tree.Except;
+import com.facebook.presto.sql.tree.Execute;
+import com.facebook.presto.sql.tree.Explain;
+import com.facebook.presto.sql.tree.Expression;
+import com.facebook.presto.sql.tree.FrameBound;
+import com.facebook.presto.sql.tree.GroupBy;
+import com.facebook.presto.sql.tree.GroupingElement;
+import com.facebook.presto.sql.tree.GroupingSets;
+import com.facebook.presto.sql.tree.Identifier;
+import com.facebook.presto.sql.tree.Insert;
+import com.facebook.presto.sql.tree.Intersect;
+import com.facebook.presto.sql.tree.Join;
+import com.facebook.presto.sql.tree.JoinCriteria;
+import com.facebook.presto.sql.tree.JoinOn;
+import com.facebook.presto.sql.tree.Lateral;
+import com.facebook.presto.sql.tree.Node;
+import com.facebook.presto.sql.tree.OrderBy;
+import com.facebook.presto.sql.tree.Prepare;
+import com.facebook.presto.sql.tree.Property;
+import com.facebook.presto.sql.tree.Query;
+import com.facebook.presto.sql.tree.QueryBody;
+import com.facebook.presto.sql.tree.QuerySpecification;
+import com.facebook.presto.sql.tree.RefreshMaterializedView;
+import com.facebook.presto.sql.tree.Relation;
+import com.facebook.presto.sql.tree.Return;
+import com.facebook.presto.sql.tree.Rollup;
+import com.facebook.presto.sql.tree.Row;
+import com.facebook.presto.sql.tree.SampledRelation;
+import com.facebook.presto.sql.tree.Select;
+import com.facebook.presto.sql.tree.SelectItem;
+import com.facebook.presto.sql.tree.ShowStats;
+import com.facebook.presto.sql.tree.SimpleGroupBy;
+import com.facebook.presto.sql.tree.SingleColumn;
+import com.facebook.presto.sql.tree.SortItem;
+import com.facebook.presto.sql.tree.Statement;
+import com.facebook.presto.sql.tree.Table;
+import com.facebook.presto.sql.tree.TableElement;
+import com.facebook.presto.sql.tree.TableSubquery;
+import com.facebook.presto.sql.tree.Union;
+import com.facebook.presto.sql.tree.Unnest;
+import com.facebook.presto.sql.tree.Values;
+import com.facebook.presto.sql.tree.Window;
+import com.facebook.presto.sql.tree.WindowFrame;
+import com.facebook.presto.sql.tree.With;
+import com.facebook.presto.sql.tree.WithQuery;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Iterables;
+
+import java.util.Iterator;
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * A default implementation of {@link AstVisitor} that reconstructs a node if any of its children is reconstructed.
+ * Expression node reconstruction is not supported and left to the users. At the moment it is only used for presto.verifier package.
+ * Generalize it with caution.
+ */
+public class DefaultTreeRewriter<C>
+        extends AstVisitor<Node, C>
+{
+    @Override
+    protected Node visitNode(Node node, C context)
+    {
+        return node;
+    }
+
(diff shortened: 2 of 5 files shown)


presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/DefaultTreeRewriter.java
```diff
@@ -0,0 +1,729 @@
...
+
+import java.util.Iterator;
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * A default implementation of {@link AstVisitor} that reconstructs a node if any of its children is reconstructed.
+ * Expression node reconstruction is not supported and left to the users.
+ */
+public class DefaultTreeRewriter<C>
```
agrawaldevesh, line 87: If this class is copied and forked, can you link to the original and mention the motivation and change ?

presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/FunctionCallRewriter.java
```diff
@@ -0,0 +1,136 @@
...
+        }
+
+        @Override
+        public int hashCode()
+        {
+            return Objects.hash(name, originalArgumentIndices);
+        }
+    }
+
+    private static final Pattern FUNCTION_CALL_SUBSTITUTION_PATTERN = Pattern.compile("/(\\w+)\\(([\\w|,]+)\\)/(\\w+)\\(([\\w|,]+)\\)/");
```
agrawaldevesh, line 69: Can you add a comment on what this regex matches ?

presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/QueryRewriter.java
```diff
@@ -159,6 +172,9 @@ public QueryObjectBundle rewriteQuery(@Language("SQL") String query, ClusterType
             ResultSetMetaData metadata = getResultMetadata((Query) statement);
             List<Identifier> columnAliases = generateStorageColumnAliases(metadata);
             Query rewrite = rewriteNonStorableColumns((Query) statement, metadata);
+            if (functionCallRewriter.isPresent()) {
```
agrawaldevesh, line 175: Why is the function call rewrite only done for select queries ?

presto-verifier/src/test/java/com/facebook/presto/verifier/rewrite/TestQueryRewriter.java
```diff
@@ -294,6 +298,160 @@ public void testRewriteNonStorableStructuredTypes()
...
+
+        assertTrue(FunctionCallRewriter.constructFunctionCallSubstituteMap("").isEmpty());
+        assertTrue(FunctionCallRewriter.constructFunctionCallSubstituteMap("/func1(a,b)//").isEmpty());
+    }
+
+    @Test
+    public void testRewriteFunctionCalls()
+    {
+        QueryRewriter queryRewriter = getQueryRewriter(Optional.of("/approx_distinct(x)/count(x)/," +
+                "/approx_percentile(x,_)/avg(x)/," +
```
agrawaldevesh, line 326: I am wary of dropping arguments like this: we may be changing what gets read transitively and maybe hiding any bugs. 

Particularly as we test newer file formats like alpha with velox. 

Is there a way to continue projecting the existing arguments somehow ? Like avg plus some function of the second argument ?

presto-verifier/src/test/java/com/facebook/presto/verifier/rewrite/TestQueryRewriter.java
```diff
@@ -294,6 +298,160 @@ public void testRewriteNonStorableStructuredTypes()
...
+                "/arbitrary(x)/min(x)/," +
+                "/map_agg(x,)/max(x)/," +
+                "/first_value(x)/min(x)/," +
+                "/map_union(x)/max(x)/," +
+                "/array_agg(x)/max(x)/"));
+
+        // Test rewriting window functions
+        assertCreateTableAs(
```
agrawaldevesh, line 334: @mbasmanova had made a comment to test some failure scenarios like where the substitute is invalid. Where are those to see that the rewrite is to blame ?

presto-verifier/src/test/java/com/facebook/presto/verifier/rewrite/TestQueryRewriter.java
spershin: There will be some limitations on 'function replace' approach, of course.
In this particular case approx_percentile() most of the time has the rest of the arguments as constants.
The case with weights is exception.
Having a generic replacement as avg(x + func1(y)) can have a problem of type incompatibility.
We could try and se if this will work all the time:
"/approx_percentile(x,y,_)/avg(x + y)/,"
"/approx_percentile(x)/avg(x)/,"

presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/QueryRewriter.java
spershin: I'm curious is where does it say it is only for "select" queries?

presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/QueryRewriter.java
```diff
@@ -95,19 +95,32 @@ public class QueryRewriter
...
+            Map<ClusterType, QualifiedName> tablePrefixes,
+            Map<ClusterType, List<Property>> tableProperties,
+            Optional<String> nonDeterministicFunctionSubstitutes)
     {
         this.sqlParser = requireNonNull(sqlParser, "sqlParser is null");
         this.typeManager = requireNonNull(typeManager, "typeManager is null");
         this.prestoAction = requireNonNull(prestoAction, "prestoAction is null");
         this.prefixes = ImmutableMap.copyOf(tablePrefixes);
         this.tableProperties = ImmutableMap.copyOf(tableProperties);
+        this.functionCallRewriter = requireNonNull(nonDeterministicFunctionSubstitutes, "nonDeterministicFunctionSubstitutes is null").map(FunctionCallRewriter::getInstance);
```
spershin, line 123: I don't understand this:
1. nonDeterministicFunctionSubstitutes is Optional<String>, meaning it can be none.
2. We use requireNonNull, meaning is cannot be 'none'?

presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/FunctionCallRewriter.java
spershin: +1
And one or two examples. It takes me a significant effort every time to understand any kind of regex - I don't work with them too often.

presto-parser/src/main/java/com/facebook/presto/sql/tree/CreateMaterializedView.java
```diff
@@ -31,6 +31,16 @@ public class CreateMaterializedView
...
+    public CreateMaterializedView(QualifiedName name, Query query, boolean notExists, List<Property> properties, Optional<String> comment)
+    {
+        super(Optional.empty());
+        this.name = requireNonNull(name, "name is null");
+        this.query = requireNonNull(query, "query is null");
+        this.notExists = notExists;
+        this.properties = ImmutableList.copyOf(requireNonNull(properties, "properties is null"));
+        this.comment = requireNonNull(comment, "comment is null");
+    }
+
```
spershin, line 43: Why is this needed in this PR?

presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/FunctionCallRewriter.java
```diff
@@ -0,0 +1,136 @@
...
+            return map.build();
+        }
+
+        Matcher matcher = FUNCTION_CALL_SUBSTITUTION_PATTERN.matcher(functionCallSubstitutes);
+
+        while (matcher.find()) {
+            String originalName = matcher.group(1);
+            List<String> originalArgumentList = ImmutableList.copyOf(matcher.group(2).split(","));
+            String substituteName = matcher.group(3);
+            List<String> substituteArgumentList = ImmutableList.copyOf(matcher.group(4).split(","));
```
spershin, line 100: This ".group(N)", is N a 1-based index?

presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/QueryRewriter.java
agrawaldevesh: The above two if clauses are for insert and ctas -- so this one is for select queries. We should be rewriting the above two as well I think. Esp for batch workload which has a lot of inserts.

presto-parser/src/main/java/com/facebook/presto/sql/tree/CreateMaterializedView.java
gggrace14: To provide a constructor without the arg Optional<NodeLocation> location. Almost all subclasses of Node has similar constructor overloading, ex., CreateTableAsSelect, and it is missing for CreateMaterializedView. This is needed by an Ast rewriter DefaultTreeRewriter, since the old NodeLocation is not valid anymore after reconstructing.

presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/FunctionCallRewriter.java
gggrace14: Split the pattern definition, and hope it makes it a bit easier to read.

presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/FunctionCallRewriter.java
gggrace14: .group(1) refers to the group wrapped by the first paranthesis. .group(0) always refers to the entire match. Ex., for a(bc), group(0) returns abc, and group(1) return bc, if matched.

(Also got my memory refreshed by reading the java doc another time)

presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/QueryRewriter.java
gggrace14: For Java,
1) Optional<String> nonDeterministicFunctionSubstitutes = Optional.empty();
2) Optional<String> nonDeterministicFunctionSubstitutes = null;

are different. 2) cannot pass requireNonNull() while 1) can.

Separately, this line calls Optional.map(), and if FunctionCallRewriter.getInstance(nonDeterministicFunctionSubstitutes.get()) returns null, Optional.map() returns Optional.empty().

presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/QueryRewriter.java
gggrace14: The if-flow here now is
```
if (statement instanceof CreateTableAsSelect) {
}
if (statement instanceof Insert) {
}
if (statement instanceof Query) {
    rewrite = (Query) functionCallRewriter.get().process(rewrite);
}
if (statement instanceof CreateView) {
}
```

I will add `functionCallRewriter.get().process()` to the rest in my 2nd PR, https://github.com/prestodb/presto/pull/18095. This PR adds it to the Query branch, only for the sake of unit tests that utilizes existing setup.

This change is split into 2 PRs,
This https://github.com/prestodb/presto/pull/20923 to add the FunctionCallRewriter class.
https://github.com/prestodb/presto/pull/18095 to invoke the FunctionCallRewriter class in the QueryRewriter class  as well as configs.

presto-verifier/src/test/java/com/facebook/presto/verifier/rewrite/TestQueryRewriter.java
gggrace14: For invalid substitution, I include the unit test testConstructMakeFunctionCallSubstituteMap() in this PR to validate the illegal substitution specification string, ex., /func1(a,b)//.

For validating illegal function signature or type signature, like /map_agg(x,)/max(x)/, I will handle this in https://github.com/prestodb/presto/pull/18095.
(The best I figured we can do in Verifier without access to HiveMetadata is to actually run the query after rewritten, fail and provide meaningful error info.)

spershin (approved):
This looks good to me, but I'm not a committer, so someone else would have to approve this change as well.
cc @mbasmanova 

If there are minor things, we can add them in the next PR or even after, when we start running this for Presto Native and learn some lesson on practice.

agrawaldevesh (approved):

mbasmanova (approved):

Issue #20886, opened 2023-09-15
Allow clients to receive query results in PrestoPage format


mbasmanova:
Presto query results today are returned in JSON format: https://prestodb.io/docs/current/develop/client-protocol.html

This is slow and difficult to parse for some clients (e.g. Aggregation Fuzzer in Velox: https://github.com/facebookincubator/velox/issues/6595).

It would be helpful to introduce a new HTTP Header that a client can specify to receive results in PrestoPage format: https://prestodb.io/docs/current/develop/serialized-page.html

The results can still be JSON, but the data would be a string that contains base64-encoded PrestoPage(s).

CC: @aditi-pandit @tdcmeehan @majetideepak

tdcmeehan:
If the primary motivation is better performance, I believe @sachdevs attempted to do this a few years ago and didn't find as significant performance improvements as we expected in the client's ability to fetch lots of data.  I think this is because typically the bottleneck is network and we go through a single coordinator process, fetching page one at a time. I would recommend an approach that allows us to separate metadata from data altogether, and would allow the client to fetch data from workers with multiple threads (e.g. the [v2 protocol](https://github.com/prestodb/presto/wiki/presto-v2-client-protocol)).

If the primary motivation is ease of parsing in a C++ library, have you considered the following:

1) I believe Meta has a C++ client which already parses the JSON data without issue (and with sufficient performance).  Could this be open sourced, or could a modified version be put in open source?
2) Could the output be written to a local file?  In this mode, we don't need any non-backward compatible changes in the client API, you would simply add a session param to save the data locally in the coordinator, or try to persist to a local file system (as I believe Hive can be configured to do), and not attempt to parse the JSON data blob.

mbasmanova:
Relevant change: "Make query result HTTP response compression configurable" - https://github.com/prestodb/presto/pull/15393

mbasmanova:
@tdcmeehan Tim, thank you for your comments.

There are two motivations: (1) performance in fuzzer / test, i.e. when running single-node Presto cluster on localhost; (2) ease of parsing in the client (Velox).

JSON serializing and parsing is quite expensive and will slow down tests considerably, hence, there is a desire to avoid JSON serde altogether. See #15393 for a somewhat related perf optimization for localhost.

Therefore, I still think it would be helpful to allow the client to request results in PrestoPage format and have the server produce these without serializing into JSON. What do you think?

mbasmanova:
CC: @agrawaldevesh

[...]

PR #20932, opened 2023-09-21
Allow Presto clients to receive query results in binary format

mbasmanova:
Introduce 'binaryResults=true' query parameter to request query results returned in binary format https://prestodb.io/docs/current/develop/serialized-page.html

With binaryResults=true, the result json will contain 'binaryData' field of type array with one or
more pages of the results. The 'data' field will not be included in the result JSON.

See #20886

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
+++ b/presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
@@ -50,6 +50,7 @@ import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
+import io.airlift.slice.DynamicSliceOutput;
 import io.airlift.units.DataSize;
 import io.airlift.units.Duration;
 
@@ -61,6 +62,7 @@ import javax.ws.rs.core.UriBuilder;
 import javax.ws.rs.core.UriInfo;
 
 import java.net.URI;
+import java.util.Base64;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
@@ -81,6 +83,7 @@ import static com.facebook.presto.execution.QueryState.FAILED;
 import static com.facebook.presto.execution.QueryState.WAITING_FOR_PREREQUISITES;
 import static com.facebook.presto.server.protocol.QueryResourceUtil.toStatementStats;
 import static com.facebook.presto.spi.StandardErrorCode.GENERIC_INTERNAL_ERROR;
+import static com.facebook.presto.spi.page.PagesSerdeUtil.writeSerializedPage;
 import static com.facebook.presto.util.Failures.toFailure;
 import static com.google.common.base.MoreObjects.firstNonNull;
 import static com.google.common.base.Preconditions.checkArgument;
@@ -94,6 +97,7 @@ import static java.util.Objects.requireNonNull;
 class Query
 {
     private static final Logger log = Logger.get(Query.class);
+    private static final Base64.Encoder BASE64_ENCODER = Base64.getEncoder();
 
     private final QueryManager queryManager;
     private final TransactionManager transactionManager;
@@ -307,7 +311,7 @@ class Query
         return removedSessionFunctions;
     }
 
-    public synchronized ListenableFuture<QueryResults> waitForResults(long token, UriInfo uriInfo, String scheme, Duration wait, DataSize targetResultSize)
+    public synchronized ListenableFuture<QueryResults> waitForResults(long token, UriInfo uriInfo, String scheme, Duration wait, DataSize targetResultSize, boolean binaryResults)
     {
         // before waiting, check if this request has already been processed and cached
         Optional<QueryResults> cachedResult = getCachedResult(token);
@@ -323,7 +327,7 @@ class Query
                 timeoutExecutor);
 
         // when state changes, fetch the next result
-        return Futures.transform(futureStateChange, ignored -> getNextResultWithRetry(token, uriInfo, scheme, targetResultSize), resultsProcessorExecutor);
+        return Futures.transform(futureStateChange, ignored -> getNextResultWithRetry(token, uriInfo, scheme, targetResultSize, binaryResults), resultsProcessorExecutor);
     }
 
     private synchronized ListenableFuture<?> getFutureStateChange()
@@ -376,9 +380,9 @@ class Query
         return Optional.empty();
     }
 
-    private synchronized QueryResults getNextResultWithRetry(long token, UriInfo uriInfo, String scheme, DataSize targetResultSize)
+    private synchronized QueryResults getNextResultWithRetry(long token, UriInfo uriInfo, String scheme, DataSize targetResultSize, boolean binaryResults)
     {
-        QueryResults queryResults = getNextResult(token, uriInfo, scheme, targetResultSize);
+        QueryResults queryResults = getNextResult(token, uriInfo, scheme, targetResultSize, binaryResults);
         if (queryResults.getError() == null || !queryResults.getError().isRetriable()) {
             return queryResults;
         }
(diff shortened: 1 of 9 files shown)


tdcmeehan:
Please consider adding either a query or path parameter to toggle the client results, as clients may set arbitrary headers, and without migration of clients to support this format, it will cause them to break unexpectedly (the `data` field will be null, which will indicate query completion).

mbasmanova:
> Please consider adding either a query or path parameter to toggle the client results, as clients may set arbitrary headers, and without migration of clients to support this format, it will cause them to break unexpectedly (the `data` field will be null, which will indicate query completion).

@tdcmeehan Tim, thank you for reviewing. I updated to use query parameter binaryResults=true instead of the header. Would you, please, take another look?

presto-main/src/test/java/com/facebook/presto/server/TestServer.java
```diff
@@ -135,6 +146,63 @@ public void testServerStarts()
...
+                    .setUri(nextUri)
+                    .build();
+            queryResults = client.execute(nextRequest, createJsonResponseHandler(QUERY_RESULTS_CODEC));
+
+            assertNull(queryResults.getData());
+            if (queryResults.getBinaryData() != null) {
+                data.addAll(queryResults.getBinaryData());
+            }
+        }
+        assertNull(queryResults.getError());
```
tdcmeehan, line 184: Perhaps print out the error in case this test becomes flaky

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
```diff
@@ -442,25 +447,48 @@ private synchronized QueryResults getNextResult(long token, UriInfo uriInfo, Str
         // last page is removed.  If another thread observes this state before the response is cached
         // the pages will be lost.
         Iterable<List<Object>> data = null;
+        List<Object> binaryData = null;
```
tdcmeehan, line 450: Nit: don't we know that the parameter of the List is String?

tdcmeehan (commented):
In general, whereas `data` is `Object` because we don't know the underlying types (primitives, lists, etc.), don't we know that `binaryData` is `String` since it's Base64 encoded?

mbasmanova:
@tdcmeehan Good points, Tim. Updated. Please, take another look.

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
```diff
@@ -442,25 +447,48 @@ private synchronized QueryResults getNextResult(long token, UriInfo uriInfo, Str
...
-            long bytes = 0;
             long rows = 0;
+            long bytes = 0;
             long targetResultBytes = targetResultSize.toBytes();
-            while (bytes < targetResultBytes) {
-                SerializedPage serializedPage = exchangeClient.pollPage();
-                if (serializedPage == null) {
-                    break;
+            if (binaryResults) {
+                binaryData = new ArrayList<>();
```
tdcmeehan, line 456: Nit: use immutable list (consistent with below branch)?

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
```diff
@@ -442,25 +447,48 @@ private synchronized QueryResults getNextResult(long token, UriInfo uriInfo, Str
...
+                        break;
+                    }
+
+                    rows += serializedPage.getPositionCount();
+                    bytes += serializedPage.getSizeInBytes();
+
+                    DynamicSliceOutput sliceOutput = new DynamicSliceOutput(1000);
+                    PagesSerdeUtil.writeSerializedPage(sliceOutput, serializedPage);
+
+                    String encodedPage = Base64.getEncoder().encodeToString(sliceOutput.slice().byteArray());
```
tdcmeehan, line 469: Nit: Add a constant?

```suggestion
                    String encodedPage = BASE64_ENCODER.encodeToString(sliceOutput.slice().byteArray());
```

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
```diff
@@ -442,25 +447,48 @@ private synchronized QueryResults getNextResult(long token, UriInfo uriInfo, Str
...
+                    SerializedPage serializedPage = exchangeClient.pollPage();
+                    if (serializedPage == null) {
+                        break;
+                    }
+
+                    rows += serializedPage.getPositionCount();
+                    bytes += serializedPage.getSizeInBytes();
+
+                    DynamicSliceOutput sliceOutput = new DynamicSliceOutput(1000);
+                    PagesSerdeUtil.writeSerializedPage(sliceOutput, serializedPage);
```
tdcmeehan, line 467: Nit: static import?

```suggestion
                    writeSerializedPage(sliceOutput, serializedPage);
```

tdcmeehan (approved):
LGTM % a few more nits, feel free to merge after addressing!

mbasmanova:
@tdcmeehan Tim, thank you for the review and offline discussion. Addressed all comments and updated the PR.

steveburnett (approved):
LGTM! (docs)

presto-main/src/main/java/com/facebook/presto/server/protocol/QueuedStatementResource.java
```diff
@@ -371,16 +375,19 @@ private static URI getQueryHtmlUri(QueryId queryId, UriInfo uriInfo, String xFor
...
     }
 
-    private static URI getQueuedUri(QueryId queryId, String slug, long token, UriInfo uriInfo, String xForwardedProto, String xPrestoPrefixUrl)
+    private static URI getQueuedUri(QueryId queryId, String slug, long token, UriInfo uriInfo, String xForwardedProto, String xPrestoPrefixUrl, boolean binaryResults)
     {
-        URI uri = uriInfo.getBaseUriBuilder()
+        UriBuilder uriBuilder = uriInfo.getBaseUriBuilder()
                 .scheme(getScheme(xForwardedProto, uriInfo))
-                .replacePath("/v1/statement/queued/")
+                .replacePath("/v1/statement/queued")
```
agrawaldevesh, line 382: Just a Q : Why was the trailing slash removed ?

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
```diff
@@ -442,25 +447,51 @@ private synchronized QueryResults getNextResult(long token, UriInfo uriInfo, Str
...
-            long bytes = 0;
             long rows = 0;
+            long bytes = 0;
             long targetResultBytes = targetResultSize.toBytes();
-            while (bytes < targetResultBytes) {
-                SerializedPage serializedPage = exchangeClient.pollPage();
-                if (serializedPage == null) {
-                    break;
+            if (binaryResults) {
+                ImmutableList.Builder<String> pages = ImmutableList.builder();
```
agrawaldevesh, line 456: Nit: How do you feel about merging these two of the loops ? 

Also is it expensive to recreate DynamicSliceOutput(1000) each time in the loop ? No idea, so just asking :-)

presto-main/src/test/java/com/facebook/presto/server/TestServer.java
```diff
@@ -135,6 +147,63 @@ public void testServerStarts()
...
+                    .build();
+            queryResults = client.execute(nextRequest, createJsonResponseHandler(QUERY_RESULTS_CODEC));
+
+            assertNull(queryResults.getData());
+            if (queryResults.getBinaryData() != null) {
+                data.addAll(queryResults.getBinaryData());
+            }
+        }
+
+        if (queryResults.getError() != null) {
```
agrawaldevesh, line 183: nit: Is there a Util class where we can create this helper method that is an improvement on assertNull(queryResults.getError()) that is used in a couple of places now (this whole check and fail descriptively thing) :-)

Issue #20848, opened 2023-09-13
[Native] CTAS E2E roundtrip test


yzhang1991:
Write in C++ and read back in Java and see if result match.

pdabre12:
@yzhang1991  Can you please describe the issue?

yzhang1991:
@majetideepak can you help give some guidance on how the local roundtrip test can be done?

majetideepak:
https://github.com/prestodb/presto/blob/master/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeTpchQueries.java 
should give a direction. Here, we use the Java worker to write and the Bative worker to read. 
We can create a similar test and use a Native worker to write CTAS and a Java worker to read,

majetideepak:
@pramodsatya has been working with E2E tests lately. He can give further guidance.

PR #20936, opened 2023-09-21
[native] Add CTAS tests

pdabre12:
Resolves: https://github.com/prestodb/presto/issues/20848

### DIFF
--- a/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeCtasQueries.java
+++ b/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeCtasQueries.java
@@ -1,0 +1,107 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.nativeworker;
+
+import com.facebook.presto.Session;
+import com.facebook.presto.testing.QueryRunner;
+import com.facebook.presto.tests.AbstractTestQueryFramework;
+import org.testng.annotations.Test;
+
+import static com.facebook.presto.nativeworker.NativeQueryRunnerUtils.createLineitem;
+import static com.facebook.presto.nativeworker.NativeQueryRunnerUtils.createNation;
+import static com.facebook.presto.nativeworker.NativeQueryRunnerUtils.createOrders;
+
+public abstract class AbstractTestNativeCtasQueries
+        extends AbstractTestQueryFramework
+{
+    @Override
+    protected void createTables()
+    {
+        QueryRunner queryRunner = (QueryRunner) getExpectedQueryRunner();
+        createLineitem(queryRunner);
+        createOrders(queryRunner);
+        createNation(queryRunner);
+    }
+
+    @Test
+    public void testCreateTableAsSelect()
+    {
+        assertUpdate("CREATE TABLE IF NOT EXISTS test_ctas AS SELECT name, regionkey FROM nation", "SELECT count(*) FROM nation");
+        assertTableColumnNames("test_ctas", "name", "regionkey");
+        assertUpdate("DROP TABLE test_ctas");
+        assertCreateTableAsSelect(
+                "test_select",
+                "SELECT orderdate, orderkey, totalprice FROM orders",
+                "SELECT count(*) FROM orders");
+
+        assertCreateTableAsSelect(
+                "test_group",
+                "SELECT orderstatus, sum(totalprice) x FROM orders GROUP BY orderstatus",
+                "SELECT count(DISTINCT orderstatus) FROM orders");
+
+        assertCreateTableAsSelect(
+                "test_join",
+                "SELECT count(*) x FROM lineitem JOIN orders ON lineitem.orderkey = orders.orderkey",
+                "SELECT CAST(1 as BIGINT)");
+
+        assertCreateTableAsSelect(
+                "test_limit",
+                "SELECT orderstatus FROM orders ORDER BY orderstatus LIMIT 10",
+                "SELECT CAST(10 as BIGINT)");
+
+        assertCreateTableAsSelect(
+                "test_unicode",
+                "SELECT '\u2603' unicode",
+                "SELECT CAST(1 as BIGINT)");
+
+        assertCreateTableAsSelect(
+                "test_with_no_data",
+                "SELECT * FROM orders WITH NO DATA",
+                "SELECT * FROM orders LIMIT 0",
+                "SELECT CAST(0 as BIGINT)");
+
+//        Tests for CREATE TABLE with UNION ALL: exercises PushTableWriteThroughUnion optimizer
+
+        assertCreateTableAsSelect(
+                "test_union_all",
+                "SELECT orderdate, orderkey, totalprice FROM orders WHERE orderkey % 2 = 0 UNION ALL " +
+                        "SELECT orderdate, orderkey, totalprice FROM orders WHERE orderkey % 2 = 1",
+                "SELECT orderdate, orderkey, totalprice FROM orders",
+                "SELECT count(*) FROM orders");
+
+//        Setting the `redistribute_writes` system property enables the redistribution of data before writing to mitigate the performance
+//        impact in case the data is skewed. The default value is `true`.
+        assertCreateTableAsSelect(
+                Session.builder(getSession()).setSystemProperty("redistribute_writes", "true").build(),
+                "test_union_all_redistributed_writes",
+                "SELECT CAST(orderdate AS DATE) orderdate, orderkey, totalprice FROM orders UNION ALL " +
+                        "SELECT DATE '2000-01-01', 1234567890, 1.23",
+                "SELECT CAST(orderdate AS DATE) orderdate, orderkey, totalprice FROM orders UNION ALL " +
+                        "SELECT DATE '2000-01-01', 1234567890, 1.23",
+                "SELECT count(*) + 1 FROM orders");
+
+        assertCreateTableAsSelect(
+                Session.builder(getSession()).setSystemProperty("redistribute_writes", "false").build(),
+                "test_union_all",
+                "SELECT CAST(orderdate AS DATE) orderdate, orderkey, totalprice FROM orders UNION ALL " +
+                        "SELECT DATE '2000-01-01', 1234567890, 1.23",
+                "SELECT CAST(orderdate AS DATE), orderkey, totalprice FROM orders UNION ALL " +
+                        "SELECT DATE '2000-01-01', 1234567890, 1.23",
+                "SELECT count(*) + 1 FROM orders");
+
+        assertExplainAnalyze("EXPLAIN ANALYZE CREATE TABLE analyze_test AS SELECT orderstatus FROM orders");
+        assertQuery("SELECT * from analyze_test", "SELECT orderstatus FROM orders");
+        assertUpdate("DROP TABLE analyze_test");
+    }
+}
--- a/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/TestPrestoNativeCtasQueriesParquet.java
+++ b/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/TestPrestoNativeCtasQueriesParquet.java
@@ -1,0 +1,37 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.nativeworker;
+
+import com.facebook.presto.testing.ExpectedQueryRunner;
+import com.facebook.presto.testing.QueryRunner;
+import org.testng.annotations.Test;
+
+@Test(groups = {"parquet"})
+public class TestPrestoNativeCtasQueriesParquet
+        extends AbstractTestNativeCtasQueries
+{
+    @Override
+    protected QueryRunner createQueryRunner()
+            throws Exception
+    {
+        return PrestoNativeQueryRunnerUtils.createNativeQueryRunner(true, "PARQUET");
+    }
+
+    @Override
+    protected ExpectedQueryRunner createExpectedQueryRunner()
+            throws Exception
+    {
+        return PrestoNativeQueryRunnerUtils.createJavaQueryRunner("PARQUET");
+    }
+}


pdabre12:
@majetideepak Please review this PR.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeCtasQueries.java
```diff
@@ -0,0 +1,152 @@
...
+import static com.facebook.presto.nativeworker.NativeQueryRunnerUtils.*;
+import static java.lang.String.format;
+import static org.testng.Assert.assertFalse;
+import static org.testng.Assert.assertTrue;
+
+public abstract class AbstractTestNativeCtasQueries
+        extends AbstractTestQueryFramework {
+
+    @Override
+    protected void createTables() {
```
pramodsatya, line 31: Why are all the tables created here? Only the tables nation, region and orders seem to be used in the tests.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeCtasQueries.java
```diff
@@ -0,0 +1,152 @@
...
+        createCustomer(queryRunner);
+        createBucketedCustomer(queryRunner);
+        createPart(queryRunner);
+        createPartSupp(queryRunner);
+        createRegion(queryRunner);
+        createSupplier(queryRunner);
+    }
+
+    @Test
+    public void testCreateTableAsSelect() {
```
pramodsatya, line 48: This appears identical to the function `testCreateTableAsSelect` in `presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestDistributedQueries.java`. It could be reused by adding another abstract class for CTAS tests.

presto-native-execution/pom.xml
```diff
@@ -156,6 +156,11 @@
             <groupId>com.facebook.airlift</groupId>
             <artifactId>log-manager</artifactId>
         </dependency>
+        <dependency>
```
pramodsatya, line 159: This can be reverted after the test utility functions are moved to `AbstractTestQueryFramework.java`.

pramodsatya (commented):
Thanks for adding the tests @pdabre12, did an initial parse and left some comments.

majetideepak:
We want Parquet as the file format as well.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeCtasQueries.java
```diff
@@ -0,0 +1,152 @@
...
+
+    protected void assertCreateTableAsSelect(String table, @Language("SQL") String query, @Language("SQL") String expectedQuery, @Language("SQL") String rowCountQuery)
+    {
+        assertCreateTableAsSelect(getSession(), table, query, expectedQuery, rowCountQuery);
+    }
+
+    protected void assertCreateTableAsSelect(Session session, String table, @Language("SQL") String query, @Language("SQL") String expectedQuery, @Language("SQL") String rowCountQuery) {
+        assertUpdate(session, "CREATE TABLE " + table + " AS " + query, rowCountQuery);
+        assertQuery(session, "SELECT * FROM " + table, expectedQuery);
+        assertUpdate(session, "DROP TABLE " + table);
```
czentgr, line 139: Do you need to test DROP TABLE in the CTAS test?
You would need to allow DROP TABLE in https://github.com/prestodb/presto/blob/dd5b168ff5bdf306de70be19cc37c9797ca00981/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/NativeQueryRunnerUtils.java

and add `hive.allow-drop-table=true` as a hive property so you can drop the table - after you found that this is missing.

pdabre12: Thanks @czentgr for the comments. I did exactly that and it works , but these tests will only work locally/on my branch unless `hive.allow-drop-table=true` is being set in the main codebase. Should I add the property `hive.allow-drop-table=true` in the codebase and try to raise a different PR mentioning the access issues?

czentgr: That's why I'm asking if you truly need to test DROP table here. After all, this is a pure catalog operation which certainly is tested and possibly doesn;t need to be done for the native worker as it is unrelated.

pdabre12: It's not needed per say. Part of the reason why I was using them was because I don't want many tables to get created and persist there. I still do not know whether the tables are deleted after the tests are ran or they persist. If they are deleted, DROP operation is not needed.

czentgr: Yes, normally a good though to clean up. In this case, afaik, it is not needed because they won't persist past the test run in the hive metastore. Note how new tables are created at the beginning of the test as well.

pdabre12: Okay, understood.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeCtasQueries.java
```diff
@@ -0,0 +1,152 @@
...
+                "SELECT orderdate, orderkey, totalprice FROM orders UNION ALL " +
+                        "SELECT DATE '2000-01-01', 1234567890, 1.23",
+                "SELECT count(*) + 1 FROM orders");
+
+        assertExplainAnalyze();
+        assertQuery("SELECT * from analyze_test", "SELECT orderstatus FROM orders");
+        assertUpdate("DROP TABLE analyze_test");
+
+    }
+    protected void assertCreateTableAsSelect(String table, @Language("SQL") String query, @Language("SQL") String rowCountQuery)
```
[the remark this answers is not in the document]
pdabre12: I did that , but now I am not using the DROP operation , hence need to override the method.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeCtasQueries.java
pdabre12: I am adding some more queries later on, so need to override it anyways.

presto-native-execution/pom.xml
pdabre12: Because the test utility function was overridden , had to keep this change.

pdabre12:
@pramodsatya Made some changes based on your comments and added some files, some of the test cases are failing. 
I am creating an excel sheet showcasing all the passing/failing operations.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/TestPrestoNativeCtasQueriesParquetUsingJSON.java
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.nativeworker;
+
+import com.facebook.presto.testing.ExpectedQueryRunner;
+import com.facebook.presto.testing.QueryRunner;
+import org.testng.annotations.Test;
+
+@Test(groups = {"parquet"})
```
pramodsatya, line 20: Testing with thrift format is enough and the test files using Json format can be removed.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeCtasQueries.java
```diff
@@ -0,0 +1,127 @@
...
+        assertExplainAnalyze("EXPLAIN ANALYZE CREATE TABLE analyze_test AS SELECT orderstatus FROM orders");
+        assertQuery("SELECT * from analyze_test", "SELECT orderstatus FROM orders");
+    }
+
+    @Override
+    protected void assertCreateTableAsSelect(Session session, String table, @Language("SQL") String query, @Language("SQL") String expectedQuery, @Language("SQL") String rowCountQuery)
+    {
+        assertUpdate(session, "CREATE TABLE " + table + " AS " + query, rowCountQuery);
+        assertQuery(session, "SELECT * FROM " + table, expectedQuery);
+        assertFalse(getQueryRunner().tableExists(session, table));
```
pramodsatya, line 122: This should be reverted since the table is not being dropped. This function can also be moved to AbstractTestQueryFramework so the dependency on intellij annotations can be removed.

pramodsatya (commented):
Thanks for making the changes @pdabre12, left some comments. Could you please look into the test failures and ensure they are passing? The changes to submodule velox also need to be reverted.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/TestPrestoNativeCtasQueriesDwrfUsingThrift.java
```diff
@@ -0,0 +1,34 @@
...
+import com.facebook.presto.testing.QueryRunner;
+
+public class TestPrestoNativeCtasQueriesDwrfUsingThrift
+        extends AbstractTestNativeCtasQueries
+{
+    @Override
+    protected QueryRunner createQueryRunner()
+            throws Exception
+    {
+        return PrestoNativeQueryRunnerUtils.createNativeQueryRunner(true);
```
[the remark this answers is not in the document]
pdabre12: I wasn't sure whether its supported yet , but I will revoke the changes for now.

pdabre12:
Test cases document link: https://ibm.box.com/s/dyhiq2re1izs90wtikxh5psebhzok0a7

pramodsatya (approved):
Thanks for adding the tests, lgtm.

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueryFramework.java
```diff
@@ -339,6 +341,50 @@ protected void assertTableColumnNames(String tableName, String... columnNames)
...
+
+    protected void assertCreateTableAsSelect(Session session, String table, @Language("SQL") String query, @Language("SQL") String expectedQuery, @Language("SQL") String rowCountQuery)
+    {
+        assertUpdate(session, "CREATE TABLE " + table + " AS " + query, rowCountQuery);
+        assertQuery(session, "SELECT * FROM " + table, expectedQuery);
+        assertUpdate(session, "DROP TABLE " + table);
+        assertFalse(getQueryRunner().tableExists(session, table));
+    }
+
+    protected void assertCreateTableAsSelectWithoutDrop(String table, @Language("SQL") String query, @Language("SQL") String rowCountQuery)
```
aditi-pandit, line 362: It might be better to add an optional boolean parameter to drop the table in assertCreateTableAsSelect instead of adding all these new variations.

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestDistributedQueries.java
```diff
@@ -371,35 +371,6 @@ public void testExplainAnalyzeDDL()
         computeActual("EXPLAIN ANALYZE DROP TABLE orders");
     }
 
-    private void assertExplainAnalyze(@Language("SQL") String query)
```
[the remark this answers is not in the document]
pdabre12: To minimize code duplication as these functions are used in different tests files,  I moved them to 
`presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueryFramework.java`

presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedTestBase.java
```diff
@@ -220,11 +218,4 @@ public void testStringFilters()
         assertQuery("SELECT count(*) FROM test_varcharn_filter WHERE shipmode = 'AIR            '", "VALUES (0)");
         assertQuery("SELECT count(*) FROM test_varcharn_filter WHERE shipmode = 'NONEXIST'", "VALUES (0)");
     }
-
-    private void assertExplainAnalyze(@Language("SQL") String query)
```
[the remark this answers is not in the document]
pdabre12: I moved the function to `presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueryFramework.java`, hence removed it from here.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeCtasQueries.java
```diff
@@ -0,0 +1,104 @@
...
+import static com.facebook.presto.nativeworker.NativeQueryRunnerUtils.createNation;
+import static com.facebook.presto.nativeworker.NativeQueryRunnerUtils.createOrders;
+
+public abstract class AbstractTestNativeCtasQueries
+        extends AbstractTestQueryFramework
+{
+    @Override
+    protected void createTables()
+    {
+        QueryRunner queryRunner = (QueryRunner) getExpectedQueryRunner();
```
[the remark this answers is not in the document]
pdabre12: Will do so in a separate PR after this.

majetideepak:
@pdabre12 can we first refactor the functions and then add CTAS in a separate PR? It is hard to review otherwise.
I also see some commented code under `assertExplainAnalyze`.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeCtasQueries.java
majetideepak: I think we should stick with the DROP operation always. I agree with @pdabre12 that cleaning up is ideal. Enabling the API to not delete the tables can lead to bugs.
I don't see any harm in adding `hive.allow-drop-table=true` to the Native runner. I assume the Java runner has this as well.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeCtasQueries.java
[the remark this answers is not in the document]
czentgr: Isn't this is the Java query runner (not the native one) to set up the tables from the tpch and create some data to be later used as input for the CTAS of the native query runner?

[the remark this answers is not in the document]
pdabre12: Yes, my bad, you are right , it is the Java Runner .

pdabre12:
@majetideepak Please take a look.

majetideepak:
@pdabre12 I want to avoid refactoring existing API. Can we extend the native runner from AbstractTestDistributedQueries to get additional API?

pdabre12:
> @pdabre12 I want to avoid refactoring the existing API. Can we extend the AbstractTestNativeCtasQueries from AbstractTestDistributedQueries to get the additional API?

Thanks for the review, addressed the comment. 

The one reason I did not want to extend the `AbstractTestNativeCtasQueries` from `AbstractTestDistributedQueries` is while running test cases for a child class,  the test cases in the parent class will be run again.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeCtasQueries.java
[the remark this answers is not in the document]
majetideepak: Should we not be testing the create table part with the Native Query RUnner?
What is the purpose of this test?

pdabre12:
This PR is dependent on https://github.com/prestodb/presto/pull/21811.

majetideepak:
@pdabre12 can you share how much time these new tests take? It is harder to find this from the CI job.

pdabre12:
> @pdabre12 can you share how much time these new tests take? It is harder to find this from the CI job.

It takes approx ~43 seconds for the new tests to run.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeCtasQueries.java
```diff
@@ -0,0 +1,115 @@
...
+        createLineitem(queryRunner);
+        createOrders(queryRunner);
+        createNation(queryRunner);
+    }
+
+    @Test
+    public void testCreateTableAsSelect()
+    {
+        assertUpdate("CREATE TABLE IF NOT EXISTS test_ctas AS SELECT name, regionkey FROM nation", "SELECT count(*) FROM nation");
```
majetideepak, line 40: Should we test `CREATE TABLE` here without `IF NOT EXISTS`?

pdabre12: I think it might be a good idea to test `CREATE TABLE` without `IF NOT EXISTS` here.
In the other test cases except this, we are testing the `CREATE TABLE` without `IF NOT EXISTS` using the `assertCreateTableAsSelect()` function defined here: https://github.com/prestodb/presto/blob/3df58d1e0f801a331aa9e461375c1027ba5f866f/presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueryFramework.java#L378

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeCtasQueries.java
```diff
@@ -0,0 +1,115 @@
...
+        assertTableColumnNames("test_ctas", "name", "regionkey");
+        assertUpdate("DROP TABLE test_ctas");
+        //Some connectors support CREATE TABLE AS but not the ordinary CREATE TABLE. Let's test CTAS IF NOT EXISTS with a table that is guaranteed to exist.
+        assertUpdate("CREATE TABLE IF NOT EXISTS nation AS SELECT orderkey, discount FROM lineitem", 0);
```
majetideepak, line 44: I think we can skip this test. We are not testing any Prestissimo path here.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeCtasQueries.java
```diff
@@ -0,0 +1,115 @@
...
+        assertTableColumnNames("nation", "nationkey", "name", "regionkey", "comment");
```
majetideepak, line 45: Remove this as well.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeCtasQueries.java
```diff
@@ -0,0 +1,115 @@
...
+                "SELECT CAST(10 as BIGINT)");
+
+        assertCreateTableAsSelect(
+                "test_unicode",
+                "SELECT '\u2603' unicode",
+                "SELECT CAST(1 as BIGINT)");
+
+        assertCreateTableAsSelect(
+                "test_with_data",
+                "SELECT * FROM orders WITH DATA",
```
majetideepak, line 74: This is again a coordinator path as this is the default. Let's remove this test.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeCtasQueries.java
```diff
@@ -0,0 +1,115 @@
...
+
+        assertCreateTableAsSelect(
+                "test_union_all",
+                "SELECT orderdate, orderkey, totalprice FROM orders WHERE orderkey % 2 = 0 UNION ALL " +
+                        "SELECT orderdate, orderkey, totalprice FROM orders WHERE orderkey % 2 = 1",
+                "SELECT orderdate, orderkey, totalprice FROM orders",
+                "SELECT count(*) FROM orders");
+
+        assertCreateTableAsSelect(
+                Session.builder(getSession()).setSystemProperty("redistribute_writes", "true").build(),
```
majetideepak, line 94: Can you add a comment on what `redistribute_writes` does? What is the default value?

majetideepak (commented):
@pdabre12 some comments. Thanks.

pdabre12:
@majetideepak Addressed the comments, PTAL.
Thanks.

PR #20937, opened 2023-09-21
[Iceberg] Add changelog table

ZacBlanco:
## Description

This commit exposes a new system table called `$changelog` that
allows users to return a table containing information about when
which snapshots specific records were added or deleted (or updated).

The schema of the changelog table is static (except for the `rowdata` column):

```
| operation: VARCHAR | ordinal: bigint | snapshotid: bigint | rowdata: row(<parent table schema>) |
```

To query the changelog table, users simply need to provide the starting
snapshot ID for which they would like to query. The table name should be
in the format:

```sql
... FROM "<table_name>@<begin snapshot>$changelog@<end snapshot>"
```

The below query is an example of how the changelog table can be used to view the time that a particular record was inserted

```

SELECT committed_at, change.operation, rowdata.orderkey as orderkey, ordinal
FROM "orders$snapshots" as snap
JOIN "orders@5922410230513748202$changelog" as change 
ON change.snapshotid = snap.snapshot_id LIMIT 3;

                committed_at                 | operation | orderkey | ordinal
---------------------------------------------+-----------+----------+---------
 2023-09-21 16:36:49.098 America/Los_Angeles | INSERT    |    18019 |       0
 2023-09-21 16:36:49.098 America/Los_Angeles | INSERT    |    18018 |       0
 2023-09-21 16:36:49.098 America/Los_Angeles | INSERT    |    18017 |       0
```

At a high-level the implementation strategy needed to expose information
from Iceberg's `IncrementalChangelogScan` tasks which differs from a
normal `TableScan`. The plan tasks for incremental scans include the
information on operation type, ordinal, and snapshot ID which are are not
contained in the normal table scan plan.

When the connector metadata queries for a changelog table type, it
returns a schema with a static 4 columns, where the first 3 are static
VARCHAR, BIGINT, BIGINT, and the final is a row type whose inner types
are dependent upon the queried table.

To accomplish this the following changes were made
- logic to generate the changelog table schema
- a new piece of info on iceberg splits containing the original table's
column handles which is used during split sourcing
- a new page source which constructs the correct format page based on
the original table schema
- finally, some small refactoring of the IcebergNativeMetadata and
IcebergHiveMetadata classes to consolidate certain bits of logic for the
ConnectorMetadata

## Motivation and Context

There has been ongoing work to implement incremental sample maintenance within Presto to support more up-to-date optimizer statistics. Implementing this changelog table is another step towards reaching that goal.

## Impact

A new "system" table that users can query from the Iceberg connector.

## Test Plan

- simple queries
- join queries on different fields of the changelog table
- different combinations of column projections on the changelog table

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* A new changelog table has been added to the Iceberg connector
```


ZacBlanco:
I recognize that this is a large PR. In terms of review it might be helpful to review in chunks, with each review corresponding to a specific chunk of functionality. Below is my recommended breakdown for review (in no particular order):

1. `presto-main/*`: changes  to allow a "specialized" (read, generic) aggregation function as a connector specific plugin
2. `presto-iceberg/**/function/changelog` and `IcebergPlugin.java`: Implementation of the `apply_changelog` function and use of the new `@FunctionInstance` annotation.
3. `presto-iceberg/**/Iceberg*Metadata*.java`: Metadata-related changes to support the changelog table
4. `presto-iceberg/**/changelog/*` and `Iceberg*Split*.java` and `IcebergPage*.java`: Changes related to providing changelog data.
5. Remaining files: mostly small changes to existing classes or utilities to support one or more of the other major sections above.

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -352,6 +352,72 @@ as a part of a SQL query by appending name to the table.
...
+table, while also making the raw data available to query.
+
+The result of a changelog query always returns a static schema with 4 columns:
+
+1. ``operation``: (``VARCHAR``) indicating whether the row was inserted,
+   updated, or deleted.
+2. ``ordinal``: (``int``) A number indicating a relative order that a particular
+   change needs to be applied to the table relative to all other changes.
+3. ``snapshotid``: (``bigint``) Represents the snapshot a row-level
+   change was made in
```
steveburnett, line 369:
```suggestion
   change was made in.
```

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -352,6 +352,72 @@ as a part of a SQL query by appending name to the table.
...
+4. ``rowdata``: (``row(T)``) which includes the data for the particular row. The
+   inner values of this type match the schema of the parent table.
+
+The changelog table can be queried with the following name format
+
+.. code-block:: sql
+
+    ... FROM "<table>@<begin snapshot ID>$changelog@[<end snapshot ID>]"
+
+- ``<table>`` is the name of the table
```
steveburnett, line 379:
```suggestion
- ``<table>`` is the name of the table.
```

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -352,6 +352,72 @@ as a part of a SQL query by appending name to the table.
...
+- ``<begin snapshot ID>`` is the snapshot of the table you want to begin viewing
```
steveburnett, line 380:
```suggestion
- ``<begin snapshot ID>`` is the snapshot of the table you want to begin viewing.
```

steveburnett (commented):
LGTM! (docs)

I added a few very small punctuation suggestions for grammatical consistency, but didn't find anything significant to suggest.

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -352,6 +352,72 @@ as a part of a SQL query by appending name to the table.
...
+
+    ... FROM "<table>@<begin snapshot ID>$changelog@[<end snapshot ID>]"
+
+- ``<table>`` is the name of the table.
+- ``<begin snapshot ID>`` is the snapshot of the table you want to begin viewing
+  changes from.
+- ``<end snapshot ID>`` is the last snapshot for which you want to view changes.
+  This parameter is optional. If left out, the most current snapshot of the
+  table is used.
+
```
steveburnett, line 385: I was a little confused here, let me explain why. 
You wrote "The result of a changelog query always returns a static schema with 4 columns:"
and this first example I find six columns, not all of which match the four. 

Rereading I realize that this six-column example is about finding the start and optional ending snapshot ID that are then used in the actual changelog query. Do I understand this correctly?

Maybe, after the sentence "If left out, the most current snapshot of the table is used." you could add an explanatory introduction here, something like

"To find the snapshot_id value or values, run a command similar to the following:

SELECT * FROM "<table>$snapshots";

For example, to query for the snapshots of the orders table, run the following command: "

My example isn't formatted fully, so please use or adapt it and format such text as "snapshot_id" and "orders" as inline code.

steveburnett (changes requested):
Nice work! I was a little confused at one point. ! explain why, what I think I understand, and assuming I correctly understand provide the direction of a suggested change to make. 

I hope this makes sense. Let me know if my assumptions are wrong or if you have questions about this.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/changelog/ChangelogPageSource.java
```diff
@@ -0,0 +1,160 @@
...
+ *     <li>primary key (T) - this column is the type and value of the primary key that identifies the row</li>
+ *     <li>snapshot ID (long) - the snapshot id which this change belongs</li>
+ *     <li>data (row) - a column containing the data corresponding to this change</li>
+ * </ul>
+ */
+public class ChangelogPageSource
+        implements ConnectorPageSource
+{
+    private final ConnectorPageSource delegate;
+    private final ConnectorSplit split;
```
ClarenceThreepwood, line 55: unused

presto-iceberg/src/main/java/com/facebook/presto/iceberg/changelog/ChangelogPageSource.java
```diff
@@ -0,0 +1,160 @@
...
+    private final ChangelogSplitInfo changelogSplitInfo;
+    private final List<IcebergColumnHandle> desiredColumns;
+    private final List<IcebergColumnHandle> delegateColumns;
```
ClarenceThreepwood, line 58: unused

presto-iceberg/src/main/java/com/facebook/presto/iceberg/changelog/ChangelogPageSource.java
```diff
@@ -0,0 +1,160 @@
...
+
+    public ChangelogPageSource(ConnectorPageSource delegate, ChangelogSplitInfo changelogSplitInfo, ConnectorSplit split, List<IcebergColumnHandle> desiredColumns, List<IcebergColumnHandle> delegateColumns)
+    {
+        this.delegate = requireNonNull(delegate, "delegate is null");
+        this.split = requireNonNull(split, "split is null");
+        this.changelogSplitInfo = requireNonNull(changelogSplitInfo, "changelogSplitInfo is null");
+        this.desiredColumns = requireNonNull(desiredColumns, "columns is null");
+        this.delegateColumns = requireNonNull(delegateColumns, "delegateColumns is null");
+        this.innerRowType = RowType.from(delegateColumns.stream().map(x -> new RowType.Field(Optional.of(x.getName()), x.getType())).collect(Collectors.toList()));
+//        this.primaryKeyIndex = IntStream.range(0, delegateColumns.size())
```
ClarenceThreepwood, line 69: pls remove commented code

presto-iceberg/src/main/java/com/facebook/presto/iceberg/changelog/ChangelogPageSource.java
```diff
@@ -0,0 +1,160 @@
...
+    @Override
+    public void close()
+            throws IOException
+    {
+        delegate.close();
+    }
+
+    public enum ChangelogSchemaColumns
+    {
+        OPERATION((source, page) -> RunLengthEncodedBlock.create(VARCHAR, Slices.utf8Slice(source.changelogSplitInfo.getOperation().toString()), page.getPositionCount())),
```
ClarenceThreepwood, line 133: ChangelogOperation appears to have two flavors of update - UPDATE_BEFORE and UPDATE_AFTER. Can you describe the difference between them?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/changelog/ChangelogSplitInfo.java
```diff
@@ -0,0 +1,80 @@
...
+{
+    private final ChangelogOperation operation;
+    private final long changeOrdinal;
+    private final long snapshotId;
+    private final List<IcebergColumnHandle> icebergColumns;
+
+    @JsonCreator
+    public ChangelogSplitInfo(
+            @JsonProperty("operation") ChangelogOperation operation,
+            @JsonProperty("changeOrdinal") long changeOrdinal,
```
ClarenceThreepwood, line 36: just "ordinal" to be consistent?

presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergTableChangelog.java
```diff
@@ -0,0 +1,200 @@
...
+                        "('ordinal', 'bigint', '', '')," +
+                        "('snapshotid', 'bigint', '', '')," +
+                        "('rowdata', 'row(\"orderkey\" bigint, \"custkey\" bigint, \"orderstatus\" varchar, \"totalprice\" double, \"orderdate\" date, \"orderpriority\" varchar, \"clerk\" varchar, \"shippriority\" integer, \"comment\" varchar)', '', '')");
+    }
+
+    @Test
+    public void testBasicSelect()
+    {
+        for (long id : snapshots) {
+            assertQuerySucceeds(String.format("SELECT * FROM \"ctas_orders@%d$changelog\"", id));
```
ClarenceThreepwood, line 75: In this file it would also be good to validate results in addition to checking whether the query succeeds.
For e.g. select a subset of columns from the changelog table (I believe only snapshotid is non-deterministic?) and use assertQuery where possible

presto-main/src/main/java/com/facebook/presto/metadata/FunctionExtractor.java
```diff
@@ -64,6 +65,11 @@ public static List<? extends SqlFunction> extractFunctions(Class<?> clazz)
             return SqlInvokedScalarFromAnnotationsParser.parseFunctionDefinition(clazz);
         }
 
+        List<? extends SqlFunction> annotatedFunctionInstances = parseFunctionsFromAnnotation(clazz);
```
ClarenceThreepwood, line 68: Can you explain why we cannot just use SqlAggregationFunction::createFunctionByAnnotations ?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/function/changelog/ApplyChangelogFunction.java
```diff
@@ -0,0 +1,174 @@
...
+import static com.facebook.presto.operator.aggregation.AggregationUtils.generateAggregationName;
+import static com.facebook.presto.spi.function.Signature.typeVariable;
+import static com.facebook.presto.spi.function.aggregation.AggregationMetadata.ParameterMetadata.ParameterType.BLOCK_INDEX;
+import static com.facebook.presto.spi.function.aggregation.AggregationMetadata.ParameterMetadata.ParameterType.BLOCK_INPUT_CHANNEL;
+import static com.facebook.presto.spi.function.aggregation.AggregationMetadata.ParameterMetadata.ParameterType.INPUT_CHANNEL;
+import static com.facebook.presto.spi.function.aggregation.AggregationMetadata.ParameterMetadata.ParameterType.STATE;
+import static com.facebook.presto.util.Reflection.methodHandle;
+import static com.google.common.collect.ImmutableList.toImmutableList;
+
+public class ApplyChangelogFunction
```
ClarenceThreepwood, line 54: Any tests for this?

ClarenceThreepwood (changes requested):
Nice work Zac! 
Some nits and questions. 
Main comment - consider restructuring this PR into two commits. One focusing on the sql function and the other introducing the changelog table

presto-docs/src/main/sphinx/connector/iceberg.rst
ZacBlanco: I agree with your analysis the text didn't describe why we queried the `$snapshots` table first. I added a few sentences beforehand to describe the example first. Please let me know if this helps clarify!

presto-docs/src/main/sphinx/connector/iceberg.rst
steveburnett: This was a careless mistake I made. I'm glad you noticed and didn't accept this suggestion!

presto-docs/src/main/sphinx/connector/iceberg.rst
steveburnett: Looks great, thanks!

steveburnett: This definitely clarifies the description. Nice work!

steveburnett (approved):
LGTM! (docs)

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
```diff
@@ -185,7 +202,10 @@ public Map<SchemaTableName, List<ColumnMetadata>> listTableColumns(ConnectorSess
         ImmutableMap.Builder<SchemaTableName, List<ColumnMetadata>> columns = ImmutableMap.builder();
         for (SchemaTableName table : tables) {
             try {
-                columns.put(table, getTableMetadata(session, table).getColumns());
+                IcebergTableName tableName = IcebergTableName.from(table.getTableName());
+                if (!tableName.isSystemTable()) {
```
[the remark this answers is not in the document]
ZacBlanco: Without this, the `TestIcebergSystemTables` suite fails - it prevents returning results for columns when the table is a system table. This path is triggered due to the current implementation of the `MetadataManager`

The call site for this `listTableColumns` method is here:

https://github.com/prestodb/presto/blob/0.284/presto-main/src/main/java/com/facebook/presto/metadata/MetadataManager.java#L558

However, this method is called for [_all_ of the metadata](https://github.com/prestodb/presto/blob/0.284/presto-main/src/main/java/com/facebook/presto/metadata/MetadataManager.java#L558) implementations in [`listConnectorIds`](https://github.com/prestodb/presto/blob/0.284/presto-main/src/main/java/com/facebook/presto/metadata/CatalogMetadata.java#L124-L127), not just the `SystemTable` ones. Thus, in order to return the proper columns for system tables, we need to make sure this method returns nothing for non-system tables.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergPageSourceProvider.java
```diff
@@ -648,13 +649,18 @@ public ConnectorPageSource createPageSource(
...
+            List<ColumnHandle> desiredColumns,
             SplitContext splitContext)
     {
         IcebergSplit split = (IcebergSplit) connectorSplit;
         IcebergTableLayoutHandle icebergLayout = (IcebergTableLayoutHandle) layout;
         IcebergTableHandle table = icebergLayout.getTable();
 
+        List<ColumnHandle> columns = desiredColumns;
+        if (split.getChangelogSplitInfo().isPresent()) {
+            columns = (List<ColumnHandle>) (List<?>) split.getChangelogSplitInfo().get().getIcebergColumns();
```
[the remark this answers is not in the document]
ZacBlanco: So the `desiredColumns` field represents the list of columns the query is interested in reading. i.e. if you had table `t(c1 int, c2 int)`, and queried `SELECT c1 FROM t` this list would only contain the columns from `c1`.

In the case of the changelog, if we queried the changelog with

```sql
SELECT ordinal FROM t$changelog
```

Then the `desiredColumns` list is populated with the handle to the `ordinal` column. However, that column doesn't actually exist in the parquet file. We need to pass a valid set of columns to the `createPageDataSource` function to build a parquet/ORC reader that only reads the desired columns. In the case of the changelog, we will always read the full set.

In summary, This conditional exists because we need to pass the "real" column set from the source table schema into the `createPageDataSource`, rather than the "virtual" columns from the changelog schema to properly read the source data.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/changelog/ChangelogPageSource.java
ZacBlanco: [An example exists in the Iceberg documentation](https://iceberg.apache.org/docs/latest/spark-procedures/#prepost-update-images).

Essentially `UPDATE_BEFORE` and `UPDATE_AFTER` amount to is that they reflect the value of a row before an update and after an update within a particular ordinal. Whether updates are encoded like this depend on the table and Spark configuration used to perform the update.

As explained in the docs, if you had an update encoded as

| id | name | operation |
|-|-|-|
| 3 | Robert | DELETE |
| 3 | Dan | INSERT |

Spark could also encode the same information as:


| id | name | operation |
|-|-|-|
| 3 | Robert | UPDATE_BEFORE |
| 3 | Dan | UPDATE_AFTER |

presto-main/src/main/java/com/facebook/presto/metadata/FunctionExtractor.java
ZacBlanco: If you implement an aggregation function with annotations through `@AggregationFunction` the type constraints _must_ be specified in your subsequent aggregation function annotations. Thus, if you wanted to implement a function for more than one type, you need a method and annotation for each specific input type. As an example of multiple accepted types using the annotations, see something like [`AverageAggregations`](https://github.com/prestodb/presto/blob/0.284/presto-main/src/main/java/com/facebook/presto/operator/aggregation/AverageAggregations.java)

However, if you want to implement a function which can be specialized for _any_ type, then you need to extend the `SqlAggregationFunction` class and then pass an _instance_ of that function to the `BuiltInFunctionAndTypeManager#function` method. An example of this type of implementation can be found in [`ChecksumAggregationFunction`](https://github.com/prestodb/presto/blob/0.284/presto-main/src/main/java/com/facebook/presto/operator/aggregation/ChecksumAggregationFunction.java), but there are also many more implemented this way. 

However, since you need an instance of the aggregation class to register a generic aggregation function, the current plugin interface is insufficient because all you can do is pass a `Class<?>`. Thus, I added the ability to extract a static instance of the function which the `FunctionAndTypeManager` can use execute the function on any generic type.

tdcmeehan: I'm not sure I follow--can you reference `BuildSetDigestAggregation` and see if this does what you need it to do?  It is not specialized for any particular type.

ZacBlanco: Thanks for the pointer Tim - I didn't come across this before. . Most of the other aggregation function that accept a type parameter that I found previously all seemed to be implemented without the `@TypeParameter` annotation. I'll try to implement it using that

ZacBlanco: I made an intermediate PR which will allow us to simplify this implementation in #21316

ZacBlanco: The implementation has been updated to use the annotation framework and the rest of the changes around function parsing were reverted

presto-iceberg/src/main/java/com/facebook/presto/iceberg/function/changelog/ApplyChangelogFunction.java
```diff
@@ -0,0 +1,87 @@
...
+import com.facebook.presto.spi.function.InputFunction;
+import com.facebook.presto.spi.function.OutputFunction;
+import com.facebook.presto.spi.function.SqlType;
+import com.facebook.presto.spi.function.TypeParameter;
+import io.airlift.slice.Slice;
+import org.apache.iceberg.ChangelogOperation;
+
+import java.util.Optional;
+
+@AggregationFunction("apply_changelog")
```
tdcmeehan, line 33: Should these functions be hidden?  Do we expect people to use them directly?

ZacBlanco: I'm not exactly clear on what the difference between hidden and non-hidden functions are. I couldn't find it in the docs.

To give you an idea on the use case for this function, there will eventually be some system (may be internal or external to Presto) which runs a query like the following:

```SQL
SELECT 
    rowdata."${primary_key}", 
    apply_changelog(ordinal, operation, rowdata order by ordinal) as rowdata
FROM "${tableName}$changelog"
GROUP BY rowdata."${primary_key}"
```

Most of the time, users probably won't be running this directly but whatever application that performs sample maintenance will need to have access to the function

tdcmeehan: Hidden functions don't show up in `SHOW FUNCTIONS`, and we may deliberately leave them undocumented, but they are otherwise accessible.  Seems to fit this use case.

aaneja (approved):

ClarenceThreepwood (approved):

ClarenceThreepwood:
Please squash/merge commits appropriately

s-akhtar-baig (commented):
Hi @ZacBlanco! I had a few comments, let me know what you think. Thanks!

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergTableName.java
```diff
@@ -40,23 +48,47 @@ public class IcebergTableName
     private final TableType tableType;
     private final Optional<Long> snapshotId;
 
-    public IcebergTableName(String tableName, TableType tableType, Optional<Long> snapshotId)
+    private final Optional<Long> changelogEndSnapshot;
+
+    private static final Set<TableType> SYSTEM_TABLES = Sets.immutableEnumSet(FILES, MANIFESTS, PARTITIONS, HISTORY, SNAPSHOTS, PROPERTIES);
+
+    @JsonCreator
+    public IcebergTableName(
```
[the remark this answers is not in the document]
ZacBlanco: I agree that reading `getTableName().getTableName()` is confusing. What do you think about `TableNameMetadata`?

This class is used quite widely throughout the connector and I think refactoring in this PR would create a lot of noise in the changelist for review. What do you think about a follow-up PR to rename after this one is merged?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergPageSourceProvider.java
```diff
@@ -680,7 +686,11 @@ public ConnectorPageSource createPageSource(
                 table.getPredicate(),
                 splitContext.isCacheable());
 
-        return new IcebergPageSource(icebergColumns, partitionKeys, dataPageSource, session.getSqlFunctionProperties().getTimeZoneKey());
+        ConnectorPageSource dataSource = new IcebergPageSource(icebergColumns, partitionKeys, dataPageSource, session.getSqlFunctionProperties().getTimeZoneKey());
+        if (split.getChangelogSplitInfo().isPresent()) {
+            dataSource = new ChangelogPageSource(dataSource, split.getChangelogSplitInfo().get(), (List<IcebergColumnHandle>) (List<?>) desiredColumns, icebergColumns);
+        }
+        return dataSource;
```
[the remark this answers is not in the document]
ZacBlanco: This won't work because the first argument of the `new ChangelogPageSource` requires the `dataSource`.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/function/changelog/ChangelogRecord.java
```diff
@@ -0,0 +1,159 @@
...
+    public ChangelogRecord(Type stateType)
+    {
+        this.stateType = requireNonNull(stateType, "type is null");
+        this.serializedType = getSerializedRowType(stateType);
+        this.lastOperation = Slices.EMPTY_SLICE;
+        this.lastOrdinal = -1; // ensures this assumes the values from the first merge.
+    }
+
+    public ChangelogRecord(Type stateType, int lastOrdinal, Slice lastOperation, Block lastRow)
+
```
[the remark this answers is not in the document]
ZacBlanco: fixed, thanks

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergTableName.java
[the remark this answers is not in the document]
s-akhtar-baig: Yes, that sounds good!

s-akhtar-baig (approved):
Added a minor comment but everything else looks good to me! Thanks @ZacBlanco!

presto-iceberg/src/main/java/com/facebook/presto/iceberg/changelog/ChangelogPageSource.java
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * This is a page source which provides pages for the Changelog of an iceberg
+ * table.
+ * <br>
+ * The schema of the changelog table is always fixed to have the following columns:
+ *
+ * <ul>
+ *     <li>operation type (varchar) - a string representation of an enum value from {@link org.apache.iceberg.ChangelogOperation}</li>
+ *     <li>ordinal (int) - the order in which the change i applied relative to all other changes in the log</li>
```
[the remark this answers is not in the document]
ZacBlanco: fixed, thanks for pointing it out!

presto-iceberg/src/main/java/com/facebook/presto/iceberg/function/changelog/ApplyChangelogFunction.java
```diff
@@ -0,0 +1,89 @@
...
+        state.set(record);
+    }
+
+    @CombineFunction
+    public static void combine(@TypeParameter("T") Type type,
+            @AggregationState ApplyChangelogState state,
+            @AggregationState ApplyChangelogState otherState)
+    {
+        ChangelogRecord r = Optional.ofNullable(state.get())
+                .map(actual -> Optional.ofNullable(otherState.get()).map(actual::merge).orElse(actual)).orElse(otherState.get());
```
tdcmeehan, line 66: I find this nested optional block to be confusing.  I'm wondering if simple if/else might be more LOCs yet more readable.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/function/changelog/ChangelogRecord.java
```diff
@@ -0,0 +1,158 @@
...
+                    }
+                    lastOperation = other.lastOperation;
+                    lastRow = other.lastRow;
+                    lastOrdinal = other.lastOrdinal;
+                    break;
+                case DELETE:
+                    // in the case of delete, do nothing, keep the current state
+                    break;
+                default:
+                    if (!other.lastOperation.toStringUtf8().isEmpty()) {
```
tdcmeehan, line 111: Why would this be empty?

tdcmeehan (commented):
Mostly nits

presto-iceberg/src/main/java/com/facebook/presto/iceberg/changelog/ChangelogSplitSource.java
```diff
@@ -0,0 +1,139 @@
...
+            IncrementalChangelogScan tableScan,
+            double minimumAssignedSplitWeight)
+    {
+        this.session = requireNonNull(session, "session is null");
+        this.columnHandles = getColumns(table.schema(), typeManager);
+        this.table = requireNonNull(table, "table is null");
+        this.tableScan = requireNonNull(tableScan, "tableScan is null");
+        this.minimumAssignedSplitWeight = minimumAssignedSplitWeight;
+        this.fileScanTaskIterator = tableScan.planFiles().iterator();
+        closer.register(fileScanTaskIterator);
```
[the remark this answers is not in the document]
ZacBlanco: I was just following the pattern from the `IcebergSplitSource` for consistency. I can simplify it

presto-iceberg/src/main/java/com/facebook/presto/iceberg/function/changelog/ApplyChangelogFunction.java
ZacBlanco: I agree, the nested optionals can be confusing at first, but using them prevents NPEs. The `if`-condition version of the function would look like:

```java
        ChangelogRecord record;
        if (state.getRecord() == null && otherState.getRecord() == null) {
            return;
        } else if (state.getRecord() == null && otherState.getRecord() != null) {
            record = otherState.getRecord();
        } else if (state.getRecord() != null && otherState.getRecord() == null) {
            record = state.getRecord();
        } else {
            record = state.getRecord().merge(otherState.getRecord());
        }
        state.setRecord(record);
```

tdcmeehan (approved):

presto-iceberg/src/main/java/com/facebook/presto/iceberg/function/changelog/ApplyChangelogState.java
```diff
@@ -0,0 +1,135 @@
...
+
+    Optional<ChangelogRecord> getRecord();
+
+    void setRecord(@Nullable ChangelogRecord value);
+
+    class Single
+            implements ApplyChangelogState
+    {
+        private final Type innerType;
+        ChangelogRecord record;
```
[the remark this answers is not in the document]
ZacBlanco: fixed

tdcmeehan (approved):

PR #20940, opened 2023-09-22
Document DICTIONARY AND RLE encodings in SerializedPage Wire Format

mbasmanova:
```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-docs/src/main/sphinx/develop/serialized-page.rst
+++ b/presto-docs/src/main/sphinx/develop/serialized-page.rst
@@ -212,6 +212,20 @@ from the null flags.
 .. image:: serialized-page-row-column.png
   :width: 600
 
+DICTIONARY Encoding
+-------------------
+
+* Number of rows - 4 bytes
+* Dictionary values column. This column itself is a serialized block whose encoding can be any of the encodings mentioned in this document.
+* Indices - number of rows * 4 bytes; 4 bytes per index
+* Dictionary ID - 24 bytes
+
+RLE Encoding
+------------
+
+* Number of rows - 4 bytes
+* Single-row constant value column
+
 Additional Usage
 ----------------
 


presto-docs/src/main/sphinx/develop/serialized-page.rst
```diff
@@ -212,6 +212,20 @@ from the null flags.
 .. image:: serialized-page-row-column.png
   :width: 600
 
+DICTIONARY Encoding
+-------------------
+
+* Number of rows - 4 bytes
+* Dictionary values column
```
tdcmeehan, line 219: Correct me if I'm wrong, but I believe this may be a nested dictionary.  At the very least, should we consider mentioning that this values type is recursive (i.e. could be other types listed in this doc)?

presto-docs/src/main/sphinx/develop/serialized-page.rst
```diff
@@ -212,6 +212,20 @@ from the null flags.
 .. image:: serialized-page-row-column.png
   :width: 600
 
+DICTIONARY Encoding
```
tdcmeehan, line 215: Should we add diagrams here and for RLE, similar to the other entries?

tdcmeehan (commented):
Thank you for adding this documentation.

presto-docs/src/main/sphinx/develop/serialized-page.rst
mbasmanova: It would nice, but I can't find the original PowerPoint I used to create these diagrams.

presto-docs/src/main/sphinx/develop/serialized-page.rst
mbasmanova: I believe this is so. It would be similar to how 'elements' column of the ARRAY can have any encoding. 

Do you have any suggestion about where / how to mention this?

tdcmeehan: Perhaps something like this: `Dictionary values column.  This is column is itself a serialized block, meaning its format may be that of any type mentioning in this document.`

mbasmanova: @tdcmeehan Thank you, Tim. Updated. Would you take another look?

presto-docs/src/main/sphinx/develop/serialized-page.rst
tdcmeehan: No problem, but please add an issue and add the `docs` label post-merge so we don't lose track of it.

tdcmeehan (approved):
LGTM, @steveburnett would you like to have a quick glance?

steveburnett:
> LGTM, @steveburnett would you like to have a quick glance?

LGTM! (docs)

PR #20413, opened 2023-07-28
Fix join reordering for non-trivial equi join conditions

aaneja:
## Description
Enhance join-reordering to work with non-simple equi-join predicates

## Motivation and Context
Join predicates like `left.key = right0.key1 + right1.key2` can reduce the join space by appearing as Project nodes or Join noes with no equi-join clauses in the join graph. This commit fixes this behavior by removing any intermediate Projects in the join graph and only creating them on-the-fly while choosing the join order

## Impact
Queries with completely connected join graphs would have CrossJoin's in them, depending on the order of specification of the FROM clause
```
presto:tiny> explain select count(*) from customer c, partsupp ps, orders o, supplier s where s.suppkey = ps.suppkey and c.custkey = o.custkey and s.nationkey + ps.partkey = c.nationkey;
                                                                                                                               Query Plan                                                                                               >
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------->
 - Output[_col0] => [count:bigint]                                                                                                                                                                                                      >
         _col0 := count (1:16)                                                                                                                                                                                                          >
     - Aggregate(FINAL) => [count:bigint]                                                                                                                                                                                               >
             count := "presto.default.count"((count_16)) (1:16)                                                                                                                                                                         >
         - LocalExchange[SINGLE] () => [count_16:bigint]                                                                                                                                                                                >
             - RemoteStreamingExchange[GATHER] => [count_16:bigint]                                                                                                                                                                     >
                 - Aggregate(PARTIAL) => [count_16:bigint]                                                                                                                                                                              >
                         count_16 := "presto.default.count"(*) (1:16)                                                                                                                                                                   >
                     - InnerJoin[("suppkey" = "suppkey_5") AND (nationkey) = ((nationkey_8) + (partkey))][$hashvalue_20, $hashvalue_21] => []                                                                                           >
                             Estimates: {source: CostBasedSourceInfo, rows: 2400 (21.09kB), cpu: 11881391400.00, memory: 178200.00, network: 178200.00}                                                                                 >
                             Distribution: REPLICATED                                                                                                                                                                                   >
                         - Project[projectLocality = LOCAL] => [partkey:bigint, suppkey:bigint, nationkey:bigint, $hashvalue_20:bigint]                                                                                                 >
                                 Estimates: {source: CostBasedSourceInfo, rows: 120000000 (1.01GB), cpu: 7561381500.00, memory: 175500.00, network: 175500.00}                                                                          >
[...]

PR #20943, opened 2023-09-22
Revert "Enhance join reordering to work with non-simple equi-join clauses"

aaneja:
This reverts commit 009c06bfcd7e7b54c8b15fab5eb60e2b34f79d9f. from #20413

There is a subtle bug handling ProjectNodes between Join nodes that only do variable renames.
Reverting for now, while I figure out a better solution


feilong-liu (approved):

ajaygeorge:
@aaneja Please remember to add a test case surfacing the bug when you add a fix for this. (ideally as a safety net even before the fix)

ajaygeorge (approved):

PR #20558, opened 2023-08-14
Fix structural type map render

GaryHo34:
## Description
Fix the issue by parsing the value again via `jsonCodec` for ARRAY and ROW type.
For MAP type, `jsonCodec` cannot recognize the string `{a=b}`. 
I manually created a logic to parse it. Based on the assumption that the value must be a map.

## Motivation and Context
Fix the open issue [Structural Type as Map Key cannot be rendered by CLI](https://github.com/prestodb/presto/issues/7546)

## Impact
N/A

## Test Plan
N/A

## Contributor checklist

- [ ] Please make sure your submission complies with our [development](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#development), [formatting](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#formatting), [commit message](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#commit-formatting-and-pull-requests), and [attribution guidelines](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#attribution).
- [ ] PR description addresses the issue accurately and concisely.  If the change is non-trivial, a GitHub Issue is referenced.
- [ ] Documented new properties (with its default value), SQL syntax, functions, or other functionality.
- [ ] If release notes are required, they follow the [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines).
- [ ] Adequate tests were added if applicable.
- [ ] CI passed.

## Release Notes
Please follow [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines) and fill in the release notes below.

```
== NO RELEASE NOTE ==
```

Issue #20588, opened 2023-08-17
[Native] Added e2e test for binomial_cdf()


yzhang1991:
binomial_cdf() was added in https://github.com/facebookincubator/velox/pull/4686

PR #20949, opened 2023-09-23
Make parseToJson optional

GaryHo34:
To compatible with the clients who are dependent on previous formatting, make parseToJson an optional feature. Developers can toggle it off by setting `nested-data-serialize-enabled` in config.properties.

## Description
This PR aims to enable user to choose from legacy formatting or new formatting of the response data.

## Motivation and Context
The previous [#20558](https://github.com/prestodb/presto/pull/20558) fixed the nested data render issue but changed the behavior of response data.

## Test Plan
Tested with the `testExplicitPropertyMappings` and `testDefaults` funcions in `TestServerConfig.java`.

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==
General Changes
* Add support for serializing nested data structures via server-side configuration properties. The new pre-serialization feature can be disabled server-wide by using the configuration property: ``nested-data-serialization-enabled``.
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/server/ServerConfig.java
+++ b/presto-main/src/main/java/com/facebook/presto/server/ServerConfig.java
@@ -39,6 +39,7 @@ public class ServerConfig
     private boolean queryResultsCompressionEnabled = true;
     private NodePoolType poolType = DEFAULT;
     private Duration clusterStatsExpirationDuration = new Duration(0, MILLISECONDS);
+    private boolean nestedDataSerializationEnabled = true;
 
     public boolean isResourceManager()
     {
@@ -199,6 +200,18 @@ public class ServerConfig
     public ServerConfig setClusterStatsExpirationDuration(Duration clusterStatsExpirationDuration)
     {
         this.clusterStatsExpirationDuration = clusterStatsExpirationDuration;
+        return this;
+    }
+
+    public boolean isNestedDataSerializationEnabled()
+    {
+        return nestedDataSerializationEnabled;
+    }
+
+    @Config("nested-data-serialization-enabled")
+    public ServerConfig setNestedDataSerializationEnabled(boolean nestedDataSerializationEnabled)
+    {
+        this.nestedDataSerializationEnabled = nestedDataSerializationEnabled;
         return this;
     }
 }
--- a/presto-tests/src/test/java/com/facebook/presto/server/TestQueryResource.java
+++ b/presto-tests/src/test/java/com/facebook/presto/server/TestQueryResource.java
@@ -351,4 +351,32 @@ public class TestQueryResource
         assertEquals(running, expectedRunning);
         assertEquals(queued, expectedQueued);
     }
+
+    @Test
+    public void testShouldReturnLegacyDataOnNestDataSerializationDisabled()
+            throws Exception
+    {
+        runner = createQueryRunner(ImmutableMap.of("nested-data-serialization-enabled", "false"));
+        server = runner.getCoordinator();
+
+        String sql = "SELECT MAP(ARRAY['a'], ARRAY[1])";
+        URI uri = uriBuilderFrom(server.getBaseUrl().resolve("/v1/statement")).build();
+
+        Request request = preparePost()
+                .setHeader(PRESTO_USER, "user")
+                .setUri(uri)
+                .setBodyGenerator(createStaticBodyGenerator(sql, UTF_8))
+                .build();
+
+        String body = client.execute(request, createStringResponseHandler()).getBody();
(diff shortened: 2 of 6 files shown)


presto-main/src/main/java/com/facebook/presto/server/ServerConfig.java
```diff
@@ -201,4 +202,16 @@ public ServerConfig setClusterStatsExpirationDuration(Duration clusterStatsExpir
         this.clusterStatsExpirationDuration = clusterStatsExpirationDuration;
         return this;
     }
+
+    public boolean isNestedDataSerializeEnabled()
```
tdcmeehan, line 206: Here and elsewhere: `serialize` => `serialization`

tdcmeehan (changes requested):
One nit otherwise looks good

amitkdutta:
@GaryHo34 Wondering if we can add tests with for cases described here https://github.com/prestodb/presto/pull/20558#issuecomment-1731864115
basically set `nested-data-serialize-enabled` to false and check if old format is kept intact.
CC: @kaikalur @jainxrohit

presto-main/src/main/java/com/facebook/presto/server/ServerConfig.java
GaryHo34: Done!

GaryHo34:
> @GaryHo34 Wondering if we can add tests with for cases described here [#20558 (comment)](https://github.com/prestodb/presto/pull/20558#issuecomment-1731864115) basically set `nested-data-serialize-enabled` to false and check if old format is kept intact. CC: @kaikalur @jainxrohit

Ok, let me try this

presto-tests/src/test/java/com/facebook/presto/server/TestQueryResource.java
```diff
@@ -351,4 +351,32 @@ private void assertStateCounts(List<BasicQueryInfo> infos, int expectedFinished,
...
+    public void testShouldReturnLegacyDataOnNestDataSerializationDisabled()
+            throws Exception
+    {
+        runner = createQueryRunner(ImmutableMap.of("nested-data-serialization-enabled", "false"));
+        server = runner.getCoordinator();
+
+        String sql = "SELECT MAP(ARRAY['a'], ARRAY[1])";
+        URI uri = uriBuilderFrom(server.getBaseUrl().resolve("/v1/statement")).build();
+
+        Request request = preparePost()
```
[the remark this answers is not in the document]
GaryHo34: I want to retrieve the data in its string format before it is deserialized. This will allow me to ensure that the returned string aligns with our expectations. I believe that the getQueryResults function uses the JSON codec, which will deserialize the response.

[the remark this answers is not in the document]
tdcmeehan: Yeah, seems sensible to keep it as is since it's so specific to this test.

tdcmeehan (approved):

tdcmeehan:
Any other feedback folks?

pranjalssh (approved):
Looks good. Please add a release note.

presto-main/src/main/java/com/facebook/presto/server/ServerConfig.java
```diff
@@ -39,6 +39,7 @@ public class ServerConfig
     private boolean queryResultsCompressionEnabled = true;
     private NodePoolType poolType = DEFAULT;
     private Duration clusterStatsExpirationDuration = new Duration(0, MILLISECONDS);
+    private boolean nestedDataSerializationEnabled = true;
```
agrawaldevesh, line 42: Why is the default true ? Like why go against backward compatibility ?

tdcmeehan: We got consensus in #20588 that given this is a bug fix, the default should be to fix it, but allow a way to back out of it in case any reliance on this unfortunate behavior has been developed. Release notes will indicate to users the change in behavior, which is expected as any system evolves.

agrawaldevesh:
Thanks for adding this flag but I am wondering why:

1. This is not a session property via the SystemSessionProperties like others with whatever default you want ? 
2. Why is this in ServerConfig specifically ? (I understand some precedence there but think about it -- this is not about the "cluster" per se). A less bad home might have been QueryManagerConfig for that matte or FeaturesConfig.

@GaryHo34 is it possible to redo this with the above suggestions ? 

cc: @amitkdutta

tdcmeehan:
@agrawaldevesh I don't think selectively enabling this bug fix was an anticipated use case, but if you think it would be useful please raise an issue.

Issue #20926, opened 2023-09-21
Allow Parquet writer to write both V1 and V2 files


yingsu00:
Presto Parquet writer can only write V2 files. We want to support writing both V1 and V2 files and control the version by a toggle.

## Expected Behavior or Use Case
Introduce a session property to specify the Parquet file version. If V1 is specified, then file needs to be written in V1.

## Presto Component, Service, or Connector
Parquet Writer

## Possible Implementation
Refer to Trino https://github.com/trinodb/trino/pull/9497 and https://github.com/trinodb/trino/pull/9611

## Context
https://github.com/prestodb/presto/issues/17240

PR #19983, opened 2023-06-26
Match hive support for type widening in the hive connector

mblanco-denodo:
Test plan - Added test to hive module. Executed hive and parquet tests locally.
```
== RELEASE NOTES ==

General Changes
* Added type widening support to the parquet column readers (int, bigint, and float) to be able to perform type conversions between inmediate numerical types
* Updated docs accordingly (hive connector - schema evolution)

Hive Changes
* Changes in ParquetPageSourceFactory so that the type-check does not fail with the new widenings supported
```

PR #20957, opened 2023-09-25
Support Parquet writer versions V1 and V2

nmahadevuni:
Cherry-pick of https://github.com/trinodb/trino/pull/9497 and https://github.com/trinodb/trino/pull/9611

## Description
Adds support for writing Parquet data versions V1 and V2 using a config and session property.

## Motivation and Context
Fixes https://github.com/prestodb/presto/issues/20926

## Impact
No external changes. Default Parquet version is set to V2 which is currently used in Presto.

## Test Plan
Amended the ParquetTester.java to run round trip tests for both writer versions.

```
== RELEASE NOTES ==

Hive Changes
* Support for Parquet writer versions V1 and V2 is now added for Hive and Iceberg catalogs. It can be toggled using session property `parquet_writer_version` and config property `hive.parquet.writer.version`. Valid values for these properties are PARQUET_1_0 and PARQUET_2_0. Default is PARQUET_2_0. E.g., set session parquet_writer_version=PARQUET_1_0 / hive.parquet.writer.version=PARQUET_1_0
```

### DIFF
--- a/presto-hive/src/main/java/com/facebook/presto/hive/HiveSessionProperties.java
+++ b/presto-hive/src/main/java/com/facebook/presto/hive/HiveSessionProperties.java
@@ -22,6 +22,7 @@ import com.facebook.presto.spi.session.PropertyMetadata;
 import com.google.common.collect.ImmutableList;
 import io.airlift.units.DataSize;
 import io.airlift.units.Duration;
+import org.apache.parquet.column.ParquetProperties;
 
 import javax.inject.Inject;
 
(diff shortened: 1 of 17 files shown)


nmahadevuni:
@yingsu00 Can you please take a look?

yingsu00 (approved):
@nmahadevuni Thanks for the nice work. It'll be best if the release note can include an example so the users know exactly how the session property should be configured. Other than that it's just a couple of nits.

yingsu00:
@nmahadevuni there is an easyCLA issue.

nmahadevuni:
> @nmahadevuni there is an easyCLA issue.

That's for the co-author who is not registered in LF easyCLA. There is no way to fix that as my previous issue created with Linux foundation about it was closed saying that its expected and there is no way to remove that.

yingsu00:
@nmahadevuni Make yourself as the main author. If that doesn't work, remove the co-author information and add a line in the commit message "co-authored by ..."

nmahadevuni:
@zhenxiao @shangxinli Can you please review this PR?

mbasmanova:
This change seems to have collided with another change and together they are causing compilation failures.

See https://github.com/prestodb/presto/pull/19983#issuecomment-1748748461

tdcmeehan:
I raised #21043 to fix the build

majetideepak:
@nmahadevuni, @yingsu00 I see from the Trino commit in the https://github.com/trinodb/trino/pull/9611 description that Trino switched the default to PARQUET_1_0. Is there a reason why Presto chose PARQUET_2_0?
The reason being Parquet V2 spec has not been finalized.
https://github.com/trinodb/trino/issues/7953#issuecomment-872544273

yingsu00:
This PR merely added a config property that allows the user to choose v1 or v2. Before this change, Presto always write V2 and user cannot change it. To avoid disruption to existing users we didn't change the default version.

Issue #20910, opened 2023-09-19
[Native] - `sum(distinct )` gives incorrect results


mknegi:
## Expected Behavior
SELECT max(orderstatus), COUNT(orderkey), sum(DISTINCT orderkey) FROM orders

## Current Behavior
```
java.lang.AssertionError: For query: 
 SELECT max(orderstatus), COUNT(orderkey), sum(DISTINCT orderkey) FROM orders
not equal
Actual rows (up to 100 of 1 extra rows shown, 1 rows in total):
    [P, 15000, null]
Expected rows (up to 100 of 1 missing rows shown, 1 rows in total):
    [P, 15000, 449872500]
```

Property optimizer.optimize-mixed-distinct-aggregations was enabled in the presto to run this query.

aditi-pandit:
The issue here is in the GroupIdNode implementation.

Simplifying the query to 
` select COUNT(orderkey), count(distinct orderkey) from ( values 1, 5, 10) as t(orderkey);`

EXPLAIN plan for this query is
```
Fragment 0 [SINGLE]                                                                                                                                         
     Output layout: [count, count_3]                                                                                                                         
     Output partitioning: SINGLE []                                                                                                                          
     Stage Execution Strategy: UNGROUPED_EXECUTION                                                                                                           
     - Output[PlanNodeId 11][_col0, _col1] => [count:bigint, count_3:bigint]                                                                                 
             _col0 := count (1:35)                                                                                                                           
             _col1 := count_3 (1:52)                                                                                                                         
         - Project[PlanNodeId 211][projectLocality = LOCAL] => [count_3:bigint, count:bigint]                                                                
                 count := COALESCE(expr_17, BIGINT'0') (1:35)                                                                                                
             - LocalExchange[PlanNodeId 287][ROUND_ROBIN] () => [count_3:bigint, expr_17:bigint]                                                             
                 - Aggregate(FINAL)[PlanNodeId 210] => [count_3:bigint, expr_17:bigint]                                                                      
                         count_3 := "presto.default.count"((count_18)) (1:52)                                                                                
                         expr_17 := "presto.default.arbitrary"((arbitrary)) (1:35)                                                                           
                     - LocalExchange[PlanNodeId 303][SINGLE] () => [count_18:bigint, arbitrary:bigint]                                                       
                         - Aggregate(PARTIAL)[PlanNodeId 301] => [count_18:bigint, arbitrary:bigint]                                                         
[...]

PR #20964, opened 2023-09-26
[native] Fix GROUPING SETS with multiple aliased columns

aditi-pandit:
GROUPING SETS can be specified with multiple alias columns of the same input column. This is typically used to compute multiple mixed aggregations on the same key.

e.g. in queries
`select COUNT(orderkey), count(distinct orderkey) from orders;`

or

`SELECT lna, lnb, SUM(quantity) FROM (SELECT linenumber lna, linenumber lnb, CAST(quantity AS BIGINT) quantity FROM lineitem) GROUP BY GROUPING SETS ((lna, lnb), (lna), (lnb), ()) 
`

Presto built plan fragments correctly specifying the output column name for such cases. But the Velox plan conversion always mapped to the input column losing the distinction of the output group keys. This change retains the output column name thus disambiguating aliased columns.

Fixes https://github.com/prestodb/presto/issues/20910

```
== NO RELEASE NOTE ==
```


mbasmanova (approved):
@aditi-pandit Thank you, Aditi.

PR #20969, opened 2023-09-26
[native-pos] Disable AsyncDataCache (thus prefetching)

Yuhta:
We see negative performance impact (5% ~ 30%) when split prefetch is enabled,
hence adding these config options to disable it.


presto-spark-base/src/main/java/com/facebook/presto/spark/execution/property/NativeExecutionSystemConfig.java
```diff
@@ -323,6 +325,18 @@ public int getNumIoThreads()
...
     }
 
+    @Config(NUM_CONNECTOR_IO_THREADS)
+    public NativeExecutionSystemConfig setNumConnectorIoThreads(int numConnectorIoThreads)
+    {
+        this.numConnectorIoThreads = numConnectorIoThreads;
+        return this;
+    }
+
+    public int getNumConnectorIoThreads()
```
shrinidhijoshi, line 335: Also add a call to this function in the `getAllProperties` function on L127.

shrinidhijoshi:
@Yuhta Please update the commit message (and PR description) with brief explanation of why this change is needed.

shrinidhijoshi (approved):
We tried to do a test offline but the internal tooling was slowing us down. As this is non-breaking and very specific to cpp process, I am approving this. We can come back and fix if something is missing

mbasmanova (commented):
@Yuhta Use [native-pos] prefix for Presto-on-Spark specific changes.

mbasmanova (commented):
@Yuhta What is the definition of num-connector-io-threads property? Is it documented already? If not, would you document it?

mbasmanova (commented):
> We see negative performance impact (5% ~ 30%) when split prefetch is enabled,

What is causing that?

mbasmanova (commented):
@Yuhta Since you have [native-pos] prefix, adding "for Presto on Spark" is redundant. Let's remove from both PR title and commit message.

presto-spark-base/src/main/java/com/facebook/presto/spark/execution/property/NativeExecutionSystemConfig.java
```diff
@@ -59,6 +59,9 @@ public class NativeExecutionSystemConfig
     // TODO: others use "-" separator and this property use _ separator. Fix them.
     private static final String HTTP_EXEC_THREADS = "http_exec_threads";
     private static final String NUM_IO_THREADS = "num-io-threads";
+    private static final String ASYNC_DATA_CACHE_ENABLED = "async-data-cache-enabled";
```
mbasmanova, line 62: Do we have these properties documented? If not, please, document them.

mbasmanova (commented):
> We see negative performance impact (5% ~ 30%) when split prefetch is enabled,

@Yuhta What's the cause of performance impact?

presto-spark-base/src/main/java/com/facebook/presto/spark/execution/property/NativeExecutionSystemConfig.java
Yuhta: Documented in where they are defined

Yuhta: Where should I put it?

presto-docs/src/main/sphinx/develop/presto-native.rst
```diff
@@ -126,4 +126,34 @@ There is also an additional parameter:
...
+--------------------------------
+
+``num-connector-io-threads``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+    * **Type** ``integer``
+    * **Default value:** ``30``
+
+    Size of IO executor for connectors to do preload/prefetch.  Prefetch is
+    disabled if set to zero.
```
steveburnett, line 141:
```suggestion
    disabled if ``num-connector-io-threads`` is set to zero.
```
Suggest explicitly stating what to set to 0 to disable prefetch.

presto-docs/src/main/sphinx/develop/presto-native.rst
```diff
@@ -126,4 +126,34 @@ There is also an additional parameter:
...
+
+``async-data-cache-enabled``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+    * **Type** ``bool``
+    * **Default value:** ``true``
+
+    Whether async data cache is enabled.  Disable this also disable split
```
steveburnett, line 149:
```suggestion
    Whether async data cache is enabled.  Setting ``async-data-cache-enabled`` to ``false`` disables split
```
Suggesting specifying what is done and its result.

presto-docs/src/main/sphinx/develop/presto-native.rst
```diff
@@ -126,4 +126,34 @@ There is also an additional parameter:
...
+    prefetching in table scan.
+
+``async-cache-ssd-gb``
+^^^^^^^^^^^^^^^^^^^^^^
+
+    * **Type** ``integer``
+    * **Default value:** ``0``
+
+    Size of the SSD cache when async data cache is enabled.  Must be zero if
+    async data cache is disabled.
```
steveburnett, line 159:
```suggestion
    ``async-data-cache-enabled`` is disabled.
```

steveburnett (changes requested):
Thanks for the documentation! I added a few suggestions to make some statements more explicit and easier to understand. If I have made a suggestion that changes the meaning in a way you disagree with, please help me understand the meaning.

steveburnett (approved):
LGTM! (docs)

presto-docs/src/main/sphinx/develop/presto-native.rst
```diff
@@ -126,4 +126,34 @@ There is also an additional parameter:
...
+    * **Default value:** ``true``
+
+    Whether async data cache is enabled.  Setting ``async-data-cache-enabled``
+    to ``false`` disables split prefetching in table scan.
+
+``async-cache-ssd-gb``
+^^^^^^^^^^^^^^^^^^^^^^
+
+    * **Type** ``integer``
+    * **Default value:** ``0``
```
mbasmanova, line 156: Are we missing some other configs re: SSD. I assume one needs to specify rott directory or something, no?

presto-docs/src/main/sphinx/develop/presto-native.rst
```diff
@@ -126,4 +126,34 @@ There is also an additional parameter:
...
+``async-data-cache-enabled``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+    * **Type** ``bool``
+    * **Default value:** ``true``
+
+    Whether async data cache is enabled.  Setting ``async-data-cache-enabled``
+    to ``false`` disables split prefetching in table scan.
+
+``async-cache-ssd-gb``
```
mbasmanova, line 152: consistency: async-data-cache-enabled vs. async-cache-ssd-gb - one says data-cache, the other just cache. Would be nice to fix.

presto-docs/src/main/sphinx/develop/presto-native.rst
```diff
@@ -126,4 +126,34 @@ There is also an additional parameter:
...
     is valid.
     If the time period exceeds the parameter value, the request is rejected as
-    authentication failure (HTTP 401).
\ No newline at end of file
+    authentication failure (HTTP 401).
+
+Async Data Cache and Prefetching
+--------------------------------
+
+``num-connector-io-threads``
```
mbasmanova, line 134: Is this specific to Hive connector?

presto-spark-base/src/main/java/com/facebook/presto/spark/execution/property/NativeExecutionSystemConfig.java
```diff
@@ -104,6 +107,9 @@ public class NativeExecutionSystemConfig
     private String httpsCertPath = "";
     private String httpsKeyPath = "";
     private int numIoThreads = 30;
+    private boolean asyncDataCacheEnabled;
```
mbasmanova, line 110: Should these have default values? true, 0 and 30 (according to documentation).

presto-docs/src/main/sphinx/develop/presto-native.rst
Yuhta: That will be a big change, rather not changing it here

mbasmanova: Understood. Perhaps, file a GitHub issue about this.

presto-spark-base/src/main/java/com/facebook/presto/spark/execution/property/NativeExecutionSystemConfig.java
Yuhta: We set it to false here to disable it.  The default values are for when it is not overwritten here.

mbasmanova: Would you explicitly set it or make a comment that default is 'false'? 

```
private boolean asyncDataCacheEnabled = false;
```

mbasmanova: > The default values are for when it is not overwritten here.

I'm not sure I understand.... The user reading the documentation would assume that default is what they get without changing anything. It looks like actual default is false, no?

Yuhta: Setting it explicitly violate a lint rule.  I will put a comment.

The default is you don't overwrite it in application code.  Here POS is overwriting it explicitly so the defaults no longer applies.  Do you have a place I can write POS defaults?

presto-docs/src/main/sphinx/develop/presto-native.rst
Yuhta: Just you will need to set SSD size in another property to indicate how much space is available

presto-docs/src/main/sphinx/develop/presto-native.rst
Yuhta: #21185

mbasmanova (approved):
Accepting to unblock, but I'm very confused by how these properties are set and what the defaults are. It would be nice to clear this up in a follow up.

aditi-pandit:
@Yuhta , @mbasmanova : We noticed this negative performance impact internally for Prestissimo as well. Did you'll only see it for POS, but not Prestissimo ? (Just curious as you are only turning off POS configs)

We should turn it off in https://github.com/prestodb/presto/blob/master/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/NativeQueryRunnerUtils.java#L33 config as well. wdyt ?

Yuhta:
@aditi-pandit We have different deployments of Prestissimo and some do turn the option off.  You can use the worker config file to overwrite the system config.  POS needs special treatment because the configuration is a little bit weird and worker config does not work out of box there.

PR #20970, opened 2023-09-27
Correct the counting for resource group's running tasks

hantangwangd:
## Description

Fix the counting for resource group's running tasks. Previously it will always be 0, and that would cause resource group manager's "max-total-running-task-count-to-not-execute-new-query" not working.

## Test Plan

. Renovate test case TestQueryTaskLimit.testQueuingWhenTaskLimitExceeds()

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/execution/resourceGroups/InternalResourceGroup.java
+++ b/presto-main/src/main/java/com/facebook/presto/execution/resourceGroups/InternalResourceGroup.java
@@ -15,7 +15,6 @@ package com.facebook.presto.execution.resourceGroups;
 
 import com.facebook.airlift.stats.CounterStat;
 import com.facebook.presto.execution.ManagedQueryExecution;
-import com.facebook.presto.execution.SqlQueryExecution;
 import com.facebook.presto.execution.resourceGroups.WeightedFairQueue.Usage;
 import com.facebook.presto.metadata.InternalNodeManager;
 import com.facebook.presto.server.QueryStateInfo;
@@ -687,8 +686,7 @@ public class InternalResourceGroup
     {
         if (subGroups().isEmpty()) {
             return runningQueries.stream()
-                    .filter(SqlQueryExecution.class::isInstance)
-                    .mapToInt(query -> ((SqlQueryExecution) query).getRunningTaskCount())
+                    .mapToInt(query -> query.getRunningTaskCount())
                     .sum();
         }
 
--- a/presto-tests/src/test/java/com/facebook/presto/tests/TestQueryTaskLimit.java
+++ b/presto-tests/src/test/java/com/facebook/presto/tests/TestQueryTaskLimit.java
@@ -28,6 +28,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 
+import static com.facebook.presto.execution.QueryState.FAILED;
 import static com.facebook.presto.execution.QueryState.QUEUED;
 import static com.facebook.presto.execution.QueryState.RUNNING;
 import static com.facebook.presto.execution.TestQueryRunnerUtil.createQuery;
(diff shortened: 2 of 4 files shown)


presto-tests/src/test/java/com/facebook/presto/tests/TestQueryTaskLimit.java
```diff
@@ -90,16 +90,31 @@ public void testExceedTaskLimit()
...
+                    LONG_LASTING_QUERY);
             waitForQueryState(queryRunner, firstQuery, RUNNING);
 
-            queryRunner.getCoordinator().getResourceGroupManager().get().setTaskLimitExceeded(true);
+            // wait for a while, the first query will schedule at lease 2 tasks, so exceed resource group manager's total task limit
+            try {
+                Thread.sleep(500);
+            }
+            catch (Exception e) {
+                // do noting
```
tdcmeehan, line 108:
```suggestion
                // do nothing
```

presto-tests/src/test/java/com/facebook/presto/tests/TestQueryTaskLimit.java
```diff
@@ -90,16 +90,31 @@ public void testExceedTaskLimit()
...
+
+        try (DistributedQueryRunner queryRunner = createQueryRunner(defaultSession, extraProperties)) {
+            QueryId firstQuery = createQuery(queryRunner, newSession("test", ImmutableSet.of(), null),
+                    LONG_LASTING_QUERY);
             waitForQueryState(queryRunner, firstQuery, RUNNING);
 
-            queryRunner.getCoordinator().getResourceGroupManager().get().setTaskLimitExceeded(true);
+            // wait for a while, the first query will schedule at lease 2 tasks, so exceed resource group manager's total task limit
+            try {
+                Thread.sleep(500);
```
tdcmeehan, line 105: Can you think of a way to check for the query's state?  Because of GC, or other environmental issues, 500 millis may not consistently give the same desired result.

presto-main/src/main/java/com/facebook/presto/execution/resourceGroups/InternalResourceGroup.java
```diff
@@ -687,8 +687,8 @@ public int getRunningTaskCount()
     {
         if (subGroups().isEmpty()) {
             return runningQueries.stream()
-                    .filter(SqlQueryExecution.class::isInstance)
-                    .mapToInt(query -> ((SqlQueryExecution) query).getRunningTaskCount())
+                    .filter(LocalDispatchQuery.class::isInstance)
```
tdcmeehan, line 690: Just so I understand, the core of this bug is that `SqlQueryExecution.getRunningTaskCount()` always returns 0?

presto-tests/src/test/java/com/facebook/presto/tests/TestQueryTaskLimit.java
```diff
@@ -90,16 +90,31 @@ public void testExceedTaskLimit()
...
+            }
 
             QueryId secondQuery = createQuery(queryRunner, newSession("test", ImmutableSet.of(), null), LONG_LASTING_QUERY);
+
+            // When current running tasks exceeded limit, the following query would be queued
             waitForQueryState(queryRunner, secondQuery, QUEUED);
 
-            queryRunner.getCoordinator().getResourceGroupManager().get().setTaskLimitExceeded(false);
+            // When first query is cancelled, the second query would be pass to run
+            queryRunner.getCoordinator().getDispatchManager().cancelQuery(firstQuery);
```
tdcmeehan, line 117: Wait for cancellation of first query before proceeding to check next query?

presto-main/src/main/java/com/facebook/presto/execution/resourceGroups/InternalResourceGroup.java
hantangwangd: The core reason is, query instances in InternalResourceGroup.runningQueries is of type LocalDispatchQuery, not of type SqlQueryExecution, so the filter will filter out all running query instances.

presto-tests/src/test/java/com/facebook/presto/tests/TestQueryTaskLimit.java
hantangwangd: It's a great guidance, I will think about how to achieve it.

presto-tests/src/test/java/com/facebook/presto/tests/TestQueryTaskLimit.java
hantangwangd: The logic to check next query is a while loop to wait for desired state. As the first query did not finish it's cancellation, the second query would continuously in QUEUED state, so function waitForQueryState(queryRunner, secondQuery, RUNNING) would continuously loop util the first query finish it's cancellation and the second query start to run. So, here we can call waitForQueryState(queryRunner, secondQuery, RUNNING) immediately, and no need to wait.

presto-main/src/main/java/com/facebook/presto/execution/resourceGroups/InternalResourceGroup.java
tdcmeehan: Oh, good catch!

presto-tests/src/test/java/com/facebook/presto/tests/TestQueryTaskLimit.java
tdcmeehan: Yes, implicitly it's the same, but it would be nice to check that the query is cancelled explicitly (rather than timed out for example).

presto-main/src/main/java/com/facebook/presto/execution/resourceGroups/InternalResourceGroup.java
tdcmeehan: Let's add `getRunningTaskCount` to `ManagedQueryExecution` interface and check instances of the interface.

presto-tests/src/test/java/com/facebook/presto/tests/TestQueryTaskLimit.java
hantangwangd: OK, I get your point. You are right, It's better to check explicitly the first query is cancelled. I will fix it.

hantangwangd:
@tdcmeehan I have fixed all your suggestions above, and squashed the commits. Would you please take a look at it? Thanks a lot.

presto-tests/src/test/java/com/facebook/presto/tests/TestQueryTaskLimit.java
```diff
@@ -135,4 +149,32 @@ private void waitForQueryToBeKilled(DistributedQueryRunner queryRunner)
             MILLISECONDS.sleep(10);
         }
     }
+
+    private void waitForQueryToBeCancelled(DistributedQueryRunner queryRunner, QueryId queryId)
```
tdcmeehan, line 153: Use `TestQueryRunnerUtil.waitForQuerState`?

hantangwangd: TestQueryRunnerUtil.waitForQueryState() can only check the query's state has changed to FAILED. Is it enough to make sure that the query is actually been cancelled rather than timed out?

tdcmeehan: I think that's fine.

hantangwangd: OK, I have fixed it.

tdcmeehan:
CC: @swapsmagic @MnO2 @abhiseksaikia @agrawaldevesh FYI

tdcmeehan (approved):

PR #20973, opened 2023-09-27
[native] Add retry for abort results

pranjalssh:
## Description
* If abortResults call fails, it leads to query failure. So we should retry them. We cannot simply fire and forget them as query may get stuck otherwise. Added tests for the same.

* Refactored tests to separate out case when producer always fails.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/PrestoExchangeSource.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoExchangeSource.cpp
@@ -140,11 +140,11 @@ folly::SemiFuture<PrestoExchangeSource::Response> PrestoExchangeSource::request(
 
   promise_ = std::move(promise);
   failedAttempts_ = 0;
-  retryState_ =
+  dataRequestRetryState_ =
       RetryState(std::chrono::duration_cast<std::chrono::milliseconds>(
                      SystemConfig::instance()->exchangeMaxErrorDuration())
                      .count());
-  doRequest(retryState_.nextDelayMs(), maxBytes, maxWaitSeconds);
+  doRequest(dataRequestRetryState_.nextDelayMs(), maxBytes, maxWaitSeconds);
 
   return future;
 }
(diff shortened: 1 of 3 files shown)


presto-native-execution/presto_cpp/main/PrestoExchangeSource.h
```diff
@@ -212,15 +214,15 @@ class PrestoExchangeSource : public velox::exec::ExchangeSource {
   const std::shared_ptr<folly::IOThreadPoolExecutor> exchangeExecutor_;
 
   std::shared_ptr<http::HttpClient> httpClient_;
-  RetryState retryState_;
+  RetryState requestRetryState_;
```
xiaoxmeng, line 217:
```
s/requestRetryState_/dataRetryState_/?
```

presto-native-execution/presto_cpp/main/PrestoExchangeSource.cpp
```diff
@@ -380,44 +380,59 @@ void PrestoExchangeSource::acknowledgeResults(int64_t ackSequence) {
 }
 
 void PrestoExchangeSource::abortResults() {
+  bool abortIssued = false;
+  if (!abortResultsIssued_.compare_exchange_strong(abortIssued, true)) {
```
xiaoxmeng, line 384:
```
if (abortResultsIssued_.exchange(true)) {
  return;
}
```

xiaoxmeng (approved):
@pranjalssh LGTM. Please shadow this in presto batch or deltoid a bit before land. Thanks!

presto-native-execution/presto_cpp/main/PrestoExchangeSource.cpp
```diff
@@ -380,44 +380,59 @@ void PrestoExchangeSource::acknowledgeResults(int64_t ackSequence) {
...
           const std::string errMsg = fmt::format(
-              "Abort results failed: {}, path {}", statusCode, self->basePath_);
+              "Abort results failed: {}, path {}",
+              error.value(),
+              self->basePath_);
           LOG(ERROR) << errMsg;
-          onFinalFailure(errMsg, queue);
-        } else {
-          self->abortResultsSucceeded_.store(true);
+          return onFinalFailure(errMsg, queue);
```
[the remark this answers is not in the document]
pranjalssh: We need to call queue->setError() so we need it

mbasmanova (commented):
@pranjalssh Would you update commit message and PR description to explain why it is necessary to retry abort API calls? Why can't we use fire-and-forget semantics for these?

PR #20974, opened 2023-09-27
Fix prune output and local exchange optimizer for HBO

feilong-liu:
## Description
Fix two bugs for HBO 1) update the prune unreferenced output optimizer to include the stats equivalent node when creating new plan nodes 2) change the add local exchange optimizer to not propagate the stats equivalent node if the original node is table writer and a new table writer merger node is added.

## Motivation and Context
HBO relies on statsEquivalentPlanNode to hash and match history nodes. However, in the prune unreferenced output optimizer, the statsEquivalentPlanNode is dropped when creating new plan nodes. This will lead to miss of plan hash and failed to record/use the history statistics.

Another bug is in the add local exchange optimizer. For table writer node, it will add a new table writer merge node above. However, these two nodes are not stats equivalent, hence cannot reuse the stats equivalent node.

## Impact
Improve the coverage of HBO stats

## Test Plan
Bug fix

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/AddLocalExchanges.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/AddLocalExchanges.java
@@ -564,6 +564,7 @@ public class AddLocalExchanges
                             new TableWriterNode(
                                     originalTableWriterNode.getSourceLocation(),
                                     originalTableWriterNode.getId(),
+                                    originalTableWriterNode.getStatsEquivalentPlanNode(),
                                     originalTableWriterNode.getSource(),
                                     originalTableWriterNode.getTarget(),
                                     variableAllocator.newVariable("partialrowcount", BIGINT),
@@ -587,6 +588,7 @@ public class AddLocalExchanges
                             new TableWriterNode(
                                     originalTableWriterNode.getSourceLocation(),
                                     originalTableWriterNode.getId(),
+                                    originalTableWriterNode.getStatsEquivalentPlanNode(),
                                     exchange.getNode(),
                                     originalTableWriterNode.getTarget(),
                                     variableAllocator.newVariable("partialrowcount", BIGINT),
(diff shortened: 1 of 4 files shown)


presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/AddLocalExchanges.java
```diff
@@ -923,8 +923,10 @@ private PlanWithProperties deriveProperties(PlanNode result, List<StreamProperti
         private PlanWithProperties accept(PlanNode node, StreamPreferredProperties context)
         {
             PlanWithProperties result = node.accept(this, context);
+            // TableWriter and TableWriterMergeNode has different output
+            boolean passStatsEquivalentPlanNode = !(node instanceof TableWriterNode && result.getNode() instanceof TableWriterMergeNode);
             return new PlanWithProperties(
-                    result.getNode().assignStatsEquivalentPlanNode(node.getStatsEquivalentPlanNode()),
+                    passStatsEquivalentPlanNode ? result.getNode().assignStatsEquivalentPlanNode(node.getStatsEquivalentPlanNode()) : result.getNode(),
```
mlyublena, line 929: do we expect to see more false positives because of this? that is, plans that shouldn't match now match

feilong-liu: This is the only case I am aware of now, in which the result node has different number of rows compared with the original node.

mlyublena (approved):

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/AddLocalExchanges.java
pranjalssh: Can we do this when creating TableWriterMergeNode? Can we add a test for this?

feilong-liu: I tried but it seems non-trivial to change the logic when creating the TableWriterMergeNode. The accept method is the default one always called.

Added a test for it to make sure that no stats equivalent node is propagated to TableWriterMergeNode

pranjalssh (approved):

jainxrohit (approved):
LGTM

PR #20977, opened 2023-09-28
Fix bug in maxExecutors config tuning in AQE codepath

pgupta2:
maxExecutors config tuning currently only works for non-AQE execution path.
Moving it at common place so that it works for both AQE and non-AQE 
execution mode

```
== NO RELEASE NOTE ==
```


vermapratyush (approved):
LGTM

singcha (approved):

NikhilCollooru (approved):

PR #20996, opened 2023-09-29
Template for security Vulnerability Request

Akanksha-kedia:
This template is suitable as it provides clear instructions for users about which version of the project are currently supported and how to report security vulnerabilities.

### DIFF
--- a/.github/ISSUE_TEMPLATE/SECURITY.md
+++ b/.github/ISSUE_TEMPLATE/SECURITY.md
@@ -1,0 +1,10 @@
+# Security Policy
+
+## Supported Versions
+
+The current release is the supported version. Security fixes are released together
+with all other fixes in each new release.
+
+## Reporting a Vulnerability
+
+If you discover a security vulnerability, please report to **security@presto.io**.


tdcmeehan:
@Akanksha-kedia I think the issue is different.  In the example in the PR, it seems unexpected that both don't fail.  Could you check the behavior of other engines, in particular Postgres?

presto-docs/src/main/sphinx/functions/array.rst
```diff
@@ -158,6 +158,12 @@ Array Functions
                                        -1,
                                        IF(cardinality(x) = cardinality(y), 0, 1))); -- [[1, 2], [2, 3, 1], [4, 2, 1, 4]]
 
+        SELECT array_sort(ARRAY[ARRAY[1, null], ARRAY[1, 2]]);
+
+        Note : Presto does not support comparison of nested arrays that contain null elements.
```
steveburnett, line 163: Suggest formatting the indentation of the Note so that the Note is text and not part of the code block, but is still included in the entry for "array_sort(array(T), function(T, T, int)) -> array(T)". 

Screenshot of local build shows how the Note's inclusion in the code block causes highlighting and emphasis that may not be your intention. 

Screenshot 2023-09-29 at 11 12 40 AM

steveburnett (changes requested):
One comment about the Note being inside in the code block above it causing the text to be highlighted in ways you might not have intended. Suggest make the Note text, indenting it like the first paragraph (the one beginning with "Sorts and returns the ``array``" in the entry "array_sort(array(T), function(T, T, int)) -> array(T)"

Akanksha-kedia:
> @Akanksha-kedia I think the issue is different. In the example in the PR, it seems unexpected that both don't fail. Could you check the behavior of other engines, in particular Postgres?

@tdcmeehan  i checked trino, its behaving as Prestodb is behaving. We need either  to document or make the necessary fix as per issue raised.

tdcmeehan:
Trino is not a good engine to check, as we both might have inherited the same bug.

I see Postgres validates the arrays to check if they have nulls

```sql
select sort(array[array[1, null], array[2, null]]);
```

```
ERROR: array must not contain nulls
```

Seems to be the same thing Velox wants to do.

Akanksha-kedia:
> Trino is not a good engine to check, as we both might have inherited the same bug.
> 
> I see Postgres validates the arrays to check if they have nulls
> 
> ```sql
> select sort(array[array[1, null], array[2, null]]);
> ```
> 
> ```
> ERROR: array must not contain nulls
> ```
> 
> Seems to be the same thing Velox wants to do.

What next with this ?Any insights?

tdcmeehan:
@Akanksha-kedia it looks like you've repurposed this old PR into a new PR for security vulnerabilities.  To prevent confusion, please raise a new PR with the template, as there are links to the old issue (questions around `array_sort` semantics).

PR #21007, opened 2023-09-29
[AutoAccept][Codemod][ThriftFormatFBSource] fbcode/github/presto-trunk/presto-docs/src/main/sphinx/include

aristidisp:
Reviewed By: aristidisp

Differential Revision: D49552047


steveburnett (changes requested):
local doc build fails with 

"Running Sphinx v4.4.0

Sphinx version error:
The sphinxcontrib.applehelp extension used by this project needs at least Sphinx v5.0; it therefore cannot be built with this version.
make: *** [html] Error 2"

Related to PR #21708.

PR #21011, opened 2023-10-01
[native] Fix Timestamp construction for negative timestamps

lingbin:
In Velox, `Timestamp::seconds_` is the starting point of one second,
and `Timestamp::nanos_` is the offset from the starting point (in
nanoseconds). So when converting milliseconds to seconds, we should
always round down, including negative values. This means that it is
wrong to use division and modulo directly. For example, for `-400ms`,
it should be converted into `(-1, 6 * 10^8)` instead of
`(0, -4 * 10^8)`.

`velox::Timestamp::fromMillis()` utility function can be used to
achieve this.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/presto_protocol/Base64Util.cpp
+++ b/presto-native-execution/presto_cpp/presto_protocol/Base64Util.cpp
@@ -137,8 +137,7 @@ velox::VectorPtr readScalarBlock(
           velox::AlignedBuffer::allocate<velox::Timestamp>(positionCount, pool);
       auto* rawTimestamps = timestamps->asMutable<velox::Timestamp>();
       for (auto i = 0; i < positionCount; i++) {
-        rawTimestamps[i] = velox::Timestamp(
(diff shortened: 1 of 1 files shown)


mbasmanova (approved):
Thanks.

zacw7:
Thank you for the quick fix and the very detailed commit message. Could you please shorten the commit message title a bit so that the `...` won't show up at the end before we merge this change? It's advised to follow https://cbea.ms/git-commit/.

lingbin:
@zacw7  Thanks for the guidance. I have modified the title so that `...` will no longer appear, please take a look again.

zacw7:
> @zacw7 Thanks for the guidance. I have modified the title so that `...` will no longer appear, please take a look again.

Could you please edit the commit message as well? Thanks!

lingbin:
@zacw7 The commit message has been guaranteed to be no more than 72 characters before line wrapping. Have I missed any other format constraints on the commit message?

zacw7 (approved):

PR #21015, opened 2023-10-02
[native]Improve spill directory failure handling

xiaoxmeng:
There are three problems found in Meta internal arbitration and spilling test about
task spilling directory setup:
(1) velox task creation failure handle path doesn't clear velox task before allowing
create the next velox task which can lead to the second velox task creation (retry)
fail with unrelated errors;
(2) set spilling path before the spill path creation succeed. The better is to create the
spilling directory and then set it task to avoid another exception on task destruction
which removes a non-exist spill directory/

This PR fixes this issue with unit test. Note the spill directory setup could fail because
of transient remote storage error as we found in Meta internal testing.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/TaskManager.cpp
+++ b/presto-native-execution/presto_cpp/main/TaskManager.cpp
@@ -52,7 +52,6 @@ static void maybeSetupTaskSpillDirectory(
   const auto includeNodeInSpillPath =
       SystemConfig::instance()->includeNodeInSpillPath();
   auto nodeConfig = NodeConfig::instance();
-
   const auto taskSpillDirPath = TaskManager::buildTaskSpillDirectoryPath(
       baseSpillDirectory,
       nodeConfig->nodeInternalAddress(),
@@ -60,12 +59,12 @@ static void maybeSetupTaskSpillDirectory(
       execTask.queryCtx()->queryId(),
       execTask.taskId(),
       includeNodeInSpillPath);
-  execTask.setSpillDirectory(taskSpillDirPath);
   // Create folder for the task spilling.
   auto fileSystem =
       velox::filesystems::getFileSystem(taskSpillDirPath, nullptr);
   VELOX_CHECK_NOT_NULL(fileSystem, "File System is null!");
   fileSystem->mkdir(taskSpillDirPath);
+  execTask.setSpillDirectory(taskSpillDirPath);
 }
 
 // Keep outstanding Promises in RequestHandler's state itself.
@@ -436,20 +435,28 @@ std::unique_ptr<TaskInfo> TaskManager::createOrUpdateTask(
         return std::make_unique<TaskInfo>(prestoTask->updateInfoLocked());
       }
 
-      execTask = exec::Task::create(
+      // Uses a temp variable to store the created velox task to destroy it
+      // under presto task lock if spill directory setup fails. Otherwise, the
+      // concurrent task creation retry from the coordinator might see the
+      // unexpected state in presto task left by the previously failed velox
+      // task which hasn't been destroyed yet, such as the task pool in query's
+      // root memory pool.
+      auto newExecTask = exec::Task::create(
           taskId, planFragment, prestoTask->id.id(), std::move(queryCtx));
-      auto baseSpillDir = *(baseSpillDir_.rlock());
-      maybeSetupTaskSpillDirectory(planFragment, *execTask, baseSpillDir);
+      // TODO: move spill directory creation inside velox task execution
+      // whenever spilling is triggered. It will reduce the unnecessary file
+      // operations on remote storage.
+      const auto baseSpillDir = *(baseSpillDir_.rlock());
+      maybeSetupTaskSpillDirectory(planFragment, *newExecTask, baseSpillDir);
 
-      prestoTask->task = execTask;
+      prestoTask->task = std::move(newExecTask);
       prestoTask->info.needsPlan = false;
       startTask = true;
-    } else {
-      execTask = prestoTask->task;
     }
+    execTask = prestoTask->task;
   }
   // Outside of prestoTask->mutex.
-  VELOX_CHECK(
+  VELOX_CHECK_NOT_NULL(
       execTask,
       "Task update received before setting a plan. The splits in "
       "this update could not be delivered for {}",
--- a/presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp
+++ b/presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp
@@ -134,6 +134,12 @@ class Cursor {
   uint64_t sequence_ = 0;
 };
 
+void setAggregationSpillConfig(
+    std::map<std::string, std::string>& queryConfigs) {
+  queryConfigs.emplace(core::QueryConfig::kSpillEnabled, "true");
+  queryConfigs.emplace(core::QueryConfig::kAggregationSpillEnabled, "true");
+}
+
 static const uint64_t kGB = 1024 * 1024 * 1024ULL;
 
 class TaskManagerTest : public testing::Test {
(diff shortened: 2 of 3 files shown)


presto-native-execution/presto_cpp/main/TaskManager.cpp
```diff
@@ -42,30 +42,37 @@ namespace {
...
+        velox::filesystems::getFileSystem(taskSpillDirPath, nullptr);
+    VELOX_CHECK_NOT_NULL(fileSystem, "File System is null!");
+    fileSystem->mkdir(taskSpillDirPath);
+  } catch (...) {
+    LOG(WARNING) << "Failed to create spill directory for velox task "
+                 << veloxTask.taskId() << ". Error: "
+                 << folly::exceptionStr(std::current_exception());
+    return;
+  }
+  veloxTask.setSpillDirectory(taskSpillDirPath);
```
tanjialiang, line 75: Thanks, curious what will happen when path creation fails and we still set the path.

tanjialiang (approved):

presto-native-execution/presto_cpp/main/TaskManager.cpp
xiaoxmeng: It will get another exception on rmdir and might distract the debug but the query fails so it won't get executed.

presto-native-execution/presto_cpp/main/TaskManager.cpp
```diff
@@ -42,30 +42,37 @@ namespace {
...
+    // Create folder for the task spilling.
+    auto fileSystem =
+        velox::filesystems::getFileSystem(taskSpillDirPath, nullptr);
+    VELOX_CHECK_NOT_NULL(fileSystem, "File System is null!");
+    fileSystem->mkdir(taskSpillDirPath);
+  } catch (...) {
+    LOG(WARNING) << "Failed to create spill directory for velox task "
+                 << veloxTask.taskId() << ". Error: "
+                 << folly::exceptionStr(std::current_exception());
+    return;
```
mbasmanova, line 73: Why is it OK to ignore this error? If we fail to setup spill directory and the query runs out of memory, will we get some cryptic error after wasting a lot of compute?

presto-native-execution/presto_cpp/main/TaskManager.cpp
```diff
@@ -424,7 +431,7 @@ std::unique_ptr<TaskInfo> TaskManager::createOrUpdateTask(
     const protocol::OutputBuffers& outputBuffers,
     std::shared_ptr<velox::core::QueryCtx> queryCtx,
     long startProcessCpuTime) {
-  std::shared_ptr<exec::Task> execTask;
+  std::shared_ptr<exec::Task> veloxTask;
```
mbasmanova, line 434: Nice rename, but it would be better to do this in a separate PR. Mixing trivial refactoring with meaningful changes makes it hard to review the code.

presto-native-execution/presto_cpp/main/TaskManager.cpp
```diff
@@ -436,21 +443,24 @@ std::unique_ptr<TaskInfo> TaskManager::createOrUpdateTask(
         return std::make_unique<TaskInfo>(prestoTask->updateInfoLocked());
       }
 
-      execTask = exec::Task::create(
-          taskId, planFragment, prestoTask->id.id(), std::move(queryCtx));
-      auto baseSpillDir = *(baseSpillDir_.rlock());
-      maybeSetupTaskSpillDirectory(planFragment, *execTask, baseSpillDir);
+      {
+        auto newVeloxTask = exec::Task::create(
```
mbasmanova, line 447: Why do we need this variable?

presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp
```diff
@@ -1132,6 +1132,67 @@ TEST_F(TaskManagerTest, checkBatchSplits) {
...
+      const std::string badSpillPath = "/nonexist/presto_native/path";
+      taskManager_->setBaseSpillDirectory(badSpillPath);
+    }
+
+    std::map<std::string, std::string> queryConfigs;
+    queryConfigs.emplace(core::QueryConfig::kTestingSpillPct, "100");
+    queryConfigs.emplace(core::QueryConfig::kSpillEnabled, "true");
+    queryConfigs.emplace(core::QueryConfig::kAggregationSpillEnabled, "true");
+
+    auto planNodeIdGenerator = std::make_shared<core::PlanNodeIdGenerator>();
```
mbasmanova, line 1167: planNodeIdGenerator is not needed for plans without joins

presto-native-execution/presto_cpp/main/TaskManager.cpp
xiaoxmeng: We currently create spilling directory at the startup of each query so the query might not spill. So instead of failing query, we can proceed the query without spilling enabled. We could later add a thread-safe method at query level to create a spilling directory when needs if this happens too often in practice.

mbasmanova: I'm not sure I understand. What would cause a failure to setup spill directory? If this is cause by bad config, where base spill directory doesn't exist than it might be better to fail early, no?

presto-native-execution/presto_cpp/main/TaskManager.cpp
xiaoxmeng: This is to destroy the velox task under presto task lock so as to ensure the second presto task update call won't see the leftover state from the previous failed velox task such as the task memory pool under query root memory pool. Here is the event sequence:
T1: task manager receive request from coordinator to create velox task;
T2: task manager holds the presto task lock to create a new velox task;
T3: the spilling directory setup fails and throw, but the new velox task hans't destroyed yet;
T4: task manager releases the presto task lock and second create velox task request comes;
T5: task manager creates a new velox task for the same presto task but the memory pool of the previously created velox task still stays in the query root memory pool;
T6: the second velox task creation fails because of the duplicate task memory pool name.

presto-native-execution/presto_cpp/main/TaskManager.cpp
xiaoxmeng: Not bad config. From the failure that we have been seen in Meta internal testing, it is caused by some remote storage error (some internal timeout at storage backend). The spilling directory setup happens on the query start but not when we trigger spills.

presto-native-execution/presto_cpp/main/TaskManager.cpp
xiaoxmeng: I removed it as maybeSetupDirectory now tolerates the failure.

presto-native-execution/presto_cpp/main/TaskManager.cpp
xiaoxmeng: Reverted.

presto-native-execution/presto_cpp/main/TaskManager.cpp
```diff
@@ -52,20 +52,27 @@ static void maybeSetupTaskSpillDirectory(
...
       nodeConfig->nodeId(),
       execTask.queryCtx()->queryId(),
       execTask.taskId(),
       includeNodeInSpillPath);
+
+  try {
+    // Create folder for the task spilling.
+    auto fileSystem =
+        velox::filesystems::getFileSystem(taskSpillDirPath, nullptr);
+    VELOX_CHECK_NOT_NULL(fileSystem, "File System is null!");
```
mbasmanova, line 67: Do we want to catch and ignore this failure as well?

presto-native-execution/presto_cpp/main/TaskManager.cpp
```diff
@@ -52,20 +52,27 @@ static void maybeSetupTaskSpillDirectory(
...
+    fileSystem->mkdir(taskSpillDirPath);
```
mbasmanova, line 68: Maybe add a comment here to explain that this might be a transient failure and we want to process with task creation and execution hoping spilling won't be needed.

However, wouldn't it be confusing if queries fail with OOM errors while cluster is configured with spilling enabled?

presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp
```diff
@@ -1132,6 +1132,66 @@ TEST_F(TaskManagerTest, checkBatchSplits) {
   ASSERT_EQ(resultOrFailure.status, nullptr);
 }
 
+TEST_F(TaskManagerTest, spillDirectoryNonExist) {
```
mbasmanova, line 1135: nit: NotExist -> Missing

presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp
```diff
@@ -1132,6 +1132,66 @@ TEST_F(TaskManagerTest, checkBatchSplits) {
...
+  // Cleanup old tasks between test iterations.
+  taskManager_->setOldTaskCleanUpMs(0);
+  for (bool spillDirExist : {false, true}) {
+    SCOPED_TRACE(fmt::format("spillDirExist: {}", spillDirExist));
+    auto spillDir = setupSpillPath();
+
+    const int numBatchesPerFile = 5;
```
mbasmanova, line 1142: Do we need 5 batches of 1000 rows each and 5 files? Can we make 1 batch with 10 rows and use Values node instead of TableScan? This will make this test run faster.

presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp
```diff
@@ -1132,6 +1132,66 @@ TEST_F(TaskManagerTest, checkBatchSplits) {
...
+      taskManager_->setBaseSpillDirectory(
+          fmt::format("{}/presto_native/path", spillDir->path));
+    } else {
+      // Set bad spill path.
+      const std::string badSpillPath = "/nonexist/presto_native/path";
+      taskManager_->setBaseSpillDirectory(badSpillPath);
+    }
+
+    std::map<std::string, std::string> queryConfigs;
+    queryConfigs.emplace(core::QueryConfig::kTestingSpillPct, "100");
```
mbasmanova, line 1163: Would it make sense to extract a helper function to "enable spilling"?

mbasmanova (commented):
@xiaoxmeng Thank you for explaining and making the PR easier to review. Looks good % a couple of questions.

mbasmanova (commented):
@xiaoxmeng Would you update commit message to provide more details about the changes?

presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp
xiaoxmeng: Not required

presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp
xiaoxmeng: Sure.

presto-native-execution/presto_cpp/main/TaskManager.cpp
```diff
@@ -436,20 +435,22 @@ std::unique_ptr<TaskInfo> TaskManager::createOrUpdateTask(
         return std::make_unique<TaskInfo>(prestoTask->updateInfoLocked());
       }
 
-      execTask = exec::Task::create(
+      auto newExecTask = exec::Task::create(
```
mbasmanova, line 438: nit: maybe add a comment to explain the use a separate variable

presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp
```diff
@@ -1132,6 +1138,71 @@ TEST_F(TaskManagerTest, checkBatchSplits) {
   ASSERT_EQ(resultOrFailure.status, nullptr);
 }
 
+TEST_F(TaskManagerTest, buildSpillDirectryFailure) {
```
mbasmanova, line 1141: typo: Directry -> Directory

presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp
```diff
@@ -1132,6 +1138,71 @@ TEST_F(TaskManagerTest, checkBatchSplits) {
...
+  // Cleanup old tasks between test iterations.
+  taskManager_->setOldTaskCleanUpMs(0);
+  for (bool buildSpillDirectoryFailure : {false}) {
+    SCOPED_TRACE(fmt::format(
+        "buildSpillDirectoryFailure: {}", buildSpillDirectoryFailure));
+    auto spillDir = setupSpillPath();
+
+    auto filePaths = makeFilePaths(1);
```
mbasmanova, line 1149: Why do we need to write data to files? Can we not use values(batches) node in the plan?

presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp
```diff
@@ -1132,6 +1138,71 @@ TEST_F(TaskManagerTest, checkBatchSplits) {
...
+    updateRequest.session.systemProperties = queryConfigs;
+    // Create task will fail if the spilling directory setup fails.
+    if (buildSpillDirectoryFailure) {
+      VELOX_ASSERT_THROW(
+          createOrUpdateTask(taskId, updateRequest, planFragment), "Mkdir");
+    } else {
+      createOrUpdateTask(taskId, updateRequest, planFragment);
+    }
+    {
+      auto taskMap = taskManager_->tasks();
```
mbasmanova, line 1182: Maybe move this block into 'else' branch above.

mbasmanova (approved):
@xiaoxmeng Please, update PR description to match the code. Consider explaining why directory creation may fail as it is non-intuitive. Please, update commit message to add details.

PR #21030, opened 2023-10-03
[native] Add connector.id to connector config

vermapratyush:
## Description
Generate and store identity for use by filesystem to uniquely identify a 
catalog even if the connector.name is same.

## Motivation and Context
Fixes https://github.com/facebookincubator/velox/issues/6633

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/PrestoServer.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoServer.cpp
@@ -68,6 +68,7 @@ constexpr char const* kHttps = "https";
 constexpr char const* kTaskUriFormat =
     "{}://{}:{}"; // protocol, address and port
 constexpr char const* kConnectorName = "connector.name";
+constexpr char const* kConnectorId = "connector.id";
 
 protocol::NodeState convertNodeState(presto::NodeState nodeState) {
   switch (nodeState) {
@@ -691,6 +692,10 @@ std::vector<std::string> PrestoServer::registerConnectors(
           fileName.substr(0, fileName.size() - kPropertiesExtension.size());
 
       auto connectorConf = util::readConfig(entry.path());
+      // Store catalogName for use by filesystem to uniquely identify
+      // a catalog even if the connector.name is same.
+      connectorConf.emplace(kConnectorId, catalogName);
+
       PRESTO_STARTUP_LOG(INFO)
           << "Registered properties from " << entry.path() << ":\n"
           << stringifyConnectorConfig(connectorConf);


presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -68,6 +68,7 @@ constexpr char const* kHttps = "https";
 constexpr char const* kTaskUriFormat =
     "{}://{}:{}"; // protocol, address and port
 constexpr char const* kConnectorName = "connector.name";
+constexpr char const* kFileSystemIdentity = "fs.identity";
```
xiaoxmeng, line 71: Why we need this identity? Thanks!

majetideepak (changes requested):
@vermapratyush The `connectorId` (catalog name in Presto) is tied to the connector instance. Can we use that as the ID?

spershin:
> @vermapratyush The `connectorId` (catalog name in Presto) is tied to the connector instance. Can we use that as the ID?

@majetideepak 
I don't see "connectorId" is being used in the catalog config files in Meta.
What is being used is "connector.name".
Example:
connector.name=xdb

But the connector.name can be the same in multiple config files for different catalogs.

So, we are looking at something unique for the file system creation in Velox.
At that unique is seems to be a combination of the config file name + connector.name.
Actually just the config file name would suffice too - it is unique, but maybe nicer to add that connector.name.

Hence the new property.
cc @xiaoxmeng , @vermapratyush

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -691,6 +692,12 @@ std::vector<std::string> PrestoServer::registerConnectors(
           fileName.substr(0, fileName.size() - kPropertiesExtension.size());
 
       auto connectorConf = util::readConfig(entry.path());
+      // Generate and store identity for use by filesystem to uniquely identify
+      // a catalog even if the connector.name is same.
+      connectorConf.emplace(
+          kFileSystemIdentity,
+          fmt::format("{}-{}", fileName, connectorConf[kConnectorName]));
```
spershin, line 699: @vermapratyush 
I think we should use "catalogName" here, not "fileName"?
We don't need the extension.

vermapratyush: Updated.

spershin (approved):
Looks good.
Thank you!

vermapratyush:
@majetideepak 

Similar to HdfsFilesystem, which leverage host and port to create an identity https://github.com/facebookincubator/velox/blob/4da8fb60908e603363ff6ebc68b6f766bf424f1c/velox/connectors/hive/storage_adapters/hdfs/HdfsFileSystem.cpp#L101-L110

It is able to decide if a new filesystem instance needs to be [created or not](https://github.com/facebookincubator/velox/blob/4da8fb60908e603363ff6ebc68b6f766bf424f1c/velox/connectors/hive/storage_adapters/hdfs/RegisterHdfsFileSystem.cpp#L44), our aim is to achieve the same using `fs.identity`. Currently in Meta Warmstorage there is no such identity within the config(connector) file.

If `connectorId` is accessible inside filesystem that will also serve the purpose. But is there a way to access this from inside of filesystem?

majetideepak:
@vermapratyush, thanks for the clarification. Why not add `fs.identity = warmStorage1` explicitly in your config? And your warm storage FS can look for this value.
Adding this field for every config is redundant.

majetideepak:
IMO just like S3, HDFS, Meta's warm storage must define an identity for the FS explicitly.

vermapratyush:
@majetideepak We did discuss that internally but thought it might be useful for other filesystems as well to be able to make this distinction.

Thanks for the input. Let me follow this up internally.

majetideepak:
@vermapratyush Thanks! Other filesystems such as S3, HDFS, depend mainly on the properties of the filesystem such as endpoint, credentials, etc.

spershin:
@majetideepak 
Requiring users to add "fs.identity" manually to config files increase a chance of human error.
Especially when they have dozens or more different configs in their system.
The code can take care very easily and automatically about this field and eliminate the human error.

I'm trying to understand what in this change got you opposing it so much?
Is there an issue I don't see?
Thank you.

majetideepak:
@spershin `fs.identity` seems very specific to the filesystem API. It also is simply the combination of `catalogName- conenctorName`. We should avoid adding such special configs and new concepts if possible.
I see your point about human error. Should we instead add `connector.id` as is automatically in the config? We can use this for other features as well in the future. You can add a function 
```
   std::string fsIdentity(Config* config) {
       return config->connector.name + config->connector.id;  
   }
```

vermapratyush:
@majetideepak `config->connector.id` currently does not exists. Do you mean changing `config["fs.identity"]` to `config["connector.id"]` ?
Also, having the unique identifier as `connector.id + connector.name` isn't important, anything that is unique works eg `connector.id`.

majetideepak:
> Do you mean changing `config["fs.identity"]` to `config["connector.id"]`

@vermapratyush Yes, this is what I meant. It is better to define an existing concept rather than a new one.

vermapratyush:
Thanks @majetideepak . Updated the PR

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -691,6 +692,10 @@ std::vector<std::string> PrestoServer::registerConnectors(
           fileName.substr(0, fileName.size() - kPropertiesExtension.size());
 
       auto connectorConf = util::readConfig(entry.path());
+      // Store catalogName for use by filesystem to uniquely identify
```
majetideepak, line 695: Perhaps `Store catalogName as 'connector.id' to uniquely identify a connector instance`?
Multiline comments also need `///`

majetideepak:
Please update the commit's body as well. We just need to mention the uniqueness property.

mbasmanova (changes requested):
Folks, there is an ongoing discussion about this. Please, hold off on changes until it is resolved.

PR #21031, opened 2023-10-03
[native] Add list all task endpoint to server operation

tanjialiang:
## Description
* Refactored some API naming
* Added list all task endpoint to server operation that lists all current existing tasks in the server along with their current state.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/PrestoServerOperations.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoServerOperations.cpp
@@ -84,8 +84,12 @@ void PrestoServerOperations::runOperation(
       case ServerOperation::Target::kVeloxQueryConfig:
         http::sendOkResponse(
             downstream, veloxQueryConfigOperation(op, message));
-      case ServerOperation::Target::kDebug:
-        http::sendOkResponse(downstream, debugOperation(op, message));
+        break;
+      case ServerOperation::Target::kTask:
+        http::sendOkResponse(downstream, taskOperation(op, message));
+        break;
+      case ServerOperation::Target::kServer:
+        http::sendOkResponse(downstream, serverOperation(op, message));
         break;
     }
   } catch (const velox::VeloxUserError& ex) {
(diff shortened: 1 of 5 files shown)


presto-native-execution/presto_cpp/main/PrestoServerOperations.cpp
```diff
@@ -185,22 +189,42 @@ std::string PrestoServerOperations::veloxQueryConfigOperation(
...
         return fmt::format("No task found with id {}", id);
       }
       return task->second->toJsonString();
     }
+    case ServerOperation::Action::kListAll: {
+      std::stringstream oss;
+      oss << "[";
+      for (const auto& task : taskMap) {
+        oss << task.first << "(" << taskStateString(task.second->task->state())
+            << "),\n";
```
spershin, line 213: Imagine you run this for a cluster of 1000 workers.... Having 1 line per task would be overkill.

presto-native-execution/presto_cpp/main/PrestoServerOperations.cpp
```diff
@@ -185,22 +189,42 @@ std::string PrestoServerOperations::veloxQueryConfigOperation(
...
+      if (task == taskMap.end()) {
         return fmt::format("No task found with id {}", id);
       }
       return task->second->toJsonString();
     }
+    case ServerOperation::Action::kListAll: {
+      std::stringstream oss;
+      oss << "[";
+      for (const auto& task : taskMap) {
+        oss << task.first << "(" << taskStateString(task.second->task->state())
```
spershin, line 212: I think the underlying Velox task could be null.

presto-native-execution/presto_cpp/main/PrestoServerOperations.cpp
xiaoxmeng: We could pass count with zero to fetch all, otherwise stop at the specified count.

presto-native-execution/presto_cpp/main/PrestoServerOperations.cpp
```diff
@@ -185,22 +189,42 @@ std::string PrestoServerOperations::veloxQueryConfigOperation(
...
+std::string PrestoServerOperations::taskOperation(
     const ServerOperation& op,
     proxygen::HTTPMessage* message) {
+  if (taskManager_ == nullptr) {
+    return "Task Manager not found";
+  }
+  const auto taskMap = taskManager_->tasks();
   switch (op.action) {
-    case ServerOperation::Action::kTask: {
+    case ServerOperation::Action::kGetDetail: {
```
xiaoxmeng, line 200: We can combine the two operations into one
kTask:
id=, then look for all the tasks
count=0, then fetch all the tasks without limitation.

xiaoxmeng (commented):
@tanjialiang thanks for the change % comments.

presto-native-execution/presto_cpp/main/PrestoServerOperations.cpp
tanjialiang: This is not supposed to be run on all workers at the same time. But yeah we can have that count feature

presto-native-execution/presto_cpp/main/PrestoServerOperations.cpp
tanjialiang: Yeah merging them would be nice concise-wise. Though separating them can make it easier to get the meaning directly from the target/action level. Shall we keep it separated in this case? Thanks

presto-native-execution/presto_cpp/main/PrestoServerOperations.cpp
```diff
@@ -185,22 +189,59 @@ std::string PrestoServerOperations::veloxQueryConfigOperation(
...
     }
+    case ServerOperation::Action::kListAll: {
+      uint32_t limit;
+      try {
+        const auto& limitStr = message->getQueryParam("limit");
+        limit = limitStr == proxygen::empty_string
+            ? std::numeric_limits<uint32_t>::max()
+            : stoi(limitStr);
+      } catch (std::exception& ex) {
+        VELOX_FAIL(ex.what());
```
xiaoxmeng, line 216: Use VELOX_USER_FAIL? Thanks!

presto-native-execution/presto_cpp/main/PrestoServerOperations.cpp
```diff
@@ -185,22 +189,59 @@ std::string PrestoServerOperations::veloxQueryConfigOperation(
...
+      }
+      std::stringstream oss;
+      oss << "[";
+      uint32_t count = 0;
+      for (const auto& task : taskMap) {
+        const auto& veloxTask = task.second->task;
+        if (veloxTask == nullptr) {
```
xiaoxmeng, line 223: Maybe print out the task id but null for velox task state?
task id, velox state?

xiaoxmeng (approved):
@tanjialiang LGTM. Thanks!

presto-native-execution/presto_cpp/main/PrestoServerOperations.cpp
```diff
@@ -185,22 +189,58 @@ std::string PrestoServerOperations::veloxQueryConfigOperation(
...
+      }
+      std::stringstream oss;
+      oss << "[";
+      uint32_t count = 0;
+      for (const auto& task : taskMap) {
+        const auto& veloxTask = task.second->task;
+        if (++count > limit) {
+          break;
+        }
+        oss << task.first << "("
```
aditi-pandit, line 226: @tanjialiang : If there are more than limit tasks, maybe we should specify in the result string the number of tasks in taskMap also.

tanjialiang: yeah good point

PR #21037, opened 2023-10-04
Implement random shuffle of probe side for skew mitigation

mlyublena:
* Randomize the skewed side of a join by transforming:
 *     join (REPLICATED)
 *       - scan S
 *       - scan T
 *
 * into
 *
 *     join (REPLICATED)
 *       - Exchange(ROUND ROBIN)
 *         - scan S
 *       - scan T

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* ...
* ...

Hive Changes
* ...
* ...
```

```
== NO RELEASE NOTE ==
```


ClarenceThreepwood:
Design question: What if there is an upstream operator that depends on the output of this join to be partitioned a particular way? For e.g. if there is an upstream join or groupby that expects the output of the current replicated join to be partitioned on the join key of (S), i.e something like (t1 JOIN (t2 JOIN t3 ON t2.b = t3.c) ON t1.a = t2.b). 
Now if this rule randomly repartitions the join input, and this plan is changed after AddExchanges is done, would that not result in wrong results?

Another way to put it - from what I can tell, changing any partitioning/distribution after AddExchanges is probably not a good idea. Thoughts?

ClarenceThreepwood (commented):
Design question: What if there is an upstream operator that depends on the output of this join to be partitioned a particular way? For e.g. if there is an upstream join or groupby that expects the output of the current replicated join to be partitioned on the join key of (S), i.e something like (t1 JOIN (t2 JOIN t3 ON t2.b = t3.c) ON t1.a = t2.b).
Now if this rule randomly repartitions the join input, and this plan is changed after AddExchanges is done, would that not result in wrong results?

Another way to put it - from what I can tell, changing any partitioning/distribution after AddExchanges is probably not a good idea. Thoughts?

jaystarshot:
Just some thoughts, please consider as nits. 
Maybe we can add this as a new type of join like SKEWED? 
Also can we incorporate this in determine join distribution rule and the reorder joins rule?

PR #21045, opened 2023-10-05
Support implicit mapping of Presto datatype Char to Iceberg datatype String

dmariamgeorge:
## Description
This PR addresses the `Type not supported for Iceberg: char(3)` error which is getting thrown when creating an Iceberg table using the "Create Table As Select" (CTAS) operation, when the source table contains character columns. 
This fix supports the mapping of Presto datatype Char to the equivalent datatype String in Iceberg. 

## Motivation and Context
Creating an Iceberg table using CTAS fails when the source table has character columns. It is unable to create table as the query throws error `Type not supported for Iceberg: char(3)`. The expected behaviour should be that CHAR fields in the source table should be implicitly mapped to String when used in a CTAS for an Iceberg table. Without the implicit mapping of Char to String the process of using CTAS when the source table has character columns, becomes very cumbersome with the need to specify and cast each individual column. With this fix, the CHAR fields in the source table can be implicitly mapped to the String data type in the Iceberg table. 

## Impact
The fix supports the implicit mapping of Char data type to String data type, when creating CTAS table containing character columns in the source table.

## Test Plan
Tested the fix by creating a table in DB2 having column where the data type is CHAR.
Tried creating an Iceberg table using CTAS, selecting from the created DB2 table. Successfully executed the query. 

## Contributor checklist

- [x] CI passed.

## Release Notes
```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/TypeConverter.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/TypeConverter.java
@@ -161,6 +161,9 @@ public final class TypeConverter
         if (type instanceof VarcharType) {
             return Types.StringType.get();
         }
+        if (type instanceof CharType) {
+            return Types.StringType.get();
+        }
         if (type instanceof VarbinaryType) {
             return Types.BinaryType.get();
         }
--- a/presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedSmokeTestBase.java
+++ b/presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedSmokeTestBase.java
@@ -987,4 +987,14 @@ public class IcebergDistributedSmokeTestBase
                         "  (NULL, NULL, NULL, NULL, 2e0, NULL, NULL)");
         dropTable(session, tableName);
     }
+
+    @Test
+    public void testCharTable()
+    {
+        String tableName = "test_char_table";
+        assertUpdate(format("CREATE TABLE %s (char_column CHAR(5))", tableName));
+        assertUpdate(format("INSERT INTO %s VALUES ('abcde'), ('12345')", tableName), 2);
+        assertQuery(format("SELECT * FROM %s", tableName), "VALUES ('abcde'), ('12345')");
+        dropTable(getSession(), tableName);
+    }
 }


tdcmeehan:
Thank you for this contribution.  Please add a test in `IcebergDistributedSmokeTestBase` that shows round trip ability to read and write to a table that has the char type.

Please also rebase to address the build failure on main. 

[It would be very nice to add to the documentation the current supported types in the Iceberg connector](https://github.com/prestodb/presto/blob/154ceeb8d18289ae5feea38f93922023dcf14ae9/presto-docs/src/main/sphinx/connector/iceberg.rst), though we can save this for another issue.

hantangwangd:
We should take length in into consideration when we want to support CharType in Iceberg table, or we may meet a lot of failure in test, for example when CharType column appeared in filter predicate, see AbstractTestDistributedQueries.testStringFilters(). May be a feasible solution is that, in Iceberg metadata layer, we use FixedType(length) to express CharType(length), and in Parquet layer, we use FIXED_LEN_BYTE_ARRAY or BINARY(in order to save space) to express it.

dmariamgeorge:
> Thank you for this contribution. Please add a test in `IcebergDistributedSmokeTestBase` that shows round trip ability to read and write to a table that has the char type.
> 
> Please also rebase to address the build failure on main.
> 
> [It would be very nice to add to the documentation the current supported types in the Iceberg connector](https://github.com/prestodb/presto/blob/154ceeb8d18289ae5feea38f93922023dcf14ae9/presto-docs/src/main/sphinx/connector/iceberg.rst), though we can save this for another issue.

@tdcmeehan - Could you please show an example to understand more on the test that shows round trip ability to read and write to a table?

tdcmeehan:
@dmariamgeorge we just need to add tests that prove the mapping works.  The test file I referenced should give sufficient examples of the flavor of testing.

tdcmeehan:
@dmariamgeorge just wanted to confirm you're still working on this?

dmariamgeorge:
> @dmariamgeorge just wanted to confirm you're still working on this?

@tdcmeehan - Yes, I'm working on this.

dmariamgeorge:
Hi @tdcmeehan 

As you have suggested, I have added a test in IcebergDistributedSmokeTestBase.java to validate the char type. Could you please a take look into the same? Thanks!

tdcmeehan:
@dmariamgeorge please squash commits. LGTM

tdcmeehan:
@dmariamgeorge commits are still not squashed

tdcmeehan:
@dmariamgeorge please notify the reviewer once changes are addressed.

This looks good, but please rebase to address the test failure, and follow the instructions posted by `linux-foundation-easycla` to sign the CLA.

dmariamgeorge:
Hi @tdcmeehan 

I have addressed the suggested changes. I have rebased to address the test failure, and have signed the CLA. Thanks!

tdcmeehan:
Upon more consideration, I have a question.  If we implicitly cast CHAR(3) to VARCHAR (mapped to Iceberg's String type), are there any unintended consequences wrt blank padding?  For example, supposing we create a table with CHAR(3), we'll map it to String.  If I insert a single character `'A'`, then I believe ANSI semantics are such that if I were to select it back, it should return `'A  '` (blank padded with two extra blank spaces).  How does DB2 and other engines handle this mapping?

dmariamgeorge:
> Upon more consideration, I have a question. If we implicitly cast CHAR(3) to VARCHAR (mapped to Iceberg's String type), are there any unintended consequences wrt blank padding? For example, supposing we create a table with CHAR(3), we'll map it to String. If I insert a single character `'A'`, then I believe ANSI semantics are such that if I were to select it back, it should return `'A '` (blank padded with two extra blank spaces). How does DB2 and other engines handle this mapping?

@tdcmeehan - Tested the behaviour by creating a table in DB2 having a column named emp_name of type char(3). Inserted a row having single character, 'B' into the table. Then, created a table in Iceberg using the table in DB2 through ctas. CHAR(3) is implicitly mapped to VARCHAR with no extra blank spaces. 

Screenshot 2023-12-19 at 12 36 43 PM

tdcmeehan:
@dmariamgeorge My concern is I am not sure it's the correct behavior to coerced `CHAR(n)` to `String`, since there are blank padding semantics that must be lost in translation.  Could you take inventory of how other engines handle this?

hantangwangd:
Agree with @tdcmeehan, I have the same concern.

gupteaj:
we may need to preserve blank padding in this case as suggested by @tdcmeehan , @hantangwangd . 
If possible , we can add session / presto property that will support both modes ( trim blank spaces and preserve blank spaces ). 
It should support similar case as varchar type 

```
presto:test> create table t1_str ( a int, b varchar);
presto:test> insert into t1_str values ( 1, 'abc  ');
presto:test> insert into t1_str values ( 1, 'abcd  ');
presto:test> select b, length(b) as length from t1_str;

   b    | length 
--------+--------
 abcd   |      6 
 abc    |      5 
(2 rows)

```

dmariamgeorge:
> Upon more consideration, I have a question. If we implicitly cast CHAR(3) to VARCHAR (mapped to Iceberg's String type), are there any unintended consequences wrt blank padding? For example, supposing we create a table with CHAR(3), we'll map it to String. If I insert a single character `'A'`, then I believe ANSI semantics are such that if I were to select it back, it should return `'A '` (blank padded with two extra blank spaces). How does DB2 and other engines handle this mapping?

Upon casting char(n) to varchar in Iceberg, could find that leading blank spaces are preserved while the trailing blank spaces are not preserved. 

```
presto> create table postgresql.tm_lakehouse_engine.yolo (name char(10));
CREATE TABLE
presto> insert into postgresql.tm_lakehouse_engine.yolo values ('a   '),('   a'),('   a   ');
INSERT: 3 rows

Query 20240128_070033_00010_zzpx5, FINISHED, 1 node
Splits: 35 total, 35 done (100.00%)
[Latency: client-side: 0:43, server-side: 0:43] [0 rows, 0B] [0 rows/s, 0B/s]

presto> select name, length(name) as name_length from postgresql.tm_lakehouse_engine.yolo;
    name    | name_length 
------------+-------------
 a          |          10 
    a       |          10 
    a       |          10 
(3 rows)

Query 20240128_070141_00011_zzpx5, FINISHED, 1 node
Splits: 17 total, 17 done (100.00%)
[Latency: client-side: 0:21, server-side: 0:21] [3 rows, 0B] [0 rows/s, 0B/s]

presto> create table iceberg.org.ctas as select * from postgresql.tm_lakehouse_engine.yolo;
CREATE TABLE: 3 rows

Query 20240128_070318_00012_zzpx5, FINISHED, 1 node
Splits: 19 total, 19 done (100.00%)
[Latency: client-side: 0:23, server-side: 0:23] [3 rows, 0B] [0 rows/s, 0B/s]

presto> select name, length(name) as name_length from iceberg.org.ctas;
 name | name_length 
------+-------------
 a    |           1 
    a |           4 
    a |           4 
(3 rows)
```

tdcmeehan:
@dmariamgeorge how would you propose we address this limitation?

dmariamgeorge:
> @dmariamgeorge how would you propose we address this limitation?

@tdcmeehan - Trying to identify the area of code where the trimming of trailing spaces occurs, and checking if a fix can be added to resolve the limitation.

tdcmeehan:
Please add an issue specifying how you will address this design limitation, then feel free to reopen (as this PR is very stale).

PR #21048, opened 2023-10-05
Support deleting whole partitions in Iceberg table

hantangwangd:
## Description

We can use `delete from <table> [where <predicate>];` to delete one or more whole partitions or all data in Iceberg table when some conditions are met:
 - The predicate could be wholly pushed down and exactly enforced.
 - The predicate Indicated one or more whole partitions in Iceberg table.

## Motivation and Context

This PR allows deleting partition level data from Iceberg tables.

## Test Plan

 - Make sure filter push down optimization do not affect existing test cases in IcebergDistributedTestBase
 - Newly added test case in IcebergDistributedTestBase.testDelete()

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* The Iceberg connector now supports `DELETE FROM <table> [where <filter>]` statements to delete one or more entire partitions
```

### DIFF
--- a/presto-docs/src/main/sphinx/connector/iceberg.rst
+++ b/presto-docs/src/main/sphinx/connector/iceberg.rst
@@ -511,6 +511,25 @@ dropping the table from the metadata catalog using ``TRUNCATE TABLE``.
     -----------+------+-----------+---------
     (0 rows)
 
+DELETE
+^^^^^^^^
+
+The iceberg connector can delete data in one or more entire partitions from tables by using ``DELETE FROM``. For example, to delete from the table ``lineitem``::
+
+     DELETE FROM lineitem;
+
+     DELETE FROM lineitem WHERE linenumber = 1;
+
+     DELETE FROM lineitem WHERE linenumber not in (1, 3, 5, 7) and linestatus in ('O', 'F');
+
+.. note::
+
+    Columns in the filter must all be identity transformed partition columns of the target table.
+
+    Filtered columns only support comparison operators, such as EQUALS, LESS THAN, or LESS THAN EQUALS.
+
+    Deletes must only occur on the latest snapshot.
+
 DROP TABLE
 ^^^^^^^^^^^
 
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
@@ -49,11 +49,14 @@ import org.apache.iceberg.DataFiles;
 import org.apache.iceberg.DeleteFiles;
 import org.apache.iceberg.FileScanTask;
 import org.apache.iceberg.PartitionField;
+import org.apache.iceberg.PartitionSpec;
 import org.apache.iceberg.PartitionSpecParser;
 import org.apache.iceberg.Schema;
 import org.apache.iceberg.SchemaParser;
 import org.apache.iceberg.Table;
+import org.apache.iceberg.TableScan;
 import org.apache.iceberg.Transaction;
+import org.apache.iceberg.expressions.Expression;
 import org.apache.iceberg.io.CloseableIterable;
 import org.apache.iceberg.types.Type;
 import org.apache.iceberg.types.TypeUtil;
@@ -65,11 +68,14 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.OptionalLong;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.stream.Collectors;
 
 import static com.facebook.presto.common.type.BigintType.BIGINT;
+import static com.facebook.presto.iceberg.ExpressionConverter.toIcebergExpression;
 import static com.facebook.presto.iceberg.IcebergColumnHandle.primitiveIcebergColumnHandle;
 import static com.facebook.presto.iceberg.IcebergErrorCode.ICEBERG_INVALID_SNAPSHOT_ID;
 import static com.facebook.presto.iceberg.IcebergTableProperties.FILE_FORMAT_PROPERTY;
@@ -89,6 +95,7 @@ import static com.facebook.presto.spi.StandardErrorCode.NOT_SUPPORTED;
 import static com.google.common.base.Verify.verify;
 import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.google.common.collect.ImmutableMap.toImmutableMap;
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
 import static java.lang.String.format;
 import static java.util.Collections.singletonList;
 import static java.util.Objects.requireNonNull;
@@ -414,6 +421,7 @@ public abstract class IcebergAbstractMetadata
                 name.getTableName(),
                 name.getTableType(),
                 resolveSnapshotIdByName(table, name),
+                name.getSnapshotId().isPresent(),
                 TupleDomain.all());
     }
 
@@ -438,6 +446,101 @@ public abstract class IcebergAbstractMetadata
         return getIcebergSystemTable(tableName, icebergTable);
     }
 
+    @Override
+    public void truncateTable(ConnectorSession session, ConnectorTableHandle tableHandle)
+    {
+        IcebergTableHandle handle = (IcebergTableHandle) tableHandle;
+        Table icebergTable = getIcebergTable(session, handle.getSchemaTableName());
+        try (CloseableIterable<FileScanTask> files = icebergTable.newScan().planFiles()) {
+            removeScanFiles(icebergTable, files);
+        }
+        catch (IOException e) {
+            throw new PrestoException(GENERIC_INTERNAL_ERROR, "failed to scan files for delete", e);
+        }
+    }
+
+    @Override
+    public ConnectorTableHandle beginDelete(ConnectorSession session, ConnectorTableHandle tableHandle)
+    {
+        IcebergTableHandle handle = (IcebergTableHandle) tableHandle;
+        if (handle.isSnapshotSpecified()) {
+            throw new PrestoException(NOT_SUPPORTED, "This connector do not allow delete data at specified snapshot");
+        }
+        throw new PrestoException(NOT_SUPPORTED, "This connector only supports delete where one or more partitions are deleted entirely");
+    }
+
+    @Override
+    public boolean supportsMetadataDelete(ConnectorSession session, ConnectorTableHandle tableHandle, Optional<ConnectorTableLayoutHandle> tableLayoutHandle)
+    {
+        IcebergTableHandle handle = (IcebergTableHandle) tableHandle;
+        if (handle.isSnapshotSpecified()) {
+            return false;
+        }
+
+        if (!tableLayoutHandle.isPresent()) {
+            return true;
+        }
+
+        // Allow metadata delete for range filters on partition columns.
+        IcebergTableLayoutHandle layoutHandle = (IcebergTableLayoutHandle) tableLayoutHandle.get();
+
+        TupleDomain<ColumnHandle> domainPredicate = layoutHandle.getTupleDomain();
+        if (domainPredicate.isAll()) {
+            return true;
+        }
+
+        Set<Integer> predicateColumnIds = domainPredicate.getDomains().get().keySet().stream()
+                .map(IcebergColumnHandle.class::cast)
+                .map(IcebergColumnHandle::getId)
+                .collect(toImmutableSet());
+        Table icebergTable = getIcebergTable(session, handle.getSchemaTableName());
+
+        boolean supportsMetadataDelete = true;
+        for (PartitionSpec spec : icebergTable.specs().values()) {
+            // Currently we do not support delete when any partition columns in predicate is not transform by identity()
+            Set<Integer> partitionColumnSourceIds = spec.fields().stream()
+                    .filter(field -> field.transform().isIdentity())
+                    .map(PartitionField::sourceId)
+                    .collect(Collectors.toSet());
+
+            if (!partitionColumnSourceIds.containsAll(predicateColumnIds)) {
+                supportsMetadataDelete = false;
+                break;
+            }
+        }
+
+        return supportsMetadataDelete;
+    }
+
+    @Override
+    public OptionalLong metadataDelete(ConnectorSession session, ConnectorTableHandle tableHandle, ConnectorTableLayoutHandle tableLayoutHandle)
+    {
+        IcebergTableHandle handle = (IcebergTableHandle) tableHandle;
+        IcebergTableLayoutHandle layoutHandle = (IcebergTableLayoutHandle) tableLayoutHandle;
+
+        Table icebergTable;
+        try {
+            icebergTable = getIcebergTable(session, handle.getSchemaTableName());
+        }
+        catch (Exception e) {
+            throw new TableNotFoundException(handle.getSchemaTableName());
+        }
+
+        TableScan scan = icebergTable.newScan();
+        TupleDomain<ColumnHandle> domainPredicate = layoutHandle.getTupleDomain();
+        if (!domainPredicate.isAll()) {
+            Expression filterExpression = toIcebergExpression(handle.getPredicate());
+            scan = scan.filter(filterExpression);
+        }
+
+        try (CloseableIterable<FileScanTask> files = scan.planFiles()) {
+            return OptionalLong.of(removeScanFiles(icebergTable, files));
+        }
+        catch (IOException e) {
+            throw new PrestoException(GENERIC_INTERNAL_ERROR, "failed to scan files for delete", e);
+        }
+    }
+
     /**
      * Deletes all the files within a particular scan
      *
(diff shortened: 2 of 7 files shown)


presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
```diff
@@ -432,6 +438,98 @@ public Optional<SystemTable> getSystemTable(ConnectorSession session, SchemaTabl
...
+                return OptionalLong.of(removeScanFiles(icebergTable, files));
+            }
+            catch (IOException e) {
+                throw new PrestoException(GENERIC_INTERNAL_ERROR, "failed to scan files for delete", e);
+            }
+        }
+        else {
+            Expression filterExpression = toIcebergExpression(handle.getPredicate());
+            try (CloseableIterable<FileScanTask> files = icebergTable.newScan()
+                    .filter(filterExpression).useSnapshot(handle.getSnapshotId().get()).planFiles()) {
```
ZacBlanco, line 524: So it seems like you've considered the case where a user tries to delete metadata from the snapshot of a table, but I don't fully understand the semantics of how that would work. If you delete data in older snapshots that doesn't propagate to newer snapshots does it?

e.g. you insert record 1 and record 2 in snapshot A, you insert record 3 in snapshot B. All records are in the same partition. Then you run `DELETE FROM "<table>@a" WHERE ....` for  snapshot C. This deletes the partition containing records 1, 2, and 3. However, In snapshot C does this mean when you query you will only get record 3, or do you get no records?

I have a few questions I'd like to answer:

1. If we do support deleting partitions from a particular snapshot, what are the correct semantics? Should we document it? Does iceberg document it?
2. Otherwise, if there isn't a clear definition from Iceberg, personally I think we could disallow metadata deletes in the case a user queries an iceberg table like `"orders@<snapshot>"`. What is your opinion?
3. In a similar vein, how about partition evolution, and that in combination with querying at specific snapshots? Will this handle the case where you try to delete a partition which was not yet created/specified at a particular snapshot?

If we do support all of these case (deleting at particular snapshots + partition evolution), then I think we should also add some tests for them.

ZacBlanco (changes requested):
My main concern is about semantics of delete with snapshots/partition evolution. It's not clear to me how exactly this is handled. Also, we should probably update the docs. Looks good otherwise though!!

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
hantangwangd: Great questions. Originally I did not consider metadata deletes with user given snapshot, just sorry for overlooked this kind of usage. But now I think we could support deleting partitions from a particular snapshot, considering iceberg's specification about metadata and snapshot. In your example, record 3 should be the remaining record, because even in the same partition, data of different snapshots would be maintained in different files. I will do some tests to confirm that.

As for partition evolution, in current implementation, the requirements for supporting metadata deletes has been checked for each partition specs, I think that could make sure there is no problem. I will add some tests for it.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
```diff
@@ -432,6 +438,98 @@ public Optional<SystemTable> getSystemTable(ConnectorSession session, SchemaTabl
...
+        // Allow metadata delete for range filters on partition columns.
+        IcebergTableLayoutHandle layoutHandle = (IcebergTableLayoutHandle) tableLayoutHandle.get();
+
+        TupleDomain<ColumnHandle> domainPredicate = layoutHandle.getTupleDomain();
+        if (domainPredicate.isAll()) {
+            return true;
+        }
+
+        Set<String> predicateColumnNames = domainPredicate.getDomains().get().keySet().stream()
+                .map(columnHandle -> ((IcebergColumnHandle) columnHandle).getName())
```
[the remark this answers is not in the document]
hantangwangd: Smells better, I will fix it.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
ZacBlanco: Sounds good! Please let me know if you find anything in the Iceberg docs

hantangwangd: Iceberg specification do not define such behavior. And in iceberg-api, it's interface DeleteFiles declares that all it's behavior of file deletions would be applied to the latest table snapshot. So, I think it's better to explicitly disallow user given snapshot for metadata deletes. Such behavior may lead in confusion, for example, when we try to delete some rows in an older snapshot which are modified or deleted in newer snapshots.

As for partition evolution, it seems like we haven't consider partition spec's evolution when try to alter table,
for example rename column, add column or drop column. So I plan to make some PRs to support that. When it's done, we can add test cases for partition evolution.

Others are all fixed, please take a look.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedTestBase.java
```diff
@@ -62,9 +62,54 @@ public void testRenameColumn()
...
+        assertEquals(totalCount - countPart1F - countPart2O - countPartOther, newTotalCount);
+        assertQuerySucceeds("DROP TABLE test_partitioned_drop");
+
+        // Do not support delete with filters about non-identity partition column
+        String errorMessage = "This connector only supports delete where one or more partitions are deleted entirely";
+        assertUpdate("CREATE TABLE test_partitioned_drop WITH (partitioning = ARRAY['bucket(orderkey, 2)', 'linenumber', 'linestatus']) as select * from lineitem", totalCount);
+        assertQueryFails("DELETE FROM test_partitioned_drop WHERE orderkey = 1", errorMessage);
+        assertQueryFails("DELETE FROM test_partitioned_drop WHERE partkey > 100", errorMessage);
+        assertQueryFails("DELETE FROM test_partitioned_drop WHERE linenumber = 1 and orderkey = 1", errorMessage);
+        assertQuerySucceeds("DROP TABLE test_partitioned_drop");
```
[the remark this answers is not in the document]
hantangwangd: Now we explicitly do not allow such behavior, tests have been added.

ZacBlanco (changes requested):
Looks good! Mostly just small nits now. I think we should also update the docs to specify this is now supported and limited to partition-only and without specific snapshots

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergTableHandle.java
```diff
@@ -33,19 +33,22 @@ public class IcebergTableHandle
...
     private final TupleDomain<IcebergColumnHandle> predicate;
+    private final boolean specifiedSnapshot;
 
     @JsonCreator
     public IcebergTableHandle(
             @JsonProperty("schemaName") String schemaName,
             @JsonProperty("tableName") String tableName,
             @JsonProperty("tableType") TableType tableType,
             @JsonProperty("snapshotId") Optional<Long> snapshotId,
+            @JsonProperty("specifiedSnapshot") boolean specifiedSnapshot,
```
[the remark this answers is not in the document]
hantangwangd: Currently logic in IcebergAbstractMetadata.getTableHandle(...) is that, if there is a specified snapshot, we will use it, or we will pick up the newest snapshot, so the snapshot will always exists here weather or not provided a specified one.

So if we want to do what you say, we need firstly modify the semantic logic of IcebergAbstractMetadata.getTableHandle(...), as well as all other codes that rely on this semantic logic, for example in IcebergSplitManager. I wonder whether it is properly to do the job in this PR, or we should do it in a separately PR, what's your opinion?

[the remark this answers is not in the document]
ZacBlanco: I see. I was under the impression this was populated only if the user specified it. Hence being `Optional`. I am ok to punt that to a different PR.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergPlanOptimizer.java
```diff
@@ -132,6 +151,41 @@ public PlanNode visitFilter(FilterNode filter, RewriteContext<Void> context)
...
             }
+
+            if (TRUE_CONSTANT.equals(decomposedFilter.getRemainingExpression()) && simplifiedColumnDomain.equals(entireColumnDomain)) {
+                Set<Integer> predicateColumnIds = simplifiedColumnDomain.getDomains().get().keySet().stream()
+                        .map(IcebergColumnHandle::getId)
+                        .collect(toImmutableSet());
+
+                IcebergTableHandle tableHandle = (IcebergTableHandle) handle.getConnectorHandle();
+                IcebergAbstractMetadata metadata = (IcebergAbstractMetadata) transactionManager.get(handle.getTransaction());
+                Table icebergTable = IcebergUtil.getIcebergTable(metadata, session, tableHandle.getSchemaTableName());
```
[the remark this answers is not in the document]
hantangwangd: Fixed.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
```diff
@@ -432,6 +440,107 @@ public Optional<SystemTable> getSystemTable(ConnectorSession session, SchemaTabl
...
+        Table icebergTable;
+        try {
+            icebergTable = getIcebergTable(session, handle.getSchemaTableName());
+        }
+        catch (Exception e) {
+            throw new TableNotFoundException(handle.getSchemaTableName());
+        }
+
+        TupleDomain<ColumnHandle> domainPredicate = layoutHandle.getTupleDomain();
+        if (domainPredicate.isAll()) {
```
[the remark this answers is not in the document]
hantangwangd: Thanks for the suggestion, I have fixed it.

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -511,6 +511,25 @@ dropping the table from the metadata catalog using ``TRUNCATE TABLE``.
...
+
+     DELETE FROM lineitem WHERE linenumber = 1;
+
+     DELETE FROM lineitem WHERE linenumber not in (1, 3, 5, 7) and linestatus in ('O', 'F');
+
+.. note::
+
+    Columns in the filter must all be identity transformed partition columns of the target table.
+
+    Do not support non-comparison operator for the filter columns, for example ADD, SUBTRACT, MODULUS, etc.
```
ZacBlanco, line 529:
```suggestion
    Filtered columns only support comparison operators such as EQUALS, LESS THAN, LESS THAN EQUALS, etc.
```

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -511,6 +511,25 @@ dropping the table from the metadata catalog using ``TRUNCATE TABLE``.
...
+
+    Disallow to specify snapshot for target table. Always delete on currently newest snapshot.
```
ZacBlanco, line 531:
```suggestion
    Snapshots may not be specified on the target table. Deletes must only occur on the latest snapshot.
```

ZacBlanco (approved):
Two minor grammatical suggestions on the docs, but @steveburnett might have better ones. Otherwise LGTM!

hantangwangd:
> Two minor grammatical suggestions on the docs, but @steveburnett might have better ones. Otherwise LGTM!

@ZacBlanco Thanks a lot, I will squash them into one commit later.

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -511,6 +511,25 @@ dropping the table from the metadata catalog using ``TRUNCATE TABLE``.
     -----------+------+-----------+---------
     (0 rows)
 
+DELETE
+^^^^^^^^
+
+The iceberg connector can delete data in one or more entire partitions from tables by using ``DELETE FROM TABLE``::
```
steveburnett, line 517:
```suggestion
The iceberg connector can delete data in one or more entire partitions from tables by using ``DELETE FROM``::
```

``DELETE FROM TABLE`` implies that those three words are required. Also consider adding a following sentence introducing the example and the example table name lineitem, something like 

"For example, to delete from the table ``lineitem``:  "

presto-docs/src/main/sphinx/connector/iceberg.rst
steveburnett: Latin such as etc., e.g, i.e, should be avoided. For an example of a standard for this, see the [GitLab recommended word list entry for etc.](https://docs.gitlab.com/ee/development/documentation/styleguide/word_list.html#etc).  To address this adapting @ZacBlanco's suggestion, consider something like this. 

```suggestion
    Filtered columns only support comparison operators, such as EQUALS, LESS THAN, or LESS THAN EQUALS.
```

steveburnett (changes requested):
Nice work! I made a couple of suggestions for consistency and clarity.

hantangwangd:
> Nice work! I made a couple of suggestions for consistency and clarity.

@steveburnett Thanks a lot. I have update the docs following your suggestion, and squash them into one commit. Please take a look.

steveburnett (approved):
LGTM! (docs) 

Updated my local branch just now and built the docs locally to pick up and review the latest changes as they will be seen in production, everything looks great.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergTableHandle.java
```diff
@@ -33,19 +33,22 @@ public class IcebergTableHandle
     private final TableType tableType;
     private final Optional<Long> snapshotId;
     private final TupleDomain<IcebergColumnHandle> predicate;
+    private final boolean specifiedSnapshot;
```
ChunxuTang, line 36: Not sure whether this is a good name for a boolean variable. `specifiedSnapshot` sounds like a noun. Do you mean `snapshotSpecificed`?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
```diff
@@ -450,17 +554,4 @@ private long removeScanFiles(Table icebergTable, Iterable<FileScanTask> scan)
...
-    {
-        IcebergTableHandle handle = (IcebergTableHandle) tableHandle;
-        Table icebergTable = getIcebergTable(session, handle.getSchemaTableName());
-        try (CloseableIterable<FileScanTask> files = icebergTable.newScan().planFiles()) {
-            removeScanFiles(icebergTable, files);
-        }
-        catch (IOException e) {
-            throw new PrestoException(GENERIC_INTERNAL_ERROR, "failed to scan files for delete", e);
-        }
-    }
```
ChunxuTang, line 465: I noticed that in the new changes, you still have a `truncateTable` function. Are there any differences between the deleted and the new `truncateTable`?

presto-main/src/main/java/com/facebook/presto/util/GraphvizPrinter.java
```diff
@@ -50,6 +50,7 @@
 import com.facebook.presto.sql.planner.plan.JoinNode;
 import com.facebook.presto.sql.planner.plan.LateralJoinNode;
 import com.facebook.presto.sql.planner.plan.MergeJoinNode;
+import com.facebook.presto.sql.planner.plan.MetadataDeleteNode;
```
ChunxuTang, line 53: May I know the purpose of the changes in this file? Looks like for now the changes are unused. Are you preparing any changes on the UI side?

ChunxuTang (commented):
@hantangwangd Thanks for your work! This is a nice feature that would benefit the connector users.
Just left some minor comments.

ChunxuTang:
@hantangwangd BTW, do you want to add a release note?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergTableHandle.java
hantangwangd: Indeed, yes. I will fix it.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
hantangwangd: They are exactly the same. The purpose of moving `truncateTable` function up is just to leave the private function `removeScanFiles` at the bottom.

presto-main/src/main/java/com/facebook/presto/util/GraphvizPrinter.java
hantangwangd: When query state changes, QueryMonitor would invoke method `createGraphvizQueryPlan` to get a 'graphvizPlan' for queryMetadata. If we do not implement the method for MetadataDeleteNode here, it will always print error message like '...Node com.facebook.presto.sql.planner.plan.MetadataDeleteNode does not have a Graphviz visitor...' when we try to execute metadata delete.

hantangwangd:
> @hantangwangd BTW, do you want to add a release note?

@ChunxuTang Sure, release note have been added.

presto-main/src/main/java/com/facebook/presto/util/GraphvizPrinter.java
ChunxuTang: Thanks for the clarification.

ChunxuTang (approved):
LGTM. @hantangwangd Thanks for the work again.

@ChunxuTang It's my pleasure.

ChunxuTang:
Re-running the failed test.

hantangwangd:
> LGTM. @hantangwangd Thanks for the work again.

@ChunxuTang It's my pleasure.

PR #19904, opened 2023-06-16
Extend payload join optimization for cases without map columns; fix b…

mlyublena:
A collection of bug fixes for the payload join optimization, including the following:
+ fixed cases where not all collected join keys belong to the left-most table
+ abort optimization if the rewrite hides required columns
+ abort rewrite for operators other than Select/Project/Scan/Join. This fixes wrong result bugs when the query contains aggregates and other operators that may change the cardinality of intermediate levels
+ remove restriction for the fact table to have map or array columns. In the future we'll incorporate HBO to make a cost-based decision when to apply this optimization

PR #21050, opened 2023-10-05
Remove redundant cast to varchar in join clause

feilong-liu:
## Description
Rewrite query such as `select select * from orders o join customer c on cast(o.custkey as varchar) = cast(c.custkey as varchar)` to `select select * from orders o join customer c on o.custkey = c.custkey`

## Motivation and Context
Varchar is expensive to hash, and join with varchar keys can be expensive. For the cases above, removing the cast can reduce the cost of join operation.

## Impact
Test in production queries show it decrease both CPU and latency by half.

## Test Plan
Unit tests and [verifier run](https://www.internalfb.com/intern/presto/verifier/results/?test_id=141530)

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Add an optimization, which removes the redundant cast to varchar expression in join condition. The optimizer is controlled by session property `remove_redundant_cast_to_varchar_in_join` and enabled by default.
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java
@@ -283,6 +283,8 @@ public class FeaturesConfig
     private boolean preProcessMetadataCalls;
     private boolean useHBOForScaledWriters;
 
+    private boolean removeRedundantCastToVarcharInJoin = true;
+
     public enum PartitioningPrecisionStrategy
     {
         // Let Presto decide when to repartition
@@ -2814,6 +2816,19 @@ public class FeaturesConfig
     public FeaturesConfig setUseHBOForScaledWriters(boolean useHBOForScaledWriters)
     {
         this.useHBOForScaledWriters = useHBOForScaledWriters;
+        return this;
+    }
+
+    public boolean isRemoveRedundantCastToVarcharInJoin()
+    {
+        return removeRedundantCastToVarcharInJoin;
+    }
+
+    @Config("optimizer.remove-redundant-cast-to-varchar-in-join")
+    @ConfigDescription("If both left and right side of join clause are varchar cast from int/bigint, remove the cast")
+    public FeaturesConfig setRemoveRedundantCastToVarcharInJoin(boolean removeRedundantCastToVarcharInJoin)
+    {
+        this.removeRedundantCastToVarcharInJoin = removeRedundantCastToVarcharInJoin;
         return this;
     }
 }
--- a/presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/RemoveRedundantCastToVarcharInJoinClause.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/RemoveRedundantCastToVarcharInJoinClause.java
@@ -1,0 +1,170 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.sql.planner.iterative.rule;
+
+import com.facebook.presto.Session;
+import com.facebook.presto.common.type.Type;
+import com.facebook.presto.common.type.VarcharType;
+import com.facebook.presto.matching.Captures;
+import com.facebook.presto.matching.Pattern;
+import com.facebook.presto.metadata.FunctionAndTypeManager;
+import com.facebook.presto.spi.plan.PlanNode;
+import com.facebook.presto.spi.plan.ProjectNode;
+import com.facebook.presto.spi.relation.CallExpression;
+import com.facebook.presto.spi.relation.RowExpression;
+import com.facebook.presto.spi.relation.VariableReferenceExpression;
+import com.facebook.presto.sql.planner.iterative.Rule;
+import com.facebook.presto.sql.planner.plan.JoinNode;
+import com.facebook.presto.sql.relational.FunctionResolution;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+import static com.facebook.presto.SystemSessionProperties.isRemoveRedundantCastToVarcharInJoinEnabled;
+import static com.facebook.presto.common.type.BigintType.BIGINT;
+import static com.facebook.presto.common.type.IntegerType.INTEGER;
+import static com.facebook.presto.sql.planner.PlannerUtils.addProjections;
+import static com.facebook.presto.sql.planner.plan.Patterns.join;
+import static com.facebook.presto.sql.relational.Expressions.castToBigInt;
+
+/**
+ * Remove redundant cast to varchar in join condition for queries like `select select * from orders o join customer c on cast(o.custkey as varchar) = cast(c.custkey as varchar)`
+ * Transform from
+ * <pre>
+ *     - Join
+ *          left_cast = right_cast
+ *          - Project
+ *              left_cast := cast(lkey as varchar)
+ *              - TableScan
+ *                  lkey BIGINT
+ *          - Project
+ *              right_cast := cast(rkey as varchar)
+ *              - TableScan
+ *                  rkey BIGINT
+ *
+ * </pre>
+ * into
+ * <pre>
+ *     - Join
+ *          new_lkey = new_rkey
+ *          - Project
+ *              left_cast := cast(lkey as varchar)
+ *              new_lkey := lkey
+ *              - TableScan
+ *                  lkey BIGINT
+ *          - Project
+ *              right_cast := cast(rkey as varchar)
+ *              new_rkey := rkey
+ *              - TableScan
+ *                  rkey BIGINT
+ * </pre>
+ * We will rely on optimizations later to remove unnecessary cast (if not used) and identity projection here.
+ * <p>
+ * Notice that we do not apply similar optimizations to queries with similar join condition like `cast(bigint as varchar) = varchar`. In general it can be converted to
+ * `bigint = try_cast(varchar as bigint)` as if the varchar here cannot be converted to bigint, try_cast will return null and will not match anyway. However, a special case is
+ * varchar begins with 0. `select cast(92 as varchar) = '092'` is false, but `select 92 = try_cast('092' as bigint)` returns true.
+ */
+public class RemoveRedundantCastToVarcharInJoinClause
+        implements Rule<JoinNode>
+{
+    private static final List<Type> TYPE_SUPPORTED = ImmutableList.of(INTEGER, BIGINT);
+    private final FunctionAndTypeManager functionAndTypeManager;
+    private final FunctionResolution functionResolution;
+
+    public RemoveRedundantCastToVarcharInJoinClause(FunctionAndTypeManager functionAndTypeManager)
+    {
+        this.functionAndTypeManager = functionAndTypeManager;
+        this.functionResolution = new FunctionResolution(functionAndTypeManager.getFunctionAndTypeResolver());
+    }
+
+    @Override
+    public boolean isEnabled(Session session)
+    {
+        return isRemoveRedundantCastToVarcharInJoinEnabled(session);
+    }
+
+    @Override
+    public Pattern<JoinNode> getPattern()
+    {
+        return join();
+    }
+
+    @Override
+    public Result apply(JoinNode node, Captures captures, Context context)
+    {
+        PlanNode leftInput = context.getLookup().resolve(node.getLeft());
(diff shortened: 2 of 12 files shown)


presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestDistributedQueries.java
```diff
@@ -1253,7 +1257,7 @@ public void testPayloadJoins()
                 "SELECT l.* FROM (select orderkey+1 as ok1, * from lineitem_map) l left join orders o on (l.ok1 = o.orderkey+1) left join part p on (l.partkey=p.partkey)",
                 "SELECT l.* FROM (select *, cast(orderkey as int) as ok from lineitem_map) l left join (select *, cast(orderkey as int) as ok from orders) o on (l.ok = o.ok) left join (select *, cast(partkey as int) as pk from part) p on (l.ok=p.pk)",
                 "with lm as (select quantity + 1 as q1, * from lineitem_map) SELECT l.* FROM (select * from lm) l left join orders o on (l.orderkey = o.orderkey) left join part p on (l.partkey=p.partkey)",
-
+                // This is the query which fail when REMOVE_REDUNDANT_CAST_TO_VARCHAR_IN_JOIN set to true
                 "SELECT l.* FROM lineitem_map l left join orders o on (cast(l.orderkey as varchar) = cast(o.orderkey as varchar)) left join part p on (l.partkey=p.partkey)",
```
feilong-liu, line 1261: @mlyublena The optimization added surfaced a bug in payload join optimizer, i.e. when payload join optimization is not enabled, this query runs successfully with the optimization in this PR, but when payload join optimization is enabled, this query will fail with the optimization in this PR.

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java
```diff
@@ -7187,4 +7188,25 @@ public void testRewriteContainsToIn()
...
+        assertQuery(session, "select *, cast(o.custkey as varchar), cast(c.custkey as varchar) from orders o join customer c on cast(o.custkey as varchar) = cast(c.custkey as varchar)");
+        // Do not trigger optimization
+        assertQuery(session, "select * from customer c join orders o on cast(acctbal as varchar) = cast(totalprice as varchar)");
+
+        Session disabled = Session.builder(getSession())
+                .setSystemProperty(REMOVE_REDUNDANT_CAST_TO_VARCHAR_IN_JOIN, "false")
+                .build();
+        assertUpdate("create table lineitem_map as select *, map(ARRAY[1,3], ARRAY[2,4]) as m1, map(ARRAY[1,3], ARRAY[2,4]) as m2 from lineitem", 60175);
+        assertUpdate("create table part_map as select *, map(ARRAY[1,3], ARRAY[2,4]) as m3, map(ARRAY[1,3], ARRAY[2,4]) as m4 from part", 2000);
+        assertQueryWithSameQueryRunner(session, "SELECT l.* FROM lineitem_map l left join orders o on (cast(l.orderkey as varchar) = cast(o.orderkey as varchar)) left join part p on (l.partkey=p.partkey)", disabled);
```
feilong-liu, line 7210: Add the query found in payload join test here, make sure that it succeeds with this optimization, with payload join disabled.

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestDistributedQueries.java
mlyublena: There is an outstanding PR that fixes a lot of bugs in payload join optimizer: https://github.com/prestodb/presto/pull/19904 
My guess is that the case you found is already fixes there

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/RemoveRedundantCastToVarcharInJoinClause.java
```diff
@@ -0,0 +1,179 @@
...
+        newRightAssignmentsBuilder.putAll(rightProject.getAssignments().getMap());
+        Map<VariableReferenceExpression, RowExpression> newRightAssignments = newRightAssignmentsBuilder.build();
+
+        PlanNode newLeftProject = addProjections(leftProject.getSource(), context.getIdAllocator(), newLeftAssignments);
+        PlanNode newRightProject = addProjections(rightProject.getSource(), context.getIdAllocator(), newRightAssignments);
+
+        return Result.ofPlanNode(new JoinNode(node.getSourceLocation(), context.getIdAllocator().getNextId(), node.getType(), newLeftProject, newRightProject, joinClauseBuilder.build(), node.getOutputVariables(), node.getFilter(), Optional.empty(), Optional.empty(), node.getDistributionType(), node.getDynamicFilters()));
+    }
+
+    private RowExpression castToBigInt(RowExpression rowExpression)
```
pranjalssh, line 163: This can be put as a util in Expressions.java

pranjalssh (approved):

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/RemoveRedundantCastToVarcharInJoinClause.java
```diff
@@ -0,0 +1,179 @@
...
+ *                  lkey BIGINT
+ *          - Project
+ *              right_cast := cast(rkey as varchar)
+ *              new_rkey := rkey
+ *              - TableScan
+ *                  rkey BIGINT
+ * </pre>
+ * We will rely on optimizations later to remove unnecessary cast (if not used) and identity projection here.
+ *
+ * Notice that we do not apply similar optimizations to queries with similar join condition like `cast(bigint as varchar) = varchar`. In general it can be converted to
```
mlyublena, line 78: should we allow this transformation for all non-lossy/deterministic casts?

presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java
```diff
@@ -282,6 +282,8 @@ public class FeaturesConfig
 
     private boolean preProcessMetadataCalls;
 
+    private boolean removeRedundantCastToVarcharInJoin = true;
```
mlyublena, line 285: why do we need a session property? isn't this always beneficial?

feilong-liu: Just in case we need to revert it if any bugs found later.

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/RemoveRedundantCastToVarcharInJoinClause.java
feilong-liu: I applied for cast(bigint/integer as varchar) as varchar has high cost for hash operation. Other cases may not show as big improvement here.

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/RemoveRedundantCastToVarcharInJoinClause.java
```diff
@@ -0,0 +1,170 @@
...
+        ProjectNode rightProject = (ProjectNode) rightInput;
+
+        ImmutableList.Builder<JoinNode.EquiJoinClause> joinClauseBuilder = ImmutableList.builder();
+        ImmutableMap.Builder<VariableReferenceExpression, RowExpression> newLeftAssignmentsBuilder = ImmutableMap.builder();
+        ImmutableMap.Builder<VariableReferenceExpression, RowExpression> newRightAssignmentsBuilder = ImmutableMap.builder();
+        boolean isChanged = false;
+        for (JoinNode.EquiJoinClause equiJoinClause : node.getCriteria()) {
+            RowExpression leftProjectAssignment = leftProject.getAssignments().getMap().get(equiJoinClause.getLeft());
+            RowExpression rightProjectAssignment = rightProject.getAssignments().getMap().get(equiJoinClause.getRight());
+            if (!isValidCast(leftProjectAssignment) || !isValidCast(rightProjectAssignment)) {
```
mlyublena, line 124: perhaps rename to isSupportedCast

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/RemoveRedundantCastToVarcharInJoinClause.java
```diff
@@ -0,0 +1,170 @@
...
+    {
+        return join();
+    }
+
+    @Override
+    public Result apply(JoinNode node, Captures captures, Context context)
+    {
+        PlanNode leftInput = context.getLookup().resolve(node.getLeft());
+        PlanNode rightInput = context.getLookup().resolve(node.getRight());
+        if (!(leftInput instanceof ProjectNode) || !(rightInput instanceof ProjectNode)) {
```
mlyublena, line 111: so that transformation won't work if the projection happens some levels down the tree?
for example
```
(select * from (select *, cast(key as varchar) as newkey) where ...) join T on newkey=cast(... as varchar)
```

presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java
```diff
@@ -2816,4 +2818,17 @@ public FeaturesConfig setUseHBOForScaledWriters(boolean useHBOForScaledWriters)
         this.useHBOForScaledWriters = useHBOForScaledWriters;
         return this;
     }
+
+    public boolean isRemoveRedundantCastToVarcharInJoinEnabled()
```
mlyublena, line 2822: nit: do we need "Enabled" in the name?

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/RemoveRedundantCastToVarcharInJoinClause.java
feilong-liu: Correct, it will not work. I tried to do it, but it will complicated this rule as we not only need to extract this pattern traversing the tree, but also need to propagate the variables used in the cast expression (it not propagated already) to the join node. Since this rule already covered the query which motivated this optimization, I decided to do it in a simplified way for now.

presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java
```diff
@@ -2816,4 +2818,17 @@ public FeaturesConfig setUseHBOForScaledWriters(boolean useHBOForScaledWriters)
...
+    {
+        return removeRedundantCastToVarcharInJoin;
+    }
+
+    @Config("optimizer.remove-redundant-cast-to-varchar-in-join")
+    @ConfigDescription("If both left and right side of join clause are varchar cast from int/bigint, remove the cast")
+    public FeaturesConfig setRemoveRedundantCastToVarcharInJoinEnabled(boolean removeRedundantCastToVarcharInJoin)
```
mlyublena, line 2829: same here: remove "Enabled" from name?

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/RemoveRedundantCastToVarcharInJoinClause.java
```diff
@@ -0,0 +1,170 @@
...
+            if (!isSupportedCast(leftProjectAssignment) || !isSupportedCast(rightProjectAssignment)) {
+                joinClauseBuilder.add(equiJoinClause);
+                continue;
+            }
+
+            RowExpression leftAssignment = ((CallExpression) leftProjectAssignment).getArguments().get(0);
+            RowExpression rightAssignment = ((CallExpression) rightProjectAssignment).getArguments().get(0);
+
+            if (!leftAssignment.getType().equals(rightAssignment.getType())) {
+                leftAssignment = castToBigInt(functionAndTypeManager, leftAssignment);
```
mlyublena, line 133: we don't need to do this if the expressions are already bigints

mlyublena (approved):
LGTM

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/RemoveRedundantCastToVarcharInJoinClause.java
feilong-liu: This is checked in `castToBigInt` function, with
```
        if (rowExpression.getType().equals(BIGINT)) {
            return rowExpression;
        }
```

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java
```diff
@@ -7187,4 +7188,18 @@ public void testRewriteContainsToIn()
...
     }
+
+    @Test
+    public void testRemoveRedundantCastToVarcharInJoinClause()
+    {
+        Session session = Session.builder(getSession())
+                .setSystemProperty(REMOVE_REDUNDANT_CAST_TO_VARCHAR_IN_JOIN, "true")
+                .build();
+        // Trigger optimization
+        assertQuery(session, "select * from orders o join customer c on cast(o.custkey as varchar) = cast(c.custkey as varchar)");
```
mlyublena, line 7199: can you add a case where there is a cross join on top of the join you're optimizing? I think your optimization covers it, but cross join can be sensitive if new columns are introduced or reordered in one of the children, so it's good to add a test to make sure there are no regressions
something like this:
```
select * from 
(select * from orders o join customer c on cast(o.custkey as varchar) = cast(c.custkey as varchar)) r, nation;
```

PR #21051, opened 2023-10-06
[native pos] Trigger a retry without broadcast if broadcast size exceeds the limit

singcha:
## Description
Add functionality to : 
1. Return broadcast size and number of rows written
2. Fail fast on driver with broadcast OOM when broadcast size exceeds configured broadcast limit

## Motivation and Context
This change enables driver to fail fast when broadcast table size exceeds configured broadcast limit. It also enables auto-retry(If config is enabled) with broadcast disabled.

## Test Plan
Tested with newly added test

== NO RELEASE NOTE ==

### DIFF
--- a/presto-native-execution/presto_cpp/main/operators/BroadcastFactory.cpp
+++ b/presto-native-execution/presto_cpp/main/operators/BroadcastFactory.cpp
@@ -73,6 +73,7 @@ BroadcastFileWriter::BroadcastFileWriter(
     : fileSystem_(std::move(fileSystem)),
       filename_(filename),
       numRows_(0),
+      maxSerializedSize_(0),
       pool_(pool),
       serde_(std::make_unique<serializer::presto::PrestoVectorSerde>()),
       inputType_(inputType) {}
@@ -83,21 +84,32 @@ void BroadcastFileWriter::collect(const RowVectorPtr& input) {
 
 void BroadcastFileWriter::noMoreData() {}
 
-// TODO: Add file stats - size, checksum, number of rows.
 RowVectorPtr BroadcastFileWriter::fileStats() {
   // No rows written.
   if (numRows_ == 0) {
     return nullptr;
   }
 
-  auto data = BaseVector::create<FlatVector<StringView>>(VARCHAR(), 1, pool_);
-  data->set(0, StringView(filename_));
+  auto fileNameVector =
+      BaseVector::create<FlatVector<StringView>>(VARCHAR(), 1, pool_);
+  fileNameVector->set(0, StringView(filename_));
+  auto maxSerializedSizeVector =
+      BaseVector::create<FlatVector<int64_t>>(BIGINT(), 1, pool_);
+  maxSerializedSizeVector->set(0, maxSerializedSize_);
+  auto numRowsVector =
+      BaseVector::create<FlatVector<int64_t>>(BIGINT(), 1, pool_);
+  numRowsVector->set(0, numRows_);
+
   return std::make_shared<RowVector>(
       pool_,
-      ROW({"filepath"}, {VARCHAR()}),
+      ROW({"filepath", "maxserializedsize", "numrows"},
+          {VARCHAR(), BIGINT(), BIGINT()}),
       nullptr,
       1,
-      std::vector<VectorPtr>({std::move(data)}));
+      std::vector<VectorPtr>(
+          {std::move(fileNameVector),
+           std::move(maxSerializedSizeVector),
+           std::move(numRowsVector)}));
 }
 
 void BroadcastFileWriter::initializeWriteFile() {
@@ -121,6 +133,7 @@ void BroadcastFileWriter::write(const RowVectorPtr& rowVector) {
   auto serializer = serde_->createSerializer(inputType_, numRows, arena.get());
 
   serializer->append(rowVector, folly::Range(&allRows, 1));
+  maxSerializedSize_ += serializer->maxSerializedSize();
   IOBufOutputStream out(*pool_);
   serializer->flush(&out);
   auto iobuf = out.getIOBuf();
--- a/presto-native-execution/src/test/java/com/facebook/presto/spark/TestPrestoSparkNativeSimpleQueries.java
+++ b/presto-native-execution/src/test/java/com/facebook/presto/spark/TestPrestoSparkNativeSimpleQueries.java
@@ -19,6 +19,7 @@ import com.facebook.presto.testing.QueryRunner;
 import com.facebook.presto.tests.AbstractTestQueryFramework;
 import org.testng.annotations.Test;
 
+import static com.facebook.presto.SystemSessionProperties.JOIN_DISTRIBUTION_TYPE;
 import static com.facebook.presto.nativeworker.NativeQueryRunnerUtils.createBucketedCustomer;
 import static com.facebook.presto.nativeworker.NativeQueryRunnerUtils.createBucketedLineitemAndOrders;
 import static com.facebook.presto.nativeworker.NativeQueryRunnerUtils.createCustomer;
@@ -34,6 +35,8 @@ import static com.facebook.presto.nativeworker.NativeQueryRunnerUtils.createPart
 import static com.facebook.presto.nativeworker.NativeQueryRunnerUtils.createPrestoBenchTables;
 import static com.facebook.presto.nativeworker.NativeQueryRunnerUtils.createRegion;
 import static com.facebook.presto.nativeworker.NativeQueryRunnerUtils.createSupplier;
+import static com.facebook.presto.spark.PrestoSparkSessionProperties.SPARK_BROADCAST_JOIN_MAX_MEMORY_OVERRIDE;
+import static com.facebook.presto.spark.PrestoSparkSessionProperties.SPARK_RETRY_ON_OUT_OF_MEMORY_BROADCAST_JOIN_ENABLED;
 
 public class TestPrestoSparkNativeSimpleQueries
         extends AbstractTestQueryFramework
@@ -165,5 +168,35 @@ public class TestPrestoSparkNativeSimpleQueries
         finally {
             getQueryRunner().execute("DROP TABLE IF EXISTS avg_partial_states");
         }
+    }
+
+    @Test
+    public void testRetryOnOutOfMemoryBroadcastJoin()
+    {
+        String query = "select l.orderkey from lineitem l join orders o on l.orderkey = o.orderkey ";
+
+        Session session = getSessionWithBroadcastJoinDistribution("10B", false);
+        // Query should fail with broadcast join OOM & retry disabled.
+        assertQueryFails(
+                session,
+                query,
+                "Query exceeded per-node broadcast memory limit of 10B \\[Max serialized broadcast size: .*kB\\]");
+
+        Session expectedSession = Session.builder(getSession())
+                .setSystemProperty(JOIN_DISTRIBUTION_TYPE, "BROADCAST")
+                .build();
+        Session actualSession = getSessionWithBroadcastJoinDistribution("10B", true);
+
+        // Query should succeed with broadcast join OOM & retry enabled.
+        assertQuery(actualSession, query, expectedSession, query);
+    }
+
+    private Session getSessionWithBroadcastJoinDistribution(String broadcastJoinMaxMemory, Boolean retryOnBroadcastOutOfMemory)
+    {
+        return Session.builder(getSession())
+                .setSystemProperty(JOIN_DISTRIBUTION_TYPE, "BROADCAST")
+                .setSystemProperty(SPARK_BROADCAST_JOIN_MAX_MEMORY_OVERRIDE, broadcastJoinMaxMemory)
+                .setSystemProperty(SPARK_RETRY_ON_OUT_OF_MEMORY_BROADCAST_JOIN_ENABLED, Boolean.toString(retryOnBroadcastOutOfMemory))
+                .build();
     }
 }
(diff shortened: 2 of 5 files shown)


presto-native-execution/presto_cpp/main/operators/BroadcastFactory.cpp
```diff
@@ -115,6 +127,7 @@ void BroadcastFileWriter::write(const RowVectorPtr& rowVector) {
 
   initializeWriteFile();
   numRows_ += numRows;
+  retainedSize_ += rowVector->retainedSize();
```
mbasmanova, line 130: RetainedSize depends on vector encoding. I think we should use estimateFlatSize instead.

This commits add both size and number of rows. Please, update commit message to reflect that. Also, make sure that all lines are within 78 chars in the commit message.

presto-spark-base/src/main/java/com/facebook/presto/spark/PrestoSparkMemoryBasedBroadcastDependency.java
```diff
@@ -81,6 +85,32 @@ public Broadcast<List<PrestoSparkSerializedPage>> executeBroadcast(JavaSparkCont
         return broadcastVariable;
     }
 
+    private long getNativeInMemoryBytes(List<PrestoSparkSerializedPage> prestoSparkSerializedPages, PagesSerde pagesSerde)
+    {
+        return prestoSparkSerializedPages.stream()
+                .map(prestoSparkSerializedPage -> PrestoSparkUtils.toSerializedPage(prestoSparkSerializedPage))
+                .map(serializedPage -> pagesSerde.deserialize(serializedPage))
+                // Extract fileSize.
```
mbasmanova, line 93: Is this file size or sum of "retainedSize" of the vectors? Does this code run on the producer or consumer?

mbasmanova (commented):
Please, update commit message for clarity. Please, explain whether the error happens in the task that writes broadcast data or the task that reads broadcast data. Please, explain what does "too large" means and what happens as a result of this failure.

```
[native pos] Fail fast if broadcast data is too large
```

presto-spark-base/src/main/java/com/facebook/presto/spark/PrestoSparkMemoryBasedBroadcastDependency.java
singcha: Its sum of "retainedSize". will update the comment to reflect "estimateFlatSize".
This code runs on driver. We can say its consumer but actual consumers are executors/native process running on executors.

presto-native-execution/presto_cpp/main/operators/BroadcastFactory.cpp
singcha: As discussed offline, updated code with maxSerializedSize.

presto-native-execution/presto_cpp/main/operators/BroadcastFactory.cpp
```diff
@@ -83,21 +84,32 @@ void BroadcastFileWriter::collect(const RowVectorPtr& input) {
...
   if (numRows_ == 0) {
     return nullptr;
   }
 
-  auto data = BaseVector::create<FlatVector<StringView>>(VARCHAR(), 1, pool_);
-  data->set(0, StringView(filename_));
+  auto fileNameVector =
+      BaseVector::create<FlatVector<StringView>>(VARCHAR(), 1, pool_);
+  fileNameVector->set(0, StringView(filename_));
+  auto retainedSizeVector =
```
mbasmanova, line 96: > retainedSizeVector

name needs updating

singcha:
@mbasmanova - Thanks Masha for review! I have addressed the comments. Can you please review it again!

mbasmanova (commented):
@singcha Commit message needs updating as it still refers to "retained" size. I also suggest to add more details for the benefit of future readers.

```
[native pos] Report an estimate of broadcast size and row count to the driver

Report an exact number of rows broadcasted along with an estimate of broadcast
size in bytes. The estimate is based on serialized data size and roughly
matches the size of the broadcast file. This size may be different from the
size of the hash table that will be built from the broadcast data on the
reducer.

The estimate of the broadcast size will be used to check whether total size of
broadcast data is within a configured limit. If not, the map stage will be
retried with broadcast disabled.
```

mbasmanova (commented):
Commit message needs updating for clarity.

```
[native pos] Fail fast on the driver if broadcast size exceeds the limit

Check the sum of broadcast size estimates from all workers against a limit and throw
 ExceededMemoryLimit if limit is exceeded.

This check runs once after all tasks have completed.

If the limit is exceed, the query will be retried without broadcast.

Depends on session property spark_retry_on_out_of_memory_broadcast_join_enabled.
```


> This check runs once after all tasks have completed.

I wonder why not run the check multiple times. Each time another task complete, check and fail in-progress tasks fast.

mbasmanova (commented):
@singcha Where / how do we specify the limit?

presto-native-execution/src/test/java/com/facebook/presto/spark/TestPrestoSparkNativeSimpleQueries.java
```diff
@@ -166,4 +169,34 @@ public void testAggregationCompanionFunction()
...
+        // Query should fail with broadcast join OOM
+        assertQueryFails(
+                session,
+                query,
+                "Query exceeded per-node broadcast memory limit of 10B \\[Compressed broadcast size: .*kB\\]");
+
+        Session expectedSession = Session.builder(getSession())
+                .setSystemProperty(JOIN_DISTRIBUTION_TYPE, "BROADCAST")
+                .build();
+        Session actualSession = Session.builder(getSession())
```
mbasmanova, line 193: Consider extracting a helper method with a boolean parameter to create a session with broadcast limit and retry enabled/disabled to avoid copy-paste.

presto-spark-base/src/main/java/com/facebook/presto/spark/PrestoSparkMemoryBasedBroadcastDependency.java
```diff
@@ -81,6 +85,32 @@ public Broadcast<List<PrestoSparkSerializedPage>> executeBroadcast(JavaSparkCont
...
+                .map(prestoSparkSerializedPage -> PrestoSparkUtils.toSerializedPage(prestoSparkSerializedPage))
+                .map(serializedPage -> pagesSerde.deserialize(serializedPage))
+                // Extract sum of maxSerializedSize.
+                .flatMap(page -> IntStream.range(0, page.getPositionCount())
+                        .mapToObj(position -> BigintType.BIGINT.getObjectValue(null, page.getBlock(1), position)))
+                .mapToLong(Long.class::cast)
+                .sum();
+    }
+
+    private long getUncompressedBroadcastSizeInBytes(List<PrestoSparkSerializedPage> broadcastValue)
```
shrinidhijoshi, line 100: Nit: change parameter name to `prestoSparkSerializedPages` as `broadcastValue` has no meaning inside of this function

presto-spark-base/src/main/java/com/facebook/presto/spark/PrestoSparkMemoryBasedBroadcastDependency.java
```diff
@@ -60,19 +70,13 @@ public Broadcast<List<PrestoSparkSerializedPage>> executeBroadcast(JavaSparkCont
...
-        long compressedBroadcastSizeInBytes = broadcastValue.stream()
-                .mapToInt(page -> page.getBytes().length)
-                .sum();
-        long uncompressedBroadcastSizeInBytes = broadcastValue.stream()
-                .mapToInt(page -> page.getUncompressedSizeInBytes())
-                .sum();
-
         long maxBroadcastSizeInBytes = maxBroadcastSize.toBytes();
-
+        long compressedBroadcastSizeInBytes = isNativeExecution ? getNativeMaxSerializedBytes(broadcastValue, pagesSerde) : getCompressedBroadcastSizeInBytes(broadcastValue);
```
shrinidhijoshi, line 74: Is there a reason we care about compressed size check ? We should be able to rely only on uncompressed size check for OOMing ..

singcha:
> @singcha Where / how do we specify the limit?

Added in commit message. Its determined by session property: spark_broadcast_join_max_memory_override.
 If this property is not set, broadcast limit is: Min(query_max_broadcast_memory,
  query.max-broadcast-memory, query.max-memory-per-node)

presto-spark-base/src/main/java/com/facebook/presto/spark/PrestoSparkMemoryBasedBroadcastDependency.java
singcha: I just left the java behavior unchanged. At the same time, I can't see why it will ever be more than uncompressed size. I have updated the code to remove this check

presto-native-execution/presto_cpp/main/operators/BroadcastFactory.cpp
```diff
@@ -83,21 +84,32 @@ void BroadcastFileWriter::collect(const RowVectorPtr& input) {
...
   }
 
-  auto data = BaseVector::create<FlatVector<StringView>>(VARCHAR(), 1, pool_);
-  data->set(0, StringView(filename_));
+  auto fileNameVector =
+      BaseVector::create<FlatVector<StringView>>(VARCHAR(), 1, pool_);
+  fileNameVector->set(0, StringView(filename_));
+  auto maxSerializedSizeVector =
+      BaseVector::create<FlatVector<int64_t>>(BIGINT(), 1, pool_);
+  maxSerializedSizeVector->set(0, maxSerializedSize_);
```
pgupta2, line 98: QQ: Shouldn't we be reporting back deserializedSize instead? HT in memory will be stored in deserialized format. Right?

presto-spark-base/src/main/java/com/facebook/presto/spark/PrestoSparkMemoryBasedBroadcastDependency.java
```diff
@@ -40,17 +44,23 @@ public class PrestoSparkMemoryBasedBroadcastDependency
     private final long queryCompletionDeadline;
     private Broadcast<List<PrestoSparkSerializedPage>> broadcastVariable;
     private final Set<PrestoSparkServiceWaitTimeMetrics> waitTimeMetrics;
+    private final boolean isNativeExecution;
 
-    public PrestoSparkMemoryBasedBroadcastDependency(RddAndMore<PrestoSparkSerializedPage> broadcastDependency, DataSize maxBroadcastSize, long queryCompletionDeadline, Set<PrestoSparkServiceWaitTimeMetrics> waitTimeMetrics)
+    public PrestoSparkMemoryBasedBroadcastDependency(RddAndMore<PrestoSparkSerializedPage> broadcastDependency,
```
pgupta2, line 49: I am bit confused here. Shouldn't this code live inside `PrestoSparkStorageBasedBroadcastDependency`? `PrestoSparkMemoryBasedBroadcastDependency` is used when HT is directly collected on Driver side and broadcasted while StorageBasedBroadcastDependency is used when HT is written to storage and only FileHandles information is collected on driver.

pgupta2 (changes requested):

presto-native-execution/presto_cpp/main/operators/BroadcastFactory.cpp
singcha: Had a discussion with @mbasmanova. This is reasonable estimate to row container's size.

presto-spark-base/src/main/java/com/facebook/presto/spark/PrestoSparkMemoryBasedBroadcastDependency.java
singcha: The naming of this class assumes that when we get presto page, it will be in memory broadcast. While for Sapphire-Velox, we return broadcast info as presto pages.

singcha: I have  moved out logic to dedicated class PrestoSparkNativeStorageBasedBroadcastDepency

presto-spark-base/src/main/java/com/facebook/presto/spark/PrestoSparkNativeStorageBasedDependency.java
```diff
@@ -0,0 +1,103 @@
...
+    {
+        List<PrestoSparkSerializedPage> broadcastValue = broadcastDependency.collectAndDestroyDependenciesWithTimeout(computeNextTimeout(queryCompletionDeadline), MILLISECONDS, waitTimeMetrics).stream()
+                .map(Tuple2::_2)
+                .collect(toList());
+
+        // release memory retained by the RDD (splits and dependencies)
+        broadcastDependency = null;
+
+        long maxBroadcastSizeInBytes = maxBroadcastSize.toBytes();
+        long uncompressedBroadcastSizeInBytes = getNativeMaxSerializedBytes(broadcastValue, pagesSerde);
```
pgupta2, line 74: `uncompressedBroadcastSizeInBytes` -> `serializedBroadcastSizeInBytes`

pgupta2 (approved):

shrinidhijoshi (approved):
LGTM! Thanks for addressing the comments.

You still need committer's stamp for the `presto-native-execution` module changes. 
@amitkdutta  Can you please help here ?

amitkdutta (approved):
Accepting since already 2 accepts are there.

PR #21056, opened 2023-10-06
[native] Stop using PlanBuilder::finalAggregation with resultTypes API

mbasmanova:
It is not sufficient to specify resultTypes for final and intermediate aggregations. We need to specify raw input types instead.

https://github.com/facebookincubator/velox/issues/6506

PlanBuilder APIs that take result types are being replaced with APIs that take raw input types.

Remove the only usage of the API that's going away from TaskManagerTest.cpp

```
== NO RELEASE NOTE ==
```


xiaoxmeng (approved):
@mbasmanova LGTM. Thanks!

PR #21076, opened 2023-10-09
[native][circle] lower parallelism to avoid an oom issue

mshang816:
lower parallelism of native build to avoid an oom issue

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -113,7 +113,7 @@ jobs:
               -DPRESTO_ENABLE_JWT=ON \
               -DCMAKE_PREFIX_PATH=/usr/local \
               -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
-            ninja -C _build/debug -j 8
+            ninja -C _build/debug -j 7
             ccache -s
       - save_cache:
           name: "Save CCache cache"


.circleci/continue_config.yml
```diff
@@ -113,7 +113,7 @@ jobs:
               -DPRESTO_ENABLE_JWT=ON \
               -DCMAKE_PREFIX_PATH=/usr/local \
               -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
-            ninja -C _build/debug -j 8
+            ninja -C _build/debug -j 7
```
agrawaldevesh, line 116: Wait. Why just go from 8 to 7, if you want to reduce the likelyhood of OOM: Why not something like 4 or 2 for that matter ?

mshang816: > Wait. Why just go from 8 to 7, if you want to reduce the likelyhood of OOM: Why not something like 4 or 2 for that matter ?
@agrawaldevesh  We can do 4 or 2 but that would impact our build performance.

pranjalssh (approved):
Lets also add a comment so we can tweak in future as well.

.circleci/continue_config.yml
pranjalssh: 1/8th of 64GB is 8GB so can be substantial reduction in memory. Lets not hurt performance by reducing too much

PR #21080, opened 2023-10-09
[native][circleci] Add link targets to link_job_pool with a lower number of max concurrent jobs

mshang816:
We have several resource-intensive link target which will cause the ld get killed due to low memory. Instead of lowering the parallelism of the entire build (using -j) we put those jobs into a job pool (link_job_pool) a set a lower upper bound of max job. This will fix the previous build failures with "collect2: fatal error: ld terminated with signal 9 [Killed]”

Test Plans:
Make sure all circle ci runs are green

```
== NO RELEASE NOTE ==
```


spershin (approved):

pranjalssh (approved):

tdcmeehan:
It appears most `linux-build-and-unit-test` runs fail again after this change.

Issue #21086, opened 2023-10-10
HiveExternalWorkerQueryRunner fail to start with NullPointerException on DATA_DIR


yingsu00:
The following error was shown when starting HiveExternalWorkerQueryRunner in IntelliJ, with all required settings in https://github.com/prestodb/presto/tree/master/presto-native-execution#development

```
.../Applications/IntelliJ IDEA.app/Contents/lib/idea_rt.jar com.facebook.presto.nativeworker.HiveExternalWorkerQueryRunner /Users/yingsu
Connected to the target VM, address: '127.0.0.1:53995', transport: 'socket'
2023-10-09T23:00:38.372-0500	INFO	main	com.facebook.airlift.log.Logging	Logging to stderr
2023-10-09T23:02:37.831-0500	INFO	main	stderr	Exception in thread "main" 
2023-10-09T23:02:37.832-0500	INFO	main	stderr	java.lang.NullPointerException
2023-10-09T23:02:37.832-0500	INFO	main	stderr		at sun.nio.fs.UnixPath.normalizeAndCheck(UnixPath.java:77)
2023-10-09T23:02:37.832-0500	INFO	main	stderr		at sun.nio.fs.UnixPath.<init>(UnixPath.java:71)
2023-10-09T23:02:37.832-0500	INFO	main	stderr		at sun.nio.fs.UnixFileSystem.getPath(UnixFileSystem.java:280)
2023-10-09T23:02:37.832-0500	INFO	main	stderr		at java.nio.file.Paths.get(Paths.java:84)
2023-10-09T23:02:37.833-0500	INFO	main	stderr		at com.facebook.presto.nativeworker.PrestoNativeQueryRunnerUtils.createJavaQueryRunner(PrestoNativeQueryRunnerUtils.java:103)
2023-10-09T23:02:37.833-0500	INFO	main	stderr		at com.facebook.presto.nativeworker.PrestoNativeQueryRunnerUtils.createJavaQueryRunner(PrestoNativeQueryRunnerUtils.java:97)
2023-10-09T23:02:37.833-0500	INFO	main	stderr		at com.facebook.presto.nativeworker.HiveExternalWorkerQueryRunner.main(HiveExternalWorkerQueryRunner.java:32)
Disconnected from the target VM, address: '127.0.0.1:53995', transport: 'socket'

Process finished with exit code 1

```

## Your Environment

* Presto version used:
* Storage (HDFS/S3/GCS..):
* Data source and connector used:
[...]

PR #21087, opened 2023-10-10
Fix NullPointerException for HiveExternalWorkerQueryRunner

yingsu00:
Resolves https://github.com/prestodb/presto/issues/21086

## Description
The documentation https://github.com/prestodb/presto/tree/master/presto
-native-execution#development says the DATA_DIR needs to be set as an
environment variable, however, the dataDirectory in createJavaQueryRunner()
was assigned to System.getProperty("DATA_DIR"), which gets null and causes
NullPointerException. Furthermore, System.getProperty() and System.getenv()
were used randomly in the native worker query runners. This is confusing
and could cause more future issues. System.getProperty() reads the Java
system properties that are passed in using -Dkey=value, while System.getenv()
reads environment variables. They cannot be used interchangably and would
return null if the corresponding values are not set. While the documentation
uses environment variables, CircleCI jobs rely on system variables. In order
to make the query runners more robust, it's better to support both of them.
This commit introduces HiveTestUtils.getProperty() static method to read
both. If both system property and environment variable are not set, an
exception would be thrown. If both of them are set but the values are
different, an exception would also be thrown.

## Motivation and Context
HiveExternalWorkerQueryRunner was unable to start

## Impact
none

## Test Plan
HiveExternalWorkerQueryRunner was able start after change

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-hive/src/test/java/com/facebook/presto/hive/HiveTestUtils.java
+++ b/presto-hive/src/test/java/com/facebook/presto/hive/HiveTestUtils.java
@@ -77,12 +77,14 @@ import io.airlift.slice.Slice;
 
 import java.math.BigDecimal;
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 
 import static com.facebook.airlift.json.JsonCodec.jsonCodec;
 import static com.facebook.airlift.json.smile.SmileCodec.smileCodec;
 import static com.facebook.presto.common.type.Decimals.encodeScaledValue;
 import static com.facebook.presto.hive.HiveDwrfEncryptionProvider.NO_ENCRYPTION;
+import static java.lang.String.format;
 import static java.util.stream.Collectors.toList;
 
 public final class HiveTestUtils
@@ -266,5 +268,29 @@ public final class HiveTestUtils
     public static Slice longDecimal(String value)
     {
         return encodeScaledValue(new BigDecimal(value));
+    }
+
+    public static Optional<String> getProperty(String name)
+    {
+        String systemPropertyValue = System.getProperty(name);
+        String environmentVariableValue = System.getenv(name);
+        if (systemPropertyValue == null) {
+            if (environmentVariableValue == null) {
+                return Optional.empty();
+            }
+            else {
+                return Optional.of(environmentVariableValue);
+            }
+        }
+        else {
+            if (environmentVariableValue != null && !systemPropertyValue.equals(environmentVariableValue)) {
+                throw new IllegalArgumentException(format("%s is set in both Java system property and environment variable, but their values are different. The Java system property value is %s, while the" +
+                                " environment variable value is %s. Please use only one value.",
+                        name,
+                        systemPropertyValue,
+                        environmentVariableValue));
+            }
+            return Optional.of(systemPropertyValue);
+        }
     }
 }
--- a/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeRemoteFunctions.java
+++ b/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeRemoteFunctions.java
@@ -18,12 +18,12 @@ import com.facebook.presto.tests.AbstractTestQueryFramework;
 import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Test;
 
+import static com.facebook.presto.hive.HiveTestUtils.getProperty;
 import static com.facebook.presto.nativeworker.NativeQueryRunnerUtils.createLineitem;
 import static com.facebook.presto.nativeworker.PrestoNativeQueryRunnerUtils.REMOTE_FUNCTION_CATALOG_NAME;
 import static com.facebook.presto.nativeworker.PrestoNativeQueryRunnerUtils.REMOTE_FUNCTION_JSON_SIGNATURES;
 import static com.facebook.presto.nativeworker.PrestoNativeQueryRunnerUtils.setupJsonFunctionNamespaceManager;
 import static com.facebook.presto.nativeworker.PrestoNativeQueryRunnerUtils.startRemoteFunctionServer;
-import static java.util.Objects.requireNonNull;
 
 @Test(groups = "remote-function")
 public abstract class AbstractTestNativeRemoteFunctions
@@ -49,12 +49,8 @@ public abstract class AbstractTestNativeRemoteFunctions
     {
         // Initialize the remote function thrift server process.
         String propertyName = "REMOTE_FUNCTION_SERVER";
-        remoteFunctionServerBinaryPath = System.getenv(propertyName);
-        if (remoteFunctionServerBinaryPath == null) {
-            remoteFunctionServerBinaryPath = System.getProperty(propertyName);
-        }
-
-        requireNonNull(remoteFunctionServerBinaryPath, "Remote function server path missing. Add -DREMOTE_FUNCTION_SERVER=<path/to/binary> to your JVM arguments.");
+        remoteFunctionServerBinaryPath = getProperty(propertyName).orElseThrow(() -> new NullPointerException("Remote function server path missing. Add -DREMOTE_FUNCTION_SERVER=<path/to/binary>" +
+                " to your JVM arguments, or create an environment variable REMOTE_FUNCTION_SERVER with value <path/to/binary>."));
         remoteFunctionServerUds = startRemoteFunctionServer(remoteFunctionServerBinaryPath);
     }
 
--- a/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/PrestoNativeQueryRunnerUtils.java
+++ b/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/PrestoNativeQueryRunnerUtils.java
@@ -31,6 +31,7 @@ import java.nio.file.Paths;
 import java.util.Optional;
 import java.util.UUID;
 
+import static com.facebook.presto.hive.HiveTestUtils.getProperty;
 import static com.facebook.presto.nativeworker.NativeQueryRunnerUtils.getNativeWorkerHiveProperties;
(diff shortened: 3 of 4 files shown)


presto-hive/src/test/java/com/facebook/presto/hive/HiveTestUtils.java
```diff
@@ -267,4 +269,28 @@ public static Slice longDecimal(String value)
...
         return encodeScaledValue(new BigDecimal(value));
     }
+
+    public static Optional<String> getProperty(String name)
+    {
+        String systemPropertyValue = System.getProperty(name);
+        String environmentVariableValue = System.getenv(name);
+        if (systemPropertyValue == null) {
+            if (environmentVariableValue == null) {
+                return Optional.empty();
```
ZacBlanco, line 279: if you're returning optional values, and the result is `null`, why not use `Optional.ofNullable`?

presto-hive/src/test/java/com/facebook/presto/hive/HiveTestUtils.java
```diff
@@ -267,4 +269,28 @@ public static Slice longDecimal(String value)
...
+            }
+            else {
+                return Optional.of(environmentVariableValue);
+            }
+        }
+        else {
+            if (environmentVariableValue != null && !systemPropertyValue.equals(environmentVariableValue)) {
```
ZacBlanco, line 286: I'm glad you're adding this check. It will definitely prevent even more confusing situations.

presto-hive/src/test/java/com/facebook/presto/hive/HiveTestUtils.java
```diff
@@ -267,4 +269,28 @@ public static Slice longDecimal(String value)
...
+        String environmentVariableValue = System.getenv(name);
+        if (systemPropertyValue == null) {
+            if (environmentVariableValue == null) {
+                return Optional.empty();
+            }
+            else {
+                return Optional.of(environmentVariableValue);
+            }
+        }
+        else {
```
ZacBlanco, line 285: IMO the nesting of conditionals is confusing to read. Here's how I would structure it:

```java
Optional<String> systemPropertyValue = Optional.ofNullable(System.getProperty(name));
Optional<String> environmentVariableValue = Optional.ofNullable(System.getenv(name));

if (!systemPropertyValue.isPresent() && !environmentVariableValue.isPresent()) {
    return Optional.empty();
}

if (!systemPropertyValue.isPresent() && environmentVariableValue.isPresent()) {
    return environmentVariableValue;
}

if (!systemPropertyValue.equals(environmentVariableValue)) {
  throw new ....
}

return systemPropertyValue;
```

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeRemoteFunctions.java
```diff
@@ -49,12 +50,10 @@ protected void preInitQueryRunners()
     {
         // Initialize the remote function thrift server process.
         String propertyName = "REMOTE_FUNCTION_SERVER";
-        remoteFunctionServerBinaryPath = System.getenv(propertyName);
-        if (remoteFunctionServerBinaryPath == null) {
-            remoteFunctionServerBinaryPath = System.getProperty(propertyName);
-        }
+        remoteFunctionServerBinaryPath = getProperty(propertyName).orElse(null);
```
ZacBlanco, line 53: I think it's better to throw errors when the system property needs to be set. Maybe use `orElseThrow` here?

```suggestion
        remoteFunctionServerBinaryPath = getProperty(propertyName).orElseThrow(() -> ...);
```

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/PrestoNativeQueryRunnerUtils.java
```diff
@@ -39,14 +40,12 @@
...
-    private static final Logger log = Logger.get(PrestoNativeQueryRunnerUtils.class);
-
-    private static final String DEFAULT_STORAGE_FORMAT = "DWRF";
-
     // The unix domain socket (UDS) used to communicate with the remote function server.
     public static final String REMOTE_FUNCTION_UDS = "remote_function_server.socket";
     public static final String REMOTE_FUNCTION_JSON_SIGNATURES = "remote_function_server.json";
     public static final String REMOTE_FUNCTION_CATALOG_NAME = "remote";
+    private static final Logger log = Logger.get(PrestoNativeQueryRunnerUtils.class);
+    private static final String DEFAULT_STORAGE_FORMAT = "DWRF";
```
ZacBlanco, line 48: The `log` should probably be moved back to the top

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/PrestoNativeQueryRunnerUtils.java
```diff
@@ -39,14 +40,12 @@
...
 {
-    private static final Logger log = Logger.get(PrestoNativeQueryRunnerUtils.class);
-
-    private static final String DEFAULT_STORAGE_FORMAT = "DWRF";
-
     // The unix domain socket (UDS) used to communicate with the remote function server.
     public static final String REMOTE_FUNCTION_UDS = "remote_function_server.socket";
     public static final String REMOTE_FUNCTION_JSON_SIGNATURES = "remote_function_server.json";
     public static final String REMOTE_FUNCTION_CATALOG_NAME = "remote";
+    private static final Logger log = Logger.get(PrestoNativeQueryRunnerUtils.class);
```
aditi-pandit, line 47: Please don't move these if there are no changes.

presto-hive/src/test/java/com/facebook/presto/hive/HiveTestUtils.java
yingsu00: @ZacBlanco In this way the systemPropertyValue will be checked multiple times while my original code does not do the if check more than necessary. It would also create the Optional objects more than necessary. In the past I've seen big perf and reliability problems by creating Optional objects too many and too fast. But since this is just a one-time off thing I'm good with both ways. I'll update the PR using your rewrite.

presto-hive/src/test/java/com/facebook/presto/hive/HiveTestUtils.java
yingsu00: Changed to your rewritings anyways.

presto-hive/src/test/java/com/facebook/presto/hive/HiveTestUtils.java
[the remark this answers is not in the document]
yingsu00: I updated to Zac's suggestion so this line is no longer there.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/PrestoNativeQueryRunnerUtils.java
yingsu00: @aditi-pandit These changes were added by the code formatter, but I think it makes sense because the private vars should be after the public ones. I thought it may not be necessary to create a separate commit for this small change, but I can do it if you insist.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeRemoteFunctions.java
yingsu00: UPdated

presto-hive/src/test/java/com/facebook/presto/hive/HiveTestUtils.java
yingsu00: @ZacBlanco @aditi-pandit This block is missing one check and it should be like this:
```
public static Optional<String> getProperty(String name)
    {
        Optional<String> systemPropertyValue = Optional.ofNullable(System.getProperty(name));
        Optional<String> environmentVariableValue = Optional.ofNullable(System.getenv(name));

        if (!systemPropertyValue.isPresent() && !environmentVariableValue.isPresent()) {
            return Optional.empty();
        }

        if (!systemPropertyValue.isPresent() && environmentVariableValue.isPresent()) {
            return environmentVariableValue;
        }

        if (environmentVariableValue.isPresent() && !systemPropertyValue.equals(environmentVariableValue)) {.  // need to have environmentVariableValue.isPresent() && 
            throw new IllegalArgumentException(format("%s is set in both Java system property and environment variable, but their values are different. The Java system property value is %s, while the" +
                            " environment variable value is %s. Please use only one value.",
                    name,
                    systemPropertyValue,
                    environmentVariableValue));
        }

        return systemPropertyValue;
    }
```
I think this is actually harder to understand than my original code. What do you think?

ZacBlanco (approved):

presto-hive/src/test/java/com/facebook/presto/hive/HiveTestUtils.java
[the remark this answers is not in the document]
yingsu00: > systemPropertyValue is not null is inferred from the if condition at this point. So the check is not needed.

I took a look again but this check is on environmentVariableValue, not systemPropertyValue. We still need to check it here.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/PrestoNativeQueryRunnerUtils.java
aditi-pandit: That's fine. There isn't a need for separate commit.

aditi-pandit (approved):

kewang1024 (commented):
Is the test failing due to setting `DWORKER_COUNT=0`?

yingsu00:
> Is the test failing due to setting DWORKER_COUNT=0?

@kewang1024 Yes you are right. Thanks for pointing it out.

tdcmeehan (approved):

PR #21093, opened 2023-10-10
[native][circleci] Fix native build failure

mshang816:
## Description
Even with previous PRs like https://github.com/prestodb/presto/pull/21093, the native ci jobs are still failing. So it looks like the previous changes do not work very well. A few more changes:
* change link_job_pool to presto_link_job_pool. link_job_pool is used by velox. Make the change to enable finer grained control from our side.
* add more targets to the job pool
* more testing to make sure the fix does work:
- compare results between 2 and 8 and 2 passes while fails which is expected
- try to disable ccache to see if jobs still pass. 

## Test Plan
Make sure all circle ci runs are green

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/CMakeLists.txt
+++ b/presto-native-execution/CMakeLists.txt
@@ -207,4 +207,12 @@ if(PRESTO_ENABLE_JWT)
   add_compile_definitions(PRESTO_ENABLE_JWT)
 endif()
 
+if("${MAX_LINK_JOBS}")
+  set_property(GLOBAL APPEND PROPERTY JOB_POOLS
+                                      "presto_link_job_pool=${MAX_LINK_JOBS}")
+else()
+  set_property(GLOBAL APPEND PROPERTY JOB_POOLS "presto_link_job_pool=8")
+endif()
+set(CMAKE_JOB_POOL_LINK presto_link_job_pool)
+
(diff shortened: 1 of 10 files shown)


.circleci/continue_config.yml
```diff
@@ -112,7 +112,6 @@ jobs:
               -DPRESTO_ENABLE_REMOTE_FUNCTIONS=ON \
               -DPRESTO_ENABLE_JWT=ON \
               -DCMAKE_PREFIX_PATH=/usr/local \
-              -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
```
majetideepak, line 115: Why are we removing the ccache?

majetideepak: I just noticed `[TEST_ONLY]`. My bad.

mshang816: > I just noticed `[TEST_ONLY]`. My bad.

@majetideepak I remove ccache for testing because I suspect that without ccache the failure is harder to surface. Will add it back later in a formal fix

pranjalssh (approved):

kewang1024 (approved):

gggrace14 (approved):

PR #21097, opened 2023-10-10
Fix RowExpressionTest to be forward compatible with variant.toJson 

kgpai:
This is required so we can land https://github.com/facebookincubator/velox/pull/6216 . 

## Description
Makes a test forward compatible.

```
== NO RELEASE NOTE ==
```


majetideepak (approved):
Thanks, @kgpai

xiaoxmeng (approved):

PR #21105, opened 2023-10-11
Remove storageFormat from dataDirectory in QueryRunners

yingsu00:
## Description
The Java HiveQueryRunner takes a program argument as the input dataDirectory so the users can persist any data in any format at this location. However PrestoNativeQueryRunnerUtils always adds a storageFormat string to the end of the input dataDirectory, and this makes the existing user defined data directories unusable and previously created schemas and data files invisible. The QueryRunners shall allow the users to designate the data directory by themselves. This commit removes the storageFormat from dataDirectory path.

## Motivation and Context
Be able to read the data previously created in HiveQueryRunner

## Impact
Data will be stored at $DATA_DIR/hive_data

## Test Plan
Before the change
```
presto:tpch> show schemas;
        Schema
----------------------
 __temporary_tables__
 information_schema
 tpcds
 tpcds_bucketed
 tpch
 tpch_bucketed
(6 rows)
```
After change
```
presto:tpch> show schemas;
           Schema
-----------------------------
 __temporary_tables__
 information_schema
 swiftly
 temp
 test
 test1
 tmp
 tpcds
 tpcds_1g_varchar
 tpcds_bucketed
 tpch
 tpch10g
 tpch1_dwrf
 tpch1_dwrf_partitioned
 tpch1g
 tpch_10
 tpch_100_dwrf
 tpch_100_parquet
 tpch_100_parquet_compressed
 tpch_10_parquet
 tpch_1_parquet
 tpch_1_parquet_partitioned
 tpch_1_parquet_smallfiles
 tpch_1_partitioned
 tpch_bucketed
 tpch_bucketed_tinytiny
 tpch_tiny
 tpch_tiny_bucketed
 tpch_tiny_parquet
 tpch_tiny_parquet_plain
 tpch_tinytiny
(31 rows)
```
The other schemas were created using HiveQueryRunner before.

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```


aditi-pandit:
@yingsu00 : This was added once we started running both DWRF and Parquet tests (around May). And the separation seems useful when debugging tests.

This current change might be disruptive now for folks who have been using the new format for a while.

@amitkdutta : What do you think about this change ? I remember there were issues when the DATA_DIR was flipped to use the Storage format suffix as well.

yingsu00:
@aditi-pandit HiveQueryRunner uses location "$(DATA_DIR)/hive_data" because it would append "hive_data" at the end of user supplied DATA_DIR, but the HiveExternalWorkerQueryRunner uses "$(DATA_DIR)/$(storageFormat)/hive_data", e.g. "/Users/yingsu/dwrf/hive_data". At least we need to align the two query runners. Also, "/Users/yingsu/dwrf/hive_data" doesn't read quite nicely. User would expect something like "/Users/yingsu/hive_data/dwrf/" with "hive_data" first, since DWRF or Parquet are both Hive formats.

yingsu00:
cc @frankobe

yingsu00:
@aditi-pandit I just realized that if we add storage format behind hive_data, e.g. "/Users/yingsu/hive_data/dwrf/", then we won't be able to see data under "/Users/yingsu/hive_data/parquet/" simultaneously when we run HiveQueryRunner or HiveExternalWorkerQueryRunner. This is not desired behavior because when you run the standalone query runners to debug queries, you will want to see all data in all formats so you can compare the behavior between different storage formats. This can only be achieved with the changes in this PR. If any specific tests need to identify the file formats, they can add the file format as part of the path themselves.

aditi-pandit:
@yingsu00 : Let me outline few scenarios :

- With this change HiveQueryRunner and HiveExternalWorkerQueryRunner can access the same files in the data_directories. This part is reasonable.
- Right now the storage format is part of the data directory path, but the actual files could be PARQUET or DWRF based on the storage format. Right now, the tests assume the default storage format so the placement of files is alright within the scope of the test. However, when running standalone with the presto client it could lead to confusion.  Removing the storage format from the directory path is reasonable then.
- What is the impact to current users ? Right now, they could have 2 duplicate schemas with the same name and tables with different storage formats all under DATA_DIR. After this change a table in a schema could have only one storage format. You would need to outline some steps to migrate the tables.

kewang1024:
> @yingsu00 : Let me outline few scenarios :
> 
> * With this change HiveQueryRunner and HiveExternalWorkerQueryRunner can access the same files in the data_directories. This part is reasonable.
> * Right now the storage format is part of the data directory path, but the actual files could be PARQUET or DWRF based on the storage format. Right now, the tests assume the default storage format so the placement of files is alright within the scope of the test. However, when running standalone with the presto client it could lead to confusion.  Removing the storage format from the directory path is reasonable then.
> * What is the impact to current users ? Right now, they could have 2 duplicate schemas with the same name and tables with different storage formats all under DATA_DIR. After this change a table in a schema could have only one storage format. You would need to outline some steps to migrate the tables.

From the scenarios listed about, looks like the conclusion is that it's reasonable to remove the storage format from the directory?

PR #21122, opened 2023-10-11
[native] Add spill max file size to session property

tanjialiang:
```
== NO RELEASE NOTE ==
```


xiaoxmeng (approved):
@tanjialiang LGTM. Thanks!

presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
```diff
@@ -1525,6 +1526,11 @@ public SystemSessionProperties(
                                 "0 is the initial spilling level, -1 means unlimited.",
                         4,
                         false),
+                integerProperty(
+                        NATIVE_MAX_SPILL_FILE_SIZE,
```
amitkdutta, line 1530: Are we planning to use this to rotate spill files (e.g. if spill file size reach to a limit, a new file will be created). Like log rotating? If so - how do we know the file size? Another option is byte size per file - which Presto server already keeps track of (presto_cpp.spilled_bytes counter)

xiaoxmeng: @amitkdutta we keeps track of the current written spill file size, and switch to a new file if the current spill file reaches to this limit.

ajaygeorge (approved):

PR #21130, opened 2023-10-12
[native]Send completed get data result if velox task has finished

xiaoxmeng:
The following race condition can cause table writer hang when scale writer is enabled

T1 coordinator schedule a new task writer
T2 one or any source task has produced all the data and all have been consumed. It triggers task
     to complete
T3 the new task writer tries to fetch data from the closed source, and find the source
     task is not running, then wait for timeout and return an empty result to retry from the writer
     task
T4 the new task writer keeps retrying and after the presto task has been reclaimed, the new data
     fetch will triggers a new presto task creation and continue retrying with empty result.

The short-term fix for this is to return completion result in T3 if the velox task is in finished state
instead of returning empty result. A completed task will stay in one-minute which is sufficient
long in practice to handle this race condition. Have verified in meta internal testing.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/TaskManager.cpp
+++ b/presto-native-execution/presto_cpp/main/TaskManager.cpp
@@ -93,6 +93,14 @@ std::unique_ptr<Result> createTimeOutResult(long token) {
   return result;
 }
 
+std::unique_ptr<Result> createCompleteResult(long token) {
+  auto result = std::make_unique<Result>();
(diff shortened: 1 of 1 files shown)


mbasmanova (approved):
Thanks. Please, update commit message to add details. Consider mentioning that this happens only in case when scaled_writers is enabled.

kewang1024 (approved):

Issue #20487, opened 2023-08-04
[BUG] - Reading wrong timestamp values in iceberg table when table is partitioned on timestamp


SAITEJA245:
Here is my table structure from the cli

```
presto:db> show create table event_new;
             Create Table             
--------------------------------------
 CREATE TABLE iceberg.db.event_new (  
    "bot_ref" integer,                
    "user_id" varchar,                
    "event_type" varchar,             
    "eventdata" varchar,              
    "triggeredby" varchar,            
    "timestamp" timestamp,            
    "platform" varchar,               
    "session_id" varchar,             
    "message_id" varchar,             
    "parent_id" real                  
 )                                    
 WITH (                               
    format = 'PARQUET',               
    partitioning = ARRAY['timestamp'] 
 )                                    
(1 row)
=========================spark-sql============================
spark-sql> show create table local.db.event_new;
CREATE TABLE local.db.event_new (
  `bot_ref` INT,
  `user_id` STRING,
  `event_type` STRING,
  `eventData` STRING,
  `triggeredBy` STRING,
  `timestamp` TIMESTAMP,
  `platform` STRING,
  `session_id` STRING,
  `message_id` STRING,
  `parent_id` FLOAT)
USING iceberg
PARTITIONED BY (timestamp)
LOCATION 'hdfs://hbase-test-1.dev.local:9000/hbase/warehouse/db/event_new'
TBLPROPERTIES(
  'current-snapshot-id' = '5602301537647142919',
  'format' = 'iceberg/parquet',
  'format-version' = '2')
```

as seen above the table is portioned on timestamp column
iceberg table data and metadata is stored in HADOOP FILE SYSTEM 

query results
```
presto:db> select timestamp, count(*) from event_new group by timestamp  order by count(*) desc;
         timestamp         | _col1  
---------------------------+--------
 +54918-05-21 00:00:00.000 | 311174 
 +54921-02-14 00:00:00.000 | 296473 
 +54923-11-11 00:00:00.000 | 267328 
 +54915-08-25 00:00:00.000 | 258690 
```

```
spark-sql> select timestamp, count(*) from local.db.event_new group by timestamp  order by count(*) desc;
2022-12-13 00:00:00	311174
2022-12-14 00:00:00	296473
2022-12-15 00:00:00	267328
2022-12-12 00:00:00	258690

```
due to this i am not able to query from hive with timestamp filters

## Your Environment
* Presto version used: `presto-server-0.282`
[...]

PR #21156, opened 2023-10-16
Fix Iceberg table partitioned by columns of TimestampType

hantangwangd:
## Description

In current implementation, when Iceberg table was partitioned by columns of TimestampType, we would meet some problems when insert, select, delete from it, or query system table "$partitions". Also see #20487 

The main reason is, Iceberg maintain data of TimestampType as long of microseconds, but presto by default handle data of TimestampType as long of milliseconds. This would introduce problems when iceberg use the long value to handle partitions.

This PR fix the problems when we use partition columns of TimestampType in IcebergTable.

## Test Plan

 - Make sure this fix do not affect existing test cases in IcebergDistributedTestBase
 - Newly added test case in IcebergDistributedTestBase.testPartitionedByTimestampType()

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergPageSource.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergPageSource.java
@@ -20,6 +20,7 @@ import com.facebook.presto.common.block.RunLengthEncodedBlock;
 import com.facebook.presto.common.type.DecimalType;
 import com.facebook.presto.common.type.Decimals;
 import com.facebook.presto.common.type.TimeZoneKey;
+import com.facebook.presto.common.type.TimestampType;
 import com.facebook.presto.common.type.Type;
 import com.facebook.presto.common.type.VarbinaryType;
 import com.facebook.presto.common.type.VarcharType;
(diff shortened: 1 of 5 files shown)


gupteaj (approved):
Changes looks good to me.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergPageSource.java
```diff
@@ -242,4 +247,16 @@ private static Object deserializePartitionValue(Type type, String valueString, S
...
+    private Block nativeValueToBlock(Type type, Object prefilledValue)
+    {
+        if (prefilledValue != null && type instanceof TimestampType && ((TimestampType) type).getPrecision() == MILLISECONDS) {
+            BlockBuilder blockBuilder = type.createBlockBuilder(null, 1);
+            writeNativeValue(type, blockBuilder, MICROSECONDS.toMillis((long) prefilledValue));
+            return blockBuilder.build();
+        }
+        else {
+            return Utils.nativeValueToBlock(type, prefilledValue);
+        }
```
kiersten-stokes, line 260: I think we can simplify this a little bit

```suggestion
        if (prefilledValue != null && type instanceof TimestampType && ((TimestampType) type).getPrecision() == MILLISECONDS) {
            return Utils.nativeValueToBlock(type, 
 MICROSECONDS.toMillis((long) prefilledValue));
        }
        return Utils.nativeValueToBlock(type, prefilledValue);
```

kiersten-stokes (approved):
Tested this out and it looks good to me! Just had one NIT below

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergPageSource.java
hantangwangd: Fixed, thanks.

tdcmeehan (approved):

PR #21173, opened 2023-10-17
Update the snappy-java version

Akanksha-kedia:
Screenshot 2023-10-17 at 5 18 34 PM


skairali (changes requested):
<dependency>
            <groupId>org.xerial.snappy</groupId>
            <artifactId>snappy-java</artifactId>
            <version>1.1.10.4</version>
            <scope>runtime</scope>
        </dependency>

this looks like a new dependency addition and not update

why are we updating presto elasticsearch dependency?

skairali (changes requested):
I see so many exclusions

The PR does not indicate the purpose of excluding or introducing dependencies

Akanksha-kedia:
> I see so many exclusions
> 
> The PR does not indicate the purpose of excluding or introducing dependencies

transistive dependecies have to be excluded and its not a new dependencies added.

jaystarshot:
I see Snappy being defined in presto root and used in other modules too. Why not change the version there? Why is this being added as an dependency in new modules?

Akanksha-kedia:
@skairali @jaystarshot @tdcmeehan please review and let me know anything missed.

Akanksha-kedia:
@tdcmeehan please review

jaystarshot:
Please add a release note 
`Updated the snappy version to 1.1.10.4`
Otherwise lgtm

PR #21182, opened 2023-10-18
[native] Allow exchange payload copy to happen in driver thread

tanjialiang:
Add option to allow the exchange response payload copy and memory allocation to happen in driver thread to reduce load in IO thread pool. This feature is configurable via system config property exchange.immediate-buffer-transfer
```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/PrestoExchangeSource.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoExchangeSource.cpp
@@ -62,7 +62,9 @@ std::string bodyAsString(
   auto iobufs = response.consumeBody();
   for (auto& body : iobufs) {
     oss << std::string((const char*)body->data(), body->length());
-    pool->free(body->writableData(), body->capacity());
+    if (pool != nullptr) {
+      pool->free(body->writableData(), body->capacity());
+    }
   }
   return oss.str();
 }
@@ -83,6 +85,8 @@ PrestoExchangeSource::PrestoExchangeSource(
       port_(baseUri.port()),
       clientCertAndKeyPath_(clientCertAndKeyPath),
       ciphers_(ciphers),
+      immediateBufferTransfer_(
+          SystemConfig::instance()->exchangeImmediateBufferTransfer()),
       driverExecutor_(driverExecutor),
       httpExecutor_(httpExecutor) {
   folly::SocketAddress address;
@@ -95,12 +99,13 @@ PrestoExchangeSource::PrestoExchangeSource(
       SystemConfig::instance()->exchangeRequestTimeout());
   VELOX_CHECK_NOT_NULL(driverExecutor_);
   VELOX_CHECK_NOT_NULL(httpExecutor_);
+  VELOX_CHECK_NOT_NULL(pool_);
   auto* ioEventBase = httpExecutor_->getEventBase();
   httpClient_ = std::make_shared<http::HttpClient>(
       ioEventBase,
       address,
       timeoutMs,
-      pool_,
+      immediateBufferTransfer_ ? pool_ : nullptr,
       clientCertAndKeyPath_,
       ciphers_,
       [](size_t bufferBytes) {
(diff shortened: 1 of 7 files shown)


presto-native-execution/presto_cpp/main/PrestoExchangeSource.cpp
```diff
@@ -54,15 +54,18 @@ void onFinalFailure(
 
 std::string bodyAsString(
     http::HttpResponse& response,
-    memory::MemoryPool* pool) {
+    memory::MemoryPool* pool,
```
xiaoxmeng, line 57: If pool is null, then no free? So we don't need immediateBufferTransfer?

xiaoxmeng (commented):
@tanjialiang thanks for looking into this!

presto-native-execution/presto_cpp/main/PrestoExchangeSource.cpp
tanjialiang: this pool is passed in from PrestoExchangeSource which will always be non-null. So we need to rely on the immediateBufferTransfer flag to tell if we shall free. Note that only inside HttpClient/HttpResponse we can tell based on nullness of pool

presto-native-execution/presto_cpp/main/http/HttpClient.cpp
```diff
@@ -116,10 +115,52 @@ void HttpResponse::append(std::unique_ptr<folly::IOBuf>&& iobuf) {
...
+std::vector<std::unique_ptr<folly::IOBuf>> HttpResponse::consumeBody(
+    velox::memory::MemoryPool* pool) {
+  VELOX_CHECK_NULL(pool_);
+  VELOX_CHECK(!hasError());
+  uint64_t totalBytes{0};
+  for (const auto& iobuf : bodyChain_) {
+    VELOX_CHECK(!iobuf->isChained());
+    totalBytes += iobuf->length();
+  }
+  pool->maybeReserve(totalBytes);
```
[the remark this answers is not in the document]
tanjialiang: Yeah makes sense.

presto-native-execution/presto_cpp/main/PrestoExchangeSource.cpp
xiaoxmeng: Can callsite of bodyAsString to determine pass pool or not based on immediateBufferTransfer? So the callee doesn't need to know the details and simply based on if pool is null or not? Thanks!

presto-native-execution/presto_cpp/main/http/HttpClient.h
```diff
@@ -70,11 +70,23 @@ class HttpResponse {
...
+  /// Consumes the response body. The memory of body will be transferred to the
+  /// memory to be allocated from 'pool'.
+  std::unique_ptr<folly::IOBuf> consumeBody(velox::memory::MemoryPool* pool);
+
   void freeBuffers();
 
   std::string dumpBodyChain() const;
 
  private:
+  // The append operation that transfers the buffer memory to memory allocated
```
xiaoxmeng, line 82:
```
// Copies the 'iobuf' to velox memory 'pool_' and free 'iobuf' immediately.
```

presto-native-execution/presto_cpp/main/http/HttpClient.h
```diff
@@ -70,11 +70,23 @@ class HttpResponse {
...
+  // from 'pool_'.
+  void appendWithCopy(std::unique_ptr<folly::IOBuf>&& iobuf);
+
+  // The append operation that shallowly append 'iobuf' without transferring its
```
xiaoxmeng, line 86:
```
// Appends the 'iobuf' to 'bodyChain_', and copies them all once into a single large velox buffer after receives the entire http response payload.
```

xiaoxmeng (approved):
@tanjialiang LGTM. Thanks!

PR #21197, opened 2023-10-20
[native pos] Disable old task cleanup routine

vermapratyush:
## Motivation and Context
PrestoServer contains oldTaskCleanup routine which cleans up tasks which are not running or have not recieved heart-beat in a long while. This is meant to detect zombie or stuck tasks, this is not relevant in Presto-On-Spark as Spark speculative execution takes care of it.

## Impact
In production we have seen legitimate tasks getting cleanup, especially jobs which are long running or have skew. Disabling old task cleanup fixes the problem.
When task cleanup routine takes effect, we see that JAVA keeps on pooling as the task has been removed from `TaskResource::taskMap_` thus making no progress.

## Test Plan
Verified on internal build on queries where we were seeing old task cleanup take action.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
+++ b/presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
@@ -86,7 +86,10 @@ void PeriodicTaskManager::start() {
 
   VELOX_CHECK_NOT_NULL(taskManager_);
   addTaskStatsTask();
-  addOldTaskCleanupTask();
+
+  if (SystemConfig::instance()->enableOldTaskCleanUp()) {
+    addOldTaskCleanupTask();
+  }
 
   if (memoryAllocator_ != nullptr) {
     addMemoryAllocatorStatsTask();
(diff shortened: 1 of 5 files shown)


mbasmanova (commented):
Consider adding a flag to enable / disable this logic instead of setting very high timeout.

mbasmanova (approved):
Looks good. Please, follow-up to document this and other configuration properties in presto-docs/src/main/sphinx/develop/presto-native.rst

vermapratyush:
@orestobot kick off tests

vermapratyush:
@prestobot kick off tests

presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
```diff
@@ -86,7 +86,10 @@ void PeriodicTaskManager::start() {
 
   VELOX_CHECK_NOT_NULL(taskManager_);
   addTaskStatsTask();
-  addOldTaskCleanupTask();
+
+  if (SystemConfig::instance()->enableOldTaskCleanUp()) {
```
xiaoxmeng, line 90: @pratyakshsharma what's the problem if we cleanup old tasks after 1 min by default in Spark-on-Prestissimo? Thanks!

vermapratyush: @xiaoxmeng Not all Presto APIs are implemented by Sapphire, for example heartbeat API. This results in prestissimo to think that a long running task is stale and it gets purged.
Due to the isolated execution of tasks in Sapphire, this feature is not needed at all.

PR #21203, opened 2023-10-20
Add null stats for join probe

feilong-liu:
## Description

Track number of probe keys and number of null probe keys in HBO, and use it to enable outer join null salt.

## Motivation and Context
Currently HBO only records the null key stats for join build side, and use it to enable null salt in HBO. This PR also add probe side information and use it to enable null salt in HBO too.

## Impact
Resolve the probe side skew in null join.

## Test Plan
Add unit test
And [end to end test on tracking](https://fburl.com/scuba/presto_query_plan_stats_inc_archive/onrle7dz)

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Track null probe keys in join and use it to enable outer join null salt in history based optimization.
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/operator/JoinProbe.java
+++ b/presto-main/src/main/java/com/facebook/presto/operator/JoinProbe.java
@@ -55,6 +55,7 @@ public class JoinProbe
     private final boolean probeMayHaveNull;
 
     private int position = -1;
+    private int nullRowCount;
 
     private JoinProbe(int[] probeOutputChannels, Page page, Page probePage, @Nullable Block probeHashBlock)
     {
@@ -79,6 +80,7 @@ public class JoinProbe
     public long getCurrentJoinPosition(LookupSource lookupSource)
     {
         if (probeMayHaveNull && currentRowContainsNull()) {
+            ++nullRowCount;
             return -1;
         }
         if (probeHashBlock != null) {
@@ -116,5 +118,15 @@ public class JoinProbe
             }
         }
         return false;
+    }
+
+    public int getNullRowCount()
+    {
+        return nullRowCount;
+    }
+
+    public int getPositionCount()
+    {
+        return positionCount;
     }
 }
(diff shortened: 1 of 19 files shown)


presto-main/src/main/java/com/facebook/presto/operator/JoinProbe.java
```diff
@@ -79,6 +80,7 @@ public boolean advanceNextPosition()
     public long getCurrentJoinPosition(LookupSource lookupSource)
     {
         if (probeMayHaveNull && currentRowContainsNull()) {
+            ++nullRowCount;
```
feilong-liu, line 83: Piggyback the count here, without adding overhead in counting.

kaikalur (approved):
Looks good.

High level observation - salting on the inner side (right side of left and left side of right) is unnecessary. We should instead be filtering them out.

feilong-liu:
> Looks good.
> 
> High level observation - salting on the inner side (right side of left and left side of right) is unnecessary. We should instead be filtering them out.

Sure, I think this should be in a separate rule, and with this null filter, we will not hit the null skew in the inner side. Looks like that we already have one such rule `AddNotNullFiltersToJoinNode` which is default to false and we need to enable it.

kaikalur:
[like]  Sreeni Viswanadha reacted to your message:

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/RandomizeNullKeyInOuterJoin.java
```diff
@@ -386,6 +384,15 @@ public PlanNode visitJoin(JoinNode joinNode, RewriteContext<Set<VariableReferenc
             return newJoinNode;
         }
 
+        private boolean hasNullSkew(JoinNodeStatsEstimate joinEstimate)
```
pranjalssh, line 387: Should this be based on ratio as compared to counts?

pranjalssh (approved):

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/RandomizeNullKeyInOuterJoin.java
kaikalur: Yeah but counts are better because irrespective of how big the table is you have to shuffle these anyway so in some sense it's not related to ratio per se

PR #21216, opened 2023-10-23
Use velox's OutputBufferManager instead of old PartitionedOutputBufferManager

kewang1024:
```
== NO RELEASE NOTE ==
```


kewang1024:
Use velox's OutputBufferManager instead of old PartitionedOutputBufferManager since the refactor on velox side https://github.com/facebookincubator/velox/pull/7102 is merged

PR #21226, opened 2023-10-23
Fix logging of plan optimizer

feilong-liu:
## Description
Change the result type of PlanOptimizer, so that it not only returns the new plan node, but also returns whether the optimizer is triggered or not.

## Motivation and Context
Currently, the logging of PlanOptimizer is to compare the old and new plan root with the equal operator. However, this leads to false positives, where any non meaningful changes (for example reorder of outputs etc) will show as optimizer triggered. It also relies on proper equal() override for all data structures in query plans.

In this PR, I change the PlanOptimizer to explicitly return whether an optimizer is triggered or not. I've implemented them for all existing plan optimizers.

## Impact
Fix logging of optimizers, and eliminate false positives.

## Test Plan
Existing unit tests.

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Fix logging of optimizer triggering, eliminating false positives where optimizer is logged as triggered but actually not. The PlanOptimizer now return both result plan and whether the plan changed or not.
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/AddExchanges.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/AddExchanges.java
@@ -164,10 +164,11 @@ public class AddExchanges
     }
 
     @Override
-    public PlanNode optimize(PlanNode plan, Session session, TypeProvider types, VariableAllocator variableAllocator, PlanNodeIdAllocator idAllocator, WarningCollector warningCollector)
+    public PlanOptimizerResult optimize(PlanNode plan, Session session, TypeProvider types, VariableAllocator variableAllocator, PlanNodeIdAllocator idAllocator, WarningCollector warningCollector)
     {
         PlanWithProperties result = new Rewriter(idAllocator, variableAllocator, session, partitioningProviderManager).accept(plan, PreferredProperties.any());
-        return result.getNode();
+        boolean optimizerTriggered = PlanNodeSearcher.searchFrom(result.getNode()).where(node -> node instanceof ExchangeNode && ((ExchangeNode) node).getScope().isRemote()).findFirst().isPresent();
+        return PlanOptimizerResult.optimizerResult(result.getNode(), optimizerTriggered);
     }
 
     private class Rewriter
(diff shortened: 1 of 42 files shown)


mlyublena:
Thanks for fixing these!

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/RandomizeNullKeyInOuterJoin.java
```diff
@@ -366,6 +374,7 @@ public PlanNode visitJoin(JoinNode joinNode, RewriteContext<Set<VariableReferenc
                             Optional.empty(),
```
mlyublena, line 374: should we also remove these inlined calls? This should now be properly handled outside in the main optimizer loop

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/AddExchanges.java
```diff
@@ -164,10 +164,11 @@ public AddExchanges(Metadata metadata, SqlParser parser, PartitioningProviderMan
     }
 
     @Override
-    public PlanNode optimize(PlanNode plan, Session session, TypeProvider types, VariableAllocator variableAllocator, PlanNodeIdAllocator idAllocator, WarningCollector warningCollector)
+    public PlanOptimizerResult optimize(PlanNode plan, Session session, TypeProvider types, VariableAllocator variableAllocator, PlanNodeIdAllocator idAllocator, WarningCollector warningCollector)
     {
         PlanWithProperties result = new Rewriter(idAllocator, variableAllocator, session, partitioningProviderManager).accept(plan, PreferredProperties.any());
-        return result.getNode();
+        boolean optimizerTriggered = PlanNodeSearcher.searchFrom(result.getNode()).where(node -> node instanceof ExchangeNode && ((ExchangeNode) node).getScope().isRemote()).findFirst().isPresent();
```
mlyublena, line 170: are we sure that this optimizer added the exchange node and it wasn't there before?

mlyublena (approved):

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/RandomizeNullKeyInOuterJoin.java
feilong-liu: Sure, let me remove this code too.

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/AddExchanges.java
feilong-liu: Yes, we only have this optimizer to add remote exchange node.

pranjalssh (approved):
Will all new optimizers will have to implement isOptimizerTriggered method. So all spi optimizers and our internal optimizers api might change. Please add that in release notes

PR #21227, opened 2023-10-23
[native] Avoid moving Lambda's arguments in toVeloxExpr().

spershin:
## Description
   In some cases we can visit this method with the same lambda
   more than once and having zero arguments and non-zero types would trigger a
   check down the stack.
   So, we make sure we don't mutate lambda here, while we investigate how that
   can happen and validate such behavior or fix a bug.

```
== NO RELEASE NOTE ==
```


gggrace14 (approved):
Looks good to me!

PR #21229, opened 2023-10-24
Fix OptimizedPartitionedOutputOperator memory usage

pranjalssh:
## Description

decodedBlock are kept in each of the partition buffers. We should let them go when we don't need them anymore.

In some rare cases of wide columns, some buffer overheads are higher than expected, and can lead to query using too much memory.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/operator/repartition/AbstractBlockEncodingBuffer.java
+++ b/presto-main/src/main/java/com/facebook/presto/operator/repartition/AbstractBlockEncodingBuffer.java
@@ -377,6 +377,9 @@ public abstract class AbstractBlockEncodingBuffer
             bufferAllocator.returnArray(nullsBuffer);
             nullsBuffer = null;
         }
+
+        // Recycle the decodedBlock. It will be set by setupDecodedBlocksAndPositions in the next call.
+        decodedBlock = null;
     }
 
     protected void serializeNullsTo(SliceOutput output)
--- a/presto-main/src/main/java/com/facebook/presto/operator/repartition/MapBlockEncodingBuffer.java
+++ b/presto-main/src/main/java/com/facebook/presto/operator/repartition/MapBlockEncodingBuffer.java
@@ -222,6 +222,8 @@ public class MapBlockEncodingBuffer
             bufferAllocator.returnArray(offsets);
             offsets = null;
         }
+
+        columnarMap = null;
     }
 
     @Override


presto-main/src/main/java/com/facebook/presto/operator/repartition/AbstractBlockEncodingBuffer.java
```diff
@@ -377,6 +377,9 @@ public void noMoreBatches()
             bufferAllocator.returnArray(nullsBuffer);
             nullsBuffer = null;
         }
+
```
agrawaldevesh, line 380: I am still trying to wrap my head around this change: Is this eliminating a leak in that references are never being cleaned or is it shortening the time that references are held ? Because I guess that decodedBlock will anyway be reset when appendData is called when the next page needs to be partitioned ? So it seems that we are cleaning up things for a very short amount of time ?

presto-main/src/main/java/com/facebook/presto/operator/repartition/MapBlockEncodingBuffer.java
```diff
@@ -222,6 +222,8 @@ public void noMoreBatches()
             bufferAllocator.returnArray(offsets);
             offsets = null;
         }
+
```
agrawaldevesh, line 225: same for this since this is also set up along with decodedBlock.

agrawaldevesh:
Should we also set decodedBlocks = null before the call to blockLeaseCloser.close() in the optimized partition output operator ?

presto-main/src/main/java/com/facebook/presto/operator/repartition/AbstractBlockEncodingBuffer.java
pranjalssh: These buffers are around 200mb(as discussed in chat). They will be reset when next data is appended to the same buffer(we have 300+ partition buffers). So keeping 300*200MB is too big

pranjalssh:
@agrawaldevesh  they will already be reset by then so wouldnt matter

presto-main/src/main/java/com/facebook/presto/operator/repartition/AbstractBlockEncodingBuffer.java
agrawaldevesh: I see. 300 partition buffers due to the hash partition count ? Can you add more color why it would be so delayed to clear the buffers ? Like addPage should be called very frequently. So we are holding onto memory very briefly less now. Why is that important?

pranjalssh: Yes but holding 200MB 300 times(for each hash partition) will not scale as it will equal to 60GB. This is aggravated by the fact that retained size of pages are 200MB(they should be much smaller. our page serialized size limit is 4MB) - which is a bug we need to fix as well.

agrawaldevesh: Exactly. So that is why I am curious why this fix works: it does not change the fact that 200x300 still, albeit for less duration. The question is how less ??

Btw, I would not index too much on retained size. This memory accounting stuff is always bogus. The big downside of over accounting is that system will be underutilized unnecessarily. Bad but not ooming.

arhimondr:
CC @yingsu00

agrawaldevesh (approved):
discussed offline

presto-main/src/main/java/com/facebook/presto/operator/repartition/MapBlockEncodingBuffer.java
```diff
@@ -222,6 +222,8 @@ public void noMoreBatches()
...
+        columnarMap = null;
```
yingsu00, line 226: What about the columnarArray and columnarRow in their specific buffers?

yingsu00 (commented):
@arhimondr I'm not sure how much this PR can help. I remember the fundamental problem was that the ColumnarXXX of some of the decoded dictionary blocks explode in size, or the estimated size for them was fundamentally smaller than actual. Will you folks be able to see if it's the estimated size being off? If yes the estimation code will need to be fixed. If the ColumnarZXX size indeed increases a lot after decoding blocks, it'll be good to find out why it is the case, then we can think of fixes.

presto-main/src/main/java/com/facebook/presto/operator/repartition/MapBlockEncodingBuffer.java
pranjalssh: Can you point which field you mean here?

pranjalssh:
@yingsu00 this PR helps a lot in some cases. In some cases these decoded blocks can be upto 200 MB(which is a bug in itself, but we need more time to investigate). If we keep each of these for all partition buffers, this can be several gigabytes of memory. As we don't need them, and this PR is doing the right thing code-wise - we can free them immediately so as to not cause cgroup ooms on our hosts.

So hoping we can move forward now - and next step is to figure out why we so large sized decoded blocks

pranjalssh:
Even if block sizes are 4MB, and our hash partition count is 1000, it can use upto 4GB of heap memory without any reservation

presto-main/src/main/java/com/facebook/presto/operator/repartition/MapBlockEncodingBuffer.java
yingsu00: @pranjalssh Never mind, these are not persisted in their specific classes. Sorry for the confusion.

yingsu00 (approved):

mbasmanova (commented):
@pranjalssh Any particular reason this fix comes without a test?

PR #21235, opened 2023-10-24
Publish query plan while a query is in running state

singcha:
## Description
Fixes a bug where query plan is not visible on presto query tool when a query is running with AQE enabled.

## Motivation and Context
This fixes a bug where AdaptiveQueryExecution doesn't publish query event before completion of query. This is deviation from queries running with AQE disabled where one can see query plan while query is running. This is required to keep the behavior consistent between aqe & non-aqe execution and ease debugging.

## Test Plan
Ran verifier with sampled queries from production and validated on presto query tool.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-spark-base/src/main/java/com/facebook/presto/spark/execution/PrestoSparkAdaptiveQueryExecution.java
+++ b/presto-spark-base/src/main/java/com/facebook/presto/spark/execution/PrestoSparkAdaptiveQueryExecution.java
@@ -65,6 +65,7 @@ import com.facebook.presto.sql.planner.plan.PlanFragmentId;
 import com.facebook.presto.sql.planner.plan.RemoteSourceNode;
 import com.facebook.presto.sql.planner.sanity.PlanChecker;
 import com.facebook.presto.transaction.TransactionManager;
+import com.google.common.collect.ImmutableList;
 import com.google.common.util.concurrent.UncheckedExecutionException;
 import io.airlift.units.Duration;
 import org.apache.spark.MapOutputStatistics;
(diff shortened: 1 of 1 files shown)


presto-spark-base/src/main/java/com/facebook/presto/spark/execution/PrestoSparkAdaptiveQueryExecution.java
```diff
@@ -247,6 +251,17 @@ protected List<Tuple2<MutablePartitionId, PrestoSparkSerializedPage>> doExecute(
...
+        queryMonitor.queryUpdatedEvent(
+                createQueryInfo(
+                        session,
+                        query,
+                        PLANNING,
+                        Optional.of(planAndMore),
+                        sparkQueueName,
+                        Optional.empty(),
+                        queryStateTimer,
+                        Optional.of(createStageInfo(session.getQueryId(), planFragmenter.fragmentQueryPlan(session, planAndMore.getPlan(), warningCollector), ImmutableList.of())),
```
shrinidhijoshi, line 263: Looks like we are using the static plan fragmenter to display the fragmented plan. While this fixes the regression, I wonder if there are cases where we might display the wrong information. Might be an edge case but something to think about. 

I also don't see a easy to fix this edge case, because even if we somehow figured out a way to publish few fragments at a time, the QueryInfo object expects a top down fragmented plan v/s bottom up plan.

shrinidhijoshi (commented):
Thanks for the fix @singcha. Left some comments

presto-spark-base/src/main/java/com/facebook/presto/spark/execution/PrestoSparkAdaptiveQueryExecution.java
singcha: Thats true. It may show stale info when runtime optimization kicks off which will get updated once we have completion event. Currently it just captures the plan we start with and the final plan once query completes.

pgupta2 (approved):

shrinidhijoshi (approved):

PR #21236, opened 2023-10-24
Add histograms for optimizer cost calculation

### LINKED ISSUE #22042: Add histogram statistic type

Histograms are a valuable data structure which can be used to generate accurate cardinality estimates of filters applied to tables or partitions. Generally they can be queried to _approximately_ answer the following questions:

1. What percent of values fall below X?
2. What percent of values fall above X?
3. What percent of values are equal to X?

During the query optimization phase, having answers to these questions will provide us with better cardinality estimates for cost-based rules, such as join re-ordering. Incorrect estimates can lead to potentially worse plans, while correct estimates will lead us to plans optimal for execution time.

By supporting histograms for the optimizer's internal representation we can generate better plans inside of the optimizer.

Presto should utilize histograms where available and fall back to the existing behavior if they don't in order to preserve backwards compatibility.

Query optimizer and connectors which can provide statistics.

## Possible Implementation

See #21236 

## Context

See first section.

ZacBlanco:
## Description

Closes #22042 

This commits provides two critical changes:

1. Adds a new enum value to ColumnStatisticType: Histogram.
2. Utilizes the new histograms in optimizer's cost calculations

## Histogram SPI Additions

With this change, a new column statistic type for histograms is introduced. In addition, a new SPI class `ConnectorHistogram` is also introduced. This interface is designed to be able to be implemented by either the connectors or in the main presto codebase. This should allow connectors to return histogram statistics in any format regardless of the source as long as they implement the interface.

The API is straightforward and includes 2 methods.

- cumulativeProbability(double value, boolean inclusive): -> CDF function
- inverseCumulativeProbability(double probability) -> inverse CDF function

A simple reference implementation is provided inside of
UniformDistributionHistogram. This implementation results in the same logic and same plans as the previous cost-based calculations. When the optimizer uses it, the calculations end up with the same number as prior to this PR, but now just utilizing the histogram API.

Additionally, to propagate statistics further up into a plan, another implementation of histograms is provided inside of DomainConstrainedHistogram. This class is used to bound a source histogram with a given domain as additional filters may be applied further up in the plan.

### Cost Calculations

Previously all cost calculations were performed inside of the ComparisonStatsCalculator using logic from the StatisticRange class to calculate the filter factors of overlapping and intersecting ranges. Instead, this change introduces cost calculation using the new histogram model and API. The core of the filter proportion calculation logic exists in the new HistogramCalculator utility class.

If the underlying Histogram implementation is swapped from the UniformDistributionHistogram, then the stats calculator will calculate the costs using the histogram information correctly.

If for some reason the new logic breaks existing plans for queries outside of the TPC-H/TPC-DS tests we have in the codebase, a new session flag `optimizer.use-histograms` is provided in order to fall back on the old logic.

### Testing

I verified through additional unit tests and the existing testing infrastructure that the implementation here results in the same cost and filter proportion calculations as the previous logic.

I also did a few small runs with TPC-DS SF1 on a real cluster with and without the cost calculations to verify there were no serious regressions

## Motivation and Context

By using histograms for cost calculations, we can get much more accurate predictions on how predicates will
affect the output of a node, potentially resulting in more cost-efficient plans.

## Impact

1. A new value in the `ColumnStatisticType` enum: `HISTOGRAM`
2. A new SPI interface to be implemented by connectors or inside `presto-main`: `ConnectorHistogram`

## Test Plan

1. Use existing test coverage to ensure no regressions
2. Extensive unit testing for corner cases without statistics (a majority of cases) to ensure no regressions

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Add histogram column statistic to Presto for the optimizer. Connectors can now implement support for them.
```

### DIFF
--- a/presto-benchto-benchmarks/src/test/java/com/facebook/presto/sql/planner/AbstractCostBasedPlanTest.java
+++ b/presto-benchto-benchmarks/src/test/java/com/facebook/presto/sql/planner/AbstractCostBasedPlanTest.java
@@ -14,6 +14,7 @@
 
 package com.facebook.presto.sql.planner;
 
+import com.facebook.presto.Session;
 import com.facebook.presto.spi.ConnectorTableHandle;
 import com.facebook.presto.spi.plan.AggregationNode;
 import com.facebook.presto.spi.plan.JoinDistributionType;
@@ -37,6 +38,7 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.stream.Stream;
 
+import static com.facebook.presto.SystemSessionProperties.OPTIMIZER_USE_HISTOGRAMS;
 import static com.facebook.presto.spi.plan.JoinDistributionType.REPLICATED;
 import static com.facebook.presto.spi.plan.JoinType.INNER;
 import static com.facebook.presto.sql.Optimizer.PlanStage.OPTIMIZED_AND_VALIDATED;
@@ -76,9 +78,40 @@ public abstract class AbstractCostBasedPlanTest
         assertEquals(generateQueryPlan(read(queryResourcePath)), read(getQueryPlanResourcePath(queryResourcePath)));
     }
 
+    @Test(dataProvider = "getQueriesDataProvider")
+    public void histogramsPlansMatch(String queryResourcePath)
+    {
+        String sql = read(queryResourcePath);
+        Session histogramSession = Session.builder(getQueryRunner().getDefaultSession())
+                .setSystemProperty(OPTIMIZER_USE_HISTOGRAMS, "true")
+                .build();
+        Session noHistogramSession = Session.builder(getQueryRunner().getDefaultSession())
+                .setSystemProperty(OPTIMIZER_USE_HISTOGRAMS, "false")
+                .build();
+        String regularPlan = generateQueryPlan(sql, noHistogramSession);
+        String histogramPlan = generateQueryPlan(sql, histogramSession);
+        if (!regularPlan.equals(histogramPlan)) {
+            assertEquals(histogramPlan, read(getHistogramPlanResourcePath(getQueryPlanResourcePath(queryResourcePath))));
+        }
+    }
+
     private String getQueryPlanResourcePath(String queryResourcePath)
     {
         return queryResourcePath.replaceAll("\\.sql$", ".plan.txt");
+    }
+
+    private String getHistogramPlanResourcePath(String regularPlanResourcePath)
+    {
+        Path root = Paths.get(regularPlanResourcePath);
+        return root.getParent().resolve("histogram/" + root.getFileName()).toString();
+    }
+
+    private Path getResourceWritePath(String queryResourcePath)
+    {
+        return Paths.get(
+                getSourcePath().toString(),
+                "src/test/resources",
+                getQueryPlanResourcePath(queryResourcePath));
     }
 
     public void generate()
@@ -90,12 +123,24 @@ public abstract class AbstractCostBasedPlanTest
                     .parallel()
                     .forEach(queryResourcePath -> {
                         try {
-                            Path queryPlanWritePath = Paths.get(
-                                    getSourcePath().toString(),
-                                    "src/test/resources",
-                                    getQueryPlanResourcePath(queryResourcePath));
+                            Path queryPlanWritePath = getResourceWritePath(queryResourcePath);
                             createParentDirs(queryPlanWritePath.toFile());
-                            write(generateQueryPlan(read(queryResourcePath)).getBytes(UTF_8), queryPlanWritePath.toFile());
+                            Session histogramSession = Session.builder(getQueryRunner().getDefaultSession())
+                                    .setSystemProperty(OPTIMIZER_USE_HISTOGRAMS, "true")
+                                    .build();
+                            Session noHistogramSession = Session.builder(getQueryRunner().getDefaultSession())
+                                    .setSystemProperty(OPTIMIZER_USE_HISTOGRAMS, "false")
+                                    .build();
+                            String sql = read(queryResourcePath);
+                            String regularPlan = generateQueryPlan(sql, noHistogramSession);
+                            String histogramPlan = generateQueryPlan(sql, histogramSession);
+                            write(regularPlan.getBytes(UTF_8), queryPlanWritePath.toFile());
+                            // write out the histogram plan if it differs
+                            if (!regularPlan.equals(histogramPlan)) {
+                                Path histogramPlanWritePath = getResourceWritePath(getHistogramPlanResourcePath(queryResourcePath));
+                                createParentDirs(histogramPlanWritePath.toFile());
+                                write(histogramPlan.getBytes(UTF_8), histogramPlanWritePath.toFile());
+                            }
                             System.out.println("Generated expected plan for query: " + queryResourcePath);
                         }
                         catch (IOException e) {
@@ -120,10 +165,15 @@ public abstract class AbstractCostBasedPlanTest
 
     private String generateQueryPlan(String query)
     {
+        return generateQueryPlan(query, getQueryRunner().getDefaultSession());
+    }
+
+    private String generateQueryPlan(String query, Session session)
+    {
         String sql = query.replaceAll("\\s+;\\s+$", "")
                 .replace("${database}.${schema}.", "")
                 .replace("\"${database}\".\"${schema}\".\"${prefix}", "\"");
-        Plan plan = plan(sql, OPTIMIZED_AND_VALIDATED, false);
+        Plan plan = plan(session, sql, OPTIMIZED_AND_VALIDATED, false);
 
         JoinOrderPrinter joinOrderPrinter = new JoinOrderPrinter();
         plan.getRoot().accept(joinOrderPrinter, 0);
--- a/presto-benchto-benchmarks/src/test/resources/sql/presto/tpcds/histogram/q85.plan.txt
+++ b/presto-benchto-benchmarks/src/test/resources/sql/presto/tpcds/histogram/q85.plan.txt
@@ -1,0 +1,38 @@
+local exchange (GATHER, SINGLE, [])
+    remote exchange (GATHER, SINGLE, [])
+        final aggregation over (r_reason_desc)
+            local exchange (GATHER, SINGLE, [])
+                remote exchange (REPARTITION, HASH, [r_reason_desc])
+                    partial aggregation over (r_reason_desc)
+                        join (INNER, REPLICATED):
+                            join (INNER, REPLICATED):
+                                join (INNER, PARTITIONED):
+                                    remote exchange (REPARTITION, HASH, [cd_demo_sk, cd_education_status, cd_marital_status])
+                                        scan customer_demographics
+                                    local exchange (GATHER, SINGLE, [])
+                                        remote exchange (REPARTITION, HASH, [cd_education_status_3, cd_marital_status_2, wr_refunded_cdemo_sk])
+                                            join (INNER, PARTITIONED):
+                                                remote exchange (REPARTITION, HASH, [wr_refunded_addr_sk])
+                                                    join (INNER, PARTITIONED):
+                                                        remote exchange (REPARTITION, HASH, [ws_item_sk, ws_order_number])
+                                                            join (INNER, REPLICATED):
+                                                                scan web_sales
+                                                                local exchange (GATHER, SINGLE, [])
+                                                                    remote exchange (REPLICATE, BROADCAST, [])
+                                                                        scan date_dim
+                                                        local exchange (GATHER, SINGLE, [])
+                                                            remote exchange (REPARTITION, HASH, [wr_item_sk, wr_order_number])
+                                                                join (INNER, REPLICATED):
+                                                                    scan web_returns
+                                                                    local exchange (GATHER, SINGLE, [])
+                                                                        remote exchange (REPLICATE, BROADCAST, [])
+                                                                            scan customer_demographics
+                                                local exchange (GATHER, SINGLE, [])
+                                                    remote exchange (REPARTITION, HASH, [ca_address_sk])
+                                                        scan customer_address
+                                local exchange (GATHER, SINGLE, [])
+                                    remote exchange (REPLICATE, BROADCAST, [])
+                                        scan web_page
+                            local exchange (GATHER, SINGLE, [])
+                                remote exchange (REPLICATE, BROADCAST, [])
+                                    scan reason
--- a/presto-hive/src/test/java/com/facebook/presto/hive/TestHiveIntegrationSmokeTest.java
+++ b/presto-hive/src/test/java/com/facebook/presto/hive/TestHiveIntegrationSmokeTest.java
@@ -4435,39 +4435,39 @@ public class TestHiveIntegrationSmokeTest
 
         assertQuery(format("SHOW STATS FOR (SELECT * FROM %s WHERE p_varchar = 'p1')", tableName),
                 "SELECT * FROM VALUES " +
-                        "('c_boolean', null, 2.0E0, 0.5E0, null, null, null), " +
-                        "('c_bigint', null, 2.0E0, 0.5E0, null, '0', '1'), " +
-                        "('c_double', null, 2.0E0, 0.5E0, null, '1.2', '2.2'), " +
-                        "('c_timestamp', null, 2.0E0, 0.5E0, null, null, null), " +
-                        "('c_varchar', 8.0E0, 2.0E0, 0.5E0, null, null, null), " +
-                        "('c_varbinary', 8.0E0, null, 0.5E0, null, null, null), " +
-                        "('c_array', 176.0E0, null, 0.5, null, null, null), " +
-                        "('p_varchar', 8.0E0, 1.0E0, 0.0E0, null, null, null), " +
-                        "(null, null, null, null, 4.0E0, null, null)");
+                        "('c_boolean', null, 2.0E0, 0.5E0, null, null, null, null), " +
+                        "('c_bigint', null, 2.0E0, 0.5E0, null, '0', '1', null), " +
+                        "('c_double', null, 2.0E0, 0.5E0, null, '1.2', '2.2', null), " +
+                        "('c_timestamp', null, 2.0E0, 0.5E0, null, null, null, null), " +
+                        "('c_varchar', 8.0E0, 2.0E0, 0.5E0, null, null, null, null), " +
+                        "('c_varbinary', 8.0E0, null, 0.5E0, null, null, null, null), " +
+                        "('c_array', 176.0E0, null, 0.5, null, null, null, null), " +
+                        "('p_varchar', 8.0E0, 1.0E0, 0.0E0, null, null, null, null), " +
+                        "(null, null, null, null, 4.0E0, null, null, null)");
         assertQuery(format("SHOW STATS FOR (SELECT * FROM %s WHERE p_varchar = 'p2')", tableName),
                 "SELECT * FROM VALUES " +
-                        "('c_boolean', null, 2.0E0, 0.5E0, null, null, null), " +
-                        "('c_bigint', null, 2.0E0, 0.5E0, null, '1', '2'), " +
-                        "('c_double', null, 2.0E0, 0.5E0, null, '2.3', '3.3'), " +
-                        "('c_timestamp', null, 2.0E0, 0.5E0, null, null, null), " +
-                        "('c_varchar', 8.0E0, 2.0E0, 0.5E0, null, null, null), " +
-                        "('c_varbinary', 8.0E0, null, 0.5E0, null, null, null), " +
-                        "('c_array', 96.0E0, null, 0.5, null, null, null), " +
-                        "('p_varchar', 8.0E0, 1.0E0, 0.0E0, null, null, null), " +
-                        "(null, null, null, null, 4.0E0, null, null)");
+                        "('c_boolean', null, 2.0E0, 0.5E0, null, null, null, null), " +
+                        "('c_bigint', null, 2.0E0, 0.5E0, null, '1', '2', null), " +
+                        "('c_double', null, 2.0E0, 0.5E0, null, '2.3', '3.3', null), " +
+                        "('c_timestamp', null, 2.0E0, 0.5E0, null, null, null, null), " +
+                        "('c_varchar', 8.0E0, 2.0E0, 0.5E0, null, null, null, null), " +
+                        "('c_varbinary', 8.0E0, null, 0.5E0, null, null, null, null), " +
+                        "('c_array', 96.0E0, null, 0.5, null, null, null, null), " +
+                        "('p_varchar', 8.0E0, 1.0E0, 0.0E0, null, null, null, null), " +
+                        "(null, null, null, null, 4.0E0, null, null, null)");
 
         // non existing partition
         assertQuery(format("SHOW STATS FOR (SELECT * FROM %s WHERE p_varchar = 'p3')", tableName),
                 "SELECT * FROM VALUES " +
-                        "('c_boolean', null, 0E0, 0E0, null, null, null), " +
-                        "('c_bigint', null, 0E0, 0E0, null, null, null), " +
-                        "('c_double', null, 0E0, 0E0, null, null, null), " +
-                        "('c_timestamp', null, 0E0, 0E0, null, null, null), " +
-                        "('c_varchar', 0E0, 0E0, 0E0, null, null, null), " +
-                        "('c_varbinary', null, 0E0, 0E0, null, null, null), " +
-                        "('c_array', null, 0E0, 0E0, null, null, null), " +
-                        "('p_varchar', 0E0, 0E0, 0E0, null, null, null), " +
-                        "(null, null, null, null, 0E0, null, null)");
+                        "('c_boolean', null, 0E0, 0E0, null, null, null, null), " +
+                        "('c_bigint', null, 0E0, 0E0, null, null, null, null), " +
+                        "('c_double', null, 0E0, 0E0, null, null, null, null), " +
+                        "('c_timestamp', null, 0E0, 0E0, null, null, null, null), " +
+                        "('c_varchar', 0E0, 0E0, 0E0, null, null, null, null), " +
+                        "('c_varbinary', null, 0E0, 0E0, null, null, null, null), " +
+                        "('c_array', null, 0E0, 0E0, null, null, null, null), " +
+                        "('p_varchar', 0E0, 0E0, 0E0, null, null, null, null), " +
+                        "(null, null, null, null, 0E0, null, null, null)");
 
         assertUpdate(format("DROP TABLE %s", tableName));
     }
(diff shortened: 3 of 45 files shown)


ZacBlanco:
A few TODOs just to give an idea of where I want to improve this PR:

1. Add a configuration property to choose to use histogram calculation over the previous logic - potentially with a log message when configured if the new logic is different than the old logic? 
  - Do you any thoughts on whether I should include the log message? It was useful for debugging but not sure if it would be good to include for the PR to upstream.
2. Clean up the logic inside of `HistogramCalculator`. The existing implementation is somewhat hacky and I did what I needed to do in order to pass the unit tests, but I think there are some easy simplifications of the logic. One of the challenges is that by using `Estimate` as the result for mins/max we lose some granularity as we don't know if an Unknown value is Infinite or NaN, which caused the logic to become quite a bit more complex than in the StatisticRange class.

presto-spi/src/main/java/com/facebook/presto/spi/statistics/ConnectorHistogram.java
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * Returns the total number of distinct values in the distribution which
+     * occur in the distribution at the value corresponding to the given
+     * percentile.
+     *
+     * @param percentile the percentile to query
+     * @return the number of distinct values falling less than the percentile
+     */
+    Estimate cumulativeDistinctValues(double percentile);
```
aaneja, line 63: Did you find a histogram implementation that supports this ? If not, lets punt on having this in the interface for now

presto-spi/src/main/java/com/facebook/presto/spi/statistics/ConnectorHistogram.java
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * @return the total number of distinct values in the distribution.
+     */
+    Estimate distinctValues();
```
aaneja, line 68: Do we need this  ? By definition this is `cumilativeDistinctValues(1.0)` ?

presto-main/src/test/java/com/facebook/presto/cost/TestHistogramCalculator.java
```diff
@@ -0,0 +1,157 @@
...
+    {
+        return new StatisticRange(low, high, NaN);
+    }
+
+    private static StatisticRange unboundedRange(double distinctValues)
+    {
+        return new StatisticRange(NEGATIVE_INFINITY, POSITIVE_INFINITY, distinctValues);
+    }
+
+    private static void assertFilterFactor(StatisticRange range, ConnectorHistogram histogram, Estimate expected)
```
aaneja, line 133: nit: Keep the call list order same as below : `Estimate expected, StatisticRange range, ConnectorHistogram histogram)`

aaneja (dismissed):
LGTM % some questions

presto-spi/src/main/java/com/facebook/presto/spi/statistics/ConnectorHistogram.java
ZacBlanco: The T-digest implementation of this class would not be able to support this. However, with the modifications I mentioned in my presentation the other week, it could be possible.

If we create a bucket-based histogram implementation (e.g. reconstructing Spark's histogram data structure) we might be able to properly compute this.

presto-spi/src/main/java/com/facebook/presto/spi/statistics/ConnectorHistogram.java
ZacBlanco: yeah, I agree we can probably just leave this out

presto-spi/src/main/java/com/facebook/presto/spi/statistics/ColumnStatisticType.java
```diff
@@ -21,7 +21,8 @@ public enum ColumnStatisticType
     NUMBER_OF_DISTINCT_VALUES("approx_distinct"),
     NUMBER_OF_NON_NULL_VALUES("count"),
     NUMBER_OF_TRUE_VALUES("count_if"),
-    TOTAL_SIZE_IN_BYTES("sum_data_size_for_stats");
+    TOTAL_SIZE_IN_BYTES("sum_data_size_for_stats"),
+    HISTOGRAM("tdigest_agg");
```
ZacBlanco, line 25: I've put this here for now as it's the only available function currently that can generate a histogram in Presto. This may be updated in a follow-up PR to something like KLL

presto-spi/src/main/java/com/facebook/presto/spi/statistics/ConnectorHistogram.java
ZacBlanco: I removed this from the API for now

ZacBlanco:
The header check is failing on `presto-benchto-benchmarks/src/test/resources/sql/presto/tpcds/q85.plan.txt`. This file never had a license header and is only used for checking test output. I think the check can be safely ignored.

aaneja (commented):
LGTM mostly. Still reviewing tests

presto-hive/src/test/java/com/facebook/presto/hive/TestShowStats.java
```diff
@@ -44,97 +44,97 @@ public void testShowStats()
...
+                        "   (null, null, null, null, 5.0, null, null, null))");
 
         assertQuery("SHOW STATS FOR (SELECT * FROM nation_partitioned WHERE regionkey > 0 and regionkey < 4)",
                 "SELECT * FROM (VALUES " +
-                        "   ('regionkey', null, 3.0, 0.0, null, 1, 3), " +
-                        "   ('nationkey', null, 5.0, 0.0, null, 1, 24), " +
-                        "   ('name', 109.0, 5.0, 0.0, null, null, null), " +
-                        "   ('comment', 1199.0, 5.0, 0.0, null, null, null), " +
-                        "   (null, null, null, null, 15.0, null, null))");
+                        "   ('regionkey', null, 3.0, 0.0, null, 1, 3, null), " +
```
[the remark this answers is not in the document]
ZacBlanco: Currently, it's up to the connector to support collecting them. If they are not collected/returned by a connector, they won't appear in show stats. I would need to add support for histograms in the Hive connector to write a test where they appear, which is out of scope for this PR.

After this PR is merged I will introduce support in the Iceberg connector which should include a test for show stats.

ZacBlanco:
I ran a quick experiment using TPC-DS SF1 to verify the performance isn't changed much with real data.
image

I've also provided the raw data for these plots

[tpcds-sf1-results.tar.gz](https://github.com/prestodb/presto/files/14149532/tpcds-sf1-results.tar.gz)

presto-benchto-benchmarks/src/test/java/com/facebook/presto/sql/planner/AbstractCostBasedPlanTest.java
```diff
@@ -78,7 +78,7 @@ public void test(String queryResourcePath)
 
     private String getQueryPlanResourcePath(String queryResourcePath)
     {
-        return queryResourcePath.replaceAll("\\.sql$", ".plan.txt");
+        return queryResourcePath.replaceAll("\\.sql$", ".plan");
```
aaneja, line 81: Is this rename strictly necessary ?

ZacBlanco: I put it there just to appease the CI header check. For whatever reason we have a CI license header check in the presto repo that uses a script from the velox repo. So the only way to get the PR check to pass is to modify the script in the velox repo, change the github check, or modify the filenames so the velox header check doesn't detect it.

I chose the latter, but can revert that commit. I did it at least temporarily to make the CI happy.

aaneja: >we have a CI license header check in the presto repo that uses a script from the velox repo.
Let's make an issue for this

ZacBlanco: I've filed #21875

aaneja:
Just some nits and high level questions. I think the [header check](https://app.circleci.com/pipelines/github/prestodb/presto/12906/workflows/4640dbd8-6c09-4256-9ec4-6a535fd71310/jobs/49822) is failing again btw

ZacBlanco:
> ... I think the [header check](https://app.circleci.com/pipelines/github/prestodb/presto/12906/workflows/4640dbd8-6c09-4256-9ec4-6a535fd71310/jobs/49822) is failing again btw

I reverted the commit that originally fixed the header check. #21875 was filed for it. Deepak said he would address it but that hasn't been completed yet. For now, We should be able to ignore the `ci/header-check`.

steveburnett:
Suggest rephrasing release note entry to something like the following, based on the [Order of changes](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines#order-of-changes) in the [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines): 

```
== RELEASE NOTES ==

General Changes
* Add histogram ColumnStatisticType to Presto for the optimizer and connectors that support them
```

presto-main/src/main/java/com/facebook/presto/cost/UniformDistributionHistogram.java
```diff
@@ -0,0 +1,121 @@
...
+    public double getHighValue()
+    {
+        return highValue;
+    }
+
+    @Override
+    public Estimate cumulativeProbability(double value, boolean inclusive)
+    {
+        if (isNaN(lowValue) ||
+                isNaN(highValue) ||
```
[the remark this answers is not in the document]
ZacBlanco: The constructor actually allows NaN values. It checks if `low <= high` iff both values are not `NaN`

presto-main/src/main/java/com/facebook/presto/cost/StatsNormalizer.java
```diff
@@ -125,6 +125,9 @@ private VariableStatsEstimate normalizeVariableStats(VariableReferenceExpression
         return VariableStatsEstimate.buildFrom(variableStats)
                 .setDistinctValuesCount(distinctValuesCount)
                 .setNullsFraction(nullsFraction)
+                .setHistogram(DisjointRangeDomainHistogram
+                        .addDisjunction(variableStats.getHistogram(),
+                                new StatisticRange(variableStats.getLowValue(), variableStats.getHighValue(), distinctValuesCount)))
                 .build();
```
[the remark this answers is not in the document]
ZacBlanco: good point. Fixed

presto-main/src/test/java/com/facebook/presto/cost/TestDisjointRangeDomainHistogram.java
```diff
@@ -0,0 +1,288 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static org.testng.Assert.assertEquals;
+
+public class TestDisjointRangeDomainHistogram
+        extends TestHistogram
+{
+    /**
+     * A uniform base with 2 ranges that are within the
```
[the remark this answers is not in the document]
ZacBlanco: fixed

presto-main/src/test/java/com/facebook/presto/cost/TestDisjointRangeDomainHistogram.java
```diff
@@ -0,0 +1,288 @@
...
+        ConnectorHistogram source = new UniformDistributionHistogram(0, 100);
+        ConnectorHistogram constrained = DisjointRangeDomainHistogram
+                .addDisjunction(source, StatisticRange.fromRange(Range.open(0d, 25d)));
+        constrained = DisjointRangeDomainHistogram
+                .addDisjunction(constrained, StatisticRange.fromRange(Range.open(75d, 100d)));
+        assertEquals(87.5, constrained.inverseCumulativeProbability(0.75).getValue());
+        assertEquals(0.0, constrained.inverseCumulativeProbability(0.0).getValue());
+        assertEquals(100, constrained.inverseCumulativeProbability(1.0).getValue());
+        assertEquals(25, constrained.inverseCumulativeProbability(0.5).getValue());
+    }
```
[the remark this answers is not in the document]
ZacBlanco: Fixed

presto-main/src/main/java/com/facebook/presto/cost/PlanNodeStatsEstimateMath.java
```diff
@@ -264,6 +267,7 @@ private static VariableStatsEstimate addColumnStats(
                 .setStatisticsRange(sum)
                 .setAverageRowSize(newAverageRowSize)
                 .setNullsFraction(newNullsFraction)
+                .setHistogram(DisjointRangeDomainHistogram.addDisjunction(leftStats.getHistogram(), rightRange))
```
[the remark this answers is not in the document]
ZacBlanco: The `RangeAdditionStrategy` _almost_ always expands the entire histogram range - The callees all reference methods from `StatisticRange#addSstatsAnd<X>DistinctValues` (or so I thought). It's just the distinct values that change with the strategy, so it wouldn't have an effect on the histogram's output.

See https://github.com/prestodb/presto/blob/61794f05632a8bf84a7f3b2b97916e127d403713/presto-main/src/main/java/com/facebook/presto/cost/StatisticRange.java#L152-L173

Now that you mentioned this though, I did see `addStatsAndIntersect` which doesn't actually always expand the range. So I will need to handle this case. Though it's only called when an `IntersectNode` is in the plan, which I don't believe any of the standard benchmarks generate.

The latest commit fixes this by introducing an enum wrapper around the function instead. We now check for `INTERSECT` to do the conjunction, otherwise use disjunction.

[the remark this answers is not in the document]
ZacBlanco: See 

- https://github.com/ZacBlanco/prestodb/blob/7c44a1ea7fd3dc0a296e719d9b559c02bf65da09/presto-main/src/main/java/com/facebook/presto/cost/PlanNodeStatsEstimateMath.java#L152-L169
- https://github.com/ZacBlanco/prestodb/blob/7c44a1ea7fd3dc0a296e719d9b559c02bf65da09/presto-main/src/main/java/com/facebook/presto/cost/PlanNodeStatsEstimateMath.java#L287-L289

presto-main/src/main/java/com/facebook/presto/cost/UniformDistributionHistogram.java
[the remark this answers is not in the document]
aaneja: My bad, read the verify incorrectly

presto-main/src/main/java/com/facebook/presto/cost/PlanNodeStatsEstimateMath.java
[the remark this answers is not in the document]
aaneja: Should we add tests `TestPlanNodeStatsEstimateMath` to prove correctness ?

aaneja:
From a testing perspective, I think we should add some E2E SQL tests (ie `presto-tests`) - these can use the existing abstract tests but with `optimizer_use_histograms` turned on to prove no regressions
Additionally, we could add some presto-main unit tests targeted towards filter & join selectivity. If we can augment the plan matcher to match on node cardinality, we can add tighter assertions w.r.t selectivity with and without histograms

ZacBlanco:
Currently, I have enabled this feature by default because it should be backwards compatible in nearly all cases. So the tests in `presto-main` and `presto-tests` should be running with this on already.

> Additionally, we could add some presto-main unit tests targeted towards filter & join selectivity. If we can augment the plan matcher to match on node cardinality, we can add tighter assertions w.r.t selectivity with and without histograms

I think it's probably a good idea to check that plans are equivalent with the flag on/off but I feel the gain is marginal given that we already have many unit tests as well as the TPC-DS/H cost plan tests. I feel it might be better to add similar tests to the TPC-DS/H cost plan tests but check the cost estimates rather than plan shape?

aaneja:
My bad - I didn't realize histograms option was defaulting to ON. Should we be shipping with this turned on though ? There have been regressions due to optimizer features, so it's best to have users opt-in and report results to us rather than surprise them

>I feel it might be better to add similar tests to the TPC-DS/H cost plan tests but check the cost estimates rather than plan shape  

Yes this is exactly what I meant by matching on node cardinality (row counts). If we augment the [PlanMatchPattern](https://github.com/prestodb/presto/blob/eb94b3f4b83dae444519c618b42b4c853d4380b0/presto-main/src/test/java/com/facebook/presto/sql/planner/assertions/PlanMatchPattern.java) and related classes to match for row counts, we could add a [planner test](https://github.com/prestodb/presto/tree/eb94b3f4b83dae444519c618b42b4c853d4380b0/presto-main/src/test/java/com/facebook/presto/sql/planner) that does strong assertions on expected cardinality

ZacBlanco:
IMO, setting it to on by default would be ideal. My worry is that people might forget to set the property once we introduce support in the connectors. It would make for a confusing situation where even if the metastore has histograms stored and generated by presto that they aren't utilized.

I understand we tend to tend to be conservative with these types so I've updated the default to off. In a later PR where I intend to introduce the support in Iceberg, I will make sure the documentation very clearly states the session property should be set to `true` to take advantage of the histograms.

---

Re: testing

Now that the default is off, I've explicitly set up some tests to run in a parameterized way such that the histogram and non-histogram code paths are stressed.

See:

- AbstractCostBasedPlanTest
- AbstractTestComparisonStatsCalculator
- AbstractTestFilterStatsCalculator

These suites are where I had to make changes previously due to cost changes when histograms defaulted to on, so the new tests changes should ensure that we don't have any major regressions.

presto-spi/src/main/java/com/facebook/presto/spi/statistics/ColumnStatisticType.java
tdcmeehan: qdigest_agg should also be able to do this.

ZacBlanco: Ah you're right. Currently, tdigest  is just the default value since one is required. Connectors can (need?) to specify the function used for the histogram statistic anyways since the metastore(s) might support only some formats (e.g. Hive only supporting KLL). I don't think the default here matters all that much

presto-main/src/main/java/com/facebook/presto/cost/PlanNodeStatsEstimateMath.java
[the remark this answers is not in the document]
ZacBlanco: done

ZacBlanco:
@aaneja and I had a discussion about the disjoint histogram which had us asking the question about what Spark does for filter estimation in `OR` cases when histograms are involved.

The relevant block of code is here: https://github.com/apache/spark/blob/ce93c9fd86715e2479552628398f6fc11e83b2af/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/FilterEstimation.scala#L82-L121

They don't change the logic when histograms are involved or not histograms or not. Basically the filter factor is always calculated as `percent1 + percent2 - (percent1 * percent2)`. There is a separate chunk of code for calculations using histograms if they exist

- For numeric comparisons: https://github.com/apache/spark/blob/ce93c9fd86715e2479552628398f6fc11e83b2af/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/FilterEstimation.scala#L493-L528
- For equality: https://github.com/apache/spark/blob/ce93c9fd86715e2479552628398f6fc11e83b2af/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/statsEstimation/FilterEstimation.scala#L555-L581

aaneja (approved):

ClarenceThreepwood (dismissed):

presto-main/src/main/java/com/facebook/presto/sql/rewrite/ShowStatsRewrite.java
```diff
@@ -303,6 +305,7 @@ private Row createColumnStatsRow(String columnName, Type type, ColumnStatistics
             rowValues.add(NULL_DOUBLE);
             rowValues.add(toStringLiteral(type, columnStatistics.getRange().map(DoubleRange::getMin)));
             rowValues.add(toStringLiteral(type, columnStatistics.getRange().map(DoubleRange::getMax)));
+            rowValues.add(columnStatistics.getHistogram().map(Objects::toString).<Expression>map(StringLiteral::new).orElse(NULL_VARCHAR));
```
jaystarshot, line 308: We might want to show the histogram boundaries like [postgres](https://www.postgresql.org/docs/current/view-pg-stats.html)

presto-main/src/main/java/com/facebook/presto/cost/VariableStatsEstimate.java
```diff
@@ -237,9 +265,16 @@ public Builder setDistinctValuesCount(double distinctValuesCount)
...
+        {
+            this.histogram = Optional.of(histogram);
+            return this;
+        }
+
         public VariableStatsEstimate build()
         {
-            return new VariableStatsEstimate(lowValue, highValue, nullsFraction, averageRowSize, distinctValuesCount);
+            return new VariableStatsEstimate(lowValue, highValue, nullsFraction, averageRowSize, distinctValuesCount,
+                    histogram.orElseGet(() -> new UniformDistributionHistogram(lowValue, highValue)));
```
jaystarshot, line 277: Here we will set and use the default histogram even without the session property. Is that intended or should we make sure that we use a noop version without?

presto-main/src/main/java/com/facebook/presto/cost/HistogramCalculator.java
```diff
@@ -0,0 +1,167 @@
...
+     *
+     * @param range the intersecting range with the histogram
+     * @param histogram the source histogram
+     * @param totalDistinctValues the total number of distinct values in the domain of the histogram
+     * @param useHeuristics whether to return heuristic values based on constants and/or distinct
+     * value counts. If false, {@link Estimate#unknown()} will be returned in any case a
+     * heuristic would have been used
+     * @return an estimate, x, where 0.0 <= x <= 1.0.
+     */
+    public static Estimate calculateFilterFactor(StatisticRange range, ConnectorHistogram histogram, Estimate totalDistinctValues, boolean useHeuristics)
```
jaystarshot, line 48: Nit: Is it better to have this control given to the ConnectorHistogram itself?

presto-main/src/main/java/com/facebook/presto/cost/VariableStatsEstimate.java
ZacBlanco: This is a good point. It will populate the histogram in `VariableStatsEstimate` without the property set. However, all of the code which actually uses the histograms for calculations is gated by the property here:

https://github.com/prestodb/presto/pull/21236/commits/838628b0574c3606302ba9e9f219b53c201d3c70#diff-3e8f0b799abcb81b4285a5d98fde631da64d29d6fd95e952b3583ffa949301f6R170

I think the tough thing about changing this is that we would need to pass `Session` everywhere in order to check the property to determine whether or not the histogram should be included on the stats objects. This isn't the only location in the code where it includes histograms. There are others such as [here](https://github.com/prestodb/presto/pull/21236/commits/838628b0574c3606302ba9e9f219b53c201d3c70#diff-22fa802402957ec23b9dfd10c46f8058045665fa29a8e9abe67398ae6be6f071R136) and [here](https://github.com/prestodb/presto/pull/21236/commits/838628b0574c3606302ba9e9f219b53c201d3c70#diff-22fa802402957ec23b9dfd10c46f8058045665fa29a8e9abe67398ae6be6f071R287-R290). As long as the calculation done in `ComparisonStatsCalculator` is the only location that histograms are used, I think it is safe to just add the histogram here by default

presto-spi/src/main/java/com/facebook/presto/spi/statistics/ConnectorHistogram.java
```diff
@@ -0,0 +1,64 @@
...
+ * domains map to real values.
+ * <br>
+ * Null values should not be represented in underlying histogram implementation.
+ * When calculating filter statistics using the {@link ColumnStatisticType#NUMBER_OF_NON_NULL_VALUES}
+ * are used to account for nulls in cost-based calculations.
+ *
+ * @see ColumnStatisticType#NUMBER_OF_NON_NULL_VALUES
+ */
+@JsonTypeInfo(use = JsonTypeInfo.Id.MINIMAL_CLASS, property = "@class")
+public interface ConnectorHistogram
```
jaystarshot, line 34: Are there examples of connector specific histograms. I'm curious about the reasons why different connectors might have distinct, incompatible histograms?

presto-main/src/main/java/com/facebook/presto/sql/rewrite/ShowStatsRewrite.java
ZacBlanco: Currently it is left up to the implementation's `toString` method to do this. With the `UniformHistogram`, this does show the boundaries. Other planned histogram implementation such as with T-Digests or KLL sketches don't have "boundaries" in the traditional sense of binned histograms because they choose to represent the data differently. I think it makes more sense to leave it to the `ConnectorHistogram` implementation. If there were a binned representation I would want to print the boundaries.

presto-spi/src/main/java/com/facebook/presto/spi/statistics/ConnectorHistogram.java
ZacBlanco: Hive 4 is planning to support storage of Apache Datasketches KLL sketches. Spark has its own standard binned histogram representation that is stored in a Base64-encoded binary format on the table properties. Presto also supports creating T-Digests through `tdigest_agg`. Iceberg could support any of these histogram representations in its Puffin file format. There are numerous histograms implementations as well as supported storage locations usually dependent upon the connector. I want to support reading or writing any type of histogram from any type of storage.

The actual implementation of any histogram is not included in this PR though. It is just the framework for the optimizer to make use of them.

presto-main/src/main/java/com/facebook/presto/cost/HistogramCalculator.java
ZacBlanco: I think that would lead to a lot of duplicate code in the implementations. At the core, histograms are able to answer basic questions about the data distribution which are the `cumulativeProbability` and `inverseCumulativeProbability` methods on the `ConnectorHistogram` interface. As long as those two methods can be implemented properly by a histogram you can answer any question the optimizer needs for comparison cost calculations.

Is there a specific reason that you think this should be handled by the individual implementations? Pushing this down to the implementation could not only have more potential for duplicate code but also lead to more errors if the implementations aren't written correctly.

presto-main/src/main/java/com/facebook/presto/cost/VariableStatsEstimate.java
jaystarshot: What if we store session in the implementations of the **ConnectorHistorgram** and throw exceptions when used without?

presto-main/src/main/java/com/facebook/presto/sql/rewrite/ShowStatsRewrite.java
jaystarshot: Sg

presto-main/src/main/java/com/facebook/presto/cost/HistogramCalculator.java
jaystarshot: Sounds good, I was wondering adding this  as a default method but as you said its the same for all histograms

jaystarshot:
Overall looks good, @tdcmeehan can you please review the SPI changes. This PR doesn't include loading the histograms from connectors and the SPI changes are currently noop. So, I'm uncertain if this PR is the most appropriate for these changes.

presto-main/src/main/java/com/facebook/presto/cost/VariableStatsEstimate.java
```diff
@@ -79,6 +83,16 @@ public VariableStatsEstimate(
         checkArgument(distinctValuesCount >= 0 || isNaN(distinctValuesCount), "Distinct values count should be non-negative, got: %s", distinctValuesCount);
         // TODO normalize distinctValuesCount for an empty range (or validate it is already normalized)
         this.distinctValuesCount = distinctValuesCount;
+        this.histogram = histogram;
```
feilong-liu, line 86: Can histogram be null?
Use requireNotNull() in the assignment if null is not expected

presto-main/src/main/java/com/facebook/presto/cost/VariableStatsEstimate.java
ZacBlanco: Even if you stored the session in the histogram, it would still require changing all of the creation sites of the histogram (and hence the creation sites of `VariableStatisticsEstimate`). Also, the property controlling histogram use is defined in `SystemSessionProperties`. I don't think there is much benefit doing it that way.

I went through a few options and I think the one that will touch the smallest amount of code will require at least having conditionals at all call sites where the histograms are created/added to stats estimates in the optimizer. From a cursory search it is about ~10 locations, plus a good handful of function signature changes to pass the session. Other options such as changing `VariableStatisticEstimate` to store a `Session` member will require 1-200 of LoC change.

Aside from a future change accidentally using the histogram without gating it by the flag, is there any other specific concern with including the histograms on the `VariableStatsEstimate` even when the feature is disabled?

presto-benchto-benchmarks/src/test/resources/sql/presto/tpcds/histogram/q85.plan.txt
```diff
@@ -0,0 +1,38 @@
+local exchange (GATHER, SINGLE, [])
```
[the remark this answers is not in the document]
ZacBlanco: The only difference in calculations among all of the test cases in the code base were these two:

https://github.com/prestodb/presto/blob/838628b0574c3606302ba9e9f219b53c201d3c70/presto-main/src/test/java/com/facebook/presto/cost/TestFilterStatsCalculatorHistograms.java#L31-L70

Due to the current design of the histogram API, the histograms don't use the same heuristics as the non-histogram calculation in cases with infinite bounds on one side. This only occurs for tables without statistics. It's mainly because the use of `Estimate` in the histogram API doesn't have a way of distinguishing `NaN` from an infinite bound, which creates ambiguity when determining which heuristic value to utilize.

presto-hive/src/test/java/com/facebook/presto/hive/TestHiveIntegrationSmokeTest.java
```diff
@@ -4435,39 +4435,39 @@ public void testCollectColumnStatisticsOnCreateTable()
...
-                        "('c_boolean', null, 2.0E0, 0.5E0, null, null, null), " +
-                        "('c_bigint', null, 2.0E0, 0.5E0, null, '0', '1'), " +
-                        "('c_double', null, 2.0E0, 0.5E0, null, '1.2', '2.2'), " +
-                        "('c_timestamp', null, 2.0E0, 0.5E0, null, null, null), " +
-                        "('c_varchar', 8.0E0, 2.0E0, 0.5E0, null, null, null), " +
-                        "('c_varbinary', 8.0E0, null, 0.5E0, null, null, null), " +
-                        "('c_array', 176.0E0, null, 0.5, null, null, null), " +
-                        "('p_varchar', 8.0E0, 1.0E0, 0.0E0, null, null, null), " +
-                        "(null, null, null, null, 4.0E0, null, null)");
+                        "('c_boolean', null, 2.0E0, 0.5E0, null, null, null, null), " +
```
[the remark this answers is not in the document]
ZacBlanco: yes

presto-main/src/main/java/com/facebook/presto/cost/JoinStatsRule.java
```diff
@@ -246,12 +247,14 @@ private PlanNodeStatsEstimate filterByAuxiliaryClause(PlanNodeStatsEstimate stat
                 .setNullsFraction(0)
                 .setStatisticsRange(intersect)
                 .setDistinctValuesCount(retainedNdv)
+                .setHistogram(addConjunction(leftStats.getHistogram(), intersect))
```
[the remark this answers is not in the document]
ZacBlanco: Histogram should always be non-null. If it was supposed to be nullable I would add an `@Nullable` annotation.

If a histogram is not set in `VariableStatisticsEstimate.Builder`, then it will default to a `UniformDistributionHistogram` with bounds equivalent to the low/high value set in the builder. By default `lowValue` is -inf, `maxValue` is +inf if stats are not available. Otherwise, if they are it will use the correct low/high bounds in the uniform histogram. I can add a `requireNonNull` check in the constructor to ensure it is never null.

presto-main/src/main/java/com/facebook/presto/cost/PlanNodeStatsEstimateMath.java
```diff
@@ -119,16 +122,18 @@ public static PlanNodeStatsEstimate capStats(PlanNodeStatsEstimate stats, PlanNo
...
+            double newHigh = min(symbolStats.getHighValue(), capSymbolStats.getHighValue());
+            newSymbolStats.setLowValue(newLow);
+            newSymbolStats.setHighValue(newHigh);
 
             double numberOfNulls = stats.getOutputRowCount() * symbolStats.getNullsFraction();
             double capNumberOfNulls = cap.getOutputRowCount() * capSymbolStats.getNullsFraction();
             double cappedNumberOfNulls = min(numberOfNulls, capNumberOfNulls);
             double cappedNullsFraction = cappedRowCount == 0 ? 1 : cappedNumberOfNulls / cappedRowCount;
             newSymbolStats.setNullsFraction(cappedNullsFraction);
+            newSymbolStats.setHistogram(addConjunction(symbolStats.getHistogram(), new StatisticRange(newLow, newHigh, 0)));
```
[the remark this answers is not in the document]
ZacBlanco: See other comment - histogram should always be available, even without stats

presto-main/src/main/java/com/facebook/presto/cost/DisjointRangeDomainHistogram.java
```diff
@@ -0,0 +1,352 @@
...
+        implements ConnectorHistogram
+{
+    private final ConnectorHistogram source;
+    // use RangeSet as the internal representation of the ranges, but the constructor arguments
+    // use StatisticRange to support serialization and deserialization.
+    private final RangeSet<Double> rangeSet;
+    private final Range<Double> sourceSpan;
+
+    @JsonCreator
+    public DisjointRangeDomainHistogram(@JsonProperty ConnectorHistogram source, @JsonProperty Collection<StatisticRange> ranges)
```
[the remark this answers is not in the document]
ZacBlanco: Good catch. Thanks. Seems it is not actually required but I think it is better to explicitly include it

presto-benchto-benchmarks/src/test/resources/sql/presto/tpcds/histogram/q85.plan.txt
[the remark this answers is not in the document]
feilong-liu: Also to confirm, when we disable using histogram, the behavior of current CBO will stay the same, is this correct?

feilong-liu (dismissed):
LGTM.

Just to confirm, current CBO will not be affected with this feature to be off?

presto-benchto-benchmarks/src/test/resources/sql/presto/tpcds/histogram/q85.plan.txt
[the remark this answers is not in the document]
ZacBlanco: yes, the calculations stay the same as long as the flag is set to false (the default)

ZacBlanco:
> LGTM.
> 
> Just to confirm, current CBO will not be affected with this feature to be off?

Yes, see: https://github.com/prestodb/presto/pull/21236#discussion_r1540381711

presto-main/src/main/java/com/facebook/presto/cost/VariableStatsEstimate.java
jaystarshot: Okay I see, nope that was the only concern.

tdcmeehan (approved):

feilong-liu:
@ZacBlanco @tdcmeehan  We observe latency regression/query timeout caused by this PR, detail in https://github.com/prestodb/presto/pull/22661, and need to revert this PR

PR #21245, opened 2023-10-26
[native] Translate array agg config to velox

pranjalssh:
We need to add logic to presto native to translate deprecated.legacy-array-agg=true Presto config to presto.array_agg.ignore_nulls = true Velox config.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/common/Configs.cpp
+++ b/presto-native-execution/presto_cpp/main/common/Configs.cpp
@@ -192,6 +192,7 @@ SystemConfig::SystemConfig() {
           STR_PROP(kInternalCommunicationJwtEnabled, "false"),
           STR_PROP(kInternalCommunicationSharedSecret, ""),
           NUM_PROP(kInternalCommunicationJwtExpirationSeconds, 300),
+          BOOL_PROP(kUseLegacyArrayAgg, false),
       };
 }
 
@@ -497,6 +498,10 @@ int32_t SystemConfig::internalCommunicationJwtExpirationSeconds() const {
       .value();
 }
 
+bool SystemConfig::useLegacyArrayAgg() const {
+  return optionalProperty<bool>(kUseLegacyArrayAgg).value();
+}
+
 NodeConfig::NodeConfig() {
   registeredProps_ =
       std::unordered_map<std::string, folly::Optional<std::string>>{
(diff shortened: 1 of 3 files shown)


presto-native-execution/presto_cpp/main/common/Configs.h
```diff
@@ -339,8 +341,14 @@ class SystemConfig : public ConfigBase {
   static constexpr std::string_view kInternalCommunicationJwtExpirationSeconds{
       "internal-communication.jwt.expiration-seconds"};
 
+  /// Use legacy version of array_agg which ignores nulls.
```
xiaoxmeng, line 344:
```
nit: s/Use/Uses/
```

presto-native-execution/presto_cpp/main/common/Configs.h
```diff
@@ -526,7 +538,15 @@ class BaseVeloxQueryConfig : public ConfigBase {
...
   BaseVeloxQueryConfig();
 
+  virtual ~BaseVeloxQueryConfig() = default;
+
+  void initialize(const std::string& filePath) override;
+
   static BaseVeloxQueryConfig* instance();
+
+ private:
+  /// Update velox config with values from presto system config.
```
xiaoxmeng, line 548:
```
// for non-public comment in header.
```

presto-native-execution/presto_cpp/main/common/Configs.cpp
```diff
@@ -660,4 +665,14 @@ BaseVeloxQueryConfig* BaseVeloxQueryConfig::instance() {
   return instance.get();
 }
 
+void BaseVeloxQueryConfig::initialize(const std::string& filePath) {
+  ConfigBase::initialize(filePath);
+  update(*SystemConfig::instance());
```
xiaoxmeng, line 670: Do we need to update here if have called update(*SystemConfig::instance()) in ctor? Thanks!

Also can we test this in BaseVeloxQueryConfigTest? Thanks!

xiaoxmeng (approved):
@pranjalssh thanks for the change % comments. Could we add a test to capture this behavior? Thanks!

presto-native-execution/presto_cpp/main/common/Configs.cpp
pranjalssh: Initial ctor will load from default values. SystemConfig may be initialized from a file after BaseVeloxQueryConfig is created - so we load again.

PR #21253, opened 2023-10-26
[native pos] Rely on Spark to decide if stage writes to a shuffle

arhimondr:
## Description

Only register shuffle handles for shuffle stages by including stage id into shuffle handle key

## Motivation and Context

Remove extra conditions that may not always be accurate. 

For example a table writer stage is not always guaranteed to be consumed by a stage running on coordinator, hence may require to write into shuffle.

The extra conditions were necessary as sometimes the `nativeInputs.getShuffleWriteDescriptor()` is available for non shuffle stages due to a problem in PrestoSparkNativeExecutionShuffleManager:

1. `ShuffleManager#getWriter` is correctly called by the Spark engine only for shuffle output stages
2. `partitionIdToShuffleHandle` uses mapId as a key
3. `PrestoSparkNativeExecutionShuffleManager#getShuffleHandle` is called for a different stage (non shuffle stage), but since the key only contains "mapId" the method may return a handle even for a stage consumed by the driver.

The idea is to extend the key to also include the stage id to resolve this issue. Once `nativeInputs.getShuffleWriteDescriptor()` correctly returns `Optional#empty()` for non shuffle stages the extra checks in PrestoSparkNativeTaskExecutorFactory can be removed.

## Impact

Shuffle is always accurately enabled for shuffle stages

## Test Plan

Existing integration tests + additional test to cover corner case when extra conditions result into an incorrect decision

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeGeneralQueries.java
+++ b/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeGeneralQueries.java
@@ -1240,6 +1240,30 @@ public abstract class AbstractTestNativeGeneralQueries
         }
     }
 
+    @Test
+    public void testUnionAllInsert()
+    {
+        String tableName = generateRandomTableName();
+        try {
+            String union = "SELECT orderkey * 2 orderkey " +
+                    "FROM ( " +
+                    "  SELECT orderkey " +
+                    "  FROM orders " +
+                    "  UNION ALL " +
+                    "  SELECT orderkey " +
+                    "  FROM orders " +
+                    ") " +
+                    "UNION ALL " +
+                    "SELECT orderkey " +
+                    "FROM orders";
+            getQueryRunner().execute(format("CREATE TABLE %s AS %s", tableName, union));
+            assertQuery(format("SELECT * FROM %s", tableName), union);
+        }
+        finally {
+            dropTableIfExists(tableName);
+        }
+    }
+
     private void assertQueryResultCount(String sql, int expectedResultCount)
     {
         assertEquals(getQueryRunner().execute(sql).getRowCount(), expectedResultCount);
--- a/presto-native-execution/src/test/java/com/facebook/presto/spark/PrestoSparkNativeQueryRunnerUtils.java
+++ b/presto-native-execution/src/test/java/com/facebook/presto/spark/PrestoSparkNativeQueryRunnerUtils.java
@@ -17,8 +17,6 @@ import com.facebook.airlift.log.Logging;
 import com.facebook.presto.hive.metastore.Database;
 import com.facebook.presto.hive.metastore.ExtendedHiveMetastore;
 import com.facebook.presto.nativeworker.PrestoNativeQueryRunnerUtils;
-import com.facebook.presto.spark.classloader_interface.PrestoSparkNativeExecutionShuffleManager;
-import com.facebook.presto.spark.classloader_interface.PrestoSparkNativeExecutionShuffleManager.StageAndMapId;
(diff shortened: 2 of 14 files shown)


shrinidhijoshi:
@arhimondr Thanks for taking a closer look at this and fixing the root cause.

Along with adding `stageId` as part of the key should we also remove entries from `partitionIdToShuffleHandle` , `shuffleIdToBaseShuffleHandle` maps in `writer.stop()` ? 

It seems like doing either of these would fix the issue, but doing both would be cleaner.

arhimondr:
@shrinidhijoshi 

> It seems like doing either (1) or (2) would fix the root cause issue, but doing both would be cleaner. Is that understanding correct ?

I'm afraid (2) might not work when two tasks from two different stages are executed at the same time by the same worker node.

> remove entries from partitionIdToShuffleHandle , shuffleIdToBaseShuffleHandle maps in writer.stop() ?

Though I think it might be worth doing anyway to avoid leaving garbage behind that may use unnecessary memory. Let me try.

arhimondr:
@shrinidhijoshi Updated

shrinidhijoshi (approved):
Thanks for the updates. LGTM!

mbasmanova (approved):
Thanks. Assuming @shrinidhijoshi already reviewed throughly.

arhimondr:
@shrinidhijoshi Unfortunately I had to remove the `assertShuffleMetadata` check. 

Previously the check was verifying the type of the `ShuffleHandle`. I preserved this verification by implementing it as an assertion inside the PrestoSparkNativeExecutionShuffleManager.

Initially I decided to keep the `assertShuffleMetadata` and verify the state of the `PrestoSparkNativeExecutionShuffleManager` to make sure the handles are properly removed. 

However it creases a problem when tests are run on the CI. Since the tests are executing in parallel a race condition is possible (`assertShuffleMetadata` when another test is in a middle of execution). Since the `PrestoSparkNativeExecutionShuffleManager` global it fails as the handles maps are not empty.

Given the `PrestoSparkNativeExecutionShuffleManager` is a global static variable that is shared between test classes the only real way of preserving the check is to run tests sequentially in a single thread (as TestNG doesn't allow to register a global trigger). However running tests sequentially will increase the CI time significantly so I decided to remove the check for now.

I also tried to think of different options (for example somehow try to identify what shuffle handles belong to the query just executed) but couldn't find a clear way of doing so. 

Please let me know if it is still good to go.

shrinidhijoshi:
@arhimondr Yeah. I think it's ok to remove this. Looks like this test specifically relied on`PrestoSparkNativeExecutionShuffleManager` not cleaning up after itself. Now that this PR fixes that, this test becomes invalid. 

Also, would it make sense to add a unit test for PrestoSparkNativeExecutionShuffleManager. Which would test the sequence  - `registerShuffle` -> `getWriter` -> `writer.stop()` ?

arhimondr:
> Also, would it make sense to add a unit test for PrestoSparkNativeExecutionShuffleManager. Which would test the sequence - registerShuffle -> getWriter -> writer.stop() ?

Good point. I think we can add a unit test and assume that Spark will do the right thing and always properly call `Writer#stop`. Let me do that.

arhimondr:
Tried implementing a unit test. Needed to create a ShuffleHandle. To create a ShuffleHandle a ShuffleDependency must be declared. To create a ShuffleDependency i need to create a number of other objects eventually running in creating a SparkContext which fails with a "ClassNotFoundException". Feels like a rabbit hole.

Discussed offline, seems like it's not worth it. Decided to move without it.

shrinidhijoshi:
Looks like merge conflict because another diff added a unit test in the same class. We can fix that and merge this

arhimondr:
Updated

PR #21256, opened 2023-10-27
[native] Drop redundant sorting keys from Window and TopNRowNumber nodes

mbasmanova:
Velox doesn't allow partitioning and sorting keys to overlap since
it is redundant to sort by partitioning keys.

Drop sorting keys present in partitioning keys during Presto-to-Velox
plan translation.

```
== NO RELEASE NOTE ==
```


xiaoxmeng (approved):
@mbasmanova thanks for the fix!

PR #21264, opened 2023-10-27
[native] Offload application work from io thread pool to cpu thread pool

tanjialiang:
Currently http requests come and are directly handled by the request receiving thread in an io thread pool. This is dangerous because if the handling io thread's work is not light enough it will block other incoming requests and requests will pile up in io threads' queues, resulting in long response time and in the worst case unresponsive server. The adoption of HTTPS makes the situation even worse because unlike HTTP's connection establishment which happens in kernel threads, HTTPS's connection establishment requires more RTT and an application io thread to handle.

This PR offloads any operations unrelated to work dispatching and response returning to another cpu thread pool called HttpProcessingExecutor to release the possible jam that might happen in the request handling io thread pool. We don't offload it directly to driverCPUPool because we don't want big volume of requests interfere with the heavy processing driver threads.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/QueryContextManager.cpp
+++ b/presto-native-execution/presto_cpp/main/QueryContextManager.cpp
@@ -31,6 +31,13 @@ static std::shared_ptr<folly::CPUThreadPoolExecutor>& executor() {
   return executor;
 }
 
+std::shared_ptr<folly::CPUThreadPoolExecutor>& httpProcessingExecutor() {
+  static auto executor = std::make_shared<folly::CPUThreadPoolExecutor>(
+      SystemConfig::instance()->numQueryThreads(),
+      std::make_shared<folly::NamedThreadFactory>("HttpProcessing"));
+  return executor;
+}
+
 std::shared_ptr<folly::IOThreadPoolExecutor> spillExecutor() {
   const int32_t numSpillThreads = SystemConfig::instance()->numSpillThreads();
   if (numSpillThreads <= 0) {
@@ -44,6 +51,10 @@ std::shared_ptr<folly::IOThreadPoolExecutor> spillExecutor() {
 
 folly::CPUThreadPoolExecutor* driverCPUExecutor() {
   return executor().get();
+}
+
+folly::CPUThreadPoolExecutor* httpProcessingExecutorPtr() {
+  return httpProcessingExecutor().get();
 }
 
 folly::IOThreadPoolExecutor* spillExecutorPtr() {
--- a/presto-native-execution/presto_cpp/main/TaskManager.cpp
+++ b/presto-native-execution/presto_cpp/main/TaskManager.cpp
@@ -714,7 +714,6 @@ folly::Future<std::unique_ptr<protocol::TaskInfo>> TaskManager::getTaskInfo(
     std::optional<protocol::TaskState> currentState,
     std::optional<protocol::Duration> maxWait,
     std::shared_ptr<http::CallbackRequestHandlerState> state) {
-  auto eventBase = folly::EventBaseManager::get()->getEventBase();
   auto [promise, future] =
       folly::makePromiseContract<std::unique_ptr<protocol::TaskInfo>>();
   auto prestoTask = findOrCreateTask(taskId);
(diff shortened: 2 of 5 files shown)


lingbin:
From the title, I am curious about the reason for putting IO operations in the driver thread. In my understanding, IO operations may hang or be very slow, and the driver thread in the pipeline model does not want to be blocked due to external factors.

I may have missed something. It would be great if the reason can be explained in the commit message. Thanks.

tanjialiang:
> From the title, I am curious about the reason for putting IO operations in the driver thread. In my understanding, IO operations may hang or be very slow, and the driver thread in the pipeline model does not want to be blocked due to external factors.
> 
> I may have missed something. It would be great if the reason can be explained in the commit message. Thanks.

This PR is not ready for review yet. But to answer your question:
> IO operations may hang or be very slow
Not so sure about what you mean by IO operations. But generally speaking, IO thread operations shall be super light and fast, so that it can spare thread resource for larger RPS. Here instead of doing actually Presto/Velox application work in IO threads, we only let IO threads do request dispatching work, in order to make it super light.

> the driver thread in the pipeline model does not want to be blocked due to external factors.
That's a good question, and in order to address this we are using another cpu pool for handling the middle layer work.(httpProcessingExecutor) I think I might want to change the title to remove this confusion. Commit messages and description will also be added when the PR is ready for review. Currently it is a draft PR that is not meant for review.

presto-native-execution/presto_cpp/main/TaskManager.cpp
```diff
@@ -794,18 +795,17 @@ folly::Future<std::unique_ptr<Result>> TaskManager::getResults(
   // destruction. To avoid the BrokenPromise exception, fulfill the promise
   // with incomplete empty pages.
   promiseHolder->atDestruction(
-      [token](folly::Promise<std::unique_ptr<Result>> promise) {
+      [token](folly::Promise<std::unique_ptr<Result>> p) {
```
xiaoxmeng, line 798:
```
s/p/promise/
```

presto-native-execution/presto_cpp/main/TaskManager.cpp
```diff
@@ -872,24 +872,24 @@ folly::Future<std::unique_ptr<Result>> TaskManager::getResults(
...
     promiseHolder->promise.setException(e);
-    return std::move(future).via(eventBase);
+    return std::move(future).via(executor);
   }
 }
 
 folly::Future<std::unique_ptr<protocol::TaskStatus>> TaskManager::getTaskStatus(
     const TaskId& taskId,
     std::optional<protocol::TaskState> currentState,
     std::optional<protocol::Duration> maxWait,
```
xiaoxmeng, line 890: Shall we pass the executor to TaskManager?

presto-native-execution/presto_cpp/main/QueryContextManager.cpp
```diff
@@ -31,6 +31,13 @@ static std::shared_ptr<folly::CPUThreadPoolExecutor>& executor() {
   return executor;
 }
 
+std::shared_ptr<folly::CPUThreadPoolExecutor>& httpProcessingExecutor() {
+  static auto executor = std::make_shared<folly::CPUThreadPoolExecutor>(
```
xiaoxmeng, line 35: Why not let PrestoServer to own this executor? Rename it to httpServerExecutor? Thanks!

xiaoxmeng (commented):
@tanjialiang Thanks for the optimization!

presto-native-execution/presto_cpp/main/QueryContextManager.cpp
tanjialiang: Putting all executors in a single place for better maintenance. httpServerExecutor sounds like httpExecSrv which is the executor for iopool.
I'm thinking moving these pools into another file in another PR for isolation. They don't seem to belong to QueryContextManager

presto-native-execution/presto_cpp/main/TaskManager.cpp
tanjialiang: Yeah I guess we can do that

spershin (approved):
@tanjialiang 
Thanks!
Looks good.

Let's add (in a followup change) a latency counter for the new executor like we have for the driverCPU and HTTP exec thread.

tanjialiang:
> Details

Yeah we can have that as a followup

presto-native-execution/presto_cpp/main/QueryContextManager.cpp
xiaoxmeng: I think Presto server currently owns these executors. Let's maintain there why we need a separate file?

PR #21265, opened 2023-10-28
[native] Convert input field names to lower case in plan conversion

kevinwilfong:
Presto is case insensitive, and so always converts column names and field names to lower case in its plan.

When Presto constructs the Velox query plan it passes in the types for Hive tables as they were read from the metastore, with capital letters if they were there in the metastore. Velox uses these types to construct the output type of the TableScan.

This leads to errors as the output of the TableScan will have capital letters in field names, while field accesses in the rest of the plan will use the field names Presto used in its original plan, which are all lower case.

Converting the field names from the types in the input Hive tables to lower case fixes this inconsistency.

### DIFF
--- a/presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp
+++ b/presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp
@@ -37,6 +37,7 @@
 #include "velox/common/compression/Compression.h"
 // clang-format on
 
+#include <folly/String.h>
 #include <folly/container/F14Set.h>
 
 using namespace facebook::velox;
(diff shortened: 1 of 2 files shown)


presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp
```diff
@@ -799,6 +800,56 @@ std::unique_ptr<common::Filter> toFilter(
   VELOX_UNSUPPORTED("Unsupported filter found.");
 }
 
+template <TypeKind KIND>
+TypePtr fieldNamesToLowerCase(const TypePtr& type) {
+  return type;
```
majetideepak, line 805: Why don't we convert the type name to lowercase here?

kevinwilfong: we only need to convert field names in rows

this is the default implementation which is used for primitives (they don't have fields, and they can't contain rows)

I added a specialization for rows to convert field names, and specializations for arrays and maps to recursively handle any rows they contain

majetideepak (approved):

xiaoxmeng (approved):
@kevinwilfong LGTM. Thanks!

amitkdutta (approved):
Thanks @kevinwilfong

PR #21277, opened 2023-10-31
Increase timeout when starting test query runner

mshang816:
## Description
Increase timeout when starting test query runner to 60 seconds as in some testing environments query runner takes a longer time to start up. 

## Impact
N/A

## Test Plan
Make sure all CI runs are green

```
== NO RELEASE NOTE ==
```


amitkdutta (approved):
Thanks @mshang816

NikhilCollooru (approved):

PR #21287, opened 2023-10-31
Add support for user provided execution strategy

singcha:
Adds support for specifying execution strategy using ``spark_execution_strategies`` session property

## Description
Currently an execution strategy is determined after job failure by retry handler. This limits us from specifying execution strategy based on previous job execution.
This change adds support for specifying execution strategy using ``spark_execution_strategies`` session property

## Test Plan
- Added unit tests

```
== NO RELEASE NOTE ==
```


shrinidhijoshi (commented):
Thanks for working on this @singcha . Something to consider w.r.t to implementation

While the intent is to expose the existing retry config mechanism out into the main input for the job, one adverse effect of this could be that we now have a 1 more way of specifying configs for presto-on-spark job. I see following side-effects specially internally at Meta.

  a. Users copy-pasting this new config class: This will be another knob that users will tune and copy-paste into pipelines. 

  b. Newer abstraction means more complexity in debugging workflows for the team, among the already existing myriad of sources of configuration. 

  c. Precedence: From the current approach it seems like this takes precedence over other configs specified, including user configs. 
     
While all these can be solved by their own means, my concern is - do we need to expose this new abstraction, instead maybe the existing orchestrator (like `dataswarm` at Meta) just specify the necessary configs ?

singcha:
> Thanks for working on this @singcha . Something to consider w.r.t to implementation
> 
> While the intent is to expose the existing retry config mechanism out into the main input for the job, one adverse effect of this could be that we now have a 1 more way of specifying configs for presto-on-spark job. I see following side-effects specially internally at Meta.
> 
> a. Users copy-pasting this new config class: This will be another knob that users will tune and copy-paste into pipelines.
> 
> b. Newer abstraction means more complexity in debugging workflows for the team, among the already existing myriad of sources of configuration.
> 
> c. Precedence: From the current approach it seems like this takes precedence over other configs specified, including user configs.
> 
> While all these can be solved by their own means, my concern is - do we need to expose this new abstraction, instead maybe the existing orchestrator (like `dataswarm` at Meta) just specify the necessary configs ?

Thanks for review.
So, the idea is for users to never specify this and wrappers/orchestrator/dataswarm will populate this config based on historical run of a query.
Implementing the logic of retry and determining the final values in orchestrator requires re-implementing same logic twice at two different places. This can lead to inconsistent behavior. Following current approach, we are specifying execution strategy from dataswarm and leaving the actual values to be determined by presto code.

presto-spark-base/src/main/java/com/facebook/presto/spark/PrestoSparkConfig.java
```diff
@@ -257,6 +260,19 @@ public boolean isRetryOnOutOfMemoryBroadcastJoinEnabled()
         return retryOnOutOfMemoryBroadcastJoinEnabled;
     }
 
+    public List<String> getExecutionStrategies()
+    {
+        return executionStrategies;
+    }
+
+    @Config("spark.execution-strategies")
+    @ConfigDescription("Execution strategy to use")
```
pgupta2, line 269: nit: Execution strategies to use for query execution.

presto-spark-base/src/main/java/com/facebook/presto/spark/PrestoSparkConfig.java
```diff
@@ -49,6 +51,7 @@ public class PrestoSparkConfig
     private int splitAssignmentBatchSize = 1_000_000;
     private double memoryRevokingThreshold;
     private double memoryRevokingTarget;
+    private List<String> executionStrategies = ImmutableList.of();
```
pgupta2, line 54: Should we call it `queryExecutionStrategies` instead of just `executionStrategies`?

presto-spark-base/src/main/java/com/facebook/presto/spark/PrestoSparkSessionProperties.java
```diff
@@ -138,6 +139,15 @@ public PrestoSparkSessionProperties(PrestoSparkConfig prestoSparkConfig)
                         "When revoking memory, try to revoke so much that memory pool is filled below target at the end",
                         prestoSparkConfig.getMemoryRevokingTarget(),
                         false),
+                new PropertyMetadata<>(
+                        SPARK_EXECUTION_STRATEGIES,
+                        "Execution strategies to be applied while running the query",
+                        VARCHAR,
+                        List.class,
+                        prestoSparkConfig.getExecutionStrategies(),
```
pgupta2, line 147: I don't see a use case to have a config property for this session property. Do you see any valid use case for this? If not, we can just  set the default value to Empty list.

presto-spark-base/src/test/java/com/facebook/presto/spark/PrestoSparkQueryRunner.java
```diff
@@ -505,23 +507,35 @@ public MaterializedResult execute(String sql)
...
             }
 
             throw failure;
         }
     }
 
-    private MaterializedResult executeWithRetryStrategies(
+    private List<ExecutionStrategy> getExecutionStrategies(Session session)
+    {
+        List<String> executionStrategiesToApply = PrestoSparkSessionProperties.getExecutionStrategies(session);
```
pgupta2, line 524: Use static import for `getExecutionStrategies`.

presto-spark-base/src/test/java/com/facebook/presto/spark/PrestoSparkQueryRunner.java
```diff
@@ -505,23 +507,35 @@ public MaterializedResult execute(String sql)
...
+        Set<String> allExecutionStrategies = Arrays.stream(ExecutionStrategy.values()).map(v -> v.name()).collect(Collectors.toSet());
+        return executionStrategiesToApply
+                .stream()
+                .filter(tag -> allExecutionStrategies.contains(tag))
```
pgupta2, line 528: nit: `tag` -> `strategy`

presto-spark-base/src/test/java/com/facebook/presto/spark/TestPrestoSparkQueryRunner.java
```diff
@@ -1340,6 +1341,103 @@ public void testCreatePartitionedTable()
...
+
+        session = Session.builder(getSession())
+                .setSystemProperty(JOIN_DISTRIBUTION_TYPE, "BROADCAST")
+                .setSystemProperty(STORAGE_BASED_BROADCAST_JOIN_ENABLED, "true")
+                .setSystemProperty(SPARK_BROADCAST_JOIN_MAX_MEMORY_OVERRIDE, "10B")
+                .setSystemProperty(SPARK_RETRY_ON_OUT_OF_MEMORY_BROADCAST_JOIN_ENABLED, "false")
+                .setSystemProperty(SPARK_EXECUTION_STRATEGIES, "DISABLE_BROADCAST_JOIN")
+                .build();
+
+        // Query should succeed since broadcast join will be disabled due to present of DISABLE_BROADCAST_JOIN tag
```
pgupta2, line 1368: Wondering if we should update this comment and similarly comments in below test cases to callout that we are passing SPARK_QUERY_EXECUTION_STRATEGIES explicitly and disabling the retry. The query should succeed since DISABLE_BROADCAST_JOIN strategy will be applied in the first run itself.

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestAggregations.java
```diff
@@ -216,7 +215,7 @@ public void testCountDistinct()
                         "  GROUP BY orderkey " +
                         "HAVING COUNT(DISTINCT partkey) != CARDINALITY(ARRAY_DISTINCT(ARRAY_AGG(partkey))))",
                 "VALUES 0");
-        assertQuery(builder(TEST_SESSION)
+        assertQuery(builder(getSession())
```
pgupta2, line 218: Is this a valid change?

presto-spark-base/src/main/java/com/facebook/presto/spark/PrestoSparkSessionProperties.java
```diff
@@ -138,6 +139,15 @@ public PrestoSparkSessionProperties(PrestoSparkConfig prestoSparkConfig)
...
+                        false,
+                        value -> Splitter.on(',').trimResults().omitEmptyStrings().splitToList(value.toString()),
```
pgupta2, line 149: Does it make sense to validate the passed strategies to make sure that these are valid? Currently, we do this validation later and basically skips the invalid values. We should basically fail if the passed strategies are incorrect immediately.  Maybe we can extend the decoder function and perform validation inside decoder itself?
```
    private List<ExecutionStrategy> getExecutionStrategies(Map<String, String> sessionProperties)
    {
        Set<String> allExecutionStrategies = Arrays.stream(ExecutionStrategy.values()).map(v -> v.name()).collect(Collectors.toSet());
        String executionStrategies = sessionProperties.getOrDefault(SPARK_EXECUTION_STRATEGIES, "");
        return Splitter.on(',').trimResults().omitEmptyStrings().splitToList(executionStrategies)
                .stream()
                .filter(tag -> allExecutionStrategies.contains(tag))
                .map(t -> ExecutionStrategy.valueOf(t))
                .collect(Collectors.toList());
    }
```

pgupta2 (changes requested):

presto-spark-base/src/main/java/com/facebook/presto/spark/PrestoSparkSessionProperties.java
singcha: I dont see any valid use case. Reverted config property.

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestAggregations.java
singcha: Yes, individual runners should be running with their session defaults to ensure that runtime specific configs are available. Without this change, presto-on-spark code runs into error when it tries to access configs exclusively applicable to presto-on-spark.

presto-spark-base/src/main/java/com/facebook/presto/spark/PrestoSparkSessionProperties.java
singcha: Yes, added validator

presto-spark-base/src/test/java/com/facebook/presto/spark/PrestoSparkQueryRunner.java
```diff
@@ -505,23 +507,33 @@ public MaterializedResult execute(String sql)
     public MaterializedResult execute(Session session, String sql)
     {
         try {
-            return executeWithRetryStrategies(session, sql, ImmutableList.of());
+            List<ExecutionStrategy> executionStrategies = getExecutionStrategies(session);
```
pgupta2, line 510: Can't we just inline this? I don't see `getExecutionStrategies(Session session)` being called from any other place.

presto-spark-launcher/src/main/java/com/facebook/presto/spark/launcher/PrestoSparkRunner.java
```diff
@@ -97,6 +100,7 @@ public void run(
             Optional<String> queryDataOutputLocation)
     {
         IPrestoSparkQueryExecutionFactory queryExecutionFactory = driverPrestoSparkService.getQueryExecutionFactory();
+        List<ExecutionStrategy> executionStrategies = getExecutionStrategies(sessionProperties);
```
pgupta2, line 103: Same here. Lets just inline `getExecutionStrategies(Map<String, String> sessionProperties)` here.

pgupta2 (approved):

shrinidhijoshi (approved):

shrinidhijoshi:
Looks like we need committers stamp as you have a change in `presto-tests` module.

ajaygeorge (approved):
stamping since this is already reviewed.

PR #21290, opened 2023-10-31
Add noisy_approx_distinct_sfm and related functions

jonhehir:
## Description

The new `SfmSketch` type corresponds to the "Sketch-Flip-Merge" data structure of http://proceedings.mlr.press/v202/hehir23a/hehir23a.pdf. In addition to the type, the following functions have been added to support noisy cardinality estimation:
`cardinality(SfmSketch)`, `merge(SfmSketch)`, `merge_sfm(Array<SfmSketch>)`, `noisy_approx_distinct_sfm`, `noisy_approx_set_sfm`, `noisy_empty_approx_set_sfm`

These functions are essentially equivalent to the HyperLogLog functions such as `approx_distinct`, `approx_set`, etc., but using the SFM sketch in lieu of HLL for extra noise.

**Documentation and release notes will be forthcoming in a separate PR.**

## Motivation and Context

These functions are added as part of the `noisyaggregation` suite. A notable difference between the theoretical SFM sketch and this implementation is that the aggregators (e.g., `noisy_approx_set_sfm`) return `NULL` when aggregating empty sets. Strictly speaking, this is a violation of the differential privacy guarantee of the theoretical SFM sketch, though this may be patched with use of the `noisy_empty_approx_set_sfm` function.

## Impact

This does not affect any existing functionality.

## Test Plan

Unit tests have been added for the `SfmSketch` structure itself and the corresponding Presto type and functions.

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/SfmSketchAggregationUtils.java
+++ b/presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/SfmSketchAggregationUtils.java
@@ -1,0 +1,149 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.operator.aggregation.noisyaggregation;
+
+import com.facebook.presto.common.block.BlockBuilder;
+import com.facebook.presto.operator.aggregation.noisyaggregation.sketch.SfmSketch;
+import com.facebook.presto.spi.PrestoException;
+import io.airlift.slice.Slice;
+
+import java.lang.invoke.MethodHandle;
+
+import static com.facebook.presto.common.type.BigintType.BIGINT;
+import static com.facebook.presto.common.type.VarbinaryType.VARBINARY;
+import static com.facebook.presto.spi.StandardErrorCode.INVALID_FUNCTION_ARGUMENT;
+import static com.facebook.presto.util.Failures.checkCondition;
+import static com.facebook.presto.util.Failures.internalError;
+
+public final class SfmSketchAggregationUtils
+{
+    public static final int DEFAULT_BUCKET_COUNT = 4096;
+    public static final int DEFAULT_PRECISION = 24;
+
+    private SfmSketchAggregationUtils() {}
+
+    public static void ensureStateInitialized(SfmSketchState state, double epsilon, long numberOfBuckets, long precision)
+    {
+        if (state.getSketch() != null) {
+            return; // already initialized
+        }
+
+        validateSketchParameters(epsilon, (int) numberOfBuckets, (int) precision);
+        SfmSketch sketch = SfmSketch.create((int) numberOfBuckets, (int) precision);
+        state.setSketch(sketch);
+        state.setEpsilon(epsilon);
+    }
+
+    public static void validateSketchParameters(double epsilon, int numberOfBuckets, int precision)
+    {
+        checkCondition(epsilon > 0,
+                INVALID_FUNCTION_ARGUMENT, "epsilon must be positive");
+        checkCondition(numberOfBuckets > 0 && (numberOfBuckets & (numberOfBuckets - 1)) == 0,
+                INVALID_FUNCTION_ARGUMENT, "numberOfBuckets must be a power of 2");
+        checkCondition(precision > 0 && precision % Byte.SIZE == 0,
+                INVALID_FUNCTION_ARGUMENT, "precision must be a multiple of %s", Byte.SIZE);
+    }
+
+    public static void addHashToSketch(SfmSketchState state, long hash, double epsilon, long numberOfBuckets, long precision)
+    {
+        ensureStateInitialized(state, epsilon, numberOfBuckets, precision);
+        state.getSketch().addHash(hash);
+    }
+
+    public static long hashLong(MethodHandle methodHandle, long value)
+    {
+        long hash;
+        try {
+            hash = (long) methodHandle.invokeExact(value);
+        }
+        catch (Throwable t) {
+            throw internalError(t);
+        }
+        return hash;
+    }
+
+    public static long hashDouble(MethodHandle methodHandle, double value)
+    {
+        long hash;
+        try {
+            hash = (long) methodHandle.invokeExact(value);
+        }
+        catch (Throwable t) {
+            throw internalError(t);
+        }
+        return hash;
+    }
+
+    public static long hashSlice(MethodHandle methodHandle, Slice value)
+    {
+        long hash;
+        try {
+            hash = (long) methodHandle.invokeExact(value);
+        }
+        catch (Throwable t) {
+            throw internalError(t);
+        }
+        return hash;
+    }
+
+    public static void mergeStates(SfmSketchState state, SfmSketchState otherState)
+    {
+        SfmSketch sketch = state.getSketch();
+        SfmSketch otherSketch = otherState.getSketch();
+        if (sketch == null) {
+            state.setSketch(otherSketch);
+            state.setEpsilon(otherState.getEpsilon());
+        }
+        else {
+            try {
+                // Throws if the sketches are incompatible (e.g., different bucket counts/size)
+                // Catch and throw a PrestoException
+                state.mergeSketch(otherSketch);
+            }
+            catch (IllegalArgumentException e) {
+                throw new PrestoException(INVALID_FUNCTION_ARGUMENT, e.getMessage(), e);
+            }
+        }
+    }
+
+    public static void writeCardinality(SfmSketchState state, BlockBuilder out)
+    {
+        SfmSketch sketch = state.getSketch();
+        if (sketch == null) {
+            // In the event we process no rows, output NULL.
+            // Note: Although the SfmSketch is differentially private, this particular output is not.
+            // We cannot output anything DP here because the state will not include epsilon if we processed no rows.
+            out.appendNull();
+        }
+        else {
+            sketch.enablePrivacy(state.getEpsilon());
+            BIGINT.writeLong(out, sketch.cardinality());
+        }
+    }
+
+    public static void writeSketch(SfmSketchState state, BlockBuilder out)
+    {
+        SfmSketch sketch = state.getSketch();
+        if (sketch == null) {
+            // In the event we process no rows, output NULL.
+            // Note: Although the SfmSketch is differentially private, this particular output is not.
+            // We cannot output anything DP here because the state will not include epsilon if we processed no rows.
+            out.appendNull();
+        }
+        else {
+            sketch.enablePrivacy(state.getEpsilon());
+            VARBINARY.writeSlice(out, sketch.serialize());
+        }
+    }
+}
--- a/presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/SfmSketchMergeAggregation.java
+++ b/presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/SfmSketchMergeAggregation.java
@@ -1,0 +1,69 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.operator.aggregation.noisyaggregation;
+
+import com.facebook.presto.common.block.BlockBuilder;
+import com.facebook.presto.operator.aggregation.noisyaggregation.sketch.SfmSketch;
+import com.facebook.presto.spi.PrestoException;
+import com.facebook.presto.spi.function.AggregationFunction;
+import com.facebook.presto.spi.function.AggregationState;
+import com.facebook.presto.spi.function.CombineFunction;
+import com.facebook.presto.spi.function.InputFunction;
+import com.facebook.presto.spi.function.OutputFunction;
+import com.facebook.presto.spi.function.SqlType;
+import com.facebook.presto.type.SfmSketchType;
+import io.airlift.slice.Slice;
+
+import static com.facebook.presto.common.type.VarbinaryType.VARBINARY;
+import static com.facebook.presto.spi.StandardErrorCode.INVALID_FUNCTION_ARGUMENT;
+
+@AggregationFunction("merge")
+public final class SfmSketchMergeAggregation
+{
+    private SfmSketchMergeAggregation() {}
+
+    @InputFunction
+    public static void input(@AggregationState SfmSketchState state, @SqlType(SfmSketchType.NAME) Slice value)
+    {
+        SfmSketch sketch = SfmSketch.deserialize(value);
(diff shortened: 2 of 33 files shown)


jonhehir:
cc @duykienvp

jonhehir:
Updated the PR to include the changes from https://github.com/prestodb/airlift/pull/65 (which are being reverted in https://github.com/prestodb/airlift/pull/66) and to rebase onto a fresh copy of `master`. This is ready to review now.

cc @gcormode

presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/sketch/Bitmap.java
```diff
@@ -0,0 +1,194 @@
...
+import org.openjdk.jol.info.ClassLayout;
+
+import static com.google.common.base.Preconditions.checkArgument;
+
+/**
+ * A level of abstraction over the bitmaps used in sketches such as LPCA and SFM.
+ * These are essentially arrays of booleans that support flipping and applying randomized response.
+ * Concretely, these are stored as byte arrays.
+ */
+public class Bitmap
```
mlyublena, line 28: I'm wondering if Bitmaps are useful in other parts of the code (and in that case perhaps the class belongs in a different location)
but I guess for now we can leave them here

mlyublena (approved):
Looks good from Presto's POV

presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/SfmSketchAggregationUtils.java
```diff
@@ -0,0 +1,151 @@
...
+        if (sketch == null) {
+            state.setSketch(otherSketch);
+            state.setEpsilon(otherState.getEpsilon());
+            state.addMemoryUsage(otherSketch.getRetainedSizeInBytes());
+        }
+        else {
+            try {
+                // Throws if the sketches are incompatible (e.g., different bucket counts/size)
+                // Catch and throw a PrestoException
+                sketch.mergeWith(otherSketch);
```
arhimondr, line 114: same memory accounting related question here

presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/sketch/Bitmap.java
```diff
@@ -0,0 +1,194 @@
...
+
+    public Bitmap(int length)
+    {
+        validateLength(length);
+        bitmap = new byte[length / Byte.SIZE];
+    }
+
+    private Bitmap(byte[] bytes)
+    {
+        bitmap = bytes;
```
arhimondr, line 42: nit: requireNonNull

presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/sketch/Bitmap.java
```diff
@@ -0,0 +1,194 @@
...
+    {
+        return new Bitmap(bytes);
+    }
+
+    public static Bitmap fromSliceInput(SliceInput input, int length)
+    {
+        validateLength(length);
+        byte[] bytes = new byte[length / Byte.SIZE];
+        for (int i = 0; i < bytes.length; i++) {
+            bytes[i] = input.readByte();
```
arhimondr, line 55: input.readBytes can be more efficient

presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/sketch/Bitmap.java
arhimondr, line 28: Have you considered https://github.com/RoaringBitmap/RoaringBitmap?

presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/sketch/SfmSketch.java
```diff
@@ -0,0 +1,426 @@
...
+    private double randomizedResponseProbability;
+    private Bitmap bitmap;
+
+    private SfmSketch(Bitmap bitmap, int indexBitLength, int precision, double randomizedResponseProbability, RandomizationStrategy randomizationStrategy)
+    {
+        validatePrefixLength(indexBitLength);
+        validatePrecision(precision, indexBitLength);
+        validateRandomizedResponseProbability(randomizedResponseProbability);
+
+        this.bitmap = bitmap;
```
arhimondr, line 72: nit: requireNonNull

presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/sketch/SfmSketch.java
```diff
@@ -0,0 +1,426 @@
...
+        this.indexBitLength = indexBitLength;
+        this.precision = precision;
+        this.randomizedResponseProbability = randomizedResponseProbability;
+        this.randomizationStrategy = randomizationStrategy;
```
arhimondr, line 76: nit: requireNonNull

presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/sketch/SfmSketch.java
```diff
@@ -0,0 +1,426 @@
...
+
+    private static final byte FORMAT_TAG = 7;
+    private static final int MAX_ESTIMATION_ITERATIONS = 1000;
+    private static final int INSTANCE_SIZE = ClassLayout.parseClass(SfmSketch.class).instanceSize();
+
+    private final int indexBitLength;
+    private final int precision;
+    private final RandomizationStrategy randomizationStrategy;
+
+    private double randomizedResponseProbability;
```
arhimondr, line 63: How difficuilt would it be to refactor the tests and make these two fields final?

presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/sketch/SecureRandomizationStrategy.java
```diff
@@ -0,0 +1,43 @@
...
+ * Note: Due to finite-precision implementation details, usage of floating-point functions
+ * as random noise, while cryptographically secure, may leak information from a privacy context.
+ * See "On Significance of the Least Significant Bits For Differential Privacy" by Mironov
+ * and use judiciously.
+ */
+public class SecureRandomizationStrategy
+        extends RandomizationStrategy
+{
+    private static final int INSTANCE_SIZE = ClassLayout.parseClass(SecureRandomizationStrategy.class).instanceSize();
+    private static final SecureRandom random = new SecureRandom();
```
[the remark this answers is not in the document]
jonhehir: Per https://arxiv.org/pdf/2002.04049.pdf, we should use a cryptographically secure PRNG. I was under the impression that `SecureRandom` would essentially use a secure seed, then deterministically generate from a PRNG, though on closer inspection, that appears to depend on the exact implementation being used. Since I'm not an expert here, I'm open to suggestions on how best to proceed here to balance security and performance.

presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/sketch/SecureRandomizationStrategy.java
```diff
@@ -0,0 +1,43 @@
...
+
+    public SecureRandomizationStrategy() {}
+
+    public long getRetainedSizeInBytes()
+    {
+        return INSTANCE_SIZE;
```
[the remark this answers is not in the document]
jonhehir: Does it need to be accounted for if it's static? I believe in a previous code review, the suggestion was to make it static to avoid having to figure out the memory usage of a `Random` object.

presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/SfmSketchState.java
```diff
@@ -0,0 +1,34 @@
...
+
+@AccumulatorStateMetadata(stateSerializerClass = SfmSketchStateSerializer.class, stateFactoryClass = SfmSketchStateFactory.class)
+public interface SfmSketchState
+        extends AccumulatorState
+{
+    SfmSketch getSketch();
+
+    void setSketch(SfmSketch value);
+
+    void addMemoryUsage(long value);
```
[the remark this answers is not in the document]
jonhehir: Good question! This was done to mimic the design in `HyperLogLog`-related classes. I like the idea of deferring the responsibility to `setSketch()`, though I'd need to also account for memory when merging a sketch in state.

presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/SfmSketchStateFactory.java
```diff
@@ -0,0 +1,166 @@
...
+        public SfmSketch getSketch()
+        {
+            return sketches.get(getGroupId());
+        }
+
+        @Override
+        public void setSketch(SfmSketch value)
+        {
+            requireNonNull(value, "value is null");
+            sketches.set(getGroupId(), value);
```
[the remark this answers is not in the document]
jonhehir: That's correct. There used to be such a line here, but I deleted it, since it was a no-op. I can put it back, though, just in case anything ever changes in the future.

presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/SfmSketchMergeAggregation.java
```diff
@@ -0,0 +1,70 @@
...
+            // if state sketch is empty, add this sketch to the state
+            state.setSketch(sketch);
+            state.setEpsilon(0); // not used
+            state.addMemoryUsage(sketch.getRetainedSizeInBytes());
+        }
+        else {
+            // if state already has a sketch, merge in the current sketch
+            try {
+                // throws if the sketches are incompatible (e.g., different bucket count/size)
+                previous.mergeWith(sketch);
```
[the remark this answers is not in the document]
jonhehir: Correct, it doesn't change the footprint. I don't anticipate it changing in the future, but I assume it's cheap enough to make the calls that we can do it anyway just in case.

presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/SfmSketchAggregationUtils.java
jonhehir: As above, memory usage won't change here, but I'll add the accounting back for future-proof-ness.

presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/sketch/Bitmap.java
jonhehir: I hadn't! Thanks for the link. In this case, I'm not sure a compressed bitmap will offer much value here, since in the end we scatter random bits throughout the bitmap. It'd be worth confirming in the future, but for now, I think we can just keep the uncompressed bitmap.

That said, the `Bitmap` class should probably extend Java's `BitSet`.

presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/sketch/SfmSketch.java
jonhehir: Making `randomizedResponseProbability` final would require returning copies rather than mutating in `enablePrivacy` and `mergeWith`. `mergeWith` in particular was designed to mutate to match the functionality in `HyperLogLog`, `QuantileDigest`, etc.

`bitmap` seems like it should be final, but I'll take a closer look.

jonhehir:
@arhimondr, I pushed changes to address most of these concerns. I believe the outstanding questions surround the use (and memory accounting) of `SecureRandom`. (See responses above.) Let me know how you'd like to proceed, thanks!

presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/SfmSketchMergeAggregation.java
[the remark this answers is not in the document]
jonhehir: > I don't anticipate it changing in the future...

Oops, I changed it already! Now that I've updated `Bitmap` to wrap Java's `BitSet`, the actual length of serialized bytes is variable-length. In practice, this means that serialized state will vary in size and should be a good bit smaller during much of the typical aggregation, since the bits in the bitmap will concentrate in the lower positions. The final result will usually be about the same size, though, since random bits will be scattered throughout the bitmap.

presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/SfmSketchAggregationUtils.java
jonhehir: (This is no longer true and memory accounting has been adjusted.)

presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/sketch/Bitmap.java
jonhehir: As mentioned above, the `Bitmap` class now wraps `BitSet`.

presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/sketch/SfmSketch.java
jonhehir: `bitmap` is now final.

presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/sketch/Bitmap.java
```diff
@@ -0,0 +1,172 @@
...
+    @Override
+    public Bitmap clone()
+    {
+        return Bitmap.fromBytes(length, bitSet.toByteArray());
+    }
+
+    public long getRetainedSizeInBytes()
+    {
+        // Under the hood, BitSet stores a long[] array of BitSet.size() bits
+        return INSTANCE_SIZE + BITSET_INSTANCE_SIZE + bitSet.size() / Byte.SIZE;
```
arhimondr, line 97: nit: `sizeOfLongArray(bitSet.size() / Long.SIZE)` to account for the `long[]` object headers.

presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/sketch/SecureRandomizationStrategy.java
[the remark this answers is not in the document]
arhimondr: Hmm, actually I'm not sure if it's safe to have it `static`. It is not thread safe.

presto-main/src/main/java/com/facebook/presto/operator/aggregation/noisyaggregation/sketch/SecureRandomizationStrategy.java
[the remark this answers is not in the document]
arhimondr: `NativePRNGNonBlocking` is used for secure_random UDF in Presto: https://github.com/prestodb/presto/blob/master/presto-main/src/main/java/com/facebook/presto/operator/scalar/MathFunctions.java#L102

Here's a description on how it operates: https://www.synopsys.com/blogs/software-security/proper-use-of-javas-securerandom.html

In short it initializes the seed from `/dev/urandom` what doesn't block if the OS experiencing lack of entropy. The algorithm is based on SHA1PRNG which is considered cryptographically secure.

From what i understand the paper states that:

> “Random numbers” should be generated by a cryptographically secure pseudo-random number generator (CSPRNG).

It looks like pseudo-random should be fine, as long as it is cryptographically secure.

jonhehir:
Thanks once again for your thoughtful comments, Andrii! In the latest commit, we're using the same `SecureRandom` providers as in `MathFunctions` (and centralized this logic into `com.facebook.presto.util`), and the static property in `SecureRandomizationStrategy` is gone. To avoid the need for memory estimation of the `SecureRandom` instance, we're no longer storing the `randomizationStrategy` property in `SfmSketch` itself, but passing it as a method to the two methods that depend on a source of randomness. Let me know if you have any further questions/concerns!

arhimondr (approved):

PR #21294, opened 2023-11-01
Disallow dropping partition column in Iceberg table explicitly

hantangwangd:
## Description

When we try to drop a partition column in Iceberg table, we should firstly remove it from the partition spec, then drop the column in table schema. But currently Iceberg have some problems in this scenario, see issue https://github.com/apache/iceberg/issues/4563. This PR explicitly disallow drop partition column until Iceberg fixed it.

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
@@ -378,6 +378,17 @@ public abstract class IcebergAbstractMetadata
         IcebergTableHandle icebergTableHandle = (IcebergTableHandle) tableHandle;
         IcebergColumnHandle handle = (IcebergColumnHandle) column;
         Table icebergTable = getIcebergTable(session, icebergTableHandle.getSchemaTableName());
+
+        // Currently drop partition column used in any partition specs of a table would introduce some problems in Iceberg.
+        // So we explicitly disallow dropping partition columns until Iceberg fix this problem.
+        // See https://github.com/apache/iceberg/issues/4563
+        boolean shouldNotDropPartitionColumn = icebergTable.specs().values().stream()
+                .flatMap(partitionSpec -> partitionSpec.fields().stream())
+                .anyMatch(field -> field.sourceId() == handle.getId());
+        if (shouldNotDropPartitionColumn) {
+            throw new PrestoException(NOT_SUPPORTED, "This connector does not support dropping columns which exist in any of the table's partition specs");
+        }
+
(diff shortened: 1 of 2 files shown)


tdcmeehan:
Isn't it true that this issue only affects prior (historical) specs, and not the latest?  So for example, it doesn't affect brand new tables, and it doesn't affect newly added columns?

hantangwangd:
That's right, because dropping columns used in current partition spec would explicitly fail in Iceberg implementation. 

So when we try to drop a partition column, we must firstly drop it from current partition spec. That means, the column exists in a historical spec now. Then we will meet various problems subsequently after dropping it from schema. In my experiment, simultaneously drop the column from partition spec and schema in a single transaction would meet problems in the issue too.

Furthermore,  this PR can also prevent Iceberg table falling into this situation in case that, for example, if we support dropping column from partition spec only.

tdcmeehan (approved):
Makes sense.  LGTM!

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
```diff
@@ -378,6 +378,17 @@ public void dropColumn(ConnectorSession session, ConnectorTableHandle tableHandl
...
         Table icebergTable = getIcebergTable(session, icebergTableHandle.getSchemaTableName());
+
+        // Currently drop partition column used in any partition specs of a table would introduce some problems in Iceberg.
+        // So we explicitly disallow dropping partition columns until Iceberg fix this problem.
+        // See https://github.com/apache/iceberg/issues/4563
+        boolean shouldNotDropPartitionColumn = icebergTable.specs().values().stream()
+                .flatMap(partitionSpec -> partitionSpec.fields().stream())
+                .anyMatch(field -> field.sourceId() == handle.getId());
+        if (shouldNotDropPartitionColumn) {
+            throw new PrestoException(NOT_SUPPORTED, "This connector does not support drop partition columns");
```
ZacBlanco, line 389: nit: Personally, I would make the error message more specific
```suggestion
            throw new PrestoException(NOT_SUPPORTED, "This connector does not support dropping columns which exist in any of the table's partition specs");
```

ZacBlanco (approved):
LGTM!

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
hantangwangd: Fixed!

PR #19933, opened 2023-06-22
Add timeout for HBO stats fetch

feilong-liu:
Fix issue https://github.com/prestodb/presto/issues/20355
Depended by https://github.com/facebookexternal/presto-facebook/pull/2394

- Add timeout to HBO optimizers, will not populate stats equivalent nodes field if timeout
- Add queries which timeouts to log

### Test plan - (Please fill in how you tested your changes)

Tested locally end to end.

```
== RELEASE NOTES ==

General Changes
* Add timeout for HBO optimizer, timeout set by session parameter `history_based_optimizer_timeout_limit`
```

PR #21308, opened 2023-11-03
Skip history optimizer query registration if previous registration timeout

feilong-liu:
## Description

In current code, HBO optimizer runs twice. If in the first run, the query registration timeout, it's very likely that the second run will timeout too. Instead of run both and timeout twice, in this PR, I add code to record the timeout of the first run, so that the  second run will skip if the first run timeout.

## Motivation and Context
Reduce latency for HBO.

## Impact
HBO latency for queries which timeout during HBO query registration will be half as of now. For example, if timeout limit is 10 seconds, and the total latency caused by HBO optimizer which timeout will be 2*10 = 20 seconds. After this change, it will be 10 seconds.

## Test Plan
Test locally end to end, optimizer latency decreased from 25.44s to 10.2s

<br class="Apple-interchange-newline">

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Improve history optimizer latency for queries which timeout during query registration in history optimizer
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/cost/HistoryBasedPlanStatisticsCalculator.java
+++ b/presto-main/src/main/java/com/facebook/presto/cost/HistoryBasedPlanStatisticsCalculator.java
@@ -79,10 +79,15 @@ public class HistoryBasedPlanStatisticsCalculator
     @Override
     public boolean registerPlan(PlanNode root, Session session, long startTimeInNano, long timeoutInMilliseconds)
     {
+        // If previous registration timeout for this query, this run is likely to timeout too, return false.
+        if (historyBasedStatisticsCacheManager.historyBasedQueryRegistrationTimeout(session.getQueryId())) {
+            return false;
+        }
         ImmutableList.Builder<PlanNodeWithHash> planNodesWithHash = ImmutableList.builder();
         Iterable<PlanNode> planNodeIterable = forTree(PlanNode::getSources).depthFirstPreOrder(root);
         for (PlanNode plan : planNodeIterable) {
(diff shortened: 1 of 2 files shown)


presto-main/src/main/java/com/facebook/presto/cost/HistoryBasedPlanStatisticsCalculator.java
```diff
@@ -142,7 +151,7 @@ private Map<PlanCanonicalizationStrategy, PlanNodeWithHash> getPlanNodeHashes(Pl
 
     private PlanNodeStatsEstimate getStatistics(PlanNode planNode, Session session, Lookup lookup, PlanNodeStatsEstimate delegateStats)
     {
-        if (!useHistoryBasedPlanStatisticsEnabled(session) || historyBasedStatisticsCacheManager.loadHistoryFailed(session.getQueryId())) {
```
feilong-liu, line 145: `loadHistoryFailed` is a macro optimization for HBO latency added in https://github.com/prestodb/presto/pull/19933. For queries which do not load any history statistics, we will skip attempt to read history stats. However, by the design, all statistics are cached during query registration, and the read here will be reading from cache, hence the saving is small. On the other hand, we run HBO optimizer twice, having one run reading empty does not mean the second time will still be empty. Hence remove this macro optimization here.

jaystarshot (approved):

mlyublena (approved):

ajaygeorge (approved):
LGTM

PR #21313, opened 2023-11-06
[native] add a test case for text reader

mshang816:
## Description
* add a test case for text reader and exclude it in all-non-parquet-tests profile
* add tests case for non-json/text reads to their own group
To make it easier to run the tests when text reader is enabled.

## Impact
N/A

## Test Plan
Make sure all CI runs are green

```
== NO RELEASE NOTE ==
```


amitkdutta (approved):
Thanks @mshang816

PR #21325, opened 2023-11-06
[native] Add caching of parsed Types

kevinwilfong:
We've seen cases of queries that spend a large amount of time just parsing types when converting the Presto Plan to Velox. This seems to be because it parses the same large Row Types that are used across many field accesses.

Adding caching within a request shows a substantial decrease in the amount of time it takes to do the conversion.

Notably, this helps with timeouts we're seeing making calls from the coordinator to create tasks on the Workers.


xiaoxmeng (approved):
@kevinwilfong nice catch. Thanks for the optimization!

xiaoxmeng (approved):

majetideepak:
@kevinwilfong I am adding Presto parser support using Flex/Bison in Velox. https://github.com/facebookincubator/velox/pull/7568
The end goal is to replace Antlr with that and remove a dependency.
I will add support for caching.
Is there a benchmark to evaluate the performance?

presto-native-execution/presto_cpp/main/types/TypeParser.h
```diff
@@ -0,0 +1,31 @@
...
+#include "presto_cpp/main/types/antlr/TypeSignatureBaseVisitor.h"
+
+namespace facebook::presto {
+
+class TypeParser {
+ public:
+  velox::TypePtr parse(const std::string& text) const;
+
+ private:
+  mutable std::unordered_map<std::string, velox::TypePtr> cache_;
```
majetideepak, line 28: Any reason not to use the `SimpleLRUCache` from Velox?
We use that to cache file handles
https://github.com/facebookincubator/velox/blob/main/velox/connectors/hive/FileHandle.h#L62

majetideepak:
I am worried that without a bound, the cache might grow too big in a production system.

PR #21330, opened 2023-11-07
[native] Rename TestNativeProbabilityFunctionQueries to TestPresto...

mshang816:
## Description
For the test to be picked up by circle ci. 
CircleCI will only pick up TestPrestoNative*.java for testing so TestNativeProbabilityFunctionQueries never got run in CI before. 
Also fix a few failed test cases:
* removing ';' which will fail the query
* enlarge scale to 0.1 which must be greater than 0  

## Test Plan
Existing CI jobs

```
== NO RELEASE NOTE ==
```


mshang816:
c.c. @gopukrishnasIBM who added the test cases

majetideepak (approved):

PR #21333, opened 2023-11-07
Update Drift dependency to use Header transport by default (2nd attempt)

sotodel:
The original attempt failed in certain jobs because they were using an extremely outdated version of Java. Those have since been updated.

Original:

This commit updates Drift version to 1.37. Thrift client/servers will now provide/expect the "thrift" ALPN value, which is meant to identify and select Header, Framed, and Unframed transports. Additionally, Drift Thrift clients will now use Header transport by default.

== RELEASE NOTES ==

General Changes

- Updated drift version to 1.37


jaystarshot:
Release note can just be 

> Updated drift version to 1.37

ajaygeorge (approved):
LGTM

PR #21336, opened 2023-11-08
[native] Create ExchangeSource instances via make_shared().

spershin:
We've seen weird crashed in jemalloc_je_safety_check_fail() and one suspect is
instances of classes inherited from ExchangeSource are being created as
unique_ptr, while ExchangeSource inherits from std::enable_shared_from_this
and is stored as shared_ptr.

This change fixes this discrepancy along with few cosmetic changes.

```
== NO RELEASE NOTE ==
```


xiaoxmeng (approved):
@spershin LGTM. Thanks!

PR #21354, opened 2023-11-09
[native] Advance Velox

kevinwilfong:
To pull in https://github.com/facebookincubator/velox/pull/7479 which we can use to address some errors we've been seeing when the Coordinator gets the status from the Worker.


amitkdutta (approved):
Looks good. Thanks @kevinwilfong

PR #21368, opened 2023-11-10
[native] Clean old tasks based on termination time

kevinwilfong:
Currently we clean old tasks based on end time. End time is set when the task finishes processing. Significant time may pass between that point and when downstream workers finish consuming results, when the task is terminated. In such situations the task is cleaned up immediately after being terminated, leading to race conditions with in flight status requests.

Using the termination time is safer and a closer analog to what is used in Java.


spershin (approved):
Thanks!

PR #21370, opened 2023-11-13
fix the issue of `hudi-connector` getting no partitions.

songpcmusic:
…tionPredicates use HashMap.

## Description
fix the issue of `hudi-connector` getting no partitions, due to partitionPredicates use HashMap, domains in incorrect order.

## Motivation and Context
When I execute a SQL to query Hudi data, HudiPartitionManager#getEffectivePartitions return 0 partitions, but 2 partitions exist. ThriftHiveMetastore#getPartitionNamesByFilter convert Predicate to parts, but items order of parts is incorrect, do not match the partition order, the root case is HudiPartitionManager using HashMap to partitionPredicate.

## Impact
public List<String> getEffectivePartitions(
            ConnectorSession connectorSession,
            ExtendedHiveMetastore metastore,
            SchemaTableName schemaTableName,
            TupleDomain<ColumnHandle> constraintSummary)
{}

### DIFF
--- a/presto-hudi/src/main/java/com/facebook/presto/hudi/HudiPartitionManager.java
+++ b/presto-hudi/src/main/java/com/facebook/presto/hudi/HudiPartitionManager.java
@@ -33,7 +33,7 @@ import com.google.common.collect.ImmutableMap;
 import javax.inject.Inject;
 
 import java.time.ZoneId;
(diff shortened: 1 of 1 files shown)


jaystarshot:
Thanks for your contribution! Can you add a test too?

Issue #21372, opened 2023-11-13
Documentation - update Deploy Presto From a Docker Image doc with new Presto Docker image


steveburnett:
The topic [Installing and Running the Presto Docker container](https://prestodb.io/docs/0.284/installation/deploy-docker.html#installing-and-running-the-presto-docker-container) on [Deploy Presto From a Docker Image](https://prestodb.io/docs/0.284/installation/deploy-docker.html) is out of date because the Docker hub ahanaio account has been shut down as mentioned in [PR 21341](https://github.com/prestodb/presto/issues/21341#issuecomment-1806310576). 

<img width="764" alt="Screenshot 2023-11-13 at 1 21 11 PM" src="https://github.com/prestodb/presto/assets/7013443/826bfc55-c08a-45c4-bec4-4cb02d2c0fdb">

Need to update steps 1 and 2 to update the commands
`docker pull ahanaio/prestodb-sandbox`
and 
`docker run -p 8080:8080 --name presto ahanaio/prestodb-sandbox`

to pull a Presto Docker image from one of the ones present on the [prestodb repos page on DockerHub](https://hub.docker.com/u/prestodb). 

If the presto Docker image to be downloaded instead of ahanaio/prestodb-sandbox requires the user to do manual configurations that were not required with the ahanaio/prestodb-sandbox, steps to configure the replacement presto Docker image need to be added to the documentation.

steveburnett:
What docker pull command should the user run in place of `docker pull ahanaio/prestodb-sandbox` ?

I've tried to pull the images on https://hub.docker.com/r/prestodb/presto and https://hub.docker.com/r/prestodb/prestodb-sandbox using the "docker pull" command on each page.

`docker pull prestodb/presto`

`docker pull prestodb/prestodb-sandbox`

Both of these commands return the same error:

"Error response from daemon: manifest for prestodb/presto:latest not found: manifest unknown: manifest unknown"

@tdcmeehan and @wanglinsong please advise. Thanks!

steveburnett:
Offline discussion recommends the latest non-edge image on the [prestodb/presto DockerHub page](https://hub.docker.com/r/prestodb/presto/tags), so the current command versions should be

`docker pull prestodb/presto:0.284`
and 
`docker run -p 8080:8080 --name presto prestodb/presto:0.284`

Issue #21341, opened 2023-11-08
404 on dockerhub / cant pull image


hbazan-pp:
Dockerhub is returning 404 on https://hub.docker.com/r/ahanaio/prestodb-sandbox
![image](https://github.com/prestodb/presto/assets/90913601/71a3898c-3ad5-4177-ad7d-7310013b6e02)

and getting similar error when pulling image:
![image](https://github.com/prestodb/presto/assets/90913601/62c80b0d-b13a-41e9-8726-617beef1accf)

## Your Environment
I don't think it is related to my environment in any way
## Expected Behavior
Image should be found and pulled

## Current Behavior
404 or access denied error

## Steps to Reproduce
`docker image pull ahanaio/prestodb-sandbox:0.284
`

steveburnett:
The 404'd link https://hub.docker.com/r/ahanaio/prestodb-sandbox is present on the Presto website, on  https://prestodb.io/getting-started/ . When a solution to this 404 is found, that page may need to be updated.

hbazan-pp:
but also docker image pull fails

b-schmeling:
Having the same issue.

fastali:
@steveburnett 
docker seems to have removed this account from the hub or perhaps user deleted the account themselves. what are you guys going to do about it?

tdcmeehan:
Hello folks--can you please use the images found in Presto Foundation? https://hub.docker.com/repository/docker/prestodb/presto/tags?page=1&ordering=last_updated

hamer101:
Just to mention, one has to specify the version to be pulled since getting just the latest results in 
> Error response from daemon: manifest for prestodb/presto:latest not found: manifest unknown: manifest unknown

hbazan-pp:
> Hello folks--can you please use the images found in Presto Foundation? https://hub.docker.com/repository/docker/prestodb/presto/tags?page=1&ordering=last_updated

which one?
 prestodb/presto:0.284 (there's no latest tag) fails to start because there's no /opt/presto-server/etc/config.properties file
there's one named presto-sandbox (like the ahanaio one) but it has no tags at all so it fails to pull

wanglinsong:
Docker hub ahanaio account is discontinued.
https://hub.docker.com/u/prestodb is the one used in OSS Presto (

MasterOdin:
[...]

PR #21375, opened 2023-11-13
[DOC] update deploy-docker.rst

steveburnett:
## Description
Update [Deploy Presto From a Docker Image](https://prestodb.io/docs/0.284/installation/deploy-docker.html) to use Prestodb/presto image instead of the no-longer-available ``ahanaio/prestodb-sandbox`` as described in [Issue 21341](https://github.com/prestodb/presto/issues/21341). 

Because the new Docker image does not include configuration files needed by Presto, I added steps to create the needed configuration files with the help of @kiersten-stokes and @yhwang - thanks to both of you! 

Because the new Docker image does not include the Presto CLI, I deleted the steps describing the use of the Presto CLI. A later improvement to this to consider would be to re-add or reference the steps in the [Command Line Interface](https://prestodb.io/docs/0.284/installation/cli.html) doc, but the removal of this material does not affect the content as it relates to the existing title [Deploy Presto From a Docker Image](https://prestodb.io/docs/0.284/installation/deploy-docker.html). 

## Motivation and Context
Updates the documentation to not refer to the no-longer-available Ahana sandbox Docker image, and to use a currently available Docker image. 

Fixes [Issue 21372](https://github.com/prestodb/presto/issues/21372). 

## Impact
None.

## Test Plan
Local doc builds for formatting and presentation. For the content, ran all commands as described.  

## Contributor checklist

- [x] CI passed.

## Release Notes
```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-docs/src/main/sphinx/installation/deploy-docker.rst
+++ b/presto-docs/src/main/sphinx/installation/deploy-docker.rst
@@ -4,28 +4,26 @@ Deploy Presto From a Docker Image
 
 These steps were developed and tested on Mac OS X, on both Intel and Apple Silicon chips. 
 
-Following these steps, you will:
+Follow these steps to:
 
-- install the command line tools for brew, docker, and `colima <https://github.com/abiosoft/colima>`_ tools
+- install the command line tools for brew, docker, and `Colima <https://github.com/abiosoft/colima>`_
 - verify your Docker setup
-- pull the Docker image of the Presto server and the Presto CLI
-- start your local Presto server and Presto CLI
-- query your local Presto server using the Presto CLI
-- query a remote Presto server using the Presto CLI
+- pull the Docker image of the Presto server
+- start your local Presto server
 
-Installing brew, docker, and colima
+Installing brew, Docker, and Colima
 ===================================================
 
-This task shows how to install brew, then to use brew to install docker and colima. 
+This task shows how to install brew, then to use brew to install Docker and Colima. 
 
-Note: If you have Docker installed you can skip steps 1-3. It is recommended that you 
+Note: If you have Docker installed you can skip steps 1-3, but you should 
 verify your Docker setup by running the command in step 4.
 
 1. If you do not have brew installed, run the following command:
 
    ``/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"``
 
-2. To install the docker command line and `colima <https://github.com/abiosoft/colima>`_ tools, run the following command:
+2. To install the Docker command line and `Colima <https://github.com/abiosoft/colima>`_ tools, run the following command:
 
    ``brew install docker colima``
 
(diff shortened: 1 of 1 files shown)


presto-docs/src/main/sphinx/installation/deploy-docker.rst
```diff
@@ -48,81 +46,56 @@ verify your Docker setup by running the command in step 4.
...
 ====================================
-When you no longer want the Presto Docker container, run the following two commands: 
+To remove the Presto Docker container, run the following two commands: 
 
 ``docker stop presto``
 
 ``docker rm presto``
 
-The commands return the name of the container when they are successful. 
+The commands return the name of the container ``''presto`` when they are successful.
```
kiersten-stokes, line 101:
```suggestion
The commands return the name of the container ``presto`` when they are successful. 
```

kiersten-stokes (approved):
LGTM! 

Agree that we can add more details later or in other sections if it turns out that we need to

presto-docs/src/main/sphinx/installation/deploy-docker.rst
```diff
@@ -48,81 +46,56 @@ verify your Docker setup by running the command in step 4.
...
+    -XX:+UseGCOverheadLimit
+    -XX:+ExplicitGCInvokesConcurrent
+    -XX:+HeapDumpOnOutOfMemoryError
+    -XX:+ExitOnOutOfMemoryError
+    -Djdk.attach.allowAttachSelf=true
+     
+4. To start the Presto server in the Docker container, run a command similar to the following example:
 
-5. To use the Presto CLI in your local Docker image to connect to a remote Presto server, run the following command: 
+   ``docker run -p 8080:8080 -it -v </local/path/to/config.properties>:/opt/presto-server/etc/config.properties -v </local/path/to/jvm.config>:/opt/presto-server/etc/jvm.config --name presto prestodb/presto:0.284``
```
kiersten-stokes, line 83: Now that I'm thinking about it, it would be more correct to mount the volume as a directory. The instructions would have to change slightly in order to direct users to create a directory on their local system that includes the `config.properties` and `jvm.config` files above. I believe it would also have to include a minimal `node.properties` with the following content:

```
node.environment=test
```

Then the command would be simplified to:

`docker run -p 8080:8080 -it -v </path/to/local/config/dir>:/opt/presto-server/etc --name presto prestodb/presto:0.284`

This would make it possible to restart the container as needed (among other things) without erroring out

kiersten-stokes (commented):
Even though this is the simplest way to get things up and running, it dawned on me that it perhaps would be better with the below suggested change that would require a few more lines of instruction

presto-docs/src/main/sphinx/installation/deploy-docker.rst
```diff
@@ -48,81 +46,56 @@ verify your Docker setup by running the command in step 4.
...
 
-   ``docker exec -it presto presto-cli --server Presto-endpoint-URL --user username --password``
+   Modify this example and replace ``</local/path/to/config.properties>`` and ``</local/path/to/jvm.config>`` with the paths to the ``config.properties`` and ``jvm.config`` files that you created.
 
-   You will be prompted to enter the password for the Presto user, then the ``presto>`` prompt is shown.
+   This command also assigns the name ``presto`` for the downloaded image ``prestodb/presto:0.284``.
```
yhwang, line 87:
```suggestion
   This command also assigns the name ``presto`` for the newly-created container that uses the downloaded image ``prestodb/presto:0.284``.
```

I guess this would be more accurate :-)

presto-docs/src/main/sphinx/installation/deploy-docker.rst
steveburnett: Nice suggestion! That would be a good way to approach this, if we were going to stay with the existing state of the software. 

In the interests of time for the current issue, I'd prefer not to do this as we're looking at other approaches such as proposing changes to the Presto Docker container that would remove the need for the current doc workaround, and we may not need these additional steps at all in future. Does that make sense?

steveburnett:
> LGTM!
> 
> Agree that we can add more details later or in other sections if it turns out that we need to

Exactly - I'd to approach to this part of the docs with a more modular approach if we can, for better maintainability and to make individual doc PRs smaller and easy to review and update.

presto-docs/src/main/sphinx/installation/deploy-docker.rst
kiersten-stokes: Makes complete sense!

tdcmeehan (approved):

tdcmeehan:
@wanglinsong can we hotfix 0.284 so these docs end up on the main site?

PR #21379, opened 2023-11-14
[native] Set connection time out so connection error can be retried timely

Yuhta:
Also adjust default error duration to a larger amount more stable on large cluster.


spershin (approved):
Looks good.

PR #21382, opened 2023-11-14
[native] Rename session property to confirm to Presto standard

bikramSingh91:
Updates the session property `native_debug.validate_output_from_operators`
to `native_debug_validate_output_from_operators`.


mbasmanova (approved):
Thanks. This is not a config, but session property.

`Rename session property to confirm to Presto standard`

kewang1024:
Required statuses must pass before merging

Looks like there is a required test that is failing

PR #21385, opened 2023-11-14
[native]Fix duplicate root memory pool in case of background arbitration

xiaoxmeng:
If a task from a query completes before the other tasks get scheduled on the same
worker, then the query context might have been destroyed before creating the other
tasks from the same query. The query context manager maintains a query context
cache but it only holds the weak pointer. So that when the next task from the same
query comes, it will create a new query context with the same query id and creates
a root memory pool with the same name from the memory manager. If there is a
background memory arbitration process, then the old memory pool instance might 
still be held by the memory arbitration process and run into duplicate root memory pool
issue.
This PR fixes this issue by appending a monotonically increasing id to root memory
pool name to ensure the instance name is unique.


mbasmanova (approved):
Thank you for the fix.

PR #21386, opened 2023-11-15
[native] Ensure graceful shutdown of exchange source connection pools

Yuhta:
CPU executor needs to be joined before we destroy the connection pools and HTTP IO executors, because there might be still operations in flight need to be finished/aborted using the IO pools.


xiaoxmeng (approved):
@Yuhta thanks for the fix!

PR #21403, opened 2023-11-16
[native] Fix PrestoQueryRunner in case query fails

Yuhta:
In case the query fails, `PrestoQueryRunner::execute` throws an exception.  This triggers the destructor of `SessionPool` to be called in current thread instead of the event loop thread.  Fix this by wrapping the ownership transfer inside `SCOPE_EXIT`.


mbasmanova (approved):
@Yuhta Jimmy, thank you for a quick fix.

CC: @duanmeng

PR #21406, opened 2023-11-16
[native pos] Ensure BroadcastFileReader#next is called outside the lock

arhimondr:
## Description

Delayed locking till we have a page to return

## Motivation and Context

Deadlock is possible in a following scenario:

- BroadcastExchangeSource::request acquires ExchangeQueue#mutex
- BroadcastFileReader#next tries to allocate some memory when ExchangeQueue#mutex is acquired
- MemoryManager initiates memory revoke via SharedArbitrator
- SharedArbitrator waits for all task drivers to be suspended
- The driver initiated an arbitration is considered suspended via MemoryReclaimer#enterArbitration
- Yet other drivers waiting on the shared ExchangeQueue#mutex are not considered suspended
- SharedArbitrator ends up waiting for the other threads to get suspended, hence a deadlock

## Impact

Some Presto on Spark queries are timing out when running with native execution enabled

## Test Plan

- CI
- TBD

## Contributor checklist

- [] Adequate tests were added if applicable.
- [] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/operators/BroadcastExchangeSource.cpp
+++ b/presto-native-execution/presto_cpp/main/operators/BroadcastExchangeSource.cpp
@@ -34,26 +34,28 @@ folly::SemiFuture<BroadcastExchangeSource::Response>
 BroadcastExchangeSource::request(
     uint32_t /*maxBytes*/,
     uint32_t /*maxWaitSeconds*/) {
-  std::vector<velox::ContinuePromise> promises;
-  int64_t totalBytes = 0;
-  {
-    std::lock_guard<std::mutex> l(queue_->mutex());
-    if (atEnd_) {
-      return folly::makeFuture(Response{0, true});
-    }
+  if (atEnd_) {
+    return folly::makeFuture(Response{0, true});
+  }
 
-    if (!reader_->hasNext()) {
-      atEnd_ = true;
-      queue_->enqueueLocked(nullptr, promises);
-    } else {
-      auto buffer = reader_->next();
-      totalBytes = buffer->size();
-      auto ioBuf = folly::IOBuf::wrapBuffer(buffer->as<char>(), buffer->size());
-      queue_->enqueueLocked(
-          std::make_unique<velox::exec::SerializedPage>(
-              std::move(ioBuf), [buffer](auto& /*unused*/) {}),
-          promises);
-    }
+  atEnd_ = !reader_->hasNext();
+  int64_t totalBytes = 0;
+  std::unique_ptr<velox::exec::SerializedPage> page;
+  if (!atEnd_) {
+    // Read outside the lock to avoid a potential deadlock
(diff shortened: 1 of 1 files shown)


singcha (approved):
Thanks for making this change. 
Would you please update description to reflect what changed - Delayed locking till we have a page to return

arhimondr:
cc: @mbasmanova @xiaoxmeng @patrickstuedi

mbasmanova (approved):
Thank you for the fix. Looks good. It might be helpful to add a comment to the code to explain that locking needs to be minimal and not include calls to the reader_.

patrickstuedi:
LGTM, thanks.

presto-native-execution/presto_cpp/main/operators/BroadcastExchangeSource.cpp
```diff
@@ -34,26 +34,29 @@ folly::SemiFuture<BroadcastExchangeSource::Response>
 BroadcastExchangeSource::request(
     uint32_t /*maxBytes*/,
     uint32_t /*maxWaitSeconds*/) {
-  std::vector<velox::ContinuePromise> promises;
+  if (atEnd_) {
```
spershin, line 37: We access atEnd_, reader_ w/o a lock?
Is this ok?
What is std::lock_guard<std::mutex> l(queue_->mutex()); suppose to protect?

arhimondr: It is supposed to protect access to the `queue` which can happen from multiple threads (e.g.: when different ExchangeSource's trying to add some data into the queue).

The `atEnd` field is a local field of an `ExchangeSource`. In my understanding `ExchangeSource` should not be accessed concurrently from multiple threads. The `ExchangeClient` picks an `ExchangeSource` to load under a lock and removes it from the queue to prevent other threads from calling into the same `ExchangeSource`.

presto-native-execution/presto_cpp/main/operators/BroadcastExchangeSource.cpp
```diff
@@ -34,26 +34,29 @@ folly::SemiFuture<BroadcastExchangeSource::Response>
...
-      queue_->enqueueLocked(
-          std::make_unique<velox::exec::SerializedPage>(
-              std::move(ioBuf), [buffer](auto& /*unused*/) {}),
-          promises);
-    }
+    atEnd_ = !hasNext;
+    queue_->enqueueLocked(std::move(page), promises);
   }
   for (auto& promise : promises) {
     promise.setValue();
```
spershin, line 62: Let's take
```
atEnd_ = !hasNext;
```
out of the lock protection.
Otherwise, there is a question: why is it protected here, but not in the beginning of the method.

spershin (approved):
Assuming that only queue_ needs protection.
Thanks, @arhimondr !

PR #21409, opened 2023-11-17
Fix getting type signature for UDT from uppercase type string

hantangwangd:
## Description

In some cases, we might want to get a type signature from an uppercase type define string, for example in `PrestoResultSet.getColumnInfo()`. That would cause some exception for user defined type, as QualifiedObjectName accept name string in lowercase. This PR fix the problem.

## Test Plan

 - enhanced test case `TestJdbcResultSet.testObjectTypes()`
 - newly added test case `TestTypeSignature.parseNamedTypeSignature()`

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-jdbc/src/test/java/com/facebook/presto/jdbc/TestJdbcResultSet.java
+++ b/presto-jdbc/src/test/java/com/facebook/presto/jdbc/TestJdbcResultSet.java
@@ -14,7 +14,14 @@
 package com.facebook.presto.jdbc;
 
 import com.facebook.airlift.log.Logging;
+import com.facebook.presto.functionNamespace.SqlInvokedFunctionNamespaceManagerConfig;
+import com.facebook.presto.functionNamespace.execution.NoopSqlFunctionExecutor;
+import com.facebook.presto.functionNamespace.execution.SqlFunctionExecutors;
+import com.facebook.presto.functionNamespace.testing.InMemoryFunctionNamespaceManager;
 import com.facebook.presto.server.testing.TestingPrestoServer;
+import com.facebook.presto.spi.function.FunctionImplementationType;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Lists;
 import org.testng.annotations.AfterClass;
 import org.testng.annotations.AfterMethod;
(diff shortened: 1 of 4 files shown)


hantangwangd:
@pranjalssh Could you please review for this?

tdcmeehan (approved):
LGTM.  Thoughts @ClarenceThreepwood @feilong-liu @jaystarshot ?

presto-jdbc/src/test/java/com/facebook/presto/jdbc/TestJdbcResultSet.java
```diff
@@ -205,6 +221,16 @@ public void testObjectTypes()
...
+
+        statement.execute("CREATE TYPE cat.sch.dist AS integer");
+        checkRepresentation("CAST(1 AS cat.sch.dist)", Types.JAVA_OBJECT, (rs, column) -> {
+            assertEquals(rs.getObject(1), 1);
+        });
+
+        statement.execute("CREATE TYPE cat.sch.pair AS (fst integer, snd varchar)");
+        checkRepresentation("CAST((1,'1001') AS cat.sch.pair)", Types.JAVA_OBJECT, (rs, column) -> {
+            assertEquals(rs.getObject(1), Lists.newArrayList(1, "1001"));
+        });
```
feilong-liu, line 233: Not sure if I understand correctly, do we need to add tests like "CREATE TYPE CAT.SCH.PAIR ..." to test the fix here?

feilong-liu (approved):
LGTM except one noob question

presto-jdbc/src/test/java/com/facebook/presto/jdbc/TestJdbcResultSet.java
hantangwangd: This test case covers the scenario I mentioned earlier in this PR description. That is when we try to build and get a PrestoResultSet, we pass an uppercase type define string to get a type signature, that would cause exception for user defined type. The corresponding code is in ``PrestoResultSet.getColumnInfo()``.

So I think maybe it's better to add test case for this scenario here.

PR #21285, opened 2023-10-31
[native] Support system tables

arhimondr:
## Description

Querying system tables with native execution enabled may result in incorrect result.

## Motivation and Context

Native and Java based executions have different partitioning functions implementation hence queries such the one below may return incorrect result.

```
SELECT *
FROM  (SELECT DISTINCT regionkey FROM table) t 
INNER JOIN  (SELECT regionkey FROM table$partitions) p 
ON t.regionkey = p.regionkey
```

## Impact

Incorrect results returned when querying system tables with native execution enabled

## Test Plan

Integration test

## Contributor checklist

- [x] Please make sure your submission complies with our [development](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#development), [formatting](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#formatting), [commit message](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#commit-formatting-and-pull-requests), and [attribution guidelines](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#attribution).
- [x] PR description addresses the issue accurately and concisely.  If the change is non-trivial, a GitHub Issue is referenced.
- [x] Documented new properties (with its default value), SQL syntax, functions, or other functionality.
- [x] If release notes are required, they follow the [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines).
- [x] Adequate tests were added if applicable.
- [ ] CI passed.

## Release Notes
Please follow [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines) and fill in the release notes below.

If release note is NOT required, use:

```
== NO RELEASE NOTE ==
```

PR #21416, opened 2023-11-18
 [native] SystemConnector to query system.runtime.tasks table

aditi-pandit:
## Description
SystemConnector is a Presto Connector for system tables. System tables include runtime schema tables like system.runtime.{nodes|tasks|queries|transactions}, properties tables (table properties, schema properties, column properties, analyze properties), hive & iceberg metadata tables.

SystemConnector tables are unique in a way that all of them are populated from metadata structures on the co-ordinator (and optionally from workers). This metadata can be internal process metadata for the runtime tables or metadata obtained from HMS/Iceberg catalog.

The distribution of the SystemTable can be ALL_NODES, ALL_COORDINATORS, SINGLE_COORDINATOR (from https://github.com/prestodb/presto/blob/master/presto-spi/src/main/java/com/facebook/presto/spi/SystemTable.java#L24)

Only one table 'system.runtime.tasks' is populated on ALL_NODES. So this table gets results from the co-ordinator as well as the workers.

In the past, querying this table was broken on Prestissimo since there was no system catalog/connector on workers. This PR enhances the native workers with a System connector/catalog that is used to populate the tasks table. The SystemConnector uses Presto TaskManager task map to populate this table.

There is one more design point. The Java co-ordinator is not fully compatible with native workers. They use different hash functions and different intermediate results for aggregations. So some changes were needed for running the system connector on the co-ordinator. 

The changes are :

- https://github.com/prestodb/presto/pull/21725 
- https://github.com/prestodb/presto/pull/21285 : This PR adds a GATHER exchange after a system table TableScanNode so that all the data gathers at a worker to make the partitioning consistent with all other tables. 

Both of these planning rules are applicable for tasks table as well since it generates data at both co-ordinator and workers.

## Motivation and Context
system.runtime.tasks table is very frequently used in deployment scripts. Querying this table was broken in Prestissimo.

## Test Plan
Added e2e tests

```
== RELEASE NOTES ==

General Changes
* Add support for querying system.runtime.tasks table in native clusters

```

### DIFF
--- a/presto-native-execution/presto_cpp/main/PrestoServer.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoServer.cpp
@@ -21,6 +21,7 @@
 #include "presto_cpp/main/Announcer.h"
 #include "presto_cpp/main/PeriodicTaskManager.h"
 #include "presto_cpp/main/SignalHandler.h"
+#include "presto_cpp/main/SystemConnector.h"
 #include "presto_cpp/main/TaskResource.h"
 #include "presto_cpp/main/common/ConfigReader.h"
 #include "presto_cpp/main/common/Counters.h"
@@ -232,6 +233,17 @@ void PrestoServer::run() {
       std::make_unique<IcebergPrestoToVeloxConnector>("iceberg"));
   registerPrestoToVeloxConnector(
       std::make_unique<TpchPrestoToVeloxConnector>("tpch"));
+  // Presto server uses system catalog or system schema in other catalogs
+  // in different places in the code. All these resolve to the SystemConnector.
+  // Depending on where the operator or column is used, different prefixes can
+  // be used in the naming. So the protocol class is mapped
+  // to all the different prefixes for System tables/columns.
+  registerPrestoToVeloxConnector(
+      std::make_unique<SystemPrestoToVeloxConnector>("$system"));
+  registerPrestoToVeloxConnector(
+      std::make_unique<SystemPrestoToVeloxConnector>("system"));
+  registerPrestoToVeloxConnector(
+      std::make_unique<SystemPrestoToVeloxConnector>("$system@system"));
 
   initializeVeloxMemory();
   initializeThreadPools();
@@ -438,6 +450,7 @@ void PrestoServer::run() {
   }
   prestoServerOperations_ =
       std::make_unique<PrestoServerOperations>(taskManager_.get(), this);
+  registerSystemConnector();
 
   // The endpoint used by operation in production.
   httpServer_->registerGet(
@@ -939,6 +952,15 @@ std::vector<std::string> PrestoServer::registerConnectors(
   return catalogNames;
 }
 
+void PrestoServer::registerSystemConnector() {
+  PRESTO_STARTUP_LOG(INFO) << "Registering system catalog "
+                           << " using connector SystemConnector";
+  VELOX_CHECK(taskManager_);
+  auto systemConnector =
+      std::make_shared<SystemConnector>("$system@system", taskManager_.get());
+  velox::connector::registerConnector(systemConnector);
+}
+
 void PrestoServer::unregisterConnectors() {
   PRESTO_SHUTDOWN_LOG(INFO) << "Unregistering connectors";
   auto connectors = facebook::velox::connector::getAllConnectors();
@@ -959,6 +981,7 @@ void PrestoServer::unregisterConnectors() {
     }
   }
 
+  facebook::velox::connector::unregisterConnector("$system@system");
   PRESTO_SHUTDOWN_LOG(INFO)
       << "Unregistered " << connectors.size() << " connectors";
 }
--- a/presto-native-execution/presto_cpp/main/SystemConnector.cpp
+++ b/presto-native-execution/presto_cpp/main/SystemConnector.cpp
@@ -1,0 +1,396 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "presto_cpp/main/SystemConnector.h"
+#include "presto_cpp/main/PrestoTask.h"
+#include "presto_cpp/main/TaskManager.h"
+
+#include "velox/type/Timestamp.h"
+
+namespace facebook::presto {
+
+using namespace velox;
+
+namespace {
+
+static const std::string kRuntimeSchema = "runtime";
+static const std::string kTasksTable = "tasks";
+
+} // namespace
+
+const velox::RowTypePtr SystemTableHandle::taskSchema() {
+  static std::vector<std::string> kTaskColumnNames = {
+      "node_id",
+      "task_id",
+      "stage_execution_id",
+      "stage_id",
+      "query_id",
+      "state",
+      "splits",
+      "queued_splits",
+      "running_splits",
+      "completed_splits",
+      "split_scheduled_time_ms",
+      "split_cpu_time_ms",
+      "split_blocked_time_ms",
+      "raw_input_bytes",
+      "raw_input_rows",
+      "processed_input_bytes",
+      "processed_input_rows",
+      "output_bytes",
+      "output_rows",
+      "physical_written_bytes",
+      "created",
+      "start",
+      "last_heartbeat",
+      "end"};
+
+  static std::vector<velox::TypePtr> kTaskColumnTypes = {
+      velox::VARCHAR(),   velox::VARCHAR(),   velox::VARCHAR(),
+      velox::VARCHAR(),   velox::VARCHAR(),   velox::VARCHAR(),
+      velox::BIGINT(),    velox::BIGINT(),    velox::BIGINT(),
+      velox::BIGINT(),    velox::BIGINT(),    velox::BIGINT(),
+      velox::BIGINT(),    velox::BIGINT(),    velox::BIGINT(),
+      velox::BIGINT(),    velox::BIGINT(),    velox::BIGINT(),
+      velox::BIGINT(),    velox::BIGINT(),    velox::TIMESTAMP(),
+      velox::TIMESTAMP(), velox::TIMESTAMP(), velox::TIMESTAMP()};
+  static const RowTypePtr kTaskSchema =
+      ROW(std::move(kTaskColumnNames), std::move(kTaskColumnTypes));
+  return kTaskSchema;
+}
+
+SystemTableHandle::SystemTableHandle(
+    std::string connectorId,
+    std::string schemaName,
+    std::string tableName)
+    : ConnectorTableHandle(std::move(connectorId)),
+      schemaName_(std::move(schemaName)),
+      tableName_(std::move(tableName)) {
+  VELOX_USER_CHECK_EQ(
+      schemaName_,
+      kRuntimeSchema,
+      "SystemConnector supports only runtime schema");
+  VELOX_USER_CHECK_EQ(
+      tableName_, kTasksTable, "SystemConnector supports only tasks table");
+}
+
+std::string SystemTableHandle::toString() const {
+  return fmt::format("schema: {} table: {}", schemaName_, tableName_);
+}
+
+SystemDataSource::SystemDataSource(
+    const std::shared_ptr<const RowType>& outputType,
+    const std::shared_ptr<connector::ConnectorTableHandle>& tableHandle,
+    const std::unordered_map<
+        std::string,
+        std::shared_ptr<connector::ColumnHandle>>& columnHandles,
+    const TaskManager* taskManager,
+    velox::memory::MemoryPool* FOLLY_NONNULL pool)
+    : taskManager_(taskManager), pool_(pool) {
+  auto systemTableHandle =
+      std::dynamic_pointer_cast<SystemTableHandle>(tableHandle);
+  VELOX_CHECK_NOT_NULL(
+      systemTableHandle,
+      "TableHandle must be an instance of SystemTableHandle");
+
+  outputColumnMappings_.reserve(outputType->names().size());
+  auto taskSchema = systemTableHandle->taskSchema();
+  for (const auto& outputName : outputType->names()) {
+    auto it = columnHandles.find(outputName);
+    VELOX_CHECK(
+        it != columnHandles.end(),
+        "ColumnHandle is missing for output column '{}'",
+        outputName);
+
+    auto handle = std::dynamic_pointer_cast<SystemColumnHandle>(it->second);
+    VELOX_CHECK_NOT_NULL(
+        handle,
+        "ColumnHandle must be an instance of SystemColumnHandle "
+        "for '{}' on table '{}'",
+        handle->name(),
+        systemTableHandle->tableName());
+
+    auto columnIndex = taskSchema->getChildIdxIfExists(handle->name());
+    VELOX_CHECK(
+        columnIndex.has_value(),
+        "Column {} not found in SystemTable task schema",
+        handle->name());
+    outputColumnMappings_.push_back(columnIndex.value());
+  }
+
+  outputType_ = outputType;
+}
+
+void SystemDataSource::addSplit(
+    std::shared_ptr<connector::ConnectorSplit> split) {
+  VELOX_CHECK_NULL(
+      currentSplit_,
+      "Previous split has not been processed yet. Call next() to process the split.");
+  currentSplit_ = std::dynamic_pointer_cast<SystemSplit>(split);
+  VELOX_CHECK(currentSplit_, "Wrong type of split for SystemDataSource.");
+}
+
+#define SET_TASK_COLUMN(value)            \
+  int j = 0;                              \
+  for (const auto& taskEntry : taskMap) { \
+    auto task = taskEntry.second;         \
+    auto taskInfo = taskInfos[j];         \
+    flat->set(j, value);                  \
+    j++;                                  \
+  }
+
+#define SET_TASK_FMT_COLUMN(value)        \
+  int j = 0;                              \
+  std::string temp;                       \
+  for (const auto& taskEntry : taskMap) { \
+    auto task = taskEntry.second;         \
+    auto taskInfo = taskInfos[j];         \
+    temp = fmt::format("{}", value);      \
+    flat->set(j, StringView(temp));       \
+    j++;                                  \
+  }
+
+RowVectorPtr SystemDataSource::getTaskResults() {
+  auto taskMap = taskManager_->tasks();
+  auto numRows = taskMap.size();
+
+  std::vector<protocol::TaskInfo> taskInfos;
+  taskInfos.reserve(numRows);
+  for (const auto& taskEntry : taskMap) {
+    taskInfos.push_back(taskEntry.second->updateInfo());
+  }
+
+  auto result = std::dynamic_pointer_cast<RowVector>(
+      BaseVector::create(outputType_, numRows, pool_));
+
+  static constexpr int64_t kNanosecondsInMillisecond = 1'000'000;
+  auto toMillis = [](int64_t nanos) -> int64_t {
+    return nanos / kNanosecondsInMillisecond;
+  };
+  for (auto i = 0; i < outputColumnMappings_.size(); i++) {
+    result->childAt(i)->resize(numRows);
+    auto taskColumn = outputColumnMappings_.at(i);
+    auto taskEnum = TaskColumnEnum(taskColumn);
+    switch (taskEnum) {
+      case TaskColumnEnum::kNodeId: {
+        auto flat = result->childAt(i)->as<FlatVector<StringView>>();
+        SET_TASK_COLUMN(StringView(taskInfo.nodeId));
+        break;
+      }
+
+      case TaskColumnEnum::kTaskId: {
+        auto flat = result->childAt(i)->as<FlatVector<StringView>>();
+        SET_TASK_COLUMN(StringView(taskInfo.taskId));
+        break;
+      }
+
+      case TaskColumnEnum::kStageExecutionId: {
+        auto flat = result->childAt(i)->as<FlatVector<StringView>>();
+        SET_TASK_FMT_COLUMN(task->id.stageExecutionId());
+        break;
+      }
+
+      case TaskColumnEnum::kStageId: {
+        auto flat = result->childAt(i)->as<FlatVector<StringView>>();
+        SET_TASK_FMT_COLUMN(task->id.stageId());
+        break;
+      }
+
+      case TaskColumnEnum::kQueryId: {
+        auto flat = result->childAt(i)->as<FlatVector<StringView>>();
+        SET_TASK_FMT_COLUMN(task->id.queryId());
+        break;
+      }
+
+      case TaskColumnEnum::kState: {
+        auto flat = result->childAt(i)->as<FlatVector<StringView>>();
+        SET_TASK_FMT_COLUMN(json(taskInfo.taskStatus.state).dump());
+        break;
+      }
+
+      case TaskColumnEnum::kSplits: {
+        auto flat = result->childAt(i)->as<FlatVector<int64_t>>();
+        SET_TASK_COLUMN(taskInfo.stats.totalDrivers);
+        break;
+      }
+
+      case TaskColumnEnum::kQueuedSplits: {
+        auto flat = result->childAt(i)->as<FlatVector<int64_t>>();
+        SET_TASK_COLUMN(taskInfo.stats.queuedDrivers);
+        break;
+      }
+
+      case TaskColumnEnum::kRunningSplits: {
+        auto flat = result->childAt(i)->as<FlatVector<int64_t>>();
+        SET_TASK_COLUMN(taskInfo.stats.runningDrivers);
+        break;
+      }
+
+      case TaskColumnEnum::kCompletedSplits: {
+        auto flat = result->childAt(i)->as<FlatVector<int64_t>>();
+        SET_TASK_COLUMN(taskInfo.stats.completedDrivers);
+        break;
+      }
+
+      case TaskColumnEnum::kSplitScheduledTimeMs: {
+        auto flat = result->childAt(i)->as<FlatVector<int64_t>>();
+        SET_TASK_COLUMN(toMillis(taskInfo.stats.totalScheduledTimeInNanos));
+        break;
+      }
+
+      case TaskColumnEnum::kSplitCpuTimeMs: {
+        auto flat = result->childAt(i)->as<FlatVector<int64_t>>();
+        SET_TASK_COLUMN(toMillis(taskInfo.stats.totalCpuTimeInNanos));
+        break;
+      }
+
+      case TaskColumnEnum::kSplitBlockedTimeMs: {
+        auto flat = result->childAt(i)->as<FlatVector<int64_t>>();
+        SET_TASK_COLUMN(toMillis(taskInfo.stats.totalBlockedTimeInNanos));
+        break;
+      }
+
+      case TaskColumnEnum::kRawInputBytes: {
+        auto flat = result->childAt(i)->as<FlatVector<int64_t>>();
+        SET_TASK_COLUMN(taskInfo.stats.rawInputDataSizeInBytes);
+        break;
+      }
+
+      case TaskColumnEnum::kRawInputRows: {
+        auto flat = result->childAt(i)->as<FlatVector<int64_t>>();
+        SET_TASK_COLUMN(taskInfo.stats.rawInputPositions);
+        break;
+      }
+
+      case TaskColumnEnum::kProcessedInputBytes: {
+        auto flat = result->childAt(i)->as<FlatVector<int64_t>>();
+        SET_TASK_COLUMN(taskInfo.stats.processedInputDataSizeInBytes);
+        break;
+      }
+
+      case TaskColumnEnum::kProcessedInputRows: {
+        auto flat = result->childAt(i)->as<FlatVector<int64_t>>();
+        SET_TASK_COLUMN(taskInfo.stats.processedInputPositions);
+        break;
+      }
+
+      case TaskColumnEnum::kOutputBytes: {
+        auto flat = result->childAt(i)->as<FlatVector<int64_t>>();
+        SET_TASK_COLUMN(taskInfo.stats.outputDataSizeInBytes);
+        break;
+      }
+
+      case TaskColumnEnum::kOutputRows: {
+        auto flat = result->childAt(i)->as<FlatVector<int64_t>>();
+        SET_TASK_COLUMN(taskInfo.stats.outputPositions);
+        break;
+      }
+
+      case TaskColumnEnum::kPhysicalWrittenBytes: {
+        auto flat = result->childAt(i)->as<FlatVector<int64_t>>();
+        SET_TASK_COLUMN(taskInfo.stats.physicalWrittenDataSizeInBytes);
+        break;
+      }
+
+      case TaskColumnEnum::kCreated: {
+        auto flat = result->childAt(i)->as<FlatVector<Timestamp>>();
+        SET_TASK_COLUMN(velox::Timestamp::fromMillis(task->createTimeMs));
+        break;
+      }
+
+      case TaskColumnEnum::kStart: {
+        auto flat = result->childAt(i)->as<FlatVector<Timestamp>>();
+        SET_TASK_COLUMN(
+            velox::Timestamp::fromMillis(task->firstSplitStartTimeMs));
+        break;
+      }
+
+      case TaskColumnEnum::kLastHeartBeat: {
+        auto flat = result->childAt(i)->as<FlatVector<Timestamp>>();
+        SET_TASK_COLUMN(velox::Timestamp::fromMillis(task->lastHeartbeatMs));
+        break;
+      }
+
+      case TaskColumnEnum::kEnd: {
+        auto flat = result->childAt(i)->as<FlatVector<Timestamp>>();
+        SET_TASK_COLUMN(velox::Timestamp::fromMillis(task->lastEndTimeMs));
+        break;
+      }
+    }
+  }
+  return result;
+}
+
+std::optional<RowVectorPtr> SystemDataSource::next(
+    uint64_t size,
+    velox::ContinueFuture& /*future*/) {
+  if (!currentSplit_) {
+    return nullptr;
+  }
+
+  auto result = getTaskResults();
+  completedRows_ += result->size();
+  completedBytes_ += result->estimateFlatSize();
+
+  currentSplit_ = nullptr;
+
+  return result;
+}
+
+std::unique_ptr<velox::connector::ConnectorSplit>
+SystemPrestoToVeloxConnector::toVeloxSplit(
+    const protocol::ConnectorId& catalogId,
+    const protocol::ConnectorSplit* const connectorSplit) const {
+  auto systemSplit = dynamic_cast<const protocol::SystemSplit*>(connectorSplit);
+  VELOX_CHECK_NOT_NULL(
+      systemSplit, "Unexpected split type {}", connectorSplit->_type);
+  return std::make_unique<SystemSplit>(
+      catalogId,
+      systemSplit->tableHandle.schemaName,
+      systemSplit->tableHandle.tableName);
+}
+
+std::unique_ptr<velox::connector::ColumnHandle>
+SystemPrestoToVeloxConnector::toVeloxColumnHandle(
+    const protocol::ColumnHandle* column,
+    const TypeParser& typeParser) const {
+  auto systemColumn = dynamic_cast<const protocol::SystemColumnHandle*>(column);
+  VELOX_CHECK_NOT_NULL(
+      systemColumn, "Unexpected column handle type {}", column->_type);
+  return std::make_unique<SystemColumnHandle>(systemColumn->columnName);
+}
+
+std::unique_ptr<velox::connector::ConnectorTableHandle>
+SystemPrestoToVeloxConnector::toVeloxTableHandle(
+    const protocol::TableHandle& tableHandle,
+    const VeloxExprConverter& exprConverter,
+    const TypeParser& typeParser,
+    std::unordered_map<
+        std::string,
+        std::shared_ptr<velox::connector::ColumnHandle>>& assignments) const {
+  auto systemLayout =
+      std::dynamic_pointer_cast<const protocol::SystemTableLayoutHandle>(
+          tableHandle.connectorTableLayout);
+  VELOX_CHECK_NOT_NULL(
+      systemLayout, "Unexpected table handle type {}", tableHandle.connectorId);
+  return std::make_unique<SystemTableHandle>(
+      tableHandle.connectorId,
+      systemLayout->table.schemaName,
+      systemLayout->table.tableName);
+}
+
+std::unique_ptr<protocol::ConnectorProtocol>
+SystemPrestoToVeloxConnector::createConnectorProtocol() const {
+  return std::make_unique<protocol::SystemConnectorProtocol>();
+}
+} // namespace facebook::presto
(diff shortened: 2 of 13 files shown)


presto-native-execution/presto_cpp/main/SystemConnector.h
```diff
@@ -0,0 +1,178 @@
...
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+#pragma once
+
+#include "presto_cpp/main/SystemSplit.h"
+#include "presto_cpp/presto_protocol/presto_protocol.h"
```
majetideepak, line 17: why do we need the protocol header here?

presto-native-execution/presto_cpp/main/SystemConnector.cpp
```diff
@@ -0,0 +1,174 @@
...
+#include "presto_cpp/main/PrestoTask.h"
+#include "presto_cpp/main/TaskManager.h"
+
+namespace facebook::presto {
+
+using namespace velox;
+
+SystemTableHandle::SystemTableHandle(std::string connectorId, std::string schemaName, std::string tableName) :
+      ConnectorTableHandle(std::move(connectorId)), schemaName_(std::move(schemaName)), tableName_(std::move(tableName)){
+  VELOX_USER_CHECK_EQ(schemaName_, "runtime", "SystemConnector supports only runtime schema");
```
majetideepak, line 24: use constants for `"runtime"` and `"tasks"` here and below.

presto-native-execution/presto_cpp/main/SystemConnector.cpp
```diff
@@ -0,0 +1,174 @@
...
+std::optional<RowVectorPtr> SystemDataSource::next(
+    uint64_t size,
+    velox::ContinueFuture& /*future*/) {
+  if (!currentSplit_) {
+    return nullptr;
+  }
+
+  TaskMap taskMap = taskManager_->tasks();
+  auto numRows = taskMap.size();
+  getTaskResults(taskMap, taskTableResult_);
```
majetideepak, line 159: We can avoid `taskTableResult_` if we pass the desired columns (`outputColumnMappings_`) and `result` vector to `getTaskResults`.

presto-native-execution/presto_cpp/main/SystemConnector.h
```diff
@@ -0,0 +1,178 @@
...
+
+  const velox::RowTypePtr& taskSchema() {
+    return kTaskSchema_;
+  }
+
+ private:
+  const std::string schemaName_;
+  const std::string tableName_;
+
+  velox::RowTypePtr kTaskSchema_;
```
majetideepak, line 62: This can be a `static const`

presto-native-execution/presto_cpp/main/SystemConnector.cpp
```diff
@@ -0,0 +1,174 @@
...
+namespace facebook::presto {
+
+using namespace velox;
+
+SystemTableHandle::SystemTableHandle(std::string connectorId, std::string schemaName, std::string tableName) :
+      ConnectorTableHandle(std::move(connectorId)), schemaName_(std::move(schemaName)), tableName_(std::move(tableName)){
+  VELOX_USER_CHECK_EQ(schemaName_, "runtime", "SystemConnector supports only runtime schema");
+  VELOX_USER_CHECK_EQ(tableName_, "tasks", "SystemConnector supports only tasks table");
+
+  std::vector<std::string> kTaskColumnNames =
```
majetideepak, line 27: const

presto-native-execution/presto_cpp/main/SystemConnector.cpp
```diff
@@ -0,0 +1,174 @@
...
+      {"node_id", "task_id", "stage_execution_id", "stage_id", "query_id", "state", "splits", "queued_splits", "running_splits", "completed_splits",
+       "split_scheduled_time_ms", "split_cpu_time_ms", "split_blocked_time_ms", "raw_input_bytes", "raw_input_rows", "processed_input_bytes",
+       "processed_input_rows", "output_bytes", "output_rows", "physical_written_bytes", "created", "start", "last_heartbeat", "end"};
+  std::vector<velox::TypePtr> kTaskColumnTypes =
```
majetideepak, line 31: const

presto-native-execution/presto_cpp/presto_protocol/Connectors.cpp
```diff
@@ -48,6 +48,12 @@ void registerTpchConnector() {
   registerConnector("tpch", "tpch");
 }
 
+void registerSystemConnector() {
+  registerConnector("$system", "$system");
```
majetideepak, line 52: Is there any significance to each of the catalog names?
How do `$system`, `system`, and `$system@system` differ?

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeGeneralQueries.java
```diff
@@ -1337,6 +1337,11 @@ public void testSelectFieldsWithCapitalLetters()
         }
     }
 
+    @Test
+    public void testTaskTable() {
```
majetideepak, line 1341: Is it guaranteed to have at least 2 tasks? Just ensuring the test will be stable.
Can we add more tests corresponding to the `select * from system.runtime.tasks` tasks?

majetideepak (commented):
@aditi-pandit nice change! some comments.

mbasmanova (commented):
@aditi-pandit Is there a design document about this connector? I assume SystemConnector requires access to Metastore and we don't have that on the worker. Hence, wondering how this will work?

mbasmanova:
CC: @spershin

aditi-pandit:
> @aditi-pandit Is there a design document about this connector? I assume SystemConnector requires access to Metastore and we don't have that on the worker. Hence, wondering how this will work?

@mbasmanova : This SystemConnector code was to query the tasks table. That seemed the only part of the SystemConnector that was needed at the worker for Prestissimo. 

Other tables like nodes, queries were populated from in-memory structures in the co-ordinator itself. Any code accessing Metastore (like TablePropertiesSystemTable say) seemed to be required only at the co-ordinator part of the connector.

I just spent a day on this prototype to wire the pieces. I haven't put together a design doc.

presto-native-execution/presto_cpp/main/SystemConnector.cpp
```diff
@@ -0,0 +1,251 @@
...
+  for (auto i = 0; i < taskTableResult->type()->size(); i++) {
+    taskTableResult->childAt(i)->resize(numRows);
+  }
+
+  auto toMillis = [](int64_t nanos) -> int64_t { return nanos / 1'000'000; };
+  int i = 0;
+  for (const auto& taskEntry : taskMap) {
+    auto task = taskEntry.second;
+    auto taskInfo = task->updateInfo();
+    taskTableResult->childAt(0)->asFlatVector<StringView>()->set(
```
mbasmanova, line 158: nit: it is hard to tell what column 0 is or what column 15 is. Perhaps, defined constants for column names and use them here and when building kTaskColumnNames.

presto-native-execution/presto_cpp/main/SystemConnector.cpp
```diff
@@ -0,0 +1,251 @@
...
+        i, taskInfo.stats.processedInputDataSizeInBytes);
+    taskTableResult->childAt(16)->asFlatVector<int64_t>()->set(
+        i, taskInfo.stats.processedInputPositions);
+    taskTableResult->childAt(17)->asFlatVector<int64_t>()->set(
+        i, taskInfo.stats.outputDataSizeInBytes);
+    taskTableResult->childAt(18)->asFlatVector<int64_t>()->set(
+        i, taskInfo.stats.outputPositions);
+    taskTableResult->childAt(19)->asFlatVector<int64_t>()->set(
+        i, taskInfo.stats.physicalWrittenDataSizeInBytes);
+    // Setting empty TIMESTAMP for below fields as string -> Timestamp is not
```
mbasmanova, line 202: This is hacky. Velox reports timestamps as integers, but then we use PrestoTask::updateInfoLocked to convert these to strings, and then complain here that we cannot convert them back to integers. Please, consider skipping unnecessary conversions.

```
  prestoTaskStats.createTime =
      util::toISOTimestamp(taskStats.executionStartTimeMs);
  prestoTaskStats.firstStartTime =
      util::toISOTimestamp(taskStats.firstSplitStartTimeMs);
  prestoTaskStats.lastStartTime =
      util::toISOTimestamp(taskStats.lastSplitStartTimeMs);
  prestoTaskStats.lastEndTime =
      util::toISOTimestamp(taskStats.executionEndTimeMs);

  /// Epoch time (ms) when task starts to run
  uint64_t executionStartTimeMs{0};

```

presto-native-execution/presto_cpp/main/SystemConnector.cpp
mbasmanova: getTaskResults can be a private method and it can return RowVector directly.

mbasmanova (commented):
@aditi-pandit Aditi, thank you for clarifying. It is interesting that tasks table is populated on the workers. I wonder why. All the information is available on the coordinator. CC: @tdcmeehan

tdcmeehan:
> @aditi-pandit Aditi, thank you for clarifying. It is interesting that tasks table is populated on the workers. I wonder why. All the information is available on the coordinator. CC: @tdcmeehan

I think the reason for this is because historically you could always deploy Presto in a mode where many or all of the workers also functioned as coordinators.  In this mode, any single coordinator would only know of the tasks that are local to that coordinator.

mbasmanova:
> I think the reason for this is because historically you could always deploy Presto in a mode where many or all of the workers also functioned as coordinators. In this mode, any single coordinator would only know of the tasks whose queries are local to that coordinator.

@tdcmeehan Tim, thank you for clarifying. I didn't know about this deployment scheme. I'm not sure I understand how this works though. When there are multiple coordinators, wouldn't query results depend on which coordinator is being asked to process the query? Are you saying that in this setup a query can be routed to any coordinator and the results are expected to be the same? I guess in this case it is necessary to ask all the workers to report their tasks since as you pointed out a single coordinator knows about a subset of tasks only.

tdcmeehan:
> > I think the reason for this is because historically you could always deploy Presto in a mode where many or all of the workers also functioned as coordinators. In this mode, any single coordinator would only know of the tasks whose queries are local to that coordinator.
> 
> @tdcmeehan Tim, thank you for clarifying. I didn't know about this deployment scheme. I'm not sure I understand how this works though. When there are multiple coordinators, wouldn't query results depend on which coordinator is being asked to process the query? Are you saying that in this setup a query can be routed to any coordinator and the results are expected to be the same? I guess in this case it is necessary to ask all the workers to report their tasks since as you pointed out a single coordinator knows about a subset of tasks only.

In this scheme, queries are sticky to a single coordinator (after you POST a query, each `nextUri` always returns the host of the local coordinator).  It's presumed there's something in front of the cluster to distribute the query creation, like a load balancer.

mbasmanova:
@tdcmeehan Tim, thank you for clarifying. One more follow-up question. In a multi-coordinator deployment, do all workers report themselves to all coordinators or a given worker is fixed-assigned to just one coordinator? In other words, do we have N coordinators managing a shared pool of workers or we have just N "mini" clusters that are independent of each other?

tdcmeehan:
> @tdcmeehan Tim, thank you for clarifying. One more follow-up question. In a multi-coordinator deployment, do all workers report themselves to all coordinators or a given worker is fixed-assigned to just one coordinator? In other words, do we have N coordinators managing a shared pool of workers or we have just N "mini" clusters that are independent of each other?

Workers report themselves to a single discovery service, which is replicated to other works in an eventually consistent manner.  Originally, when this system connector was written, there was no concept of shared resources (e.g. resource groups, global memory management, etc.) and it relied purely on individual backpressure from workers, although there are now tools to help make that work.

mbasmanova:
@tdcmeehan Tim, I wonder if it still makes sense to support this deployment model. What do you think? Does it makes sense to consider it when thinking about native workers?

tdcmeehan:
Tactically and short term, I think it would be great to support this if there was an easy and not hacky way to get it to work with #21725 and #21285.  But given that most people would be deploying Presto for their large to medium size data lakes, I don't think an Impala-style deployment model makes sense for Presto's future, and personally I feel comfortable saying we can deprecate it in the future.

That being said, system tables in the coordinator present a challenge for what I feel is one of the end goals of moving to native, which is simplifying our Java code.  I'd like to think about a way to move this to C++ so it doesn't need to be present in the Java SPI (thinking way ahead in the future, if the only reason we retain page source providers is for system tables, I think it would be worthwhile to think about how to move system tables to C++).  So I'd like to revisit the presumption at some point that system tables must be coordinator-provided tables, since even now that's not necessarily true.

aditi-pandit:
@mbasmanova, @tdcmeehan : If we want to stay with this approach of getting tasks table on worker we could modify https://github.com/prestodb/presto/pull/21725 and https://github.com/prestodb/presto/pull/21285 to not perform those rewrites for system.runtime.tasks table specifically as it based on the worker. 

https://github.com/prestodb/presto/pull/21725 could work un-modified as well. It would just mean that we don't allow partial agg over the tasks table which might not be a big deal unless a massive numbers of queries are scheduled in the cluster.

wdyt ?

majetideepak:
The other fixable issue we are hitting internally in a large setup when querying system tables is that the Native worker does not handle chunked HTTP responses yet. @tdcmeehan do you know what causes a chunked HTTP response from the coordinator? I tried reproducing with a large system table (many entries) but I could not.

https://github.com/prestodb/presto/blob/master/presto-native-execution/presto_cpp/main/PrestoExchangeSource.cpp#L246

mbasmanova:
@majetideepak Chunked response used to be produced by task/async endpoint which was removed in https://github.com/prestodb/presto/pull/21772 . You should not be seeing issues if you update past that PR.

majetideepak:
@mbasmanova thank you for the pointer!

presto-native-execution/presto_cpp/main/SystemConnector.h
aditi-pandit: removed.

presto-native-execution/presto_cpp/main/SystemConnector.h
```diff
@@ -0,0 +1,211 @@
...
+};
+
+class SystemDataSource : public velox::connector::DataSource {
+ public:
+  SystemDataSource(
+      const std::shared_ptr<const velox::RowType>& outputType,
```
mbasmanova, line 70: RowTypePtr

presto-native-execution/presto_cpp/main/SystemConnector.h
```diff
@@ -0,0 +1,211 @@
...
+      const std::shared_ptr<velox::connector::ConnectorTableHandle>&
+          tableHandle,
+      const std::unordered_map<
+          std::string,
+          std::shared_ptr<velox::connector::ColumnHandle>>& columnHandles,
+      const TaskManager* taskManager,
+      velox::memory::MemoryPool* FOLLY_NONNULL pool);
```
mbasmanova, line 77: We stopped using FOLLY_NONNULL in Velox. Do we still need these in Prestissimo?

presto-native-execution/presto_cpp/main/SystemConnector.cpp
```diff
@@ -0,0 +1,297 @@
...
+    std::string tableName)
+    : ConnectorTableHandle(std::move(connectorId)),
+      schemaName_(std::move(schemaName)),
+      tableName_(std::move(tableName)) {
+  VELOX_USER_CHECK_EQ(
+      schemaName_, "runtime", "SystemConnector supports only runtime schema");
+  VELOX_USER_CHECK_EQ(
+      tableName_, "tasks", "SystemConnector supports only tasks table");
+
+  std::vector<std::string> kTaskColumnNames = {
```
mbasmanova, line 34: static

presto-native-execution/presto_cpp/main/SystemConnector.cpp
```diff
@@ -0,0 +1,297 @@
...
+      "processed_input_bytes",
+      "processed_input_rows",
+      "output_bytes",
+      "output_rows",
+      "physical_written_bytes",
+      "created",
+      "start",
+      "last_heartbeat",
+      "end"};
+  std::vector<velox::TypePtr> kTaskColumnTypes = {
```
mbasmanova, line 59: static

presto-native-execution/presto_cpp/main/SystemConnector.cpp
```diff
@@ -0,0 +1,297 @@
...
+        std::shared_ptr<connector::ColumnHandle>>& columnHandles,
+    const TaskManager* taskManager,
+    velox::memory::MemoryPool* FOLLY_NONNULL pool)
+    : taskManager_(taskManager), pool_(pool) {
+  auto systemTableHandle =
+      std::dynamic_pointer_cast<SystemTableHandle>(tableHandle);
+  VELOX_CHECK_NOT_NULL(
+      systemTableHandle,
+      "TableHandle must be an instance of SystemTableHandle");
+  VELOX_USER_CHECK_EQ(
```
mbasmanova, line 89: These checks are redundant as they duplicate checks in SystemTableHandle's constructor.

presto-native-execution/presto_cpp/main/SystemConnector.cpp
mbasmanova: @aditi-pandit What do you think?

presto-native-execution/presto_cpp/main/SystemConnector.cpp
```diff
@@ -0,0 +1,297 @@
...
+}
+
+void SystemDataSource::addSplit(
+    std::shared_ptr<connector::ConnectorSplit> split) {
+  VELOX_CHECK_NULL(
+      currentSplit_,
+      "Previous split has not been processed yet. Call next() to process the split.");
+  currentSplit_ = std::dynamic_pointer_cast<SystemSplit>(split);
+  VELOX_CHECK(currentSplit_, "Wrong type of split for SystemDataSource.");
+  VELOX_USER_CHECK_EQ(
```
mbasmanova, line 132: duplicate checks?

presto-native-execution/presto_cpp/main/SystemConnector.cpp
```diff
@@ -0,0 +1,297 @@
...
+  for (auto i = 0; i < taskTableResult->type()->size(); i++) {
+    taskTableResult->childAt(i)->resize(numRows);
+  }
+
+  auto toMillis = [](int64_t nanos) -> int64_t { return nanos / 1'000'000; };
+  int i = 0;
+  for (const auto& taskEntry : taskMap) {
+    auto task = taskEntry.second;
+    auto taskInfo = task->updateInfo();
+    taskTableResult->childAt(0)->asFlatVector<StringView>()->set(
```
mbasmanova, line 157: > taskTableResult->childAt(0)->asFlatVector<StringView>()

These dynamic casts per row are expensive.

presto-native-execution/presto_cpp/main/SystemConnector.cpp
```diff
@@ -0,0 +1,297 @@
...
+        i, StringView(taskInfo.nodeId));
+    taskTableResult->childAt(1)->asFlatVector<StringView>()->set(
+        i, StringView(taskInfo.taskId));
+    std::string temp = fmt::format("{}", task->id.stageExecutionId());
+    taskTableResult->childAt(2)->asFlatVector<StringView>()->set(
```
mbasmanova, line 162: This code is hard to read because it is not clear what column 2 or 5 means. Perhaps, create an enum.

presto-native-execution/presto_cpp/main/SystemConnector.cpp
mbasmanova: @aditi-pandit What do you think?

mbasmanova (commented):
Some comments.

presto-native-execution/presto_cpp/main/SystemConnector.cpp
aditi-pandit: RowTypePtr requires non-const column names and types list. So this cannot be changed.

presto-native-execution/presto_cpp/main/SystemConnector.cpp
aditi-pandit: RowTypePtr requires non-const column names and types list. So this cannot be changed.

presto-native-execution/presto_cpp/main/SystemConnector.cpp
aditi-pandit: I was being conservative but these can be removed.

presto-native-execution/presto_cpp/main/SystemConnector.cpp
aditi-pandit: Yeah removed these as well.

presto-native-execution/presto_cpp/main/SystemConnector.cpp
aditi-pandit: I've changed all of these to static now in a slightly different implementation.

presto-native-execution/presto_cpp/main/SystemConnector.cpp
aditi-pandit: Same for this as well.

presto-native-execution/presto_cpp/main/SystemConnector.cpp
aditi-pandit: Am tracking these separately in the PrestoTask now.

presto-native-execution/presto_cpp/main/SystemConnector.cpp
aditi-pandit: Have added an enum now.

presto-native-execution/presto_cpp/main/SystemConnector.cpp
aditi-pandit: That's a good idea. Have changed the code.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeGeneralQueries.java
aditi-pandit: Added comment.

presto-native-execution/presto_cpp/main/SystemConnector.cpp
```diff
@@ -0,0 +1,445 @@
...
+  VELOX_CHECK_NULL(
+      currentSplit_,
+      "Previous split has not been processed yet. Call next() to process the split.");
+  currentSplit_ = std::dynamic_pointer_cast<SystemSplit>(split);
+  VELOX_CHECK(currentSplit_, "Wrong type of split for SystemDataSource.");
+}
+
+SystemDataSource::TaskColumnEnum SystemDataSource::getTaskColumnEnum(
+    int32_t taskColumn) {
+  switch (taskColumn) {
```
mbasmanova, line 141: This shouldn't be necessary. See https://en.cppreference.com/w/cpp/language/enum

aditi-pandit: So going from enum -> int is generally type-safe. The other way (which I needed) is not and could error out in different compilers. It took me a while to play with the syntax here and I've found something that works. Hopefully this should work out.

aditi-pandit:
@mbasmanova, @majetideepak : Thanks for your previous input. This code is looking good for a full review now. Looking forward to your comments.

presto-native-execution/presto_cpp/presto_protocol/Connectors.cpp
aditi-pandit: So while the connector is named @system@system, the protocol objects have @system, system and $system@system usages of connector. I don't know the origin of this and its hard to trace from the code.

@tdcmeehan : Do you have any insights here ?

aditi-pandit:
> @mbasmanova, @tdcmeehan : Thanks for the discussion. It has been informative.
> 
> If we want to stay with this approach of getting tasks table on worker we could modify #21725 and #21285 to not perform those rewrites for system.runtime.tasks table specifically as it based on the worker.
> 
> #21725 could work un-modified as well. It would just mean that we don't allow partial agg over the tasks table which might not be a big deal unless a massive numbers of queries are scheduled in the cluster.
> 
> wdyt ?

tasks table gets data from all-nodes, so both the co-ordinator and workers. Since the co-ordinator generates data, both previous planner rules are also applicable.

presto-native-execution/presto_cpp/main/SystemConnector.h
```diff
@@ -0,0 +1,233 @@
...
+};
+
+class SystemTableHandle : public velox::connector::ConnectorTableHandle {
+ public:
+  explicit SystemTableHandle(
+      std::string connectorId,
+      std::string schemaName,
+      std::string tableName);
+
+  ~SystemTableHandle() override {}
```
arhimondr, line 44: Is this override necessary?

presto-native-execution/presto_cpp/main/SystemConnector.h
```diff
@@ -0,0 +1,233 @@
...
+  }
+
+ private:
+  const std::string name_;
+};
+
+class SystemTableHandle : public velox::connector::ConnectorTableHandle {
+ public:
+  explicit SystemTableHandle(
+      std::string connectorId,
```
arhimondr, line 40: Do we usually prefer to pass by value and then move? Or pass by const reference and copy? When do we prefer one over the other?

presto-native-execution/presto_cpp/main/SystemConnector.h
```diff
@@ -0,0 +1,233 @@
...
+    return completedRows_;
+  }
+
+  uint64_t getCompletedBytes() override {
+    return completedBytes_;
+  }
+
+  std::unordered_map<std::string, velox::RuntimeCounter> runtimeStats()
+      override {
+    // TODO: Which stats do we want to expose here?
```
arhimondr, line 98: nit: Unless somebody comes up with some useful statistics to expose we should probably drop the comment

presto-native-execution/presto_cpp/main/SystemConnector.h
```diff
@@ -0,0 +1,233 @@
...
+
+class SystemConnector : public velox::connector::Connector {
+ public:
+  SystemConnector(
+      const std::string& id,
+      std::shared_ptr<const velox::Config> /*properties*/,
+      folly::Executor* /*executor*/)
+      : Connector(id) {}
+
+  void setTaskManager(const TaskManager* taskManager) {
```
arhimondr, line 155: Having a setter makes it prone to a `null` pointer related bugs (if somebody creates a `SystemConnector` but forgets to set a TaskManager)

I wonder if we even need the `SystemConnectorFactory`? Since the connector is manually registered I wonder whether it makes sense to simply take the `TaskManager` in the constructor and create it without using a factory?

presto-native-execution/presto_cpp/main/SystemConnector.h
```diff
@@ -0,0 +1,233 @@
...
+  const TaskManager* taskManager_;
+};
+
+class SystemConnectorFactory : public velox::connector::ConnectorFactory {
+ public:
+  static constexpr const char* kSystemConnectorName{"$system"};
+
+  SystemConnectorFactory() : SystemConnectorFactory(kSystemConnectorName) {}
+
+  explicit SystemConnectorFactory(const char* connectorName)
```
arhimondr, line 196: Otherwise maybe we can pass the `TaskManager` to the `SystemConnectorFactory` and pass it to the `SystemConnector`?

presto-native-execution/presto_cpp/main/SystemConnector.cpp
```diff
@@ -0,0 +1,389 @@
...
+namespace facebook::presto {
+
+using namespace velox;
+
+namespace {
+
+static const std::string kRuntimeSchema = "runtime";
+static const std::string kTasksTable = "tasks";
+
+static std::vector<std::string> kTaskColumnNames = {
```
arhimondr, line 29: Should we make the `kTaskColumnNames` and `kTaskColumnTypes` to be local variables in `SystemTableHandle::taskSchema` to avoid potential use-after-move?

presto-native-execution/presto_cpp/main/SystemConnector.cpp
```diff
@@ -0,0 +1,389 @@
...
+void SystemDataSource::addSplit(
+    std::shared_ptr<connector::ConnectorSplit> split) {
+  VELOX_CHECK_NULL(
+      currentSplit_,
+      "Previous split has not been processed yet. Call next() to process the split.");
+  currentSplit_ = std::dynamic_pointer_cast<SystemSplit>(split);
+  VELOX_CHECK(currentSplit_, "Wrong type of split for SystemDataSource.");
+}
+
+#define SET_TASK_COLUMN(value)            \
```
arhimondr, line 139: Have you considered a template (+ lambda?)

arhimondr (commented):
Mostly style related comments / questions. Otherwise looks good.

presto-native-execution/presto_cpp/main/SystemConnector.h
aditi-pandit: Good catch. There is an underlying default destructor.

presto-native-execution/presto_cpp/main/SystemConnector.cpp
aditi-pandit: Good idea. Thanks !

presto-native-execution/presto_cpp/main/SystemConnector.h
aditi-pandit: @arhimondr : Good question. I prefer pass by const ref and copy to avoid use after move at the caller. But I've seen pass by value and move as a common pattern in Velox as well.

Lets make this pass by const ref and copy to be safe.

presto-native-execution/presto_cpp/main/SystemConnector.h
aditi-pandit: Good point. This code can be simplified and made more robust after removing SystemConnectorFactory. And there isn't a need for it really.

presto-native-execution/presto_cpp/main/SystemConnector.h
aditi-pandit: Not needed anymore.

presto-native-execution/presto_cpp/main/SystemConnector.cpp
```diff
@@ -0,0 +1,386 @@
...
+  std::vector<protocol::TaskInfo> taskInfos;
+  taskInfos.reserve(numRows);
+  for (const auto& taskEntry : taskMap) {
+    taskInfos.push_back(taskEntry.second->updateInfo());
+  }
+
+  auto result = std::dynamic_pointer_cast<RowVector>(
+      BaseVector::create(outputType_, numRows, pool_));
+
+  auto toMillis = [](int64_t nanos) -> int64_t { return nanos / 1'000'000; };
```
majetideepak, line 172: `constexpr int64_t kNanosecondsInMillisecond = 1'000'000;`

presto-native-execution/presto_cpp/main/SystemConnector.h
```diff
@@ -0,0 +1,206 @@
+/*
```
majetideepak, line 1: We can may be add a `connectors` folder under `presto_cpp`

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -939,6 +947,16 @@ std::vector<std::string> PrestoServer::registerConnectors(
   return catalogNames;
 }
 
+void PrestoServer::registerSystemConnector() {
+  PRESTO_STARTUP_LOG(INFO) << "Registering system catalog "
+                           << " using connector SystemConnector";
+  VELOX_CHECK(taskManager_);
+  auto systemConnector =
+      std::make_shared<SystemConnector>("$system@system", taskManager_.get());
```
majetideepak, line 955: Should we use `$system` to match the `presto_protocol.yml` change?

presto-native-execution/presto_cpp/presto_protocol/presto_protocol.yml
```diff
@@ -82,13 +84,15 @@ AbstractClasses:
...
         - { name: RemoteTransactionHandle,  key: $remote }
+        - { name: SystemTransactionHandle,  key: $system }
 
     ConnectorTableLayoutHandle:
       super: JsonEncodedSubclass
       subclasses:
         - { name: HiveTableLayoutHandle,    key: hive }
         - { name: IcebergTableLayoutHandle, key: hive-iceberg }
         - { name: TpchTableLayoutHandle,    key: tpch }
+        - { name: SystemTableLayoutHandle,  key: $system }
```
majetideepak, line 95: Why don't we need `system` and `$system@system` here?

majetideepak (commented):
@aditi-pandit some comments. Thanks!

presto-native-execution/presto_cpp/main/SystemConnector.cpp
```diff
@@ -0,0 +1,386 @@
...
+        break;
+      }
+
+      case TaskColumnEnum::kState: {
+        auto flat = result->childAt(i)->as<FlatVector<StringView>>();
+        SET_TASK_FMT_COLUMN(json(taskInfo.taskStatus.state).dump());
+        break;
+      }
+
+      case TaskColumnEnum::kSplits: {
```
aditi-pandit, line 214: @yzhang1991 : Splits/Drivers calculation.

presto-native-execution/presto_cpp/main/SystemConnector.h
aditi-pandit: @majetideepak : I had given this a thought. 

But since this code depends on TaskManager, that could introduce a circular dependency between main and connectors. So this connector is a good fit for the main folder. Other connectors in the future could be moved to a connectors folder.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -939,6 +947,16 @@ std::vector<std::string> PrestoServer::registerConnectors(
...
+  VELOX_CHECK(systemConnector);
```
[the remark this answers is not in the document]
aditi-pandit: This stayed over from the construction with a factory. Removed it.

presto-native-execution/presto_cpp/main/SystemConnector.cpp
aditi-pandit: @majetideepak : So my reason for not using the lambda was because it would be called per row per column basis. And I thought that would be a bit inefficient. 

Though I agree that the template/lambda use might reduce the code size.

presto-native-execution/presto_cpp/main/SystemConnector.h
```diff
@@ -0,0 +1,206 @@
...
+      const std::shared_ptr<velox::common::Filter>& /*filter*/) override {
+    VELOX_NYI("Dynamic filters not supported by SystemConnector.");
+  }
+
+  std::optional<velox::RowVectorPtr> next(
+      uint64_t size,
+      velox::ContinueFuture& future) override;
+
+  uint64_t getCompletedRows() override {
+    return completedRows_;
```
czentgr, line 87: Wondering why we have this and `completedBytes` as private members that are not used? They always return 0. Something for later?
Also the type of members is `size_t` but returned is `uint64_t` which is the same currently but why the discrepancy?

czentgr (commented):
This is nice and a great tutorial on how to implement basic connectors!

presto-native-execution/presto_cpp/presto_protocol/presto_protocol.yml
aditi-pandit: So we need the protocol json classes to be generated only once in this script. There isn't a need for all 3 catalog name mappings here. The mapping of the protocol to the key/catalog name happens in the PrestoToVeloxConnector code now. So that's where we have the 3 catalog name mappings.

presto-native-execution/presto_cpp/main/SystemConnector.h
aditi-pandit: @czentgr : Good catch. I updated both completedRows and completedBytes in SystemDataSource::getOutput() now. These are used to populate TableScan stats for Presto.

The size_t and uint64_t appears in many places for this code in different connectors. Maybe just something that has got over-looked.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -232,6 +233,12 @@ void PrestoServer::run() {
       std::make_unique<IcebergPrestoToVeloxConnector>("iceberg"));
   registerPrestoToVeloxConnector(
       std::make_unique<TpchPrestoToVeloxConnector>("tpch"));
+  registerPrestoToVeloxConnector(
+      std::make_unique<SystemPrestoToVeloxConnector>("$system"));
```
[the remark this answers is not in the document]
aditi-pandit: Done.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
aditi-pandit: I actually changed presto_protocol.yml to $system@system as that seemed to work out better.

aditi-pandit:
@majetideepak : Have addressed your review comments. Would appreciate another pass. Thanks !

presto-native-execution/presto_cpp/main/SystemConnector.cpp
```diff
@@ -0,0 +1,392 @@
...
+  for (const auto& taskEntry : taskMap) { \
+    auto task = taskEntry.second;         \
+    auto taskInfo = taskInfos[j];         \
+    temp = fmt::format("{}", value);      \
+    flat->set(j, StringView(temp));       \
+    j++;                                  \
+  }
+
+RowVectorPtr SystemDataSource::getTaskResults() {
+  TaskMap taskMap = taskManager_->tasks();
```
majetideepak, line 160: nit: auto

presto-native-execution/presto_cpp/main/SystemConnector.cpp
```diff
@@ -0,0 +1,392 @@
...
+      case TaskColumnEnum::kCreated: {
+        auto flat = result->childAt(i)->as<FlatVector<Timestamp>>();
+        SET_TASK_COLUMN(velox::Timestamp::fromMillis(task->createTimeMs));
+        break;
+      }
+
+      case TaskColumnEnum::kStart: {
+        auto flat = result->childAt(i)->as<FlatVector<Timestamp>>();
+        SET_TASK_COLUMN(
+            velox::Timestamp::fromMillis(task->firstSplitStartTimeMs));
```
majetideepak, line 310: Why is this not `taskInfo.stats.firstStartTime`?
Same above `taskInfo.stats.createTime`?

majetideepak (commented):
@aditi-pandit few comments. Thanks!

presto-native-execution/presto_cpp/main/SystemConnector.h
```diff
@@ -0,0 +1,206 @@
...
+
+  const std::string& schemaName() {
+    return schemaName_;
+  }
+
+  const std::string& tableName() {
+    return tableName_;
+  }
+
+  const velox::RowTypePtr& taskSchema();
```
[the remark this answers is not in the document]
aditi-pandit: Done

presto-native-execution/presto_cpp/main/SystemConnector.cpp
```diff
@@ -0,0 +1,392 @@
...
+    VELOX_CHECK(
+        it != columnHandles.end(),
+        "ColumnHandle is missing for output column '{}'",
+        outputName);
+
+    auto handle = std::dynamic_pointer_cast<SystemColumnHandle>(it->second);
+    VELOX_CHECK_NOT_NULL(
+        handle,
+        "ColumnHandle must be an instance of SystemColumnHandle "
+        "for '{}' on table '{}'",
```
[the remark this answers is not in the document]
aditi-pandit: Good catch.

presto-native-execution/presto_cpp/main/SystemConnector.cpp
aditi-pandit: taskInfo.stats.firstStartTime and createTime are serialized std::strings and we would need to re-parse them here. So instead we went with maintaining the original u_int64t ms values.

majetideepak (approved):
Thanks, @aditi-pandit

PR #21417, opened 2023-11-18
[native]Add server memory pushback configs

xiaoxmeng:
Add server memory pushback configs:

- system-mem-pushback-enabled is set to enable or disable server memory pushback
- system-mem-limit-gb specifies the server memory limit to trigger server memory pushback
- system-mem-shrink-gb specifies the memory to shrink when server memory pushback is
- triggered.

Server memory pushback is reclaim server memory usage to help get the server out of memory
condition. Since getting the server memory is platform dependent, the actual low memory 
condition detection can be implemented by deriving presto server's addAdditionalPeriodicTasks
to add a periodic running task to detect and trigger server memory pushback.
And we can reclaim memory by shrinking the async data cache: AsyncDataCache::shrink


amitkdutta (approved):
Thanks @xiaoxmeng

PR #21426, opened 2023-11-21
Refactor assertCreateTableAsSelect() and assertExplainAnalyze() code

pdabre12:
Refactored the assertCreateTableAsSelect() and assertExplainAnalyze() to reduce code duplication.

### DIFF
--- a/presto-accumulo/src/test/java/com/facebook/presto/accumulo/TestAccumuloDistributedQueries.java
+++ b/presto-accumulo/src/test/java/com/facebook/presto/accumulo/TestAccumuloDistributedQueries.java
@@ -85,19 +85,22 @@ public class TestAccumuloDistributedQueries
         this.assertCreateTableAsSelect(
                 "test_group",
                 "SELECT orderstatus, sum(totalprice) x FROM orders GROUP BY orderstatus",
-                "SELECT count(DISTINCT orderstatus) FROM orders");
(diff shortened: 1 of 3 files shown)


pdabre12:
@majetideepak  Please review.

majetideepak:
@pdabre12 Can you give a motivation for this refactor? Do we ever call `assertCreateTableAsSelect` with `allowDropTable` set to `false`?

pdabre12:
@majetideepak Yes, it will be used in #20936.

majetideepak:
> @majetideepak Yes, it will be used in #20936.

Why do we not want to delete the tables in that PR?

pdabre12:
There is a small discussion about this in PR #20936. 
Basically in order to drop tables, I need to set a system property "hive"  which would mean modifying the file https://github.com/prestodb/presto/blob/dd5b168ff5bdf306de70be19cc37c9797ca00981/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/NativeQueryRunnerUtils.java

majetideepak:
> There is a small discussion about this in PR #20936.

I commented there. Let's add the system property to the NativeQueryRunner in a separate PR.
I don't think we should allow not dropping the tables.

pdabre12:
@majetideepak 
Added the changes in #21641. 
Closing this PR as it's no longer needed.

PR #21427, opened 2023-11-21
[native] Fix promise_ race in PrestoExchangeSource.

spershin:
We've noticed periodic crashes in PrestoExchangeSource:: abortResults() and/or
PrestoExchangeSource:: close(). Stack was a bit different at times.

Investigating it for a while we saw that this happens when query fails in an operator and
fails fairly early ( the 1st 10-20 seconds of execution).
After a while we managed to get an ASAN crash on it and found out that std::move(promise_)
triggered ASAN on moving the context_ string in the promise_.
That was a good indication that there is a race of promise_ being moved out or moved in.
Also explained weird crashes we could not track down using non-ASAN ways, since the move
was corrupting the memory, rather than pointer being null or memory being freed.

Looking at the code we saw that the only call that accesses promise_ (mutates it) unprotected
was in PrestoExchangeSource::request().

This fix it to protect promise_ mutation with the queue_'s mutex like in other code sites.

After the fix the query that was close to 100% triggering the crash didn't cause crash
anymore (36 attempts).
In addition, no zombie tasks were left in the system either.

Saying that, we might want to revisit the call sequence that results in PrestoExchangeSource::request()
call, since it looks like it is being done in parallel with close() call and potentially can leave some futures
unfulfilled. I haven't noticed this being the case with this fix, however.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/PrestoExchangeSource.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoExchangeSource.cpp
@@ -139,14 +139,17 @@ folly::SemiFuture<PrestoExchangeSource::Response> PrestoExchangeSource::request(
   // Before calling 'request', the caller should have called
   // 'shouldRequestLocked' and received 'true' response. Hence, we expect
   // requestPending_ == true, atEnd_ == false.
-  // This call cannot be made concurrently from multiple threads.
   VELOX_CHECK(requestPending_);
-  VELOX_CHECK(!promise_.valid() || promise_.isFulfilled());
-
+  // This call cannot be made concurrently from multiple threads, but other
+  // calls that mutate promise_ can be called concurrently.
   auto promise = VeloxPromise<Response>("PrestoExchangeSource::request");
   auto future = promise.getSemiFuture();
+  {
+    std::lock_guard<std::mutex> l(queue_->mutex());
+    VELOX_CHECK(!promise_.valid() || promise_.isFulfilled());
+    promise_ = std::move(promise);
+  }
 
-  promise_ = std::move(promise);
   failedAttempts_ = 0;
   dataRequestRetryState_ =
       RetryState(std::chrono::duration_cast<std::chrono::milliseconds>(


presto-native-execution/presto_cpp/main/PrestoExchangeSource.cpp
```diff
@@ -139,14 +139,19 @@ folly::SemiFuture<PrestoExchangeSource::Response> PrestoExchangeSource::request(
...
+  // calls that mutate members like promise_ can be called concurrently.
+  auto future{folly::SemiFuture<Response>::makeEmpty()};
+  {
+    std::lock_guard<std::mutex> l(queue_->mutex());
+    VELOX_CHECK(requestPending_);
+    VELOX_CHECK(!promise_.valid() || promise_.isFulfilled());
 
-  auto promise = VeloxPromise<Response>("PrestoExchangeSource::request");
-  auto future = promise.getSemiFuture();
+    auto promise = VeloxPromise<Response>("PrestoExchangeSource::request");
```
xiaoxmeng, line 150: You can create future and promise outside the lock and move to promise_ inside the lock? Is it hard to create a unit test to trigger this race condition? Thanks!

spershin: True, we can create promise outside the lock. I've move.

I don't know how to create that race. :)
It is probably possible with a lot of thinking and code reading...

spershin: @xiaoxmeng 
Actually...
If we do this:
```
  VELOX_CHECK(requestPending_);
  auto promise = VeloxPromise<Response>("PrestoExchangeSource::request");
  auto future = promise.getSemiFuture();
  {
    std::lock_guard<std::mutex> l(queue_->mutex());
    VELOX_CHECK(!promise_.valid() || promise_.isFulfilled());
    promise_ = std::move(promise);
  }
```
We might have unfulfilled promise destruction (if VELOX_CHECK() under the lock throws).
Do we care much about creating promise outside of the lock?

xiaoxmeng: The unfilled promise won't get destroyed under the lock? We first check promise_ if fails, then we throw. The unfilled promise is destroyed out side the lock? thanks!

xiaoxmeng (approved):

xiaoxmeng (approved):

PR #21433, opened 2023-11-21
[native]Add hook to stop additional periodic task

xiaoxmeng:


gggrace14 (approved):
I'm okay with it functionally. However why not remove the code for now and revert the change once we need it again?

xiaoxmeng:
> I'm okay with it functionally. However why not remove the code for now and revert the change once we need it again?

We will add back stats things to ease the migration. Thanks!

PR #21436, opened 2023-11-22
Quick stats

aaneja:
It is possible to derive stats for a Hive table or partition by utilizing metadata from table formats or other such sources. This commit adds a QuickStatsBuilder class that can use these sources to build stats when no stats are available from HMS

## Description
Stats built are cached for future use (caching duration configurable)
The time to build stats can be kept bounded using a config property too

A ParquetQuickStatsBuilder is also added, which utilizes stats stored in parquet file footers

Other additions
- JMX metrics to measure hit rate for the quick stats provider
- RuntimeStats to measure planning time impact of building
these stats

See https://prestodb.io/blog/2023/07/20/quick-stats-presto-blog/ for more details
Slides [here](https://docs.google.com/presentation/d/1yHd0GMPbSqcuJa2bOoVIYrEDfmAkAo3vtTfGlGocwQM/edit?usp=sharing), Presto day recording [here](https://www.youtube.com/watch?v=NotgeUG8V4E)

## Motivation and Context
Missing stats cause the planner to fly blind. Any stats, even if they are incomplete can help the planner build better plans

## Impact
Newly online Hive tables/partitions do not need to run ANALYZE (which can be very costly to run over large datasets) before being made available to Presto

## Test Plan
- Unit tests added
- E2E test added in TestParquetDistributedQueries
- Tested against TPCH queries manually and [saw improvements](https://docs.google.com/presentation/d/1yHd0GMPbSqcuJa2bOoVIYrEDfmAkAo3vtTfGlGocwQM/edit?usp=sharing)

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

Hive Changes
* Added quick stats, a mechanism to build stats from metadata for tables/partitions that are missing stats
```


presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java
```diff
@@ -0,0 +1,308 @@
...
+            inProgressReaperExecutor.schedule(() -> {
+                inProgressBuilds.remove(partitionKey);
+                partitionStatisticsCompletableFuture.get().cancel(true);
+            }, reaperExpiryMillis, MILLISECONDS);
+            return Instant.now();
+        });
+
+        CompletableFuture<PartitionStatistics> future = partitionStatisticsCompletableFuture.get();
+        // If a background call to build quick stats was started, wait for it to complete
+        // This way, the first query that initiated the quick stats call for this partition will wait for the stats to be built (unless it times out)
```
ZacBlanco, line 225: Curious if there was any other discussion about this behavior? Why not let all queries looking for this data wait for stats? Or just let all queries with in-progress jobs just return empty stats? Is there a particular difference between the query that initiated vs later ones that makes this behavior is desirable?

I feel like preventing some queries from completing, but not others would look odd from the users perspective and/or more difficult to debug. If you let all queries through regardless of in-progress status, you might not benefit from the stats immediately, but at least you're not waiting the whole time for a large table, especially if the query would be quick otherwise (e.g. scanning just a tiny partition?)

OTOH, blocking all queries due to in-progress stats might lead to undesirable performance behavior, but it might be more understandable for the user that all queries are stuck in planning. It may be easier to draw the conclusion that stats quickstats collection is the culprit.

I don't have a strong opinion of which approach to take. Just wanted some discussion.

presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java
```diff
@@ -0,0 +1,308 @@
...
+        try {
+            fs = hdfsEnvironment.getFileSystem(hdfsContext, path);
+        }
+        catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+
+        // Resolve the list of files using the directory lister.
+        // If the directoryLister is a CachingDirectoryLister, this listing call will get/put the call from cache
+        ImmutableList<HiveFileInfo> fileList = ImmutableList.copyOf(directoryLister.list(fs, resolvedTable, path, partition, nameNodeStats, hiveDirectoryContext));
```
ZacBlanco, line 286: Do we ever need all the files at once? It might be better to use the iterator API instead to collect the stats rather than aggregate the entire file list here. For large partitions you might even be able to start collecting stats before finishing listing which could improve the wall time due to pipelining.

presto-hive/src/main/java/com/facebook/presto/hive/HiveSessionProperties.java
```diff
@@ -711,6 +713,11 @@ public HiveSessionProperties(HiveClientConfig hiveClientConfig, OrcFileWriterCon
                         PARALLEL_PARSING_OF_PARTITION_VALUES_ENABLED,
                         "Enables parallel parsing of partition values from partition names using thread pool",
                         hiveClientConfig.isParallelParsingOfPartitionValuesEnabled(),
+                        false),
+                booleanProperty(
+                        QUICK_STATS_ENABLED,
+                        "Use quick stats to resolve stats",
```
ZacBlanco, line 719: nit: I would make this more descriptive 

```suggestion
                        "Resolve some table statistics by reading parquet footers",
```

presto-hive/src/test/java/com/facebook/presto/hive/statistics/TestParquetQuickStatsBuilder.java
```diff
@@ -0,0 +1,317 @@
...
+    public static final String TEST_TABLE = "quick_stats";
+    private ParquetQuickStatsBuilder parquetQuickStatsBuilder;
+    private MetastoreContext metastoreContext;
+    private SemiTransactionalHiveMetastore metastoreMock;
+    private HdfsEnvironment hdfsEnvironment;
+    private HiveClientConfig hiveClientConfig;
+
+    public static void main(String[] args)
+            throws Exception
+    {
```
ZacBlanco, line 86: if we intend to keep, I would make it configurable with storage location?

ZacBlanco (changes requested):
Took a first pass over the code sans tests

presto-hive/src/main/java/com/facebook/presto/hive/statistics/ParquetQuickStatsBuilder.java
```diff
@@ -0,0 +1,346 @@
...
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+
+public class ParquetQuickStatsBuilder
+        implements QuickStatsBuilderInterface
+{
+    public static final Logger log = Logger.get(ParquetQuickStatsBuilder.class);
+    private static final ExecutorService footerFetchExecutor = Executors.newCachedThreadPool();
+
+    private static final Set<String> PARQUET_SERDE_CLASS_NAMES = ImmutableSet.<String>builder()
+            .add("org.apache.hadoop.hive.ql.io.parquet.serde.ParquetHiveSerDe")
```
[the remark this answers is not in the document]
aaneja: Was following the pattern from https://github.com/prestodb/presto/blob/aaf917d7f351a3dbadd8f00a0240979a1a5745b7/presto-hive/src/main/java/com/facebook/presto/hive/parquet/ParquetPageSourceFactory.java#L153-L156

IMO hardcoding is the better way since class namespaces can get refactored, but existing parquet files will have the serde name still referring to the old one

presto-hive/src/main/java/com/facebook/presto/hive/statistics/ColumnQuickStats.java
```diff
@@ -0,0 +1,125 @@
...
+    {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        ColumnQuickStats<?> that = (ColumnQuickStats<?>) o;
+        return rowCount == that.rowCount && nullsCount == that.nullsCount &&
+                java.util.Objects.equals(columnName, that.columnName) &&
```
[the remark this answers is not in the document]
aaneja: Yes since a static import for `equals` will conflict with the current `equals` method :) Will add a static import for `java.util.Objects`

presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java
```diff
@@ -0,0 +1,308 @@
...
+        return result.build();
+    }
+
+    public PartitionStatistics getQuickStats(ConnectorSession session, SemiTransactionalHiveMetastore metastore, SchemaTableName table,
+            MetastoreContext metastoreContext, String partitionId)
+    {
+        if (!isQuickStatsEnabled(session)) {
+            return empty();
+        }
+        requestCount.incrementAndGet(); // New request was made to resolve quick stats for a partition
```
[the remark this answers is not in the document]
aaneja: I'm not returning `Cache#stats()` hit/miss counts because for this use case it is not painting the full picture. We also need to track the calls that could be answered by the inline call to the quick stats provider.

presto-hive/src/main/java/com/facebook/presto/hive/statistics/ParquetQuickStatsBuilder.java
```diff
@@ -0,0 +1,346 @@
...
+                    throw new RuntimeException(e);
+                }
+            }, footerFetchExecutor);
+        }
+
+        HashMap<ColumnPath, ColumnQuickStats<?>> rolledUpColStats = new HashMap<>();
+        try {
+            // Wait for footer reads to finish
+            CompletableFuture<Void> overallCompletableFuture = CompletableFuture.allOf(footerFetchCompletableFutures);
+            overallCompletableFuture.get(footerFetchTimeoutMillis, MILLISECONDS);
```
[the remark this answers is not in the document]
aaneja: Not really. The CPU/latency impact of doing this single threaded is miniscule, even for a large number of files (the parquet metadata per file is already serialized, there is no waiting time once the futures have finished)

presto-hive/src/main/java/com/facebook/presto/hive/HiveSessionProperties.java
aaneja: This is not accurate. Parquet quick stats is one of the currently implemented ways, future implementations are an option for Hudi, Delta, Iceberg, ORC etc

presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsBuilderInterface.java
```diff
@@ -0,0 +1,30 @@
...
+import com.facebook.presto.hive.HiveFileInfo;
+import com.facebook.presto.hive.metastore.MetastoreContext;
+import com.facebook.presto.hive.metastore.SemiTransactionalHiveMetastore;
+import com.facebook.presto.spi.ConnectorSession;
+import com.facebook.presto.spi.SchemaTableName;
+
+import java.util.List;
+
+@FunctionalInterface
+public interface QuickStatsBuilderInterface
```
tdcmeehan, line 26:
```suggestion
public interface QuickStatsBuilder
```

presto-hive/src/main/java/com/facebook/presto/hive/statistics/ParquetQuickStatsBuilder.java
```diff
@@ -0,0 +1,346 @@
...
+import static com.facebook.presto.parquet.cache.MetadataReader.readFooter;
+import static java.util.Objects.requireNonNull;
+import static java.util.concurrent.CompletableFuture.supplyAsync;
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+
+public class ParquetQuickStatsBuilder
+        implements QuickStatsBuilderInterface
+{
+    public static final Logger log = Logger.get(ParquetQuickStatsBuilder.class);
+    private static final ExecutorService footerFetchExecutor = Executors.newCachedThreadPool();
```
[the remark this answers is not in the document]
ZacBlanco: I understand wanting to control size but the `newCachedThreadPool()` with no parameters basically grows unbounded as you throw work at it to prevent queuing. You might get more concurrency due to many IO operations but you can run into high context switch overheads or too many threads if ulimits are not configured high enough. I would at least use a thread pool where you can cap the threads.

I was going to suggest just using the common FJ pool. The common pool is capped at 2x core count which results in queuing but limits overheads. Since it's shared though you may run into queuing issues. Keeping separate executors is OK, but I would at least looking at capping the thread count

presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java
ZacBlanco: I understand wanting to avoid blocking for the batch use case. However, for querying data without stats loaded, will having the stats loaded actually decrease E2E query times? It's likely to be query dependent - so I'm sure for some queries it might. I'm concerned then in the common case whether it's worth blocking at all.

> A user could also use this mechanism for pre-building quick stats for newly online partitions; simply do a SHOW STATS FOR (<partition query>) before bringing the partition online, and this way all new queries will benefit from the pre-cached stats

This feels like something that should be achieved via an explicit procedure call IMO, but can understand how just dumping the query into `SHOW STATS` can be easier to load the desired partition(s).

presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java
ZacBlanco: > 1/ Most implementations will use a caching directory lister, so the Iterator is just a list wrapper over a pre-cached file list

Given the assumption from the previous discussion that the quick stats is for new or old data, how likely is the result to be in the cache in the first place?

I don't know how common it is to have partitions with 1000s of files which is likely where and iterator optimization would be of benefit. I can see it being useful if someone has poorly optimized table though with little downside to well optimized ones

presto-hive/src/main/java/com/facebook/presto/hive/HiveSessionProperties.java
ZacBlanco: I see. Maybe a section in the documentation would be helpful?

presto-hive/src/main/java/com/facebook/presto/hive/statistics/ParquetQuickStatsBuilder.java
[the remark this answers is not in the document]
ZacBlanco: Is there ever a case where this map won't be exactly the same as the one in `ParquetPageSourceFactory`? Can we just expose that set instead of creating a copy?

presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java
aaneja: >for online querying data without stats loaded, will having the stats loaded actually decrease E2E query times

The latency impact of not having stats resulting in a poor plan will dominate the E2E query times, so yes having quick stats loaded will help reduce this.
If this user is concerned about having their query wait for quick stats building, they can always `quick-stats.build-timeout` set to a low enough value acceptable to them (`1ms` e.g)

presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java
aaneja: My bad - I was mis-rembering the cache listing order. Quick stats would be the first to initiate directory listing, split loading that happens after planning will benefit from this
Updating the interace to user a vanilla Iterator :+1:

presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java
ZacBlanco: I know that stats are helpful for queries for queries which execute on an especially large dataset, but if many queries run in seconds to ms without stats, then they won't really be beneficial. At the end of the day it really depends on the workload and datasets. I personally don't have a grasp on all the workloads that people are running with Presto. What I do know is that Presto's design is catered towards fast-running analytical queries over long-running ones where stats might be more critical, hence you want the overhead of stat collection to be small.

As long as it's configurable on the session side as to how long to wait for stats, I suppose the current solution is okay

aaneja: The argument of quick vs long running queries having low/high impact from stats is bit myopic since its focused on a single query. Consider the beneficial impact of having stats available to subsequent queries and lowered cluster utilization as a result. 
To allay your concerns, I will add a session control for the build-timeout (right now it is a startup property), including setting it to `0`. This way users can choose per query the behavior they want.

steveburnett: > I will add a session control property for the build-timeout (right now it is a startup property), including setting it to `0`. This way users can choose per query the behavior they want.

Please document the new session property, when you can.

presto-hive/src/test/java/com/facebook/presto/hive/statistics/TestQuickStatsProvider.java
```diff
@@ -0,0 +1,341 @@
...
+    public void setUp()
+    {
+        metastoreMock = mock(SemiTransactionalHiveMetastore.class);
+        metastoreContext = new MetastoreContext(SESSION.getUser(),
+                SESSION.getQueryId(),
+                Optional.empty(),
+                Optional.empty(),
+                Optional.empty(),
+                false,
+                HiveColumnConverterProvider.DEFAULT_COLUMN_CONVERTER_PROVIDER);
```
ZacBlanco, line 116: nit: static import

presto-hive/src/test/java/com/facebook/presto/hive/statistics/TestQuickStatsProvider.java
```diff
@@ -0,0 +1,341 @@
...
+                        Optional.empty(),
+                        Optional.empty())));
+
+        directoryListerMock = mock(DirectoryLister.class);
+        when(directoryListerMock.list(any(), any(), any(), any(), any(), any())).thenReturn(emptyIterator());
+
+        MetastoreClientConfig metastoreClientConfig = new MetastoreClientConfig();
+        hdfsEnvironment = createTestHdfsEnvironment(hiveClientConfig, metastoreClientConfig);
+
+        ColumnQuickStats<Integer> mockIntegerColumnStats = new ColumnQuickStats<>("column", Integer.class);
```
ZacBlanco, line 161: maybe re-use methods from `TestParquetQuickStatsBuilder`?

presto-hive/src/test/java/com/facebook/presto/hive/statistics/TestQuickStatsProvider.java
```diff
@@ -0,0 +1,341 @@
...
+
+    @Test(invocationCount = 3)
+    public void testConcurrentFetchForSamePartition()
+            throws ExecutionException, InterruptedException
+    {
+        QuickStatsBuilder longRunningQuickStatsBuilderMock = mock(QuickStatsBuilder.class);
+        when(longRunningQuickStatsBuilderMock.buildQuickStats(any(), any(), any(), any(), any(), any()))
+                .thenAnswer((Answer<PartitionQuickStats>) invocationOnMock -> {
+                    // Sleep for 1.5s to simulate a long-running quick stats call
+                    Thread.sleep(1500);
```
ZacBlanco, line 221: Similar vein to other comment about CI time. I would be prudent about forcing tests to wait for any amount of time. Would prefer to reduce timeouts or find a way to better simulate long-running calls. Don't have to make any change but would be nice to make this as small as possible without jeopardizing the stability of the test.

presto-hive/src/test/java/com/facebook/presto/hive/statistics/TestQuickStatsProvider.java
```diff
@@ -0,0 +1,341 @@
...
+        assertEquals(quickStats.entrySet().size(), testPartitionsMix.size());
+        assertTrue(quickStats.keySet().containsAll(testPartitionsMix));
+        quickStats.values().forEach(ps -> assertEquals(ps, expectedPartitionStats));
+        for (String testPartition : testPartitions2) {
+            verify(quickStatsBuilderMock).buildQuickStats(any(), any(), any(), any(), eq(testPartition), any());
+        }
+        verifyNoMoreInteractions(quickStatsBuilderMock);
+    }
+
+    @Test(invocationCount = 3)
```
ZacBlanco, line 213: Why 3 invocations?

ZacBlanco (commented):
Took a pass on tests. LGTM for the most part. Mainly nits

presto-hive/src/test/java/com/facebook/presto/hive/statistics/TestParquetQuickStatsBuilder.java
```diff
@@ -0,0 +1,318 @@
...
+        else {
+            System.out.println("Warmup..");
+        }
+    }
+
+    @Test
+    public void testStatsBuildTimeIsBoundedUsingFooterFetchTimeout()
+            throws Exception
+    {
+        HiveClientConfig customHiveClientConfig = new HiveClientConfig().setParquetQuickStatsFooterFetchTimeout(new Duration(100, TimeUnit.MILLISECONDS));
```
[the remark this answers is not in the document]
aaneja: ok reducing to 10ms

presto-hive/src/test/java/com/facebook/presto/hive/statistics/TestQuickStatsProvider.java
aaneja: reducing to 50ms

presto-hive/src/test/java/com/facebook/presto/hive/statistics/TestQuickStatsProvider.java
aaneja: To remove any possible noise from thread timing - `future2` may be the one to schedule the in-progress build and that shouldn't break the test

presto-hive/src/test/java/com/facebook/presto/hive/statistics/TestQuickStatsProvider.java
aaneja: Didn't get this - what methods are you referring to ?

ZacBlanco: I meant these: https://github.com/prestodb/presto/pull/21436/files#diff-5b27abaf2f3eda821b70c980b26ea45db3bf6b495b626362ac8d350b93ff3c93R131

Am fine leaving this as is, but thought it might just be less LoC to use it rather than set each explicitly

presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java
```diff
@@ -0,0 +1,318 @@
...
+    private final DirectoryLister directoryLister;
+    private final List<QuickStatsBuilder> statsBuilderStrategies;
+    private final boolean recursiveDirWalkerEnabled;
+    private final ConcurrentHashMap<String, Instant> inProgressBuilds = new ConcurrentHashMap<>();
+    private final AtomicLong requestCount = new AtomicLong(0L);
+    private final AtomicLong succesfulResolveFromCacheCount = new AtomicLong(0L);
+    private final AtomicLong succesfulResolveFromProviderCount = new AtomicLong(0L);
+    private final Duration cacheExpiryDuration;
+    private final long reaperExpiryMillis;
+    private final Cache<String, PartitionStatistics> partitionToStatsCache;
```
ClarenceThreepwood, line 93: If you do something like

set quickStatsEnabled=true
query1
set quickStatsEnabled=false
query1

The stats computed for the first query are not used for the second even though available. Is this the expected behavior? Now if the second query were a query that references a table used in the first, and the flag were disabled by oversight, then it is a shame that the already computed stats were not used.

Related question - Did you consider making partitionToStatsCache part of CachingHiveMetastore? The advantage would be that QuickStats, if computed, would be available to all queries and comes with its own set of well-defined parameters and accounting.

presto-hive/src/main/java/com/facebook/presto/hive/statistics/ParquetQuickStatsBuilder.java
```diff
@@ -0,0 +1,344 @@
...
+        requireNonNull(table);
+        requireNonNull(metastoreContext);
+        requireNonNull(partitionId);
+        requireNonNull(files);
+
+        if (!files.hasNext()) {
+            return PartitionQuickStats.EMPTY;
+        }
+
+        StorageFormat storageFormat;
```
ClarenceThreepwood, line 251: This logic in determining storageFormat would have to be replicated for each <file format>QuickStatsBuilder. Would it make sense to move it into QuickStatsBuilder and then pick the right strategy based on the storageFormat there?

presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java
```diff
@@ -0,0 +1,318 @@
...
+        });
+
+        CompletableFuture<PartitionStatistics> future = partitionStatisticsCompletableFuture.get();
+        long quickStatsBuildTimeoutMillis = getQuickStatsBuildTimeoutMillis(session);
+
+        // A background call to build quick stats was started, and we want to wait for quick stats to be built
+        // Note : Only the first query that initiated the quick stats call for this partition will have to wait for the stats to be built
+        // Other queries will either observe an in-progress build, and get an empty quick stats returned from below
+        // Or would've already gotten the cached quick stats value
+        if (future != null && quickStatsBuildTimeoutMillis > 0) {
```
ClarenceThreepwood, line 232: Do we know how this behaves in case of a self-join? I hope we do not wind up split brained and use stats for partitions in one part of the query and not in another. Maybe another good test case

presto-hive/src/test/java/com/facebook/presto/hive/statistics/TestQuickStatsProvider.java
```diff
@@ -0,0 +1,342 @@
...
+        // Verify only one call was made for each test partition
+        assertEquals(quickStats.entrySet().size(), testPartitions.size());
+        assertTrue(quickStats.keySet().containsAll(testPartitions));
+        quickStats.values().forEach(ps -> assertEquals(ps, expectedPartitionStats));
+
+        verifyNoMoreInteractions(longRunningQuickStatsBuilderMock);
+    }
+
+    @Test
+    public void quickStatsBuildTimeIsBounded()
```
ClarenceThreepwood, line 285: Do we need a test for cache expiry as well?

presto-hive/src/main/java/com/facebook/presto/hive/statistics/ParquetQuickStatsBuilder.java
```diff
@@ -0,0 +1,344 @@
...
+    private final long footerFetchTimeoutMillis;
+
+    public ParquetQuickStatsBuilder(FileFormatDataSourceStats stats, HdfsEnvironment hdfsEnvironment, HiveClientConfig hiveClientConfig)
+    {
+        this.stats = stats;
+        this.hdfsEnvironment = hdfsEnvironment;
+        this.footerFetchTimeoutMillis = hiveClientConfig.getParquetQuickStatsFooterFetchTimeout().roundTo(MILLISECONDS);
+    }
+
+    private static void processColumnMetadata(ParquetMetadata parquetMetadata, Map<ColumnPath, ColumnQuickStats<?>> rolledUpColStats)
```
ClarenceThreepwood, line 95: nit: Move private methods to after the public ones

ClarenceThreepwood (commented):
Nice work Anant! 
My feedback consists mostly questions for my edification and suggestions exploring alternate choices

presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java
aaneja: The `quick_stats_enabled` flag is meant to provide the user a switch to use (or compute first and then use) stats outside of HMS. If the flag is`false`, I think users would be surprised if they see stats being inferred for partitions they know are missing stats. The behavior is similar to how `use_history_based_plan_statistics` controls if HBO computed plan nodes stats are used or not

I did consider having these stats persisted to HMS (say a background thread that atomically updates the metastore if stats for the partition are missing), but decided to punt on this for now

presto-hive/src/main/java/com/facebook/presto/hive/statistics/ParquetQuickStatsBuilder.java
aaneja: Yes, leaving it for a future refactor. Will add a note for this.

presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java
aaneja: Since the planner code is single threaded for an individual query, stats are resolved one PlanNode at a time. So the first call to resolve quick stats will come from the first instance of the TableScan for this table, this will resolve and cache stats. The next call from the subsequent TableScan node will pick up these cached stats.

I could confirm this with a query : `explain (format json) select 1 from region_no_stats r1, region_no_stats r2 where r1.r_regionkey = r2.r_regionkey;` , [the results](https://gist.github.com/aaneja/680aea3681ef7d592f0e20b8dbb92776#file-quickstats_selfjoin-json) show that the stats for each TableScan node were correctly resolved to -
```
 "r_regionkey<integer>" : {                                                                                                                                                                                                                                                          
                 "lowValue" : 0.0,                                                                                                                                                                                                                                                                 
                 "highValue" : 4.0,                                                                                                                                                                                                                                                                
                 "nullsFraction" : 0.0,                                                                                                                                                                                                                                                            
                 "averageRowSize" : "NaN",                                                                                                                                                                                                                                                         
                 "distinctValuesCount" : "NaN"                                                                                                                                                                                                                                                     
               } 
```

Now if the first call came in and the quick stats resolution times out, this node will see `EMPTY` stats. The next call could, in theory, see the cached stats if the quick stats resolution finishes in the background.

While this is not ideal, I don't see this as incorrect behavior - the same behavior would occur with HMS based stats too; a background HMS update whilst a query is in progress could in theory result in the same situation

presto-hive/src/test/java/com/facebook/presto/hive/statistics/TestQuickStatsProvider.java
aaneja: Since we're using a Guava Cache with well tested expiry behavior, I chose to skip this test

presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java
```diff
@@ -0,0 +1,318 @@
...
+                catch (InterruptedException | ExecutionException e) {
+                    // This should not happen because we checked that the future was completed successfully
+                    log.error(e, "Failed to get value for a quick stats future which was completed successfully");
+                    throw new RuntimeException(e);
+                }
+            }
+            else {
+                // If a future did not finish, or finished exceptionally, we do not add it to the results
+                // A new query for the same partition could trigger a successful quick stats fetch for this partition
+                result.put(partitionId, empty());
```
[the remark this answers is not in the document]
aaneja: This would be identical to current behavior where a query comes in for say 10 partitions, and only 5 of them have HMS provided stats. See [MetastoreHiveStatisticsProvider#getTableStatistics](https://github.com/prestodb/presto/blob/496e98ca25fb243ab25731d188c2fd56eba3b3e9/presto-hive/src/main/java/com/facebook/presto/hive/statistics/MetastoreHiveStatisticsProvider.java#L389) for how the rollup happens. Briefly, it aggregates the available stats and comes up with an underestimate

presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java
```diff
@@ -0,0 +1,318 @@
...
+        String partitionKey = String.join("/", table.toSchemaTablePrefix().toString(), partitionId);
+
+        PartitionStatistics cachedValue = partitionToStatsCache.getIfPresent(partitionKey);
+        if (cachedValue != null) {
+            succesfulResolveFromCacheCount.incrementAndGet();
+            return cachedValue;
+        }
+
+        // Check if we already have a quick stats build in progress for this partition key
+        if (inProgressBuilds.containsKey(partitionKey)) {
```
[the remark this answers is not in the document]
aaneja: The intent of waiting for a background quick stats fetch is that the initiator query can get the benefit of quick stats in the first go. This is useful for interactive queries and would be what users expect if they set enable quick stats.  
However if there are multiple queries for the same partition (think multiple queries from a dashboard that is loading recently uploaded data), and we block all of them if an in-progress build is occurring, we would introduce queuing in the system which we should avoid.

Related discussion on this - https://github.com/prestodb/presto/pull/21436#discussion_r1449338815

presto-hive/src/main/java/com/facebook/presto/hive/statistics/ParquetQuickStatsBuilder.java
aaneja: Can you share your codestyle rules for IntelliJ ?
Mine have constructor above static methods
image

presto-hive/src/main/java/com/facebook/presto/hive/statistics/ParquetQuickStatsBuilder.java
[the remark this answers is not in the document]
aaneja: Okay, any suggestions on sane defaults for the max threads for the bounded executors in ParquetQuickStatsBuilder and QuickStatsProvider ?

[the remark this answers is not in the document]
ZacBlanco: I think there should be a cap, but unsure what would be reasonable. Just spitting out my thoughts here

A single executor runnable can't be suspended when an IO request is made, thus if you have, say 400 outstanding requests and an executor with 2 threads, you'll only ever get a maximum concurrency of 2 requests. In reality, a properly constructed IO system should be able to submit all 400 requests using the 2 threads concurrently and return the results for each once the IO is finished, without ever blocking a thread. Unfortunately, not something we have available in Java using the `FileSystem` APIs that make the IO requests.

Since most of the requests are IO-based and not CPU-based, increasing the executor size past the thread count of the machine **is** likely to have tangible benefit from my understanding. The only worry would be memory consumption of the thread's stack space in the case of hundreds (or potentially thousands) or threads. My guess is that the cap should probably be some function of the JVM memory, not necessarily the available cores on the machine if you assume the work these threads do is IO-bound, not CPU bound.

I believe the default stack size is 1M, meaning 1024 threads would consume _up to_ 1GiB of memory. Maybe scaling the thread count to something like consuming up to 10/15% of max memory by default might be reasonable?  e.g. if the coordinator has 10G heap, then you would allow 10% of that memory to be consumed by quick stats threads IF they reach maximum stack size. Most of the time this would be much smaller as thread stack minimum size is something like 16KiB.

Example for ParquetQuickStatsBuilder

```java
private static final ExecutorService footerFetchExecutor = new ThreadPoolExecutor(1, (int) (Runtime.getRuntime().maxMemory() * 0.1 / (1024 * 1024)), 60000, MILLISECONDS, new LinkedBlockingQueue<>());
```

You could also check the JVM arguments to see if the `-Xss` parameter is set and divide by that size argument rather than the default 1BM (`(1024 * 1024)`) in the example above

presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java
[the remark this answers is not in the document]
ClarenceThreepwood: You say "I will add a session control property for the build-timeout (right now it is a startup property), including setting it to 0. This way users can choose per query the behavior they want"

But this does not change the per query behavior if timeout > 0 correct? The subsequent queries still will not wait?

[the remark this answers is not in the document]
aaneja: The build timeout only comes to play for the query that initiates the quick stats build. Any subsequent queries, if they observe an already initiated quick stats call will NOT wait, and assume EMPTY quick stats

presto-hive/src/main/java/com/facebook/presto/hive/statistics/ParquetQuickStatsBuilder.java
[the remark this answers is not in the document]
aaneja: I'm aware of the thread semantics; you're right, for IO bound tasks using the JVM threadpool is not the right idea ([virtual threads](https://openjdk.org/jeps/425) will help here) given that most FS implementations don't use non-blocking IO. However, I'm not convinced making the dynamic calculation is the right choice given that 1/We haven't done something similar in other bg threadpools 2/The connectors can (and do) use their own bg threadpools -this would our 10-15% memory usage assumption incorrect

I would rather use a (configurable) constant value for the max sizes. I think we should keep them conservative to start off
1. `100` for the QuickStatsBuilder - This would limit the max number of partitons scanned at any time to 100
2.  `500` for the ParquetQuickStats builder - Assuming 5 files per partition.

Of course these wont limit the actual partition/files that a query with quick stats can use since the bg builds will be queued up when we hit the limits. I will add threadpool monitoring metrics so that actual threadpool sizes can be monitored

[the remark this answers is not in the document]
ZacBlanco: I'm fine setting to the default values you suggested.

My main reason for suggesting a dynamic default is because I've seen that once you start trying to scale to different hardware (up or down), it can take a fair bit of time to tune and debug which static configs need to be changed for a given deployment. A dynamic value as the default can potentially prevent the need to manually tune which makes the software more user friendly. I know it's generally not the norm with these types of systems, but I haven't seen strong arguments as to why we shouldn't be trying to tune some of the defaults dynamically based on the current hardware/JVM limits. More of a philosophical debate, but I would personally like to see things move in that direction

ZacBlanco (approved):
Updates LGTM. I think the only missing bits are documentation of quick stats as a whole and associated config properties

presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java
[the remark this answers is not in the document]
aaneja: Prototype commit to have more than one query wait for stats to be built is : https://github.com/prestodb/presto/pull/21436/commits/e3e001ec26e20093cc11e4064ea90f6af6f0decf

Lemme know your thoughts ?

presto-docs/src/main/sphinx/connector/hive.rst
```diff
@@ -654,6 +654,83 @@ This query will collect statistics for 2 partitions with keys:
 * ``partition1_value1, partition1_value2``
 * ``partition2_value1, partition2_value2``
 
+Quick Stats
+--------------------------------------
+
+The Hive connector can build basic statistics for partitions with missing statistics
+by examining file or table metadata. For example, parquet footers can be used to infer
```
steveburnett, line 661:
```suggestion
by examining file or table metadata. For example, Parquet footers can be used to infer
```

presto-docs/src/main/sphinx/connector/hive.rst
```diff
@@ -654,6 +654,83 @@ This query will collect statistics for 2 partitions with keys:
...
+row counts, number of null and min/max values. These statistics help in query planning,
```
steveburnett, line 662:
```suggestion
row counts, number of nulls, and min/max values. These statistics help in query planning,
```
I'm incorporating your typo correction into my suggestion, which by itself I'm merely suggesting adding a comma.

presto-docs/src/main/sphinx/connector/hive.rst
```diff
@@ -654,6 +654,83 @@ This query will collect statistics for 2 partitions with keys:
...
+and serve as interim cache of stats for partitions which haven't yet been ANALYZEd
```
steveburnett, line 663:
```suggestion
and serve as an interim cache of stats for partitions which haven't yet been ANALYZEd. 
```
Minor grammar and punctuation. ANALYZEd feels a little odd - not wrong, but I'm not sure when I read this what it means. Do you have another way that you could describe what you mean here?

presto-docs/src/main/sphinx/connector/hive.rst
```diff
@@ -654,6 +654,83 @@ This query will collect statistics for 2 partitions with keys:
...
+   -
+
+      - Property Name
+      - Description
+      - Default
+   -
+
+      - ``hive.quick-stats.enabled``
+      - Enable stats collection through quick stats providers. Also
+         toggleable through the ``quick_stats_enabled`` session property
```
steveburnett, line 680:
```suggestion
        toggleable through the ``quick_stats_enabled`` session property.
```
Every text line after the first in the Description column has one extra space indent at the beginning. I'll try to edit these here, but if you can do a local build of the docs it'd help. I'm including screenshot of the build of the current file, and a screenshot of my local edited build to show the difference. I'm sorry this is going to be a lot of comments for almost entirely single-space deleting nits! 
current
Screenshot 2024-02-07 at 12 24 08 PM

my edited and local-built example
Screenshot 2024-02-07 at 12 33 55 PM

presto-docs/src/main/sphinx/connector/hive.rst
```diff
@@ -654,6 +654,83 @@ This query will collect statistics for 2 partitions with keys:
...
+      - ``false``
+   -
+
+      - ``hive.quick-stats.max-concurrent-calls``
+      - Quick stats are built for multiple partitions concurrently. This
+         property sets the maximum number of concurrent builds that can
```
steveburnett, line 686:
```suggestion
        property sets the maximum number of concurrent builds that can
```

presto-docs/src/main/sphinx/connector/hive.rst
```diff
@@ -654,6 +654,83 @@ This query will collect statistics for 2 partitions with keys:
...
+         be made
```
steveburnett, line 687:
```suggestion
        be made.
```

presto-docs/src/main/sphinx/connector/hive.rst
```diff
@@ -654,6 +654,83 @@ This query will collect statistics for 2 partitions with keys:
...
+      - ``100``
+   -
+
+      - ``hive.quick-stats.inline-build-timeout``
+      - Duration the query that initiates a quick stats build for a
+         partition should wait before timing out and returning empty
```
steveburnett, line 693:
```suggestion
        partition should wait before timing out and returning empty
```

presto-docs/src/main/sphinx/connector/hive.rst
```diff
@@ -654,6 +654,83 @@ This query will collect statistics for 2 partitions with keys:
...
+         stats. Also toggleable through the
```
steveburnett, line 694:
```suggestion
        stats. Also toggleable through the
```

presto-docs/src/main/sphinx/connector/hive.rst
```diff
@@ -654,6 +654,83 @@ This query will collect statistics for 2 partitions with keys:
...
+         ``quick_stats_inline_build_timeout`` session property
```
steveburnett, line 695:
```suggestion
        ``quick_stats_inline_build_timeout`` session property.
```

presto-docs/src/main/sphinx/connector/hive.rst
```diff
@@ -654,6 +654,83 @@ This query will collect statistics for 2 partitions with keys:
...
+      - ``60s``
+   -
+
+      - ``hive.quick-stats.background-build-timeout``
+      - If a query observes that quick stats are already being built for
```
steveburnett, line 700:
```suggestion
      - If a query observes that quick stats are being built for
```

presto-docs/src/main/sphinx/connector/hive.rst
```diff
@@ -654,6 +654,83 @@ This query will collect statistics for 2 partitions with keys:
...
+         a partition, this is the duration it will wait for those stats
```
steveburnett, line 701:
```suggestion
        a partition, this is the duration it waits for those stats
```

presto-docs/src/main/sphinx/connector/hive.rst
```diff
@@ -654,6 +654,83 @@ This query will collect statistics for 2 partitions with keys:
...
+         to be built before returning empty stats
```
steveburnett, line 702:
```suggestion
        to be built before returning empty stats.
```

presto-docs/src/main/sphinx/connector/hive.rst
```diff
@@ -654,6 +654,83 @@ This query will collect statistics for 2 partitions with keys:
...
+      - ``0s``
+   -
+
+      - ``hive.quick-stats.cache-expiry``
+      - Duration to retain the stats in the quick stats in-memory cache
```
steveburnett, line 707:
```suggestion
      - Duration to retain the stats in the quick stats in-memory cache.
```
Nit: If one item in a list needs punctuation (a period), they all do. (I wanted to explain why I made this and some of the similar nit suggestions.)

presto-docs/src/main/sphinx/connector/hive.rst
```diff
@@ -654,6 +654,83 @@ This query will collect statistics for 2 partitions with keys:
...
+      - ``24h``
+   -
+
+      - ``hive.quick-stats.reaper-expiry``
+      - If the quick stats build for a partition is stuck (say due to
```
steveburnett, line 712:
```suggestion
      - If the quick stats build for a partition is stuck (for example, due to
```

presto-docs/src/main/sphinx/connector/hive.rst
```diff
@@ -654,6 +654,83 @@ This query will collect statistics for 2 partitions with keys:
...
+         long-running IO), a reaper thread will terminate any background
```
steveburnett, line 713:
```suggestion
        long-running IO), a reaper thread terminates any background
```

presto-docs/src/main/sphinx/connector/hive.rst
```diff
@@ -654,6 +654,83 @@ This query will collect statistics for 2 partitions with keys:
...
+         build threads so that a new fetch could be triggered afresh.
```
steveburnett, line 714:
```suggestion
        build threads so that a new fetch could be triggered afresh.
```

presto-docs/src/main/sphinx/connector/hive.rst
```diff
@@ -654,6 +654,83 @@ This query will collect statistics for 2 partitions with keys:
...
+         This property controls the duration, after a background build
```
steveburnett, line 715:
```suggestion
        This property controls the duration, after a background build
```

presto-docs/src/main/sphinx/connector/hive.rst
```diff
@@ -654,6 +654,83 @@ This query will collect statistics for 2 partitions with keys:
...
+         thread is started, for the reaper to perform the termination
```
steveburnett, line 716:
```suggestion
        thread is started, for the reaper to perform the termination.
```

presto-docs/src/main/sphinx/connector/hive.rst
```diff
@@ -654,6 +654,83 @@ This query will collect statistics for 2 partitions with keys:
...
+      - ``5m``
+   -
+
+      - ``hive.quick-stats.parquet.max-concurrent-calls``
+      - Multiple parquet file footers are read and processed
+         concurrently. This property sets the maximum number of such
```
steveburnett, line 722:
```suggestion
        concurrently. This property sets the maximum number of
```

presto-docs/src/main/sphinx/connector/hive.rst
```diff
@@ -654,6 +654,83 @@ This query will collect statistics for 2 partitions with keys:
...
+         concurrent calls that can be made
```
steveburnett, line 723:
```suggestion
        concurrent calls that can be made.
```

presto-docs/src/main/sphinx/connector/hive.rst
```diff
@@ -654,6 +654,83 @@ This query will collect statistics for 2 partitions with keys:
...
+      - ``500``
+   -
+
+      - ``hive.quick-stats.parquet.file-metadata-fetch-timeout``
+      - Duration after which the parquet quick stats builder will fail
+         and return empty stats.
```
steveburnett, line 729:
```suggestion
        and return empty stats.
```

steveburnett (changes requested):
Thanks, this looks great! My review has a lot of comments that are almost entirely nits to fix the formatting of the Descriptions in the table. 

If you have time, you could do a local build to check if the formatting in the table is good after addressing these changes. I'll check with a new local build after you update the branch whether or not you do anyway, so don't worry about it if it's inconvenient. Thanks!

presto-docs/src/main/sphinx/connector/hive.rst
aaneja: My bad, I used pandoc to convert a markdown table to rst and missed this. Will fix this up

presto-hive/src/main/java/com/facebook/presto/hive/HiveClientConfig.java
```diff
@@ -1586,6 +1591,67 @@ public boolean isHudiMetadataEnabled()
...
+        this.quickStatsInProgressReaperExpiry = reaperExpiry;
+        return this;
+    }
+
+    public Duration getQuickStatsReaperExpiry()
+    {
+        return this.quickStatsInProgressReaperExpiry;
+    }
+
+    @Config("hive.quick-stats.parquet.footer-fetch-timeout")
```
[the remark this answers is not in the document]
aaneja: I've renamed this to `hive.quick-stats.parquet.file-metadata-fetch-timeout`. I kept the `parquet` prefix, since 1/ I think users will want individual knobs per quick stats provider 2/Migrating user configs if/when we add a new provider is avoided

presto-docs/src/main/sphinx/connector/hive.rst
aaneja: I've rewritten this to `These 'quick' statistics help in query planning,
and serve as as a temporary source of stats for partitions which haven't had ANALYZE run on
them.`

presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java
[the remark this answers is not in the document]
aaneja: Squashed this in

presto-docs/src/main/sphinx/connector/hive.rst
```diff
@@ -654,6 +654,88 @@ This query will collect statistics for 2 partitions with keys:
...
+      - If the quick stats build for a partition is stuck (for example, due to
+        a long-running IO operation), a reaper job terminates any background
+        build threads so that a new fetch could be triggered afresh.
+        This property controls the duration, after a background build
+        thread is started, for the reaper to perform the termination.
+      - ``5m``
+   -
+
+      - ``hive.quick-stats.parquet.max-concurrent-calls``
+      - Multiple parquet file footers are read and processed
```
steveburnett, line 726:
```suggestion
      - Multiple Parquet file footers are read and processed
```

presto-docs/src/main/sphinx/connector/hive.rst
```diff
@@ -654,6 +654,88 @@ This query will collect statistics for 2 partitions with keys:
...
+        concurrently. This property sets the maximum number of
+        concurrent calls that can be made.
+      - 500
+   -
+
+      - ``hive.quick-stats.parquet.file-metadata-fetch-timeout``
+      - Duration after which the parquet quick stats builder will fail
```
steveburnett, line 733:
```suggestion
      - Duration after which the Parquet quick stats builder will fail
```

steveburnett (changes requested):
Thanks for the update! Tthe table formatting looks perfect in a new local build.  

Two nits I should have noticed and mentioned yesterday, but this should be everything.

steveburnett (approved):
LGTM! (docs)

Thanks for the quick turnaround today! Pull of updated branch, new docs build, full review of new content. No further changes requested.

ZacBlanco (dismissed):

ClarenceThreepwood (dismissed):

presto-hive/src/main/java/com/facebook/presto/hive/statistics/ParquetQuickStatsBuilder.java
```diff
@@ -0,0 +1,366 @@
...
+
+        if (rolledUpColStats.isEmpty()) {
+            return PartitionQuickStats.EMPTY;
+        }
+        return new PartitionQuickStats(partitionId, rolledUpColStats.values(), filesCount);
+    }
+
+    @Managed
+    @Nested
+    public ThreadPoolExecutorMBean getExecutor()
```
tdcmeehan, line 350: Should we add additional stats, such as how long it took to open a footer, how many bytes were read, and simple call count metrics like how many times quick stats were requested and how many files were opened?

aaneja: I have simple call count metrics in the QuickStatsProvider (as JMX)-
 
https://github.com/aaneja/presto/blob/39f417eebe6b4a7d2912efb97589cbee634762cd/presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java#L120-L136

The latency to build Quick stats via a provider (as runtime stats) which are a proxy for the worst case footer fetch latency (since we fetch footers concurrently)-

https://github.com/aaneja/presto/blob/39f417eebe6b4a7d2912efb97589cbee634762cd/presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java#L351

Number of footers that were scanned for a partition  (as runtime stats) -

https://github.com/aaneja/presto/blob/39f417eebe6b4a7d2912efb97589cbee634762cd/presto-hive/src/main/java/com/facebook/presto/hive/statistics/ParquetQuickStatsBuilder.java#L323-L324

I can track total bytes fetched, but since parquet footers are usually small (expected footer size 16K bytes) and fetched in a single FS call, we can estimate them via `file_count*16K`. FS stats from the coordinator (like `PrestoS3FileSystemStats`) should also give us a detailed picture for debugging

presto-hive/src/main/java/com/facebook/presto/hive/statistics/ColumnQuickStats.java
```diff
@@ -0,0 +1,125 @@
...
+ */
+public class ColumnQuickStats<T extends Comparable<T>>
+{
+    private final String columnName;
+    private final Type statType;
+    private long rowCount;
+    private long nullsCount;
+
+    private T minValue;
+    private T maxValue;
```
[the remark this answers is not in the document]
aaneja: I decided to punt on this. `ColumnQuickStats` needs to be a mutable POJO to support additively building stats from metadata, so we need to track individual values in a form that is easy to mutate

If/When we have a need to track a new stat or track a stats without having the base stats (null count, row count) available, we can refactor this class to return a built `Map<ColumnStatisticType,StatsObject>` which would represent available stats.

presto-hive/src/main/java/com/facebook/presto/hive/statistics/MetastoreHiveStatisticsProvider.java
```diff
@@ -98,20 +103,23 @@ public class MetastoreHiveStatisticsProvider
...
         requireNonNull(metastore, "metastore is null");
         this.statisticsProvider = (session, table, hivePartitions) -> getPartitionsStatistics(session, metastore, table, hivePartitions);
+        this.quickStatsProvider = quickStatsProvider;
     }
 
     @VisibleForTesting
     MetastoreHiveStatisticsProvider(PartitionsStatisticsProvider statisticsProvider)
     {
         this.statisticsProvider = requireNonNull(statisticsProvider, "statisticsProvider is null");
+        this.quickStatsProvider = null;
```
tdcmeehan, line 119: It would be nice to ensure this is never null.  If this is for testing, can we wire in a dummy quick stats provider?

tdcmeehan (commented):
Two minor questions

presto-hive/src/main/java/com/facebook/presto/hive/statistics/ParquetQuickStatsBuilder.java
aaneja: I've added some more JMX metrics - 

https://github.com/aaneja/presto/blob/90fcf6772af5612f7d798b3a28b857233ceacb1e/presto-hive/src/main/java/com/facebook/presto/hive/statistics/ParquetQuickStatsBuilder.java#L256-L275

https://github.com/aaneja/presto/blob/90fcf6772af5612f7d798b3a28b857233ceacb1e/presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java#L140-L145

tdcmeehan (approved):

PR #21440, opened 2023-11-26
[native] Make HTTP client connection pool configurable and disable by default

Yuhta:
If too many endpoints are pinned to the same event base, it will create some congestions on these connections.  Make it default to disabled for now until we find a smarter way to balance the event bases and connections.

### DIFF
--- a/presto-native-execution/presto_cpp/main/http/HttpClient.h
+++ b/presto-native-execution/presto_cpp/main/http/HttpClient.h
@@ -124,6 +124,8 @@ class HttpClient : public std::enable_shared_from_this<HttpClient> {
       const std::string& ciphers = "",
       std::function<void(int)>&& reportOnBodyStatsFunc = nullptr);
 
+  ~HttpClient();
+
   // TODO Avoid copy by using IOBuf for body
   folly::SemiFuture<std::unique_ptr<HttpResponse>> sendRequest(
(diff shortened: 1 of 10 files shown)


spershin (approved):
Looks good!
Thanks!

presto-native-execution/presto_cpp/main/http/HttpClient.h
```diff
@@ -136,7 +138,8 @@ class HttpClient : public std::enable_shared_from_this<HttpClient> {
 
  private:
   folly::EventBase* const eventBase_;
-  proxygen::SessionPool* const sessionPool_;
+  proxygen::SessionPool* sessionPool_;
```
xiaoxmeng, line 141: Put const members first? Thanks!

presto-native-execution/presto_cpp/main/http/HttpClient.h
```diff
@@ -136,7 +138,8 @@ class HttpClient : public std::enable_shared_from_this<HttpClient> {
...
+  std::unique_ptr<proxygen::SessionPool> sessionPoolHolder_;
```
xiaoxmeng, line 142: Can you comment when sessionPoolHolder_ is created? Thanks!

xiaoxmeng (approved):

PR #21443, opened 2023-11-27
[native] Change HTTP executor num threads config param to HW multiplier.

spershin:
* Change HTTP executor num threads config param to be HW multiplier, instead of
being the number of threads.
* Made startup and shutdown messages more friendly.
* Joining two HTTP executors during shutdown, so if they have some threads stuck, we will see which executor is affected.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/PrestoServer.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoServer.cpp
@@ -114,7 +114,6 @@ void PrestoServer::run() {
   auto nodeConfig = NodeConfig::instance();
   auto baseVeloxQueryConfig = BaseVeloxQueryConfig::instance();
   int httpPort{0};
-  int httpExecThreads{0};
 
   std::string certPath;
   std::string keyPath;
@@ -173,7 +172,6 @@ void PrestoServer::run() {
     }
 
     nodeVersion_ = systemConfig->prestoVersion();
-    httpExecThreads = systemConfig->httpExecThreads();
     environment_ = nodeConfig->nodeEnvironment();
     nodeId_ = nodeConfig->nodeId();
     address_ = nodeConfig->nodeInternalAddress(
@@ -311,7 +309,8 @@ void PrestoServer::run() {
       systemConfig->numIoThreads(),
       std::make_shared<folly::NamedThreadFactory>("PrestoWorkerNetwork"));
 
-  PRESTO_STARTUP_LOG(INFO) << "Exchange Http executor has "
+  PRESTO_STARTUP_LOG(INFO) << "Exchange Http IO executor '"
+                           << exchangeHttpExecutor_->getName() << "' has "
                            << exchangeHttpExecutor_->numThreads()
                            << " threads.";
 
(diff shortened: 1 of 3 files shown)


presto-native-execution/presto_cpp/main/common/Configs.h
```diff
@@ -164,8 +164,10 @@ class SystemConfig : public ConfigBase {
       "task.max-drivers-per-task"};
   static constexpr std::string_view kConcurrentLifespansPerTask{
       "task.concurrent-lifespans-per-task"};
-  static constexpr std::string_view kHttpExecThreads{"http_exec_threads"};
-  static constexpr std::string_view kNumHttpCpuThreads{"num-http-cpu-threads"};
+  static constexpr std::string_view kHttpServerNumExecThreadsHwMultiplier{
+      "http-server.num-exec-threads-hw-multiplier"};
```
Yuhta, line 168: Maybe `http-server.num-io-threads-hw-multiplier`?

presto-native-execution/presto_cpp/main/common/Configs.h
```diff
@@ -429,9 +431,9 @@ class SystemConfig : public ConfigBase {
 
   int32_t concurrentLifespansPerTask() const;
 
-  int32_t httpExecThreads() const;
+  int32_t numHttpIoThreadsHwMultiplier() const;
```
xiaoxmeng, line 434: Do we want to use a double so we can configure it to a fraction of hw concurrency? Thanks!

spershin: Maybe this is not a bad idea.
I thought about it before, but not this time.

presto-native-execution/presto_cpp/main/common/Configs.h
```diff
@@ -164,8 +164,17 @@ class SystemConfig : public ConfigBase {
       "task.max-drivers-per-task"};
   static constexpr std::string_view kConcurrentLifespansPerTask{
       "task.concurrent-lifespans-per-task"};
-  static constexpr std::string_view kHttpExecThreads{"http_exec_threads"};
-  static constexpr std::string_view kNumHttpCpuThreads{"num-http-cpu-threads"};
```
amitkdutta, line 168: This means the new deployment scripts need to pass new name. And new deployment script can not be used to update existing binary anymore because existing binary does not recognize new name. This might require adding versioning in deployment scripts.

spershin: @amitkdutta 
We discussed this. We don't need the versioning here.
Test and dev environment would use the default values for a very short time.
Prod won't be pushed.
We'll update the scripts as soon as this one lands.

Yuhta (approved):

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -621,9 +649,12 @@ void PrestoServer::stop() {
   // Make sure we only go here once.
   auto shutdownOnsetSec = SystemConfig::instance()->shutdownOnsetSec();
   if (!shuttingDown_.exchange(true)) {
-    PRESTO_SHUTDOWN_LOG(INFO) << "Initiating shutdown. Will wait for "
-                              << shutdownOnsetSec << " seconds.";
-    this->setNodeState(NodeState::SHUTTING_DOWN);
+    PRESTO_SHUTDOWN_LOG(INFO) << "Shutdown has been requested. "
+                                 "Setting node state to 'shutting down'.";
+    setNodeState(NodeState::SHUTTING_DOWN);
```
xiaoxmeng, line 654: setNodeState is pretty lightweight. Shall we just merge the two logs into one? thanks!

xiaoxmeng (approved):
@spershin LGTM % one nit. Thanks!

presto-native-execution/presto_cpp/main/PrestoServer.cpp
spershin: @xiaoxmeng 
I split it up, because there was confusion some people expressed about the previous message.
So, now we have two clear messages, so there is very little confusion.
Also, potentially, if in the future setNodeState() gets more complicated and can get stuck, we'll see it in the log.
One extra message per start/stop is not an issue.

PR #21445, opened 2023-11-28
Support insert values with batch line of parameters

hantangwangd:
## Description

This PR support PreparedStatement.addBatch()/PreparedStatement.executeBatch() for ``insert values`` statement.

The change are broken into 4 commits, corresponding 4 main parts.

1. Add syntax support for execute batch using multiple row of data
2. Support multiple line parameters in Analysis, and make corresponding adjustment
3. Customize the execution & output refactor logic for ``Prepare/ExecuteBatch``
4. Refactor jdbc client to support ``PreparedStatement.addBatch()/PreparedStatement.executeBatch()`` for insert values

### Syntax Support

Support syntax ``EXECUTE BATCH <queryId> USING (v1, v2), (v3, v4), (v5, v6)``. It is necessary to let the server aware that the current executing statement is a ``execute batch``, so it could customize the return result.

### Internal Support Multiple Line Parameters

Use data structure ``MultiLineParameters`` to replace originally ``Map<NodeRef<Parameter>, Expression>``, and parameters are parsed and maintained in a multiple line form.

Refactor the logic involving ``ValuesNode`` to support multiple line parameters, and leave other statement logic unaffected by using ``MultiLineParameters.getFirstRowOfParametersIfExists()``.

### Execution & Output Refactor Logic for ``Prepare/Execute batch``

``Prepare`` statement would return a result of type VarbinaryType to jdbc client, which could be deserialized to an object instance of ``ResultForPrepare``. It contains message ``insertValuesWithParameter`` and ``typesForValidate``, which could support the jdbc client to perform verification in advance.

``Execute batch`` statement would return a result of type ArrayType[Integer], which would be transformed to a int[] by jdbc client and return to end users.

So we designed an extra architecture to support statements of different categories to register their own ``ExecuteAndOutputHandler`` at dispatch stage, run their customized execution logic and store the intermediate message at running stage, and eventually use the intermediate message to refactor the output result.

### Refactor JDBC Client

Currently we support ``PreparedStatement.addBatch()/PreparedStatement.executeBatch()`` only for insert values. The function behavior is comply with java.sql api and try to consistent with MySQL.

## Motivation and Context

We would want the support by PreparedStatement.addBatch()/PreparedStatement.executeBatch() when we try to load data from somewhere else like a stream datasource or some other storages unsupported by presto engine yet. We would read and collect a batch of data and insert them all at once rather than insert the data one single row at a time. 

Concatenating an insert values string for multiple rows is inefficient and fragile, especially when involving various data types. So, it is very necessary to support addBatch()/executeBatch() for ``insert values`` PreparedStatement.

## Impact

- New PreparedStatement functions support
- Internal refactors to support multiple row parameters
- An extra architecture to support the customized logic of execution and return result refactor

## Test Plan

 - Newly added test class ``TestJdbcPreparedStatementBatch``
 - Should not affect existing test cases

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-analyzer/src/main/java/com/facebook/presto/sql/analyzer/MultiLineParameters.java
+++ b/presto-analyzer/src/main/java/com/facebook/presto/sql/analyzer/MultiLineParameters.java
@@ -1,0 +1,99 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.sql.analyzer;
+
+import com.facebook.presto.spi.PrestoException;
+import com.facebook.presto.sql.tree.Expression;
+import com.facebook.presto.sql.tree.NodeRef;
+import com.facebook.presto.sql.tree.Parameter;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableMap.Builder;
+
+import java.util.List;
+import java.util.Map;
+
+import static com.facebook.presto.spi.StandardErrorCode.INVALID_ARGUMENTS;
+import static java.util.Objects.requireNonNull;
+
+public class MultiLineParameters
+{
+    public static final MultiLineParameters EMPTY = new MultiLineParameters(ImmutableList.of(), ImmutableList.of());
+
+    private final List<NodeRef<Parameter>> parameterRefs;
+    private final List<List<Expression>> parameterExpressions;
+
+    public MultiLineParameters(List<NodeRef<Parameter>> parameterRefs, List<List<Expression>> parameterExpressions)
+    {
+        this.parameterRefs = requireNonNull(parameterRefs, "parameterRefs is null");
+        this.parameterExpressions = requireNonNull(parameterExpressions, "parameterExpressions is null");
+        for (List<Expression> expressions : parameterExpressions) {
+            if (expressions.size() != parameterRefs.size()) {
+                throw new PrestoException(INVALID_ARGUMENTS, "Parameters not compatible");
+            }
+        }
+    }
+
+    public static MultiLineParameters from(Map<NodeRef<Parameter>, Expression> parameters)
+    {
+        requireNonNull(parameters, "parameters is null");
+        if (parameters.isEmpty()) {
+            return EMPTY;
+        }
+        ImmutableList.Builder refsBuilder = ImmutableList.builder();
+        ImmutableList.Builder expressionsBuilder = ImmutableList.builder();
+        parameters.entrySet().stream().forEach(entry -> {
+            refsBuilder.add(entry.getKey());
+            expressionsBuilder.add(entry.getValue());
+        });
+        return new MultiLineParameters(refsBuilder.build(), expressionsBuilder.build());
+    }
+
+    public int size()
+    {
+        return parameterExpressions.size();
+    }
+
+    public Map<NodeRef<Parameter>, Expression> getFirstRowOfParametersIfExists()
+    {
+        if (parameterExpressions.isEmpty()) {
+            return ImmutableMap.of();
+        }
+        List<Expression> singleRowExpressions = parameterExpressions.get(0);
+        Builder mapBuilder = ImmutableMap.builder();
+        for (int i = 0; i < parameterRefs.size(); i++) {
+            mapBuilder.put(parameterRefs.get(i), singleRowExpressions.get(i));
+        }
+        return mapBuilder.build();
+    }
+
+    public Map<NodeRef<Parameter>, Expression> getRowOfParameters(int rowIdx)
+    {
+        if (rowIdx >= parameterExpressions.size()) {
+            throw new PrestoException(INVALID_ARGUMENTS, "Invalidate rowIdx: " + rowIdx);
+        }
+
+        // rowIdx < 0 implies no parameters exists
+        if (rowIdx < 0) {
+            return ImmutableMap.of();
+        }
+
+        List<Expression> singleRowExpressions = parameterExpressions.get(rowIdx);
+        Builder mapBuilder = ImmutableMap.builder();
+        for (int i = 0; i < parameterRefs.size(); i++) {
+            mapBuilder.put(parameterRefs.get(i), singleRowExpressions.get(i));
+        }
+        return mapBuilder.build();
+    }
+}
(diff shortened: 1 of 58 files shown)


presto-jdbc/src/main/java/com/facebook/presto/jdbc/PrestoPreparedStatement.java
```diff
@@ -131,11 +133,59 @@ public boolean execute()
...
+        // if there is no parameter set, return an empty array directly
+        if (batchExecuteSql == null) {
+            return new int[] {};
+        }
+        super.executeBatch(batchExecuteSql);
+        ResultSet rs = getResultSet();
+
+        // executeBatch() would get a result of Array[Integer] type from server, and transform it to a int[]
+        int[] res = new int[0];
+        while (rs.next()) {
```
tdcmeehan, line 152: This `rs` should just return one result, right?

presto-jdbc/src/main/java/com/facebook/presto/jdbc/PrestoPreparedStatement.java
```diff
@@ -131,11 +133,59 @@ public boolean execute()
...
+        if (!isInsertValuesWithParameter()) {
+            throw new SQLException("Execute batch is only supported for insert values with parameter statement");
+        }
+        String batchExecuteSql = getBatchExecuteSql();
+        // if there is no parameter set, return an empty array directly
+        if (batchExecuteSql == null) {
+            return new int[] {};
+        }
+        super.executeBatch(batchExecuteSql);
+        ResultSet rs = getResultSet();
```
tdcmeehan, line 148: Wrap with a try-with-resources?

presto-main/src/main/java/com/facebook/presto/execution/CallTask.java
```diff
@@ -131,7 +131,7 @@ else if (i < procedure.getArguments().size()) {
 
         // get argument values
         Object[] values = new Object[procedure.getArguments().size()];
-        Map<NodeRef<Parameter>, Expression> parameterLookup = parameterExtractor(call, parameters);
+        Map<NodeRef<Parameter>, Expression> parameterLookup = parameterExtractor(call, parameters).getFirstRowOfParametersIfExists();
```
tdcmeehan, line 134: Should we add a method on `MultiLineParameters` that doesn't just returns the empty map if it doesn't exist, but throws with an exception to fail fast and alert on the bug?

tdcmeehan (changes requested):

presto-main/src/main/java/com/facebook/presto/server/protocol/ExecuteAndOutputHandler.java
```diff
@@ -0,0 +1,117 @@
...
+        public OutputResultData(boolean needRefactor, Iterable<List<Object>> value)
+        {
+            this.needRefactor = needRefactor;
+            this.value = requireNonNull(value, "value is null");
+        }
+
+        public boolean isNeedRefactor()
+        {
+            return needRefactor;
+        }
```
tdcmeehan, line 80: Why is this called `needRefactor`?

tdcmeehan (changes requested):

tdcmeehan:
It's not clear to me why we need to introduce a SQL grammar change in order to support batch inserts.  Can you help clarify why this is needed?  I do not see it in the SQL spec.

hantangwangd:
> It's not clear to me why we need to introduce a SQL grammar change in order to support batch inserts. Can you help clarify why this is needed? I do not see it in the SQL spec.

Thanks for pointing out the concern about SQL grammar spec. Sorry that I haven't paid a lot attention on it. I just considered the grammar ``EXECUTE BATCH`` as an internal mechanism expanded from ``EXECUTE`` to support multiple line of parameters.

Currently we do not support semicolon separated multiple sql statements, and this approach is also a bit inefficient for immutable file based storage such like Iceberg. So we try to expand the mechanism which are used to support ``EXECUTE`` to support ``PreparedStatement.executeBatch()``. That is, we send multiple line of parameters and tell server that this is an ``execute batch`` statement. Then the server will transform the multiple line parameters into a ValuesNode and handle them in one insert statement, and construct the result data of ``Array[Integer]`` type.

So, when we execute ``PreparedStatement.executeBatch()``, we should have a way to tell server that the current statement is an ``execute batch``. That's important because the server would handle the parameters and construct the result based on this message, or it may have some difficulty in distinguishing ``ExecuteBatch`` from ``Execute``.

When considering the implementation approach, using a syntax of ``EXECUTE BATCH`` is more explicit and less effect the origin ``EXECUTE``, but not absolutely necessary. We can also use a statement level session property to tell server that the current statement is a ``execute batch``. 

So, what's your opinion? If you think that the grammar expand is inappropriate, we can use statement level session property instead. Or any other better ways?

presto-analyzer/src/main/java/com/facebook/presto/sql/analyzer/MultiLineParameters.java
```diff
@@ -0,0 +1,99 @@
...
+public class MultiLineParameters
+{
+    public static final MultiLineParameters EMPTY = new MultiLineParameters(ImmutableList.of(), ImmutableList.of());
+
+    private final List<NodeRef<Parameter>> parameterRefs;
+    private final List<List<Expression>> parameterExpressions;
+
+    public MultiLineParameters(List<NodeRef<Parameter>> parameterRefs, List<List<Expression>> parameterExpressions)
+    {
+        this.parameterRefs = requireNonNull(parameterRefs, "parameterRefs is null");
```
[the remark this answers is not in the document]
hantangwangd: OK, fixed!

presto-analyzer/src/main/java/com/facebook/presto/sql/analyzer/MultiLineParameters.java
```diff
@@ -0,0 +1,99 @@
...
+        this.parameterExpressions = requireNonNull(parameterExpressions, "parameterExpressions is null");
+        for (List<Expression> expressions : parameterExpressions) {
+            if (expressions.size() != parameterRefs.size()) {
+                throw new PrestoException(INVALID_ARGUMENTS, "Parameters not compatible");
+            }
+        }
```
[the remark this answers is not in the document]
hantangwangd: Done!

presto-jdbc/src/main/java/com/facebook/presto/jdbc/PrestoPreparedStatement.java
hantangwangd: Yes, that's right.

presto-main/src/main/java/com/facebook/presto/server/protocol/ExecuteAndOutputHandler.java
hantangwangd: Oh, maybe an inaccurate wording. What I mean is if need to overwrite the original columns or result data, so I have renamed it to ``needOverwrite``. Is it OK?

tdcmeehan:
Could we implement this feature as being transactional?  I.e., within an open transaction, we collect the insert values, and at the end insert in one batch?

tdcmeehan:
Overall, I am not in favor of introducing non-spec compliant syntax changes to support JDBC specific API.  Some lines of investigation I would encourage:

1) The CLI already has a dependency on the Presto parser.  I wonder if a client side change could allow us to support batched prepared statements?
2) Does the client protocol need to be updated to support the sending of *parameters* in addition to *statements*?  We may not need SQL to execute the query: we could simply allow one to POST the parameters in some well defined format.  This could be done through a new endpoint.

hantangwangd:
@tdcmeehan OK, I got it. Thanks for your guideline. I will make some investigation on it.

PR #21446, opened 2023-11-28
[native] Remove 'http_exec_threads' from config Props in PrestoNativeQueryRunnerUtils

spershin:
The parameter has been replaced by "http-server.num-io-threads-hw-multiplier".

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/PrestoNativeQueryRunnerUtils.java
+++ b/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/PrestoNativeQueryRunnerUtils.java
@@ -175,7 +175,6 @@ public class PrestoNativeQueryRunnerUtils
                         Files.write(tempDirectoryPath.resolve("velox.properties"), "".getBytes());
                         String configProperties = format("discovery.uri=%s%n" +
                                 "presto.version=testversion%n" +
-                                "http_exec_threads=8%n" +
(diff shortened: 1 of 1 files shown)


xiaoxmeng (approved):

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/PrestoNativeQueryRunnerUtils.java
```diff
@@ -175,7 +175,6 @@ public static QueryRunner createNativeQueryRunner(
                         Files.write(tempDirectoryPath.resolve("velox.properties"), "".getBytes());
                         String configProperties = format("discovery.uri=%s%n" +
                                 "presto.version=testversion%n" +
-                                "http_exec_threads=8%n" +
```
majetideepak, line 178: Should we use `http-server.num-io-threads-hw-multiplier` instead of removing this?

spershin: @majetideepak 
The previous one was setting value to 8. With multiplier it is not possible to achieve the same result, because every HW has its own concurrency.
I was thinking on keeping it at 0.5 (mac, for example, has HW = 10), but decided to remove it altogether and use default.
If any issues arise - we'll bring it back.

PR #21466, opened 2023-12-01
[native] Move 4 num threads config params to HW multiplier.

spershin:
Extent of changes:
```
    "num-io-threads" -> "exchange.http-client.num-io-threads-hw-multiplier"
    kNumIoThreads -> kExchangeHttpClientNumIoThreadsHwMultiplier
    30 -> 1.0
    numIoThreads() -> exchangeHttpClientNumIoThreadsHwMultiplier()

    "num-connector-io-threads" -> "connector.prefetch.num-io-threads-hw-multiplier"
    kNumConnectorIoThreads -> kConnectorPrefetchNumIoThreadsHwMultiplier
    30 -> 1.0
    numConnectorIoThreads() -> connectorPrefetchNumIoThreadsHwMultiplier()

    "num-query-threads" -> "driver.num-cpu-threads-hw-multiplier"
    kNumQueryThreads -> kDriverNumCpuThreadsHwMultiplier
    std::thread::hardware_concurrency() * 4 -> 4.0
    numQueryThreads() -> driverNumCpuThreadsHwMultiplier()

    "num-spill-threads" -> "spiller.num-cpu-threads-hw-multiplier"
    kNumSpillThreads -> kSpillerNumCpuThreadsHwMultiplier
    std::thread::hardware_concurrency() -> 1.0
    numSpillThreads() -> spillerNumCpuThreadsHwMultiplier()
```

Also modified NativeExecutionSystemConfig and TestNativeExecutionSystemConfig accordingly.

Tested new config props locally on mac.

HW multiplier (or hardware concurrency multiplier) is a floating point number used in calculating
how many threads we would use a certain executor using this formula:
hw_concurrency x multiplier.

```
== NO RELEASE NOTE ==
```


presto-docs/src/main/sphinx/develop/presto-native.rst
```diff
@@ -141,15 +141,15 @@ There is also an additional parameter:
...
-    * **Type** ``integer``
-    * **Default value:** ``30``
-    * **Presto on Spark default value:** ``0``
+    * **Type** ``double``
+    * **Default value:** ``1.0``
+    * **Presto on Spark default value:** ``0.0``
 
     Size of IO executor for connectors to do preload/prefetch.  Prefetch is
-    disabled if ``num-connector-io-threads`` is set to zero.
+    disabled if ``connector.prefetch.num-io-threads-hw-multiplier`` is set to zero.
```
kewang1024, line 152: QQ: Why do we need connector here? is this config somehow related to different connectors?

kewang1024 (approved):
changes LGTM!
Question: what is hw-multiplier? maybe clarify that in commit description?

presto-native-execution/presto_cpp/main/common/Configs.h
```diff
@@ -186,12 +186,26 @@ class SystemConfig : public ConfigBase {
...
-  static constexpr std::string_view kNumConnectorIoThreads{
-      "num-connector-io-threads"};
-  static constexpr std::string_view kNumQueryThreads{"num-query-threads"};
-  static constexpr std::string_view kNumSpillThreads{"num-spill-threads"};
+  /// Floating point number used in calculating how many threads we would use
+  /// for IO executor for connectors to do preload/prefetch:
+  /// hw_concurrency x multiplier.
+  /// If 0.0 then connector preload/prefetch is disabled.
+  /// 0.0 is default.
+  static constexpr std::string_view kConnectorPrefetchNumIoThreadsHwMultiplier{
```
xiaoxmeng, line 194:
```
s/kConnectorPrefetchNumIoThreadsHwMultiplier/kConnectorNumIoThreadsHwMultiplier/
```
We don't need prefetch prefix

presto-native-execution/presto_cpp/main/common/Configs.h
```diff
@@ -335,6 +349,12 @@ class SystemConfig : public ConfigBase {
   static constexpr std::string_view kExchangeEnableConnectionPool{
       "exchange.http-client.enable-connection-pool"};
 
+  /// Floating point number used in calculating how many threads we would use
+  /// for Exchange HTTP client IO executor: hw_concurrency x multiplier.
+  /// 1.0 is default.
+  static constexpr std::string_view kExchangeHttpClientNumIoThreadsHwMultiplier{
```
xiaoxmeng, line 355: I'd just call kHttpClientNumIoThreadsHwMultiplier

spershin:
> changes LGTM! Question: what is hw-multiplier? maybe clarify that in commit description?

@kewang1024 

Thanks!
The header has comments for each config prop:
```
  /// Floating point number used in calculating how many threads we would use
  /// for IO executor for connectors to do preload/prefetch:
  /// hw_concurrency x multiplier.
  /// If 0.0 then connector preload/prefetch is disabled.
  /// 0.0 is default.
```

I can add this to the description as well.

presto-native-execution/presto_cpp/main/PrestoServer.h
```diff
@@ -174,7 +174,7 @@ class PrestoServer {
   std::unique_ptr<folly::IOThreadPoolExecutor> cacheExecutor_;
 
   // Executor for async IO for connectors.
-  std::unique_ptr<folly::IOThreadPoolExecutor> connectorIoExecutor_;
+  std::unique_ptr<folly::IOThreadPoolExecutor> connectorPrefetchIoExecutor_;
```
xiaoxmeng, line 177: I will keep the current variable name

presto-native-execution/presto_cpp/main/common/Configs.h
spershin: > kHttpClientNumIoThreadsHwMultiplier

@xiaoxmeng 

It is used by the exchange only.

presto-native-execution/presto_cpp/main/common/Configs.h
spershin: Discussed offline. Will remove 'Prefetch'.

steveburnett (approved):
LGTM! (docs)
Local build of docs, new additions look good.

spershin:
> @spershin I see that this change is related to #21443 I wonder if it is safe to use `std::thread::hardware_concurrency();` given the possible environments that Presto-on-Spark (Spark executors) can run in OSS and in bumblebee containers at Meta.
> 
> I suspect it might return the total cores on the machine v/s cores usable by the container, which might result in creating many more threads for this codepath than expected. Would be good to confirm

@shrinidhijoshi 
Indeed it will return the HW cores.
For use cases sharing with other loads this needs to be considered and values below 1.0 can be used to specify less threads than HW concurrency

shrinidhijoshi:
I see

> values below 1.0 can be used to specify less threads than HW concurrency

This might not be feasible to achieve in practice because most containers are not aware what fraction they are of the workload on the machine, but rather only aware of their absolute share. Like in the case of bumblebee, you would receive the `spark.executor.cores` value. The total number of cores on the machine is completely unpredictable at config specification time.

spershin:
@shrinidhijoshi 

About "worker.cores".
That does not strike me as a good idea.
1. It is redundant - since if just gives another lever to control the same outcome, but if improperly managed the result would be not what expected.
2. Why container environments cannot just inject hw_multiplier the way they want directly?

So basically, let say your container wants presto_cpp to use only 25% of cores.
In your scenario you will need:
1. Pass worker.cores = Pass std::thread::hardware_concurrency * 0.25
2. Pass hw_multiplier = 1.0.
3. Note that this step is very sensitive because we have multiple executors and you need to make sure you pass proper multiplier for all of them considering what you passed in worker.cores.

In the current scenario you need:
1. Pass hw_multiplier = 0.25.
2. Pass other multipliers as needed, knowing they are multiplers to the absolute HW concurency.

The latter seems simpler to me.
Also less space for human error.

spershin:
@shrinidhijoshi 
> This might not be feasible to achieve in practice because most containers are not aware what fraction they are of the workload on the machine, but rather only aware of their absolute share. Like in the case of bumblebee, you would receive the `spark.executor.cores` value. The total number of cores on the machine is completely unpredictable at config specification time.

Worth considering using HW concurrency in bumblebee too?
How does it work with some abstract number when "The total number of cores on the machine is completely unpredictable at config specification time"?

Since bumblebee's numbers (of thread) from your description are "taken from the ceiling", why not take the multipliers "from the ceiling" too?

shrinidhijoshi:
I can see the merit of the approach you are suggesting, but

> Why container environments cannot just inject hw_multiplier the way they want directly?

I believe this is the gist of the issue I am pointing at and why it might not be feasible to implement your approach.

 The concept of `fraction of the host` (hw_multiplier) is an internal metric tracked by these container managers. They won't / can't expose this detail to the application. They interface with the application using absolute numbers like `milliCores` and `memBytes`. These millicores and memBytes form the basis of constructing all the derived application level properties.

shrinidhijoshi:
> Worth considering using HW concurrency in bumblebee too?

While I am not an expert on bin packing in Resource Schedulers. I can guess why Tangram/Tupperware/Kubernetes/Mesos all operate on absolute numbers. 

If you think about how application requested resources to begin with, you would have requested using millicores and membytes. Ex: I want 1000 containers of 10G size with 8cores. This interface is universally accepted for most Resource Managers. 

Once the Resource Schedulers find a spot for your container and start it up, given your application specified the request, it already knows once running inside the container what resources are available to it. So we build up the configuration from those to parameters/properties.
 
Having the application use a new fraction/hw_multiplier would result in a redundant way of recomputing the same values what you already specified in request. 

If the container implementation ensured that the system introspection calls like `std::thread::hardware_concurrency` returned your sandbox values (request.millicores, request.memBytes) then there would be no issue, but as we don't control the sandbox/container implementation we cannot rely on introspection to figure out available resources, and have to instead use absolute-number/application properties.

spershin:
Chatted offline with @shrinidhijoshi and realized that there was confusion between config parameters:
"num-query-threads" -> "driver.num-cpu-threads-hw-multiplier" is used to define the number of threads in the Driver CPU executor.
While sapphire cares about
task.max-drivers-per-task=16, which is used to define task parallelism.

As long as Driver CPU executor's size is bigger than task.max-drivers-per-task (and by default it is 4xHW, which is not changing and sapphire is using it) the change will not affect sapphire.

shrinidhijoshi (approved):
LGTM! Thanks for the insightful discussion @spershin

spershin:
> LGTM! Thanks for the insightful discussion @spershin

@shrinidhijoshi 

Thanks for reviewing with consideration!

arhimondr (approved):

Issue #9553, opened 2017-12-14
Aggregation with  structural type causes GC pressure and excessive cross region reference


wenleix:
We have seen the following two issues in production when performing aggregations with structural type with huge number of groups. There are two separate cases:

- We see a lot of `SingleMapBlock` when the aggregation function is like `MAX_BY(map, x)` or `ARBITRARY(map)`, since an `SingleMapBlock` will be used for each group state.
  
  This issue can also happen for `SingleRowBlock` (for row type) or `XArrayBlock` (for array type). We have also seen a massive number of `Slice` in heap for `MAX(varchar)`. 

- We see a lot of `MapBlockBuilder` for `ARRAY_AGG(map)`. 

  In general, this doesn't restricts to map type, since we could see a lot of `BlockBuilder` for `ARRAY_AGG(x)` on any type, as long as the aggregation produces many groups.

wenleix:
For `array_agg`, one solution is store one huge flattened `ArrayBlockBuilder` for all groups, and we don't have per group `ArrayBlockBuilder`. We need to have a linked list for each group to construct the aggregated array per group. This is similar to the histogram aggregation improvement in https://github.com/prestodb/presto/pull/8918. 

This idea should also be applicable to the following aggregation functions: `map_agg`, `map_unoin ` and `multimap_agg`

wenleix:
Similarly, for `ARBITRARY` , one idea is to store a huge flattened `ArrayBlockBuilder` for all groups.  Another integer array needs to be maintained for the mapping from group id to the position in the `ArrayBlockBuilder`. Since for `ARBITRARY`, element only doesn't need to be updated per group, we don't need to handle update case.

While `MIN_BY` and `MAX_BY` can also use this method to store the flattened array for all groups, they need to handle updates for group. One way would be to always append to the huge flattened `ArrayBlockBuilder` and update the mapping from group id to position. However, this requires periodic compaction when the `ArrayBlockBuilder` wastes too much memory.

wenleix:
Per discussion with @highker  and @haozhun . There is a easier solution for `ARBITRARY`, `MIN_BY`, `MAX_BY`.

[...]

PR #21510, opened 2023-12-11
Add option to add limit to number of groups for khyperloglog_agg function

feilong-liu:
## Description
Add an option to limit the number of groups for the khyperloglog_agg. This limit is set in feature config, with property `khyperloglog-agg-group-limit`.

## Motivation and Context
This is to solve the same problem as mentioned in https://github.com/prestodb/presto/issues/9553 for the `khyperloglog_agg` function.
In this aggregation function, it will create multiple `HyperLogLog` objects for the same group. It can lead to cross heap memory region reference, which increase the native memory usage as described in the issue.
In this PR, I add a limit on the number of groups this aggregation can have. This limit defaults to 0 which means no limit, i.e. the current behavior.

## Impact
Have a way to limit native memory usage for this aggregation.

## Test Plan
Existing unit tests
Also run [verifier suites](https://our.internmc.facebook.com/intern/presto/verifier/results/?test_id=150748) to make sure that the aggregation function still returns the same results

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Add an feature config property `khyperloglog-agg-group-limit` to set the maximum number of groups `khyperloglog_agg` can have. It will fail the query when the limit is exceeded. The default is 0 which means no limit.
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/type/khyperloglog/KHyperLogLogAggregationFunction.java
+++ b/presto-main/src/main/java/com/facebook/presto/type/khyperloglog/KHyperLogLogAggregationFunction.java
@@ -11,30 +11,141 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package com.facebook.presto.type.khyperloglog;
 
+import com.facebook.presto.bytecode.DynamicClassLoader;
 import com.facebook.presto.common.block.BlockBuilder;
-import com.facebook.presto.common.type.StandardTypes;
-import com.facebook.presto.spi.function.AggregationFunction;
-import com.facebook.presto.spi.function.AggregationState;
-import com.facebook.presto.spi.function.CombineFunction;
-import com.facebook.presto.spi.function.InputFunction;
-import com.facebook.presto.spi.function.LiteralParameters;
-import com.facebook.presto.spi.function.OutputFunction;
-import com.facebook.presto.spi.function.SqlType;
+import com.facebook.presto.common.type.Type;
+import com.facebook.presto.metadata.BoundVariables;
+import com.facebook.presto.metadata.FunctionAndTypeManager;
+import com.facebook.presto.metadata.SqlAggregationFunction;
+import com.facebook.presto.operator.aggregation.AccumulatorCompiler;
+import com.facebook.presto.operator.aggregation.BuiltInAggregationFunctionImplementation;
+import com.facebook.presto.spi.PrestoException;
+import com.facebook.presto.spi.function.AccumulatorState;
+import com.facebook.presto.spi.function.AccumulatorStateSerializer;
+import com.facebook.presto.spi.function.aggregation.Accumulator;
+import com.facebook.presto.spi.function.aggregation.AggregationMetadata;
+import com.facebook.presto.spi.function.aggregation.GroupedAccumulator;
+import com.google.common.collect.ImmutableList;
 import io.airlift.slice.Slice;
 import io.airlift.slice.XxHash64;
 
-@AggregationFunction("khyperloglog_agg")
+import java.lang.invoke.MethodHandle;
+import java.util.List;
+
+import static com.facebook.presto.common.type.TypeSignature.parseTypeSignature;
+import static com.facebook.presto.operator.aggregation.AggregationUtils.generateAggregationName;
+import static com.facebook.presto.spi.StandardErrorCode.INVALID_FUNCTION_ARGUMENT;
+import static com.facebook.presto.spi.function.Signature.typeVariable;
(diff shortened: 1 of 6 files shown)


mlyublena (approved):
LGTM

presto-main/src/main/java/com/facebook/presto/type/khyperloglog/KHyperLogLogStateFactory.java
```diff
@@ -65,6 +81,9 @@ public void setGroupId(long groupId)
         @Override
         public void ensureCapacity(long size)
         {
+            if (groupLimit > 0 && size > groupLimit) {
+                throw new PrestoException(NOT_SUPPORTED, "GroupedKHyperLogLogState number of groups exceed limit");
```
kaikalur, line 85: Mention what the limit was

kaikalur (approved):

presto-main/src/main/java/com/facebook/presto/type/khyperloglog/KHyperLogLogAggregationFunction.java
```diff
@@ -11,76 +11,177 @@
...
+import com.facebook.presto.spi.function.AccumulatorState;
+import com.facebook.presto.spi.function.AccumulatorStateSerializer;
+import com.facebook.presto.spi.function.aggregation.Accumulator;
+import com.facebook.presto.spi.function.aggregation.AggregationMetadata;
+import com.facebook.presto.spi.function.aggregation.GroupedAccumulator;
+import com.google.common.collect.ImmutableList;
 import io.airlift.slice.Slice;
 import io.airlift.slice.XxHash64;
 
-@AggregationFunction("khyperloglog_agg")
```
agrawaldevesh, line 29: Can you explain more about these changes, is this some new style of writing agg functions ?

agrawaldevesh (commented):
I am curious how would we generalize this to cover say other expensive agg states down the line with minimal code ? Not something we ought to fix now, but what are your thoughts ? Should these checks go elsewhere then ?

presto-main/src/main/java/com/facebook/presto/type/khyperloglog/KHyperLogLogAggregationFunction.java
feilong-liu: This is an easier way to write aggregation function, corresponding to the ParametricAggregation class. The input, combine and output function together with the annotations can be used to generate the aggregation function. However, it's also not flexible. As we need to specify the limit on the number of groups, I choose to refactor the code to override functions directly.

feilong-liu:
> I am curious how would we generalize this to cover say other expensive agg states down the line with minimal code ? Not something we ought to fix now, but what are your thoughts ? Should these checks go elsewhere then ?

All aggregations are compiled by AccumulatorCompiler to generate bytecode, we may need to change here if want to add support for all functions in one place.

tdcmeehan:
> I am curious how would we generalize this to cover say other expensive agg states down the line with minimal code ? Not something we ought to fix now, but what are your thoughts ? Should these checks go elsewhere then ?

In the past, we've rewritten aggregation functions to "flatten" their internal state.  For example, for `array_agg`, we put all the values in one giant array for all groups, and do some record keeping to track which sections of the array go to which group.  This is expensive and not generalizable, though.

One potential option suggested by @ZacBlanco is to modify the code generation to specify a max grouping and make that configurable per-function, so that we don't need to do this type of code change for each aggregation function that we discover is expensive.  It is unfortunate in this PR that we revert from using the aggregation function framework, as it makes the code harder to read.

feilong-liu:
> > I am curious how would we generalize this to cover say other expensive agg states down the line with minimal code ? Not something we ought to fix now, but what are your thoughts ? Should these checks go elsewhere then ?
> 
> In the past, we've rewritten aggregation functions to "flatten" their internal state. For example, for `array_agg`, we put all the values in one giant array for all groups, and do some record keeping to track which sections of the array go to which group. This is expensive and not generalizable, though.
> 
> One potential option suggested by @ZacBlanco is to modify the code generation to specify a max grouping and make that configurable per-function, so that we don't need to do this type of code change for each aggregation function that we discover is expensive. It is unfortunate in this PR that we revert from using the aggregation function framework, as it makes the code harder to read.

The ideal case is to fix the algorithm if possible, like the `arrray_agg`, and apply limit if cannot avoid. Fortunately, this will not be a problem for Prestissimo. Change accumulator interface is a more general solution, but with more complex changes. Since this is the only aggregation we found expensive and cannot be simplified with algorithm change, I incline to keep the changes minimal and generalize it if we find more cases

ajaygeorge (approved):
LGTM

PR #21513, opened 2023-12-11
Apply INCREASE_CONTAINER_SIZE execution strategy for user supplied error codes

pgupta2:
INCREASE_CONTAINER_SIZE execution strategy is quite generic
and can be applied to multiple error codes. Adding a session
property to pass a string of comma separated error_codes which
should be retried with this execution strategy.

### DIFF
--- a/presto-spark-base/src/main/java/com/facebook/presto/spark/PrestoSparkSessionProperties.java
+++ b/presto-spark-base/src/main/java/com/facebook/presto/spark/PrestoSparkSessionProperties.java
@@ -25,6 +25,8 @@ import java.util.List;
 import java.util.Map;
 
 import static com.facebook.presto.common.type.VarcharType.VARCHAR;
+import static com.facebook.presto.spark.SparkErrorCode.SPARK_EXECUTOR_OOM;
+import static com.facebook.presto.spi.StandardErrorCode.EXCEEDED_LOCAL_MEMORY_LIMIT;
 import static com.facebook.presto.spi.session.PropertyMetadata.booleanProperty;
 import static com.facebook.presto.spi.session.PropertyMetadata.dataSizeProperty;
 import static com.facebook.presto.spi.session.PropertyMetadata.doubleProperty;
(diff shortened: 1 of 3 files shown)


presto-spark-base/src/main/java/com/facebook/presto/spark/util/PrestoSparkFailureUtils.java
```diff
@@ -103,8 +104,10 @@ private static List<ExecutionStrategy> getRetryExecutionStrategies(Session sessi
             strategies.add(DISABLE_BROADCAST_JOIN);
         }
 
-        if (isRetryOnOutOfMemoryWithIncreasedMemoryEnabled(session)
-                && (errorCode.equals(EXCEEDED_LOCAL_MEMORY_LIMIT.toErrorCode()) || errorCode.equals(SPARK_EXECUTOR_OOM.toErrorCode()))) {
+        if (isRetryOnOutOfMemoryWithIncreasedMemoryEnabled(session) &&
+                (errorCode.equals(EXCEEDED_LOCAL_MEMORY_LIMIT.toErrorCode())
```
singcha, line 108: Can we default new property with `EXCEEDED_LOCAL_MEMORY_LIMIT, SPARK_EXECUTOR_OOM` and remove this condition

singcha (approved):
Thanks for updating the PR.

NikhilCollooru (approved):

presto-spark-base/src/main/java/com/facebook/presto/spark/PrestoSparkSessionProperties.java
```diff
@@ -184,6 +185,15 @@ public PrestoSparkSessionProperties(PrestoSparkConfig prestoSparkConfig)
                         true,
                         value -> MAP_SPLITTER.split(nullToEmpty((String) value)),
                         value -> value),
+                new PropertyMetadata<>(
+                        SPARK_RETRY_ON_OUT_OF_MEMORY_WITH_INCREASED_MEMORY_SETTINGS_ERROR_CODES,
+                        "Error Codes to retry with increase memory settings",
+                        VARCHAR,
+                        List.class,
+                        ImmutableList.of("EXCEEDED_LOCAL_MEMORY_LIMIT", "SPARK_EXECUTOR_OOM"),
```
shrinidhijoshi, line 193: Nit: Would be good to derive these from the enum values from their respective error code provider - `StandardErrorCode.*` 
 and `SparkErrorCode.*` instead of hardcoded string values

shrinidhijoshi (approved):

shrinidhijoshi (approved):

Issue #21461, opened 2023-11-30
Presto allows parsing of invalid timezone names


pedroerp:
Presto's timezone parsing is too permissive and allows for timezone names that are not valid according to IANA's timezone database, in addition to other bugs:

https://en.wikipedia.org/wiki/List_of_tz_database_time_zones

The following are examples of non-compliant timezone names that are allowed:

```
ETC/+06:00
ETC/+06
ETC/+6
ETC/UTC*
ETC/UT*
```
-> the only official formats are "ETC/GMT+1" and similar

```
EST
``` 
-> not supported, but it should be

timezones in the format:
```
ETC/GMT+10:00
```
besides not being official, return the wrong result (don't flip the sign as they should)

```
+1
+01
```
officially also don't exist (only "+01:00")

## Expected Behavior
These should fail to parse, though it might break backwards compatibility.

## Current Behavior
They are parsed successfully (incorrectly).

## Steps to Reproduce
```
presto> select from_unixtime(1698528090, 'ETC/+06:00');
             _col0              
--------------------------------
 2023-10-29 03:21:30.000 +06:00 
```

and other variations described above.

Cc: @mbasmanova @zacw7

mbasmanova:
CC: @kaikalur @tdcmeehan @aditi-pandit @majetideepak

tdcmeehan:
Though this breaks backwards compatibility, these timezones seem obscure and it may be acceptable to just warn folks that such timezones will no longer be supported via a release note.

The class that has this bug is `TimeZoneKey`.  The `normalizeZoneId` method in this class appears to have some bugs, and it appears it is overly permissive in its parsing of `etc` timezones.

We could also fix up our documentation to say explicitly what values are allowed for the timezone.  AFAIA, these are tz identifier values, and GMT is such a value whereas EST is not.

pedroerp:
> AFAIA, these are tz identifier values, and GMT is such a value whereas EST is not

I thought it was weird as well, but it's listed here as a valid timezone name:

https://en.wikipedia.org/wiki/List_of_tz_database_time_zones

tdcmeehan:
I think the Wikipedia article is mixing up abbreviations with canonical zone IDs.  To be fair, they don't appear cleanly delineated in any documentation I can find from IANA, and zdump happily interchanges these abbreviations with canonical ids.

[...]

PR #21521, opened 2023-12-12
Fixed parsing of invalid timezone names #21461

karadyauran:
## Description
I have modified the logic in the `normalizeZoneId` method to address a specific case when the input zone starts with "ETC/". Previously, the method returned the `zoneId` as is if it did not match the pattern "^(ETC/)?GMT[+-]\\d{1,2}(:\\d{2})?$". I have updated this check to return the original `zoneId` instead of `zoneId` without the "ETC/" prefix.

## Motivation and Context
https://github.com/prestodb/presto/issues/21461

### Public API Changes
- [x] No changes to public APIs.

## Test Plan
- [x] **Scenario 1:** Positive Test
  - [x] Steps to Reproduce: TimeZoneKey.getTimeZoneKey("ETC/GMT+01")
  - [x] Expected Result: "+01:00"
  - [x] Actual Result: "+01:00"

- [x] **Scenario 1:** Negative Test
  - [x] Steps to Reproduce: TimeZoneKey.getTimeZoneKey("ETC/GMT*")
  - [x] Expected Result: TimeZoneNotSupportedException
  - [x] Actual Result: TimeZoneNotSupportedException

```
== NO RELEASE NOTE ==
```


tdcmeehan:
@karadyauran thank you for your PR.  

Please note, per the message above, at some point you'll need to sign the CLA.  Please follow the instructions from @linux-foundation-easycla above to get that completed.

Regarding the implementation: please note that there's a well known list of timezones that support ETC.  These are found here: https://data.iana.org/time-zones/tzdb-2021a/etcetera and here: https://data.iana.org/time-zones/tzdb-2021a/backward

It may make the most sense to test by prefix: if the prefix is `Etc/`, then it must be followed by one of : `GMT`, `Greenwich`, `UCT`, `Universal`, `UTC`, and `Zulu`.  If the prefix is `Etc/GMT`, then it must be followed by `0` or `[+-` plus any number from 0 to 12.

I think you removed the sign flipping logic, but that needs to stay: etc timezones flip it.

Let's modify the unit tests in `TestTimeZoneKey` to assert this new restrictive behavior.

Finally, we can update the release note section of your PR to note the more restrictive parsing.

karadyauran:
Ok, I understand the comments, I will fix the parsing shortly!

I've already signed the CLA.

Issue #21527, opened 2023-12-13
Security vulnerabilities issues with spring boot maven plugin and spring core


anilsomisetty:
Module: presto-benchto-benchmarks

Dependencies:

1. org.springframework.boot.spring-boot-maven-plugin:1.2.5.RELEASE

      Vulnerabilities:

    - [CVE-2023-37460](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2023-37460)
    - [CVE-2022-4245](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-4245)
    - [CVE-2022-4244](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-4244)
    - [CVE-2021-26291](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-26291)
    - [CVE-2018-1002200](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2018-1002200)

2. org.springframework.spring-core: 4.1.6.RELEASE (child dependency of com.teradata.benchto.benchto-driver:0.4)

      Vulnerabilities:

      - [CVE-2023-20863](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2023-20863)
      - [CVE-2023-20861](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2023-20861)
      - [CVE-2022-22971](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-22971)
      - [CVE-2022-22970](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-22970)
      - [CVE-2022-22968](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-22968)
      - [CVE-2018-15756](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2018-15756)
      - [CVE-2018-1275](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2018-1275)
      - [CVE-2018-1272](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2018-1272)
      - [CVE-2018-1271](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2018-1271)
      - [CVE-2018-1270](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2018-1270)
[...]

PR #21528, opened 2023-12-13
Upgraded spring boot maven plugin and spring core version to resolve multiple CVE

anilsomisetty:
## Description
This PR fixes the issue https://github.com/prestodb/presto/issues/21527

## Test Plan
I have ran the presto-benchto-benchmarks module test cases after upgrading the dependency versions and all tests are passed.

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-benchto-benchmarks/pom.xml
+++ b/presto-benchto-benchmarks/pom.xml
@@ -26,6 +26,14 @@
             <groupId>com.teradata.benchto</groupId>
             <artifactId>benchto-driver</artifactId>
             <version>0.4</version>
+            <!-- Excluding the spring-core vulnerable dependency from this
+                        latest version of benchto driver -->
+            <exclusions>
+                <exclusion>
+                    <groupId>org.springframework</groupId>
+                    <artifactId>spring-core</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
 
         <dependency>
@@ -82,6 +90,15 @@
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
         </dependency>
+
+        <!-- Adding the latest spring-core version compatible with java 8 because
+                    vulnerable version is excluded above in benchto driver-->
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-core</artifactId>
+            <version>5.3.31</version>
+        </dependency>
+
     </dependencies>
 
     <build>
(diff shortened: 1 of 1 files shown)


presto-benchto-benchmarks/pom.xml
```diff
@@ -26,6 +26,12 @@
             <groupId>com.teradata.benchto</groupId>
             <artifactId>benchto-driver</artifactId>
             <version>0.4</version>
+            <exclusions>
+                <exclusion>
```
tdcmeehan, line 30: Please add a comment explaining why we need to do this

presto-benchto-benchmarks/pom.xml
```diff
@@ -82,6 +88,13 @@
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
         </dependency>
+
+        <dependency>
+            <groupId>org.springframework</groupId>
```
tdcmeehan, line 93: Please add a comment explaining why we need to do this

tdcmeehan (changes requested):
LGTM % add some comments so we understand why we need to do this.  Ideally, we would migrate the dependency to something we own and can fix directly.

presto-benchto-benchmarks/pom.xml
anilsomisetty: Hi @tdcmeehan ,
For benchto-driver - version 0.4 is the latest I see in the maven repository but it has a child dependency spring-web which in turn has a child dependency spring-core version 4.1.6.RELEASE which has the 14 CVE reported, so I have excluded that dependency to resolve it.

presto-benchto-benchmarks/pom.xml
anilsomisetty: Since I have excluded the spring-core dependency above in teradata benchto-driver I have added a upgraded version of it which is compatible with java 8 so that code flow doesn't break without it

presto-benchto-benchmarks/pom.xml
tdcmeehan: Oh yes I understand, I mean please add a little comment in the POM file explaining this, since we don't directly use this dependency in the project.

presto-benchto-benchmarks/pom.xml
tdcmeehan: Ditto--see above.

presto-benchto-benchmarks/pom.xml
anilsomisetty: Sorry, I have misunderstood, I have added comments in the POM file for these changes, please review

tdcmeehan (approved):
LGTM--thanks for fixing these CVEs!

PR #21532, opened 2023-12-13
Add file create config to hive session properties

tanjialiang:
This session property will allow file creation option setting on per query basis. This functionality is for Presto native only.
```
== NO RELEASE NOTE ==
```


xiaoxmeng (commented):
@tanjialiang do we want to add native prefix and remove the native prefix in Prestissimo? Thanks!

presto-hive/src/main/java/com/facebook/presto/hive/HiveSessionProperties.java
```diff
@@ -147,6 +147,7 @@ public final class HiveSessionProperties
     private static final String READ_TABLE_CONSTRAINTS = "read_table_constraints";
     public static final String READ_MASKED_VALUE_ENABLED = "read_null_masked_parquet_encrypted_value_enabled";
     public static final String PARALLEL_PARSING_OF_PARTITION_VALUES_ENABLED = "parallel_parsing_of_partition_values_enabled";
+    public static final String FILE_CREATE_CONFIG = "file_create_config";
```
xiaoxmeng, line 150: native_file_create_config?

xiaoxmeng (dismissed):
@tanjialiang thanks!

presto-native-execution/presto_cpp/main/QueryContextManager.cpp
```diff
@@ -76,10 +87,12 @@ toConnectorConfigs(const protocol::SessionRepresentation& session) {
   std::unordered_map<std::string, std::unordered_map<std::string, std::string>>
       connectorConfigs;
   for (const auto& entry : session.catalogProperties) {
```
xiaoxmeng, line 89:
```
  for (const auto& entry : session.catalogProperties) {
    for (const auto& property : entry.second) {
      connectorConfigs.emplace(
          toVeloxConnectorConfig(property.first), property.second);
    }
  }
```

tdcmeehan:
I have some basic questions if you please could take a moment to help me understand.

Can you help to give some examples why we need file creation option setting on a per query basis?

This is a session property: when do we expect the user of Presto to supply a value?  And what does the value look like?  Is it a path to a config file local to the worker?

tanjialiang:
> I have some basic questions if you please could take a moment to help me understand.
> 
> Can you help to give some examples why we need file creation option setting on a per query basis?
> 
> This is a session property: when do we expect the user of Presto to supply a value? And what does the value look like? Is it a path to a config file local to the worker?

It will be a free form config that is targeted to give underlying file system the configs to create a file. This is implementation dependent so different file systems might have their own sets of configs (And that's why it's free form). For example you can specify block size to write or encodings to apply to the files. This property will be passed down to the underlying file system impl and it will be responsible to parse the configs.

tdcmeehan:
As an example, suppose my file system implementation is HDFS.  Does this mean that I will pass a config file path local to the worker to a .properties file that specifies individual HDFS properties?

tanjialiang:
> As an example, suppose my file system implementation is HDFS. Does this mean that I will pass a config file path local to the worker to a .properties file that specifies individual HDFS properties?

Not really. You do not supply a path. An example could be file_create_config="hdfs.write-file-block-size=10MB;hdfs.write-file-buffer=1MB"

presto-native-execution/presto_cpp/main/QueryContextManager.cpp
tanjialiang: catalogProperties is a map of connectorId to a map of property pairs. This nested loop might not work in this case?

tdcmeehan:
Got it.  And basically each `;` separated value is expected to be a connector config?

tanjialiang:
> Got it. And basically each `;` separated value is expected to be a connector config?

Yeah that is just an example. Again the form can de freely defined by the underlying plugged in file system. You can even use json to pass the properties in if you want to implement your file system to parse it that way.

tdcmeehan:
Have you considered putting this session property in the PrismConnector?

xiaoxmeng:
> Have you considered putting this session property in the PrismConnector?

The config value is in free form so any hive storage backend can leverage to optimize the physical data storage? Thanks!

tdcmeehan:
Typically, connectors don't take in freeform configs, because such configs may be validated inconsistently, and are often undocumented.  Presto already has a plethora of undocumented and obscure session properties that most people don't understand.  If the idea is to allow a "generic" config to allow the underlying filesystem, I think it's best to actually explicitly enumerate in the connector which configs the connector can take in to affect filesystem properties.  Imagining this from a user's perspective, I find it difficult to understand how a user would set the session without deep experience and understanding of their underlying deployment, and also knowing the nuances of how the filesystem would choose to parse this particular config.

If you really think this is a good idea and the goal is to make it easier for Meta deployments to set custom session properties to tweak filesystem behavior on a per-query basis, then perhaps give the PrismConnector a try, since this is an internal Hive-compliant connector and accomplishes this goal in a way that unblocks your production support needs.

PR #21549, opened 2023-12-15
Fix CTE Materialization unsupported hive bucket types

### LINKED ISSUE #21540: CTE materialization internal error - Hive bucket hashCode is not supported for Hive category: STRUCT and Binary
```
"type": "java.lang.UnsupportedOperationException",
    "message": "Computation of Hive bucket hashCode is not supported for Hive category: STRUCT.",
    "suppressed": [],
    "stack": [
        "com.facebook.presto.hive.HiveBucketing.hash(HiveBucketing.java:191)",
        "com.facebook.presto.hive.HiveBucketing.hashOfList(HiveBucketing.java:268)",
        "com.facebook.presto.hive.HiveBucketing.hash(HiveBucketing.java:183)",
        "com.facebook.presto.hive.HiveBucketing.getBucketHashCode(HiveBucketing.java:113)",
        "com.facebook.presto.hive.HiveBucketing.getHiveBucket(HiveBucketing.java:89)",
        "com.facebook.presto.hive.HiveBucketFunction.getBucket(HiveBucketFunction.java:76)",
        "com.facebook.presto.operator.BucketPartitionFunction.getPartition(BucketPartitionFunction.java:47)",
        "com.facebook.presto.operator.repartition.PartitionedOutputOperator$PagePartitioner.partitionPage(PartitionedOutputOperator.java:433)",
        "com.facebook.presto.operator.repartition.PartitionedOutputOperator.addInput(PartitionedOutputOperator.java:269)",
        "com.facebook.presto.operator.Driver.processInternal(Driver.java:438)",
        "com.facebook.presto.operator.Driver.lambda$processFor$9(Driver.java:311)",
        "com.facebook.presto.operator.Driver.tryWithLock(Driver.java:732)",
        "com.facebook.presto.operator.Driver.processFor(Driver.java:304)",
        "com.facebook.presto.execution.SqlTaskExecution$DriverSplitRunner.processFor(SqlTaskExecution.java:1079)",
        "com.facebook.presto.execution.executor.PrioritizedSplitRunner.process(PrioritizedSplitRunner.java:165)",
        "com.facebook.presto.execution.executor.TaskExecutor$TaskRunner.run(TaskExecutor.java:670)",
  
    ],
```

jaystarshot:
Fixes #21540

Hive catalog does not allow bucketing on a lot of presto types including user defined types. [link](https://github.com/prestodb/presto/blob/15bcbd0e9450b91bde5c68b89f64cfb81af0d3b9/presto-hive/src/main/java/com/facebook/presto/hive/HiveBucketing.java#L142). 

Hence making cte materialization avoid those types to bucket on.  If all columns are of unsupported types, then the cte will not be materialized

Added test cases for all supported and unsupported types to make sure that the queries are successful

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

Hive Changes
* Fixed CTE Materialization for unsupported hive bucket types

### CODE REVIEW

File: presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/CteUtils.java [hunk 1]
```diff
@@ -57,14 +73,42 @@ public static Integer getCtePartitionIndex(List<VariableReferenceExpression> out
...
     {
-        return !(var.getType() instanceof RowType);
+        return isPrestoTypeBucketable(var.getType());
     }
 
+    /*
+       Currently, Hive bucketing does not support bucketing on these Presto types and user defined types
+       CHAR, DECIMAL, VARBINARY, ROW
+   */
+    public static boolean isPrestoTypeBucketable(Type type)
```
tdcmeehan, line 85: Can we think of a way to do this that doesn't require engine changes that presume the underlying connector's capabilities?  I'm just thinking how this might work if we use something like Iceberg.

tdcmeehan (changes requested):

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/CteUtils.java
jaystarshot: Yes, I thought of the same but that requires spi changes. (isTypeBucketable etc)

jaystarshot: Also not sure the iceberg connector supports temporary table functionality atm

tdcmeehan: SPI changes could be fine if they're targeted and allow the engine to function properly without baking in underlying presumptions of the storage subsystem.

presto-tpch/src/main/java/com/facebook/presto/tpch/TpchMetadata.java
```diff
@@ -448,6 +448,12 @@ public void finishStatisticsCollection(ConnectorSession session, ConnectorTableH
         // do nothing
     }
 
+    @Override
+    public boolean isPrestoTypeCompatibleWithTemporaryTableBucketing(ConnectorSession session, String catalogName, Type prestoType)
+    {
+        return true;
```
jaystarshot, line 454: Added so that the testing framwork (logicalCteOptimizer) can test successfully.

jaystarshot:
@tdcmeehan I have refactored to check via SPI, please review

presto-spi/src/main/java/com/facebook/presto/spi/connector/ConnectorMetadata.java
```diff
@@ -321,6 +321,14 @@ default ConnectorTableHandle createTemporaryTable(ConnectorSession session, List
         throw new PrestoException(NOT_SUPPORTED, "This connector does not support creating temporary tables");
     }
 
+    /**
+     * Returns true if the connector metadata allows for bucketing based on the Presto type during the creation of a temporary table.
+     */
+    default boolean isPrestoTypeCompatibleWithTemporaryTableBucketing(ConnectorSession session, String catalogName, Type prestoType)
```
tdcmeehan, line 327: Can we use a more concise name, like `isTypeBucketable`?

jaystarshot:
Actually I had hardcoded all the uses of partitioning_provider_catalog to "hive" in our testing. On correcting and using partitioning_provider_catalog property,  I see unrelated failures in different stack like [https://github.com/prestodb/presto/blob/d121d92484856665157f36a74151be3ce3a3f474/p[…]com/facebook/presto/sql/planner/optimizations/AddExchanges.java](https://github.com/prestodb/presto/blob/d121d92484856665157f36a74151be3ce3a3f474/presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/AddExchanges.java#L1476) without cte materialization
Seems like the property is used in some unrelated places like this one. Failures are the same, (hive bucketing not supported) which suggests that we should use a different session property for cte partitioning.

jaystarshot:
Should I separate those session properties in https://github.com/prestodb/presto/pull/21625 ?

jaystarshot:
Revisiting and regaining context

jaystarshot:
@tdcmeehan  During his review mentioned that the stack trace in the attached ticket ( https://github.com/prestodb/presto/issues/21540)was a bit suspicious since it suggests that a bucketFunctionType HIVE_COMPATIBLE was used. 
I digged a little and I think this issue has already been solved for exchange materialization [commit](https://github.com/prestodb/presto/commit/cb69e899efbabf35a09003852194800309f53dad). Setting the `hive.bucket_function_type_for_exchange = PRESTO_NATIVE`, works well in this case.
This property is used [here](https://github.com/prestodb/presto/blob/496e98ca25fb243ab25731d188c2fd56eba3b3e9/presto-hive/src/main/java/com/facebook/presto/hive/HiveMetadata.java#L2916) in code which was the main fix.
I think this would be the proper fix in our case too.
We might need an extension to the SPI there because we don't want to set `hive.bucket_function_type_for_exchange = PRESTO_NATIVE` for non materialized exchanges in case of cte materialization.

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/CteUtils.java
```diff
@@ -1,76 +0,0 @@
...
-import com.facebook.presto.common.type.VarcharType;
-import com.facebook.presto.common.type.Varchars;
-import com.facebook.presto.spi.PrestoException;
-import com.facebook.presto.spi.relation.VariableReferenceExpression;
-
-import java.util.List;
-
-import static com.facebook.presto.spi.StandardErrorCode.GENERIC_INTERNAL_ERROR;
-
-public class CteUtils
```
tdcmeehan, line 26: Very nice!

tdcmeehan (approved):
LGTM.  @feilong-liu would you like to have a look at the current PR again?

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/CteUtils.java
jaystarshot: @tdcmeehan  I just saw and one test case is failing
 ```
@Test
    public void testCteWithZeroLengthVarchar()
    {
        String testQuery = "WITH temp AS (" +
                "  SELECT * FROM (VALUES " +
                "    (CAST('' AS VARCHAR(0)), 9)" +
                "  ) AS t (text_column, number_column)" +
                ") SELECT * FROM temp";
        QueryRunner queryRunner = getQueryRunner();
        compareResults(queryRunner.execute(getMaterializedSession(),
                        testQuery),
                queryRunner.execute(getSession(),
                        testQuery));
    }
```

Stack Trace
```
Caused by: java.lang.RuntimeException: Varchar length 0 out of allowed range [1, 65535]
	at org.apache.hadoop.hive.serde2.typeinfo.BaseCharUtils.validateVarcharParameter(BaseCharUtils.java:32)
	at org.apache.hadoop.hive.serde2.typeinfo.VarcharTypeInfo.<init>(VarcharTypeInfo.java:33)
	at org.apache.hadoop.hive.serde2.typeinfo.TypeInfoFactory.createPrimitiveTypeInfo(TypeInfoFactory.java:159)
	at org.apache.hadoop.hive.serde2.typeinfo.TypeInfoFactory.getPrimitiveTypeInfo(TypeInfoFactory.java:117)
	at org.apache.hadoop.hive.serde2.typeinfo.TypeInfoFactory.getVarcharTypeInfo(TypeInfoFactory.java:183)
	at com.facebook.presto.hive.HiveTypeTranslator.translate(HiveTypeTranslator.java:98)
	at com.facebook.presto.hive.HiveType.toHiveType(HiveType.java:218)
	at com.facebook.presto.hive.HiveMetadata.getColumnHandles(HiveMetadata.java:3584)
	at com.facebook.presto.hive.HiveMetadata.createTemporaryTable(HiveMetadata.java:1107)


```


So we might need this check for the 0 length Varchar due to , because  write with 0 length varchar fails.

See reference - While Presto supports Varchar of length 0 (as discussed in https://github.com/trinodb/trino/issues/1136

Comment (jaystarshot) [reply presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/CteUtils.java:26]: I am hence planning to disallow cte creation if there is a 0 length varchar

Comment (jaystarshot) [reply presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/CteUtils.java:26]: But that issue should exist in ctas as well i think, will differ this test case into a new issue

Comment (jaystarshot) [reply presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/CteUtils.java:26]: Created #21791 and adding todos in test

Comment (jaystarshot) [reply presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/CteUtils.java:26]: Confirmed that ```
CREATE TABLE tmp.test2512351 AS (
    SELECT * FROM (VALUES 
        (CAST('' AS VARCHAR(0)), 9)
    ) AS t(col1, col2)
);
```
fails in Presto

jaystarshot:
cc: @tdcmeehan, need an approval again thanks! Fixed some tests which were based on the previous unsupported bucket functions

tdcmeehan (approved):

PR #21557, opened 2023-12-16
Add prestoVersion to session

NikhilCollooru:
## Description
This PR adds presto server version to the session.

## Motivation and Context
Currently we apply session properties via `SessionPropertyConfigurationManager`. But we want to apply these session properties based on the server version. To do this we should have the presto server version in the `SessionConfigurationContext` / `Session` so that the implementers of `SessionPropertyConfigurationManager.getCatalogSessionProperties()` can make a decision based on the presto version
We want to handle the scenario where we have a single file that specifies all the session properties to be applied but have multiple server deployments with different presto versions . So we want to be able to specify the minimum version from which the session property should be added. 

## Impact
None

## Test Plan
Tests already exist.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-hive/src/test/java/com/facebook/presto/hive/AbstractTestHiveClient.java
+++ b/presto-hive/src/test/java/com/facebook/presto/hive/AbstractTestHiveClient.java
@@ -1198,6 +1198,12 @@ public abstract class AbstractTestHiveClient
             {
                 return WarningCollector.NOOP;
             }
+
+            @Override
+            public String getPrestoVersion()
+            {
+                return "test_version";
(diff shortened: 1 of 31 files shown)


ajaygeorge:
can you take a look at the test failures to see if they are related

NikhilCollooru:
> can you take a look at the test failures to see if they are related

The tests came clean now.

NikhilCollooru:
Found an easier way to achieve this via Meta's implementation of SessionPropertyConfigurationManager. Closing this PR.

PR #21568, opened 2023-12-18
Add KLL Sketch Functions

ZacBlanco:
## Description

This change adds KLL Sketch support from the Apache DataSketches library. One of the benefits of KLL sketches is that the implementation supports more datatypes than the existing quantile digest and T-Digest implementations. It also has provably-bounded error compared to the T-Digest

The following datatype is added

- `kllsketch(T)`: Basically a wrapper around `varbinary`, similar to `tdigest` and `qdigest` types

The following functions are added

- `sketch_kll`: computes a KLL sketch
- `sketch_kll_with_k`: computes a kll sketch with the given value for k
- `sketch_kll_quantile`: queries the sketch for a particular quantile
- `sketch_kll_rank`: queries the sketch for a particular rank (inverse of the quantile function)

## Motivation and Context

Eventual use by the optimizer after #21236. But also good to have since HMS is [releasing support for them soon](https://issues.apache.org/jira/browse/HIVE-26221).

## Impact

Two new aggregation functions, two new scalar functions

## Test Plan

Standard aggregation unit tests and scalar function unit tests

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* New support for Apache DataSketches KLL sketch with the sketch_kll and related family of functions
```

### DIFF
--- a/presto-docs/src/main/sphinx/functions/sketch.rst
+++ b/presto-docs/src/main/sphinx/functions/sketch.rst
@@ -10,7 +10,14 @@ functions which result in full accuracy.
 Presto provides support for computing some sketches available in the `Apache
 DataSketches`_ library. 
 
-.. function:: sketch_theta(data) -> varbinary
+Theta Sketches
+--------------
+
+Theta sketches enable distinct value counting on datasets and also provide the
+ability to perform set operations. For more information on Theta sketches,
+please see the Apache Datasketches `Theta sketch documentation`_.
+
+.. function:: sketch_theta(x) -> varbinary
 
     Computes a `theta sketch`_ from an input dataset. The output from
     this function can be used as an input to any of the other ``sketch_theta_*``
@@ -28,5 +35,44 @@ DataSketches`_ library.
     the number of retained entries in the sketch.
 
 
+KLL Sketches
+------------
+
+Kll Sketches are an implementation of a quantiles sketch. For more information
+about the KLL sketch, see the Apache Datasketches `KLL Sketch documentation`_.
+
+
+.. function:: sketch_kll[T](x: T) -> kllsketch[T]
+
+    This computes a KLL Sketch. The stored form is the little-endian serialized
+    version of the Apache DataSketches KLL Sketch.
+
+.. function:: sketch_kll_with_k[T](x: T, k: int) -> kllsketch[T]
+
+    This computes a KLL Sketch using the supplied value for ``k``. The ``k`` parameter must be in
+    the range [8..65535]. It controls the accuracy of the sketch. Smaller ``k`` is less accurate but
+    consumes less storage. A larger ``k`` will be more accurate but consume more storage. For more
+    information on the ``k`` parameter, refer to the `KLL Sketch documentation`_. The serialized
+    form of the sketch returned by this function is the same as the `sketch_kll` function.
+
+.. function:: sketch_kll_quantile[T](sketch: kllsketch[T], rank: double[, inclusivity: boolean]) -> T
+
+    Computes the value in the sketch that occurs at a particular quantile. The
+    third argument refers to the inclusivity of the query. This function returns
+    values strictly less than the quantile when inclusivity is false or values
+    less than or equal to the quantile when inclusivity true. If omitted, the
+    default inclusivity is true.
+
+.. function:: sketch_kll_rank[T](sketch: kllsketch[T], quantile: T[, inclusivity: boolean]) -> double
+
+    Computes the quantile that a particular value occurs at in the sketch. The
+    third argument refers to the inclusivity of the query. Given that the sketch
+    represents a distribution of data ``X``, this function returns quantiles
+    representing ``P(T < X)`` when inclusivity is false and ``P(T <= X)`` when
+    inclusivity is true. If omitted, the default inclusivity is true.
+
+
 .. _Apache DataSketches: https://datasketches.apache.org/
 .. _theta sketch: https://datasketches.apache.org/docs/Theta/ThetaSketchFramework.html
+.. _Theta sketch documentation: https://datasketches.apache.org/docs/Theta/ThetaSketchFramework.html
+.. _KLL Sketch documentation: https://datasketches.apache.org/docs/KLL/KLLSketch.html
--- a/presto-main/src/main/java/com/facebook/presto/operator/aggregation/sketch/kll/KllSketchAggregationFunction.java
+++ b/presto-main/src/main/java/com/facebook/presto/operator/aggregation/sketch/kll/KllSketchAggregationFunction.java
@@ -1,0 +1,78 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.operator.aggregation.sketch.kll;
+
+import com.facebook.presto.common.block.BlockBuilder;
+import com.facebook.presto.spi.function.AggregationFunction;
+import com.facebook.presto.spi.function.AggregationState;
+import com.facebook.presto.spi.function.CombineFunction;
+import com.facebook.presto.spi.function.InputFunction;
+import com.facebook.presto.spi.function.OutputFunction;
+import com.facebook.presto.spi.function.SqlType;
+import com.facebook.presto.spi.function.TypeParameter;
+import io.airlift.slice.Slice;
+
+@AggregationFunction(value = "sketch_kll")
+public class KllSketchAggregationFunction
+{
+    /**
+     * In the case the underlying library changes the default value for k, we use our own default
+     * here.
+     */
+    private static final int DEFAULT_K = 200;
+    private KllSketchAggregationFunction()
+    {
+    }
+
+    @InputFunction
+    @TypeParameter("T")
+    public static void input(@AggregationState KllSketchAggregationState state, @SqlType("T") long value)
+    {
+        KllSketchWithKAggregationFunction.input(state, value, DEFAULT_K);
+    }
+
+    @InputFunction
+    @TypeParameter("T")
+    public static void input(@AggregationState KllSketchAggregationState state, @SqlType("T") double value)
+    {
+        KllSketchWithKAggregationFunction.input(state, value, DEFAULT_K);
+    }
+
+    @InputFunction
+    @TypeParameter("T")
+    public static void input(@AggregationState KllSketchAggregationState state, @SqlType("T") Slice value)
+    {
+        KllSketchWithKAggregationFunction.input(state, value, DEFAULT_K);
+    }
+
+    @InputFunction
+    @TypeParameter("T")
+    public static void input(@AggregationState KllSketchAggregationState state, @SqlType("T") boolean value)
+    {
+        KllSketchWithKAggregationFunction.input(state, value, DEFAULT_K);
+    }
+
+    @CombineFunction
+    public static void combine(@AggregationState KllSketchAggregationState state, @AggregationState KllSketchAggregationState otherState)
+    {
+        KllSketchWithKAggregationFunction.combine(state, otherState);
+    }
+
+    @TypeParameter("T")
+    @OutputFunction("kllsketch(T)")
+    public static void output(@AggregationState KllSketchAggregationState state, BlockBuilder out)
+    {
+        KllSketchWithKAggregationFunction.output(state, out);
+    }
+}
--- a/presto-main/src/main/java/com/facebook/presto/operator/aggregation/sketch/kll/KllSketchAggregationState.java
+++ b/presto-main/src/main/java/com/facebook/presto/operator/aggregation/sketch/kll/KllSketchAggregationState.java
@@ -1,0 +1,226 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.operator.aggregation.sketch.kll;
+
+import com.facebook.presto.common.array.ObjectBigArray;
+import com.facebook.presto.common.type.Type;
+import com.facebook.presto.operator.aggregation.sketch.theta.ThetaSketchStateFactory;
+import com.facebook.presto.operator.aggregation.state.AbstractGroupedAccumulatorState;
+import com.facebook.presto.spi.PrestoException;
+import com.facebook.presto.spi.function.AccumulatorState;
+import com.facebook.presto.spi.function.AccumulatorStateMetadata;
+import com.google.common.collect.ImmutableMap;
+import io.airlift.slice.Slice;
+import org.apache.datasketches.common.ArrayOfBooleansSerDe;
+import org.apache.datasketches.common.ArrayOfDoublesSerDe;
+import org.apache.datasketches.common.ArrayOfItemsSerDe;
+import org.apache.datasketches.common.ArrayOfLongsSerDe;
+import org.apache.datasketches.common.ArrayOfStringsSerDe;
+import org.apache.datasketches.kll.KllItemsSketch;
+import org.openjdk.jol.info.ClassLayout;
+
+import javax.annotation.Nullable;
+
+import java.util.Comparator;
+import java.util.Map;
+
+import static com.facebook.presto.spi.StandardErrorCode.GENERIC_INTERNAL_ERROR;
+import static com.facebook.presto.spi.StandardErrorCode.INVALID_ARGUMENTS;
+import static java.util.Objects.requireNonNull;
+
+@AccumulatorStateMetadata(stateFactoryClass = KllSketchStateFactory.class, stateSerializerClass = KllSketchStateSerializer.class)
+public interface KllSketchAggregationState
+        extends AccumulatorState
+{
+    /**
+     * These parameters were created by running a linear regression against the current datasketches
+     * library version and the true in-memory size with k = 200. Upgrades to the datasketches
+     * library should attempt to re-compute these parameters.
+     */
+    Map<Class<?>, double[]> SIZE_ESTIMATOR_PARAMETERS = ImmutableMap.<Class<?>, double[]>builder()
+            .put(double.class, new double[] {929.0746716729514, 3.3328879969131457})
+            .put(long.class, new double[] {929.920850240635, 3.3325974295999283})
+            .put(Slice.class, new double[] {1008.5483794752541, 4.234749475485517})
+            .put(boolean.class, new double[] {345.60095530779705, 14.12193288555095})
+            .build();
+
+    long SKETCH_INSTANCE_SIZE = ClassLayout.parseClass(KllItemsSketch.class).instanceSize();
+
+    @Nullable
+    <T> KllItemsSketch<T> getSketch();
+
+    void addMemoryUsage(long value);
+
+    Type getType();
+
+    <T> void setSketch(KllItemsSketch<T> sketch);
+
+    class SketchParameters<T>
+    {
+        private final Comparator<T> comparator;
+        private final ArrayOfItemsSerDe<T> serde;
+
+        public SketchParameters(Comparator<T> comparator, ArrayOfItemsSerDe<T> serde)
+        {
+            this.comparator = comparator;
+            this.serde = serde;
+        }
+
+        public Comparator<T> getComparator()
+        {
+            return comparator;
+        }
+
+        public ArrayOfItemsSerDe<T> getSerde()
+        {
+            return serde;
+        }
+    }
+
+    class Single
+            implements KllSketchAggregationState
+    {
+        private static final long INSTANCE_SIZE = ClassLayout.parseClass(Single.class).instanceSize();
+        @Nullable
+        private KllItemsSketch sketch;
+        private final Type type;
+
+        public Single(Type type)
+        {
+            this.type = requireNonNull(type, "type is null");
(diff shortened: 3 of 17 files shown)


presto-docs/src/main/sphinx/functions/sketch.rst
```diff
@@ -28,5 +35,36 @@ DataSketches`_ library.
     the number of retained entries in the sketch.
 
 
+KLL Sketches
+------------
+
+Kll Sketches are an implementation of a quantiles sketch. For more information
+about the KLL sketch, see the Apache Datasketches `KLL Sketch documentation`_
```
steveburnett, line 42:
```suggestion
about the KLL sketch, see the Apache Datasketches `KLL Sketch documentation`_.
```

steveburnett (changes requested):
Only a very tiny nit, otherwise LGTM!

presto-main/src/main/java/com/facebook/presto/operator/aggregation/sketch/kll/KllSketchAggregationState.java
```diff
@@ -0,0 +1,98 @@
...
+ * limitations under the License.
+ */
+package com.facebook.presto.operator.aggregation.sketch.kll;
+
+import com.facebook.presto.common.array.ObjectBigArray;
+import com.facebook.presto.operator.aggregation.sketch.theta.ThetaSketchStateFactory;
+import com.facebook.presto.operator.aggregation.state.AbstractGroupedAccumulatorState;
+import com.facebook.presto.spi.function.AccumulatorState;
+import com.facebook.presto.spi.function.AccumulatorStateMetadata;
+import org.apache.datasketches.kll.KllItemsSketch;
```
Yuhta, line 21: There are some pitfalls (mainly memory-wise) we use KLL in presto_cpp, not sure if they are applicable to this Java implementation or not:
1. The sketch is initialized with a fairly large pre-allocated array depending on the error bound.  This would cause large memory usage when the number of groups are high.
2. The serialized form is not very compact and can occupy more memory than it needs.  This is ok for the partial aggregation side but would cause memory issue on the final aggregation side.

ZacBlanco: Thanks for bringing up these points! Was there anything you did to mitigate them?

re: 1. I haven't checked the code, but didn't consider the performance with a high number of groups since the main target use case for introducing the function is to provide table and partition-level statistics for the optimizer. If you have any thoughts on optimization I would happily discuss.

re: 2: I haven't measured the serialized size but according to the charts on the Apache Datasketches website, a "standard" k value of 200 produces sketches in the range of 2-4k bytes serialized with floats after consuming 1M+ items. Let's generously assume that with doubles, we would consume double the space, so about 8k.

kll size.

Do you know which parameters effect the number of partial aggregations sent to the final agg? I assumed it's the size of the cluster but, I could be wrong. Even if coalescing 1k KLL sketches in one node this should only consume 8k * 1000 = ~8MB of memory. I feel that shouldn't pose any issue. But it really depends on how many sketches are sent to the final agg step.

tdcmeehan: @Yuhta what sort of memory issues were observed?

Yuhta: For 1 it's easy to fix in the sketch itself, just need to make the sketch grows the bottom layer when the number of elements it keeps is small.  You may need to submit PR to `datasketches` project to fix it though, or copy the code over.

For 2 the size of serialization also depends on how the input data looks like.  The worst case comparing to Q/T-digest comes from a lot of identical inputs.  In this case Q-digest does not increase the memory usage but KLL does, because it keeps samples of inputs instead of distribution.  This multiplied by the number of groups and the fan-in factor of final aggregation sometimes blows it up.  One mitigation we did is to compact the sketch before the serialization, this does not resolve the problem completely but we have been good since then on production data.

Both issues can be fixed later though, I just mention it here so we are not surprised when these issues showing up.

presto-main/src/main/java/com/facebook/presto/operator/aggregation/sketch/kll/KllSketchAggregationState.java
```diff
@@ -0,0 +1,98 @@
...
+        @Override
+        public <T> void setSketch(KllItemsSketch<T> sketch)
+        {
+            sketches.set(getGroupId(), requireNonNull(sketch, "kll sketch is null"));
+        }
+
+        @Override
+        public long getEstimatedSize()
+        {
+            return sketches.sizeOf() + INSTANCE_SIZE;
```
tdcmeehan, line 89: The problem with this is it only estimates the size of the list, not the size of the memory used in aggregate.  It needs to be `SIZE_OF(KLLSketch) * sketches.sizeOf() + INSTANCE_SIZE`.

One way of doing this is to keep a running total in the grouped aggregation state, and increment/decrement it whenever sketches are added.

presto-main/src/main/java/com/facebook/presto/operator/aggregation/sketch/kll/KllSketchAggregationState.java
```diff
@@ -0,0 +1,98 @@
...
+        public <T> void setSketch(KllItemsSketch<T> sketch)
+        {
+            this.sketch = sketch;
+        }
+
+        @Override
+        public long getEstimatedSize()
+        {
+            if (sketch != null) {
+                return sketch.getSerializedSizeBytes() + INSTANCE_SIZE;
```
tdcmeehan, line 61: Is this the same thing as the size in memory?  If not, we need to use a more accurate method that captures the size in memory for accounting purposes.  If they're off, then such error can grow in magnitude for large group by or when the input is skewed.

tdcmeehan (changes requested):

presto-main/src/main/java/com/facebook/presto/operator/aggregation/sketch/kll/KllSketchAggregationState.java
ZacBlanco: I always assumed it was the serialized size. Docs on the interface would be nice...

Anyways, there isn't going to be a perfect way to calculate the size, especially for variable-width types as we discussed offline. For fixed-width types, I think a heuristic might be okay. I replaced this with the following function:

```java
        static final long SKETCH_INSTANCE_SIZE = ClassLayout.parseClass(KllItemsSketch.class).instanceSize();
        @Override
        public long getEstimatedSize()
        {
            if (sketch != null) {
                // Estimating the memory is difficult without direct access to
                // the underlying implementation. We will attempt to give a
                // close estimate without underestimating to prevent crashes.
                // Overestimates might claim more resources, and prevent more
                // queries from running.
                // Since retained entries generally dominate the memory usage.
                return (SKETCH_INSTANCE_SIZE +
                        INSTANCE_SIZE +
                        (long) (sketch.getNumRetained() * retainedItemSize * 1.2));
            }

            return INSTANCE_SIZE;
        }
```

I ran some rough benchmarks using JOL's `GraphLayout#parseInstance` which attempts to accurately calculate retained memory of a single object. I compared it against the above java function to test the accuracy. Here are the results:

image

image

At around 2048 items (2^11), the error percentage drops under 1 percent, which decreases up to 0.1% after the sketch has seen > 100k items.
I might be able to modify the heuristic to account for sketches which have seen a smaller number of item. from the stream. But I think the error is within a reasonable margin? What are your thoughts on the reasonable margin of error?

ZacBlanco: FWIW I also experimented against the T-Digest using a compression factor giving comparable sizes to the above chart:
image

The maximum error here is roughly 20%, while the minimum is ~0.27%

tdcmeehan: @ZacBlanco although this seems like a moderate risk, I am wary of doing something Presto-side since upgrades might be unstable and cause the heuristic to become out much more inaccurate.

Can you estimate how does the serialized size compare to JOL's estimation?  If they're close, then perhaps we could use that (plus a fudge factor), as it seems less likely that the data structure would change without a corresponding change to the serialized layout.  Additionally, you could update the grouped version to keep track one-by-one using the same heuristic.

ZacBlanco: Sorry for the delay on this one. I spent some time analyzing to come up with a decent heuristic that
we can use for the meantime based on the serialized size of the sketch.

While it's possible to make an addition to the sketch APIs to report utilized memory it's difficult
as they make heavy use of abstract classes which makes the calculation more complex than
most of the `ClassLayout#parseClass` instances in the Presto codebase. I started a small
implementation that might work for our case but proposing and implementing that API will probably
take some time to get accepted by the community.

While we wait for that change to land in the datasketches upstream, I ran a regression analysis to
determine the best adjustment factor to use. For each datatype, you can model the true in-memory
size, $t$ as $t(x) = f_1 x + f_2$ where $x$ is the size of the serialized sketch. I think this
actually models the true size quite well because when serialized, the size will be dominated by the
number of retained items. But those retained items take up less space serialized than when in the
JVM object always by a constant factor (e.g. an instance of `Double` takes up 24 bytes, but when
serialized just 8 bytes). Any other overhead in the sketch is constant (e.g. min/max, seen count, etc)
which is why we need the constant factor adjustment. This method should work for both fixed-width
and variable-sized datatypes since the serialized size actually contains the values which contribute
to memory consumption - unlike the approach with retained items I tried above.

Using this model I ran the regression to find the best parameters of $f_1$ and $f_2$ for each
datatype. Some charts are below. For strings and numeric types the regression function seems to
pretty closely track the true usage. For booleans there is a discrepancy where the regression over-
estimates by a bit.

image

These were the parameters I found:

| Type | $f_1$ | $f_2$ |
|-----|-----|----|
| Double |  929.0746716729514 | 3.3328879969131457 |
| Long | 1008.5483794752541 | 4.234749475485517 |
| String | 345.60095530779705 | 14.12193288555095
| Boolean | 929.920850240635 | 3.3325974295999283 |

I would probably encode them into a static hashmap that gets referenced when querying the estimated size.

The ideal solution is for sure to get something upstream, but if we really care about getting accurate sizes for grouped queries, then this will be a pretty good stopgap solution with minimal error until then. The accuracy issues with booleans I am not very concerned about because there isn't really a great use case for booleans and KLL sketches anyways.

tdcmeehan: Thank you for working on this detailed analysis.  I agree that your heuristic seems to be a good stopgap for now.

ZacBlanco: The code has been updated with this approach

presto-main/src/main/java/com/facebook/presto/operator/aggregation/sketch/kll/KllSketchAggregationState.java
ZacBlanco: Fixed

steveburnett (approved):
LGTM! (docs)

New pull of branch and new local build of docs.

presto-main/src/test/java/com/facebook/presto/operator/aggregation/TestKllSketchStateSerializer.java
```diff
@@ -0,0 +1,158 @@
...
+        verifySketches(actual, state.getSketch());
+    }
+
+    private static void verifySketches(KllItemsSketch<?> expected, KllItemsSketch<?> actual)
+    {
+        for (double i = 0; i < 10; i++) {
+            double q = i / 10;
+            assertEquals(actual.getQuantile(q), expected.getQuantile(q));
+        }
+    }
```
aaneja, line 157: Can we instead use `assertEquals(actual.toString(true,true), expected.toString(true,true));` ? Would this be a stronger or weaker assertion ?

presto-main/src/test/java/com/facebook/presto/operator/aggregation/TestKllSketchStateSerializer.java
```diff
@@ -0,0 +1,158 @@
...
+        KllItemsSketch actual = sketchSupplier.get();
+        sketchInputs.stream().forEach(i -> {
+            sketch.update(i);
+            actual.update(i);
+        });
+        blockBuilder = VARBINARY.createBlockBuilder(null, 1);
+        state.setSketch(sketch);
+        serializer.serialize(state, blockBuilder);
+        Block serializedBlock = blockBuilder.build();
+        serializer.deserialize(serializedBlock, 0, state);
```
aaneja, line 147: Can you deserialize into a new empty state object instead of reusing this `state` object ?

aaneja (changes requested):

presto-main/src/main/java/com/facebook/presto/operator/aggregation/sketch/kll/KllSketchWithKAggregationFunction.java
```diff
@@ -0,0 +1,129 @@
...
+            out.appendNull();
+            return;
+        }
+        VARBINARY.writeSlice(out, Slices.wrappedBuffer(state.getSketch().toByteArray()));
+    }
+
+    private static <T> void initializeSketch(KllSketchAggregationState state, Supplier<Comparator<T>> comparator, Supplier<ArrayOfItemsSerDe<T>> serdeSupplier, long k)
+    {
+        if (k < 8 || k > 65535) {
+            throw new PrestoException(INVALID_ARGUMENTS, "k value must satisfy 8 <= k <= 65525: " + k);
```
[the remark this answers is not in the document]
aaneja: Also: Do we need to hardcode these min/max `k` constants ? Should we instead use  `DEFAULT_M` and `MAX_K` from the `KllSketch` class ?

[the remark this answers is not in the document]
ZacBlanco: `DEFAULT_M` isn't exposed publicly, but I will refer to the `MAX_K` argument

ClarenceThreepwood (dismissed):
lgtm. Just nits from my side
Good work on the memory accounting!

presto-docs/src/main/sphinx/functions/sketch.rst
```diff
@@ -28,5 +35,36 @@ DataSketches`_ library.
     the number of retained entries in the sketch.
 
 
+KLL Sketches
```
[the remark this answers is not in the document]
steveburnett: I noticed new content had been added addressing this, so I pulled the branch and made a local build of the docs just now. Here's a screenshot of the newly-added `sketch_kll_with_k` entry:
Screenshot 2024-01-24 at 9 36 22 AM

presto-main/src/test/java/com/facebook/presto/operator/aggregation/TestKllSketchStateSerializer.java
aaneja: How about `assertEquals(actual.toByteArray(), expected.toByteArray());` ? This works with the current implementation and asserts that the conversion via the new classes is lossless ?

aaneja (approved):

presto-main/src/main/java/com/facebook/presto/operator/aggregation/sketch/kll/KllSketchAggregationFunction.java
```diff
@@ -0,0 +1,78 @@
...
+import io.airlift.slice.Slice;
+
+@AggregationFunction(value = "sketch_kll")
+public class KllSketchAggregationFunction
+{
+    /**
+     * In the case the underlying library changes the default value for k, we use our own default
+     * here.
+     */
+    private static final int DEFAULT_K = 200;
```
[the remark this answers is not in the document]
ZacBlanco: I could see a case where it might be desirable to have it be configurable, but that's why we have the `sketch_kll_with_k` function so users can pass any value they want. I put this here to (1) extract constant value from all of the call sites and (2) provide a stable default in the case the library changes the default after an upgrade. A different value will affect the storage size of the sketch, so I wanted something that would be stable within the Presto code base.

presto-main/src/main/java/com/facebook/presto/operator/aggregation/sketch/kll/KllSketchStateSerializer.java
```diff
@@ -0,0 +1,72 @@
...
+            state.setSketch(null);
+            return;
+        }
+        Memory memory = Memory.wrap(VARBINARY.getSlice(block, index).toByteBuffer(), LITTLE_ENDIAN);
+        KllSketchAggregationState.SketchParameters parameters = KllSketchAggregationState.getSketchParameters(type);
+        // use heapify over wrap in order to get a writable sketch for updates and merges
+        KllItemsSketch sketch = KllItemsSketch.heapify(memory, parameters.getComparator(), parameters.getSerde());
+        state.addMemoryUsage(-getEstimatedKllInMemorySize(state.getSketch(), type.getJavaType()));
+        state.setSketch(sketch);
+        state.addMemoryUsage(getEstimatedKllInMemorySize(state.getSketch(), type.getJavaType()));
```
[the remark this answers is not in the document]
ZacBlanco: They might not - since `state` is mutable, Line 67 creates a new sketch which may have a different size than the result of `state.getSketch()` at that time. So, we need to subtract the current state (Line 68), set the new state (line 69) and re-calculate the size (Line 70)

presto-main/src/test/java/com/facebook/presto/operator/aggregation/TestKllSketchStateSerializer.java
ZacBlanco: I just tested this, but received failures. The byte arrays are not equivalent, but still produce the same results. I changed the equality check to use the same fields as in the library method instead.

ZacBlanco: Ah I realized I was using the wrong test method beforehand. I checked again with the `toByteArray` methods on this test and it works. I will update it.

tdcmeehan:
I think we need to document the serialized binary format, similar to T-Digest and QDigest.  Otherwise, they're completely unreadable by any other system.

ZacBlanco:
> I think we need to document the serialized binary format, similar to T-Digest and QDigest.

The serialized binary format is specified by the Apache Datasketches library. This is already noted in the docs changes.

https://github.com/prestodb/presto/pull/21568/files#diff-33bc3be2c487c146b213ef7b64ddc7c91956aa6e8b5d6207b217061312286b37R47-R48

If this is confusing I can try to update the phrasing

tdcmeehan:
I can't find where they explicitly document the binary format--is there a link for that?

ZacBlanco:
The binary format is not documented unfortunately. However, all versions (cpp+java) of the datasketches libraries can serialize and deserialize the sketches. e.g. java can deserialize cpp serialized sketches, and vice versa. AFAIK this is the case for all sketches coming from the datasketches library.

presto-main/src/main/java/com/facebook/presto/operator/aggregation/sketch/kll/KllSketchAggregationState.java
```diff
@@ -0,0 +1,226 @@
...
+@AccumulatorStateMetadata(stateFactoryClass = KllSketchStateFactory.class, stateSerializerClass = KllSketchStateSerializer.class)
+public interface KllSketchAggregationState
+        extends AccumulatorState
+{
+    /**
+     * These parameters were created by running a linear regression against the current datasketches
+     * library version and the true in-memory size with k = 200. Upgrades to the datasketches
+     * library should attempt to re-compute these parameters.
+     */
+    Map<Class<?>, double[]> SIZE_ESTIMATOR_PARAMETERS = ImmutableMap.<Class<?>, double[]>builder()
```
tdcmeehan, line 51: This entire class seems to be missing visibility modifiers. Could we add them?

tdcmeehan (commented):
Just one last thing

presto-main/src/main/java/com/facebook/presto/operator/aggregation/sketch/kll/KllSketchAggregationState.java
ZacBlanco: Because `KllSketchAggregationState` is an interface, modifiers on fields are redundant (`static` by default) and `private/public` are not allowed. Methods and inner classes are also public by default, hence adding would be redundant.

tdcmeehan: Oh, missed that!

tdcmeehan (approved):

ZacBlanco:
Just for some continuity on the discussion of sketch serialized format compatibility, we brought this to their community in hopes to clarify the guarantees that the maintainers provide. This resulted in some changes to the datasketches documentation in https://github.com/apache/datasketches-website/pull/162 and subsequently https://github.com/apache/datasketches-website/pull/163

tdcmeehan:
Possible follow up: if KLL has a provable error bound, could we use it to replace approx_distinct?  The only reason approx_distinct uses Q-Digest under the hood, in spite of its poor performance compared to T-Digest, is it has a provable error bound.

ZacBlanco:
I'm not familiar with the code for `approx_distinct`, but if the KLL sketch answers the same question as the Q-digest (e.g. what percentile does value X fall? or the inverse), then we should be able to replace it. Another option is you could replace the `approx_distinct` implementation entirely with a Theta sketch. I can't speak how the accuracy of a Theta sketch compares to the current `approx_distinct` implementation though.

This could be a good project/experiment for someone new to Presto

jmalkin:
The binary format is defined here: https://github.com/apache/datasketches-java/blob/master/src/main/java/org/apache/datasketches/kll/KllPreambleUtil.java

Obviously the format for items cannot be defined for arbitrary type T, and the exact size of the levels array is not fixed. But the serialized header is defined, as is the order of the non-fixed-size/format data.

leerho:
@Yuhta,
I don't understand your comment above (Jan 4th) "The serialized form is not very compact and can occupy more memory than it needs."

The KLL sketch always serializes (`toByteArray()`) into a compact form, which has no wasted space,  and grows very gradually based on the number of items fed to and retained by the sketch.  You can see the effect of this in the plot copied from our website above.  The blue curve starts very small (8 bytes) and grows gradually until it reaches ~2.5K bytes at about 1M items.

It is true that if the sketch image were compressed it would become even smaller, but that can be costly in terms of time.  That is not an option we currently offer as part of the sketch, but as long as one is using lossless compression algorithms, it is something you could do yourself. 

If you could elaborate you concern a bit more it would help us understand the issue.

Thanks!

Issue #20494, opened 2023-08-04
Iceberg MoR support - writes


tdcmeehan:
Currently, the Presto Iceberg connector doesn't support MoR tables.

This is the primary item added in Iceberg V2.  After prestodb/presto#20492, we can add support for the generation of delete list files and their addition in manifest files.  This is prioritized differently, because in Lakehouse environments one may use two different systems for production and reading of data.

[See the Iceberg spec for more details.](https://iceberg.apache.org/spec/#version-2)

PR #21571, opened 2023-12-19
Iceberg MOR support - implement row-level deletes

s-akhtar-baig:
## Description

Support row-level delete operations for Iceberg V2 tables. Use merge-on-read mode by default and create positional delete files. 

## Motivation and Context

GitHub issue: https://github.com/prestodb/presto/issues/20494

## Impact

Delete operations will succeed on partitioning columns as well as non-partitioning columns. If delete mode is set to "copy-on-write" and if filtered on non-partitioning columns, then the following error will be displayed as of this moment.

_This connector only supports delete where one or more partitions are deleted entirely_

## Test Plan

- [x] Added tests for delete operations in both modes. 
- [x] Updated existing tests to account for the new changes.

## Release Notes

```
== RELEASE NOTES ==

Iceberg Connector Changes
* Added support for row-level deletes on Iceberg V2 tables.
The delete mode can be changed from ``merge-on-read`` to ``copy-on-write`` by setting table property ``delete_mode``.
```


presto-iceberg/src/main/java/com/facebook/presto/iceberg/delete/IcebergDeletePageSink.java
```diff
@@ -0,0 +1,220 @@
...
+        RuntimeException error = null;
+        try {
+            positionDeleteWriter.getWriter().rollback();
+        }
+        catch (Throwable t) {
+            if (error == null) {
+                error = new RuntimeException("Exception during rollback");
+            }
+            throw error;
+        }
```
tdcmeehan, line 176: I believe this is the same

```suggestion
            positionDeleteWriter.getWriter().rollback();
```

s-akhtar-baig: The error message will be either "_Error rolling back write parquet to Hive_" or "_Error rolling back write to Hive_" for Parquet and Orc file writers respectively. Would that be okay?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/PartitionData.java
```diff
@@ -93,6 +93,26 @@ public String toJson()
         }
     }
 
+    public static PartitionData fromStructLike(StructLike partitionData, Type[] types)
```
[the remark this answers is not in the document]
s-akhtar-baig: Oh yes, this method needs refactoring. I have made changes, let me know how it looks.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/delete/IcebergDeletePageSink.java
tdcmeehan: You can create a new error code in IcebergErrorCode, perhaps `ICEBERG_ROLLBACK_ERROR`.  Do not throw RuntimeException since it's unclassified, throw a new PrestoException with the new error code.

steveburnett (approved):
LGTM! (docs)

Local build of the docs from this branch, everything looks fine. Thanks!

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergTableProperties.java
```diff
@@ -114,4 +125,9 @@ public static String getFormatVersion(Map<String, Object> tableProperties)
     {
         return (String) tableProperties.get(FORMAT_VERSION);
     }
+
+    public static String getDeleteMode(Map<String, Object> tableProperties)
+    {
+        return (String) tableProperties.get(DELETE_MODE);
+    }
```
hantangwangd, line 132: Seems table property `delete_mode ` has no use. The same as above, maybe we should firstly figure out whether the `delete-mode` be a connector property or a table property.

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -216,6 +216,9 @@ Property Name                                      Description
 
 ``iceberg.enable-merge-on-read-mode``              Enable reading base tables that use merge-on-read for          ``true``
                                                    updates.
+
+``iceberg.delete-mode``                            The delete mode for Iceberg V2 tables. The available values    ``merge-on-read``
+                                                   are ``copy-on-write`` and ``merge-on-read``.
```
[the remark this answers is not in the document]
s-akhtar-baig: @hantangwangd, thank you for reviewing these changes! 

My thought process was to allow users to switch between these two modes by setting the connector property. Although data rewriting is not implemented at the moment, I am assuming we want to support it in the near future. 

Yes, I am good with either keeping both or just one, considering the table property is not being used yet. 

@tdcmeehan, thoughts on this?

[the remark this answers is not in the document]
steveburnett: I ask that whatever decision (both or just one), that the properties are documented. 

If the decision is made to keep just one, the release note entry should be updated.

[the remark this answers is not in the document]
tdcmeehan: I think we should be consistent with the rest of the Iceberg ecosystem, which treat this as a table property.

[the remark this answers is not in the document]
s-akhtar-baig: That makes sense! And should we keep or remove the connector property?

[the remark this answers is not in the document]
hantangwangd: Agree with @tdcmeehan. This may be clearer and without confusion for users, especially the ones who are already working on iceberg.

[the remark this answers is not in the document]
tdcmeehan: @s-akhtar-baig let's remove it

yingsu00 (commented):
@s-akhtar-baig @tdcmeehan Overall looks good, I have a couple of questions though.
1. I see that this PR is using the UpdatablePageSource approach. Since the DeleteOperator can only set an UpdatablePageSource as its pageSource, and only TableScanOperator and ScanFilterAndProjectOperator can actually return an UpdatablePageSource object, it means that the DeleteOperator can only be in the same stage with the Scan operator that reads the target table with rows to be deleted. It cannot work with ExchangeOperator, who is also a SourceOperator, because it's addSplit() doesn't return an actual object but just `nullopt`. This means the SemiJoin to join the target table with the delete values has to be broadcast join. What will happen if both sides exceed the broadcast join limit data size? Will the query force broadcast join, or will the query fail?
2. Trino has removed the UpdatablePageSource in [Remove legacy code for DELETE and UPDATE](https://github.com/trinodb/trino/pull/15161/files#top). What's your opinion on their change? Any plans to adopt similar routes?

Btw. I attach a Presto plan and a Trino plan just for your conveniences.
Presto

image

Trino
image

hantangwangd:
> What will happen if both sides exceed the broadcast join limit data size? Will the query force broadcast join, or will the query fail?

Seems semi join for delete has been fixed to broadcast base on the optimizer `ReplicateSemiJoinInDelete`.

yingsu00:
> Seems semi join for delete has been fixed to broadcast base on the optimizer `ReplicateSemiJoinInDelete`.

Thanks @hantangwangd for your information. It seems this is a limitation for this approach. Do you know why ExchangeOperator cannot be used as the SourceOperator here so it can allow hash joins which might be beneficial for larger data sets? Also, this UpdatablePageSource way for deletes/udpates and having a delete sink in a PageSource seems a bit unusual to me. Do you folks want to consider the MergeWriter approach Trino has?

hantangwangd:
> Do you know why ExchangeOperator cannot be used as the SourceOperator here so it can allow hash joins which might be beneficial for larger data sets? Also, this UpdatablePageSource way for deletes/udpates and having a delete sink in a PageSource seems a bit unusual to me. 

I think this approach has an implicit constraint between a sourceOperator and a deleteOperator in the same driver, that is they would handle the same data file, reading from and writing delete file for it. And currently an exchangeOperator could not follow this constraint unless further customization. Maybe that is the reason for this design.

yingsu00:
> I think this approach has an implicit constraint between a sourceOperator and a deleteOperator in the same driver, that is they would handle the same data file, reading from and writing delete file for it. And currently an exchangeOperator could not follow this constraint unless further customization. Maybe that is the reason for this design.

@hantangwangd Thanks for the discussion. But why do the delete file writes have to be done in a SourceOperator, and specifically, the ScanFilterAndProject and TableScan? Can it be done just in DeleteOperator? For MOR delete, the base files are not updated at all, and the delete files are just a bunch of new files that can be written by the DeleteOperator. Even for COW delete, it doesn't have to be the scan operators that write the updated data file, as long as the updated data to be written is passed to the DeleteOperator. The data coming out of the SemiJoin can just be that. The difference is that MOR delete would produce the rows to be deleted, while COW delete produce the rows that are not deleted. In neither case, the scan doesn't have to be in the same stage as the delete.

Taking one step back, even if we don't write the delete files in DeleteOperator but still in a SourceOperator, why can't it be any SourceOperator like ExchangeOperator? As long as it pulls the data to be written from its upstream, it should be fine. They are not operating on the same data file in either MOR or COW. Why was there the constraint they have to be in the same fragment?

s-akhtar-baig:
@yingsu00, thank you for reviewing and @hantangwangd, thank you for providing details! Much appreciated!

Regarding the delete workflow in this pull request, I tried to follow Presto’s current implementation of delete and update as described at (https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/develop/delete-and-update.rst). 

As you mentioned, there are constraints with this way of implementation, one-to-one association of a data file to a delete file being one of the big ones. This presents future work opportunities including coalescing delete files and writing equality deletes on the coordinator node.   

Please let me know if I am missing anything. But, there is definitely an opportunity to optimize delete and merge operations. I propose we can create a design document and start curating an optimized workflow.

yingsu00:
@hantangwangd 

> In order to remove the constraint mentioned above, maybe we need to add a column about file path besides of rowId in the output page of ConnectorPageSourceWithRowPositions.

Agreed, I think that will work. It's going to be a RLEBlock and will take very small space. Same for the partition information the DeleteOperator needs.

> As I understand, SourceOperator just build and return a factory for DeleteOperator to create the UpdatablePageSource, which contains a supplier deleteSinkSupplier. The IcebergDeletePageSink would be created in deleteOperator and then the delete file writes would be done in deleteOperator. That's all about lazy load and parameter closure. Of course I agree that things don't have to be this way.

Yes the DeleteOperator initiate the delete by calling `pageSource().deleteRows(rowIds)`, but the `deleteSink` and `deleteSinkSupplier` are owned by the `IcebergUpdateablePageSource`, which actually write the delete files by the code within IcebergUpdateablePageSource:
```
@Override
    public void deleteRows(Block rowIds)
    {
        if (deleteSink == null) {
            deleteSink = deleteSinkSupplier.get();
        }
        deleteSink.appendPage(new Page(rowIds));
    }
```
Do the deleteSink and deleteSinkSupplier have to be in the PageSource? We both agree it's not the case. All the information can be acquired in the DeleteOperator, if it could be passed to the DeleteOperator from the SourceOperator. But a PageSource owning a sink, and a loop in the Driver pipeline seems a bit strange to me. 

By lazy loading, did you mean the information like the file path, partition spec, etc that are needed in addition to the rowIds could be lazily loaded? I think they can still be lazily loaded in the DeleteOperator even if we remove the driver loop and make the DeleteOperator own the deleteSink, but they have the same value for for every deleted rows(base file path, partition spec are the same for all rows in the same split), so there's no point or benefit to lazily load them because they're already there. They would just be some RLEBlocks and carrying them from the SourceOperator to the DeleteOperator doesn't incur much cost at all. Note that the rowIds and the delete keys cannot be lazily loaded anyways, since the SemiJoin before the DeleteOperator would need to load the values in order to perform the join. Other than these, I'm not sure what else could be lazily loaded. Maybe you refer to something else?

Of course, I'm not asking to make these changes in this PR, but I wonder if somebody can try this approach in the near future. Trino has removed the UpdatablePageSource, and there must be some valid reasons behind it. Maybe the issues we discussed above are some of their reasons too. I think the approach that doesn't introduce a sink in a source, or a loop in the driver pipeline is cleaner. What do you think?

cc @agrawalreetika

yingsu00:
hi @s-akhtar-baig, yes I know you were following the current framework. As I said, I'm not asking you to make these changes in this PR, but I wonder if somebody can do some through study and try this approach in the near future. We can merge this PR first so that we can at least support the positional delete and make improvements in the future.

> This presents future work opportunities including coalescing delete files and writing equality deletes on the coordinator node.   

Equality deletes MAY be written on the coordinator if the delete values are known at the planning time, but I would consider it as an improvement to the general implementation. In the general implementation, we will just remove the join that joins the delete values with the target table, and then the scan on the target table can be saved as well. I expect this to be much faster than the positional deletes, especially when the target table is very large. The read afterwards with the equality delete files would also run faster because many equality deletes can be interpreted as TupleDomain filters and pushed down to the readers in scan.

hantangwangd:
@yingsu00 
> Do the deleteSink and deleteSinkSupplier have to be in the PageSource? We both agree it's not the case. All the information can be acquired in the DeleteOperator, if it could be passed to the DeleteOperator from the SourceOperator. But a PageSource owning a sink, and a loop in the Driver pipeline seems a bit strange to me.

Yes, I agree with you, things do not have to be this way. Another way, for example, only passing the closure for creation logic of deleteSink and let DeleteOperator really create and hold deleteSink might be a more nature way.

As for lazy loading, I meant that in this approach the `IcebergUpdateablePageSource` would not really be created until the driver begin running, that is, `TableScanOperator.getOutput()` be invoked. So during operators construction, they have to pass a closure(or say factory). And as you say, that's also not the reason why sourceOperator must hold the deleteSink.

So overall, I agree with you that, if we further unbind the constraint, and clearly define the message to transfer, we could make the things you concerned more cleaner, and support an `exchangeOperator` to be source of `deleteOperator`.

presto-docs/src/main/sphinx/connector/iceberg.rst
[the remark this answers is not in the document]
s-akhtar-baig: I have made changes to use ``delete_mode`` as iceberg table property. I have also updated the document and release notes. fyi @hantangwangd @steveburnett @tdcmeehan. 

Please let me know if I missed anything. Thanks!

yingsu00:
@hantangwangd 

> Yes, I agree with you, things do not have to be this way. Another way, for example, only passing the closure for creation logic of deleteSink and let DeleteOperator really create and hold deleteSink might be a more nature way.

Yeah that's exactly what I meant. But by "closure", could you explain a bit more? In Trino, the info was sent as a single field of ROW type: 
image. I didn't check in detail, and I think the information needed by the DeleteOperator to create the `deleteSink` can be either passed from the driver pipeline, or from the query contexts.

> As for lazy loading, I meant that in this approach the IcebergUpdateablePageSource would not really be created until the driver begin running, that is, TableScanOperator.getOutput() be invoked. So during operators construction, they have to pass a closure(or say factory). And as you say, that's also not the reason why sourceOperator must hold the deleteSink.

As far as I can see, the HivePageSource and other Hive related connector's PageSource were all created in `TableScanOperator.getOutput()`, even for the original IcebergPageSource. It seems to me there is not much value of this "lazy" creation, because the PageSource needs to be created anyways and it's almost immediately when the driver starts to run. The `deleteSink` was created lazily though, as it's initiated by the DeleteOperator through IcebergUpdateablePageSource and it's after the SemiJoin starts to output pages. If the Semijoin doesn't produce anything, then the `deleteSink` doesn't need to be created. But we can make it lazily created anyways even if it's created and owned by the DeleteOperator itself.

hantangwangd:
@yingsu00 Thanks for providing the info format in trino. Seems its implementation approach is similar to what we discussed above.

Sorry for making the description more complicated above. The "closure" I mentioned could be simply thought of as a factory, which encloses parameters and the creation logic. It would be something like:
```
() -> new IcebergDeletePageSink(......);
```
Other than:
```
() -> {
	if (deleteSink == null) {
        deleteSink = new IcebergDeletePageSink(......);
    }
    return deleteSink;
}
```
So that `DeleteOperator` could use this factory to create and hold a deleteSink. And it is just one of the feasible implementation options to use here. Of course the message for creating `deleteSink` could also be passed by other ways as you mentioned.

And as to the 'lazy creation', what I meant is passing a creation factory as above is a pretty good way(following the current framework), as the messages for creating `IcebergDeletePageSink` would always be set before `DeleteOperator` try to actually create it. And again its just one of the feasible implementation options here.

And to put it further, when we try to support `exchangeOperator` as a sourceOperator for `deleteOperator`, we could not simply transform the message like this. For sourceOperators as tableScanOperator might live in different stages with deleteOperators. So maybe we should also encode the message into the data pages as trino does.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
```diff
@@ -689,7 +699,8 @@ public IcebergTableHandle getTableHandle(ConnectorSession session, SchemaTableNa
                 })
                 .orElseGet(() -> resolveSnapshotIdByName(table, name));
 
-        // Get Iceberg tables schema with missing filesystem metadata will fail.
+        // Get Iceberg tables schema, spec, properties, and location with missing
```
hantangwangd, line 702: nits: no `spec`

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
```diff
@@ -742,10 +755,65 @@ public void truncateTable(ConnectorSession session, ConnectorTableHandle tableHa
     public ConnectorTableHandle beginDelete(ConnectorSession session, ConnectorTableHandle tableHandle)
     {
         IcebergTableHandle handle = (IcebergTableHandle) tableHandle;
+
         if (handle.isSnapshotSpecified()) {
             throw new PrestoException(NOT_SUPPORTED, "This connector do not allow delete data at specified snapshot");
         }
-        throw new PrestoException(NOT_SUPPORTED, "This connector only supports delete where one or more partitions are deleted entirely");
+        if (getDeleteMode(getTableMetadata(session, handle).getProperties()) == RowLevelOperationMode.COPY_ON_WRITE) {
```
hantangwangd, line 762: Didn't handle already have the properties? Could we use it directly?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
```diff
@@ -742,10 +755,65 @@ public void truncateTable(ConnectorSession session, ConnectorTableHandle tableHa
...
+
+        Type[] partitionColumnTypes = icebergTable.spec().fields().stream()
+                .map(field -> field.transform().getResultType(
+                        icebergTable.schema().findType(field.sourceId())))
+                .toArray(Type[]::new);
+
+        CharSequenceSet referencedDataFiles = CharSequenceSet.empty();
+
+        for (CommitTaskData task : commitTasks) {
+            FileMetadata.Builder builder = FileMetadata.deleteFileBuilder(icebergTable.spec())
```
hantangwangd, line 793: The commitTask may use a historical spec version, so could we simply use `icebergTable.spec()` here?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/delete/IcebergDeletePageSink.java
```diff
@@ -0,0 +1,213 @@
...
+    @Override
+    public long getCompletedBytes()
+    {
+        return writtenBytes;
+    }
+
+    @Override
+    public long getSystemMemoryUsage()
+    {
+        return systemMemoryUsage;
```
hantangwangd, line 120: Seems `systemMemoryUsage ` is not set anywhere?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/equalitydeletes/EqualityDeletesSplitSource.java
```diff
@@ -107,6 +108,8 @@ private IcebergSplit splitFromDeleteFile(DeleteFile deleteFile)
                 deleteFile.format(),
                 ImmutableList.of(),
                 getPartitionKeys(specById.get(deleteFile.specId()), deleteFile.partition()),
+                PartitionSpecParser.toJson(PartitionSpec.unpartitioned()),
+                Optional.empty(),
```
hantangwangd, line 112: Is that ok? As `partitionSpec` and `partitionData` are all available, may be we can set them properly.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedSmokeTestBase.java
```diff
@@ -1233,4 +1242,80 @@ public void testUnregisterTable()
...
+        @Language("RegExp") String errorMessage = "This connector only supports delete where one or more partitions are deleted entirely";
+
+        assertUpdate("CREATE TABLE " + tableNameCor + " (id integer, value integer) WITH (format_version = '2', delete_mode = 'copy-on-write')");
+        assertUpdate("INSERT INTO " + tableNameCor + " VALUES (1, 5)", 1);
+        assertQuery("SELECT * FROM " + tableNameCor, "VALUES (1, 5)");
+        assertQueryFails("DELETE FROM " + tableNameCor + " WHERE value = 5", errorMessage);
+
+        dropTable(session, tableNameCor);
+    }
+
```
hantangwangd, line 1275:
```suggestion
@Test
    public void testDeleteOnTableWithPartitionEvolution()
    {
        String tableName = "test_delete_on_table_with_partition_evolution";
        Session session = getSession();

        assertUpdate("CREATE TABLE " + tableName + " (a integer, b varchar) WITH (format_version = '2')");
        assertUpdate("INSERT INTO " + tableName + " VALUES (1, '1001'), (2, '1002'), (3, '1003')", 3);
        assertQuery("SELECT * FROM " + tableName, "VALUES (1, '1001'), (2, '1002'), (3, '1003')");

        assertUpdate("DELETE FROM " + tableName + " WHERE b = '1001'", 1);
        assertQuery("SELECT * FROM " + tableName, "VALUES (2, '1002'), (3, '1003')");

        assertUpdate("ALTER TABLE " + tableName + " ADD COLUMN c integer WITH (partitioning = 'identity')");
        assertUpdate("INSERT INTO " + tableName + " VALUES (1, '1001', 1), (2, '1002', 1), (3, '1003', 2)", 3);
        assertQuery("SELECT * FROM " + tableName, "VALUES (2, '1002', NULL), (3, '1003', NULL), (1, '1001', 1), (2, '1002', 1), (3, '1003', 2)");

        assertUpdate("DELETE FROM " + tableName + " WHERE b = '1002'", 2);
        assertQuery("SELECT * FROM " + tableName, "VALUES (3, '1003', NULL), (1, '1001', 1), (3, '1003', 2)");

        assertUpdate("ALTER TABLE " + tableName + " ADD COLUMN d integer WITH (partitioning = 'identity')");
        assertUpdate("INSERT INTO " + tableName + " VALUES (1, '1001', 1, 1), (2, '1002', 1, 2), (3, '1003', 2, 1)", 3);

        assertUpdate("DELETE FROM " + tableName + " WHERE b = '1003'", 3);
        assertQuery("SELECT * FROM " + tableName, "VALUES (1, '1001', 1, NULL), (1, '1001', 1, 1), (2, '1002', 1, 2)");

        dropTable(session, tableName);
    }
```

This test case would fail, the reason might be the partition spec version incompatible mentioned above in `IcebergAbstractMetadata.finishDelete()`.

hantangwangd (changes requested):
Thanks for the fix. Overall looks good to me, only one problem and some nits.Please check and fix the test case I have provided.

yingsu00:
> For sourceOperators as tableScanOperator might live in different stages with deleteOperators. So maybe we should also encode the message into the data pages as trino does.

Agree. @hantangwangd @s-akhtar-baig Are you folks going to continue working on this?

hantangwangd:
> Are you folks going to continue working on this?

Hi @yingsu00 , currently I am working on trying to support grouped execution in Iceberg as described in #20420, which seems a little complex. And after that, if @s-akhtar-baig do not have enough time, of cause I can working on this.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
s-akhtar-baig: @hantangwangd, thank you for pointing these out! 

I have handled your review comments. Please take a look when you get a chance, thanks!

s-akhtar-baig:
@yingsu00, I am working on other items and may not get to this anytime soon. @hantangwangd, please feel free to work on it and let me know if I can be helpful in anyway shape or form. 

Thank you both for your feedback and sharing your insights!

presto-iceberg/src/main/java/com/facebook/presto/iceberg/CommitTaskData.java
```diff
@@ -25,19 +26,28 @@ public class CommitTaskData
     private final String path;
     private final long fileSizeInBytes;
     private final MetricsWrapper metrics;
+    private final String partitionSpecJson;
```
hantangwangd, line 29: I think `specId` would be enough, what's your opinion?

hantangwangd (approved):
LGTM! Thank your for the great work. One last nit.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/CommitTaskData.java
s-akhtar-baig: That made sense to me and I have made changes accordingly. Thanks @hantangwangd!

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -787,7 +791,7 @@ dropping the table from the metadata catalog using ``TRUNCATE TABLE``.
 DELETE
 ^^^^^^^^
 
-The Iceberg connector can delete data in one or more entire partitions from tables by using ``DELETE FROM``. For example, to delete from the table ``lineitem``::
```
tdcmeehan, line 790: Should we preserve the older version of the documentation for V1 tables?

presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedSmokeTestBase.java
```diff
@@ -1233,4 +1242,108 @@ public void testUnregisterTable()
...
+        assertUpdate("DELETE FROM " + tableNameMor + " WHERE id = 1", 1);
+        assertQuery("SELECT * FROM " + tableNameMor, "VALUES (2, 13), (3, 5), (4, 13), (5, 1), (6, 1)");
+        assertUpdate("DELETE FROM " + tableNameMor + " WHERE value = 13", 2);
+        assertQuery("SELECT * FROM " + tableNameMor, "VALUES (3, 5), (5, 1), (6, 1)");
+
+        dropTable(session, tableNameMor);
+
+        // Test with delete_mode set to copy-on-write
+        String tableNameCor = "test_delete_partitioned_cor";
+        @Language("RegExp") String errorMessage = "This connector only supports delete where one or more partitions are deleted entirely";
```
tdcmeehan, line 1305: Should the error message be changed to reflect that actually this is now configurable by a table property?

tdcmeehan (approved):
LGTM.  Any final comments @beinan @ChunxuTang @yingsu00

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -254,6 +254,10 @@ Property Name                             Description
 ``format_version``                         Optionally specifies the format version of the Iceberg
                                            specification to use for new tables, either ``1`` or ``2``.
                                            Defaults to ``1``.
+
+``delete_mode``                            Optionally specifies the write delete mode of the Iceberg
+                                           specification to use for new tables, either ``copy-on-write``
+                                           or ``merge-on-read``. Defaults to ``merge-on-read``.
```
steveburnett, line 260: As a "nice to have but not necessary in this PR", would you consider adding a Default column for this table? If not no worries, we could do it in future as a separate docs-only PR. `partitioning` and `location` don't appear to have default values which makes sense they wouldn't, so their entry in the column could be empty, or `none`.

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -797,12 +801,14 @@ The Iceberg connector can delete data in one or more entire partitions from tabl
 
 .. note::
 
-    Columns in the filter must all be identity transformed partition columns of the target table.
-
     Filtered columns only support comparison operators, such as EQUALS, LESS THAN, or LESS THAN EQUALS.
 
     Deletes must only occur on the latest snapshot.
 
+    For V1 tables, the connector can only delete data in one or more entire
```
steveburnett, line 808:
```suggestion
    For V1 tables, the Iceberg connector can only delete data in one or more entire
```

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -797,12 +801,14 @@ The Iceberg connector can delete data in one or more entire partitions from tabl
...
+    partitions. Columns in the filter must all be identity transformed partition
```
steveburnett, line 809: No change, just noting that "Columns in the filter must all be..." is clearly phrased. Nice work!

steveburnett (changes requested):
Really, this review is only a nit, a compliment, and a suggestion (Default column) that is entirely optional for you to consider if you have time.

presto-docs/src/main/sphinx/connector/iceberg.rst
steveburnett: Reviewed, nice work!

presto-docs/src/main/sphinx/connector/iceberg.rst
steveburnett: New column looks good in new build, thanks for adding it!

steveburnett (approved):
LGTM! (docs) 

Pulled just-updated branch, new local build, everything looks good. Thanks for the quick response!

If you make further changes in the docs before merge, request review from me and I'll respond as quickly as I'm able.

gupteaj (approved):

ChunxuTang (approved):
@s-akhtar-baig This work is fantastic! The row-level deletion is a feature that the community strongly needs.

Just a quick question on the backward compatibility: After the change, would the Iceberg connector still be compatible with pre-existing V1 tables?

ChunxuTang:
BTW, could you resolve the conflicts from the main branch and make sure all tests passed?

presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedSmokeTestBase.java
```diff
@@ -1232,4 +1245,130 @@ public void testUnregisterTable()
...
+
+    @Test
+    public void testDeleteOnPartitionedV1Table()
+    {
+        String tableName = "test_delete_on_partitioned_v1_table";
+
+        Session session = getSession();
+
+        String errorMessage = format("This connector only supports delete where one or more partitions are deleted entirely for table versions older than %d", MIN_FORMAT_VERSION_FOR_DELETE);
+        assertUpdate("CREATE TABLE " + tableName + " (id integer, value integer) WITH (format_version = '1', partitioning = Array['id'])");
```
s-akhtar-baig, line 1360: Added test for delete operations on V1 tables.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergUtil.java
```diff
@@ -901,9 +917,61 @@ public static Map<String, String> populateTableProperties(ConnectorTableMetadata
...
-            propertiesBuilder.put(FORMAT_VERSION, formatVersion);
+        verify(formatVersion != null, "Format version cannot be null");
+        propertiesBuilder.put(FORMAT_VERSION, formatVersion);
+
+        if (parseFormatVersion(formatVersion) < MIN_FORMAT_VERSION_FOR_DELETE) {
+            propertiesBuilder.put(DELETE_MODE, RowLevelOperationMode.COPY_ON_WRITE.modeName());
+        }
+        else {
+            RowLevelOperationMode deleteMode = IcebergTableProperties.getDeleteMode(tableMetadata.getProperties());
+            propertiesBuilder.put(DELETE_MODE, deleteMode.modeName());
```
s-akhtar-baig, line 928: Added checks to determine delete modes for older table versions vs newer.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergTableProperties.java
```diff
@@ -74,13 +77,22 @@ public IcebergTableProperties(IcebergConfig icebergConfig)
                 .add(stringProperty(
                         FORMAT_VERSION,
                         "Format version for the table",
-                        null,
+                        DEFAULT_FORMAT_VERSION,
```
s-akhtar-baig, line 80: Changed default format version to 2 to follow Iceberg's implementation (https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/TableMetadata.java#L53C20-L53C48).

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
```diff
@@ -752,10 +766,71 @@ public void truncateTable(ConnectorSession session, ConnectorTableHandle tableHa
...
+        Table icebergTable = getIcebergTable(session, handle.getSchemaTableName());
+
         if (handle.isSnapshotSpecified()) {
             throw new PrestoException(NOT_SUPPORTED, "This connector do not allow delete data at specified snapshot");
         }
-        throw new PrestoException(NOT_SUPPORTED, "This connector only supports delete where one or more partitions are deleted entirely");
+
+        int formatVersion = ((BaseTable) icebergTable).operations().current().formatVersion();
+        if (formatVersion < MIN_FORMAT_VERSION_FOR_DELETE) {
+            throw new PrestoException(NOT_SUPPORTED, format("This connector only supports delete where one or more partitions are deleted entirely for table versions older than %d", MIN_FORMAT_VERSION_FOR_DELETE));
```
s-akhtar-baig, line 777: Added a new error message for v1 tables.

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -251,14 +251,16 @@ Property Name                             Description
 ``location``                               Optionally specifies the file system location URI for
                                            the table.
 
-``format_version``                         Optionally specifies the format version of the Iceberg
+``format_version``                         Optionally specifies the format version of the Iceberg           ``2``
```
s-akhtar-baig, line 254: @steveburnett, I have made changes to the default format version and resolved other merge conflicts. Please review when you get a chance, thanks!

steveburnett (approved):
LGTM! (docs)

Pulled updated branch, made new local build of docs and reviewed new doc changes. Everything looks good, thanks!

tdcmeehan (approved):

Issue #21062, opened 2023-10-06
[tracking] Enhancements for the UX


yhwang:
When working on the UX changes, I observed some nice-to-have enhancements to the UX. I'd like to use this issue to list and track those possible enhancements:
- [ ] In the `Split` tab of the `Query details` page, the loading and error view can be merged with the regular view by leveraging the `show/hide` CSS. This can reduce a `setTimeout` call.
- [ ] It's using bootstrap v3.3.5. It would be good to upgrade to 5.3.2 (the latest version when writing this issue) and use more components/CSS the new version provides.
- [ ] When doing the UX development, it would be good to have a lightweight webserver to serve the UX pages and proxy the APIs call to a presto coordinator. This can decouple the UX webserver from the presto server and speed up the UX dev.
- [ ] Separate some JS libraries from the generated JS pages. This can speed up the page loading time and eliminate the build warning from the webpack. 

## Expected Behavior or Use Case
It would be good to have separate PRs to address each item listed above.

## Presto Component, Service, or Connector
UX

yhwang:
Feel free to add more items in the comments if you see fit and grab any item you'd like to work on. Let me assign this to myself and I will try to tackle them one by one.

yhwang:
@jaystarshot good question also suggestion. Currently, session properties are not supported. This could be done in either the `SQL Client` component or `presto-js-client`. It's a good feature. Let me add it to the list.

yhwang:
@yingsu00 I tried to run some queries to check the live plan view. I saw the` ScanFilter` like this:
<img width="210" alt="image" src="https://github.com/prestodb/presto/assets/1147320/4b2373b3-27f1-4c85-b1f2-918a263229b7">

I saw the `connectorHandle='customer.sf1.0'` and it should indicate the table. Could you elaborate more about where you want to add the table name?

ParallaxComputers:
Hi guys, I'm working on issue number 12!

ParallaxComputers:
Hi guys, I'm working on issue number 12!

VidhuDashora:
Hi guys, sorry disregard the previous comment I had to remove it because it was the wrong account. Anyways, my name is Vidhu Dashora and I've started working on number 12, 

```12. The [suggestion](https://react.dev/reference/react/PureComponent#migrating-from-a-purecomponent-class-component-to-a-function) from React.js is to use function components. Refactor class components into function components.```

tejkart:
[...]

PR #21576, opened 2023-12-20
Add SQL Client page on Web UI

yhwang:
## Description
Leverage `presto-js-client` to run SQL queries and show results on the web UX. Add the limitation on
each SQL query to only return 100 records at most.

related to #21062

## Motivation and Context
Provide the capability to directly run simple SQL queries on the web UI and check results.

## Impact
Users accessing web UI can run SQL queries and check results on the new page. 

## Test Plan
1. check out the PR change
2. run a presto server using docker
3. run the web UI and proxy the API calls to the docker container which is running as a presto server
4. check the new page

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```


yhwang:
attached a screenshot to show the UX:
Screenshot 2023-12-19 at 8 31 55 PM

yhwang:
Some limitations:
- The results are showed after all the data returns. That's because of the current `presto-js-client` behavior. I will open an issue in presto-js-client repo to seek for enhancement
- A `Limit 100` clause will be added to the SQL query if no limit detects or the existing limitation is larger than 100. This would make the web page more efficient, use less memory and less impact on the presto cluster
- No history/caching mechanism yet
- Not able to kill the query if it's running too long. Need to navigate back to the query list in the `cluster view` page, find the query and kill it. This is also related to item 1.

tdcmeehan:
CC: @alonsovb would you and your team like to have an initial look?

presto-main/src/main/resources/webapp/src/package.json
```diff
@@ -12,17 +12,22 @@
...
     "babel-loader": "9.1.3",
+    "css-loader": "^6.8.1",
     "flow-bin": "^0.215.1",
+    "style-loader": "^3.3.3",
     "webpack": "5.88.2",
     "webpack-cli": "^5.1.4",
     "webpack-dev-server": "^4.15.1"
   },
   "dependencies": {
+    "@prestodb/presto-js-client": "^0.1.0-beta.0",
```
jorgeramirezamora, line 23: Latest version is should be [presto-client-0.1.1](https://github.com/prestodb/presto-js-client/releases/tag/presto-client-0.1.1)

presto-main/src/main/resources/webapp/src/yarn.lock
```diff
@@ -1077,6 +1077,20 @@
   resolved "https://registry.yarnpkg.com/@leichtgewicht/ip-codec/-/ip-codec-2.0.4.tgz#b2ac626d6cb9c8718ab459166d4bb405b8ffa78b"
   integrity sha512-Hcv+nVC0kZnQ3tD9GVu5xSMR4VVYOteQIr/hwFPVEvPdlXqgGEuRjiheChHgdM+JyqdgNcmzZOX/tnl0JOiI7A==
 
+"@prestodb/presto-js-client@^0.1.0-beta.0":
```
jorgeramirezamora, line 1080: See my comment above

jorgeramirezamora (changes requested):
Just an initial comment

presto-main/src/main/resources/webapp/src/package.json
yhwang: thanks @jorgeramirezamora 
fixed. I didn't specify the version when I ran `yarn add @prestodb/presto-js-client` and the interesting thing is the version shows up in [here](https://www.npmjs.com/package/@prestodb/presto-js-client?activeTab=readme) is not 0.1.1. But 0.1.1 is the latest version indeed. I will leave that to you to handle it.

jorgeramirezamora: Yeah, GitHub is marking `0.1.1` as LTS but NPM is marking `0.1.0-beta.0` as LTS. Will bump it to `0.2.0` in the coming days to avoid this issue.

presto-main/src/main/resources/webapp/src/components/SQLClient.jsx
```diff
@@ -0,0 +1,277 @@
...
+            {results.data && <div className="row"><DataTable columns={getColumns()} data={results.data} theme='dark' customStyles={customStyles} striped='true' pagination /></div>}
+        </>
+    );
+}
+
+export default function SQLClientView() {
+    const [values, setValues] = React.useState({sql: '', running: false, results: undefined});
+
+    const executeSQL = (sqlInfo) => {
+        const client = createClient(sqlInfo.catalog, sqlInfo.schema);
```
alonsovb, line 217: we're creating a new client each time we want to run a query, is this intented?

yhwang: Yes, I can change it to one instance and passing `catalog` and `schema` in each query. Which one you suggest?

alonsovb: the way you did it is fine 👍

presto-main/src/main/resources/webapp/src/components/SQLClient.jsx
```diff
@@ -0,0 +1,277 @@
...
+          setValues({...values, running: true, results: undefined});
+          client.query(sqlInfo.sql).then((prestoQuery) => {
+              if (prestoQuery.data && prestoQuery.data.length > 0) {
+                setValues({...values, running: false, results: prestoQuery});
+              } else {
+                // presto-js-client doesn't return error
```
alonsovb, line 223: this seems like a good potential improvement on the presto-js-client
I created an issue for it: https://github.com/prestodb/presto-js-client/issues/11

alonsovb (approved):
the code looks good!

presto-main/src/main/resources/webapp/src/components/SQLClient.jsx
yhwang: awesome! You spotted it! I was thinking it would be good to get the queryId first and another call either pull the result or wait for the query to finish and get the whole results at once. But it's up to you and once you have new change in `presto-js-client` I can adopt it here.

yhwang:
@alonsovb @jorgeramirezamora thanks for the prompt review and comments.

presto-main/src/main/resources/webapp/src/components/SQLClient.jsx
alonsovb: sounds good, I also think we could add a couple of extra things to the js-client:
- get catalogs/schemas utility
- get query metadata

eventually the web ui should be able to use the js-client for getting all of the things instead of having to do `fetch` manually to the REST API

yhwang: I am also curious why I don't get any error. Once you have any change in presto-js-client, please tag me and I can update the code here. Thanks!

yhwang: > sounds good, I also think we could add a couple of extra things to the js-client:
> 
>     * get catalogs/schemas utility
> 
>     * get query metadata

I like this idea! +1

yhwang:
FYI,
if you want to verify this PR but don't want to build the presto-server from scratch, you can use the presto docker image
1. check out my change:
   ```
   git clone https://github.com/yhwang/presto.git -b sql-tab
   cd presto
   ```
2. start a presto server using docker image: `public.ecr.aws/oss-presto/presto:0.286-20231219171929-8cb69ea`:
   ```
   docker run --rm --name presto-server -p 8080:8080 -v ./docker/etc/catalog:/opt/presto-server/etc/catalog public.ecr.aws/oss-presto/presto:0.286-20231219171929
   ```
   This starts a presto server on port 8080 and has jmx, memory, tpch, and tpcds catalogs. The reason I pick this image is it supports multi-arch, including amd64 and arm64.
3. make sure you have node and yarn on your system and run the following command to install the dependencies:
   ```
   yarn --cwd presto-main/src/main/resources/webapp/src install
   ```
4. start the web UI:
   ```
   yarn --cwd presto-main/src/main/resources/webapp/src run serve
   ```
   It should start the web UI on port 8081, because 8080 is used by the presto server. So you can access the web UI by:
   http://localhost:8081
   or check the output messages. It tells you the address.

jorgeramirezamora (approved):
Good job @yhwang 👏🏼

tdcmeehan:
This looks amazing and I'm so excited to see this go out.

One complication on the LIMIT addition: it needs to be added only for SELECT queries or queries with CTEs and a SELECT.  LIMIT on any other type of query will result in invalid syntax.

yhwang:
@tdcmeehan thanks for the comment. Let me add an RE to detect if it's a `select` query or not. working on it

yhwang:
@tdcmeehan added a SQL parser and only add a LIMIT clause when it's a SELECT query.

alonsovb (approved):
looks good!

tdcmeehan:
There are some problems with the parser: 
Screenshot 2023-12-22 at 3 35 29 PM

I should be able to add a fully qualified table (catalog + schema + table name), but it appears the parser doesn't like that.  It works, though, if I select the schema and catalog from the dropdown and leave them unspecified in the query.

tdcmeehan:
I also tried to surround the table name with `"`, which should disambiguate, but it appears the parser doesn't accept that syntax.  Worthwhile to check if column names can also be surrounding in quotes.

yhwang:
not good for a federated query. My guess is it only supports `<schema>.<table>` for table name. Maybe we should use the ANTLR that Presto uses to generate the JS parser?

tdcmeehan:
Antlr would definitely work!

tdcmeehan:
We could do a one-time commit of the Antlr parser into the UI JS code--we only need to know about the outer shape of the query, which is probably never going to change or evolve.

yhwang:
@tdcmeehan I generated the js parser based on the `SqlBase.g4` and used it to parse and detect the `SELECT` query.

yhwang:
Note: when fully qualified table names are not used and no catalog and schema are assigned, it only returns `SYNTAX_ERR` from the presto-js-client for now.  need changes in the presto-js-client to get more detailed information.

tdcmeehan (changes requested):
Two minor changes, otherwise LGTM

1) I would try to make sure that the SQL parser is best-effort, just in case for some reason the syntax drifts between the Antlr parser on the JS side and the underlying engine.  In particular, syntax errors should be caught in the Presto server.  I think the client should just attempt to parse, then pass as-is to the server if it's unable to, to keep the source of truth as the server.
2) Could you please include instructions in `src/sql-parser/` in a README.md file to explain how to regenerate this code?

yhwang:
@tdcmeehan for item 1, the current behavior aligns with what you mentioned. the only caveat is the presto-js-client only returns `SYNTAX_ERR` for now if there is any error on the server end. Once this issue is fixed on presto-js-client, I will re-visit the SQL client code to show/retrieve the detailed error message.

For item 2, I am working on this and will provide an update soon.

tdcmeehan (approved):

PR #21580, opened 2023-12-20
Use Proper query scope in CTE Materialization

### LINKED ISSUE #21563: CTE Materialization creating wrong plans for similar named ctes in completely different scope
Materialization of CTEs involves identifying their usage through CTE names and relative paths. However, there are two identified bugs in this process, which have been noticed in 0.2% of production queries. 

In Presto SQL, the allowance for nested and similarly named CTEs in numerous instances contributes to these issues.

Observed 2 edge cases from production

edge case 1 : 
 Local scope is not always prioritized over global scope.

```
WITH test AS (SELECT 1),
_query AS (
   with test AS (
         SELECT * FROM test. <<---- marked**
 )
 SELECT * FROM test
)
SELECT * FROM _query JOIN _query on true
```
Here the marked use of test should be from outside of the current scope. Using cte_materialization_strategy='ALL' gives a graph cycles error

edge case 2:
 

```
  SELECT   *, (WITH T as (SELECT 1) SELECT * FROM T)
                       FROM (
                        WITH T AS ( 
                          SELECT ColumnA, ColumnB FROM (
                               VALUES 
                                 (1, 'A'),
                               (2, 'B'),
                                 (3, 'C'),
                                 (4, 'D')
                             ) AS TempTable(ColumnA, ColumnB)
                         )
                         SELECT * FROM T
```

Here project and from clause both have a completely new scope creating the same CTE t, hence the cte name key should have this baseline value in it so that wrong ctes are not used at the wrong places.

jaystarshot:
Fixes #21563. 

## Motivation and Context
Fixes an issue with cte refereneces referencing wrong ctes (sample examples mentioned in the issue). This was causing incorrect plans.
Each reference of the cte expands the same namedQuery, hence we can fix all these issues  with normalization of that named query object

## Contributor checklist

- [x] CI passed.
- [x] Final Prod Shadow test

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Fixed a bug in cte materialization which was causing incorrect plan generation in some cases

### DIFF
--- a/presto-hive/src/test/java/com/facebook/presto/hive/TestCteExecution.java
+++ b/presto-hive/src/test/java/com/facebook/presto/hive/TestCteExecution.java
@@ -168,6 +168,28 @@ public class TestCteExecution
     }
 
     @Test
+    public void testRedefinedCteWithSameDefinitionDifferentBase()
+    {
+        String testQuery = "SELECT (with test_base AS (SELECT colB FROM (VALUES (1)) AS TempTable(colB)), \n" +
+                "test_cte as (  SELECT colB FROM test_base)\n" +
+                "SELECT * FROM test_cte\n" +
+                "),\n" +
+                "(WITH test_base AS (\n" +
+                "    SELECT text_column\n" +
+                "    FROM (VALUES ('Some Text', 9)) AS t (text_column, number_column)\n" +
+                "), \n" +
+                "test_cte AS (\n" +
+                "    SELECT * FROM test_base\n" +
+                ")\n" +
+                "SELECT  CONCAT(text_column , 'XYZ') FROM test_cte\n" +
+                ")\n";
+        QueryRunner queryRunner = getQueryRunner();
+        compareResults(
+                queryRunner.execute(getMaterializedSession(), testQuery),
+                queryRunner.execute(getSession(), testQuery));
+    }
+
+    @Test
     public void testComplexRefinedCtesOutsideScope()
     {
         String testQuery = "WITH " +
--- a/presto-main/src/main/java/com/facebook/presto/sql/planner/SqlPlannerContext.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/planner/SqlPlannerContext.java
@@ -15,13 +15,17 @@ package com.facebook.presto.sql.planner;
 
 import com.facebook.presto.Session;
 import com.facebook.presto.spi.PrestoException;
+import com.facebook.presto.sql.analyzer.Analysis;
 import com.facebook.presto.sql.relational.SqlToRowExpressionTranslator;
+import com.facebook.presto.sql.tree.DefaultTraversalVisitor;
 import com.facebook.presto.sql.tree.Query;
+import com.facebook.presto.sql.tree.Table;
 import com.google.common.annotations.VisibleForTesting;
 
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.Stack;
+import java.util.TreeSet;
 
 import static com.facebook.presto.SystemSessionProperties.getMaxLeafNodesInPlan;
 import static com.facebook.presto.SystemSessionProperties.isLeafNodeLimitEnabled;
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by jaystarshot** [comment]
This same issue will happen in CteInformationCollector (wrong reference count will increase), thinking about adding a new field called actualCtePath in [CTEInformation](https://github.com/prestodb/presto/blob/8aebd3434b8253cc821b06621df7d5c21f61f3be/presto-spi/src/main/java/com/facebook/presto/spi/eventlistener/CTEInformation.java#L21). What do you think?

File: presto-main/src/main/java/com/facebook/presto/sql/planner/SqlPlannerContext.java [hunk 1]
```diff
@@ -64,57 +63,22 @@ public void incrementLeafNodes(Session session)
...
-        }
+        // All references of the same queryStatement fall under the same scope
+        Map<Query, String> queryNodeScopeIdMap = new HashMap<>();
 
-        public String getRawPath(String cteName)
+        public String normalize(Query query, String cteName)
         {
-            if (!this.rawCtePathMap.containsKey(cteName)) {
-                return cteName;
+            if (!queryNodeScopeIdMap.containsKey(query)) {
```
mlyublena, line 78: nit: you can use putIfAbsent

presto-main/src/main/java/com/facebook/presto/sql/planner/SqlPlannerContext.java
```diff
@@ -64,57 +63,22 @@ public void incrementLeafNodes(Session session)
...
-
-        public void pop(Query query)
-        {
-            this.cteStack.pop();
-            if (query.getWith().isPresent()) {
-                query.getWith().get().getQueries().forEach(with -> this.removeNestedCte(with.getName().toString()));
-            }
-        }
+        // All references of the same queryStatement fall under the same scope
+        Map<Query, String> queryNodeScopeIdMap = new HashMap<>();
```
mlyublena, line 74: nit: why do we convert the scope id to string?

mlyublena (approved):

jaystarshot:
```
SELECT (WITH test_base AS
       (
              SELECT colb
              FROM   (VALUES
                     (
                            1
                     )
                     ) AS temptable(colb)), test_cte AS
       (
              SELECT *
              FROM   test_base)SELECT *
FROM   test_cte ),
       (WITH test_base AS
       (
              SELECT text_column
              FROM   (VALUES
                     (
                            'Some Text',
                            9
                     )
                     ) AS t (text_column, number_column) ), test_cte AS
       (
              SELECT *
              FROM   test_base )SELECT *
FROM   test_cte )

```
Observed just one more extra edge case in prod where the definition of the test_cte remains same but the test_base changes ( referencing some other column).  In this case we are still considering both ctes to be the same.
Thinking how to fix this, because seems non trivial

presto-main/src/main/java/com/facebook/presto/sql/planner/SqlPlannerContext.java
jaystarshot: Just so that it is easier to debug and add tests, we can use hashCode but it becomes complicated

presto-hive/src/test/java/com/facebook/presto/hive/TestCteExecution.java
```diff
@@ -167,6 +167,28 @@ public void testRefinedCtesOutsideScope()
                 queryRunner.execute(getSession(), testQuery));
     }
 
+    @Test
+    public void testRedefinedCteWithSameDefinationDifferentBase()
```
feilong-liu, line 171: typo: Defination -> Definition

feilong-liu (approved):

presto-main/src/main/java/com/facebook/presto/sql/planner/SqlPlannerContext.java
jaystarshot: Keeping same since its better that the ++ happens only if its new

tdcmeehan (approved):

PR #21593, opened 2023-12-22
[native]Add system config for per-driver cpu time slicing enforcement

xiaoxmeng:
Add system config got per-driver cpu time slicing enforcement and set the
corresponding config in velox query config accordingly

### DIFF
--- a/presto-native-execution/presto_cpp/main/common/Configs.cpp
+++ b/presto-native-execution/presto_cpp/main/common/Configs.cpp
@@ -214,6 +214,7 @@ SystemConfig::SystemConfig() {
           BOOL_PROP(kUseLegacyArrayAgg, false),
           STR_PROP(kSinkMaxBufferSize, "32MB"),
           STR_PROP(kDriverMaxPagePartitioningBufferSize, "32MB"),
+          NUM_PROP(kDriverCpuTimeSliceLimitMs, 0),
       };
 }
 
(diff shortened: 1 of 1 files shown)


presto-native-execution/presto_cpp/main/common/Configs.cpp
```diff
@@ -573,6 +574,10 @@ bool SystemConfig::useLegacyArrayAgg() const {
   return optionalProperty<bool>(kUseLegacyArrayAgg).value();
 }
 
+uint32_t SystemConfig::driverCpuTimeSliceLimitMs() const {
+  return optionalProperty<bool>(kDriverCpuTimeSliceLimitMs).value();
```
tanjialiang, line 578: return optionalProperty<uint32_t>(kDriverCpuTimeSliceLimitMs).value();

presto-native-execution/presto_cpp/main/common/Configs.cpp
```diff
@@ -752,6 +760,8 @@ void BaseVeloxQueryConfig::updateLoadedValues(
       {QueryConfig::kMaxPartitionedOutputBufferSize,
        systemConfig->capacityPropertyAsBytesString(
            SystemConfig::kDriverMaxPagePartitioningBufferSize)},
+      {QueryConfig::kDriverCpuTimeSliceLimitMs,
+       std::to_string(systemConfig->driverCpuTimeSliceLimitMs())},
```
tanjialiang, line 764: shall we use folly::to\<std::string\>() ?

Also I don't quite understand this method. Why are we putting these query config into system config while initialize the system config? Thanks

tanjialiang (commented):
Thanks for the change.

presto-native-execution/presto_cpp/main/common/Configs.cpp
xiaoxmeng: Yeah, we probably doesn't need system config. Removed.

Issue #20701, opened 2023-08-29
When query a column of type map<string, string>, the keys of the map are converted to lowercase


yhwang:
When running the following query:
```
SELECT MAP(ARRAY['myFirstRow', 'mySecondRow'], ARRAY[cast(row('row1FieldValue1', 'row1FieldValue2') as row("firstField" varchar, "secondField" varchar)), cast(row('row2FieldValue1', 'row2FieldValue2') as row("firstField" varchar, "secondField" varchar))]) as mapField;
```

I got the following results:
```
                                                                   mapField                                                                    
-----------------------------------------------------------------------------------------------------------------------------------------------
 {mySecondRow={firstfield=row2FieldValue1, secondfield=row2FieldValue2}, myFirstRow={firstfield=row1FieldValue1, secondfield=row1FieldValue2}} 
(1 row)
```

The keys of the map are `firstField` and `secondField` in the query. However, they become `firstfield` and `secondfield` in the query results.

## Your Environment

* Presto version used: 0.284 (master branch)
* Storage (HDFS/S3/GCS..): N/A
* Data source and connector used: N/A
* Deployment (Cloud or On-prem): local dev env
* [Pastebin](https://pastebin.com/) link to the complete debug logs:

## Expected Behavior
The result should preserve the original keys without lowercase them

## Current Behavior
The keys of a map are converted to lowercase

## Possible Solution
Not sure this is a CLI or server side issue

## Steps to Reproduce

1. start a presto CLI
2. run the following query: `SELECT MAP(ARRAY['myFirstRow', 'mySecondRow'], ARRAY[cast(row('row1FieldValue1', 'row1FieldValue2') as row("firstField" varchar, "secondField" varchar)), cast(row('row2FieldValue1', 'row2FieldValue2') as row("firstField" varchar, "secondField" varchar))]) as mapField;`

milescrawford:
This also seems to affect the *contents* of the column, which I think is a bigger issue.

Presto is reading from a json table like this:
```
{"corpusid":186849207,"externalids":{"ACL":null,"DBLP":null,"ArXiv":null,"MAG":"2914916907","CorpusId":"186849207", ...
```

Output of presto query: `select * from table limit 1`
```
#   corpusid    corpusid    externalids ...
[...]

PR #21602, opened 2023-12-24
Preserve case for RowType's field name and JSON content when ``CAST``

hantangwangd:
## Description

Fixes issue #20701 .

In parsing stage, when we try to build a `CAST` expression, we should not simply translate the target type string to lower case, as the target type string may have some quoted field names which should retain their original case.

So as to json string, when we try to `CAST` it to other type like `RowType`, we should not roughly translate it to lower case, that would corrupt the json data.

This PR try to fix the problems above.

## Test Plan

 - Enhanced test case TestRowOperators.testJsonToRow(), adding case reservation verify for json cast to other type
 - Enhanced test case TestRowOperators.testRowCast(), adding the example mentioned in issue #20701 
 - Make sure this fix do not affect existing test cases

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-parser/src/main/java/com/facebook/presto/sql/tree/Cast.java
+++ b/presto-parser/src/main/java/com/facebook/presto/sql/tree/Cast.java
@@ -19,7 +19,6 @@ import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 
-import static java.util.Locale.ENGLISH;
 import static java.util.Objects.requireNonNull;
 
 public final class Cast
@@ -67,7 +66,7 @@ public final class Cast
         requireNonNull(type, "type is null");
 
         this.expression = expression;
-        this.type = type.toLowerCase(ENGLISH);
+        this.type = transformCase(type);
         this.safe = safe;
         this.typeOnly = typeOnly;
     }
@@ -90,6 +89,34 @@ public final class Cast
     public boolean isTypeOnly()
     {
         return typeOnly;
+    }
+
+    /**
+     * We uniformly change the whole type string to lower case,
+     *  except field names which are enclosed in double quotation marks
+     * */
+    private static String transformCase(String input)
+    {
+        if (input == null) {
+            return null;
+        }
+
+        StringBuilder sb = new StringBuilder();
+        boolean insideQuotes = false;
+        for (int i = 0, l = input.length(); i < l; i++) {
+            char ch = input.charAt(i);
+            if (ch == '"') {
+                insideQuotes = !insideQuotes;
+            }
+
+            if (insideQuotes) {
+                sb.append(ch);
+            }
+            else {
+                sb.append(Character.toLowerCase(ch));
+            }
+        }
+        return sb.toString();
     }
 
     @Override
--- a/presto-common/src/main/java/com/facebook/presto/common/type/RowType.java
+++ b/presto-common/src/main/java/com/facebook/presto/common/type/RowType.java
@@ -83,6 +83,11 @@ public class RowType
         return new Field(Optional.empty(), type);
     }
 
+    public static Field field(String name, Type type, boolean delimited)
+    {
+        return new Field(Optional.of(name), type, delimited);
+    }
+
     private static TypeSignature makeSignature(List<Field> fields)
     {
         int size = fields.size();
--- a/presto-main/src/main/java/com/facebook/presto/util/JsonUtil.java
+++ b/presto-main/src/main/java/com/facebook/presto/util/JsonUtil.java
@@ -54,7 +54,6 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
-import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
 import java.util.TreeMap;
@@ -1324,7 +1323,7 @@ public final class JsonUtil
                 if (parser.currentToken() != FIELD_NAME) {
                     throw new JsonCastException(format("Expected a json field name, but got %s", parser.getText()));
                 }
-                String fieldName = parser.getText().toLowerCase(Locale.ENGLISH);
+                String fieldName = parser.getText();
                 Integer fieldIndex = fieldNameToIndex.get().get(fieldName);
                 parser.nextToken();
                 if (fieldIndex != null) {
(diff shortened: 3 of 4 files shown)


presto-parser/src/main/java/com/facebook/presto/sql/tree/Cast.java
```diff
@@ -92,6 +91,34 @@ public boolean isTypeOnly()
...
+     * */
+    private static String transformCase(String input)
+    {
+        if (input == null) {
+            return null;
+        }
+
+        StringBuilder sb = new StringBuilder();
+        boolean insideQuotes = false;
+        for (int i = 0; i < input.length(); i++) {
```
yhwang, line 106:
```suggestion
        for (int i = 0, l = input.length(); i < l; i++) {
```

yhwang:
Sorry for the late response. Thanks for the RP. You also added the query in the issue as the test case. Nice job!

yhwang (approved):

presto-parser/src/main/java/com/facebook/presto/sql/tree/Cast.java
```diff
@@ -92,6 +91,34 @@ public boolean isTypeOnly()
         return typeOnly;
     }
 
+    /**
+     * We uniformly change the whole type string to lower case,
+     *  except field names which are enclosed in double quotation marks
+     * */
+    private static String transformCase(String input)
```
tdcmeehan, line 98: I think we should also validate that the quotes are matched--suppose someone makes a typo, it will cause the error message or output to be strange and unexpected.  I would also double check that one can't escape double quotes inside double quotes (I don't believe this is possible currently).

yhwang: not sure if the unmatched quotes would be caught by ANTLR parser first or not, my guess would be very possible

yhwang: double-check the SqlBase.g4:
```
ROW '(' identifier type (',' identifier type)* ')'

identifier
    : IDENTIFIER             #unquotedIdentifier
    | QUOTED_IDENTIFIER      #quotedIdentifier
    | nonReserved            #unquotedIdentifier
    | BACKQUOTED_IDENTIFIER  #backQuotedIdentifier
    | DIGIT_IDENTIFIER       #digitIdentifier
```
The parser would throw a syntax error if the quote doesn't end properly.

hantangwangd: I have run some test cases to make sure that ANTLR parse would indeed catch the unmatched quotes and throw a ParsingException.

Besides, seems we have already handled escape double quotes properly, for SqlBase.g4 defined escape double quotes as '""'. Traditional escape double quotes like '\"' would be consider as two chars: '\' and '"', so the parser would handle it.

tdcmeehan: Makes sense!

tdcmeehan (approved):
@hantangwangd thank you for fixing this bug!

hantangwangd:
@tdcmeehan My pleasure!

ajaygeorge:
@hantangwangd @tdcmeehan 
This commit seems to have introduced a bug. Can you please check.

Simple repro :
With this commit age field becomes null.
```
SELECT CAST(json_extract('{"1":[{"name": "John", "AGE": 30}]}', '$') AS MAP<bigint, ARRAY<ROW(name VARCHAR, age VARCHAR)>>) AS data;
            data
-----------------------------
 {1=[{name=John, age=null}]}
(1 row)
```

With this commit reverted, age field is in small caps and has a value.
```
 SELECT CAST(json_extract('{"1":[{"name": "John", "AGE": 30}]}', '$') AS MAP<bigint, ARRAY<ROW(name VARCHAR, age VARCHAR)>>) AS data;
           data
---------------------------
 {1=[{name=John, age=30}]}
(1 row)
```

hantangwangd:
@ajaygeorge OK,  I will check, thanks for pointing out.

hantangwangd:
Hi @ajaygeorge, I think this behavior meets the expectation of this commit, because the purpose is to reserve the case of words enclosed in double quotation marks in json.

So in your test case maybe we could use:
```
SELECT CAST(json_extract('{"1":[{"name": "John", "AGE": 30}]}', '$') AS MAP<bigint, ARRAY<ROW(name VARCHAR, "AGE" VARCHAR)>>) AS data;
-----------------------------
 {1=[{name=John, AGE=30}]} 
(1 row)
```

Is that OK? Any thoughts please let me know.

mbasmanova:
@hantangwangd The problem with this commit is that it produces a change in behavior for user queries. The results do not match before and after the changes in this PR. We'll have to revert this change.

hantangwangd:
Hi @mbasmanova, thanks for your information. Actually this change contains two parts of modifications. The first part is about `CAST` some data to a specified type, in which we reserve the case of type names enclosed in double quotation marks, in case that we intended to preserve the capitalization. this is a problem reported in https://github.com/prestodb/presto/issues/20701 by @yhwang. 

And the second part is about casting a json data to a row type, in which reserving case in `fieldName` would produce the behavior changes mentioned by @ajaygeorge. 

Should the two parts all be reverted? Or the first part of modification should be reserved, as it shouldn't cause any ambiguity?

jainxrohit:
> Hi @mbasmanova, thanks for your information. Actually this change contains two parts of modifications. The first part is about `CAST` some data to a specified type, in which we reserve the case of type names enclosed in double quotation marks, in case that we intended to preserve the capitalization. this is a problem reported in #20701 by @yhwang.

Preserving the case is a good change, however this results in the change of behavior. Which causes user pipelines to fail. For now lets revert this. We can reintroduce the PR with a flag for reserving the case.

mbasmanova:
CC: @aditi-pandit

hantangwangd:
> Preserving the case is a good change, however this results in the change of behavior. Which causes user pipelines to fail. For now lets revert this. We can reintroduce the PR with a flag for reserving the case.

OK, I get it. I will do this.

PR #21605, opened 2023-12-27
Iceberg: Support applying equality deletes as a join

jasonf20:
## Description
This PR adds support for applying equality deletes as a join instead of while reading the split data.

Since equality deletes often apply to a lot of files, the current
implementation ends up opening the delete files #splits * #delete_files
times which can be very large for cdc/upsert use cases.

This commit implements equality deletes as a join. A connector optimizer
is added to apply the appropriate join(s).

## Motivation and Context
Currently, the use case of frequent updates with MoR + Eventual compaction doesn't work because querying tables with more than a couple of delete files is too slow. This enables new use cases to be done using Iceberg in Presto such as CDC.

### Example
This example queries a table on a partition with ~5GB of data files + ~150 delete files. Cpu time of 39 seconds vs 5:30 hours
image
image

For each additional delete file the current implementation will need to load #splits more files. While the new method will only load 1 more file. So the difference will keep growing with more files.

**Query Plan**
```
- Output[PlanNodeId 9][_col0] => [count:bigint]
        _col0 := count (1:16)
    - Aggregate(FINAL)[PlanNodeId 4] => [count:bigint]
            count := "presto.default.count"((count_9)) (1:16)
        - LocalExchange[PlanNodeId 393][SINGLE] () => [count_9:bigint]
            - RemoteStreamingExchange[PlanNodeId 399][GATHER] => [count_9:bigint]
                - Aggregate(PARTIAL)[PlanNodeId 397] => [count_9:bigint]
                        count_9 := "presto.default.count"((build_tag)) (1:16)
                    - FilterProject[PlanNodeId 326,2][filterPredicate = IS_NULL($data_sequence_number_6), projectLocality = LOCAL] => [build_tag:varchar]
                            Estimates: {source: CostBasedSourceInfo, rows: ? (?), cpu: ?, memory: 3,825,758.00, network: 3,825,758.00}/{source: CostBasedSourceInfo, rows: ? (?), cpu: ?, memory: 3,825,758.00, network: 3,825,758.00}
                        - LeftJoin[PlanNodeId 225][("file_name_4" = "file_name_7") AND ("partition_date" = "partition_date_8") AND ($data_sequence_number_5) < ($data_sequence_number_6)][$hashvalue, $hashvalue_10] => [build_tag:varchar, $data_sequence_number_6:bigint]
                                Estimates: {source: CostBasedSourceInfo, rows: ? (?), cpu: ?, memory: 3,825,758.00, network: 3,825,758.00}
                                Distribution: REPLICATED
                                SortExpression[$data_sequence_number_6]
                            - ScanProject[PlanNodeId 0,444][table = TableHandle {connectorId='iceberg', connectorHandle='files_upsert$data_without_equality_deletes@Optional[6805049505336407491]', layout='Optional[files_upsert$data_without_equality_deletes@Optional[6805049505336407491]]'}, projectLocality = LOCAL] => [partition_date:date, build_tag:varchar, $data_sequence_number_5:bigint, file_name_4:varchar, $hashvalue:bigint]
                                    Estimates: {source: CostBasedSourceInfo, rows: 140,725,454 (1.18GB), cpu: 3,907,632,853.00, memory: 0.00, network: 0.00}/{source: CostBasedSourceInfo, rows: 140,725,454 (1.18GB), cpu: 9,081,794,792.00, memory: 0.00, network: 0.00}
                                    $hashvalue := combine_hash(combine_hash(BIGINT'0', COALESCE($operator$hash_code(file_name_4), BIGINT'0')), COALESCE($operator$hash_code(partition_date), BIGINT'0')) (1:38)
                                    partition_date := 3:partition_date:date (1:38)
                                    $data_sequence_number_5 := 2147482646:$data_sequence_number:bigint
                                    build_tag := 6:build_tag:varchar (1:38)
                                    file_name_4 := 2:file_name:varchar
                            - LocalExchange[PlanNodeId 373][HASH][$hashvalue_10] (file_name_7, partition_date_8) => [file_name_7:varchar, partition_date_8:date, $data_sequence_number_6:bigint, $hashvalue_10:bigint]
                                    Estimates: {source: CostBasedSourceInfo, rows: 122,904 (1.05MB), cpu: 10,371,138.00, memory: 0.00, network: 3,825,758.00}
                                - RemoteStreamingExchange[PlanNodeId 311][REPLICATE] => [file_name_7:varchar, partition_date_8:date, $data_sequence_number_6:bigint, $hashvalue_11:bigint]
                                        Estimates: {source: CostBasedSourceInfo, rows: 122,904 (1.05MB), cpu: 6,545,380.00, memory: 0.00, network: 3,825,758.00}
                                    - ScanProject[PlanNodeId 224,445][table = TableHandle {connectorId='iceberg', connectorHandle='files_upsert$equality_deletes@Optional[6805049505336407491]', layout='Optional[files_upsert$equality_deletes@Optional[6805049505336407491]]'}, projectLocality = LOCAL] => [file_name_7:varchar, partition_date_8:date, $data_sequence_number_6:bigint, $hashvalue_12:bigint]
                                            Estimates: {source: CostBasedSourceInfo, rows: 122,904 (1.05MB), cpu: 2,719,622.00, memory: 0.00, network: 0.00}/{source: CostBasedSourceInfo, rows: 122,904 (1.05MB), cpu: 6,545,380.00, memory: 0.00, network: 0.00}
                                            $hashvalue_12 := combine_hash(combine_hash(BIGINT'0', COALESCE($operator$hash_code(file_name_7), BIGINT'0')), COALESCE($operator$hash_code(partition_date_8), BIGINT'0'))
                                            file_name_7 := 2:file_name:varchar
                                            $data_sequence_number_6 := 2147482646:$data_sequence_number:bigint
                                            partition_date_8 := 3:partition_date:date

```

## Impact
New parameter to configure if this is enabled. 

## Test Plan
* Wrote new tests that cover edge cases with this implementation
* Changed all existing equality delete tests to also test with this feature enabled
* Tested manually with tables that I had that didn't work with Presto previously (see above)

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==


Iceberg Changes
* Add support for querying ``"$path"`` which returns the file path containing the row
* Add support for querying ``"$data_sequence_number"`` which returns the Iceberg data sequence number of the file containing the row 
* Add support for applying equality deletes as a join. This can be enabled with the session property ``iceberg.delete_as_join_pushdown_enabled``.

SPI Changes
* Add support for connectors to return joins  in ``ConnectorPlanOptimizer.optimize``
```


presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergEqualityDeleteAsJoin.java
```diff
@@ -0,0 +1,458 @@
...
+                ConnectorSession session,
+                TypeManager typeManager,
+                VariableAllocator variableAllocator)
+        {
+            this.functionResolution = functionResolution;
+            this.transactionManager = transactionManager;
+            this.idAllocator = idAllocator;
+            this.session = session;
+            this.typeManager = typeManager;
+            this.variableAllocator = variableAllocator;
```
tdcmeehan, line 152: Please validate inputs e.g. 
```suggestion
            this.functionResolution = requireNonNull(functionResolution, "functionResolution is null");
```

presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergEqualityDeleteAsJoin.java
```diff
@@ -0,0 +1,458 @@
...
+                            new SpecialFormExpression(SpecialFormExpression.Form.COALESCE, BigintType.BIGINT, deleteVersionColumns)));
+
+            return filter;
+        }
+
+        private static ImmutableMap<Set<Integer>, DeleteSetInfo> collectDeleteInformation(Table icebergTable,
+                IcebergTableHandle icebergTableHandle,
+                long snapshotId)
+        {
+            HashMap<Set<Integer>, DeleteSetInfo> deleteInformations = new HashMap<>();
```
tdcmeehan, line 251: Use an Immutable Map builder

presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergEqualityDeleteAsJoin.java
```diff
@@ -0,0 +1,458 @@
...
+            try (CloseableIterator<DeleteFile> files =
+                    IcebergUtil.getDeleteFiles(icebergTable, snapshotId, icebergTableHandle.getPredicate(), Optional.empty()).iterator()) {
```
tdcmeehan, line 253: Static import
```suggestion
                    getDeleteFiles(icebergTable, snapshotId, icebergTableHandle.getPredicate(), Optional.empty()).iterator()) {
```

presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergEqualityDeleteAsJoin.java
```diff
@@ -0,0 +1,458 @@
...
+                while (files.hasNext()) {
```
tdcmeehan, line 254: Use for-each?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergEqualityDeleteAsJoin.java
```diff
@@ -0,0 +1,458 @@
...
+                    DeleteFile delete = files.next();
+                    if (delete.content() == FileContent.EQUALITY_DELETES) {
+                        Map<Integer, PartitionFieldInfo> partitionFields = new HashMap<>();
```
tdcmeehan, line 257: Use a builder?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergEqualityDeleteAsJoin.java
```diff
@@ -0,0 +1,458 @@
...
+                        PartitionSpec partitionSpec = icebergTable.specs().get(delete.specId());
+                        Types.StructType partitionType = partitionSpec.partitionType();
+                        // PartitionField ids are unique across the entire table in v2. We can assume we are in v2 since v1 doesn't have equality deletes
+                        partitionSpec.fields().forEach(f -> partitionFields.put(f.fieldId(), new PartitionFieldInfo(partitionType.field(f.fieldId()), f)));
+                        List<Integer> partitionFieldIds = partitionSpec.fields().stream().map(PartitionField::fieldId).collect(Collectors.toList());
```
tdcmeehan, line 262:
```suggestion
                        List<Integer> partitionFieldIds = partitionSpec.fields().stream().map(PartitionField::fieldId).collect(toImmutableList());
```

presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergEqualityDeleteAsJoin.java
```diff
@@ -0,0 +1,458 @@
...
+                        HashSet<Integer> result = new HashSet<>();
+                        result.addAll(delete.equalityFieldIds());
+                        result.addAll(partitionFieldIds);
+                        deleteInformations.put(ImmutableSet.copyOf(result), new DeleteSetInfo(partitionFields, delete.equalityFieldIds()));
+                    }
+                }
+            }
+            catch (IOException e) {
+                return ImmutableMap.of();
```
tdcmeehan, line 271: I think this should fail loudly.  Consider an Iceberg exception.

tdcmeehan (changes requested):
Just leaving some generic comments on the plan optimizer, please generalize the feedback where possible.  I will need more time to review.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergEqualityDeleteAsJoin.java
jasonf20: This particular map can have duplicates added to it since there can be delete files with the same delete schema.  which fails the ImmutableMapBuilder. However, it is converted to an immutable map before being returned from this function.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergColumnHandle.java
```diff
@@ -160,6 +164,59 @@ public String toString()
         return getId() + ":" + getName() + ":" + type.getDisplayName() + ":" + columnType + ":" + requiredSubfields;
     }
 
+    public static IcebergColumnHandle pathColumnHandle()
+    {
+        return getIcebergColumnHandle(FILE_PATH);
```
tdcmeehan, line 169: Should these not just be constants?  And if so, let's just make them public static, since IcebergColumnHandle is immutable.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergMetadataColumn.java
```diff
@@ -0,0 +1,72 @@
...
+import org.apache.iceberg.MetadataColumns;
+
+import java.util.Set;
+import java.util.stream.Stream;
+
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
+
+public enum IcebergMetadataColumn
+{
+    FILE_PATH(MetadataColumns.FILE_PATH.fieldId(), "$path", VarcharType.VARCHAR, ColumnIdentity.TypeCategory.PRIMITIVE),
```
tdcmeehan, line 28: Please use static imports to clean this up a bit.

tdcmeehan (commented):
`Iceberg: Support metadata columns`

Can we update the documentation as well noting these new columns (`iceberg.rst`)?

tdcmeehan (commented):
`Allow connector optimizers to create joins`

I think this approach is OK.  I would move out `Type` and `EquiJoinClause` to top level classes because they're shared between the `ConnectorJoinNode` and the original `JoinNode`, and the inner class relationship these classes have to `ConnectorJoinNode` may imply that they're somehow specific to the connector plan optimizer somehow.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergPageSourceProvider.java
```diff
@@ -752,11 +754,17 @@ public ConnectorPageSource createPageSource(
         ConnectorPageSource dataPageSource = connectorPageSourceWithRowPositions.getConnectorPageSource();
 
         Supplier<Optional<RowPredicate>> deletePredicate = Suppliers.memoize(() -> {
+            // If equality deletes are optimized into a join they don't need to be applied here
+            List<DeleteFile> deletesToApply = split
+                    .getDeletes()
+                    .stream()
+                    .filter(deleteFile -> deleteFile.content() == POSITION_DELETES || equalityDeletesRequired)
+                    .collect(Collectors.toList());
```
tdcmeehan, line 762: `toImmutableList`

presto-iceberg/src/main/java/com/facebook/presto/iceberg/TableType.java
```diff
@@ -22,5 +22,7 @@ public enum TableType
     PARTITIONS,
     FILES,
     PROPERTIES,
-    CHANGELOG
+    CHANGELOG,
+    EQUALITY_DELETES,
+    DATE_WITHOUT_EQUALITY_DELETES
```
tdcmeehan, line 27: `DATA_WITHOUT_EQUALITY_DELETES`

tdcmeehan: The table type has a special meaning--it's parsed and can be specified through a special syntax (e.g. `table$data_without_equality_deletes`).  It would be better for this not to be exposed publicly, even in an undocumented way.

At this moment I'm not sure if `IcebergTableHandle` is the best place to put this.  As a compromise, let's explicitly disallow its use in `IcebergTableName#from`, perhaps by adding a flag on this enum which specifies if it's a public or private table type (all existing ones would be public).

tdcmeehan (changes requested):
`Iceberg: Support applying equality deletes as a join`

presto-iceberg/src/main/java/com/facebook/presto/iceberg/TableType.java
tdcmeehan: BTW, you can add a trailing comma to prevent a similar diff on line 25 for future additions.  e.g. 
```suggestion
    DATE_WITHOUT_EQUALITY_DELETES,
```

presto-iceberg/src/main/java/com/facebook/presto/iceberg/equalitydeletes/EqualityDeletesSplitSource.java
```diff
@@ -0,0 +1,114 @@
...
+    private final Map<Integer, PartitionSpec> specById;
+    private final CloseableIterator<DeleteFile> deleteFiles;
+
+    public EqualityDeletesSplitSource(
+            ConnectorSession session,
+            Table table,
+            CloseableIterable<DeleteFile> deleteFiles)
+    {
+        this.session = requireNonNull(session, "session is null");
+        this.specById = table.specs();
```
tdcmeehan, line 60: Please validate input

presto-iceberg/src/main/java/com/facebook/presto/iceberg/equalitydeletes/EqualityDeletesSplitSource.java
```diff
@@ -0,0 +1,114 @@
...
+    @Override
+    public boolean isFinished()
+    {
+        return !deleteFiles.hasNext();
+    }
+
+    @Override
+    public CompletableFuture<ConnectorSplitBatch> getNextBatch(ConnectorPartitionHandle partitionHandle, int maxSize)
+    {
+        List<ConnectorSplit> splits = new ArrayList<>();
```
tdcmeehan, line 74: Please use immutable list

presto-iceberg/src/main/java/com/facebook/presto/iceberg/equalitydeletes/EqualityDeletesSplitSource.java
```diff
@@ -0,0 +1,114 @@
...
+        Iterator<DeleteFile> iterator = limit(deleteFiles, maxSize);
+        while (iterator.hasNext()) {
```
tdcmeehan, line 76: Use for each

presto-iceberg/src/main/java/com/facebook/presto/iceberg/equalitydeletes/EqualityDeletesSplitSource.java
```diff
@@ -0,0 +1,114 @@
...
+            splits.add(toIcebergSplit(manifestReader));
+        }
+        return completedFuture(new ConnectorSplitBatch(splits, isFinished()));
+    }
+
+    @Override
+    public void close()
+    {
+        try {
+            deleteFiles.close();
```
tdcmeehan, line 87: Please apply a similar fix as in #21612, i.e. additionally set it to null or empty iterator

presto-iceberg/src/main/java/com/facebook/presto/iceberg/equalitydeletes/EqualityDeletesSplitSource.java
```diff
@@ -0,0 +1,114 @@
...
+
+    public EqualityDeletesSplitSource(
+            ConnectorSession session,
+            Table table,
+            CloseableIterable<DeleteFile> deleteFiles)
+    {
+        this.session = requireNonNull(session, "session is null");
+        this.specById = table.specs();
+        requireNonNull(deleteFiles, "deleteFiles is null");
+        this.deleteFiles = CloseableIterable.filter(deleteFiles, f -> f.content() == FileContent.EQUALITY_DELETES).iterator();
```
tdcmeehan, line 62: Please rename `f` to `deleteFile` (this project avoids single letter variable names)

presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergEqualityDeleteAsJoin.java
tdcmeehan: No problem, please add little comment explaining this

presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergEqualityDeleteAsJoin.java
```diff
@@ -0,0 +1,458 @@
...
+                ImmutableMap<VariableReferenceExpression, ColumnHandle> deleteColumnAssignments = ImmutableMap.<VariableReferenceExpression, ColumnHandle>builder()
+                        .putAll(deleteGroupInfo.allFields(icebergTable.schema()).stream().collect(Collectors.toMap(this::toVariableReference, this::toIcebergColumnHandle)))
+                        .put(joinSequenceNumber, IcebergColumnHandle.dataSequenceNumberColumnHandle())
+                        .build();
+
+                // ON source.delete_column = deletes.delete_column, ...
+                Set<ConnectorJoinNode.EquiJoinClause> clauses = deleteColumnAssignments
+                        .entrySet()
+                        .stream()
+                        .filter(f -> !IcebergMetadataColumn.isMetadataColumnId(((IcebergColumnHandle) (f.getValue())).getId()))
```
tdcmeehan, line 211: Please call it `assignment`

presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergEqualityDeleteAsJoin.java
```diff
@@ -0,0 +1,458 @@
...
+                // no equality deletes
+                return node;
+            }
+
+            // Add all the fields required by the join that were not added by the user's query
+            ImmutableMap<VariableReferenceExpression, ColumnHandle> unselectedAssignments = createAssignmentsForUnselectedFields(node, deleteSchemas, icebergTable);
+            TableScanNode updatedTableScan = createNewRoot(node, icebergTableHandle, tableName, unselectedAssignments, table);
+
+            Map<Integer, VariableReferenceExpression> reverseAssignmentsMap =
+                    updatedTableScan.getAssignments().entrySet().stream().collect(Collectors.toMap(x -> ((IcebergColumnHandle) (x.getValue())).getId(), Map.Entry::getKey));
```
tdcmeehan, line 185: Please call it `assignment`

tdcmeehan (commented):
`Iceberg: Support applying equality deletes as a join`

I think broadcast join is a global property of the query.  Is it expected that users will manually specify that they enable broadcast join?

jasonf20:
> `Iceberg: Support applying equality deletes as a join`
> 
> I think broadcast join is a global property of the query. Is it expected that users will manually specify that they enable broadcast join?

From what I understand the default is `AUTOMATIC` and that allows broadcast. Is that not the case? I was getting broadcast joins when running explain plan.

yingsu00 (commented):
@jasonf20 Hi Jason, will you be able to post the plan after this change in the PR message? Thanks!

presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergEqualityDeleteAsJoin.java
```diff
@@ -0,0 +1,467 @@
...
+                            if (!partitionField.transform().isIdentity()) {
+                                if (!partitionField.transform().isIdentity()) {
+                                    Type partitionFieldType = partitionField.transform().getResultType(sourceField.type());
+                                    VariableReferenceExpression variableReference = variableAllocator.newVariable(partitionField.name(), toPrestoType(partitionFieldType, typeManager));
+                                    IcebergColumnHandle columnHandle = new IcebergColumnHandle(
+                                            ColumnIdentity.createColumnIdentity(partitionField.name(), partitionField.fieldId(), partitionFieldType),
+                                            toPrestoType(partitionFieldType, typeManager),
+                                            Optional.empty(),
+                                            PARTITION_KEY);
+                                    unselectedAssignmentsBuilder.put(variableReference, columnHandle);
```
hantangwangd, line 381: Are non-identity partition columns necessary to add to  `unselectedAssignmentsBuilder`? As I  understand, they would not be used in join equality clause with  equal delete table scan.

hantangwangd (commented):
Overall looks good to me. Some little problems to figure out.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergColumnHandle.java
hantangwangd: +1

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -304,6 +304,38 @@ Property Name                                          Description
                                                        this size will not be cached.
 ====================================================   =============================================================   ============
 
+Extra Hidden Metadata Columns
+----------------------------
+
+The iceberg connector exposes extra hidden metadata columns. You can query these
```
steveburnett, line 310:
```suggestion
The Iceberg connector exposes extra hidden metadata columns. You can query these
```

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -304,6 +304,38 @@ Property Name                                          Description
...
+as part of a SQL query by including them in your select statement.
```
steveburnett, line 311:
```suggestion
as part of a SQL query by including them in your SELECT statement.
```

steveburnett (changes requested):
Thanks for the documentation! Two small nits only.

presto-docs/src/main/sphinx/connector/iceberg.rst
jasonf20: thanks fixed, both (+ 1 more `iceberg`)

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergUtil.java
```diff
@@ -13,6 +13,7 @@
  */
 package com.facebook.presto.iceberg;
 
+import avro.shaded.com.google.common.collect.Iterators;
```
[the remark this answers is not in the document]
jasonf20: bad import by mistake, thanks for catching this.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergEqualityDeleteAsJoin.java
```diff
@@ -0,0 +1,467 @@
...
+                                }
+                            }
+                            else if (!selectedFields.contains(sourceField.fieldId())) {
```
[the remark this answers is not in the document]
jasonf20: This is the `sourceField` of the `Identity` partition field. The initial set difference is calculated on the partition field id which is different than the source field id for identity partitions.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergEqualityDeleteAsJoin.java
```diff
@@ -0,0 +1,467 @@
...
+            ImmutableMap.Builder<VariableReferenceExpression, ColumnHandle> assignmentsBuilder = ImmutableMap.<VariableReferenceExpression, ColumnHandle>builder()
+                    .put(dataSequenceNumberVariableReference, DATA_SEQUENCE_NUMBER_COLUMN_HANDLE)
+                    .putAll(unselectedAssignments)
+                    .putAll(node.getAssignments());
+            ImmutableList.Builder<VariableReferenceExpression> outputsBuilder = ImmutableList.builder();
+            outputsBuilder.addAll(node.getOutputVariables());
+            if (!node.getAssignments().containsKey(dataSequenceNumberVariableReference)) {
+                outputsBuilder.add(dataSequenceNumberVariableReference);
+            }
+            outputsBuilder.addAll(ImmutableList.copyOf(unselectedAssignments.keySet()));
```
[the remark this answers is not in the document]
jasonf20: It's already an ImmutableMap so I suppose it's not. Changed 👍

presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergEqualityDeleteAsJoin.java
jasonf20: They are added to the join clause. Only the metadata column "$data_sequence_number" isn't. (that column is used as a filter instead).

The test `testEqualityDeletesWithHiddenPartitions` covers this case.

jasonf20:
@yingsu00 added

presto-docs/src/main/sphinx/connector/iceberg.rst
steveburnett: Thanks!

steveburnett (changes requested):
Thanks for the quick reply and fixes! I found one more nit when I did a new local build of the docs for you to consider.

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -304,6 +304,38 @@ Property Name                                          Description
...
+
+.. code-block:: text
+
+             $path                    |  regionkey
+     ---------------------------------+-----------
+      /full/path/to/file/file.parquet | 2
+
+``$data_sequence_number`` column
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+* ``$data_sequence_number``: The Iceberg data sequence number in which this row as added
```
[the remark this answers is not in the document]
jasonf20: Yea, typo. Fixed, thanks!

steveburnett (approved):
LGTM! (docs)

Pulled the branch again, new local build of the docs, everything looks good. Thanks!

presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergEqualityDeleteAsJoin.java
hantangwangd: Oh, I overlooked the return statement after identity partition column branch in `DeleteSetInfo.allFields(Schema schema)`.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergEqualityDeleteAsJoin.java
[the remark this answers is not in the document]
hantangwangd: You are right, I didn't notice that.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedTestBase.java
```diff
@@ -859,68 +868,168 @@ public void testTableWithPositionDelete(String fileFormat)
...
+            writeEqualityDeleteToNationTable(icebergTable, ImmutableMap.of("regionkey", 1L), ImmutableMap.of("nationkey_bucket", partitionsBlock.getInt(i)));
+            String updatedPartitions = partitions.stream().limit(i + 1).map(Object::toString).collect(Collectors.joining(",", "(", ")"));
+            assertQuery(session, "SELECT * FROM " + tableName, "SELECT * FROM nation WHERE NOT(regionkey = 1 AND nationkey IN " + updatedPartitions + ")");
+        }
+        testCheckDeleteFiles(icebergTable, partitions.size(), partitions.stream().map(i -> EQUALITY_DELETES).collect(Collectors.toList()));
+        assertQuery(session, "SELECT * FROM " + tableName, "SELECT * FROM nation WHERE regionkey != 1");
+        assertQuery(session, "SELECT nationkey FROM " + tableName, "SELECT nationkey FROM nation WHERE regionkey != 1");
+        assertQuery(session, "SELECT name FROM " + tableName, "SELECT name FROM nation WHERE regionkey != 1");
+    }
+
```
hantangwangd, line 968:
```suggestion

    @Test(dataProvider = "equalityDeleteOptions")
    public void testEqualityDeletesWithHiddenPartitionsEvolution()
            throws Exception
    {
        Session session = deleteAsJoinEnabled(true);
        String tableName = "test_v2_row_delete_" + randomTableSuffix();
        assertUpdate("CREATE TABLE " + tableName + "(a int, b varchar) with (format = 'parquet')");
        assertUpdate("insert into " + tableName + " values (1, '1001'), (2, '1002'), (3, '1003')", 3);
        Table icebergTable = updateTable(tableName);
        writeEqualityDeleteToNationTable(icebergTable, ImmutableMap.of("a", 2));
        assertQuery(session, "SELECT * FROM " + tableName, "values(1, '1001'), (3, '1003')");

        assertUpdate("alter TABLE " + tableName + " add column c int with(partitioning = 'bucket(2)')");
        assertUpdate("insert into " + tableName + " values (6, '1004', 1), (6, '1006', 2)", 2);
        icebergTable = updateTable(tableName);
        PartitionTransforms.ColumnTransform columnTransform = PartitionTransforms.getColumnTransform(icebergTable.spec().fields().get(0), INTEGER);
        BlockBuilder builder = BIGINT.createFixedSizeBlockBuilder(1);
        List<Integer> partitions = Arrays.asList(1, 2);
        partitions.forEach(p -> BIGINT.writeLong(builder, p));
        Block partitionsBlock = columnTransform.getTransform().apply(builder.build());
        writeEqualityDeleteToNationTable(icebergTable, ImmutableMap.of("a", 6, "c", 2),
                ImmutableMap.of("c_bucket", partitionsBlock.getInt(0)));
        assertQuery(session, "SELECT * FROM " + tableName, "values(1, '1001', NULL), (3, '1003', NULL), (6, '1004', 1)");

        assertUpdate("alter TABLE " + tableName + " add column d varchar with(partitioning = 'truncate(2)')");
        assertUpdate("insert into " + tableName + " values (6, '1004', 1, 'th001'), (6, '1006', 2, 'th002')", 2);
        icebergTable = updateTable(tableName);

        writeEqualityDeleteToNationTable(icebergTable, ImmutableMap.of("a", 6, "c", 1),
                ImmutableMap.of("c_bucket", partitionsBlock.getInt(0), "d_trunc", "th"));
        testCheckDeleteFiles(icebergTable, 3, ImmutableList.of(EQUALITY_DELETES, EQUALITY_DELETES, EQUALITY_DELETES));
        assertQuery(session, "SELECT * FROM " + tableName, "values(1, '1001', NULL, NULL), (3, '1003', NULL, NULL), (6, '1004', 1, NULL), (6, '1006', 2, 'th002')");
    }
```

Seems in this test case the record `(6, '1004', 1, NULL)` would be wrongly deleted. The reason might be that, when we try to scan files for equality delete table of specific delete schema  by IcebergUtil.getDeleteFiles(...), we should take the partition columns of delete schema as well. Or different delete schema with same equality delete columns and different partition columns would lead in some problems.

jasonf20: Great catch! thanks for writing a test to reproduce this! I pushed a fix for this.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergTableHandle.java
```diff
@@ -33,6 +33,7 @@ public class IcebergTableHandle
     private final TupleDomain<IcebergColumnHandle> predicate;
     private final boolean snapshotSpecified;
     private final Optional<String> tableSchemaJson;
+    private final Optional<Integer> partitionSpecId;
```
hantangwangd, line 36: Due to renaming partition column would cause partition spec evolution, maybe the `partitionSpecId` here should be `Optional<Set<Integer>>` or something like this. For renaming partition column would not change the partition column id and it's source column id in schema, but it would create a new partition spec.

jasonf20: Another great point! Didn't think about renaming, changed to use partition field ids.

hantangwangd (approved):
LGTM! Thanks for the great job. Please check the test failures.

tdcmeehan:
@jasonf20 this looks good to me, but can you please squash the fixup commits into their respective base commits so I can give it one last look?

tdcmeehan:
> > `Iceberg: Support applying equality deletes as a join`
> > I think broadcast join is a global property of the query. Is it expected that users will manually specify that they enable broadcast join?
> 
> From what I understand the default is `AUTOMATIC` and that allows broadcast. Is that not the case? I was getting broadcast joins when running explain plan.
> 
> I think I corrected all other comments.

I think this is safe, because Iceberg gives us the statistics we need to do this safely and consistently.

But I would add a note in the documentation explaining this.  Which reminds me that I believe this session property itself should be documented.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergSessionProperties.java
```diff
@@ -317,6 +318,11 @@ public IcebergSessionProperties(
                                 "value of 1 means a single record is equivalent to 1 millisecond of " +
                                 "time difference.",
                         icebergConfig.getStatisticSnapshotRecordDifferenceWeight(),
+                        false),
+                booleanProperty(
+                        DELETE_AS_JOIN_PUSHDOWN_ENABLED,
+                        "When enabled equality delete row filtering will be pushed down into a join.",
+                        icebergConfig.isPushdownFilterEnabled(),
```
tdcmeehan, line 325: I believe this config value only works with the native workers at present, and that's still under development.  Let's add a new config property.

jasonf20: Why would it require the native workers? I've been testing this with the Java workers I believe.

jasonf20:
@tdcmeehan I thought the param documentation is generated from the description in the code, but If that's not the case where should I go to add documentation?

Pushed a squashed version as well.

steveburnett:
> @tdcmeehan I thought the param documentation is generated from the description in the code, but If that's not the case where should I go to add documentation?
> 
> Pushed a squashed version as well.

I believe that session property documentation can be added to [Properties Reference](https://prestodb.io/docs/0.284/admin/properties.html).

jasonf20:
> I believe that session property documentation can be added to [Properties Reference](https://prestodb.io/docs/0.284/admin/properties.html).

Hmm I don't see other connector specific properties documented there, also I don't see other iceberg properties on that page. I can add it but it doesn't seem in line with the other iceberg session properties. Let me know what you think

steveburnett:
> > I believe that session property documentation can be added to [Properties Reference](https://prestodb.io/docs/0.284/admin/properties.html).
> 
> Hmm I don't see other connector specific properties documented there, also I don't see other iceberg properties on that page. I can add it but it doesn't seem in line with the other iceberg session properties. Let me know what you think

Yes, you're right! [Iceberg Connector](https://prestodb.io/docs/0.284/connector/iceberg.html#configuration-properties) doc would probably be better, similar to the [session property in the Hudi Connector doc](https://prestodb.io/docs/0.284/connector/hudi.html#configuration).

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -269,6 +269,19 @@ and a file system location of ``s3://test_bucket/test_schema/test_table``:
...
     )
 
+Session Properties
+-------------------
+
+Session properties set behaviour changes for queries executed within the given session.
+The following list is a partial list of the session properties available in the Iceberg connector.
+
+============================================= ===============================================================
+Property Name                                 Description
```
steveburnett, line 279: Consider adding a Default column and a default value, similar to the [Configuration Properties](https://prestodb.io/docs/0.284/connector/iceberg.html#configuration-properties) in the Iceberg Connector doc and the [session property in the Hudi Connector](https://prestodb.io/docs/0.284/connector/hudi.html#configuration) doc.

steveburnett (changes requested):
Thanks for adding the session property documentation! I had a couple of suggestions.

jasonf20:
@steveburnett Thanks, Done.

steveburnett (approved):
LGTM! (docs)

jasonf20:
@steveburnett Sorry for driving you mad 🤦  but I made some more doc changes after adding the value as a connector property and setting the session property to default to the connector property after discussing this with @tdcmeehan

steveburnett:
> @steveburnett Sorry for driving you mad 🤦 but I made some more doc changes after adding the value as a connector property and setting the session property to default to the connector property after discussing this with @tdcmeehan

You're really not driving me mad, honestly! I can take another look, no problem.

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -373,6 +388,38 @@ Metastore cache only caches schema and table names. Other metadata would be fetc
     hive.metastore-refresh-interval=3d
     hive.metastore-cache-maximum-size=10000000
 
+Extra Hidden Metadata Columns
```
steveburnett, line 391: I didn't bring it up originally when this topic was added to the documentation, but I am curious about the first word here: why "Extra"? Are there non-extra hidden metadata columns or tables? I'm not requesting a change here at the moment, I'm more curious about why than anything.

steveburnett (changes requested):
Thanks for the additional documentation! I had a couple of nits, and one question that's more curiosity than anything else. Let me know what you think, please.

presto-docs/src/main/sphinx/connector/iceberg.rst
jasonf20: I kept with the convention of `Extra Hidden Metadata Tables` below this section. I agree the `Extra` phrasing is a little odd. Perhaps `Hidden` would be better or nothing at all.

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -269,6 +272,18 @@ and a file system location of ``s3://test_bucket/test_schema/test_table``:
...
 
+Session Properties
+-------------------
+
+Session properties set behaviour changes for queries executed within the given session.
+
+============================================= ======================================================================
+Property Name                                 Description
+============================================= ======================================================================
+``iceberg.delete_as_join_pushdown_enabled``   Overrides the behaviour of the connector property
```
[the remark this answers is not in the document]
jasonf20: I didn't set the default since it's an override for the connector property and the default is documented there, do you think it should be here as well?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergSessionProperties.java
jasonf20: Oh I misunderstood you, I see you were talking about the default I passed here. That was a mistake, thanks for catching that!

presto-docs/src/main/sphinx/connector/iceberg.rst
steveburnett: Thanks for the explanation, and the note about hive.rst. I'll worry about it making sense later.

presto-docs/src/main/sphinx/connector/iceberg.rst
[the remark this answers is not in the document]
steveburnett: Yes, I think so. Here's my reasoning: 

* The session property (iceberg.delete_as_join_pushdown_enabled) is a separate property from the configuration property (iceberg.delete-as-join-pushdown-enabled), so I would like the doc for the session property to be self-contained. 

* As it is written now, the reader must infer that the values to set the session property to are the same as the configuration property - but it isn't stated that that is the case here. They'll probably figure it out, but why make them? Alternatively, could add "The values for `iceberg.delete_as_join_pushdown_enabled` are the same as `iceberg.delete-as-join-pushdown-enabled` ", but including the Default value for the session property _with_ the session property is shorter, less ambiguous, and doesn't require the reader to go find `iceberg.delete-as-join-pushdown-enabled` in the other table.

That was a lot of words but I wanted to explain why, and an alternative I'd considered. Does this make sense?

steveburnett (approved):
LGTM (docs)

presto-iceberg/src/main/java/com/facebook/presto/iceberg/equalitydeletes/EqualityDeletesSplitSource.java
```diff
@@ -0,0 +1,116 @@
...
+    {
+        try {
+            deleteFiles.close();
+            // TODO: remove this after org.apache.iceberg.io.CloseableIterator'withClose
+            // correct release resources holds by iterator.
+            // (and make it final)
+            deleteFiles = CloseableIterator.empty();
+        }
+        catch (IOException e) {
+            throw new PrestoException(GENERIC_INTERNAL_ERROR, e);
```
tdcmeehan, line 92: `ICEBERG_FILESYSTEM_ERROR`?

tdcmeehan (approved):
LGTM % a few more nits and % remaining feedback from @steveburnett.

CC: @mlyublena

jasonf20:
Thanks @tdcmeehan. Pushed an update.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergConfig.java
```diff
@@ -235,6 +236,19 @@ public boolean isPushdownFilterEnabled()
         return pushdownFilterEnabled;
     }
 
+    @Config("iceberg.delete-as-join-pushdown-enabled")
```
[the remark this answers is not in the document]
jasonf20: Yea I think rewrite is more intuitive and explains what is being done a little better.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergUtil.java
```diff
@@ -746,4 +764,120 @@ public static void loadCachingProperties(Map<String, String> properties, Iceberg
...
+     * @param requestedSchema If provided, only delete files with this schema will be provided
+     */
+    public static CloseableIterable<DeleteFile> getDeleteFiles(Table table,
+            long snapshot,
+            TupleDomain<IcebergColumnHandle> filter,
+            Optional<Set<Integer>> requestedPartitionSpec,
+            Optional<Set<Integer>> requestedSchema)
+    {
+        Expression filterExpression = toIcebergExpression(filter);
+        CloseableIterator<FileScanTask> fileTasks = table.newScan().useSnapshot(snapshot).filter(filterExpression).planFiles().iterator();
```
tdcmeehan, line 792: Small nit, but just to future proof this code, I'd do one of two things:

1) Add a little check to ensure that the `iterator()` method is called just once.
2) Make this lazy and move the scan planning into the `iterator()` method.

It's unexpected that multiple calls to `iterator()` will not iterate from the beginning.  My preference is on option 2.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergUtil.java
```diff
@@ -746,4 +764,120 @@ public static void loadCachingProperties(Map<String, String> properties, Iceberg
...
+    private static class DeleteFilesIterator
+            implements CloseableIterator<DeleteFile>
+    {
+        private final Set<String> seenFiles = new HashSet<>();
+        private final Map<Integer, PartitionSpec> partitionSpecsById;
+        private final CloseableIterator<FileScanTask> fileTasks;
+        private final Optional<Set<Integer>> requestedPartitionSpec;
+        private final Optional<Set<Integer>> requestedSchema;
+        Iterator<DeleteFile> currentDeletes = emptyIterator();
+        DeleteFile currentFile;
```
tdcmeehan, line 820:
```suggestion
        Iterator<DeleteFile> currentDeletes = emptyIterator();
        DeleteFile currentFile;
```
```suggestion
        private Iterator<DeleteFile> currentDeletes = emptyIterator();
        private DeleteFile currentFile;
```

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergUtil.java
```diff
@@ -746,4 +764,120 @@ public static void loadCachingProperties(Map<String, String> properties, Iceberg
...
+
+        DeleteFilesIterator(Map<Integer, PartitionSpec> partitionSpecsById,
```
tdcmeehan, line 822:
```suggestion
        private DeleteFilesIterator(Map<Integer, PartitionSpec> partitionSpecsById,
```

tdcmeehan (commented):
A few last comments

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergUtil.java
jasonf20: I changed this line so it doesn't call `.iterator()` and I only call that inside the `iterator()` method. So it should respect the `Iterable` interface now. Does that solution work?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergUtil.java
```diff
@@ -746,4 +764,120 @@ public static void loadCachingProperties(Map<String, String> properties, Iceberg
...
+                    requestedPartitionSpec.get().equals(partitionSpecsById.get(file.specId()).fields().stream().map(PartitionField::fieldId).collect(Collectors.toSet()));
+            return matchesPartition && (file.content() == FileContent.POSITION_DELETES ||
+                    !requestedSchema.isPresent() || requestedSchema.get().equals(ImmutableSet.copyOf(file.equalityFieldIds())));
+        }
+
+        @Override
+        public void close()
+                throws IOException
+        {
+            fileTasks.close();
```
tdcmeehan, line 880: I believe this might need to be explicitly set to `emptyIterator` as well, similar to https://github.com/prestodb/presto/pull/21605/commits/18f4226603ceb06659500254e4e3c1e9a71c4f86#r1442237365

yingsu00 (commented):
@jasonf20 Thanks for uploading the query plan. Could you please add another actual plan for multiple equality delete files with different equality ids? You can use "explain analyze" to get an actual plan.

Also, I wonder if you could expand the tests to test its interplay with "iceberg.pushdown_filter_enabled" with ORC files (Java Parquet doesn't support filter pushdown). It'll be great if you can add an actual plan for that case too.

presto-spi/src/main/java/com/facebook/presto/spi/plan/ConnectorJoinNode.java
```diff
@@ -0,0 +1,139 @@
...
+        super(Optional.empty(), id, statsEquivalentPlanNode);
+        this.sources = requireNonNull(sources, "sources is null");
+        this.type = requireNonNull(type, "type is null");
+        this.criteria = requireNonNull(criteria, "criteria is null");
+        this.filters = requireNonNull(filters, "filters is null");
+        this.outputVariables = requireNonNull(outputVariables, "outputVariables is null");
+    }
+
+    @Override
+    @JsonProperty
```
feilong-liu, line 54: Is this necessary? 
If Json serialization is necessary, please add a JsonCreator construction method, if not remove these 
JsonProperty annotations.

Similar for other fields

yingsu00:
@jasonf20 Will you be able to do a small comparative TPCDS run with this equality join turned on/off and with/without equality delete files to make sure there is no regression? Without any equality delete files in the data, the before/after plans should be the same and performance should also be same. With some equality delete files, the new plans have more joins and may affect the join reordering later on. If we can show there are no regressions I think the PR is good to go. Thank you!

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/ApplyConnectorOptimization.java
```diff
@@ -292,4 +301,26 @@ private static <T> boolean containsAll(Set<T> container, Collection<T> test)
...
+                    Optional.empty(),
+                    node.getType(),
+                    context.rewrite(node.getSources().get(0)),
+                    context.rewrite(node.getSources().get(1)),
+                    ImmutableList.copyOf(node.getCriteria()),
+                    node.getOutputVariables(),
+                    node.getFilters().stream().reduce((a, b) -> new SpecialFormExpression(SpecialFormExpression.Form.AND, BooleanType.BOOLEAN, a, b)),
+                    Optional.empty(),
+                    Optional.empty(),
+                    Optional.empty(),
```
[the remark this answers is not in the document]
jasonf20: The `ConnectorJoinNode` currently doesn't allow setting it since I just relied on automatic. I think it can be added when a connector wants to set it?

presto-spi/src/main/java/com/facebook/presto/spi/plan/ConnectorJoinNode.java
jasonf20: Removed the json attributes.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergUtil.java
jasonf20: Done 👍

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/ApplyConnectorOptimization.java
```diff
@@ -292,4 +301,26 @@ private static <T> boolean containsAll(Set<T> container, Collection<T> test)
...
+                    Collections.emptyMap());
```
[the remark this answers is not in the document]
jasonf20: changed

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/ApplyConnectorOptimization.java
```diff
@@ -292,4 +301,26 @@ private static <T> boolean containsAll(Set<T> container, Collection<T> test)
...
+
+    private static class ConnectorToInternalJoinRewriter
+            extends SimplePlanRewriter<Void>
+    {
+        @Override
+        public PlanNode visitConnectorJoinNode(ConnectorJoinNode node, RewriteContext<Void> context)
+        {
+            return new JoinNode(node.getSourceLocation(),
+                    node.getId(),
+                    Optional.empty(),
```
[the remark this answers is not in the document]
jasonf20: changed

jasonf20:
Hi @yingsu00 

Thanks for the comments, Adding the explain plan to the end of this comment.

Regarding `iceberg.pushdown_filter_enabled`. Based on your comment, I changed the equality delete scan node to pass TupleDomain.all() so this won't affect the generated scan nodes and is now unrelated to the filtering that this may perform on the original scan node. In theory a subset of the filter could be applied to the equality delete files, but since these are typically quite small this optimization is negligible. I think if we want add this it can be done in a seperate PR and it doesn't introduce a regression since the existing implementation doesn't filter them either. There are a few issues with adding a spec to validate this: 
1. "pushdown_filter_enabled" only works in native mode and the current spec is in Java. 
2. Equality delete files are always written as Parquet currently and the code that writes ORC equality delete files is not available due to missing packages and adding the package is non-trivial since there are version conflicts with other ORC packages. Given that the delete scan node passes `TupleDomain.all()` it feels safe to assume that the parameter won't interact with these changes without explicitly adding a spec for that.  

Regarding TPCDS testing, there is currently no API to create equality delete files in Presto so there isn't a way to run this test with TPCDS. The rewriter explictly returns the current node when there are no equality delete files. So this code shouldn't cause any change to the explain plan for queries without Equality Delete files. For tables with Equality Delete files (which can't be created using presto) the existing implementation is almost unusable for even small amounts of deletes (see initial perfomance boost in PR message). The possibility of a regression when Equality Delete files exist is mitigated by the fact that previously they were not really queryable.

Given this information, how should we proceed?


Explain plan with multiple delete schemas:
```
MaterializedResult{rows=[[Fragment 1 [HASH]
    CPU: 159.75ms, Scheduled: 2.08s, Input: 27 rows (3.37kB); per task: avg.: 6.75 std.dev.: 4.60, Output: 18 rows (2.18kB), 4 tasks
    Output layout: [nationkey, name, regionkey, comment, $data_sequence_number_15, nationkey_17, $data_sequence_number_16, regionkey_19, $data_sequence_number_18, regionkey_22, nationkey_21, $data_sequence_number_20]
    Output partitioning: SINGLE []
    Stage Execution Strategy: UNGROUPED_EXECUTION
    - Filter[PlanNodeId 132][filterPredicate = IS_NULL(COALESCE($data_sequence_number_16, $data_sequence_number_18, $data_sequence_number_20))] => [nationkey:bigint, name:varchar, regionkey:bigint, comment:varchar, $data_sequence_number_15:bigint, nationkey_17:bigint, $data_sequence_number_16:bigint, regionkey_19:bigint, $data_sequence_number_18:bigint, regionkey_22:bigint, nationkey_21:bigint, $data_sequence_number_20:bigint]
            Estimates: {source: CostBasedSourceInfo, rows: ? (?), cpu: ?, memory: 241.00, network: ?}
            CPU: 2.00ms (0.89%), Scheduled: 18.00ms (0.39%), Output: 18 rows (2.18kB)
            Input avg.: 1.56 rows, Input std.dev.: 198.51%
        - LeftJoin[PlanNodeId 131][("regionkey" = "regionkey_22") AND ("nationkey" = "nationkey_21") AND ($data_sequence_number_15) < ($data_sequence_number_20)][$hashvalue_32, $hashvalue_33] => [nationkey:bigint, name:varchar, regionkey:bigint, comment:varchar, $data_sequence_number_15:bigint, nationkey_17:bigint, $data_sequence_number_16:bigint, regionkey_19:bigint, $data_sequence_number_18:bigint, regionkey_22:bigint, nationkey_21:bigint, $data_sequence_number_20:bigint]
                Estimates: {source: CostBasedSourceInfo, rows: ? (?), cpu: ?, memory: 241.00, network: ?}
                CPU: 92.00ms (41.07%), Scheduled: 1.43s (31.00%), Output: 25 rows (3.54kB)
                Left (probe) Input avg.: 1.56 rows, Input std.dev.: 198.51%
                Right (build) Input avg.: 0.06 rows, Input std.dev.: 387.30%
                Distribution: PARTITIONED
                SortExpression[$data_sequence_number_20]
            - Project[PlanNodeId 424][projectLocality = LOCAL] => [nationkey:bigint, name:varchar, regionkey:bigint, comment:varchar, $data_sequence_number_15:bigint, nationkey_17:bigint, $data_sequence_number_16:bigint, regionkey_19:bigint, $data_sequence_number_18:bigint, $hashvalue_32:bigint]
                    Estimates: {source: CostBasedSourceInfo, rows: ? (?), cpu: ?, memory: 130.00, network: ?}
                    CPU: 3.00ms (1.34%), Scheduled: 8.00ms (0.17%), Output: 25 rows (3.48kB)
                    Input avg.: 1.56 rows, Input std.dev.: 198.51%
                    $hashvalue_32 := combine_hash(combine_hash(BIGINT'0', COALESCE($operator$hash_code(regionkey), BIGINT'0')), COALESCE($operator$hash_code(nationkey), BIGINT'0')) (1:31)
                - LeftJoin[PlanNodeId 129][("regionkey" = "regionkey_19") AND ($data_sequence_number_15) < ($data_sequence_number_18)][$hashvalue, $hashvalue_29] => [nationkey:bigint, name:varchar, regionkey:bigint, comment:varchar, $data_sequence_number_15:bigint, nationkey_17:bigint, $data_sequence_number_16:bigint, regionkey_19:bigint, $data_sequence_number_18:bigint]
                        Estimates: {source: CostBasedSourceInfo, rows: ? (?), cpu: ?, memory: 130.00, network: ?}
                        CPU: 48.00ms (21.43%), Scheduled: 533.00ms (11.57%), Output: 25 rows (3.33kB)
                        Left (probe) Input avg.: 1.56 rows, Input std.dev.: 198.51%
                        Right (build) Input avg.: 0.06 rows, Input std.dev.: 387.30%
                        Distribution: PARTITIONED
                        SortExpression[$data_sequence_number_18]
                    - RemoteSource[2] => [nationkey:bigint, name:varchar, regionkey:bigint, comment:varchar, $data_sequence_number_15:bigint, nationkey_17:bigint, $data_sequence_number_16:bigint, $hashvalue:bigint]
                            CPU: 0.00ns (0.00%), Scheduled: 0.00ns (0.00%), Output: 25 rows (3.30kB)
                            Input avg.: 1.56 rows, Input std.dev.: 198.51%
                    - LocalExchange[PlanNodeId 341][HASH][$hashvalue_29] (regionkey_19) => [regionkey_19:bigint, $data_sequence_number_18:bigint, $hashvalue_29:bigint]
                            Estimates: {source: CostBasedSourceInfo, rows: 1 (238B), cpu: 251.00, memory: 0.00, network: 65.00}
                            CPU: 0.00ns (0.00%), Scheduled: 28.00ms (0.61%), Output: 1 row (27B)
                            Input avg.: 0.06 rows, Input std.dev.: 387.30%
                        - RemoteSource[5] => [regionkey_19:bigint, $data_sequence_number_18:bigint, $hashvalue_30:bigint]
                                CPU: 0.00ns (0.00%), Scheduled: 0.00ns (0.00%), Output: 1 row (27B)
                                Input avg.: 0.06 rows, Input std.dev.: 387.30%
            - LocalExchange[PlanNodeId 342][HASH][$hashvalue_33] (regionkey_22, nationkey_21) => [regionkey_22:bigint, nationkey_21:bigint, $data_sequence_number_20:bigint, $hashvalue_33:bigint]
                    Estimates: {source: CostBasedSourceInfo, rows: 1 (275B), cpu: 564.00, memory: 0.00, network: 120.00}
                    CPU: 0.00ns (0.00%), Scheduled: 5.00ms (0.11%), Output: 1 row (36B)
                    Input avg.: 0.06 rows, Input std.dev.: 387.30%
                - Project[PlanNodeId 426][projectLocality = LOCAL] => [regionkey_22:bigint, nationkey_21:bigint, $data_sequence_number_20:bigint, $hashvalue_34:bigint]
                        Estimates: {source: CostBasedSourceInfo, rows: 1 (275B), cpu: 453.00, memory: 0.00, network: 120.00}
                        CPU: 0.00ns (0.00%), Scheduled: 0.00ns (0.00%), Output: 1 row (36B)
                        Input avg.: 0.06 rows, Input std.dev.: 387.30%
                    - RemoteSource[6] => [regionkey_22:bigint, nationkey_21:bigint, $data_sequence_number_20:bigint, $hashvalue_34:bigint, $hashvalue_35:bigint]
                            CPU: 0.00ns (0.00%), Scheduled: 0.00ns (0.00%), Output: 1 row (45B)
                            Input avg.: 0.06 rows, Input std.dev.: 387.30%

Fragment 2 [HASH]
    CPU: 70.26ms, Scheduled: 2.43s, Input: 26 rows (3.14kB); per task: avg.: 6.50 std.dev.: 1.12, Output: 25 rows (3.30kB), 4 tasks
    Output layout: [nationkey, name, regionkey, comment, $data_sequence_number_15, nationkey_17, $data_sequence_number_16, $hashvalue_28]
    Output partitioning: HASH [regionkey][$hashvalue_28]
    Stage Execution Strategy: UNGROUPED_EXECUTION
    - Project[PlanNodeId 422][projectLocality = LOCAL] => [nationkey:bigint, name:varchar, regionkey:bigint, comment:varchar, $data_sequence_number_15:bigint, nationkey_17:bigint, $data_sequence_number_16:bigint, $hashvalue_28:bigint]
            Estimates: {source: CostBasedSourceInfo, rows: ? (?), cpu: ?, memory: 65.00, network: 1,951.00}
            CPU: 6.00ms (2.68%), Scheduled: 28.00ms (0.61%), Output: 25 rows (3.30kB)
            Input avg.: 1.56 rows, Input std.dev.: 179.56%
            $hashvalue_28 := combine_hash(BIGINT'0', COALESCE($operator$hash_code(regionkey), BIGINT'0')) (1:31)
        - LeftJoin[PlanNodeId 127][("nationkey" = "nationkey_17") AND ($data_sequence_number_15) < ($data_sequence_number_16)][$hashvalue_23, $hashvalue_25] => [nationkey:bigint, name:varchar, regionkey:bigint, comment:varchar, $data_sequence_number_15:bigint, nationkey_17:bigint, $data_sequence_number_16:bigint]
                Estimates: {source: CostBasedSourceInfo, rows: ? (?), cpu: ?, memory: 65.00, network: 1,951.00}
                CPU: 51.00ms (22.77%), Scheduled: 2.30s (50.02%), Output: 25 rows (3.01kB)
                Left (probe) Input avg.: 1.56 rows, Input std.dev.: 179.56%
                Right (build) Input avg.: 0.06 rows, Input std.dev.: 387.30%
                Distribution: PARTITIONED
                SortExpression[$data_sequence_number_16]
            - RemoteSource[3] => [nationkey:bigint, name:varchar, regionkey:bigint, comment:varchar, $data_sequence_number_15:bigint, $hashvalue_23:bigint]
                    CPU: 0.00ns (0.00%), Scheduled: 0.00ns (0.00%), Output: 25 rows (3.11kB)
```

yingsu00:
@jasonf20 Thanks for your information. Regarding TPCDS testing, I thought you have other engines that can generate equality deletes, but just learnt that it's actually hard for you to generate them now. Prestissimo will support equality deletes but it's not there yet. So I'll just withdraw the ask of TPCDS tests.

What I was concerned was that this may affect the overall query shape, since it's adding some new join nodes in the logical planning phase, and the join reordering rule will be performed later and may introduce some uncertainty. Sometimes the plan shape changes because of some seemingly unrelated issues and the result may not be ideal. That's why I was hoping you could run the TPCDS benchmark to rule out the possibility of some unexpected plan change. But if testing TPCDS is hard, we will test it ourselves in the future.

jasonf20:
Thanks @yingsu00. I did validate the query shape is unchanged when disabled for queries on tables that I had with equality deletes. But not TPCDS tables, that isn't something I have access to. In my tests, when there are no equality deletes and/or when the flag is disabled the explain plan I got was the unchanged. Of course this is limited to the queries I tested but it does have some coverage. I think static analysis of the start of the rewriter can also help validate this.

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/ApplyConnectorOptimization.java
```diff
@@ -292,4 +300,26 @@ private static <T> boolean containsAll(Set<T> container, Collection<T> test)
...
+        {
+            return new JoinNode(node.getSourceLocation(),
+                    node.getId(),
+                    node.getStatsEquivalentPlanNode(),
+                    node.getType(),
+                    context.rewrite(node.getSources().get(0)),
+                    context.rewrite(node.getSources().get(1)),
+                    ImmutableList.copyOf(node.getCriteria()),
+                    node.getOutputVariables(),
+                    node.getFilters().stream().reduce((a, b) -> new SpecialFormExpression(SpecialFormExpression.Form.AND, BooleanType.BOOLEAN, a, b)),
```
aaneja, line 318: You can use https://github.com/prestodb/presto/blob/55cc36f416ce35fd3cfef033b0292aad880f95ee/presto-expressions/src/main/java/com/facebook/presto/expressions/LogicalRowExpressions.java#L121-L124

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/ApplyConnectorOptimization.java
```diff
@@ -292,4 +300,26 @@ private static <T> boolean containsAll(Set<T> container, Collection<T> test)
         }
         return true;
     }
+
+    private static class ConnectorToInternalJoinRewriter
+            extends SimplePlanRewriter<Void>
+    {
+        @Override
+        public PlanNode visitConnectorJoinNode(ConnectorJoinNode node, RewriteContext<Void> context)
```
aaneja, line 308: Maybe I'm missing something obvious - but why do we need to have a new type `ConnectorJoinNode` ? Why can't we move the `JoinNode` type to the SPI ?

aaneja (commented):
Still reviewing, will come back to this

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/ApplyConnectorOptimization.java
jasonf20: When I tried I got stuck on a lot of classes that needed to be moved as well. But I don't remember which ones at the moment, I could give it another go. However there is also some value in having a dedicated class acting as an adapter to shield the SPI from internal changes.

jasonf20: I think it was `FunctionAndTypeManager` which isn't available to the SPI. I guess we can add a helper class that is available to the callers of `getSortExpressionContext`. 
There were more classes but most of them needed to be moved to the SPI anyway for the adapter node. 

It's just a one liner.

aaneja: Okay I think this is a fair compromise

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/ApplyConnectorOptimization.java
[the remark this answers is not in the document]
feilong-liu: But this will remove the decision from DetermineJoinDistributionType optimizer?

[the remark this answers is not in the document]
jasonf20: The `ConnectorJoinNode` is converted as soon as it's returned from the connector in `ApplyConnectorOptimization` (logical). Which is before `DetermineJoinDistributionType`.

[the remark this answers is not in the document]
feilong-liu: How about ApplyConnectorOptimization (physical)? 

> The ConnectorJoinNode currently doesn't allow setting it since I just relied on automatic

Why it does not allow? Can you elaborate?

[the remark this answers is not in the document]
hantangwangd: As I understand, ApplyConnectorOptimization(physical) did not contains optimizer `IcebergEqualityDeleteAsJoin`, so it would never meet and rewrite `ConnectorJoinNode`.

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/ApplyConnectorOptimization.java
```diff
@@ -292,4 +299,26 @@ private static <T> boolean containsAll(Set<T> container, Collection<T> test)
...
+        {
+            return new JoinNode(node.getSourceLocation(),
+                    node.getId(),
+                    node.getStatsEquivalentPlanNode(),
+                    node.getType(),
+                    context.rewrite(node.getSources().get(0)),
+                    context.rewrite(node.getSources().get(1)),
+                    ImmutableList.copyOf(node.getCriteria()),
+                    node.getOutputVariables(),
+                    node.getFilters().stream().reduce(LogicalRowExpressions::and),
```
aaneja, line 317: nit: There is a variant which has a `Collection` as its input, so `Optional.of(LogicalRowExpressions.and(node.getFilters()))` will suffice

aaneja (approved):

yingsu00 (approved):

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/ApplyConnectorOptimization.java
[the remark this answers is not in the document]
jasonf20: Exactly. 
If you want me to allow it I can move `DistributionType` to the SPI package as well. I just didn't need it so I didn't want to move more classes than needed. But let me know if you prefer I move it as well.

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/ApplyConnectorOptimization.java
jasonf20: It ends up being a little more code since I need to check if the filters list is empty. 
```
node.getFilters().isEmpty() ? Optional.empty() : Optional.of(LogicalRowExpressions.and(node.getFilters())),
``` 
Which do you prefer?

jasonf20: I went with the second option so that it will create a flat Expression instead of a nested one.

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/ApplyConnectorOptimization.java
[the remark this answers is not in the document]
feilong-liu: Sounds like too much assumption and dependencies behind this.

> move DistributionType to the SPI package as well

I think it will be great, if there are no other reasons which prevents it.

tdcmeehan:
Thank you @jasonf20 for this very nice contribution, and thank you @feilong-liu @hantangwangd @steveburnett @aaneja @yingsu00 for the thorough reviews.

yingsu00:
@jasonf20 I saw you have another PR for Trino: [cebergPlugin: Performance improvements for Equality Delete files](https://github.com/trinodb/trino/pull/18397/files#top) which doesn't need to add joins. How about the comparisons of the two approches? cc @tdcmeehan

PR #21616, opened 2023-12-29
Fix filtering by subfield of RowType column in Iceberg table

hantangwangd:
## Description

When we try to query iceberg table using filters about the subfield of row type columns, the filtering condition would fail. For example, with an Iceberg table as follow:

``create table iceberg.default.test(id int, r row(a int, b varchar));``

When we query the table with sql statement ``select * from iceberg.default.test where r.a = 1``, we would get all rows in the table.

The PR fix this problem.

## Test Plan

 - Newly added test case IcebergDistributedSmokeTestBase.testFilterBySubfieldOfRowType()

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedSmokeTestBase.java
+++ b/presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedSmokeTestBase.java
@@ -893,6 +893,23 @@ public class IcebergDistributedSmokeTestBase
         dropTable(session, tableName);
     }
 
+    @Test
+    public void testFilterBySubfieldOfRowType()
+    {
+        Session session = getSession();
+        String tableName = "test_filter_by_subfieldofrow";
+
+        assertUpdate("CREATE TABLE " + tableName + " (id integer, r row(a integer, b varchar))");
+        assertUpdate("INSERT INTO " + tableName + " VALUES (1, (1, '1001'))", 1);
+        assertUpdate("INSERT INTO " + tableName + " VALUES (2, (2, '1002'))", 1);
+        assertUpdate("INSERT INTO " + tableName + " VALUES (3, (3, '1003'))", 1);
+
+        assertQuery("SELECT * FROM " + tableName + " WHERE r.a = 1", "VALUES (1, (1, '1001'))");
+        assertQuery("SELECT * FROM " + tableName + " WHERE r.b = '1003'", "VALUES (3, (3, '1003'))");
+        assertQuery("SELECT * FROM " + tableName + " WHERE r.a > 1 and r.b < '1003'", "VALUES (2, (2, '1002'))");
+        dropTable(session, tableName);
+    }
+
     protected String getLocation(String schema, String table)
     {
         return null;
--- a/presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergLogicalPlanner.java
+++ b/presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergLogicalPlanner.java
@@ -66,6 +66,7 @@ import static com.facebook.presto.common.predicate.ValueSet.ofRanges;
 import static com.facebook.presto.common.type.BigintType.BIGINT;
 import static com.facebook.presto.common.type.BooleanType.BOOLEAN;
 import static com.facebook.presto.common.type.DoubleType.DOUBLE;
+import static com.facebook.presto.common.type.IntegerType.INTEGER;
 import static com.facebook.presto.common.type.VarcharType.VARCHAR;
 import static com.facebook.presto.expressions.LogicalRowExpressions.TRUE_CONSTANT;
 import static com.facebook.presto.iceberg.IcebergAbstractMetadata.isEntireColumn;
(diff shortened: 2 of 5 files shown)


presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedSmokeTestBase.java
```diff
@@ -891,6 +891,23 @@ public void testPartitionedByDoubleWithNaNInf()
         dropTable(session, tableName);
     }
 
+    @Test
+    public void testFilterBySubfieldOfRowType()
```
ZacBlanco, line 895: Do you think this test case would fit better in `TestIcebergLogicalPlanner`?

hantangwangd: I notice test cases in ``TestIcebergLogicalPlanner `` are all about the verification of optimized plans, but my test case ``testFilterBySubfieldOfRowType()`` is about the verification of query result. I think it's not so suitable to be placed in ``TestIcebergLogicalPlanner ``.

Or do you means we should also verify the optimized plan of the statement?

imjalpreet: I think we should also validate the plan in `TestIcebergLogicalPlanner`. 

Maybe the below cases:
* Filter contains only identity partition columns
* Filter contains a partition column and a regular column
* Filter contains a partition column and a subfield of a column
* Filter contains only a subfield of a column

imjalpreet: This test(`testFilterBySubfieldOfRowType`) can remain in this class

imjalpreet (commented):
We can add the plan validation tests I mentioned here: https://github.com/prestodb/presto/pull/21616/files#r1439964052, apart from that LGTM.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedSmokeTestBase.java
hantangwangd: Done!

hantangwangd:
> We can add the plan validation tests I mentioned here: https://github.com/prestodb/presto/pull/21616/files#r1439964052, apart from that LGTM.

I have added the test cases for plan validation, please take a look, thanks! @imjalpreet @ZacBlanco

ZacBlanco (approved):
LGTM, thanks for adding the new tests!

presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergLogicalPlanner.java
```diff
@@ -112,6 +113,54 @@ protected QueryRunner createQueryRunner()
         return createIcebergQueryRunner(ImmutableMap.of("experimental.pushdown-subfields-enabled", "true"), ImmutableMap.of());
     }
 
+    @Test
+    public void testFiltersWithPushdownDisable()
+    {
+        Session sessionWithoutFilterPushdown = pushdownFilterDisabled();
```
imjalpreet, line 119: Since the filter pushdown session property is disabled by default, I think this is not required.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergLogicalPlanner.java
```diff
@@ -112,6 +113,54 @@ protected QueryRunner createQueryRunner()
...
+        assertUpdate("CREATE TABLE test_filters_with_pushdown_disable(id int, name varchar, r row(a int, b varchar)) with (partitioning = ARRAY['id'])");
+
+        // Only identity partition column predicates, would be enforced totally by tableScan
+        assertPlan(sessionWithoutFilterPushdown, "SELECT name, r FROM test_filters_with_pushdown_disable WHERE id = 10",
```
imjalpreet, line 123: We can use `com.facebook.presto.tests.AbstractTestQueryFramework#assertPlan(java.lang.String, com.facebook.presto.sql.planner.assertions.PlanMatchPattern)` which uses default session.

tdcmeehan (approved):

imjalpreet (commented):
LGTM, just some minor nits.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergLogicalPlanner.java
hantangwangd: Fixed, please take a look!

tdcmeehan (approved):

imjalpreet:
@hantangwangd It seems there were two unrelated files `presto-spi/src/main/java/com/facebook/presto/spi/ColumnHandle.java` and `presto-main/src/test/java/com/facebook/presto/sql/planner/assertions/ColumnReference.java` that have been committed along with this PR which weren't present earlier when we reviewed. Can you please confirm?

ZacBlanco:
@imjalpreet These files were both present when I reviewed

imjalpreet:
Oh, I had the original PR checkout in my local in which only two iceberg files were present before plan validation tests were added. I might have missed seeing the new files in my last review.

But it doesn't look like those changes are related to this PR, let me know if they are. As far as I understand we just needed the fix in IcebergPlanOptimizer. Please let me know in case I misunderstood something.

@hantangwangd @ZacBlanco

hantangwangd:
@imjalpreet The modify in `ColumnHandle` and `ColumnReference` are related to the newly added test cases in `TestIcebergLogicalPlanner`, in order to support the column matching for RowType `r row(a int, b varchar)` when having subfield column filter conditions like `r.a=10`.

imjalpreet:
@hantangwangd Oh, thanks for the details and confirmation. I missed that, will have a look.

PR #21620, opened 2023-12-31
[native] Advance velox and update memory code

majetideepak:
Fix JsonSignatureParserTest.complexTypes test.


xiaoxmeng (approved):
@majetideepak thanks for the fix!

PR #21626, opened 2024-01-04
Fix writtenIntermediateBytes metric to capture correct intermediate w…

jaystarshot:
The writtenIntermediateBytes metric was noop because plan.isOutputTableWriterFragment() will always be true and wrong metrics were updated when there are temporary table writes currently since there was a bug in plan Fragmenter where it would pass temporary tables' writer operator as output table writes hence populating the wrong metric.
code [link](https://github.com/prestodb/presto/blob/4214d3826d2e49c580a00974a704685045f79bf7/presto-main/src/main/java/com/facebook/presto/execution/QueryStats.java#L389).

Fixed by adding the information (whether temporary) to the tablewriterNode and filtering out those

## Impact
Fixed writtenIntermediateBytes metric. Users of CTE and Exchange materialization should now be able to see the correct populated metric

## Test Plan
Production tested

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Fixed writtenIntermediateBytes metric by ensuring its correct population through temporary table writers.

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/sql/planner/PlanFragmenter.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/planner/PlanFragmenter.java
@@ -37,7 +37,7 @@ import java.util.Set;
 import static com.facebook.presto.SystemSessionProperties.isForceSingleNodeOutput;
 import static com.facebook.presto.sql.planner.PlanFragmenterUtils.ROOT_FRAGMENT_ID;
 import static com.facebook.presto.sql.planner.PlanFragmenterUtils.finalizeSubPlan;
-import static com.facebook.presto.sql.planner.PlanFragmenterUtils.getTableWriterNodeIds;
+import static com.facebook.presto.sql.planner.PlanFragmenterUtils.getOutputTableWriterNodeIds;
 import static com.facebook.presto.sql.planner.SystemPartitioningHandle.SINGLE_DISTRIBUTION;
 import static java.util.Objects.requireNonNull;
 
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

File: presto-main/src/main/java/com/facebook/presto/sql/planner/PlanFragmenter.java [hunk 1]
```diff
@@ -81,7 +81,7 @@ public SubPlan createSubPlans(Session session, Plan plan, boolean forceSingleNod
                 sqlParser,
                 idAllocator,
                 variableAllocator,
-                getTableWriterNodeIds(plan.getRoot()));
```
jaystarshot, line 84: Main fix

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/CteUtils.java
```diff
@@ -25,6 +25,7 @@
 
 public class CteUtils
 {
+    public static final String TEMPORARY_TABLE_SCHEMA = "temporary_table_schema";
```
feilong-liu, line 28: where is this used?

jaystarshot: Removed thanks

feilong-liu (approved):
LGTM. I also briefly checked the code, looks like that the "outputTableWriterFragment" variable is only used for accounting purpose.

Can you maybe add a comment for the "outputTableWriterFragment" member in "PlanFragment" class stating that this is true only for final output fragment, but not for temporary table writer?

jaystarshot:
cc: @tdcmeehan

tdcmeehan (approved):

steveburnett:
Nit (commenting because it's not enough for a review): in Description, suggest deleting the lines in the release note entry after the line `Fixed writtenIntermediateBytes metric...`

steveburnett:
> Nit (commenting because it's not enough for a review): in Description, suggest deleting the lines in the release note entry after the line `Fixed writtenIntermediateBytes metric...`

Thanks!

PR #21640, opened 2024-01-05
[Iceberg]Fix filtering by unmatched value with Iceberg filter pushdown enabled

hantangwangd:
## Description

When filter pushdown is enabled, trying to query from an empty Iceberg table  or query with filter which do not match any partition would lead in an error. This PR fix the problem.

## Test Plan

 - Newly added test case TestIcebergLogicalPlanner.testFilterByUnmatchedValueWithFilterPushdown()

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
@@ -239,6 +239,17 @@ public abstract class IcebergAbstractMetadata
             return new ConnectorTableLayout(handle);
         }
 
+        if (!icebergTableLayoutHandle.getPartitions().isPresent()) {
+            return new ConnectorTableLayout(
+                    icebergTableLayoutHandle,
+                    Optional.empty(),
+                    TupleDomain.none(),
+                    Optional.empty(),
+                    Optional.empty(),
+                    Optional.empty(),
+                    ImmutableList.of(),
+                    Optional.empty());
+        }
         List<ColumnHandle> partitionColumns = ImmutableList.copyOf(icebergTableLayoutHandle.getPartitionColumns());
         List<HivePartition> partitions = icebergTableLayoutHandle.getPartitions().get();
 
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergUtil.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergUtil.java
@@ -550,9 +550,16 @@ public final class IcebergUtil
             List<IcebergColumnHandle> partitionColumns)
     {
         IcebergTableName name = ((IcebergTableHandle) tableHandle).getTableName();
+
+        // Empty iceberg table would cause `snapshotId` not present
+        Optional<Long> snapshotId = resolveSnapshotIdByName(icebergTable, name);
+        if (!snapshotId.isPresent()) {
+            return ImmutableList.of();
+        }
+
         TableScan tableScan = icebergTable.newScan()
                 .filter(toIcebergExpression(constraint.getSummary().simplify().transform(IcebergColumnHandle.class::cast)))
-                .useSnapshot(resolveSnapshotIdByName(icebergTable, name).get());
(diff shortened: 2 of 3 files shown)


presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
```diff
@@ -239,6 +239,17 @@ public ConnectorTableLayout getTableLayout(ConnectorSession session, ConnectorTa
             return new ConnectorTableLayout(handle);
         }
 
+        if (!icebergTableLayoutHandle.getPartitions().isPresent()) {
```
imjalpreet, line 242: @hantangwangd Can you please let us know the error due to which this was needed? We probably should not return from here since if filter pushdown is enabled we still need to compute subfieldPredicate and in turn combinedRemainingPredicate.

Also, just wanted to add, that Filter Pushdown in Iceberg connector is currently only supported with Prestissimo/Velox and not with Java Workers in case you are trying querying using Java Workers.

imjalpreet (commented):
Thanks @hantangwangd empty table check looks good. I wanted some clarity around `IcebergAbstractMetadata` change.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
hantangwangd: If `icebergTableLayoutHandle.getPartitions()` is not present, the following code:
```
        List<HivePartition> partitions = icebergTableLayoutHandle.getPartitions().get();
```
would throw an error. This situation may occur when we trying to get partitions using a filter which matches none in `IcebergUtil.getPartitions(...)`. And In such situations,  seems there is no need to compute subfieldPredicate etc.

And as I understand, connector optimizer logic is running in coordinator, this problem appears in optimization stage, so I think maybe it need to be fix. If anything I misunderstand, please let me know.

imjalpreet: > And as I understand, connector optimizer logic is running in coordinator, this problem appears in optimization stage, so I think maybe it need to be fix.

Yes, that's right, if there's a fix required here it needs to be added. I just wanted to inform to clarify.

Should we pass on the remainingPredicate from icebergTableLayoutHandle? I see its being used when MetadataQueryOptimizer comes into the picture. I haven't tried that yet so not sure if sending empty here would lead to any issues.

imjalpreet: On a quick glance, it doesn't look like it should be an issue. So, the change looks good.

imjalpreet: Another alternate solution might be to set partitions to an empty list rather than null here: https://github.com/prestodb/presto/blob/0d09809ed42534b0ac5a801d7dd25bbef3a3092a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergFilterPushdown.java#L203

@hantangwangd Then we would not need the check over here. Let me know which one you think is a better solution.

hantangwangd: @imjalpreet Thanks for the suggestion. Do not get any matching partition means the filter would cause an empty result, so I think it's better to stop immediately and return an ConnectorTableLayout with predicate `TupleDomain.none()`, which would cause an empty ValuesNode. If anything I missed, please let me know.

imjalpreet: Yeah sure looks good, both ways would solve the issue, I am good with any.

imjalpreet (approved):

ZacBlanco (approved):
LGTM!

tdcmeehan (approved):

PR #21643, opened 2024-01-05
[native]Set cpu driver slicing to 1 second in Prestissimo

xiaoxmeng:
We will remove this once Velox has default 1 second settings


gggrace14 (commented):
This change itself looks good to me that turns it on for Prestissimo.

It looks to me perf impactful. Can you help do a before & after perf comparison with prod traffic? Usually we compare total_split_cpu_time and query_execution_time.

Also just to get the knowledge, we once have this https://github.com/facebookincubator/velox/pull/4620. How are the two related and different?

gggrace14 (approved):

Issue #16268, opened 2021-06-15
Support Java 16


lucagiac81:
OpenJDK 16 introduces several new features, such as a Vector API, and it’d be beneficial to have those features available in Presto.
Currently, building/launching Presto with OpenJDK 16 causes some errors. Observed errors and proposed solutions are documented below. The errors were observed when building with -DskipTest (tests still need to be examined).
This effort would also help support JDK 17 (next LTS release) in the long run.

### spotbugs-maven-plugin

`[ERROR] Failed to execute goal com.github.spotbugs:spotbugs-maven-plugin:3.1.10:spotbugs (spotbugs) on project presto-root: Execution spotbugs of goal com.github.spotbugs:spotbugs-maven-plugin:3.1.10:spotbugs failed: Unable to load the mojo 'spotbugs' in the plugin 'com.github.spotbugs:spotbugs-maven-plugin:3.1.10'. A required class is missing: Could not initialize class org.codehaus.groovy.vmplugin.v7.Java7`

- com.github.spotbugs:spotbugs-maven-plugin:3.1.10 is specified in pom.xml of com.facebook.airlift:airbase:102 (parent POM of presto-root).
- Upgrading to spotbugs-maven-plugin:4.2.2 resolves the error.
- The latest version of airbase (110) does use spotbugs-maven-plugin:4.2.2.
  - So, updating airbase from 102 to 110 in presto-root would be sufficient.
  - However, there is no artifact for airbase:110 in Maven repository (102 is the latest published). An artifact for airbase:110 should be generated.
- For testing purposes, I specified spotbugs-maven-plugin:4.2.2 directly in presto-root pom.xml.

### provisio-maven-plugin

Several errors are caused by the following change in OpenJDK 16: [Strongly Encapsulate JDK Internals by Default](https://openjdk.java.net/jeps/396)
Example of one such error:

```
[ERROR] Failed to execute goal io.takari.maven.plugins:provisio-maven-plugin:0.1.40:provision (default-provision) on project presto-tpch: Execution default-provision of goal io.takari.maven.plugins:provisio-maven-plugin:0.1.40:provision failed: An API incompatibility was encountered while executing io.takari.maven.plugins:provisio-maven-plugin:0.1.40:provision: java.lang.ExceptionInInitializerError: null

[...]

PR #21670, opened 2024-01-10
[WIP] Optimize Hive Connector ORC Readers and Filter using Java Vector API

AbhijitKulkarni1:
## Description

c5edcc336300e21b390f475b3104d0629b754d42:
Enable Presto to be compatible with JDK19.0.2 version. Related to #16268 

93a8d8160e76b5ab67689d7e817d45fa7f0ced52:
Optimize Hive connector to harness data parallelism in ORC Readers and Filters using Java Vector API

## Motivation and Context

This PR enables Presto to be used with latest features available in JDK19.0.2 version like Java Vector API and Foreign API updates. Presto ORC module is optimized with vectorized Reader to harness data parallelism using Java Vector API. Hive Connector is updated to provide hive property "orc-use-vector-filter" to use the optimized Presto ORC to read ORC format data in vectorized way.
To start with we have improved the Double data type ORC readers and other data types will follow in future.

## Impact

c5edcc336300e21b390f475b3104d0629b754d42:
JDK19 specific maven profiles are added to update the versions of dependencies/plugins as required for Presto to be compatible with JDK19.0.2 version. These updates do not impact the default compilation with JDK8 or 11 which is officially supported.

93a8d8160e76b5ab67689d7e817d45fa7f0ced52:
No functional impact. Improved Hive Connector for ORC Readers and Filters

## Test Plan

c5edcc336300e21b390f475b3104d0629b754d42:
Test all smoke test suites

93a8d8160e76b5ab67689d7e817d45fa7f0ced52:
Added unit tests to verify the vectorization updates and performed test coverage.

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==
c5edcc336300e21b390f475b3104d0629b754d42:
General Changes
Compilation and Building:
Set the following in MAVEN_OPTS environmental variable before building the presto with JDK19.
export MAVEN_OPTS="--add-opens java.base/java.util=ALL-UNNAMED --add-opens java.base/java.lang.reflect=ALL-UNNAMED --add-opens java.base/java.text=ALL-UNNAMED --add-opens java.base/java.awt.font=ALL-UNNAMED --add-opens java.desktop/java.awt.font=ALL-UNNAMED --add-opens=java.base/java.nio=ALL-UNNAMED"
./mvnw clean install -DskipTests

Launch:
Set the following jvm options in addition to default options in etc/jvm.config file during Presto launch to run with JRE 19.

--add-opens=java.base/java.lang=ALL-UNNAMED
--add-opens=java.base/java.lang.reflect=ALL-UNNAMED
--add-opens=java.base/java.nio=ALL-UNNAMED
-Djdk.reflect.useDirectMethodHandle=false

93a8d8160e76b5ab67689d7e817d45fa7f0ced52:
General Changes
Note the following points about this optimization.
1. Updates are compatible with JDK 19.0.2 version. So use jdk 19.0.2 for compilation and execution of Presto binary.
2. The JDK 19.0.2 compilation for vector updates in presto-common, presto-orc modules are handled using maven profiles. The updates do not impact the lower version JDK compilation (JDK 8 and 11).
3. Note the following points for compilation and launch of Presto using JDK19.0.2
    i. Compilation: 
export MAVEN_OPTS="--add-opens java.base/java.util=ALL-UNNAMED --add-opens java.base/java.lang.reflect=ALL-UNNAMED --add-opens java.base/java.text=ALL-UNNAMED --add-opens java.base/java.awt.font=ALL-UNNAMED --add-opens java.desktop/java.awt.font=ALL-UNNAMED --add-opens=java.base/java.nio=ALL-UNNAMED --add-modules=jdk.incubator.vector --enable-preview"
./mvnw clean install -DskipTests

   ii. Runtime in Presto "etc\jvm.config" file set the following jvm options
--add-opens=java.base/java.lang=ALL-UNNAMED
--add-opens=java.base/java.lang.reflect=ALL-UNNAMED
--add-opens=java.base/java.nio=ALL-UNNAMED
-Djdk.reflect.useDirectMethodHandle=false
--add-modules=jdk.incubator.vector
--enable-preview

Hive Connector Changes
Addition of hive property  "orc-use-vector-filter" to enable/disable the use of vector ORC readers in Hive connector. If set to "true" in hive.properties file (hive.orc-use-vector-filter="true"), new vectorized readers and filters are used for compilation/execution. Default being "false" for current non vectorized flow.

```
== NO RELEASE NOTE ==
```

### CODE REVIEW

**Comment by tdcmeehan** [comment]
@AbhijitKulkarni1 this would be a very nice and exciting change.  I am wondering if you need any help?

**Comment by steveburnett** [comment]
Suggest a change to the current release notes entry: perhaps this draft describes the effect of this PR? 

```
== RELEASE NOTES ==
General Changes
* Improve Presto compatibility with JDK 19.0.2. 

Hive Connector Changes
* Add configuration property  `orc-use-vector-filter` to enable or disable the use of vector ORC readers.
```

**Comment by AbhijitKulkarni1** [comment]
@tdcmeehan Thanks for reaching out. We are in process of internal data review and would be able to submit PR for review shortly. Will surely connect with you in case I need any help.

**Comment by steveburnett** [comment]
Two items:

* Please add documentation for ``orc-use-vector-filter`` - probably in [Hive Configuration Properties](https://prestodb.io/docs/current/connector/hive.html#hive-configuration-properties). 

* Suggest a change to the current release notes entry: perhaps this draft describes the effect of this PR? 

```
== RELEASE NOTES ==
General Changes
* Improve Presto compatibility with JDK 19.0.2 :pr: `21760`

Hive Connector Changes
* Add configuration property  ``orc-use-vector-filter`` to enable or disable the use of vector ORC readers :pr: `21760`
```

**Comment by AbhijitKulkarni1** [comment]
@steveburnett Can you please let me know about the second point? Do you mean to add one liner release notes and move the details in other section?
- Suggest a change to the current release notes entry: perhaps this draft describes the effect of this PR?

**Comment by steveburnett** [comment]
Good question, @AbhijitKulkarni1! 

I suggest that each release note entry should be a summary of the work done in the PR. The summary version allows a reader of the new Release Notes to quickly identify if an entry is something they need more detail on. If not, they can move on to the next entry. You could examine the recent release notes, for example the [0.287 Release Notes](https://prestodb.io/docs/current/release/release-0.287.html), for examples. 

Because each release note entry ends with a link to the PR it describes - for example, if the reader wants to know what ``Improve Presto compatibility with JDK 19.0.2`` means - they can select the link to the PR and find that explanation in the Description part of your PR. 

I suggest you move much of the content you have in your release note draft into the Description of your PR. 

Each individual item does not have to be only a single line: I attempted to summarize each of yours at that high level, but you should revise my suggested draft and include detail you feel must be present in the release notes. 

For PRs that include documentation changes, a link can be added to the release note entry that links directly to the new documentation. 
See [Formatting](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines#formatting) in the Release Notes Guidelines for how to code such a link. 
For a specific example, see the [0.287 Release Notes](https://prestodb.io/docs/current/release/release-0.287.html) and the release note entry:

"* Add configuration property legacy_json_cast whose default value is true. See [Legacy Compatible Properties](https://prestodb.io/docs/current/admin/properties.html#legacy-compatible-properties). [#21869](https://github.com/prestodb/presto/pull/21869)" 

I hope this helps to answer your question, and please ask anything else you would like to know.

**Comment by AbhijitKulkarni1** [comment]
Thanks @steveburnett for the guidance. I will add the property to Hive Configurations property file hive.rst. For release notes can you please clarify if they are picked up from PR automatically during release preparation or I need to add it somewhere because I didn't find the file release-0.288.rst, Kindly let me know.

**Comment by steveburnett** [comment]
Of course! The release note entries are picked up from the PRs automatically during release preparation - if you have a correctly formatted block in your PR for your release note entries, that's all you need to do. 
The automatically collected release note entries are reviewed and revised to create the .rst file. 
My suggestions hope to reduce the work of the reviewers preparing the next release notes, which can help make the Presto release process easier and maybe even faster as well. 

If you're interested in how the release notes are created for a Presto release, you're welcome to take a look at #22647.

File: presto-orc/src/main/java/com/facebook/presto/orc/OrcSelectiveRecordReader.java [hunk 1]
```diff
@@ -188,7 +188,8 @@ public OrcSelectiveRecordReader(
             StripeMetadataSource stripeMetadataSource,
             boolean cacheable,
             RuntimeStats runtimeStats,
-            Optional<OrcFileIntrospector> fileIntrospector)
+            Optional<OrcFileIntrospector> fileIntrospector,
+            Map<String, Boolean> orcReaderUserOptions)
```
sdruzkin, line 192: Please no random maps with user options, everything related to options should go to OrcRecordReaderOptions

ZacBlanco:
First off, awesome work. I'm sure this was difficult to get working properly. Upgrading Java versions can be tricky, especially for a large legacy system like Presto that has barely migrated off of 8. One concern I have is with the choice of Java version. Java 19 is not an LTS release and support for this JVM version has already ended. It will not be receiving any further updates: https://javaalmanac.io/jdk/

It would be really great if this could target an LTS release. 21 is the latest LTS release and EOL is in 2028. I did a quick search and it seems there were some API changes between 19 and 21 in the Vector API. I'm not sure what was behind the decision to target Java 19, but it would be good to know the motivation. Not sure what others think about the Java version, but I think using LTS is desirable.

presto-docs/src/main/sphinx/connector/hive.rst
```diff
@@ -195,6 +195,8 @@ Property Name                                      Description
                                                    S3SelectPushdown.
 
 ``hive.metastore.load-balancing-enabled``          Enable load balancing between multiple Metastore instances
+
+``hive.orc-use-vector-filter``                     Enable use of vector ORC readers in compilation and execution ``false``
```
steveburnett, line 199:
```suggestion
`hive.orc-use-vector-filter``                     Enable use of vector ORC readers in compilation and          ``false``
                                                   execution
```
So tables are fussy, and leaving "execution" on line 199 causes it to break into the Default values column. When I edited hive.rst in my local build according to this screenshot, the page compiles successfully and your added property is present. 
(I think my suggestion here is correct, I'm including a screenshot of my Visual Studio Code for the suggested edit to show you exactly what worked to fix this in my local build.) 

Screenshot 2024-06-12 at 3 01 23 PM

steveburnett (changes requested):
One formatting error that I fixed locally so I could review the content. The content is good, just fix the table entry and I'm happy to approve.

AbhijitKulkarni1:
> First off, awesome work. I'm sure this was difficult to get working properly. Upgrading Java versions can be tricky, especially for a large legacy system like Presto that has barely migrated off of 8. One concern I have though is with the choice of Java version. Java 19 is not an LTS release and support for this JVM version has already ended. It will not be receiving any further updates: https://javaalmanac.io/jdk/
> 
> It would be really great if this could target an LTS release. 21 is the latest LTS release and EOL is in 2028. I did a quick search and it seems there were some API changes between 19 and 21 in the Vector API. I'm not sure what was behind the decision to target Java 19, but it would be good to know the motivation. Not sure what others think about the Java version, but I think using LTS is desirable.

Thanks for the compliments. Ensuring Presto worked with the latest Java was indeed a tough task. We started with compatibility explorations and considered Java 17 LTS, 18 for vectorized filtering with shuffle and rearrange Vector APIs to start with but chose Java 19 for "compress" Vector API (https://openjdk.org/jeps/426), which does shuffle and rearrange in one API call and performs better. I'm of the opinion that this solution should be compatible with Java 21 LTS; however, I must confirm its compatibility before moving forward with the submission.

steveburnett (approved):
LGTM! (doc)

Pull updated branch, new doc build, looks good!

presto-orc/src/main/java/com/facebook/presto/orc/OrcSelectiveRecordReader.java
AbhijitKulkarni1: Review comment fixed

presto-docs/src/main/sphinx/connector/hive.rst
AbhijitKulkarni1: Issue fixed
Fix

AbhijitKulkarni1:
> > First off, awesome work. I'm sure this was difficult to get working properly. Upgrading Java versions can be tricky, especially for a large legacy system like Presto that has barely migrated off of 8. One concern I have though is with the choice of Java version. Java 19 is not an LTS release and support for this JVM version has already ended. It will not be receiving any further updates: https://javaalmanac.io/jdk/
> > It would be really great if this could target an LTS release. 21 is the latest LTS release and EOL is in 2028. I did a quick search and it seems there were some API changes between 19 and 21 in the Vector API. I'm not sure what was behind the decision to target Java 19, but it would be good to know the motivation. Not sure what others think about the Java version, but I think using LTS is desirable.
> 
> Thanks for the compliments @ZacBlanco . Ensuring Presto worked with the latest Java was indeed a tough task. We started with compatibility explorations and considered Java 17 LTS, 18 for vectorized filtering with shuffle and rearrange Vector APIs to start with but chose Java 19 for "compress" Vector API (https://openjdk.org/jeps/426), which does shuffle and rearrange in one API call and performs better. I'm of the opinion that this solution should be compatible with Java 21 LTS; however, I must confirm its compatibility before moving forward with the submission.

Updated PR to support JDK 21.0.3 LTS version.

presto-common/src/main/java/com/facebook/presto/common/predicate/TupleDomainFilter.java
```diff
@@ -744,8 +744,8 @@ public boolean getUpperExclusive()
     class DoubleRange
             extends AbstractRange
     {
-        private final double lower;
-        private final double upper;
+        protected final double lower;
```
rschlussel, line 747: probably better to keep these fields private and access via getters.

presto-docs/src/main/sphinx/connector/hive.rst
```diff
@@ -198,6 +198,8 @@ Property Name                                      Description
 
 ``hive.skip-empty-files``                          Enable skipping empty files. Otherwise, it will produce an   ``false``
                                                    error iterating through empty files.
+
+``hive.orc-use-vector-filter``                     Enable use of vector ORC readers in compilation and execution ``false``
```
rschlussel: add a note that this requires jdk 19?

presto-hive/src/main/java/com/facebook/presto/hive/HiveClientConfig.java
```diff
@@ -1861,4 +1862,17 @@ public HiveClientConfig setLegacyTimestampBucketing(boolean legacyTimestampBucke
         this.legacyTimestampBucketing = legacyTimestampBucketing;
         return this;
     }
+
+    @Config("hive.orc-use-vector-filter")
+    @ConfigDescription("Experimental: enable vector path for orc filters")
```
rschlussel, line 1868: note that this requires jdk 19. Also, would probably remove "Experimental".  Notes like that tend to just stick around forever. Would be great to add validation that the jdk meets the minimum version and fail with invalid_configuration_property otherwise.

presto-hive/src/main/java/com/facebook/presto/hive/HiveSessionProperties.java
```diff
@@ -215,6 +216,11 @@ public HiveSessionProperties(HiveClientConfig hiveClientConfig, OrcFileWriterCon
                         "Experimental: ORC: Compression level, works only for ZSTD and ZLIB compression kinds",
                         orcFileWriterConfig.getCompressionLevel(),
                         false),
+                booleanProperty(
+                        ORC_USE_VECTOR_FILTER,
+                        "Experimental: enable vector path for orc filters",
```
rschlussel, line 221: same comment about noting the jdk requirement and probably removing "experimental". Would be great to add validation that the jdk meets the minimum version if you set this, and fail with INVALID_SESSION_PROPERTY otherwise.

presto-orc/src/main/java/com/facebook/presto/orc/reader/SelectiveStreamReaders.java
```diff
@@ -138,6 +142,44 @@ public static SelectiveStreamReader createStreamReader(
...
+                constructorArguments[1] = Optional.class;
+                constructorArguments[2] = boolean.class;
+                constructorArguments[3] = OrcLocalMemoryContext.class;
+                Constructor constructor = doubleSelectiveStreamVectorReaderClass.getDeclaredConstructor(constructorArguments);
+                return (SelectiveStreamReader) constructor.newInstance(streamDescriptor,
+                        getOptionalOnlyFilter(type, filters),
+                        outputType.isPresent(),
+                        systemMemoryContext.newOrcLocalMemoryContext(SelectiveStreamReaders.class.getSimpleName()));
+            }
+            catch (ClassNotFoundException | NoSuchMethodException | InstantiationException | IllegalAccessException | InvocationTargetException e) {
```
rschlussel, line 174: we should fail loudly here.

presto-orc/src/test/java/com/facebook/presto/orc/reader/vector/vectorReadAndFilter/TestDoubleVectorReadAndFilter.java
```diff
@@ -0,0 +1,438 @@
...
+            double[] doubleArr = new double[20];
+            for (int i = 1; i < 21; i++) {
+                doubleArr[i - 1] = (double) ((i * 10) % 80);
+            }
+            ByteBuffer bb = ByteBuffer.allocate(doubleArr.length * 8);
+            bb.order(ByteOrder.LITTLE_ENDIAN);
+            for (double d : doubleArr) {
+                bb.putDouble(d);
+            }
+            Mockito.when(inputStream.readDoubleDataInBytes(anyInt())).thenReturn(bb.array());
```
rschlussel, line 67: we don't use mocks in presto. Please stub out any classes that you don't want to create fully.

presto-orc/src/main/java/com/facebook/presto/orc/reader/DoubleSelectiveStreamReader.java
```diff
@@ -56,29 +56,29 @@ public class DoubleSelectiveStreamReader
 
     private final StreamDescriptor streamDescriptor;
     @Nullable
-    private final TupleDomainFilter filter;
-    private final boolean nonDeterministicFilter;
-    private final boolean nullsAllowed;
-    private final boolean outputRequired;
+    protected final TupleDomainFilter filter;
```
rschlussel, line 59: I don't love the idea of exposing all of these internal details to the whole package. I wonder if we can have a "FilterReader function" that gets passed into this class instead and that gets called in readWithFilter, and when the vector reader ins enable it's the VectorFilterReader and a DefaultFilterReader. then we could pass any required fields as function arguments?

presto-common/src/main/java/com/facebook/presto/common/predicate/vector/DoubleRangeVector.java
```diff
@@ -0,0 +1,73 @@
...
+import jdk.incubator.vector.VectorMask;
+
+import static jdk.incubator.vector.VectorOperators.GE;
+import static jdk.incubator.vector.VectorOperators.GT;
+import static jdk.incubator.vector.VectorOperators.LE;
+import static jdk.incubator.vector.VectorOperators.LT;
+import static jdk.incubator.vector.VectorOperators.NE;
+
+public class DoubleRangeVector
+        extends TupleDomainFilter.DoubleRange
```
rschlussel, line 27: what do we get out of extending DoubleRange? Seems like we don't use any of the DoubleRange functionality.  We should add those fields directly to DoubleRangeVector instead and make it independent of DoubleRange.

presto-common/src/main/java/com/facebook/presto/common/predicate/vector/DoubleRangeVector.java
```diff
@@ -0,0 +1,73 @@
...
+        implements TupleDomainFilterVector
+{
+    protected DoubleRangeVector(double lower, boolean lowerUnbounded, boolean lowerExclusive, double upper, boolean upperUnbounded, boolean upperExclusive, boolean nullAllowed)
```
rschlussel, line 30: Can this be private?

presto-common/pom.xml
```diff
@@ -115,4 +115,112 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+
+    <profiles>
+        <profile>
+            <id>java21PrestoCommon</id>
```
rschlussel, line 121: just curious why the build profiles are named differently for every module vs. all having a profile called java21 and javaLessThan21?

presto-orc/pom.xml
```diff
@@ -211,5 +211,139 @@
...
             </build>
         </profile>
+        <profile>
+            <id>java21PrestoORC</id>
+            <activation>
+                <jdk>21</jdk>
+            </activation>
+            <dependencies>
+                <dependency>
+                    <groupId>org.mockito</groupId>
```
rschlussel, line 221: We don't use mocks. Please create stubbed out classes instead.

rschlussel (commented):
Thanks for this change!

Can you please clean up the commits and break it into 2 commits
1. changes required for java > 19 support
2. Orc vector changes

I do feel a bit worried about how this change will affect development, and would love to hear thoughts about it. Will we need to set our java version in intellij to java 19? will it make it hard for people to avoid using features after java8 until we upgrade our required java version for the whole project?

AbhijitKulkarni1:
> Thanks for this change!
> 
> Can you please clean up the commits and break it into 2 commits
> 
> 1. changes required for java > 19 support
> 2. Orc vector changes
> 
> I do feel a bit worried about how this change will affect development, and would love to hear thoughts about it. Will we need to set our java version in intellij to java 19? will it make it hard for people to avoid using features after java8 until we upgrade our required java version for the whole project?

The updates are handled such a way that it does not affect the compilation and execution for lower java version. This is done with specific maven profiles for JDK 21 during compile time. Also for Hive connector execution, the Vector API code is executed only if the flag 'hive.orc-use-vector-filter' is set to true and we have the Vector API support from Java runtime. So these updates do support backward compatibility.

presto-orc/src/main/java/com/facebook/presto/orc/reader/SelectiveStreamReaders.java
AbhijitKulkarni1: Can you please help me know ? And BTW there is old flow available if the incubator classes are not available during runtime.

presto-common/src/main/java/com/facebook/presto/common/predicate/vector/DoubleRangeVector.java
AbhijitKulkarni1: DoubleRangeVector is an extention of the scalar equivalent and in the vector processing logic for handling the scalar tail part of execution we need the DoubleRange functionality to filter in scalar way.

wanglinsong:
Do we need to update other test workflows besides .github/workflows/hive-tests.yml?
How do we make sure other core functionalities work correctly with jdk21?

presto-orc/pom.xml
AbhijitKulkarni1: Could you kindly direct me to an example within Presto? That would be very helpful.

presto-orc/src/main/java/com/facebook/presto/orc/reader/SelectiveStreamReaders.java
rschlussel: I mean to say instead of falling back if the classes aren't available, it would be better to fail the query so people know they have misconfigured presto.

presto-orc/pom.xml
rschlussel: you can look at any classes named TestingXXX as an example. We create testing implementations of interfaces that follow the needs of a test. Can be simple or complex as needed.   Here are a few examples, and you can see how they are used in tests:  https://github.com/prestodb/presto/blob/master/presto-main/src/main/java/com/facebook/presto/testing/TestingMetadata.java, https://github.com/prestodb/presto/blob/master/presto-spi/src/test/java/com/facebook/presto/spi/ttl/TestingNodeTtlFetcher.java, https://github.com/prestodb/presto/blob/master/presto-spi/src/test/java/com/facebook/presto/spi/TestingColumnHandle.java, https://github.com/prestodb/presto/blob/master/presto-main/src/main/java/com/facebook/presto/testing/TestingAccessControlManager.java

PR #21674, opened 2024-01-11
[native]Add to handle empty data case in ShuffleReader::next

xiaoxmeng:
Add to handle empty data case in ShuffleReader::next and update UnsafeRowExchangeSource
to remove the use of hasNext(). The followup is to deprecate hasNext.


mbasmanova (approved):
Thanks. Please add [native] prefix to commit and PR titles.

tanjialiang (approved):

amitkdutta (approved):

PR #21690, opened 2024-01-12
 [native pos] Make max_spill_bytes configurable in presto-on-spark

shrinidhijoshi:
Make `max_spill_bytes` configurable in presto-on-spark and default to 500 GB

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-spark-base/src/main/java/com/facebook/presto/spark/execution/property/NativeExecutionVeloxConfig.java
+++ b/presto-spark-base/src/main/java/com/facebook/presto/spark/execution/property/NativeExecutionVeloxConfig.java
@@ -29,12 +29,16 @@ public class NativeExecutionVeloxConfig
     private static final String AGGREGATION_SPILL_ENABLED = "aggregation_spill_enabled";
     private static final String JOIN_SPILL_ENABLED = "join_spill_enabled";
     private static final String ORDER_BY_SPILL_ENABLED = "order_by_spill_enabled";
+    private static final String MAX_SPILL_BYTES = "max_spill_bytes";
 
     private boolean codegenEnabled;
     private boolean spillEnabled = true;
     private boolean aggregationSpillEnabled = true;
     private boolean joinSpillEnabled = true;
     private boolean orderBySpillEnabled = true;
+    // Velox default value is 100GB, as it is designed for Presto cluster
+    // use-case. But for presto-on-spark, 500GB is a reasonable default
+    private Long maxSpillBytes = 500L << 30;
 
     public Map<String, String> getAllProperties()
     {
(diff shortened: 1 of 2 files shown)


presto-spark-base/src/main/java/com/facebook/presto/spark/execution/property/NativeExecutionVeloxConfig.java
```diff
@@ -29,12 +29,20 @@ public class NativeExecutionVeloxConfig
...
     private boolean spillEnabled = true;
     private boolean aggregationSpillEnabled = true;
     private boolean joinSpillEnabled = true;
     private boolean orderBySpillEnabled = true;
+    // Velox default value is 100G, as it is designed for Presto cluster
+    // use-case. But for presto-on-spark, we do not have any limit
+    private Long maxSpillBytes = 0L;
+    // Same as above comments. We do not need to set limit on
+    // how many rows we can spill
+    private Long maxSpillRunRows = 0L;
```
xiaoxmeng, line 45: Let's keep this to avoid memory allocation error during the spill. We can tune this a bit later if create too many spill files. Thanks!

xiaoxmeng (approved):
@shrinidhijoshi LGTM. Thanks!

presto-spark-base/src/main/java/com/facebook/presto/spark/execution/property/NativeExecutionVeloxConfig.java
shrinidhijoshi: Sounds good. Updated.

xiaoxmeng (approved):

mbasmanova:
> For presto-on-spark we do not have those limits.

I'm not sure if that's a good idea. Network and storage are shared resources that need to be protected. We should not allow a single query to use too much of these.

shrinidhijoshi:
@mbasmanova From discussion with @xiaoxmeng we decided that instead of setting up an arbitrary number here, we can start with unlimited and observe system behavior and come back and tweak it, or set it accordingly in Meta internal config files, instead of OSS

mbasmanova:
@shrinidhijoshi I don't think "unlimited" is a safe defaults. It seems better to have a reasonable default and let individual deployments override this instead of waiting for a major SEV.

shrinidhijoshi:
That's fair point. The initial thought here was that primarily we would be using a disagg storage architecture, we aren't bound in any way how much we can spill as we are not sharing the spill capacity. Hence the 0 default.

But on second thought, your argument makes sense. So I will try to find a reasonable default for this. Do you have something in mind already ?

shrinidhijoshi:
We discussed offline and did some research on current spill sizes for presto-on-spark tasks. Based on that research 500GB seems to be a reasonable default. Update the PR to reflect the change.

cc @mbasmanova

mbasmanova (approved):
Thanks.

PR #21710, opened 2024-01-17
[native] Remove mmap arena exposure from Prestissimo

tanjialiang:
mmap arena is never used and will not be a good use case for Prestissimo. Removing the exposure of mmap arena configs to reduce confusion.
```
== NO RELEASE NOTE ==
```


xiaoxmeng (approved):
@tanjialiang LGTM. Thanks!

PR #21720, opened 2024-01-17
Add Heuristic CTE Materialization Strategy

### LINKED ISSUE #21637: Heuristic Based CTE Materialization
Add a new strategy which will auto materialize ctes if used > x times and have y joins, aggregates etc. Needs proper brainstorming

Extends https://github.com/prestodb/presto/issues/19744

jaystarshot:
1. This PR adds support Heuristic CTE materialization.
We have rolled this in one of our production cluster and confirmed that this is stable. 
Our current testing has shown **upto** 50% cpu gain and > 50% less hdfs read for affected queries.
Its heuristic so there may be cases where non-ideal plans are generated.

2. Changed cteName member variable to cteId for cte plan nodes.

3. Also fixed CteInformation to include cteId so that correct references are incremented which crucial for cte materialization.

In heuristic strategy, the relational planner adds cteReferences everywhere and the LogicalCteOptimizer decides which references to implement

Fixes #21637 

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Add Heuristic CTE Materialization strategy which auto materialized expensive ctes. This is configurable by setting ``cte_materialization_strategy`` to ``HEURISTIC`` or ``HEURISTIC_COMPLEX_QUERIES_ONLY``. (default ``NONE``)

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
+++ b/presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
@@ -25,6 +25,7 @@ import com.facebook.presto.memory.MemoryManagerConfig;
 import com.facebook.presto.memory.NodeMemoryConfig;
 import com.facebook.presto.server.security.SecurityConfig;
 import com.facebook.presto.spi.PrestoException;
+import com.facebook.presto.spi.eventlistener.CTEInformation;
 import com.facebook.presto.spi.session.PropertyMetadata;
 import com.facebook.presto.spiller.NodeSpillConfig;
 import com.facebook.presto.sql.analyzer.FeaturesConfig;
@@ -94,6 +95,7 @@ public final class SystemSessionProperties
     public static final String DISTRIBUTED_INDEX_JOIN = "distributed_index_join";
     public static final String HASH_PARTITION_COUNT = "hash_partition_count";
     public static final String CTE_HASH_PARTITION_COUNT = "cte_hash_partition_count";
+    public static final String CTE_HEURISTIC_REPLICATION_THRESHOLD = "cte_heuristic_replication_threshold";
 
     public static final String PARTITIONING_PROVIDER_CATALOG = "partitioning_provider_catalog";
 
@@ -1090,6 +1092,11 @@ public final class SystemSessionProperties
                         "Enable pushing of filters and projections inside common table expressions.",
                         featuresConfig.getCteFilterAndProjectionPushdownEnabled(),
                         false),
+                integerProperty(
+                        CTE_HEURISTIC_REPLICATION_THRESHOLD,
+                        "Used with CTE Materialization Strategy = Heuristic. CTES are only materialized if they are used greater than or equal to this number",
+                        featuresConfig.getCteHeuristicReplicationThreshold(),
+                        false),
                 new PropertyMetadata<>(
                         DEFAULT_JOIN_SELECTIVITY_COEFFICIENT,
                         "use a default join selectivity coefficient factor when column statistics are not available in a join node",
@@ -2025,6 +2032,11 @@ public final class SystemSessionProperties
         return session.getSystemProperty(CTE_HASH_PARTITION_COUNT, Integer.class);
     }
 
+    public static int getCteHeuristicReplicationThreshold(Session session)
+    {
+        return session.getSystemProperty(CTE_HEURISTIC_REPLICATION_THRESHOLD, Integer.class);
+    }
+
     public static String getPartitioningProviderCatalog(Session session)
     {
         return session.getSystemProperty(PARTITIONING_PROVIDER_CATALOG, String.class);
@@ -2033,6 +2045,15 @@ public final class SystemSessionProperties
     public static String getCtePartitioningProviderCatalog(Session session)
     {
         return session.getSystemProperty(CTE_PARTITIONING_PROVIDER_CATALOG, String.class);
+    }
+
+    public static boolean isCteMaterializationApplicable(Session session)
+    {
+        boolean isStrategyNone = getCteMaterializationStrategy(session).equals(NONE);
+        boolean hasMaterializedCTE = session.getCteInformationCollector().getCTEInformationList()
+                .stream()
+                .anyMatch(CTEInformation::isMaterialized);
+        return !isStrategyNone && hasMaterializedCTE;
     }
 
     public static ExchangeMaterializationStrategy getExchangeMaterializationStrategy(Session session)
--- a/presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java
@@ -93,9 +93,9 @@ public class FeaturesConfig
     private DataSize maxRevocableMemoryPerTask = new DataSize(500, MEGABYTE);
     private JoinReorderingStrategy joinReorderingStrategy = JoinReorderingStrategy.AUTOMATIC;
     private PartialMergePushdownStrategy partialMergePushdownStrategy = PartialMergePushdownStrategy.NONE;
-
     private CteMaterializationStrategy cteMaterializationStrategy = CteMaterializationStrategy.NONE;
     private boolean cteFilterAndProjectionPushdownEnabled;
+    private int cteHeuristicReplicationThreshold = 4;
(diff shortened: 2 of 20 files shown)

### CODE REVIEW

**Comment by jaystarshot** [comment]
@feilong-liu  @mlyublena Please review when you get time, thanks!

File: presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java [hunk 1]
```diff
@@ -357,7 +357,9 @@ public boolean isAdoptingMergedPreference()
     public enum CteMaterializationStrategy
     {
         ALL, // Materialize all CTES
-        NONE // Materialize no ctes
+        NONE, // Materialize no CTES
+        HEURISTIC,
```
mlyublena, line 361: can you add comments for the new values

presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java
```diff
@@ -357,7 +357,9 @@ public boolean isAdoptingMergedPreference()
...
+        HEURISTIC_STRICT
```
mlyublena, line 362: I wonder if there is a better name than strict: perhaps HEURISTIC_COMPLEX_QUERIES_ONLY? we already have "complex query" reference in another session property (RESTRICT_HISTORY_BASED_OPTIMIZATION_TO_COMPLEX_QUERY)

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/LogicalCteOptimizer.java
```diff
@@ -132,34 +136,65 @@ public boolean isPlanRewritten()
...
         @Override
         public PlanNode visitCteReference(CteReferenceNode node, RewriteContext<CteTransformerContext> context)
         {
-            context.get().addDependency(node.getCteName());
-            context.get().pushActiveCte(node.getCteName());
-            // So that dependent CTEs are processed properly
+            HashMap<String, CTEInformation> cteInformationMap = session.getCteInformationCollector().getCteInformationMap();
+            CTEInformation cteInfo = cteInformationMap.get(node.getCteId());
+            switch (getCteMaterializationStrategy(session)) {
+                case HEURISTIC_STRICT:
```
mlyublena, line 163: do we want to also disable materialization of VALUES nodes? so only proceed if the query is a proper join/aggregation on top of table scans

mlyublena:
LGTM modulo some nits

mlyublena (approved):

jaystarshot:
@mlyublena  Thanks for your review, I have 
1. updated the PR to include the values check
2. Added tests in case to assert no materialization in heuristic cases
3. Backported a small fix from our repo in case there is no materialization in heuristic cases (cte references were not removed)

jaystarshot:
Rebased

presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
```diff
@@ -1082,6 +1083,11 @@ public SystemSessionProperties(
                         "Enable pushing of filters and projections inside common table expressions.",
                         featuresConfig.getCteFilterAndProjectionPushdownEnabled(),
                         false),
+                integerProperty(
+                        CTE_HEURISTIC_REPLICATION_THRESHOLD,
+                        "Used with CTE Materialization Strategy = Heuristic. CTES are only materialized if they are used greater than or equal to this number",
```
mlyublena, line 1088: nit: perhaps change comment to "if they are used more than this number of times"

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/LogicalCteOptimizer.java
```diff
@@ -132,34 +139,68 @@ public boolean isPlanRewritten()
...
-                    actualSource,
-                    node.getCteName(),
-                    variableAllocator.newVariable("rows", BIGINT), node.getOutputVariables());
-            context.get().addProducer(node.getCteName(), cteProducerSource);
-            return new CteConsumerNode(node.getSourceLocation(), idAllocator.getNextId(), actualSource.getOutputVariables(), node.getCteName());
+            HashMap<String, CTEInformation> cteInformationMap = session.getCteInformationCollector().getCteInformationMap();
+            CTEInformation cteInfo = cteInformationMap.get(node.getCteId());
+            switch (getCteMaterializationStrategy(session)) {
+                case HEURISTIC_COMPLEX_QUERIES_ONLY:
+                    if (!PlanNodeSearcher.searchFrom(node).where(planNode ->
```
mlyublena, line 169: perhans refactor this expression into a function containsNodeType(node, NodeTypesList)

also: should we recursively handle CTEConsumers here? Right now we allow this if we see at least one CTE consumer as data source, but the cte producer for that one may be a trivial one (like a table scan)?

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/LogicalCteOptimizer.java
```diff
@@ -132,34 +139,68 @@ public boolean isPlanRewritten()
...
             extends SimplePlanRewriter<CteTransformerContext>
     {
         private final PlanNodeIdAllocator idAllocator;
 
         private final VariableAllocator variableAllocator;
 
-        public CteConsumerTransformer(PlanNodeIdAllocator idAllocator, VariableAllocator variableAllocator)
+        private final Session session;
+
+        private static final List<Class<? extends PlanNode>> PRECOMPUTE_PLAN_NODES = ImmutableList.of(JoinNode.class, SemiJoinNode.class, AggregationNode.class);
```
mlyublena, line 151: should we call these two COMPLEX_PLAN_NODES and COMPLEX_DATA_SOURCE_NODES or something like that to better correlate with the session property name?

mlyublena (approved):

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/LogicalCteOptimizer.java
jaystarshot: Actually heuristic strategy flow for dependent CTEs is also not exactly accurate right now (eg: if a inner cte having a join/agg is materialized, the outer cte using this inner one might not be complex). This would need a completly different implementation which we would need to address in the cost based stategy. I was wondering if we could keep heuristic strategy as simple as possible

jaystarshot: PlanNodeSearcher.searchFrom(node) searches the node with CTEReferences, those are not rewritten as CteConsumers when this search happens

jaystarshot: If we visit the children of this node first and convert them to cte producers that might fix the problem. 
Also` Right now we allow this if we see at least one CTE consumer as data source, but the cte producer for that one may be a trivial one (like a values node)?` will never happen right? Since the heuristic strategy will ensures that the inner cte will never be converted to a cte producer.

pranjalssh (dismissed):

jaystarshot:
@mlyublena @pranjalssh Thank you for the review. However I found a test case added [here](https://github.com/prestodb/presto/pull/21720/files#diff-f1b72dc5b02cc72edb905847f0c8fe2b5599a929349f00ce8c902724a9469138R315) wherein, the CTE would be materialized even if it wasn't complex if the inner CTE was complex,
hence I changed the flow a bit to accommodate this.
1. Now, we first determine if a CTE is complex and then perform the rewrite.
2. Also cleaned up some session property handling, along with some renaming in the classes for clarity.
Please re-review!
cc: @tdcmeehan

pranjalssh (approved):

jaystarshot:
@pranjalssh Need a review again due to the comment change

tdcmeehan (approved):

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/LogicalCteOptimizer.java
jaystarshot: I fixed this test case in the refactor

mlyublena (approved):

PR #21727, opened 2024-01-18
[native] Fix exchange source close race condition

tanjialiang:
The race condition can happen in following sequence:
1. T1: PrestoExchangeSource tries to make a request by calling request(), attempting to assign a new promise to promise_.
2. T2: PrestoExchangeSource is closed by calling close(). This action realizes the stored promise_ and no further cleanup of the promise_ will be performed
3. T1: new promise is assigned to promise_.

This will cause that promise_ linger forever because this exchange source is already closed before promise_ assignment. The corresponding future will never get unblocked.
```
== NO RELEASE NOTE ==
```


mbasmanova:
CC: @heary-cloud

spershin (approved):
Nice catch!
The fix should work.

xiaoxmeng (approved):
@tanjialiang thanks for the fix!

PR #21734, opened 2024-01-19
Update default spill compression codec for native engine.

kevinwilfong:
We've seen that compressing spill files can substantially reduce query execution wall time for queries that spill a lot of data.

Updating the default codec from none to snappy. Snappy was chosen as it offers a reasonable balance between the CPU cost and compression ratio.

```
== NO RELEASE NOTE ==
```


amitkdutta (approved):
Thanks @kevinwilfong. Looks good to me.

NikhilCollooru (approved):

PR #21743, opened 2024-01-20
Add TableWriterMergeNode in presto UI

kewang1024:
```
== NO RELEASE NOTE ==
```


amitkdutta (approved):
Thanks @kewang1024 Ke

kewang1024:
Tested locally
From
image
to
image

NikhilCollooru (approved):

yhwang:
Thanks for the PR. Could you give me some background about the change? It was added around Aug 2022 and the UI hasn't been updated since then?

Based on the change, the `TableWriterMergeNode` node would be properly handled by the `getChildren` API, and be depicted on the graph.

Any example SQL using the tpch catalog I can try it out?

kewang1024:
> Thanks for the PR. Could you give me some background about the change? It was added around Aug 2022 and the UI hasn't been updated since then?
> 
> Based on the change, the `TableWriterMergeNode` node would be properly handled by the `getChildren` API, and be depicted on the graph.
> 
> Any example SQL using the tpch catalog I can try it out?

What was added Aug 2022? I'm not an UI expert so might need some context

The reason for this change is that we saw for TableWriterMerge Node, it's not showing correctly on UI and found the mapping is missing

Here is the query I used to reproduce:
```
use tpch;
CREATE TABLE hive.tpch.orders_ke ( "orderkey" bigint, "custkey" bigint, "orderstatus" varchar(1), "totalprice" double, "orderdate" varchar, "orderpriority" varchar(15), "clerk" varchar(15), "shippriority" integer, "comment" varchar(79) ) WITH ( format = 'DWRF' );
insert into orders_ke select orderkey, custkey, orderstatus, totalprice, '2011-01-01', '1', 'clerk', 2, 'comment' from orders;
```

yhwang:
> What was added Aug 2022? I'm not an UI expert so it would be great if you can provide some context on the UI/front end change

The `TableWriterMergeNode` JSON data struct was added around Aug 2022 :-). And when the data struct was added, there was no corresponding change on the UI back then.

I will check the JSON data returned from the API and see if all data types are handled on the UI side later. Thanks for the PR.

kewang1024:
> > What was added Aug 2022? I'm not an UI expert so it would be great if you can provide some context on the UI/front end change
> 
> The `TableWriterMergeNode` JSON data struct was added around Aug 2022 :-). And when the data struct was added, there was no corresponding change on the UI back then.
> 
> I will check the JSON data returned from the API and see if all data types are handled on the UI side later. Thanks for the PR.

Merged the PR to unblock the debugging process since otherwise we can't see the TableMergeNode in the UI

Thanks for letting me know, please keep me posted, pretty curious what has changed

PR #21755, opened 2024-01-23
[native] Fix task endpoint of server operation

tanjialiang:
Let v1/operation/task/listAll endpoint 1) print meaningful error message when parameter 'limit' is invalid. 2) return valid json result
```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/PrestoServerOperations.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoServerOperations.cpp
@@ -207,26 +207,26 @@ std::string PrestoServerOperations::taskOperation(
     }
     case ServerOperation::Action::kListAll: {
       uint32_t limit;
+      const auto& limitStr = message->getQueryParam("limit");
       try {
-        const auto& limitStr = message->getQueryParam("limit");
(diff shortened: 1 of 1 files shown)


presto-native-execution/presto_cpp/main/PrestoServerOperations.cpp
```diff
@@ -207,26 +207,26 @@ std::string PrestoServerOperations::taskOperation(
...
       std::stringstream oss;
       oss << "[";
       uint32_t count = 0;
-      for (const auto& task : taskMap) {
-        const auto& veloxTask = task.second->task;
+      for (auto taskItr = taskMap.begin(); taskItr != taskMap.end();
+           ++taskItr) {
+        const auto& veloxTask = taskItr->second->task;
         if (++count > limit) {
           oss << "... " << (taskMap.size() - limit) << " more tasks ...\n";
```
mbasmanova, line 225: @tanjialiang Would you double check that returned JSON is valid. Is seems problematic to add "... N more tasks ..." string in the middle of JSON array.

PR #21756, opened 2024-01-23
[native] Refactor server operation endpoint

tanjialiang:
* Let task/listAll endpoint to return valid json format
* Let task/listAll endpoint print meaningful error message when 'limit' is malformatted
* Let systemConfig/getProperty endpoint print meaningful error message if property does not exist.
```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/PrestoServerOperations.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoServerOperations.cpp
@@ -140,10 +140,11 @@ std::string PrestoServerOperations::systemConfigOperation(
           "Missing 'name' parameter for '{}.{}' operation",
           ServerOperation::targetString(op.target),
           ServerOperation::actionString(op.action));
-      return fmt::format(
-          "{}\n",
-          SystemConfig::instance()->optionalProperty(name).value_or(
-              "<default>"));
+      auto valueOpt = SystemConfig::instance()->optionalProperty(name);
+      VELOX_USER_CHECK(
+          valueOpt.has_value(),
+          fmt::format("Could not find property '{}'\n", name));
+      return fmt::format("{}\n", valueOpt.value());
     }
     default:
       break;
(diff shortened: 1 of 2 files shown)


xiaoxmeng (approved):
@tanjialiang LGTM. Thanks for the quick fix!

presto-native-execution/presto_cpp/main/PrestoServerOperations.cpp
```diff
@@ -207,28 +207,27 @@ std::string PrestoServerOperations::taskOperation(
...
       oss << "[";
       uint32_t count = 0;
-      for (const auto& task : taskMap) {
-        const auto& veloxTask = task.second->task;
+      for (auto taskItr = taskMap.begin(); taskItr != taskMap.end();
+           ++taskItr) {
+        const auto& veloxTask = taskItr->second->task;
         if (++count > limit) {
-          oss << "... " << (taskMap.size() - limit) << " more tasks ...\n";
+          oss << "... " << (taskMap.size() - limit) << " more tasks ...]\n";
```
mbasmanova, line 225: Curious, how do you verify that this returns valid JSON. Would you share an example response? It feels like it is not a valid JSON.

presto-native-execution/presto_cpp/main/PrestoServerOperations.cpp
```diff
@@ -143,7 +143,7 @@ std::string PrestoServerOperations::systemConfigOperation(
       return fmt::format(
           "{}\n",
           SystemConfig::instance()->optionalProperty(name).value_or(
-              "<default>"));
+              fmt::format("Could not find property '{}'", name)));
```
mbasmanova, line 146: Can we make sure to return HTTP 500 on errors?

mbasmanova (commented):
Some questions.

presto-native-execution/presto_cpp/main/PrestoServerOperations.cpp
tanjialiang: When limit is put, the option is to return an incomplete JSON with the ... n more tasks ... at the end to indicate the whole picture.

presto-native-execution/presto_cpp/main/PrestoServerOperations.cpp
tanjialiang: Yeah we can do that

presto-native-execution/presto_cpp/main/PrestoServerOperations.cpp
tanjialiang: But we can probably change it to return a JSON with limited tasks in it and a trailing statement saying n more tasks

xiaoxmeng (approved):

PR #21761, opened 2024-01-24
[native] Add 'server/setState' and 'server/announcer' server operations.

spershin:
The new server operations can be used for debugging: isolate a bad worker
from the cluster to unblock the cluster and then debug the worker.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/PeriodicServiceInventoryManager.cpp
+++ b/presto-native-execution/presto_cpp/main/PeriodicServiceInventoryManager.cpp
@@ -46,6 +46,11 @@ void PeriodicServiceInventoryManager::stop() {
   eventBaseThread_.stop();
 }
 
+void PeriodicServiceInventoryManager::enableRequest(bool enable) {
+  LOG(INFO) << (enable ? "Enabling " : "Disabling ") << id_;
+  requestEnabled_ = enable;
+}
+
 void PeriodicServiceInventoryManager::sendRequest() {
   // stop() calls EventBase's destructor which executed all pending callbacks;
   // make sure not to do anything if that's the case
@@ -75,6 +80,12 @@ void PeriodicServiceInventoryManager::sendRequest() {
   } catch (const std::exception& ex) {
     LOG(WARNING) << "Error occurred during updating service address: "
                  << ex.what();
+    scheduleNext();
+    return;
+  }
+
+  if (!requestEnabled_) {
+    LOG(INFO) << id_ << " skipped (it is disabled).";
     scheduleNext();
     return;
   }
--- a/presto-native-execution/presto_cpp/main/PeriodicServiceInventoryManager.h
+++ b/presto-native-execution/presto_cpp/main/PeriodicServiceInventoryManager.h
@@ -33,6 +33,9 @@ class PeriodicServiceInventoryManager {
 
   void stop();
 
+  /// If disabled then won't send any requests, but keeps itself running.
+  void enableRequest(bool enable);
+
  protected:
   // Denotes whether we retry failed requests due to network errors.
   virtual bool retryFailed() {
@@ -69,6 +72,7 @@ class PeriodicServiceInventoryManager {
   folly::SocketAddress serviceAddress_;
   std::shared_ptr<http::HttpClient> client_;
   std::atomic_bool stopped_{true};
+  std::atomic_bool requestEnabled_{true};
   uint64_t failedAttempts_{0};
   uint64_t attempts_{0};
 
--- a/presto-native-execution/presto_cpp/main/PrestoServer.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoServer.cpp
@@ -113,6 +113,18 @@ bool isCacheTtlEnabled() {
 
 } // namespace
 
+std::string nodeState2String(NodeState nodeState) {
+  switch (nodeState) {
+    case presto::NodeState::kActive:
+      return "active";
+    case presto::NodeState::kInActive:
+      return "inactive";
+    case presto::NodeState::kShuttingDown:
+      return "shutting_down";
+  }
+  return fmt::format("<unknown>:{}>", static_cast<int>(nodeState));
+}
+
 PrestoServer::PrestoServer(const std::string& configDirectoryPath)
     : configDirectoryPath_(configDirectoryPath),
       signalHandler_(std::make_unique<SignalHandler>(this)),
@@ -416,7 +428,7 @@ void PrestoServer::run() {
     }
   }
   prestoServerOperations_ =
-      std::make_unique<PrestoServerOperations>(taskManager_.get());
+      std::make_unique<PrestoServerOperations>(taskManager_.get(), this);
 
   // The endpoint used by operation in production.
   httpServer_->registerGet(
@@ -721,6 +733,12 @@ void PrestoServer::stop() {
       httpServer_->stop();
       PRESTO_SHUTDOWN_LOG(INFO) << "HTTP Server stopped.";
     }
+  }
+}
+
+void PrestoServer::enableAnnouncer(bool enable) {
+  if (announcer_ != nullptr) {
+    announcer_->enableRequest(enable);
   }
 }
 
--- a/presto-native-execution/presto_cpp/main/PrestoServerOperations.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoServerOperations.cpp
@@ -15,6 +15,7 @@
 #include <velox/common/base/Exceptions.h>
 #include <velox/common/base/VeloxException.h>
 #include <velox/common/process/TraceContext.h>
+#include "presto_cpp/main/PrestoServer.h"
 #include "presto_cpp/main/ServerOperation.h"
 #include "presto_cpp/main/common/Configs.h"
 #include "presto_cpp/main/http/HttpServer.h"
@@ -242,15 +243,73 @@ std::string PrestoServerOperations::taskOperation(
 
 std::string PrestoServerOperations::serverOperation(
     const ServerOperation& op,
-    proxygen::HTTPMessage* /* unused */) {
+    proxygen::HTTPMessage* message) {
   switch (op.action) {
-    case ServerOperation::Action::kTrace: {
-      return velox::process::TraceContext::statusLine();
-    }
+    case ServerOperation::Action::kTrace:
+      return serverOperationTrace();
+    case ServerOperation::Action::kSetState:
+      return serverOperationSetState(message);
+    case ServerOperation::Action::kAnnouncer:
+      return serverOperationAnnouncer(message);
     default:
       break;
   }
   return unsupportedAction(op);
+}
+
+std::string PrestoServerOperations::serverOperationTrace() {
+  return velox::process::TraceContext::statusLine();
+}
+
+std::string PrestoServerOperations::serverOperationSetState(
+    proxygen::HTTPMessage* message) {
+  if (server_) {
+    const auto& stateStr = message->getQueryParam("state");
+    const auto prevState = server_->nodeState();
+    NodeState newNodeState{NodeState::kActive};
+    if (stateStr == "active") {
+      newNodeState = NodeState::kActive;
+    } else if (stateStr == "inactive") {
+      newNodeState = NodeState::kInActive;
+    } else if (stateStr == "shutting_down") {
+      newNodeState = NodeState::kShuttingDown;
+    } else {
+      return fmt::format(
+          "Invalid state '{}'. "
+          "Supported states are: 'active', 'inactive', 'shutting_down'. "
+          "Example: server/setState?state=shutting_down",
+          stateStr);
+    }
+    if (newNodeState != prevState) {
+      LOG(INFO) << "Setting node state to " << nodeState2String(newNodeState);
+      server_->setNodeState(newNodeState);
+    }
+    return fmt::format(
+        "New node state: '{}', previous state: '{}'.",
+        nodeState2String(newNodeState),
+        nodeState2String(prevState));
+  }
+  return "No PrestoServer to change state of (it is nullptr).";
+}
+
+std::string PrestoServerOperations::serverOperationAnnouncer(
+    proxygen::HTTPMessage* message) {
+  if (server_) {
+    const auto& actionStr = message->getQueryParam("action");
+    if (actionStr == "enable") {
+      server_->enableAnnouncer(true);
+      return "Announcer enabled";
+    } else if (actionStr == "disable") {
+      server_->enableAnnouncer(false);
+      return "Announcer disabled";
+    }
+    return fmt::format(
+        "Invalid action '{}'. "
+        "Supported actions are: 'enable', 'disable'. "
+        "Example: server/announcer?action=disable",
+        actionStr);
+  }
+  return "No PrestoServer to change announcer of (it is nullptr).";
 }
 
 } // namespace facebook::presto
--- a/presto-native-execution/presto_cpp/main/PrestoServerOperations.h
+++ b/presto-native-execution/presto_cpp/main/PrestoServerOperations.h
@@ -23,13 +23,14 @@ class ResponseHandler;
 
 namespace facebook::presto {
 
+class PrestoServer;
 struct ServerOperation;
 
 /// Static class implements Presto Server Operations.
 class PrestoServerOperations {
  public:
-  PrestoServerOperations(TaskManager* const taskManager)
-      : taskManager_(taskManager) {}
+  PrestoServerOperations(TaskManager* taskManager, PrestoServer* server)
+      : taskManager_(taskManager), server_(server) {}
 
   void runOperation(
       proxygen::HTTPMessage* message,
@@ -55,7 +56,15 @@ class PrestoServerOperations {
       const ServerOperation& op,
       proxygen::HTTPMessage* message);
 
+ private:
+  std::string serverOperationTrace();
+
+  std::string serverOperationSetState(proxygen::HTTPMessage* message);
+
+  std::string serverOperationAnnouncer(proxygen::HTTPMessage* message);
+
   TaskManager* const taskManager_;
+  PrestoServer* const server_;
 };
 
 } // namespace facebook::presto
--- a/presto-native-execution/presto_cpp/main/ServerOperation.h
+++ b/presto-native-execution/presto_cpp/main/ServerOperation.h
@@ -32,13 +32,26 @@ struct ServerOperation {
 
   /// The action this operation is trying to take
   enum class Action {
+    /// Applicable to kConnector. Clears the connector cache.
     kClearCache,
+    /// Applicable to kConnector. Returns stats of the connector cache.
     kGetCacheStats,
+    /// Applicable to kSystemConfig & kVeloxQueryConfig. Modifies the value of a
+    /// single property.
     kSetProperty,
+    /// Applicable to kSystemConfig & kVeloxQueryConfig. Returns the value of a
+    /// single property.
     kGetProperty,
+    /// Applicable to kTask. Returns detailed info on one Task.
     kGetDetail,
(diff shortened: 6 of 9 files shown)


presto-native-execution/presto_cpp/main/PrestoServerOperations.cpp
```diff
@@ -242,11 +243,56 @@ std::string PrestoServerOperations::taskOperation(
...
+      if (server_) {
+        const auto& actionStr = message->getQueryParam("action");
+        if (actionStr == "enable") {
+          server_->enableAnnouncer(true);
+          return "Announcer enabled";
+        } else if (actionStr == "disable") {
+          server_->enableAnnouncer(false);
+          return "Announcer disabled";
+        }
+        return fmt::format(
```
mbasmanova, line 289: Can we report errors using HTTP 500 ?

presto-native-execution/presto_cpp/main/PrestoServerOperations.cpp
```diff
@@ -242,11 +243,56 @@ std::string PrestoServerOperations::taskOperation(
...
+            "Invalid action '{}'. "
+            "Supported actions are: 'enable', 'disable'. "
+            "Example: server/announcer?action=disable",
+            actionStr);
+      }
+      return "No PrestoServer to change announcer of (it is nullptr).";
```
mbasmanova, line 295: ditto

mbasmanova (commented):
It would be nice to update PR description to explain the new APIs and how they can be used.

presto-native-execution/presto_cpp/main/PrestoServerOperations.cpp
spershin: I don't see why we need this. It is not error, as such.

presto-native-execution/presto_cpp/main/PrestoServerOperations.cpp
spershin: That's only for the test. In presto_cpp it will never by null.

tdcmeehan:
Why add `server/setState` when we have `/v1/info/state`?

spershin:
@tdcmeehan 

"/v1/info/state" is a getter, while "server/setState" is a way in the generic server operation debug endpoint to set (change) the state.
In addition, inside Meta, with pt tool we can do it to multiple hosts with a single pt command.

spershin:
> It would be nice to update PR description to explain the new APIs and how they can be used.

This will/can be added to the server operation docs later?

Also, the code has the LOG statement that shows it.

tdcmeehan:
> @tdcmeehan
> 
> "/v1/info/state" is a getter, while "server/setState" is a way in the generic server operation debug endpoint to set (change) the state. In addition, inside Meta, with pt tool we can do it to multiple hosts with a single pt command.

You can do that with a `PUT` to `/v1/info/state`.

spershin:
@tdcmeehan 
Inside Meta, with pt tool we can do it to multiple hosts with a single pt command.
Can put do the same?

Also nice to have debug capabilities under one umbrella. :)

tdcmeehan:
Your PT tools is probably issuing multiple GETs, one per host.  In the same way you could issue multiple PUTs.

Most Presto endpoints are RESTful, and this is just not RESTful.  It's a custom API with custom, unique behavior.  In general, for people building their own versions of pt tools, it is strange and unexpected for GETs to mutate anything.

spershin:
@tdcmeehan 

> Your PT tools is probably issuing multiple GETs, one per host. In the same way you could issue multiple PUTs.

They aren't mine PT tools. :)

However, for me, as a user of this functionality, it does not matter what pt tool is doing behind the scenes. What matters is convenience, that comes with it. It is the most important thing, not some rules. :)
I'm unaware how can I send the same request to multiple workers using curl, for example and I don't want to spend time finding it out unless I have to. If the curl way can do the same (multiple targets) - good for it, other folks can use it as their most convenient way. No one is hurt.
To add to this, pt tool supports integration with TW, (for example when you know task IDs of the workers you want to affect).

You are talking about GETs making some changes.
Well, as an engineer issuing
`curl "twshared26075.05.pnb1:7777/v1/operation/systemConfig/setProperty?name=shutdown-onset-sec&value=7"`
(for example) you should notice SET in the command and some flexibility to accept that it would change something. On top of that - why do you issue that command at all - you must have wanted that to change property in the 1st place, right?

Also, I don't see in the above 'curl' example any indication of 'GET' vs 'PUT'. Or are you talking about how it is registered in the code (via registerGet)?

So, what refactoring of the 'v1/operation' should look like from your perspective to make people happy?

tdcmeehan:
Here's an alternative that I would approve:

`PATCH` to `twshared26075.05.pnb1:7777/v1/config` -- the HTTP verb tells you that it's a patch, meaning partial update to an existing object (the `config`)

Body (MIME type `text/plain`): `shutdown-onset-sec=7` (same format as .properties file)

`GET` would just return the entire config.
`PUT` could optionally be implemented to put the entire config in situ and would have the same format as `PATCH` (same as .properties file)

tdcmeehan:
Likewise, there's already a `/v1/info/state` in Java workers that I'd encourage you to use for consistency (and this inevitably must also be in the PT tools) so I don't think we need to use this new endpoint to set the state.

presto-native-execution/presto_cpp/main/PrestoServerOperations.h
```diff
@@ -23,13 +23,14 @@ class ResponseHandler;
...
 
+class PrestoServer;
 struct ServerOperation;
 
 /// Static class implements Presto Server Operations.
 class PrestoServerOperations {
  public:
-  PrestoServerOperations(TaskManager* const taskManager)
-      : taskManager_(taskManager) {}
+  PrestoServerOperations(TaskManager* const taskManager, PrestoServer* server)
```
xiaoxmeng, line 32: NYC: drop pointer const from function input

presto-native-execution/presto_cpp/main/PrestoServerOperations.h
```diff
@@ -56,6 +57,7 @@ class PrestoServerOperations {
       proxygen::HTTPMessage* message);
 
   TaskManager* const taskManager_;
+  PrestoServer* server_;
```
xiaoxmeng, line 60:
```
PrestoServer* const server_;
```

presto-native-execution/presto_cpp/main/PeriodicServiceInventoryManager.cpp
```diff
@@ -79,6 +79,12 @@ void PeriodicServiceInventoryManager::sendRequest() {
     return;
   }
 
+  if (!requestEnabled_) {
+    LOG(INFO) << id_ << " skipped (it is disabled).";
```
xiaoxmeng, line 83: LOG_EVERTY_N?

presto-native-execution/presto_cpp/main/PeriodicServiceInventoryManager.h
```diff
@@ -33,6 +33,11 @@ class PeriodicServiceInventoryManager {
 
   void stop();
 
+  /// If disabled then won't send any requests, but keeps itself running.
+  void enableRequest(bool enable) {
+    requestEnabled_ = enable;
```
xiaoxmeng, line 38: Shall we add a log whenever requestEnabled_ is changed?
```
if (requestEnabled_.exchange(enable) != enable) {
   LOG(INFO) << enable or disable?
}
```

presto-native-execution/presto_cpp/main/PrestoServerOperations.h
xiaoxmeng, line 32: Can we access task manager from Presto server by providing an accessor in PrestoServer? Then we just need to pass PrestoServer here?

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -113,6 +113,18 @@ bool isCacheTtlEnabled() {
...
 } // namespace
 
+const char* nodeState2String(NodeState nodeState) {
+  switch (nodeState) {
+    case presto::NodeState::kActive:
+      return "active";
+    case presto::NodeState::kInActive:
+      return "inactive";
+    case presto::NodeState::kShuttingDown:
+      return "shutting_down";
```
xiaoxmeng, line 123:
```
std::string nodeState2String() {

default:
   return fmt::format("<unknow:{}>", static_cast<int>(nodeState));
```

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -724,6 +736,13 @@ void PrestoServer::stop() {
   }
 }
 
+void PrestoServer::enableAnnouncer(bool enable) {
+  if (announcer_ != nullptr) {
+    LOG(INFO) << (enable ? "Enabling announcer..." : "Disabling announcer...");
```
xiaoxmeng, line 741: If we have log in enableRequest, then we don't need log here.

presto-native-execution/presto_cpp/main/PrestoServerOperations.cpp
```diff
@@ -242,11 +243,56 @@ std::string PrestoServerOperations::taskOperation(
...
     const ServerOperation& op,
-    proxygen::HTTPMessage* /* unused */) {
+    proxygen::HTTPMessage* message) {
   switch (op.action) {
-    case ServerOperation::Action::kTrace: {
+    case ServerOperation::Action::kTrace:
       return velox::process::TraceContext::statusLine();
-    }
+    case ServerOperation::Action::kSetState:
+      if (server_) {
```
xiaoxmeng, line 251: nit: can we have a dispatch method to handle this? Thanks!

xiaoxmeng (commented):
@spershin thanks for the change % minors.

spershin:
@tdcmeehan 

That is a fair proposal ("/v1/config" and "/v1/info/state", PATCH, PUT and GET).

However, what does not seem right here is the growing number of these endpoints (and numerous versions of them (PUT, GET, PATCH)) as we continue adding to the list.
On top of that - inability to distinguish easily which is a debug endpoint (currently 'v1/operation' is the 'one that rules them all') and which is a normal one (original Presto). So basically, not under one umbrella. As the result, hard for engineers to locate/decide on which one to use and hard to maintaining separate security for debug and 'normal' endpoints if we chose (and we should).

I suggest making some exceptions for 'v1/operation'.
It can be flexible, for example, we can split all mutating ones into PUT (I'm super skeptical about the PATCH name, which so does not sound right unless you came from a strong HTTP background, I suppose) and the getters into GET flavors to appease the RESTness of the Presto protocol.

What do you think?

presto-native-execution/presto_cpp/main/ServerOperation.h
```diff
@@ -39,6 +39,8 @@ struct ServerOperation {
     kGetDetail,
     kListAll,
     kTrace,
+    kSetState,
```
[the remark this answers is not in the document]
spershin: @xiaoxmeng 
Yes, I too find it very confusing having actions so separated from entities.
Will do.

presto-native-execution/presto_cpp/main/PeriodicServiceInventoryManager.h
spershin: We do this in PrestoServer::enableAnnouncer(), which is on a slightly higher level than this base class.

presto-native-execution/presto_cpp/main/PrestoServerOperations.h
spershin: We can do this, however that would break some unit tests, where we have task manager, but not the server.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
spershin: Using 'default' is not a great idea: when new enum members are added you will not have compiler warning -> error for them not used in the switch statement.
That's why I prefer having a separate return. We can add the number though.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
spershin: He-he, sure, I can move. :)

presto-native-execution/presto_cpp/main/PrestoServerOperations.cpp
spershin: You mean a method per action?
Sure, we can do that.

presto-native-execution/presto_cpp/main/PrestoServerOperations.cpp
spershin: @mbasmanova 
I'm changing the PR according to some comment.
I was looking to address yours, but it will require a bit more intrusive change to the whole PrestoServerOperations class.

Should we do this in the follow up?

mbasmanova: @spershin Sergey, thank you for considering my comments. A follow-up would be fine with me.

xiaoxmeng (approved):
@spershin LGTM. We might consider to separate get/post in followup for mutate and non-mutate operations. And then we can apply more security check for mutate operation like using https. Thanks!

spershin:
@tdcmeehan 

Sure, I already agreed to split the 'v1/operation' endpoint into two in the followups: mutating and not.
Would splitting into PUT and GET suffice?

As I understand, that would make calling the endpoints more complicated as one will have to specify PUT/GET on top of the end point and parameters, right? That would be somewhat offset by removing action prefixes from the action name.

Also, we are going to do something like this:
convert
```
curl "hostname:7777/v1/operation/systemConfig/getProperty?name=shutdown-onset-sec"
curl "hostname:7777/v1/operation/systemConfig/setProperty?name=shutdown-onset-sec&value=7"

```
into
```
curl GET "hostname:7777/v1/operation/systemConfig/property?name=shutdown-onset-sec"
curl PUT "hostname:7777/v1/operation/systemConfig/property?name=shutdown-onset-sec&value=7"

```
right?
Would that be aligned with the REST?

tdcmeehan:
@spershin this is better, but I still have some suggestions.

See doc for HTTP `PUT`: https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PUT

The data you upload in a `PUT` is expected to be in the body.  Here, we are adding it as a query parameter.  This of course works, but it's not standard.  I'd recommend putting the value in the body (see below).

Moreover, in REST, the URL describes the object we're mutating.  `hostname:7777/v1/operation/systemConfig/property` is an improvement because we're no longer putting the verb in the URL (where we expect the object to be), but I'd recommend getting rid of `property` as it's redundant.  Just put in the property name directly in the path: e.g., instead of `systemConfig/property?name=fizzbuzz`, do `systemConfig/fizzbuzz`.

See more clear examples below.  If easy, I'd recommend also getting rid of `operation` as it doesn't seem to add much value or clarity to the endpoint name.

---

HTTP verb: `PUT`

URL: `hostname:7777/v1/systemConfig/shutdown-onset-sec`

Body (MIME type: text/plain): `7`

---

HTTP verb: `PATCH`

URL: `hostname:7777/v1/systemConfig`

Body (MIME type: text/plain): `shutdown-onset-sec=7`

---

HTTP verb: `GET`

URL: `hostname:7777/v1/systemConfig/shutdown-onset-sec`

Response body (MIME type: text/plain): `7`



---

HTTP verb: `GET`

URL: `hostname:7777/v1/systemConfig`

Body (MIME type: text/plain): all system configs in .properties format

tdcmeehan:
@spershin will you be following up on this?

spershin:
@tdcmeehan 
With all respect, not in the way you put it in the last message - it is simply another attempt to push the exactly same idea you had at the beginning of the message exchange.
Does not look like you paid attention to the arguments I provided. :(

tdcmeehan:
@spershin committers are given trust to merge code that is up to the quality of the rest of the committer group. When someone expresses a disagreement, it is your responsibility to either address it or list trade offs and put a path forward.  I really hate to give it a red check to prevent merge, but if that's what it takes to address feedback that's what we can do. My recommendation is to address all feedback all the time and drive to resolution, same as every other committer. Thanks!

spershin:
@tdcmeehan 

> committers are given trust to merge code that is up to the quality of the rest of the committer group. When someone expresses a disagreement, it is your responsibility to either address it or list trade offs and put a path forward. I really hate to give it a red check to prevent merge, but if that's what it takes to address feedback that's what we can do. My recommendation is to address all feedback all the time and drive to resolution, same as every other committer. Thanks!

Sad it had to come to this.

I had impression that I had a proposal of the path forward and it did not look like what I have started with (which you admitted), however you turned yours into the exactly the one you started with. That's not how the negotiation and discussion is done. Unless it is not a negotiation and there is only one way, pretty much (I really hope not).

So I'm proposing the path forward again and listing my arguments to support them:

1. I suggest we keep a common prefix for all operational endpoints, so it is very clear for EVERYONE and not just a handful of people with deep knowledge of Presto, that it is a special class of endpoints that are different from the classic Presto REST API. And that they should be used with care. I believe that however good documentation is - it cannot replace a good distinct readable endpoint (just like variable, function, etc names). We can change the name 'operation' to something else, if you feel it is not fitting. In addition, it would be easy to add any security on that, as it would be a single class of endpoints starting from the same object. I don't see it breaks the REST paradigm, as we can consider "operation" as an object/domain here.

2. I'm questioning "...systemConfig/shutdown-onset-sec" format, as this might be confusing to people who see that for the 1st time. Imagine "...systemConfig/testing.spill_pct", "...systemConfig/presto.array_agg.ignore_nulls". Can be anything after "systemConfig/" and throw people off the track. It could be confusing. Having "...systemConfig/property/testing.spill_pct" is removing many questions that an engineer might have. For this one - I'm flexible - we can drop it. Once again, I don't see how it breaks the REST paradigm, as the property is an object under a config.

3. Passing the actual values in the message body. That can be done if not much difference supplying this in whatever tool the people will use to do it.

4. Here is a question for a message body: why don't then we do the following: "...systemConfig/property" body("propname=val")?

5. Using PUT and GET - I already agreed with this point way above.

Let me know what you think.

tdcmeehan:
On the technical points:

I'm not a fan of keeping `operation` in the endpoint name because it's ambiguous.  I hear what you're saying about making it very clear to users that these may have unpredictable effects though, and I'm fine to agree to disagree.

On properties, I can see the argument of using `systemConfig/property/xyz`--one can also have `systemConfig/properties` to retrieve all, or some such.  Let's just get it out of the param and make it path variable, as you've done in your example.

Agreed on using the body for inputs.

> Here is a question for a message body: why don't then we do the following: 
> "...systemConfig/property" body("propname=val")?

It should be one of two things, either `systemConfig/properties`, with a `PATCH` that updates the config as you have in your example, or `systemConfig/property/propname` with a PUT updating the just the value from the body.  We're supposed to name a specific thing in the path--in this case this is the properties.  So we use the plural to signify either an update to the properties as a whole and `PATCH` to partially update, or specify the specific property in the path and use `PUT`.

I know you expressed concern that `PATCH` is too esoteric and not used commonly, but actually this is a well documented HTTP verb and I think nearly every HTTP library should support its use.  To me, using `PATCH` is very clear, because it's embedded into the verb exactly what's supposed to happen by issuing the command.

On the non-technical points:

I don't see technical discussions as negotiations.  Nor am I here to give you trouble or try to hold you up.  We want Prestissimo to get to the same level of maturity we see in Java.  One thing Java does well is it has a principled approach to naming and service discovery of its internal and external endpoints.  This is something highly valued and not desirable to throw out in our move to Prestissimo.  It's also not something I think that can be negotiated--it's a principle of design, and if we want to change this principle, that needs to be clearly elucidated to everyone why we want to suddenly do things differently.  My suggestions are definitely not law either: I'm merely trying to give you examples that I think better fit in the existing conventions and design.  You're of course free to counter with alternatives, as you've done.  What we shouldn't do though is just toss it out the window and say we're doing it differently now--that's a decision we make as a group, not on one-off bases.

PR #21763, opened 2024-01-24
[native] Let TaskManager abort existing running tasks during shutdown

tanjialiang:
During shutdown, TaskManager waits for the tasks to complete. It will potentially wait for a long time before things get shutdown. We don't really care about those running tasks when choosing to shutdown a cluster so we force detele these running tasks to make shutdown more smooth.
```
== NO RELEASE NOTE ==
```


tanjialiang:
Abandoning

PR #21765, opened 2024-01-24
[native] Add support for Iceberg connector in presto_protocol

imjalpreet:
## Description

Add Java classes from the Iceberg connector in presto_protocol.yml file.
Specialize IcebergColumnHandle to support 'operator<()'.

Depends on #21764 

## Motivation and Context

https://github.com/facebookincubator/velox/issues/5977

## Impact

With these changes, we will be able to read iceberg tables from Prestissimo

## Test Plan

Successfully ran all TPC-H queries with iceberg tables (both partitioned and non-partitioned)

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Add support for Iceberg connector in presto_protocol
```

### DIFF
--- a/presto-native-execution/presto_cpp/presto_protocol/presto_protocol.yml
+++ b/presto-native-execution/presto_cpp/presto_protocol/presto_protocol.yml
@@ -50,6 +50,7 @@ AbstractClasses:
       comparable: true
       subclasses:
         - { name: HiveColumnHandle,         key: hive }
+        - { name: IcebergColumnHandle,      key: hive-iceberg }
         - { name: TpchColumnHandle,         key: tpch }
 
     ConnectorPartitioningHandle:
@@ -63,6 +64,7 @@ AbstractClasses:
       super: JsonEncodedSubclass
       subclasses:
         - { name: HiveTableHandle,          key: hive }
+        - { name: IcebergTableHandle,       key: hive-iceberg }
         - { name: TpchTableHandle,          key: tpch }
 
     ConnectorOutputTableHandle:
@@ -85,6 +87,7 @@ AbstractClasses:
       super: JsonEncodedSubclass
       subclasses:
         - { name: HiveTableLayoutHandle,    key: hive }
+        - { name: IcebergTableLayoutHandle, key: hive-iceberg }
         - { name: TpchTableLayoutHandle,    key: tpch }
 
     ConnectorMetadataUpdateHandle:
@@ -96,6 +99,7 @@ AbstractClasses:
       super: JsonEncodedSubclass
       subclasses:
         - { name: HiveSplit,            key: hive }
+        - { name: IcebergSplit,         key: hive-iceberg }
         - { name: TpchSplit,            key: tpch }
         - { name: RemoteSplit,          key: $remote }
         - { name: EmptySplit,           key: $empty }
@@ -177,6 +181,7 @@ JavaClasses:
   - presto-hive-metastore/src/main/java/com/facebook/presto/hive/BucketFunctionType.java
   - presto-hive-common/src/main/java/com/facebook/presto/hive/CacheQuotaRequirement.java
   - presto-hive-common/src/main/java/com/facebook/presto/hive/CacheQuotaScope.java
+  - presto-hive-common/src/main/java/com/facebook/presto/hive/BaseHiveColumnHandle.java
   - presto-spi/src/main/java/com/facebook/presto/spi/relation/CallExpression.java
   - presto-hive-metastore/src/main/java/com/facebook/presto/hive/metastore/Column.java
   - presto-verifier/src/main/java/com/facebook/presto/verifier/framework/Column.java
@@ -233,6 +238,18 @@ JavaClasses:
   - presto-hive-metastore/src/main/java/com/facebook/presto/hive/HiveTableHandle.java
   - presto-hive/src/main/java/com/facebook/presto/hive/HiveTableLayoutHandle.java
   - presto-hive/src/main/java/com/facebook/presto/hive/HiveTransactionHandle.java
+  - presto-iceberg/src/main/java/com/facebook/presto/iceberg/FileContent.java
+  - presto-iceberg/src/main/java/com/facebook/presto/iceberg/FileFormat.java
+  - presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergSplit.java
+  - presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergTableHandle.java
+  - presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergTableName.java
+  - presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergTableType.java
+  - presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergTableLayoutHandle.java
+  - presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergColumnHandle.java
+  - presto-iceberg/src/main/java/com/facebook/presto/iceberg/ColumnIdentity.java
+  - presto-iceberg/src/main/java/com/facebook/presto/iceberg/delete/DeleteFile.java
+  - presto-iceberg/src/main/java/com/facebook/presto/iceberg/changelog/ChangelogOperation.java
+  - presto-iceberg/src/main/java/com/facebook/presto/iceberg/changelog/ChangelogSplitInfo.java
   - presto-tpch/src/main/java/com/facebook/presto/tpch/TpchSplit.java
   - presto-tpch/src/main/java/com/facebook/presto/tpch/TpchTableHandle.java
   - presto-tpch/src/main/java/com/facebook/presto/tpch/TpchTableLayoutHandle.java
@@ -322,4 +339,3 @@ JavaClasses:
   - presto-spi/src/main/java/com/facebook/presto/spi/plan/JoinType.java
   - presto-spi/src/main/java/com/facebook/presto/spi/plan/JoinDistributionType.java
   - presto-spi/src/main/java/com/facebook/presto/spi/plan/EquiJoinClause.java
-  - presto-hive-common/src/main/java/com/facebook/presto/hive/BaseHiveColumnHandle.java
--- a/presto-native-execution/presto_cpp/presto_protocol/presto_protocol.cpp
+++ b/presto-native-execution/presto_cpp/presto_protocol/presto_protocol.cpp
@@ -54,181 +54,6 @@ std::string json_map_key(const VariableReferenceExpression& p) {
 }
 } // namespace facebook::presto::protocol
 
-namespace facebook::presto::protocol {
-// Loosly copied this here from NLOHMANN_JSON_SERIALIZE_ENUM()
-
-// NOLINTNEXTLINE: cppcoreguidelines-avoid-c-arrays
-static const std::pair<RuntimeUnit, json> RuntimeUnit_enum_table[] =
-    { // NOLINT: cert-err58-cpp
-        {RuntimeUnit::NONE, "NONE"},
-        {RuntimeUnit::NANO, "NANO"},
-        {RuntimeUnit::BYTE, "BYTE"}};
-void to_json(json& j, const RuntimeUnit& e) {
-  static_assert(
-      std::is_enum<RuntimeUnit>::value, "RuntimeUnit must be an enum!");
-  const auto* it = std::find_if(
-      std::begin(RuntimeUnit_enum_table),
-      std::end(RuntimeUnit_enum_table),
-      [e](const std::pair<RuntimeUnit, json>& ej_pair) -> bool {
-        return ej_pair.first == e;
-      });
-  j = ((it != std::end(RuntimeUnit_enum_table))
-           ? it
-           : std::begin(RuntimeUnit_enum_table))
-          ->second;
-}
-void from_json(const json& j, RuntimeUnit& e) {
-  static_assert(
-      std::is_enum<RuntimeUnit>::value, "RuntimeUnit must be an enum!");
-  const auto* it = std::find_if(
-      std::begin(RuntimeUnit_enum_table),
-      std::end(RuntimeUnit_enum_table),
-      [&j](const std::pair<RuntimeUnit, json>& ej_pair) -> bool {
-        return ej_pair.second == j;
-      });
-  e = ((it != std::end(RuntimeUnit_enum_table))
-           ? it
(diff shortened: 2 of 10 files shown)


majetideepak:
@imjalpreet is this ready for review?

imjalpreet:
@majetideepak Yes the changes are all there in this PR. But need to rebase on master and make some minor updates based on new changes added in master. But I was facing a build error in my local most likely related to https://github.com/prestodb/presto/pull/21797 so I haven't pushed the rebased version of this branch yet.

Once I verify the changes after a successful build, I will push them as well. You can still review the PR since it mostly has all the changes or we can wait for the rebased version which has minor additions.

majetideepak:
Why do we need so many `special` changes? Are we not able to translate the Java source files directly?
We should avoid adding to `special` as they are harder to maintain.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -852,6 +853,11 @@ std::vector<std::string> PrestoServer::registerConnectors(
       PRESTO_STARTUP_LOG(INFO) << "Registering catalog " << catalogName
                                << " using connector " << connectorName;
 
+      // TODO: Temporary Change to use HiveDataSource for Iceberg Connector
```
yingsu00, line 856: Moving the discussion in #21584 to here:

Add an iceberg.properties in presto-native-execution/etc/catalog with the following
```
# The Presto "iceberg" catalog is handled by the hive connector in Presto native execution.
connector.name=hive

cache.enabled=true
```

presto-native-execution/presto_cpp/presto_protocol/Connectors.cpp
```diff
@@ -44,6 +44,10 @@ void registerHiveConnectors() {
   registerConnector("hive-hadoop2", "hive");
 }
 
+void registerIcebergConnector() {
+  registerConnector("iceberg", "iceberg");
```
yingsu00, line 48: Let's keep this as is, but add a comment saying the connectorName and connectorKey are for Presto, not Presto native.

mbasmanova:
> Why do we need so many `special` changes? Are we not able to translate the Java source files directly? We should avoid adding to `special` as they are harder to maintain.

@majetideepak @tdcmeehan Deepak, Tim, do we have any ongoing efforts to switch from JSON to Thrift for communication? JSON-based protocols are pretty challenging to maintain.

tdcmeehan:
@mbasmanova  For clarity, there are a few reasons this is not scalable:

1) We must manually update the protocol every time connectors change or there is a new connector.
2) This protocol must be built-in to the Prestissimo binaries.

While 1 is problematic, I believe 2 is the problem that should be solved first (if we must choose between the two), because while 1 is an inconvenience, 2 is more than an inconvenience: it's a blocker from real connector SPI support in C++.

Consider that, to solve 2, we must make it so that connector data structures are not built, but deserialized in a custom way (which is how it works in Java, albeit tied JSON).  The idea is to make it so that connector data structures are serialized in a custom manner, and allow connectors to self-specify how to serialize and deserialize them.  This may include Thrift, but it will be up to the connector how to accomplish that.

When thought of in this way, you're left with two parts to serialize and deserialize: the outer shell of the task update request, and the inner structures used by connectors.  The outer shell may be moved to Thrift.  We would love to get community help to migrate this as well, but I don't think it's the top priority since it changes more slowly and it doesn't block federation use cases in Prestissimo.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
imjalpreet: @yingsu00 I have added the catalog file in #21584 since the final changes are part of that PR

presto-native-execution/presto_cpp/presto_protocol/Connectors.cpp
imjalpreet: @yingsu00 I have made the change as we had discussed earlier and moved the iceberg registration as part of the hive connectors

presto-native-execution/presto_cpp/presto_protocol/special/BaseHiveColumnHandle.hpp.inc
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+namespace facebook::presto::protocol {
+
+struct BaseHiveColumnHandle : public ColumnHandle {
```
mbasmanova, line 16: Would you add comments to explain why all these classes cannot be auto-generated?

majetideepak: @mbasmanova I am working with @imjalpreet to try and automate these as well. We will add comments if we are unable to auto-generate.

mbasmanova: @majetideepak Got it. Thank you for sharing. Great to know that you are trying to avoid manual overrides. Thanks.

presto-native-execution/presto_cpp/presto_protocol/special/ColumnType.cpp.inc
aditi-pandit, line 1: Can this enum be automated as well ? I feel it should be.

aditi-pandit (commented):
We should try to automate the Column(Table)Handle and TableLayout classes.

Are we blocked on them because of the use of inheritance e.g. `struct HiveTableHandle : public BaseHiveTableHandle` ?

presto-native-execution/presto_cpp/presto_protocol/special/ColumnType.cpp.inc
imjalpreet: Oh, yes I missed removing this enum. I had already refactored on the Java side to be able to remove this special class.

imjalpreet: I will remove this in my next push

imjalpreet:
> Are we blocked on them because of the use of inheritance e.g. struct HiveTableHandle : public BaseHiveTableHandle ?

@aditi-pandit Yes, we are trying to see if it would be possible to somehow incorporate multi-level inheritance in these classes (ColumnHandle, TableHandle, and TableLayoutHandle) without the introduction of special classes.

yingsu00:
@majetideepak @mbasmanova @tdcmeehan @aditi-pandit Thanks everyone for chiming in. @imjalpreet and I had some offline discussion today. We think there are two ways to solve this problem:

1. Make presto protocol to support multi-level inheritance (This PR)
In the last a couple of days, @imjalpreet has been trying to modify the Ser/De code to make it support multi-level inheritance and was able to reduce most extra specials. In this approach, presto_protocol.yml was modified to add a "super" member in the subclasses lists in addition to the current "name" and "key" members. We also had to move the `partitionColumns` back from BaseHiveTableLayoutHandle to HiveTableLayoutHandle and IcebergTableLayoutHandle to resolve the problem the `partitionColumns` could not be deserialized because it was List<BaseHiveColumnHandle> and the deserializer doesn't know its actual type.  Jalpreet has updated this PR with this approach.

2. Only serialize leaf classes
    In this approach , BaseHiveTableLayoutHandle and BaseHiveColumnHandle will be removed from presto_protocol, and only HiveTableLayoutHandle, IcebergTableLayoutHandle and HiveColumnHandle, IcebergColumnHandle would be serialized/deserialized. Then we will remove the references of these base classes in presto-native-execution (there is only one real callsite and it can be removed). @imjalpreet can also send a PR on this.

Both approachs need additional work. Before we proceed, we'd like to get your opinions which route we are going to take. IMHO the first approach is more generic and may benefit future changes. But we may need to use a slightly different approach there: instead of adding a "super" member in presto_protocol.yml, I think it's better to list the classes in a separate entry:
```
    ConnectorTableHandle:
      super: JsonEncodedSubclass
      subclasses:
        - { name: BaseHiveTableHandle,      key: base-hive }
        - { name: TpchTableHandle,          key: tpch }
        -
    BaseHiveTableHandle:
      super: JsonEncodedSubclass
      subclasses:
        - { name: HiveTableHandle,          key: hive }
        - { name: IcebergTableHandle,       key: hive-iceberg }
```
This needs additional work in order to build the tree structure. Your opinions on which way to go is very appreciated.

majetideepak:
@imjalpreet and @yingsu00 thanks for working on this and finding solutions to avoid manual overrides.
Among the two approaches, I prefer the latter `Only serialize leaf classes` since I see no use in maintaining the inheritance information in the protocol. 
In the future, we can add support for' super' when there is a use-case for the inheritance details.

> Both approachs need additional work.

Can you share what additional work is required for the second approach?

imjalpreet:
@majetideepak I have updated the PR to remove the usage of Base classes and excluded inheritance in the protocol. 

Based on these changes I have also updated the implementation in the follow-up PR: https://github.com/prestodb/presto/pull/21584

> Can you share what additional work is required for the second approach?
I think Ying was referring to the implementation changes for the current PRs. There are no changes required in protocol generation scripts like the inheritance approach.

Please have a look at the updated changes and let me know if these look good or in case there are any questions.

yingsu00:
@majetideepak Jalpreet has created a PR for each way:

1. Multiple inheritance: [[iceberg] [native] Introduce Iceberg Connector in Prestissimo (with multi-level inheritance)](https://github.com/prestodb/presto/pull/21905#top)
2. Only leaf classes: This PR.

majetideepak (approved):
@imjalpreet Thank you for extending the protocol without a lot of specialization.
Just to confirm, the `presto_protocol.h` and `presto_protocol.cpp` files have not been manually edited correct?

majetideepak (changes requested):
@imjalpreet please fix the commit title and description. Thanks!

mbasmanova (commented):
@imjalpreet This version is looking nice, clean and simple. Curious, what was the trick that allowed to remove the specializations.

imjalpreet:
> Thank you for extending the protocol without a lot of specialization. Can you confirm the `presto_protocol.h` and `presto_protocol.cpp` files have not been manually edited?

@majetideepak yes, the files have not been manually edited. 

> please fix the commit title and description. Thanks!

Can you please let me know what would you like to fix in the commit message?

imjalpreet:
> This version is looking nice, clean and simple. Curious, what was the trick that allowed to remove the specializations.

@mbasmanova There are some enums from the iceberg library that are required in the protocol. We have created wrappers for them on the java side so that they can be automated using the yaml file. Apart from that, earlier we were trying to replicate the multi-level inheritance present on the java side in prestissimo as well. But since currently we don't have a big usage of the inheritance details in Prestissimo, we have updated our implementation to remove their usage in Prestissimo. If in the future, we see that inheritance will be useful, we can have a look at https://github.com/prestodb/presto/pull/21905 for one of the ways we can update the protocol generation process to simplify the inclusion of multi-level inheritance.

mbasmanova:
@imjalpreet Got it. Thank you for explaining.

majetideepak:
@imjalpreet can you add the description to the commit body as well? You need to add an empty line after the title and add the body.
See example below.

```
commit 1bd7843cd321023d04c6cba734b136b6c552ea96
Author: Deepak Majeti <deepak.majeti@ibm.com>
Date:   Mon Feb 5 01:49:44 2024 -0500

    [native] Remove RE2 dependency from centos setup script
    
    RE2 is installed in the Velox centos setup script
```

majetideepak:
Just these 2 lines in the commit body.
```
Add Java classes from the Iceberg connector in presto_protocol.yml file.
Specialize IcebergColumnHandle to support 'operator<()'.
```

majetideepak (approved):
Thanks, @imjalpreet

presto-native-execution/presto_cpp/presto_protocol/presto_protocol.yml
```diff
@@ -177,6 +181,7 @@ JavaClasses:
   - presto-hive-metastore/src/main/java/com/facebook/presto/hive/BucketFunctionType.java
   - presto-hive-common/src/main/java/com/facebook/presto/hive/CacheQuotaRequirement.java
   - presto-hive-common/src/main/java/com/facebook/presto/hive/CacheQuotaScope.java
+  - presto-hive-common/src/main/java/com/facebook/presto/hive/BaseHiveColumnHandle.java
```
yingsu00, line 184: Is this still needed?

yingsu00 (commented):
@imjalpreet 
I see that the Json constructor for BaseHiveTableLayoutHandle is still there. Is it still needed? I don't see it's being used. Can you please add a separate commit or PR to remove it If it's no longer needed? 

BaseHiveColumnHandle  fields are still json annotated. Are the annotations required?

presto-native-execution/presto_cpp/presto_protocol/presto_protocol.yml
imjalpreet: Yes, this is still needed. ColumnType enum is part of this class and it is required to be included in the protocol.

presto-native-execution/presto_cpp/presto_protocol/special/HiveColumnHandle.hpp.inc
```diff
@@ -12,6 +12,9 @@
  * limitations under the License.
  */
 
+// HiveColumnHandle is special since we require an implementation of
```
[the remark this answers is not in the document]
imjalpreet: Sure, I will make this change.

imjalpreet:
> I see that the Json constructor for BaseHiveTableLayoutHandle is still there. Is it still needed? I don't see it's being used. Can you please add a separate commit or PR to remove it If it's no longer needed? 
> 
> 
> 
> BaseHiveColumnHandle  fields are still json annotated. Are the annotations required? 

@yingsu00 Since, this PR only includes protocol changes I haven't refactored any presto classes here.

I will have a look at the Base* classes in Presto, verify the need for JSON constructors and refactor them as needed in a separate PR.

presto-native-execution/presto_cpp/presto_protocol/special/HiveColumnHandle.hpp.inc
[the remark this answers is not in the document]
imjalpreet: I updated the comment

yingsu00 (approved):

PR #10278, opened 2018-03-28
Use explicit Optional in ConnectorMetadata

kokosing:

PR #21767, opened 2024-01-24
[For Test]Refactor ConnectorMetadata to avoid unnecessary multi-layouts returning

hantangwangd:
## Description

For Test

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```


tdcmeehan:
I think we need to come up with a deprecation strategy.

For now, I would recommend adding a new method `getTableLayoutForConstraint` (this name makes it clear what it's supposed to do vs. `getTableLayout` which does not do filter pushdown).  `getTableLayouts` can be marked as `@Deprecated`, and we can make its default implementation return a singleton list of `getTableLayoutForConstraint`.

In the release notes, we should mark that this method is being deprecated.  Then, we should create an issue for X months down the road to delete the method.

`getTableLayouts` should always be identical to a singleton list of the result of `getTableLayoutForConstraint`.  We can add a precondition check in `MetadataManager#getLayout` that proves this (probably where the current check to ensure size = 1 is).  Then we can remove this code once deprecation is complete.

hantangwangd:
Thanks for the instruction. Sounds great, I will do it.

hantangwangd:
Hi @tdcmeehan, following your guidance, the new method ``getTableLayoutForConstraint `` has been added and ``getTableLayouts `` has been marked as ``@Deprecated``, and the release note has been added. Consulting the PR #10278, perhaps for now it's better to make ``getTableLayoutForConstraint ``'s default implementation to validate and use ``getTableLayouts ``'s result. In this way, we can leave all connector implementations unchanged and let the engine's metadata manager to use the new SPI method. What's your opinion?

presto-main/src/main/java/com/facebook/presto/metadata/MetadataManager.java
```diff
@@ -364,14 +364,10 @@ public TableLayoutResult getLayout(Session session, TableHandle table, Constrain
         CatalogMetadata catalogMetadata = getCatalogMetadata(session, connectorId);
         ConnectorMetadata metadata = catalogMetadata.getMetadataFor(connectorId);
         ConnectorSession connectorSession = session.toConnectorSession(connectorId);
-        List<ConnectorTableLayoutResult> layouts = metadata.getTableLayouts(connectorSession, connectorTable, constraint, desiredColumns);
+        ConnectorTableLayoutResult layout = metadata.getTableLayoutForConstraint(connectorSession, connectorTable, constraint, desiredColumns);
```
tdcmeehan, line 367: I would try to ensure that the results between `getTableLayouts` and `getTableLayoutForConstaint` are consistent, otherwise it might be confusing to connector authors.

Remember that this method is also used by connectors not even known to Presto (i.e. external libraries).  So we need to accomodate people who are slow to adopt to the new API.

PR #21780, opened 2024-01-25
[native] Add http request active check when fetch data from output buffer manager

xiaoxmeng:
Add http request active check when fetch data from output buffer manager
in Velox. The active check is based on whether the http callstate has been destroyed
or the associated request has expired. This is to avoid arbitrary output buffer
to load data into a destination buffer which has set notify but the associated client
request has expired. This helps to accelerate the shuffle for query with scale writer
which uses arbitrary output buffer.

Unit test is added to verify this behavior.

### DIFF
--- a/presto-native-execution/presto_cpp/main/PrestoTask.h
+++ b/presto-native-execution/presto_cpp/main/PrestoTask.h
@@ -14,6 +14,7 @@
 #pragma once
 
 #include <memory>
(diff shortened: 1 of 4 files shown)


mbasmanova (approved):
Thanks. Please, extract "Advance Velox" changes into a separate commit before merging.

tanjialiang (approved):

Issue #21653, opened 2024-01-09
Inconsistent behavior between min/max(x, n) and min_by/max_by(x, y, n) in Window operation


kagamiori:
Hi community, I noticed an inconsistency between min/max(x, n) and min_by/max_by(x, y, n) when they are used in Window operation. As an example, for the query below with min_by(x, y, n), since the window frame is `UNBOUNDED PRECEDING` to `CURRENT ROW`, when the second row is processed, the function should aggregate both rows and hence there should be two values in the result array, i.e., [1, 2]. However, the current result of the second row is [2] only.

```
SELECT
    c0,
    c1,
    c2,
    c3,
    c4,
    MIN_BY(c0, c1, c2) OVER (
        PARTITION BY
            c3
        ORDER BY
            c4 ASC
        RANGE BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW
    )
FROM (
    VALUES
        (1, 1, 10, FALSE, 0),
        (2, 2, 10, FALSE, 1)
) AS t(c0, c1, c2, c3, c4)
```

On the other hand, for the same query on min(x, n), the result of the second row is [1, 2] as expected.
```
SELECT
    c0,
    c1,
    c2,
    c3,
    MIN(c0, c1) OVER (
        PARTITION BY
            c2
        ORDER BY
            c3 ASC
        RANGE BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW
    )
FROM (
    VALUES
        (1, 10, FALSE, 0),
        (2, 10, FALSE, 1)
) AS t(c0, c1, c2, c3)
```

The difference between the min_by(x, y, n) and min(x, n) is that when aggregation result is extracted from the accumulator, min_by() allows the accumulator (i.e., `heap`) to be cleared: https://github.com/prestodb/presto/blob/cb582bce0fd18f51d8862a8b2f53a134780f41aa/presto-main/src/main/java/com/facebook/presto/operator/aggregation/minmaxby/AbstractMinMaxByNAggregationFunction.java#L147-L153
, while min() retains the content in the accumulator: https://github.com/prestodb/presto/blob/bed052afeef85c728d0c99237be1047f60e24839/presto-main/src/main/java/com/facebook/presto/operator/aggregation/AbstractMinMaxNAggregationFunction.java#L173-L181

I wonder whether this difference is expected or their behavior should be made consistent?

## Your Environment

* Presto version used: 0.286

kagamiori:
cc @mbasmanova

[...]

PR #18615, opened 2022-11-03
Fix MIN(x, n) and MAX(x, n) returning NULL on window aggregations

sviscaino:
These functions have an odd behaviour when used on window aggregations. They return NULL for all rows except the first one in the window, because the heap is being cleared by popAll [here](https://github.com/prestodb/presto/blob/b6c7a1e46992e3c703213818eb5256bce9d3498f/presto-main/src/main/java/com/facebook/presto/operator/aggregation/AbstractMinMaxNAggregationFunction.java#L167), so calling output() a second time will return NULL. We add a call to recover the state of the heap after building the output array.

Test plan -
Before
Screenshot 2022-11-03 at 12 09 02

After
Screenshot 2022-11-03 at 12 09 08

Unit test: `mvn test -Dtest=TestMinMaxNFunction` + non regression `mvn test -Dtest=TestMinMaxByNAggregation`

```
== RELEASE NOTES ==

General Changes
* Fix max(x, n) and min(x, n) returning NULL on window aggregations
```

PR #21793, opened 2024-01-26
Fix min_by/max_by for window function evaluation

feilong-liu:
## Description
Fix issue https://github.com/prestodb/presto/issues/21653

## Motivation and Context
As described in https://github.com/prestodb/presto/issues/21653, the min_by/max_by are not returning the expected result, due to the optimization in window function expression. A similar bug in min/max function was fixed in https://github.com/prestodb/presto/pull/18615

## Impact
Fix the correctness issue for max_by/min_by in window function. 
The queries with the following pattern will be impacted:
- The max_by/min_by function should take three arguments, i.e. have the n argument like max_by(x, y, n). This is because the max_by(x, y) function has a [different implementation](https://github.com/prestodb/presto/blob/master/presto-main/src/main/java/com/facebook/presto/operator/aggregation/minmaxby/AbstractMinMaxBy.java) than the max_by(x, y, n) function
- The max_by/min_by function should not have “unbounded following” in the frame definition. This is because the optimization in the window function evaluation only works when the end of frame changes for different rows in the same partition.

Notice that this change will also slightly degrade the performance of this aggregation function. The fix is to restore the accumulation state after outputting the result, which is not needed if not used in window function. However, to mitigate this, we need to know whether it's used in window function in the implementation, which makes the code complex and not worth the benefit.

## Test Plan
Tested with the query in the issue.

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Fix a bug for min_by/max_by for window function, where results are incorrect when the function specifies number of elements to keep and the window does not have “unbounded following” in the frame.
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/operator/aggregation/minmaxby/AbstractMinMaxByNAggregationFunction.java
+++ b/presto-main/src/main/java/com/facebook/presto/operator/aggregation/minmaxby/AbstractMinMaxByNAggregationFunction.java
@@ -140,16 +140,17 @@ public abstract class AbstractMinMaxByNAggregationFunction
         }
 
         Type elementType = outputType.getElementType();
+        Type keyType = heap.getKeyType();
 
         BlockBuilder arrayBlockBuilder = out.beginBlockEntry();
-        BlockBuilder reversedBlockBuilder = elementType.createBlockBuilder(null, heap.getCapacity());
-        long startSize = heap.getEstimatedSize();
-        heap.popAll(reversedBlockBuilder);
-        state.addMemoryUsage(heap.getEstimatedSize() - startSize);
+        BlockBuilder reversedValueBlockBuilder = elementType.createBlockBuilder(null, heap.getCapacity());
+        BlockBuilder reversedKeyBlockBuilder = keyType.createBlockBuilder(null, heap.getCapacity());
+        heap.popAll(reversedValueBlockBuilder, reversedKeyBlockBuilder);
 
-        for (int i = reversedBlockBuilder.getPositionCount() - 1; i >= 0; i--) {
-            elementType.appendTo(reversedBlockBuilder, i, arrayBlockBuilder);
+        for (int i = reversedValueBlockBuilder.getPositionCount() - 1; i >= 0; i--) {
+            elementType.appendTo(reversedValueBlockBuilder, i, arrayBlockBuilder);
         }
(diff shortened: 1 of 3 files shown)


presto-main/src/main/java/com/facebook/presto/operator/aggregation/minmaxby/AbstractMinMaxByNAggregationFunction.java
```diff
@@ -140,16 +140,17 @@ public static void output(ArrayType outputType, MinMaxByNState state, BlockBuild
         }
 
         Type elementType = outputType.getElementType();
+        Type keyType = heap.getKeyType();
 
         BlockBuilder arrayBlockBuilder = out.beginBlockEntry();
-        BlockBuilder reversedBlockBuilder = elementType.createBlockBuilder(null, heap.getCapacity());
-        long startSize = heap.getEstimatedSize();
-        heap.popAll(reversedBlockBuilder);
-        state.addMemoryUsage(heap.getEstimatedSize() - startSize);
```
feilong-liu, line 148: The heap was restored after the change, hence do not need to update the memory usage.

presto-main/src/main/java/com/facebook/presto/operator/aggregation/minmaxby/AbstractMinMaxByNAggregationFunction.java
```diff
@@ -140,16 +140,17 @@ public static void output(ArrayType outputType, MinMaxByNState state, BlockBuild
...
+        BlockBuilder reversedValueBlockBuilder = elementType.createBlockBuilder(null, heap.getCapacity());
+        BlockBuilder reversedKeyBlockBuilder = keyType.createBlockBuilder(null, heap.getCapacity());
+        heap.popAll(reversedValueBlockBuilder, reversedKeyBlockBuilder);
 
-        for (int i = reversedBlockBuilder.getPositionCount() - 1; i >= 0; i--) {
-            elementType.appendTo(reversedBlockBuilder, i, arrayBlockBuilder);
+        for (int i = reversedValueBlockBuilder.getPositionCount() - 1; i >= 0; i--) {
+            elementType.appendTo(reversedValueBlockBuilder, i, arrayBlockBuilder);
         }
+        heap.addAll(reversedKeyBlockBuilder, reversedValueBlockBuilder);
```
feilong-liu, line 153: Restore states

pranjalssh (approved):
LGTM

PR #21794, opened 2024-01-26
[native] Add memory reclaim stats to task runtime stats

xiaoxmeng:
Add task runtime stats to collect memory reclaim stats during a task execution.
This helps to debug how much time spent on task memory reclaim of a slow query
execution. If the memory reclaim is triggered by the task itself, then the time will
also be counted in task cpu execution time. If not, it is counted in the task's scheduled
time but not execution time as the task will be stopped before memory reclamation.

### DIFF
--- a/presto-native-execution/presto_cpp/main/PrestoTask.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoTask.cpp
@@ -440,9 +440,22 @@ protocol::TaskInfo PrestoTask::updateInfoLocked() {
 
   std::unordered_map<std::string, RuntimeMetric> taskRuntimeStats;
 
+  if (taskStats.memoryReclaimCount > 0) {
+    taskRuntimeStats["memoryReclaimCount"].addValue(
+        taskStats.memoryReclaimCount);
+    taskRuntimeStats.insert(
+        {"memoryReclaimWallNanos",
+         RuntimeMetric(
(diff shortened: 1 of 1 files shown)


presto-native-execution/presto_cpp/main/PrestoTask.cpp
```diff
@@ -440,6 +440,13 @@ protocol::TaskInfo PrestoTask::updateInfoLocked() {
 
   std::unordered_map<std::string, RuntimeMetric> taskRuntimeStats;
 
+  if (taskStats.memoryReclaimCount > 0) {
```
mbasmanova, line 443: It would be nice to report these stats live while query is running. 

There is code later that calls processTaskStats which clears taskRuntimeStats if query is not running.

Also, I noticed that Coordinator UI shows stats based on unit. Hence, you need to specify kNanos unit to ensure nice display.

For example,

```
  taskRuntimeStats.insert(
      {"nativeProcessCpuTime",
       RuntimeMetric(processCpuTime_, RuntimeCounter::Unit::kNanos)});
```

xiaoxmeng: You mean to call processTaskStats with tryToSkipIfRunning to false?

tanjialiang (approved):

mbasmanova (approved):
Thanks.

PR #21795, opened 2024-01-26
[native] Fix presto_native to also build with fmt 9+.

kgpai:
## Description
Add support in presto_native so that it can build with fmt 9+ and also work with fmt 8. 

## Motivation and Context
This change should allow you to build Prestissimo with versions of fmt above 8.  Velox updated its dependency to fmt 10. Users who update to fmt 10 will not be able to build presto_native. This change fixes presto_native so that it can build with fmt 8 and above.

## Impact
N/A

## Test Plan
Tested against fmt 10 and should be backwards compatible with fmt 8. 

```
== NO RELEASE NOTE ==
```


kgpai:
Looks like the CI image will also have to be updated to fmt 10 : https://app.circleci.com/pipelines/github/prestodb/presto/11848/workflows/c3262685-7d90-475b-aa73-be2527b83df1/jobs/45096

PR #21800, opened 2024-01-27
Make runtime stats available for DirectoryLister

abhiseksaikia:
## Description
We want to use runtimestats in DirectoryLister implementation to measure different runtime metrics

## Motivation and Context
Goal is to add a metric to measure duration it takes to initialize iterator of HiveFileInfo for one of the implementation of DirectoryLister.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-hive/src/test/java/com/facebook/presto/hive/AbstractTestHiveClient.java
+++ b/presto-hive/src/test/java/com/facebook/presto/hive/AbstractTestHiveClient.java
@@ -18,6 +18,7 @@ import com.facebook.airlift.stats.CounterStat;
 import com.facebook.presto.GroupByHashPageIndexerFactory;
 import com.facebook.presto.cache.CacheConfig;
 import com.facebook.presto.common.Page;
+import com.facebook.presto.common.RuntimeStats;
 import com.facebook.presto.common.Subfield;
 import com.facebook.presto.common.block.Block;
 import com.facebook.presto.common.function.SqlFunctionProperties;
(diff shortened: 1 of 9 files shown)


presto-spi/src/main/java/com/facebook/presto/spi/ConnectorSession.java
```diff
@@ -63,4 +64,6 @@ default boolean isReadConstraints()
     }
 
     WarningCollector getWarningCollector();
+
+    Optional<RuntimeStats> getRuntimeStats();
```
NikhilCollooru, line 68: Any reason this is made Optional ? The `Session` has this as **not** optional. I think its better if we follow the same types across the different session objects. wdyt ?

abhiseksaikia: Makes sense to change this to non-Optional given the implementation of session and RuntimeStats are on oss side.

presto-hive/src/test/java/com/facebook/presto/hive/AbstractTestHiveClient.java
```diff
@@ -1204,6 +1205,12 @@ public WarningCollector getWarningCollector()
             {
                 return WarningCollector.NOOP;
             }
+
+            @Override
+            public RuntimeStats getRuntimeStats()
+            {
+                return new RuntimeStats();
```
NikhilCollooru, line 1212: I think this should be `session.getRuntimeStats()`

NikhilCollooru (approved):

NikhilCollooru (approved):

PR #21790, opened 2024-01-25
[WIP] [native pos] Bind native execution process to specific interface

arhimondr:
To reduce likelihood of running into port allocation conflicts

## Contributor checklist

- [ ] Please make sure your submission complies with our [development](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#development), [formatting](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#formatting), [commit message](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#commit-formatting-and-pull-requests), and [attribution guidelines](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#attribution).
- [ ] PR description addresses the issue accurately and concisely.  If the change is non-trivial, a GitHub Issue is referenced.
- [ ] Documented new properties (with its default value), SQL syntax, functions, or other functionality.
- [ ] If release notes are required, they follow the [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines).
- [ ] Adequate tests were added if applicable.
- [ ] CI passed.

## Release Notes
Please follow [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines) and fill in the release notes below.

```
== RELEASE NOTES ==

General Changes
* ...
* ...

Hive Changes
* ...
* ...
```

If release note is NOT required, use:

```
== NO RELEASE NOTE ==
```

PR #21801, opened 2024-01-28
[native] Fix https socket binding.

amitkdutta:
When https is enabled, socket address is mistakenly binded with http in https://github.com/prestodb/presto/pull/21790/files#diff-87eb79cbdaa1e2812753e39ee76d4d29edf60b1d595a094ccc2423c02b5555f4R284. Here we bind it properly and also register the newly added config properfy.

### DIFF
--- a/presto-native-execution/presto_cpp/main/common/Configs.cpp
+++ b/presto-native-execution/presto_cpp/main/common/Configs.cpp
@@ -139,6 +139,7 @@ SystemConfig::SystemConfig() {
           NONE_PROP(kPrestoVersion),
           NONE_PROP(kHttpServerHttpPort),
           STR_PROP(kHttpServerReusePort, "false"),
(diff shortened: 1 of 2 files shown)


gggrace14 (approved):
I see

presto-native-execution/presto_cpp/main/common/Configs.cpp
```diff
@@ -139,6 +139,7 @@ SystemConfig::SystemConfig() {
           NONE_PROP(kPrestoVersion),
           NONE_PROP(kHttpServerHttpPort),
           STR_PROP(kHttpServerReusePort, "false"),
+          BOOL_PROP(kHttpServerBindToNodeInternalAddressOnlyEnabled, "false"),
```
arhimondr, line 142: @amitkdutta What is this map? Should we add a test to make sure this map is populated?

PR #20625, opened 2023-08-22
[native] Add e2e tests for aggregate window functions

pramodsatya:
Test plan - Adds e2e tests for window aggregate functions sum, avg, count, min, max.
```
== NO RELEASE NOTE ==
```

PR #21825, opened 2024-01-31
[Native] Fix error `java.lang.OutOfMemoryError: Compressed class space`

pramodsatya:
## Description

The following error was seen when adding presto native e2e tests for window aggregate functions in  https://github.com/prestodb/presto/pull/20625 :
```
[ERROR] java.lang.OutOfMemoryError: Compressed class space
[ERROR] Terminating due to java.lang.OutOfMemoryError: Compressed class space
```

This error is fixed by increasing the compressed class space size to 3gb in the CircleCI config for `linux-presto-e2e-tests`.

## Test Plan

The CircleCI job `linux-presto-e2e-tests` was verified to be passing with this fix: https://app.circleci.com/pipelines/github/prestodb/presto/11583/workflows/b23f1755-3811-49fd-844e-748dc2212080/jobs/43847 .

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -64,8 +64,8 @@ executors:
       - image: public.ecr.aws/oss-presto/prestocpp:root-20240129
     resource_class: 2xlarge
     environment:
-      MAVEN_OPTS: "-Xmx4G -XX:+ExitOnOutOfMemoryError"
-      MAVEN_INSTALL_OPTS: "-Xmx2G -XX:+ExitOnOutOfMemoryError"
+      MAVEN_OPTS: "-Xmx4G -XX:+ExitOnOutOfMemoryError -XX:CompressedClassSpaceSize=3g"
+      MAVEN_INSTALL_OPTS: "-Xmx2G -XX:+ExitOnOutOfMemoryError -XX:CompressedClassSpaceSize=3g"
       MAVEN_FAST_INSTALL: "-B -V --quiet -T C1 -DskipTests -Dair.check.skip-all -Dmaven.javadoc.skip=true"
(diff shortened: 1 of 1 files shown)


majetideepak:
@amitkdutta can you take a look at this change? Looks reasonable to me, but I want to double-check.

mshang816 (approved):

tdcmeehan (approved):

.circleci/continue_config.yml
```diff
@@ -57,8 +57,8 @@ executors:
       - image: prestocpp/prestocpp-avx-centos:root-20230613
     resource_class: 2xlarge
     environment:
-      MAVEN_OPTS: "-Xmx4G -XX:+ExitOnOutOfMemoryError"
-      MAVEN_INSTALL_OPTS: "-Xmx2G -XX:+ExitOnOutOfMemoryError"
+      MAVEN_OPTS: "-Xmx4G -XX:+ExitOnOutOfMemoryError -XX:CompressedClassSpaceSize=3g"
```
imjalpreet, line 60: Small recommendation: I see that we are using machine size as 2xlarge (which is 32 GB memory) we can either increase Xmx for both `MAVEN_OPTS` and `MAVEN_FINAL_OPTS` or if we don't need such large machines we can plan to shift to smaller resource class probably to save costs. 

If there are certain suites that require larger machine size, we can also segregate different suites to use different executors to optimize costs.

pramodsatya: Thanks for the suggestion @imjalpreet. @majetideepak could you please confirm if I can increase Xmx to 8G in `MAVEN_OPTS` and `MAVEN_FINAL_OPTS` in this PR?

majetideepak: We have to remember that Prestissimo workers also run on the same node and they require memory. I would prefer to limit the JVM memory to only the required since we know the memory is held up by the JVM until GC runs which is not in our control.

PR #21839, opened 2024-02-01
[native] Export averageOutputBufferWallNanos task-level stat

mbasmanova:
This statistic tracks average time a byte of data is waiting in output buffers to be fetched. It helps debug slow exchange.

This was added to Velox in https://github.com/facebookincubator/velox/pull/8534

```
== NO RELEASE NOTE ==
```


spershin (approved):

kewang1024 (approved):
Thanks for the additional helpful stats, curious, why do we report Nanos instead of just Ms second here, we anyway only have precision for Ms in velox

mbasmanova:
> Thanks for the additional helpful stats, curious, why do we report Nanos instead of just Ms second here, we anyway only have precision for Ms in velox

Simply because ms is not supported:

```
struct RuntimeCounter {
  enum class Unit { kNone, kNanos, kBytes };
```

PR #21858, opened 2024-02-05
[WIP] [PR-FOR-TESTS] Single Tenant execution POC

shrinidhijoshi:
## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* ...
* ...

Hive Changes
* ...
* ...
```

```
== NO RELEASE NOTE ==
```


presto-main/src/main/java/com/facebook/presto/execution/scheduler/FixedSubsetNodeSetSupplier.java
```diff
@@ -0,0 +1,296 @@
...
+        List<InternalNode> allocatedNodes = perQueryNodeAssignments.values().stream()
+                .flatMap(Collection::stream)
+                .collect(Collectors.toList());
+
+        List<InternalNode> currentActiveWorkerNodesInCluster = nodeManager.getNodes(ACTIVE).stream()
+                .filter(node -> !node.isResourceManager() && !node.isCoordinator() && !node.isCatalogServer())
+                .collect(toImmutableList());
+
+        List<InternalNode> avaialableNodes = currentActiveWorkerNodesInCluster.stream()
+                .filter(node -> !allocatedNodes.contains(node))
```
arhimondr, line 105: I wonder how likely these lookups to impact test results? With thousands of nodes in a cluster that may be taking a while. Might be safer to do a `Set`.

presto-main/src/main/java/com/facebook/presto/execution/scheduler/FixedSubsetNodeSetSupplier.java
```diff
@@ -0,0 +1,296 @@
...
+            InternalNodeManager nodeManager,
+            NetworkTopology networkTopology,
+            @ForNodeScheduler ScheduledExecutorService scheduledExecutorService)
+    {
+        this.nodeManager = nodeManager;
+        this.networkLocationCache = new NetworkLocationCache(networkTopology);
+        this.perQueryNodeAssignments = new HashMap<>();
+        this.pendingRequests = new PriorityQueue<>();
+        // start loop to check pending node requests every second
+        scheduledFuture = scheduledExecutorService.scheduleAtFixedRate(this::fulfillPendingRequests, 0, 100, TimeUnit.MILLISECONDS);
```
arhimondr, line 86: nit: To reduce latency we may want to trigger `fulfillPendingRequests` each time `acquire` / `release` is called.

shrinidhijoshi: Makes sense. Thats what I originally had in mind but took a shortcut for POC. But you are right we probably need to fix this for POC as well. Will update now

presto-main/src/main/java/com/facebook/presto/execution/scheduler/FixedSubsetNodeSetSupplier.java
shrinidhijoshi: Good point. Will update

presto-main/src/main/java/com/facebook/presto/execution/scheduler/NodeScheduler.java
```diff
@@ -92,12 +89,13 @@ public class NodeScheduler
     private final long maxPendingSplitsWeightPerTask;
     private final NodeTaskMap nodeTaskMap;
     private final boolean useNetworkTopology;
-    private final Duration nodeMapRefreshInterval;
+    private final Duration nodeSetRefreshInterval;
```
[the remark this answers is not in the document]
shrinidhijoshi: Sure. Do not have a strong opinion on this. Can retain it

presto-main/src/main/java/com/facebook/presto/dispatcher/LocalDispatchQuery.java
```diff
@@ -191,15 +197,29 @@ public void startWaitingForResources()
 
     private void waitForMinimumWorkers()
     {
-        ListenableFuture<?> minimumWorkerFuture = clusterSizeMonitor.waitForMinimumWorkers();
+        int nodesToAcquire = getHashPartitionCount(getSession());
+        CompletableFuture<?> waitForNodesFuture =
```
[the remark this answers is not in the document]
shrinidhijoshi: Curious why

presto-main/src/main/java/com/facebook/presto/dispatcher/LocalDispatchQuery.java
```diff
@@ -191,15 +197,29 @@ public void startWaitingForResources()
...
+                nodeScheduler.acquireNodes(stateMachine.getQueryId(), nodesToAcquire);
+
+        // Listener to release nodes after query execution completes
+        stateMachine.addStateChangeListener(newState -> {
```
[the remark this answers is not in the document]
shrinidhijoshi: Yes exactly right. We could also be more intelligent/dynamic about node acquisition depending on what we found during planning. Will see if I can get this feature in the prototype itself over the next couple days

presto-main/src/main/java/com/facebook/presto/execution/scheduler/FixedSubsetNodeSetSupplier.java
```diff
@@ -0,0 +1,340 @@
...
+                List<InternalNode> nodesForQuery = new ArrayList<>();
+                for (int i = 0; i < nodeSetAcquireRequest.getCount(); i++) {
+                    nodesForQuery.add(freeNodes.remove(i));
+                }
+                // assign the nodes to this query
+                queryNodesMap.put(nodeSetAcquireRequest.getQueryId(), nodesForQuery);
+                // register the state supplier (used for bookeeping/cleanup)
+                queryStateMap.put(nodeSetAcquireRequest.getQueryId(), nodeSetAcquireRequest.getQueryStateSupplier());
+                // Let the query scheduler know that query can now execute
+                nodeSetAcquireRequest.getCompletableFuture().complete(null);
```
agrawaldevesh, line 121: So earlier waitForMinWorkers was almost a no-op and perhaps only ever triggered when the cluster is starting up. But now it can happen because of too much query concurrency. This can creep in scheduling delays. Should we log/measure it explicitly and perhaps better yet for the POC just fail fast and ensure that we always have nodes available for the query ? (Because we can always size the query / RG concurrency)

presto-main/src/main/java/com/facebook/presto/execution/scheduler/FixedSubsetNodeSetSupplier.java
```diff
@@ -0,0 +1,340 @@
...
+        }
+
+        @Override
+        public int compareTo(@NotNull Object other)
+        {
+            NodeSetAcquireRequest otherRequest = (NodeSetAcquireRequest) other;
+            if (otherRequest.getCount() == this.getCount()) {
+                return 0;
+            }
+            return this.getCount() < otherRequest.getCount() ? -1 : 1;
```
agrawaldevesh, line 331: QueryIds contain timestamps (or maybe we should explicitly include create timestamps here) -- to break the ties when the hash-partition count is the same (which will probably happen all of the time).

presto-main/src/main/java/com/facebook/presto/execution/scheduler/FixedSubsetNodeSetSupplier.java
```diff
@@ -0,0 +1,340 @@
...
+import static java.lang.String.format;
+
+/**
+ * This WIP class is responsible for assigning a subset of the cluster nodes
+ * to a requesting query.
+ * The size of the subset is currently static, but it can by defined by configuration
+ * or can be driven by fairness algorithm depending on query metadata.
+ */
+public class FixedSubsetNodeSetSupplier
+        implements NodeSetSupplier, Closeable
```
agrawaldevesh, line 65: Maybe its too soon: But should we share some code with the DefaultNodeSupplier ? 

I am also very unclear in the difference b/w a NodeSet and the older NodeMap, can you clarify that in comments somewhere ?

presto-main/src/main/java/com/facebook/presto/execution/scheduler/FixedSubsetNodeSetSupplier.java
```diff
@@ -0,0 +1,340 @@
...
+                // allNodes = nodeManager.getAllConnectorNodes(connectorId).stream().filter(finalNodeFilterPredicate).collect(toImmutableList());
+            }
+            else {
+                activeNodes = nodeManager.getNodes(ACTIVE).stream().filter(finalNodeFilterPredicate).collect(toImmutableList());
+                // allNodes = activeNodes;
+            }
+
+            // Here we want to assign a subset of the active nodes to this query
+            if (queryNodesMap.containsKey(queryId)) {
+                activeNodes = activeNodes.stream().filter(internalNode -> queryNodesMap.get(queryId)
```
agrawaldevesh, line 196: This is a bit of a weird data structure wise: like, why is this not just ImmutableList.copyOf(queryNodesMap.get(queryId)) ?

presto-main/src/main/java/com/facebook/presto/execution/scheduler/FixedSubsetNodeSetSupplier.java
```diff
@@ -0,0 +1,340 @@
...
+                // assign the nodes to this query
+                queryNodesMap.put(nodeSetAcquireRequest.getQueryId(), nodesForQuery);
+                // register the state supplier (used for bookeeping/cleanup)
+                queryStateMap.put(nodeSetAcquireRequest.getQueryId(), nodeSetAcquireRequest.getQueryStateSupplier());
+                // Let the query scheduler know that query can now execute
+                nodeSetAcquireRequest.getCompletableFuture().complete(null);
+                log.info("Successfully fulfilled nodeAcquireRequest=%s", nodeSetAcquireRequest);
+
+                // Move to the next request
+                pendingRequests.poll();
```
agrawaldevesh, line 125: idk, should we move this poll request to the beginning of this try block ?

presto-main/src/main/java/com/facebook/presto/execution/scheduler/nodeSelection/SimpleNodeSelector.java
```diff
@@ -274,7 +274,7 @@ protected Optional<InternalNodeInfo> chooseLeastBusyNode(SplitWeight splitWeight
         return Optional.of(new InternalNodeInfo(chosenNode, false));
     }
 
-    private List<InternalNode> getEligibleNodes(int limit, NodeMap nodeMap, List<RemoteTask> existingTasks)
+    private List<InternalNode> getEligibleNodes(int limit, NodeSet nodeMap, List<RemoteTask> existingTasks)
```
agrawaldevesh, line 277: So where are we on the debate of NodeMap -> NodeSet ? Do we need to make that change ?

presto-main/src/main/java/com/facebook/presto/dispatcher/LocalDispatchQuery.java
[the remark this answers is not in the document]
agrawaldevesh: +1 on what Andrii states. Please keep this stuff consistent with the existing approach of ListenableFuture.

presto-main/src/main/java/com/facebook/presto/execution/scheduler/FixedSubsetNodeSetSupplier.java
```diff
@@ -0,0 +1,340 @@
...
+
+        // In case we have not released nodes on finished queries.
+        // This is to prevent leaking nodes over time as falsely occupied
+        for (Map.Entry<QueryId, List<InternalNode>> entry : queryNodesMap.entrySet()) {
+            if (queryStateMap.containsKey(entry.getKey()) && queryStateMap.get(entry.getKey()).get().isDone()) {
+                releaseNodes(entry.getKey(), entry.getValue().size());
+            }
+        }
+
+        List<InternalNode> freeNodes = computeFreeNodesInCluster();
```
agrawaldevesh, line 107: Should this min workers ? Like skip fullfilling if the num free nodes is less than what waitForMinWorkers originally did -- to wait for clusters to "turn up"

presto-main/src/main/java/com/facebook/presto/execution/scheduler/FixedSubsetNodeSetSupplier.java
```diff
@@ -0,0 +1,340 @@
...
+
+        NodeSetAcquireRequest nodeSetAcquireRequest = pendingRequests.peek();
+        while (nodeSetAcquireRequest != null && nodeSetAcquireRequest.getCount() <= freeNodes.size()) {
```
agrawaldevesh, line 110: I am not sure why we are looping here, instead of putting in a loop around L107 ? Consider what happens if the cluster membership changes while you are looping below.

presto-main/src/main/java/com/facebook/presto/dispatcher/LocalDispatchQuery.java
```diff
@@ -191,17 +201,47 @@ public void startWaitingForResources()
 
     private void waitForMinimumWorkers()
     {
-        ListenableFuture<?> minimumWorkerFuture = clusterSizeMonitor.waitForMinimumWorkers();
+        int nodesToAcquire = computeNodesToAcquire();
+        CompletableFuture<?> waitForNodesFuture =
+                nodeScheduler.acquireNodes(stateMachine.getQueryId(), nodesToAcquire, stateMachine::getQueryState);
```
agrawaldevesh, line 206: How do you handle cluster turnup now ? We are no longer waiting for the waitForMinWorkers ? See comment/suggestion on handling that in the fullfillPendingRequest

presto-main/src/main/java/com/facebook/presto/execution/scheduler/nodeSelection/TopologyAwareNodeSelector.java
```diff
@@ -69,7 +69,7 @@ public class TopologyAwareNodeSelector
     private final NodeSelectionStats nodeSelectionStats;
     private final NodeTaskMap nodeTaskMap;
     private final boolean includeCoordinator;
-    private final AtomicReference<Supplier<NodeMap>> nodeMap;
+    private final AtomicReference<Supplier<NodeSet>> nodeMap;
```
agrawaldevesh, line 72: Idk, I am not sure why we are going with this change from NodeMap to NodeSet, it is increasing the diff size.

agrawaldevesh (commented):
I am not sure but I am wondering if we considered some simpler approaches than the full machinery of building such a "node allocator" (FixedSubsetNodeSupplier) ?

For example, bin nodes in K groups and then modify the "node selectors" (like SimpleNodeSelector::getEligibleNodes, or selectRandomNodes for non-leaves) to choose one of the groups for the query say by hashing ?

presto-main/src/main/java/com/facebook/presto/execution/scheduler/FixedSubsetNodeSetSupplier.java
shrinidhijoshi: Yup. Current code is to quicly test the feasibility and wins. Will def look into code restructuring and cleanup. 

> NodeSet and the older NodeMap,

No difference. Will undo this renaming and retain NodeMap

presto-main/src/main/java/com/facebook/presto/execution/scheduler/FixedSubsetNodeSetSupplier.java
shrinidhijoshi: Yes. I presumed we don't need that for POC

presto-main/src/main/java/com/facebook/presto/execution/scheduler/FixedSubsetNodeSetSupplier.java
shrinidhijoshi: Good point. We discussed this and concluded that for POC result measurements, we can get this 'extra' from the QueryStats.resourceWaitingTime. So decided to leave this logic as is

presto-main/src/main/java/com/facebook/presto/execution/scheduler/FixedSubsetNodeSetSupplier.java
```diff
@@ -0,0 +1,340 @@
...
+
+    private void fulfillPendingRequests()
+    {
+        if (pendingRequests.size() == 0) {
+            return;
+        }
+
+        // In case we have not released nodes on finished queries.
+        // This is to prevent leaking nodes over time as falsely occupied
+        for (Map.Entry<QueryId, List<InternalNode>> entry : queryNodesMap.entrySet()) {
```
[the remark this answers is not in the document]
shrinidhijoshi: Yes this needs to be investigated. The reason was not obvious, so decided to add a catch-all would be quick way to get unblocked for POC.

presto-main/src/main/java/com/facebook/presto/dispatcher/LocalDispatchQuery.java
```diff
@@ -191,17 +201,47 @@ public void startWaitingForResources()
...
+    {
+        if (ALL.equals(getExchangeMaterializationStrategy(getSession()))) {
+            // TODO (shrinidhijoshi): Fix node count logic for LBM
+            // For LBM, the query.hash-partition-count represents buckets and not
+            // number of nodes. So for experiment we cap it at 600
+            return Math.min(600, getHashPartitionCount(getSession()));
+        }
+        else {
+            int hashPartitionCount = getHashPartitionCount(getSession());
+            if (hashPartitionCount > 4900) {
```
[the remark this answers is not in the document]
shrinidhijoshi: Yes. Of course. But this code is for a quick POC on a 5k node cluster.

presto-main/src/main/java/com/facebook/presto/dispatcher/LocalDispatchQuery.java
shrinidhijoshi: Good point. Will definitely look into this past the POC phase. The current POC is a very controlled environment, so didn't fix this.

presto-main/src/main/java/com/facebook/presto/execution/scheduler/FixedSubsetNodeSetSupplier.java
shrinidhijoshi: Yes this is a good edge case. The reason it does not matter for POC is that the loop is very quick. But yes in the prod version, we would like to re-compute free nodes after every query assignment

presto-main/src/main/java/com/facebook/presto/execution/scheduler/FixedSubsetNodeSetSupplier.java
shrinidhijoshi: Yes. Responded to this in cluster turn up use-case

PR #21866, opened 2024-02-05
Revert "Preserve case for RowType's field name and JSON content when ``CAST``"

ajaygeorge:
This reverts commit d5a31b3ebb22254f000dca40e420a857bfdc5e15.

## Description
This reverts the commit d5a31b3ebb22254f000dca40e420a857bfdc5e15.
which was causing issues with user queries.
See discussion in https://github.com/prestodb/presto/pull/21602#issuecomment-1924690473

## Test Plan
Unit Tests

## Contributor checklist

- [x] Please make sure your submission complies with our [development](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#development), [formatting](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#formatting), [commit message](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#commit-formatting-and-pull-requests), and [attribution guidelines](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#attribution).
- [x] PR description addresses the issue accurately and concisely.  If the change is non-trivial, a GitHub Issue is referenced.
- [x] Documented new properties (with its default value), SQL syntax, functions, or other functionality.
- [x] If release notes are required, they follow the [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines).
- [x] Adequate tests were added if applicable.
- [ ] CI passed.

## Release Notes
Please follow [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines) and fill in the release notes below.

```
== NO RELEASE NOTE ==
```

Issue #12923, opened 2019-06-10
CAST(json TO ROW) should follow the delimited field name semantics


arhimondr:
Currently the `CAST(json to ROW)` always does the case insensitive JSON object field matching. (https://github.com/prestodb/presto/blob/master/presto-main/src/main/java/com/facebook/presto/util/JsonUtil.java#L1262)

However in JSON field names are case sensitive. 

Following the SQL semantics, CAST(json to ROW(field_name VARCHAR)) should match both, `field_name` and `FIELD_NAME`, as the non delimited field of a row is case insensitive.

However, `CAST(json to ROW("field_name" VARCHAR))` should perform the case sensitive match.

arhimondr:
https://github.com/prestodb/presto/blob/master/presto-spi/src/main/java/com/facebook/presto/spi/type/RowFieldName.java#L26

Yafeng-Cheng:
Any updates on this? Camel case is basically useless in such structure if not case sensitive.

PR #21869, opened 2024-02-06
Preserve case for RowType's field name and JSON content when ``CAST``

hantangwangd:
## Description

As discussed in https://github.com/prestodb/presto/pull/21866 and https://github.com/prestodb/presto/pull/21602, add a config property ``legacy_json_cast`` to control and support the legacy behavior which do not reserve the case of field name in json when casting to row type.

The deferences in behavior with and without the config property could be found in presto-docs/src/main/sphinx/functions/json.rst.

## Impact

Add a new config property ``legacy_json_cast`` whose default value is ``true`` to support legacy behavior. When set the property ``legacy_json_cast`` to ``false`` in both coordinators and workers' config file, there will be some changes in behavior of casting from a json to a row.

## Test Plan

 - Enhanced test case TestRowOperators.testJsonToRow(), adding case reservation verify for json cast to other type
 - Adding the legacy json cast compatible cases mentioned in https://github.com/prestodb/presto/pull/21602#issuecomment-1924690473
 - Enhanced test case TestRowOperators.testRowCast(), adding the example mentioned in issue #20701 
 - Make sure this fix do not affect existing test cases

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Add configuration property `legacy_json_cast` whose default value is `true`. See [Properties Reference](http://prestodb.io/docs/current/admin/properties.html#legacy-compatible-properties). 
```


hantangwangd:
Hi @mbasmanova, I have added some notes and examples to presto-docs/src/main/sphinx/functions/json.rst, trying to describe the behavior with and without the new config property. Please take a look, thanks.

presto-docs/src/main/sphinx/functions/json.rst
```diff
@@ -71,6 +71,28 @@ Cast from JSON
         SELECT CAST(JSON '{"k1": [1, 23], "k2": 456}' AS MAP(VARCHAR, JSON)); -- {k1 = JSON '[1,23]', k2 = JSON '456'}
         SELECT CAST(JSON '[null]' AS ARRAY(JSON)); -- [JSON 'null']
 
+.. note::
+
+    When casting from ``JSON`` to ``ROW``, by default the case of field names
```
mbasmanova, line 76: @hantangwangd Thank you for adding this doc. This is helpful.

presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
```diff
@@ -300,6 +300,7 @@ public final class SystemSessionProperties
     public static final String ADD_PARTIAL_NODE_FOR_ROW_NUMBER_WITH_LIMIT = "add_partial_node_for_row_number_with_limit";
     public static final String REWRITE_CASE_TO_MAP_ENABLED = "rewrite_case_to_map_enabled";
     public static final String FIELD_NAMES_IN_JSON_CAST_ENABLED = "field_names_in_json_cast_enabled";
+    public static final String LEGACY_JSON_CAST = "legacy_json_cast";
```
mbasmanova, line 303: CC: @pranjalssh @tdcmeehan @majetideepak @aditi-pandit @spershin 

We may need to add logic to check this session property and fail fast if native workers do not support it. Otherwise, we'll get wrong results when processing queries with Prestissimo.

spershin:
@hantangwangd 

What is the use case for this feature?
Why it needs to be added?

hantangwangd:
> I wonder if it is better to not allow changing this on a per-query basis and rather have this as a cluster-wide config.

You mean that it's better to let it be a config property which could only be set in coordinator and work's config.properties?

tdcmeehan:
@spershin please see attached documentation and #20701

hantangwangd:
Hi @tdcmeehan @mbasmanova, following your guidance, property ``legacy_json_cast `` has been limited to a config property, and could not be affected by session property setting. And the default value has been changed to ``false``, document and test cases have been changed correspondingly. Please take a look, thanks!

tdcmeehan:
> I agree that it is nice when default values include latest bug fixes and not legacy behavior. At the same time, it is important to avoid regressions to existing users as well.

Of course I agree with this.  That is the purpose of my suggestion--to announce breaking changes through release notes in order to avoid regressions to existing users.

> Wondering if we already have a section at the top of release notes that lists all the breaking changes and whether it is possible to easily access a combined list of breaking changes between any two releases.

I had this thought as well--we already have a "release highlights" section that we occasionally use to highlight important updates.  We could also add a "breaking changes" section to make it very clear that user action is required.  Given our 2-month release cadence, it should be manageable for operators to understand what is required of them in between releases.

> Perhaps, we could introduce a single property that says "Do not break me". When this property is set, the defaults for breaking configs are set to match legacy behavior. This would allow existing users to set a single property once and not worry about new release breaking them. At the same time new users will not have this config set and will be getting latest and greatest.

How would you propose we eventually remove the old code under this model?  It would seem that the "do not break me" flag would just collect the buggy features, and the people who depend on it would prevent the bugs from being deleted.  If we do a flag per feature, then at least we can go back to delete the feature at a later point in time (by explicitly turning on the "legacy" behavior, the operator has been put on notice that it will eventually be deprecated and removed). 

> This is something I'm not sure about and would like to get more clarity. In JSON, keys are case sensitive, right? But in SQL, column names are not? Or is it that column names are not case sensitive, but struct fields are? Or is it that both column names and struct fields are case insensitive unless they are in double quotes?

AKAIK, if it’s not quoted, it’s lower cased, and if it’s quoted, then case is preserved.  I think struct fields behave exactly as columns.  This has actually been discussed before in #12923.

Going back to Ajay’s test case, I think the following test cases make sense:

1) Delimiter unspecified
   `SELECT CAST(json_extract('{"1":[{"name": "John", "AGE": 30}]}', '$') AS MAP<bigint, ARRAY<ROW(name VARCHAR, age VARCHAR)>>) AS data;`
   Expected: `` {1=[{name=John, age=30}]}``, ``Actual:  {1=[{name=John, age=null}]}``

2) Delimiter specified with matching case
   `SELECT CAST(json_extract('{"1":[{"name": "John", "AGE": 30}]}', '$') AS MAP<bigint, ARRAY<ROW(name VARCHAR, "AGE" VARCHAR)>>) AS data;`
   Expected: `` {1=[{name=John, AGE=30}]}``, ``Actual:  {1=[{name=John, AGE=30}]}``

3) Delimiter specified without matching case
   `SELECT CAST(json_extract('{"1":[{"name": "John", "AGE": 30}]}', '$') AS MAP<bigint, ARRAY<ROW(name VARCHAR, "age" VARCHAR)>>) AS data;`
   Expected: `` {1=[{name=John, age=null}]} ``, ``Actual:  {1=[{name=John, age=null}]} ``

I think the problem with with test case 1: it seems that if one hasn’t specified quotedness on the ROW, then it should match the JSON key. 

With the current code, there is breaking behavior in 1 and 3 (before this change, 3 would also return age).  I think the breaking behavior should only be for 3—this is because a delimiter is specified on the age, and it’s not the case of the JSON (JSON is case sensitive).  It seems that 1 should not change, because we are casting the JSON as a struct, and the field name should be accessible in a case insensitive way given no delimiter was specified.

@hantangwong thoughts on this?

hantangwangd:
> I think the problem is with test case 1: it seems that if one hasn’t specified quotedness on the ROW, then it should match the JSON key.
> 
> With the current code, there is breaking behavior in 1 and 3 (before this change, 3 would also return age). I think the breaking behavior should only be for 3—this is because a delimiter is specified on the age, and it’s not the case of the JSON (JSON is case sensitive). It seems that 1 should not change, because we are casting the JSON as a struct, and the field name should be accessible in a case insensitive way given no delimiter was specified.

@tdcmeehan It makes sense to me. Seems it may be necessary to record the status of quotation marks during parsing. I will do some investigation to implement this.

hantangwangd:
Hi @tdcmeehan, if we consider unquoted field names as case-insensitive when matching, there are still some behaviors that need to be clarified, mainly involving field names that differ only in case. Take the following test cases for example:

1) Delimiter all specified without duplicate
   `SELECT CAST(json_extract('{"1":[{"key1": "John", "KEY1": "Johnny"}]}', '$') AS MAP<bigint, ARRAY<ROW("key1" VARCHAR, "KEY1" VARCHAR)>>) AS data;`
   Expected: `` {1=[{key1=John, KEY1=Johnny}]}``

2) Delimiter specified mixed with Delimiter unspecified without duplicate
   `SELECT CAST(json_extract('{"1":[{"key1": "John", "KEY1": "Johnny"}]}', '$') AS MAP<bigint, ARRAY<ROW(key1 VARCHAR, "KEY1" VARCHAR)>>) AS data;`
   Expected: `` {1=[{key1=John, KEY1=Johnny}]}``


3) Delimiter specified mixed with Delimiter unspecified with duplicate
   `SELECT CAST(json_extract('{"1":[{"key1": "John", "KEY1": "Johnny"}]}', '$') AS MAP<bigint, ARRAY<ROW(KEY1 VARCHAR, "key1" VARCHAR)>>) AS data;`
   Expected: throw an exception or get result `` {1=[{key1=John, KEY1=Johnny}]}`` ?

4) Delimiter unspecified with multiple matching key in json
   `SELECT CAST(json_extract('{"1":[{"key1": "John", "KEY1":"Johnny"}]}', '$') AS MAP<bigint, ARRAY<ROW(key1 VARCHAR)>>) AS data;`
   Expected: ?

Are the behaviors reasonable in case 1 and 2? And what should be the appropriate behavior in case 3 and 4? Or should we prohibit such usage? What's your opinion？

tdcmeehan:
My gut feeling is 3 should work (i.e. `{1=[{key1=John, KEY1=Johnny}]}`), because we can first select out delimiter-specified fields, and choose the remaining fields after that, and 4 should throw an ambiguous column exception as the alternative is arbitrarily picking one field.  But I think we should first study the SQL spec to see if this is specified.

steveburnett (approved):
LGTM! (docs)

New local build of docs, changes look good, no additional changes requested. Thanks!

mbasmanova:
> How would you propose we eventually remove the old code under this model? It would seem that the "do not break me" flag would just collect the buggy features, and the people who depend on it would prevent the bugs from being deleted. If we do a flag per feature, then at least we can go back to delete the feature at a later point in time (by explicitly turning on the "legacy" behavior, the operator has been put on notice that it will eventually be deprecated and removed).

This sounds good on paper, but doesn't work in practice. Unfortunately, there is no way to avoid "collecting the buggy features" when it comes to user-visible behavior. Notice that deprecated.legacy-unnest-array-rows, deprecated.legacy-array-agg and a bunch of other similar flags have been 'deprecated' for many years and there is practically zero chance of them being removed.

That said, I don't think strongly about introducing "Do not break me" flag. Just wanted to share a thought that crossed my mind.

> if it’s not quoted, it’s lower cased, and if it’s quoted, then case is preserved. I think struct fields behave exactly as columns. This has actually been discussed before in https://github.com/prestodb/presto/issues/12923.

Thank you for clarifying and sharing a pointer. This is helpful and makes sense to me.

Tim, I very much appreciate 3 examples you provided. The "Expected" behavior makes sense to me, but I'm confused by "Actual". Does "Actual" mean behavior after changes in the PR? Somehow, I assumed "Actual" means current Presto Java behavior (latest merged code).

I'm seeing the same result for all 3 cases in current code (both Java and Prestissimo): `{1=[{name=John, age=30}]}`

I agree that 2nd result should be different: `{1=[{name=John, AGE=30}]}`

I'm also seeing that JSON keys are fully preserved when casting to a map. Thank feels right.

```
> SELECT CAST(json_extract('{"1":[{"name": "John", "AGE": 30}]}', '$') AS MAP<bigint, ARRAY<MAP<VARCHAR, VARCHAR>>>) AS data;
           data
---------------------------
 {1=[{name=John, AGE=30}]}
```

>  if we consider unquoted field names as case-insensitive when matching, there are still some behaviors that need to be clarified, mainly involving field names that differ only in case. 

Here, I believe 3 and 4 should result in error: Duplicate field: key1.

My thinking is the following.

* ROW(KEY1 VARCHAR, "key1" VARCHAR) defines a struct with duplicate fields (key1, key1) because the first field is normalized to lower case.

I expect this type to be reject outright. However, current behavior is not consistent. Sometimes this type is rejected, but other times it is accepted. It feels like it should be rejected in all these cases.

```
presto> select cast(row(1, 2) as ROW(KEY1 VARCHAR, "key1" VARCHAR));
  _col0
----------
 {key1=2}
(1 row)

presto> select typeof(cast(row(1, 2) as ROW(KEY1 VARCHAR, "key1" VARCHAR)));
                _col0
-------------------------------------
 row("key1" varchar, "key1" varchar)
(1 row)

presto> select cast(row(1, 2) as ROW(KEY1 VARCHAR, "key1" VARCHAR)).key1;
Query 20240209_085621_09358_9e2qz failed: Ambiguous field com.facebook.presto.common.type.RowType$Field@7b7cd13 in type row("key1" varchar, "key1" varchar)
```

* ROW(key1 VARCHAR) defines a struct with a single field: key1 (case-insensitive). Both "key1" and "KEY1" in JSON match, hence, an error.

```
presto> SELECT CAST(json_extract('{"1":[{"key1": "John", "KEY1":"Johnny"}]}', '$') AS MAP<bigint, ARRAY<ROW(key1 VARCHAR)>>) AS data;
Query 20240209_085818_09403_9e2qz failed: Cannot cast to map(bigint,array(row(key1 varchar))). Duplicate field: key1
{"1":[{"key1":"John","KEY1":"Johnny"}]}
```

* ROW("key1" VARCHAR) defines a struct with a single field: key1 (case-sensitive). Now, only "key1" in JSON matches hence, no error.

To summarize, I believe we are saying that struct fields have 2 properties: field name, a boolean indicating whether name is case-sensitive.

ROW(foo INTEGER, "Foo") has 2 distinct fields: (foo, case-insensitive) and (Foo, case-sensitive). 

```
    public static class Field
    {
        private final Type type;
        private final Optional<String> name;
        private final boolean delimited; // This flag determines whether name is case-sensitive or not.
```

Side nite: RowType in Velox doesn't have 'delimited' flag on fields. We may need to figure out how to add it.

ROW type documentation doesn't discuss 'delimited' flat at all: https://prestodb.io/docs/current/language/types.html#row

It would be nice to update. CC: @steveburnett

mbasmanova:
CC: @kagamiori @gggrace14

steveburnett:
> Side nite: RowType in Velox doesn't have 'delimited' flag on fields. We may need to figure out how to add it.
> 
> ROW type documentation doesn't discuss 'delimited' flat at all: https://prestodb.io/docs/current/language/types.html#row
> 
> It would be nice to update. CC: @steveburnett

@mbasmanova thanks for mentioning this! I can open a docs issue to update the page you link to.

tdcmeehan:
@mbasmanova one point before we move on from how to deprecate features.  I agree that there's lots of examples where no one has gone back and deleted the deprecated feature, but I think it's important that we still have the possibility of doing so because this is something we can improve on as a community.  It's really just a practice that we can get better at.  I just don't want us to give up on cleaning up the code and making it simpler.

> Tim, I very much appreciate 3 examples you provided. The "Expected" behavior makes sense to me, but I'm confused by "Actual". Does "Actual" mean behavior after changes in the PR? Somehow, I assumed "Actual" means current Presto Java behavior (latest merged code).

I meant the behavior in this PR, not on trunk.  I observed the same, that all three test cases pass and are consistent on trunk. 

I agree that it's reasonable to interpret that 3 should fail.  In reality, we lower case fields which are not delimited, and it's simpler to implement this than to first parse delimited fields, then non-delimited fields to avoid a naming collision.  I have yet to read the spec, and would appreciate if someone could invest some time on this, but perhaps we can proceed and revisit if we notice anything that contradicts what is described in the spec.

@hantangwangd does this make sense for you as well?  To summarize, I think what needs to happen is:

`SELECT CAST(json_extract('{"1":[{"name": "John", "AGE": 30}]}', '$') AS MAP<bigint, ARRAY<ROW(name VARCHAR, age VARCHAR)>>) AS data;`
Expected result: `{1=[{name=John, age=30}]}`

`SELECT CAST(json_extract('{"1":[{"name": "John", "AGE": 30}]}', '$') AS MAP<bigint, ARRAY<ROW(name VARCHAR, "age" VARCHAR)>>) AS data;`
Expected result: `{1=[{name=John, age=null}]`

`SELECT CAST(json_extract('{"1":[{"key1": "John", "KEY1": "Johnny"}]}', '$') AS MAP<bigint, ARRAY<ROW(KEY1 VARCHAR, "key1" VARCHAR)>>) AS data;`
`SELECT CAST(json_extract('{"1":[{"key1": "John", "KEY1":"Johnny"}]}', '$') AS MAP<bigint, ARRAY<ROW(key1 VARCHAR)>>) AS data;`
Expected result: `EXCEPTION`

SELECT CAST(json_extract('{"1":[{"key1": "John", "KEY1": "Johnny"}]}', '$') AS MAP<bigint, ARRAY<ROW("key1" VARCHAR, "KEY1" VARCHAR)>>) AS data;
`SELECT CAST(json_extract('{"1":[{"key1": "John", "KEY1": "Johnny"}]}', '$') AS MAP<bigint, ARRAY<ROW(key1 VARCHAR, "KEY1" VARCHAR)>>) AS data;`
Expected result: `{1=[{key1=John, KEY1=Johnny}]}`

hantangwangd:
> Here, I believe 3 and 4 should result in error: Duplicate field: key1.
> ROW(KEY1 VARCHAR, "key1" VARCHAR) defines a struct with duplicate fields (key1, key1) because the first field is normalized to lower case.

----

> In reality, we lower case fields which are not delimited, and it's simpler to implement this than to first parse delimited fields, then non-delimited fields to avoid a naming collision. 

@tdcmeehan  @mbasmanova Thanks a lot for all your examples and viewpoints, they are very guiding for me. I strongly agree that unquoted field names could be case insensitive when matching, but should always be normalized to lower case as part of RowType definition.  That is, field names as part of type definition should not be affected by actual data. Moreover, its more easy to implement.

Following the above principles, the latest examples provided by @tdcmeehan  are exactly the case. I completely agree with them.

> I expect this type to be reject outright. However, current behavior is not consistent. Sometimes this type is rejected, but other times it is accepted. It feels like it should be rejected in all these cases.

Thanks for providing this information, I would like to handle these inconsistent behaviors.

mbasmanova:
>  I agree that there's lots of examples where no one has gone back and deleted the deprecated feature, but I think it's important that we still have the possibility of doing so because this is something we can improve on as a community. It's really just a practice that we can get better at. I just don't want us to give up on cleaning up the code and making it simpler.

@tdcmeehan I'm not sure I agree with this. Large users simply cannot migrate away from legacy behaviors (array_agg, for example). If we as a community will start removing support for these, the large users will have no choice but fork the project.

hantangwangd:
@tdcmeehan @mbasmanova The behavior of casting from JSON to ROW has been modified to what was discussed and agreed above. And all the examples provided in discussions have been added to test cases. Please take a look, thanks. Any questions please let me know.

tdcmeehan:
> > I agree that there's lots of examples where no one has gone back and deleted the deprecated feature, but I think it's important that we still have the possibility of doing so because this is something we can improve on as a community. It's really just a practice that we can get better at. I just don't want us to give up on cleaning up the code and making it simpler.
> 
> @tdcmeehan I'm not sure I agree with this. Large users simply cannot migrate away from legacy behaviors (array_agg, for example). If we as a community will start removing support for these, the large users will have no choice but fork the project.

@mbasmanova like most things, I think the answer is not black or white but somewhere in the gray middle.  Feedback from lots of large users that migrations away from `array_agg` should inform our decision to delete the legacy behavior.  So for example, Meta saying we can't fix the behavior of cast from JSON to ROW because it's not feasible to correct the workload is a really important and valuable piece of feedback that should absolutely be considered. (FYI, I haven't looked into the `array_agg` backwards compatibility and have no opinion on it, I'm just using it as an example.)

That being said, there's lots of open source examples of large companies using data infrastructure that involves these companies doing some sort of manual action as part of a version upgrade, so I'm pretty confident this is something that's actually achievable.  But let me know if you think otherwise.

I also expect that most large companies have already forked.  This is because you need the optionality to, for example, fix a bug quickly and in a way that immediately resolves an issue, and it's hard to do that in an open source project without a fork.  For example, if you don't have committers yet in the project, or if the committer pool is small, it's the only way to make any real progress quickly.  I really expect nearly every large company has already forked, and that's fine; keeping all the closed source changes inside forks minimal by keeping the community aligned is what's important.  Of course that's what we're all here for and it's why we meet regularly and keep each other in the loop on all the work we're doing, so it's not really something that worries me.

presto-main/src/test/java/com/facebook/presto/operator/scalar/AbstractTestFunctions.java
```diff
@@ -206,6 +206,9 @@ protected void assertNotSupported(String projection, String message)
                 throw failure;
             }
         }
+        catch (AssertionError e) {
+            assertEquals(e.getMessage(), message);
```
tdcmeehan, line 210: Should we just ensure PrestoException is thrown?

presto-main/src/test/java/com/facebook/presto/operator/scalar/FunctionAssertions.java
```diff
@@ -487,6 +497,9 @@ public void assertInvalidCast(String projection, String message)
                 throw failure;
             }
         }
+        catch (AssertionError e) {
+            assertEquals(e.getMessage(), message);
```
tdcmeehan, line 501: Here as well -- should this just be a PrestoException

presto-common/src/main/java/com/facebook/presto/common/type/TypeSignature.java
```diff
@@ -609,6 +618,10 @@ else if (c == ',' && bracketLevel == 1) {
         return new TypeSignature(signature.substring(0, StandardTypes.ROW.length()), fields);
     }
 
+    private static void checkDuplicateAndAdd(Set<String> fieldNames, String fieldName)
+    {
+        verify(fieldNames.add(fieldName), "Duplicate field: " + fieldName);
```
tdcmeehan, line 623: Can you help me understand why we need this change?

hantangwangd: This code is trying to handle the inconsistent behaviors in various statements about RowType provided by @mbasmanova. Listing as follows: 

```
presto> select typeof(cast(row(1, 2) as ROW(KEY1 VARCHAR, "key1" VARCHAR)));
                _col0
-------------------------------------
 row("key1" varchar, "key1" varchar)
(1 row)

presto> select cast(row(1, 2) as ROW(KEY1 VARCHAR, "key1" VARCHAR)).key1;
Query 20240209_085621_09358_9e2qz failed: Ambiguous field com.facebook.presto.common.type.RowType$Field@7b7cd13 in type row("key1" varchar, "key1" varchar)

presto> select cast(row(1, 2) as ROW(KEY1 VARCHAR, "key1" VARCHAR));
  _col0
----------
 {key1=2}
(1 row)
```

This code would check the keys in analyzing stage and consistently throw an exception if existing any duplicate ones.

presto-main/src/test/java/com/facebook/presto/operator/scalar/AbstractTestFunctions.java
hantangwangd: The exception for `RowType`'s duplicate key check is thrown from `TypeSignature` in analyzing stage. It's in common module and could not recognize `PrestoException`. Seems we could not throw a `PrestoException` from common module. What your opinion, is there a better way to handle this situation?

presto-docs/src/main/sphinx/functions/json.rst
```diff
@@ -71,6 +71,32 @@ Cast from JSON
...
+        SELECT CAST(JSON '{"v1":123,"V2":"abc","v3":true}' AS ROW(v1 BIGINT, v2 VARCHAR, v3 BOOLEAN)); -- {v1=123, v2=abc, v3=true}
+        SELECT CAST(JSON '{"v1":123,"V2":"abc","v3":true}' AS ROW(v1 BIGINT, "V2" VARCHAR, v3 BOOLEAN)); -- {v1=123, V2=abc, v3=true}
+        SELECT CAST(JSON '{"v1":123,"V2":"abc","v3":true}' AS ROW(v1 BIGINT, "v2" VARCHAR, v3 BOOLEAN)); -- {v1=123, v2=null, v3=true}
+        SELECT CAST(JSON '{"v1":123,"V2":"abc", "v2":"abc2","v3":true}' AS ROW(v1 BIGINT, v2 VARCHAR, "V2" VARCHAR, v3 BOOLEAN)); -- {v1=123, v2=abc2, V2=abc, v3=true}
+
+    If the name of a field does not match (including case sensitivity), the value is null.
+
+    To totally ignore the case of field names in RowType when casting from ``JSON`` to ``ROW``
+    (for example, for legacy support) set config property ``legacy_json_cast`` to ``true``
+    in the coordinator and the worker's `config properties <https://prestodb.io/docs/0.285/installation/deployment.html#config-properties>`_.
```
steveburnett, line 89:
```suggestion
    in the coordinator and the worker's `config properties <../installation/deployment.html#config-properties>`_.
```

Fixed link from an absolute link to an older version of the docs to a link. Verified suggestion in a local build. 

As a suggestion: the destination of the link ([Config Properties](https://prestodb.io/docs/current/installation/deployment.html#config-properties) in Deploying Presto) does not mention the config property ``legacy_json_cast``. Suggest you consider adding the config property ``legacy_json_cast`` to [Properties Reference](http://prestodb.io/docs/current/admin/properties.html), and changing this link to point to the section of [Properties Reference](http://prestodb.io/docs/current/admin/properties.html) that you add it to.

presto-docs/src/main/sphinx/functions/json.rst
```diff
@@ -71,6 +71,32 @@ Cast from JSON
...
+    After setting the property, the matching would be case-insensitive as the following example::
+
+        SELECT CAST(JSON '{"v1":123,"V2":"abc","v3":true}' AS ROW(v1 BIGINT, v2 VARCHAR, v3 BOOLEAN)); -- {v1=123, v2=abc, v3=true}
+        SELECT CAST(JSON '{"v1":123,"V2":"abc","v3":true}' AS ROW(v1 BIGINT, "V2" VARCHAR, "V3" BOOLEAN)); -- {v1=123, V2=abc, V3=true}
+
+    Nevertheless, executing the following statement returns an error
```
steveburnett, line 95:
```suggestion
    The following statement returns an error
```

steveburnett (changes requested):
Took another look today and made a suggestion. Let me know what you think!

presto-docs/src/main/sphinx/functions/json.rst
hantangwangd: Hi @steveburnett, thanks for your suggestion, I'm not so sure whether it's appropriate to add `legacy_json_cast` to [Properties Reference](http://prestodb.io/docs/current/admin/properties.html), for the section is used to describe the most important config properties. Feels `legacy_json_cast` is not so important as properties like `join-distribution-type` or `experimental.spill-enabled` etc. What your opinion, do you still think it's appropriate to be added there?

steveburnett: @hantangwangd that's a good question! 

To back up a level of abstraction, I feel that the config property `legacy_json_cast` should be documented _somewhere_ in the Presto docs. If you or others disagree with that, I'm open to understanding why, and if we conclude that we don't need to document, it then we don't need to worry about where to not document it :). 

Assuming that `legacy_json_cast` should be documented in the Presto docs, let's discuss where. 

I feel that [Properties Reference](http://prestodb.io/docs/current/admin/properties.html) should be more complete of a reference than it currently is, with the primary exclusion factor being the stated "not include any connector-specific catalog configuration properties". 

For example, I was planning to update [Properties Reference](http://prestodb.io/docs/current/admin/properties.html) with the properties in [Config Properties](https://prestodb.io/docs/current/installation/deployment.html#config-properties) in Deploying Presto at some point. I've thought about investigating if we could automate generating a list of config properties from the current build of Presto, and automatically publishing that list on this page in a "docs-as-code" strategy. Therefore I thought Properties Reference would be a good place to document this new and not-connector-specific config property. 

Do you have a better idea of where to document `legacy_json_cast`? I'm less emphatic about where in the doc than I am that it should be in the documentation. Let me know what you think!

hantangwangd: @steveburnett Thanks for your detailed description. Agree that we should document the config property `legacy_json_cast`. And since you are planning to update and enlarge [Properties Reference](http://prestodb.io/docs/current/admin/properties.html), then I think it would be appropriate to add `legacy_json_cast` there. Will do it!

hantangwangd:
Hi @steveburnett, I have modified the docs following your suggestion and added the release note. Built the docs locally and verified there's no problem. Please take a look when convenient, thanks!

presto-docs/src/main/sphinx/admin/properties.rst
```diff
@@ -972,4 +972,16 @@ Logging Properties
...
+------------------------------
+
+``legacy_json_cast``
+^^^^^^^^^^^^^^^^^^^^^
+
+    * **Type:** ``boolean``
+    * **Default value:** ``false``
+
+    When casting from ``JSON`` to ``ROW``, by default the case of double quoted field names in RowType is strictly enforced when matching, and unquoted field name in RowType is case-insensitive when matching.
+    To totally ignore the case of field names in RowType when casting from ``JSON`` to ``ROW``(for legacy support), set config property ``legacy_json_cast`` to ``true``. After setting the property, the matching would be always case-insensitive.
```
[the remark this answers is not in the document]
hantangwangd: My fault, I only focus on the issue of links. Thanks for pointing out that. Fixed!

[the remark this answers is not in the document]
steveburnett: > My fault, I only focus on the issue of links. Thanks for pointing out that. Fixed!

I make this mistake myself a lot, so I notice it when it happens.

steveburnett (approved):
LGTM! (docs)

Pull of updated branch, local build, looks great. Thanks for the very quick response!

presto-main/src/test/java/com/facebook/presto/type/TestRowOperators.java
```diff
@@ -503,7 +593,10 @@ public void testRowCast()
         assertFunction("cast(row(1,null,3) as row(aa bigint, bb boolean, cc boolean)).aa", BIGINT, 1L);
         assertFunction("cast(row(null,null,null) as row(aa bigint, bb boolean, cc boolean)).aa", BIGINT, null);
 
-        assertInvalidFunction("CAST(ROW(1, 2) AS ROW(a BIGINT, A DOUBLE)).a");
```
rschlussel, line 506: I'll look at the whole PR when I have a chance, but since you've deleted this, you can now rid of the deprecated assertInvalidFunction that this is using from AbstractTestFunctions and FunctionAssertions.  This was the only place using it.

presto-common/src/main/java/com/facebook/presto/common/type/TypeSignature.java
rschlussel: This shouldn't be a verify failure if the is something the user could send us in a query.  Maybe create a custom exception and handle it in Failures.toErrorCode(). This ultimately should be a semantic exception/user error. Otherwise we will ultimately wrap this in a Presto exception as a GENERIC_INTERNAL_ERROR.

presto-main/src/test/java/com/facebook/presto/type/TestRowOperators.java
hantangwangd: Ok, thank you for the reminder, I will remove them.

presto-common/src/main/java/com/facebook/presto/common/type/TypeSignature.java
hantangwangd: Sure, got it. I will fix it.

presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java
```diff
@@ -277,6 +277,7 @@ public class FeaturesConfig
     private boolean useDefaultsForCorrelatedAggregationPushdownThroughOuterJoins = true;
     private boolean mergeDuplicateAggregationsEnabled = true;
     private boolean fieldNamesInJsonCastEnabled;
+    private boolean legacyJsonCast;
```
rschlussel, line 280: I'm not sure that defaulting to false is the right thing here. In particular, while it's definitely buggy to make the values lower case, the keys it's much less clear, as quotation marks in the json string aren't the same as quotation marks in sql.  They aren't evidence that the user wants to preserve the case.

rschlussel, line 280: It's not clear to me that this should default to false or even be considered legacy behavior.  For sure making the content lower case is a bug and should be fixed for all cases. But for the keys, quoting in json is not evidence of trying to preserve the case, it's just how you define string types, so it's not clear one way or the other to me with regards to treating them as quoted or unquoted identifiers.

rschlussel: never mind, i think i misunderstood what was happening here. it's when the row fields are quoted that this matching is case sensitive, not when the json fields are in quotes.  However, I still think it's worth considering the default value, as this is a change that will break people's production deployments. IMO, we should leave the value as true at first to leave people time to migrate/update their configurations with this new property.

presto-main/src/main/java/com/facebook/presto/util/JsonUtil.java
```diff
@@ -1300,40 +1303,72 @@ public static void parseJsonToSingleRowBlock(
...
                     throw new JsonCastException(format("Expected a json field name, but got %s", parser.getText()));
                 }
-                String fieldName = parser.getText().toLowerCase(Locale.ENGLISH);
-                Integer fieldIndex = fieldNameToIndex.get().get(fieldName);
+
+                String fieldName = parser.getText();
+                Integer fieldIndex = caseSensitiveWhenMatching.get(fieldName);
+                if (fieldIndex == null) {
+                    fieldIndex = caseInsensitiveWhenMatching.get(fieldName.toLowerCase(ENGLISH));
+                }
```
rschlussel, line 1360: for safety, throw an exception if it's not in either.

rschlussel (commented):
question about the default config value, and a small comment, but generally the change looks good.  Also, I feel this needs a highlight in the release notes and not just a regular release note, as it is a behavior change that could cause query failures.

presto-main/src/main/java/com/facebook/presto/util/JsonUtil.java
hantangwangd: Sure!

hantangwangd:
> question about the default config value, and a small comment, but generally the change looks good. Also, I feel this needs a highlight in the release notes and not just a regular release note, as it is a behavior change that could cause query failures.

@rschlussel thanks for your review! I'm OK about the default config value, as it's involving the users' current behavior. @tdcmeehan What is your opinion?

A highlight in the release notes sounds great!

tdcmeehan:
We can target this PR for 0.287 with it disabled by default, then change the default in advance of the cut for 0.288.

presto-main/src/main/java/com/facebook/presto/util/JsonUtil.java
hantangwangd: After re-check the code, I think maybe we shouldn't throw an exception when it's not in either, so that it would fall into the `parser.skipChildren()` below. The result is, when we have a field in json that is not defined and matched in the row type to cast, we would ignore the field rather than throw an exception. For example, executing the following statement:

```
SELECT CAST(JSON '{"v1":123,"v2":"abc","v3":true, "v4":"1234"}' AS ROW(v1 BIGINT, v2 VARCHAR, v3 BOOLEAN, v5 VARCHAR));
```

we would get the following result, rather than get an exception:

```
               _col0                
------------------------------------
 {v1=123, v2=abc, v3=true, V5=null} 
```
What do you think, is this behavior reasonable?

hantangwangd:
@rschlussel @tdcmeehan @steveburnett  I have changed the default config value `legacy-json-cast` to `true` for now, and modify the docs accordingly. Please take a look when convenient, thanks!

rschlussel (dismissed):

presto-docs/src/main/sphinx/admin/properties.rst
```diff
@@ -989,3 +989,16 @@ Logging Properties
...
+------------------------------
+
+``legacy_json_cast``
+^^^^^^^^^^^^^^^^^^^^^
+
+    * **Type:** ``boolean``
+    * **Default value:** ``true``
+
+    When casting from ``JSON`` to ``ROW``, by default ignore the case of field names in RowType when casting from ``JSON`` to ``ROW`` (for legacy support), the matching would be always case-insensitive.
+    To strictly enforce the case of double quoted field names in RowType when matching, set config property ``legacy_json_cast`` to ``false``. After setting the property, the matching for quoted field name would be case-sensitive, and the matching for unquoted field name remains case-insensitive.
```
steveburnett, line 1004:
```suggestion
    
    Set ``legacy_json_cast`` to ``false`` to strictly enforce the case-sensitivity of double quoted field names in ``RowType`` when matching. Matching for unquoted field names remains case-insensitive.
```
Suggest imperative mode for "Set ``legacy_json_cast`` to ``false`` ", and prefacing it with a line break to left-justify this instruction to make it more obvious to the reader what they must do if they want a change in the behavior. Separated "matching for unquoted field name remains case-insensitive" as its own sentence to improve readability. 

You use both "double quoted" and "quoted". Please check my suggestion to make sure that it says what you want it to mean, and correct it as needed.

presto-docs/src/main/sphinx/functions/json.rst
```diff
@@ -71,6 +71,30 @@ Cast from JSON
         SELECT CAST(JSON '{"k1": [1, 23], "k2": 456}' AS MAP(VARCHAR, JSON)); -- {k1 = JSON '[1,23]', k2 = JSON '456'}
         SELECT CAST(JSON '[null]' AS ARRAY(JSON)); -- [JSON 'null']
 
+.. note::
+
+    When casting from ``JSON`` to ``ROW``, by default the case of double quoted field names
```
steveburnett, line 76:
```suggestion
    When casting from ``JSON`` to ``ROW``, for legacy support the case of double quoted field names
```

presto-docs/src/main/sphinx/functions/json.rst
```diff
@@ -71,6 +71,30 @@ Cast from JSON
...
+
+        SELECT CAST(JSON '{"v1":123,"V2":"abc","v3":true}' AS ROW(v1 BIGINT, v2 VARCHAR, v3 BOOLEAN)); -- {v1=123, v2=abc, v3=true}
+        SELECT CAST(JSON '{"v1":123,"V2":"abc","v3":true}' AS ROW(v1 BIGINT, "V2" VARCHAR, "V3" BOOLEAN)); -- {v1=123, V2=abc, V3=true}
+
+    The following statement returns an error due to duplicate field::
+
+        SELECT CAST(JSON '{"v1":123,"V2":"abc","v2":"abc2","v3":true}' AS ROW(v1 BIGINT, "V2" VARCHAR, v2 VARCHAR, "V3" BOOLEAN));
+
+    To enforce the case of field names in RowType when casting from ``JSON`` to ``ROW`` set config property ``legacy_json_cast`` to ``false``
+    in the coordinator and the worker's `config properties <../admin/properties.html#legacy-compatible-properties>`_.
```
steveburnett, line 87:
```suggestion
    in the coordinator and the worker's `configuration properties <../admin/properties.html#legacy-compatible-properties>`_.
```
Minor rephrasing. Including the link to the config property documentation, as you did, is a great help for the reader.

steveburnett (changes requested):
Nice work on the docs! I made some minor suggestions and nits about formatting and phrasing. Let me know what you think!

presto-docs/src/main/sphinx/admin/properties.rst
```diff
@@ -989,3 +989,16 @@ Logging Properties
...
+Legacy Compatible Properties
+------------------------------
+
+``legacy_json_cast``
+^^^^^^^^^^^^^^^^^^^^^
+
+    * **Type:** ``boolean``
+    * **Default value:** ``true``
+
+    When casting from ``JSON`` to ``ROW``, by default ignore the case of field names in RowType when casting from ``JSON`` to ``ROW`` (for legacy support), the matching would be always case-insensitive.
```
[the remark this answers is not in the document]
hantangwangd: I have removed the second 'when casting from ``JSON`` to ``ROW``'. It seems a little duplicate, is that ok?

hantangwangd:
@steveburnett Thanks for your detailed suggestions. They are all fixed, only one minor modification. Please take a look when available.

presto-docs/src/main/sphinx/functions/json.rst
```diff
@@ -71,6 +71,30 @@ Cast from JSON
...
+    in ``RowType`` is ignored when matching. For example::
+
+        SELECT CAST(JSON '{"v1":123,"V2":"abc","v3":true}' AS ROW(v1 BIGINT, v2 VARCHAR, v3 BOOLEAN)); -- {v1=123, v2=abc, v3=true}
+        SELECT CAST(JSON '{"v1":123,"V2":"abc","v3":true}' AS ROW(v1 BIGINT, "V2" VARCHAR, "V3" BOOLEAN)); -- {v1=123, V2=abc, V3=true}
+
+    The following statement returns an error due to duplicate field::
+
+        SELECT CAST(JSON '{"v1":123,"V2":"abc","v2":"abc2","v3":true}' AS ROW(v1 BIGINT, "V2" VARCHAR, v2 VARCHAR, "V3" BOOLEAN));
+
+    To enforce the case of field names in ``RowType`` when casting from ``JSON`` to ``ROW`` set the configuration property ``legacy_json_cast`` to ``false``
```
steveburnett, line 86:
```suggestion
    To enforce the case of field names in ``RowType`` when casting from ``JSON`` to ``ROW``, set the configuration property ``legacy_json_cast`` to ``false``
```
I should have caught this previously: the sentence reads better with a comma here as a pause before the imperative instruction.

steveburnett (changes requested):
Thanks for the quick revision! One tiny nit I should have caught earlier, but everything else looks great.

presto-docs/src/main/sphinx/functions/json.rst
hantangwangd: Thanks for the point, fixed.

steveburnett (approved):
LGTM! (docs)

Pull updated branch, new local build of docs, everything looks good. Thanks!

presto-main/src/main/java/com/facebook/presto/util/Failures.java
```diff
@@ -171,6 +173,10 @@ private static ErrorCode toErrorCode(Throwable throwable)
             return SLICE_TOO_LARGE.toErrorCode();
         }
 
+        if (throwable instanceof InvalidTypeDefinitionException) {
```
tdcmeehan, line 176: Rather than catch and redefine, let's just directly throw a `PrestoException` with `INVALID_TYPE_DEFINITION`.

tdcmeehan (changes requested):
LGTM % the error handling can be cleaned up.

presto-main/src/main/java/com/facebook/presto/util/Failures.java
rschlussel: you can't throw a PrestoException from presto-common, as it's defined in presto-spi

tdcmeehan (approved):

PR #21874, opened 2024-02-07
[native] Revisit stuck operator detection.

spershin:
## Description

We have recently added the functionality to take worker offline when we detect long running operators (stuck ones or super slow). Such situation can lead to all Driver threads being blocked and already scheduled queried not progressing and new ones getting queued on the workers and on the coordinator. As the result the cluster becomes inoperational.

We observed this with just one or two workers in such condition can bring down the whole cluster.
The solution was to set the worker state to 'shutting_down' - this will signal to the coordinator to not schedule any more tasks on it, while allowing the already running tasks to complete (if they can).

The current implementation has few issues:
1. From the log it is not clear which operator got stuck.
2. When announcement is succeeded (we do and log every 30 seconds), it is not clear what state the worker is in.
3. Operator might not be truly stuck, but rather slow running, so it can eventually complete. Need to bring the worker back to the active mode.
4. Trace Context output is abrupt and confusing in the log - need to give engineer a hint what they are looking at.

So, here we change the following:
1. For each stuck operator we log the operator type, node id and the method name.
2. We add the node/worker state to the Announcement succeeded message.
3. We attach worker back when no stuck operators are detected.
4. Code changed a bit (to use folly::synchronized) to ensure we change the shuttingDown flag and the state safely.
5. Prefix the TraceContext logging with the 'title' to better understand what it is.
6. If the server does not have CoordinatorDiscoverer (and Announcer) we don't spin the WatchDog thread up. That situation is applied to the Spark on Velox.

```
============== Before:

0207 09:36:55.019454   576 PeriodicTaskManager.cpp:680] Stuck operator: tid=48549 taskId=20240207_085357_24936_99p2y.2.0.128.0 opId=2
E0207 09:36:55.019484   576 PeriodicTaskManager.cpp:693] Driver::run=1 entered 34989390 avg ms 3 max ms 769847 continuous for 31736365

W0207 09:36:55.419687   558 PeriodicTaskManager.cpp:697] Changing node status to SHUTTING_DOWN due to detected stuck drivers

I0207 12:13:40.209054   491 PeriodicServiceInventoryManager.cpp:115] Announcement succeeded: HTTP 202


============== After:

E0207 12:14:35.888252   558 PeriodicTaskManager.cpp:680] Stuck operator: tid=1040 taskId=20240207_201049_00000_2j8r9.3.0.0.0 opCall=TableScan.1::getOutput duration= 1m 20s
W0207 12:14:35.888353   558 PeriodicTaskManager.cpp:696] TraceContext::status:
Driver::run: numThreads=15 numEnters=88441 avgMs=1 maxMs=120834 continued=83672

W0207 12:14:35.888398   558 PeriodicTaskManager.cpp:699] Will detach worker due to detected stuck operators
W0207 12:14:35.888403   558 PrestoServer.cpp:763] Changing node status to SHUTTING_DOWN.

W0207 12:15:35.888581   558 PeriodicTaskManager.cpp:707] Will attach worker due to the absence of stuck operators
W0207 12:15:35.888610   558 PrestoServer.cpp:771] Changing node status to ACTIVE.

I0207 12:13:40.209054   491 PeriodicServiceInventoryManager.cpp:115] Announcement succeeded: HTTP 202. State: active.
I0207 12:14:45.136858   491 PeriodicServiceInventoryManager.cpp:115] Announcement succeeded: HTTP 202. State: shutting_down.
```

As a test I ran a query in a cluster of 60 workers with 0.1% of chance for every TableScan::getOutput() to sleep for 2 minutes, while having 1 minute threshold to detect a stuck operator.
A query (normally running in under 20 seconds) took 51 minutes to complete with the number of active workers fluctuating during the run. After the query completion all workers returned to the active state. Logs show expected output.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
+++ b/presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
@@ -120,7 +120,9 @@ void PeriodicTaskManager::start() {
     addArbitratorStatsTask();
   }
 
-  addWatchdogTask();
+  if (server_ && server_->hasCoordinatorDiscoverer()) {
+    addWatchdogTask();
+  }
 
   oneTimeRunner_.start();
 }
@@ -678,11 +680,14 @@ void PeriodicTaskManager::addWatchdogTask() {
         RECORD_METRIC_VALUE(kCounterNumTasksDeadlock, deadlockTasks.size());
         for (const auto& call : stuckOpCalls) {
           LOG(ERROR) << "Stuck operator: tid=" << call.tid
-                     << " taskId=" << call.taskId << " opId=" << call.opId;
+                     << " taskId=" << call.taskId << " opCall=" << call.opCall
+                     << " duration= " << velox::succinctMillis(call.durationMs);
         }
         RECORD_METRIC_VALUE(kCounterNumStuckDrivers, stuckOpCalls.size());
         if (!deadlockTasks.empty() || !stuckOpCalls.empty()) {
           detachWorker();
+        } else {
+          maybeAttachWorker();
         }
       },
       60'000'000, // 60 seconds
@@ -690,13 +695,19 @@ void PeriodicTaskManager::addWatchdogTask() {
 }
 
 void PeriodicTaskManager::detachWorker() {
-  LOG(ERROR) << velox::process::TraceContext::statusLine();
+  LOG(WARNING) << "TraceContext::status:\n"
+               << velox::process::TraceContext::statusLine();
   if (server_ && server_->nodeState() == NodeState::kActive) {
-    // Benefit of shutting down is that the queries that aren't stuck yet will
-    // be finished.  While stopping announcement would kill them.
-    LOG(ERROR)
-        << "Changing node status to SHUTTING_DOWN due to detected stuck drivers";
-    server_->setNodeState(NodeState::kShuttingDown);
+    LOG(WARNING) << "Will detach worker due to detected stuck operators";
+    server_->detachWorker();
+  }
+}
+
+void PeriodicTaskManager::maybeAttachWorker() {
+  if (server_ && !server_->isShuttingDown() &&
+      server_->nodeState() == NodeState::kShuttingDown) {
+    LOG(WARNING) << "Will attach worker due to the absence of stuck operators";
+    server_->maybeAttachWorker();
   }
 }
 
--- a/presto-native-execution/presto_cpp/main/PrestoServer.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoServer.cpp
@@ -268,6 +268,7 @@ void PrestoServer::run() {
         catalogNames,
         systemConfig->announcementMaxFrequencyMs(),
         sslContext_);
+    updateAnnouncerDetails();
     announcer_->start();
 
     uint64_t heartbeatFrequencyMs = systemConfig->heartbeatFrequencyMs();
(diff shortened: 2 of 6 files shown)


presto-native-execution/presto_cpp/main/PrestoServer.h
```diff
@@ -83,9 +83,16 @@ class PrestoServer {
...
-  void setNodeState(NodeState nodeState) {
-    nodeState_ = nodeState;
-  }
+  /// Set worker into the SHUTTING_DOWN state even if we aren't shutting down.
+  /// This will prevent coordinator from sending new tasks to this worker.
+  void detachWorker();
+
+  /// Set worker into the ACTIVE state if we aren't shutting down.
+  /// This will enable coordinator to send new tasks to this worker.
+  void maybeAttachWorker();
```
mbasmanova, line 92: What does 'maybe' mean here? Would you update the comment to clarify? 

Would it make sense to return a boolean indicating whether the operation succeeded or not?

presto-native-execution/presto_cpp/main/PrestoServer.h
```diff
@@ -83,9 +83,16 @@ class PrestoServer {
...
+
+  /// Changes this node's state.
+  void setNodeState(NodeState nodeState);
```
mbasmanova, line 95: How is this different from detachWorker and maybeAttachWorker?

presto-native-execution/presto_cpp/main/PrestoServer.h
```diff
@@ -214,7 +221,7 @@ class PrestoServer {
   std::unique_ptr<TaskManager> taskManager_;
   std::unique_ptr<TaskResource> taskResource_;
   std::atomic<NodeState> nodeState_{NodeState::kActive};
-  std::atomic_bool shuttingDown_{false};
+  folly::Synchronized<bool> shuttingDown_{false};
```
mbasmanova, line 224: Why is this change?

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -718,32 +719,64 @@ void PrestoServer::initializeVeloxMemory() {
...
+  if (!*readLockedShuttingDown && nodeState() == NodeState::kShuttingDown) {
+    LOG(WARNING) << "Changing node status to ACTIVE.";
+    setNodeState(NodeState::kActive);
+  }
+}
+
+void PrestoServer::setNodeState(NodeState nodeState) {
+  nodeState_ = nodeState;
+  if (announcer_ != nullptr) {
+    announcer_->setDetails(nodeState2String(nodeState_));
```
mbasmanova, line 779: Is this thread safe?

presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
```diff
@@ -678,25 +678,31 @@ void PeriodicTaskManager::addWatchdogTask() {
...
           detachWorker();
+        } else {
+          maybeAttachWorker();
         }
       },
       60'000'000, // 60 seconds
       "Watchdog");
 }
 
 void PeriodicTaskManager::detachWorker() {
```
mbasmanova, line 695: Any reason this method name doesn't have 'maybe' prefix like maybeAttachWorker?

mbasmanova (commented):
@spershin I don't have a lot of context, but have some questions.

presto-native-execution/presto_cpp/main/PrestoServer.h
spershin: @mbasmanova 
It says it in the comment: "Set worker into the ACTIVE state if we aren't shutting down.".
It does not make sense to return bool, because we aren't using it.

presto-native-execution/presto_cpp/main/PrestoServer.h
spershin: Technically, detach wouldn't detach if we are already detached.
But the end result would be 'detached'.

Attach, on the other hand, wouldn't attach if we are detached, because we are shutting down.

That's the difference.

presto-native-execution/presto_cpp/main/PrestoServer.h
spershin: @mbasmanova 

It is mentioned in the PR's description: "Code changed a bit to ensure we change the shuttingDown flag and the state safely."

The reason is that we don't want to attach the worker back if we are truly shutting down, so the two member variables need to be changed under a single lock.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
spershin: Yes, announcer's 'details' member is folly::synchronized and the announcer_ itself is freed only with the destruction of the server object.

presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
spershin: Which method? You highlighted the 'maybeAttachWorker'.

spershin:
@mbasmanova 
Thanks for reviewing! Answered the comments.
I will give this change a test run in the cluster with the stuck operator situation.

presto-native-execution/presto_cpp/main/PrestoServer.h
mbasmanova: I see. Then it seems to make sense to rename detachWorker to maybeDetachWorker for consistency.

presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
mbasmanova: I was asking why detachWorker is not named maybeDetachWorker.

spershin: @mbasmanova 

As I mentioned in another comment (above):

Technically, detach wouldn't detach if we are already detached.
But the end result would be 'detached'.

Attach, on the other hand, wouldn't attach if we are detached, because we are shutting down.

That's the difference.

mbasmanova: I see. I didn't understand it until now.

I wonder if other readers will be confused as well. Would it be possible to update comments to make things a bit clearer?

mbasmanova (approved):
@spershin Thank you for adding more details to PR description. This is very helpful.

Any chance you could share examples of 1, 2, 4 before and after to illustrate the issues and solutions a bit better. 

Thanks.

presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
```diff
@@ -120,7 +120,9 @@ void PeriodicTaskManager::start() {
     addArbitratorStatsTask();
   }
 
-  addWatchdogTask();
+  if (server_ && server_->hasCoordinatorDiscoverer()) {
```
mbasmanova, line 123: Would be nice to describe this change in the PR description and commit message.

mbasmanova (approved):
@spershin Look good % please, update commit message to add details before merging.

Issue #21878, opened 2024-02-07
ST_POINT() function Documentation is unclear


brookbi:
Can someone please document on this page whether to use (longitude, latitude) or (latitude, longitude) while running distance calculations? It's unclear to me for the ST_POINT() function listing

https://prestodb.io/docs/current/functions/geospatial.html

steveburnett:
@mbasmanova and @jagill , I see that you had commits related to this file https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/functions/geospatial.rst. Can you help? 

If you explain here how ST_POINT works to improve the existing docs, I'd be happy to open a docs PR to update this page if you'd like.

brookbi:
It appears that ST_POINT() creates a geospatial point that’s then used for other calculations, like great circle distance, and it seems like it goes in (Longitude, Latitude) order but I’m not 100%On Feb 7, 2024, at 12:09 PM, Steve Burnett ***@***.***> wrote:﻿
@mbasmanova and @jagill , I see that you had commits related to this file https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/functions/geospatial.rst. Can you help?
If you explain here how ST_POINT works to improve the existing docs, I'd be happy to open a docs PR to update this page if you'd like.

mbasmanova:
> it goes in (Longitude, Latitude) order

That's right.

steveburnett:
Thanks, I'll open a PR to update the docs. Thanks!

steveburnett:
@brookbi , I was not able to add you as a reviewer to [PR 21880](https://github.com/prestodb/presto/pull/21880). Please review it and comment - there or here - if the edit solves your concern, or if you have comments.

steveburnett:
@brookbi, @mbasmanova pointed out on [PR 21880](https://github.com/prestodb/presto/pull/21880) that the page https://prestodb.io/docs/current/functions/geospatial.html contains the following explanation: 

``Use ST_GeometryFromText and ST_GeomFromBinary functions to create geometry objects from WKT or WKB. In WKT/WKB, the coordinate order is (x, y). For spherical/geospatial uses, this implies (longitude, latitude) instead of (latitude, longitude).

The basis for the Geometry type is a plane. The shortest path between two points on the plane is a straight line. That means calculations on geometries (areas, distances, lengths, intersections, etc) can be calculated using cartesian mathematics and straight line vectors.``

[...]

PR #21880, opened 2024-02-07
[documentation] update ST_POINT function documentation

steveburnett:
## Description
Update documentation of the ST_POINT function in [Geospatial Functions](https://prestodb.io/docs/current/functions/geospatial.html) to clarify that the x and y values are longitude and latitude. Added an example copied from the introduction section of the same page.  

## Motivation and Context
Addresses [Issue 21878](https://github.com/prestodb/presto/issues/21878). 

## Impact
Documentation. 

## Test Plan
Local build of my changes. Screenshot: 
Screenshot 2024-02-07 at 1 52 04 PM

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-docs/src/main/sphinx/functions/geospatial.rst
+++ b/presto-docs/src/main/sphinx/functions/geospatial.rst
@@ -100,7 +100,8 @@ Constructors
 
 .. function:: ST_Point(x, y) -> Point
 
-    Returns a geometry type point object with the given coordinate values.
+    Returns a geometry type point object with the given longitude and latitude coordinate values. 
+    For example, ``ST_Point(-71.0882, 42.3607)``.
 
 .. function:: ST_Polygon(varchar) -> Polygon
 


presto-docs/src/main/sphinx/functions/geospatial.rst
```diff
@@ -100,7 +100,8 @@ Constructors
 
 .. function:: ST_Point(x, y) -> Point
 
-    Returns a geometry type point object with the given coordinate values.
+    Returns a geometry type point object with the given longitude and latitude coordinate values.
```
mbasmanova, line 103: We may need to be careful here not to create an impression that computations are done on a sphere. All computations are done on a 2d plane with (x, y) coordinates. One can treat a sphere as such..... with x = longiture and y = latitude... but they need to understand what they are doing... I think this is why we documented it as (x, y) and not (longitude, latitude).

steveburnett: Good point! I could add your clarification with something like the following: 

"All computations of this function are done on a 2d plane - not a sphere- with (x, y) coordinates." 

Would that work? Do you have a better suggestion?

mbasmanova: I noticed we already have this:

```
Use ST_GeometryFromText and ST_GeomFromBinary functions to create geometry objects from WKT or WKB. In WKT/WKB, the coordinate order is (x, y). For spherical/geospatial uses, this implies (longitude, latitude) instead of (latitude, longitude).

The basis for the Geometry type is a plane. The shortest path between two points on the plane is a straight line. That means calculations on geometries (areas, distances, lengths, intersections, etc) can be calculated using cartesian mathematics and straight line vectors.
```

Do we need to add anything else?

steveburnett: I saw that, and it helped me understand the context. Because the opener of the issue described that it was unclear to them whether to use (longitude, latitude) or (latitude, longitude) for ST_POINT, I wanted to revise the description of ST_POINT as that existing content didn't help them. 

How about this replacement of my current line 103, bringing your quoted material into context in the ST_POINT definition? 

"Returns a geometry type point object, using longitude and latitude coordinate values as ``x`` and ``y`` to calculate the point on a plane."

mbasmanova: > as that existing content didn't help them.

Is this the case? I don't think we even asked them if they read that paragraph, did we?

steveburnett: The existing content you quote was present on the page when they opened the issue, but I agree that we didn't ask them if they read that paragraph. What would you suggest?

mbasmanova: I suggest asking them whether they read that part of the doc. If not, we need to remind ourselves to remember to read the full page and ask folks to do the same next time.

mbasmanova: or, at least search for "lat, long" on the page before assuming the doc is lacking.

steveburnett:
Closing this PR as the prompting [Issue 21878](https://github.com/prestodb/presto/issues/21878) has been resolved.

PR #21884, opened 2024-02-08
Fix sign shift issue in presto

r-barnes:
Differential Revision: D53577764

### DIFF
--- a/presto-native-execution/presto_cpp/presto_protocol/Base64Util.cpp
+++ b/presto-native-execution/presto_cpp/presto_protocol/Base64Util.cpp
@@ -30,7 +30,7 @@ static const char* kRle = "RLE";
 static const char* kArray = "ARRAY";
 static const char* kMap = "MAP";
 static const char* kInt128Array = "INT128_ARRAY";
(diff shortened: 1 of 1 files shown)


presto-native-execution/presto_cpp/presto_protocol/Base64Util.cpp
```diff
@@ -30,7 +30,7 @@ static const char* kRle = "RLE";
 static const char* kArray = "ARRAY";
 static const char* kMap = "MAP";
 static const char* kInt128Array = "INT128_ARRAY";
-static const __int128_t kInt128Mask = ~(static_cast<__int128_t>(1) << 127);
+static const __uint128_t kInt128Mask = ~(static_cast<__uint128_t>(1) << 127);
```
spershin, line 33: Interesting that velox/type/DecimalUtil.h
has
`  static constexpr uint128_t kInt128Mask = (static_cast<uint128_t>(1) << 127);
`
And it is not really used anywhere.
I think presto_cpp should maybe use that Velox's constant.
But let's merge this one for now to move fast.

spershin (approved):

PR #21893, opened 2024-02-09
Fix min and max for inputs that include NaN values

### LINKED ISSUE #21877: Aggregate functions Max/Min return different results based on order of NaN for floating point types.
Aggregate functions Max/Min return different results based on when NaN is encountered in the input for floating point types. If Nan is the first value then irrespective of what the other values are the result is Nan. This seems wrong. 

## Expected Behavior
Max/Min shouldnt be sensitive to order of NaN and should return same result. 

## Current Behavior
```
presto:di> select max(x) from (values 4.0,nan(),null) T(x);;
 _col0
-------
   4.0
(1 row)

presto:di> select max(x) from (values nan(),4.0,null) T(x);
 _col0
-------
   NaN
(1 row)
```

## Possible Solution
The bug is likely here (https://github.com/prestodb/presto/blob/master/presto-main/src/main/java/com/facebook/presto/operator/aggregation/AbstractMinMaxAggregationFunction.java) where state is initially null and set to Nan and then subsequently all comparison against it fails.

hainenber:
## Description

min and max used to return NaN if the very first value in the input was NaN, but they ignored NaN values otherwise.

After this change min and max return NaN only if all inputs are NaN.

## Motivation and Context
Fixes #21877 

## Impact

Ensure correct value of `max/min()` not being affected by order of NaNs

## Test Plan

- Run `./mvnw clean install` in current PR
- Execute following snippet
```
select max(x) from (values 4.0,nan(),null) T(x);
select max(x) from (values nan(),4,null) T(x);
```
- Results in 
```
presto> select max(x) from (values 4.0,nan(),null) T(x);
 _col0 
-------
   4.0 
(1 row)

Query 20240209_123313_00001_skyr3, FINISHED, 1 node
Splits: 1 total, 1 done (100.00%)
[Latency: client-side: 172ms, server-side: 143ms] [0 rows, 0B] [0 rows/s, 0B/s]

presto> select max(x) from (values nan(),4,null) T(x);
 _col0 
-------
   4.0 
(1 row)

Query 20240209_123324_00002_skyr3, FINISHED, 1 node
Splits: 1 total, 1 done (100.00%)
[Latency: client-side: 64ms, server-side: 47ms] [0 rows, 0B] [0 rows/s, 0B/s]

```

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Fixes issue where NaN value changes final result of aggregation max/min function if given as first item argument.

```

### DIFF
--- a/presto-docs/src/main/sphinx/functions/aggregate.rst
+++ b/presto-docs/src/main/sphinx/functions/aggregate.rst
@@ -98,6 +98,7 @@ General Aggregate Functions
 .. function:: max(x) -> [same as input]
 
     Returns the maximum value of all input values.
+    When x is a REAL or DOUBLE, returns NaN if all input values are NaN. Ignores NaN values otherwise.
 
 .. function:: max(x, n) -> array<[same as x]>
 
@@ -106,6 +107,7 @@ General Aggregate Functions
 .. function:: min(x) -> [same as input]
 
     Returns the minimum value of all input values.
+    When x is a REAL or DOUBLE, returns NaN if all input values are NaN. Ignores NaN values otherwise.
 
 .. function:: min(x, n) -> array<[same as x]>
 
--- a/presto-main/src/main/java/com/facebook/presto/operator/aggregation/AbstractMinMaxAggregationFunction.java
+++ b/presto-main/src/main/java/com/facebook/presto/operator/aggregation/AbstractMinMaxAggregationFunction.java
@@ -278,7 +278,10 @@ public abstract class AbstractMinMaxAggregationFunction
             return;
         }
         try {
-            if ((boolean) methodHandle.invokeExact(value, state.getDouble())) {
+            double currentState = state.getDouble();
+            // If NaN is set as state previously, configure the next state with compared value
+            // to prevent failed comparisons with NaN
+            if (Double.isNaN(currentState) || (boolean) methodHandle.invokeExact(value, currentState)) {
                 state.setDouble(value);
             }
         }
--- a/presto-main/src/test/java/com/facebook/presto/operator/aggregation/TestDoubleMaxAggregation.java
+++ b/presto-main/src/test/java/com/facebook/presto/operator/aggregation/TestDoubleMaxAggregation.java
@@ -16,15 +16,50 @@ package com.facebook.presto.operator.aggregation;
 import com.facebook.presto.common.block.Block;
 import com.facebook.presto.common.block.BlockBuilder;
 import com.facebook.presto.common.type.StandardTypes;
+import com.facebook.presto.metadata.FunctionAndTypeManager;
+import com.facebook.presto.metadata.MetadataManager;
+import com.facebook.presto.spi.function.JavaAggregationFunctionImplementation;
 import com.google.common.collect.ImmutableList;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
 
 import java.util.List;
 
+import static com.facebook.presto.block.BlockAssertions.createDoublesBlock;
 import static com.facebook.presto.common.type.DoubleType.DOUBLE;
+import static com.facebook.presto.operator.aggregation.AggregationTestUtils.assertAggregation;
+import static com.facebook.presto.sql.analyzer.TypeSignatureProvider.fromTypes;
+import static java.lang.Double.NEGATIVE_INFINITY;
+import static java.lang.Double.NaN;
+import static java.lang.Double.POSITIVE_INFINITY;
 
 public class TestDoubleMaxAggregation
         extends AbstractTestAggregationFunction
 {
+    private static final String FUNCTION_NAME = "max";
+    private JavaAggregationFunctionImplementation maxFunction;
+
+    @BeforeClass
+    public void setup()
+    {
+        FunctionAndTypeManager functionAndTypeManager = MetadataManager.createTestMetadataManager().getFunctionAndTypeManager();
+        maxFunction = functionAndTypeManager.getJavaAggregateFunctionImplementation(
+                functionAndTypeManager.lookupFunction(TestDoubleMaxAggregation.FUNCTION_NAME, fromTypes(DOUBLE)));
+    }
+
+    @Test
+    public void max()
+    {
+        assertAggregation(maxFunction, 4.0, createDoublesBlock(4.0));
+        assertAggregation(maxFunction, 5.0, createDoublesBlock(2.0, 4.0, 5.0));
+        assertAggregation(maxFunction, 4.0, createDoublesBlock(4.0, NaN, null));
+        assertAggregation(maxFunction, 4.0, createDoublesBlock(NaN, 4.0, null));
+        assertAggregation(maxFunction, POSITIVE_INFINITY, createDoublesBlock(NaN, null, POSITIVE_INFINITY));
+        assertAggregation(maxFunction, POSITIVE_INFINITY, createDoublesBlock(NaN, null, NEGATIVE_INFINITY, POSITIVE_INFINITY));
+        assertAggregation(maxFunction, null, createDoublesBlock((Double) null));
+        assertAggregation(maxFunction, NaN, createDoublesBlock(NaN));
+    }
+
     @Override
     public Block[] getSequenceBlocks(int start, int length)
     {
--- a/presto-main/src/test/java/com/facebook/presto/operator/aggregation/TestDoubleMinAggregation.java
+++ b/presto-main/src/test/java/com/facebook/presto/operator/aggregation/TestDoubleMinAggregation.java
@@ -16,15 +16,50 @@ package com.facebook.presto.operator.aggregation;
 import com.facebook.presto.common.block.Block;
 import com.facebook.presto.common.block.BlockBuilder;
 import com.facebook.presto.common.type.StandardTypes;
+import com.facebook.presto.metadata.FunctionAndTypeManager;
+import com.facebook.presto.metadata.MetadataManager;
+import com.facebook.presto.spi.function.JavaAggregationFunctionImplementation;
 import com.google.common.collect.ImmutableList;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
 
 import java.util.List;
 
+import static com.facebook.presto.block.BlockAssertions.createDoublesBlock;
 import static com.facebook.presto.common.type.DoubleType.DOUBLE;
+import static com.facebook.presto.operator.aggregation.AggregationTestUtils.assertAggregation;
+import static com.facebook.presto.sql.analyzer.TypeSignatureProvider.fromTypes;
+import static java.lang.Double.NEGATIVE_INFINITY;
+import static java.lang.Double.NaN;
+import static java.lang.Double.POSITIVE_INFINITY;
 
 public class TestDoubleMinAggregation
         extends AbstractTestAggregationFunction
 {
+    private static final String FUNCTION_NAME = "min";
+    private JavaAggregationFunctionImplementation minFunction;
+
+    @BeforeClass
+    public void setup()
+    {
+        FunctionAndTypeManager functionAndTypeManager = MetadataManager.createTestMetadataManager().getFunctionAndTypeManager();
+        minFunction = functionAndTypeManager.getJavaAggregateFunctionImplementation(
+                functionAndTypeManager.lookupFunction(TestDoubleMinAggregation.FUNCTION_NAME, fromTypes(DOUBLE)));
+    }
+
+    @Test
+    public void min()
+    {
+        assertAggregation(minFunction, 4.0, createDoublesBlock(4.0));
+        assertAggregation(minFunction, 2.0, createDoublesBlock(2.0, 4.0, 5.0));
+        assertAggregation(minFunction, 4.0, createDoublesBlock(4.0, NaN, null));
+        assertAggregation(minFunction, 4.0, createDoublesBlock(NaN, 4.0, null));
+        assertAggregation(minFunction, NEGATIVE_INFINITY, createDoublesBlock(NaN, null, NEGATIVE_INFINITY));
+        assertAggregation(minFunction, NEGATIVE_INFINITY, createDoublesBlock(NaN, null, NEGATIVE_INFINITY, POSITIVE_INFINITY));
+        assertAggregation(minFunction, null, createDoublesBlock((Double) null));
+        assertAggregation(minFunction, NaN, createDoublesBlock(NaN));
+    }
+
     @Override
     public Block[] getSequenceBlocks(int start, int length)
     {


presto-main/src/main/java/com/facebook/presto/operator/aggregation/AbstractMinMaxAggregationFunction.java
```diff
@@ -274,7 +274,10 @@ private static void compareAndUpdateState(MethodHandle methodHandle, NullableDou
     {
         if (state.isNull()) {
             state.setNull(false);
-            state.setDouble(value);
+            // Not set value to NaN initially to ensure subsequent comparisons succeed
```
tdcmeehan, line 277: I don't think this will work, as I think `SELECT min(a) from (VALUES NaN()) AS t(a);` should return `NaN`.

tdcmeehan (changes requested):
Can you add some unit tests as well?

hainenber:
> Can you add some unit tests as well?

Sure, let me add one!

presto-main/src/main/java/com/facebook/presto/operator/aggregation/AbstractMinMaxAggregationFunction.java
hainenber: Approach changed in latest commit to cover your described case :D

tdcmeehan (commented):
LGTM % one nit

presto-main/src/test/java/com/facebook/presto/operator/aggregation/AbstractMinMaxAggregationFunctionTest.java
```diff
@@ -0,0 +1,65 @@
...
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+import static com.facebook.presto.block.BlockAssertions.createDoublesBlock;
+import static com.facebook.presto.common.type.DoubleType.DOUBLE;
+import static com.facebook.presto.operator.aggregation.AggregationTestUtils.assertAggregation;
+import static com.facebook.presto.sql.analyzer.TypeSignatureProvider.fromTypes;
+import static java.lang.Double.NaN;
+
+public class AbstractMinMaxAggregationFunctionTest
```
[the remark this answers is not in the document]
hainenber: Done, I've moved them to `TestDoubleMax/MinAggregation` files

hainenber:
Commits squashed and adhere to commit message guideline :D

tdcmeehan (approved):
LGTM

mbasmanova (commented):
@hainenber Would you clarify what is the new behavior and update the documentation?

CC: @steveburnett

spershin:
@hainenber @tdcmeehan @mbasmanova 
This behavior is inconsistent with array_min/array_max which both return NaN when found at least one NaN element.
Do we want to keep max and array_max inconsistent?
Or do we want to make them behave in the same way (either/or)?
Make sense to double-check min_by/max_by too.

tdcmeehan:
@spershin as mentioned in the issue, in my opinion that's a separate problem. I think it requires discussion and thought on how to make them consistent, and this fix can be considered separately.

tdcmeehan:
I also believe there's an existing issue for that somewhere.

spershin:
@tdcmeehan 

We are changing the behavior of min/max here.
Would it not be wise to hold on for a while, discuss the min/max over double issue as a whole and then roll out a change?
To, potentially, disturb users less?

tdcmeehan:
@spershin if we were to document this function better, it would be written something along these lines:

> Returns the maximum value of all input values.
> For double and float types, NaN values are ignored, except if the first value is encountered, then the function will always return NaN.

This behavior seems almost ridiculous to document, and very counterintuitive.  When I look at it through a risk/reward lense, it seems there is more harm to leave this behavior as it is because it is so unintuitive and difficult to document, and my worry is if we tie it to the larger question of how do we consistently handle NaN values in our aggregation functions, then it's going to hold up a relatively straightforward bug fix.  But I am curious if you think differently on the risk/reward for this smaller change.

spershin:
@tdcmeehan 

Looking from the position of "how will it affect existing users and their workloads" when we make a change like this.
Pipelines, dashboards and UIs can potentially start producing very different results with each such change and it can have disruptions for users/clients even if previously it wasn't quite right.

I don't really think much about the documentation, probably because I believe that not many users went and studied the exact behavior of min/max in the manual in the 1st place. Because the function is pretty simple and unless you scrutinize something you won't need that info.

I don't have a strong opinion on this topic. Simply tried to project a set of opinions I'm encountering all the time when it comes to "should we change that part of Presto, because it seems wrong" question.

mbasmanova:
I'm seeing that there are 2 distinct issues with handling NaNs. (1) different functions handle NaNs differently (for example, min vs. array_min); (2) some functions handle NaNs inconsistently (min and max return different results depending on the order of inputs). 

It might be hard to fix (1), but we should be able to fix (2). My impression is that this PR fixes (2) for min and max. BTW, there is a similar issue in min_by and max_by.

@hainenber It looks like you are fixing min and max to always ignore NaN values in the input. If that's the case, let's state this clearly in the PR description and commit message and update documentation at presto-docs/src/main/sphinx/functions/aggregate.rst

mlyublena:
I couldn't find anything in the SQL standard that talks about treatment of NaN in aggregates but I agree with comments above that at the very least the behavior should be consistent across various aggregates and built-in functions, including array aggs and least/greatest

mlyublena:
> I couldn't find anything in the SQL standard that talks about treatment of NaN in aggregates but I agree with comments above that at the very least the behavior should be consistent across various aggregates and built-in functions, including array aggs and least/greatest

Also, most systems I checked with the exception of BigQuery don't specify what behavior is followed wrt NaN (they only do so for NULLs). If we ignore NaN, it makes it consistent with NULL-value behavior.
For Bigquery, they return NaN if any of the values is NaN:
https://cloud.google.com/bigquery/docs/reference/standard-sql/aggregate_functions#min

@hainenber , have you checked what other systems do with NaNs? Postgres, MySQL, SqlServer etc?

hainenber:
> > I couldn't find anything in the SQL standard that talks about treatment of NaN in aggregates but I agree with comments above that at the very least the behavior should be consistent across various aggregates and built-in functions, including array aggs and least/greatest
> 
> Also, most systems I checked with the exception of BigQuery don't specify what behavior is followed wrt NaN (they only do so for NULLs). If we ignore NaN, it makes it consistent with NULL-value behavior. For Bigquery, they return NaN if any of the values is NaN: https://cloud.google.com/bigquery/docs/reference/standard-sql/aggregate_functions#min
> 
> @hainenber , have you checked what other systems do with NaNs? Postgres, MySQL, SqlServer etc?

This is what I've got after few rounds of searching and LLM'ing
- PostgreSQL treats NaN values as equal, and greater than all non-NaN values. (https://www.postgresql.org/docs/current/datatype-numeric.html)
- MySQL and SQL Server converts NaN as Null and perform comparison accordingly.

hainenber:
Done updating `aggregate.rst` for `min` and `max` function. I also clarified the solution in the PR message/description.

I also added tests for positive/negative infinity and null as requested.

presto-docs/src/main/sphinx/functions/aggregate.rst
```diff
@@ -98,6 +98,7 @@ General Aggregate Functions
 .. function:: max(x) -> [same as input]
 
     Returns the maximum value of all input values.
+    NaN values would be ignored in the input.
```
mbasmanova, line 101: This is not quite accurate. When all input values are NaN, the result is NaN, no?

mbasmanova (commented):
@hainenber Thank you for iterating. Looks good % comment about documentation.

CC: @steveburnett

presto-docs/src/main/sphinx/functions/aggregate.rst
mbasmanova: My impression is that the behavior is the following:

```
Returns the maximum value of all input values. 

When x is a REAL or DOUBLE returns NaN if all input values are NaN. Ignores NaN values otherwise.
```

It might be helpful to include some examples.

hainenber: Thanks for bearing with me on this PR. Done amending

mbasmanova (approved):
Thanks.

mbasmanova (commented):
@hainenber Would you update commit message to something like

```
Fix min and max for inputs that include NaN values

min and max used to return NaN if the very first value in the input was NaN, but
they ignored NaN values otherwise.

After this change min and max return NaN only if all inputs are NaN.
```

presto-docs/src/main/sphinx/functions/aggregate.rst
```diff
@@ -98,6 +98,7 @@ General Aggregate Functions
 .. function:: max(x) -> [same as input]
 
     Returns the maximum value of all input values.
+    When x is a REAL or DOUBLE returns NaN if all input values are NaN. Ignores NaN values otherwise.
```
steveburnett, line 101:
```suggestion
    When x is a REAL or DOUBLE, returns NaN if all input values are NaN. Ignores NaN values otherwise.
```
Suggest comma as a break for readability, to separate the condition from the outcome.

presto-docs/src/main/sphinx/functions/aggregate.rst
```diff
@@ -106,6 +107,7 @@ General Aggregate Functions
 .. function:: min(x) -> [same as input]
 
     Returns the minimum value of all input values.
+    When x is a REAL or DOUBLE returns NaN if all input values are NaN. Ignores NaN values otherwise.
```
steveburnett, line 110:
```suggestion
    When x is a REAL or DOUBLE, returns NaN if all input values are NaN. Ignores NaN values otherwise.
```
Suggest comma as a break for readability, to separate the condition from the outcome.

hainenber:
Hi there, this PR is free for other folks taking over. Thanks for your effort reviewing previously!

steveburnett (approved):
LGTM! (doc)

Read through the comments and discussion, pull branch, local build, everything looks good. Thanks!

rschlussel:
I see the logic behind this change, as it seems to achieve what the original function was trying to do.  However, I agree with @spershin that we should wait to fix the behavior for min and max until we have a plan for how we want to handle NaNs across the board.  This is a long standing issue, and given that we expect to have a plan for NaNs imminently, it doesn't seem worth changing the behavior for these functions 2x.
Also, the behavior here almost definitely will not be the final behavior we want for these function once we have consistent handling for NaNs.  (we would either consider nans as largest or would return nan if there are any nans in the input).

tdcmeehan:
@rschlussel while I agree with your sentiment, I still find this bug fix to be a relatively minor change in scope and an obvious bug fix.  Just in case it wasn't clear, consider once again what it would take to document the existing behavior:

> Returns the maximum value of all input values.
> For double and float types, NaN values are ignored, except if the first value is NaN, then the function will always return NaN.

After this change, the documentation would look like this:

> Returns the maximum value of all input values.
> For double and float types, NaN values are ignored.

Don't you find this to be a small, sensible change that may improve the experience of Presto users, that doesn't hold up our larger fixes related to NaN handling?

rschlussel:
That's fair. Yes, I agree the old behavior is very clearly a bug.  And given that someone has taken the time to fix it, you're right, we should merge it.  However, i think it's also true that in practice we haven't been providing any guarantees to our users regarding treatment of NaNs, and I don't think it's worth investing in fixing other similar bugs until we move toward a more cohesive solution.

presto-main/src/test/java/com/facebook/presto/operator/aggregation/TestDoubleMinAggregation.java
```diff
@@ -16,15 +16,50 @@
...
+    @BeforeClass
+    public void setup()
+    {
+        FunctionAndTypeManager functionAndTypeManager = MetadataManager.createTestMetadataManager().getFunctionAndTypeManager();
+        minFunction = functionAndTypeManager.getJavaAggregateFunctionImplementation(
+                functionAndTypeManager.lookupFunction(TestDoubleMinAggregation.FUNCTION_NAME, fromTypes(DOUBLE)));
+    }
+
+    @Test
+    public void min()
```
rschlussel, line 51: call this testMin()

presto-main/src/test/java/com/facebook/presto/operator/aggregation/TestDoubleMaxAggregation.java
```diff
@@ -16,15 +16,50 @@
...
+    @BeforeClass
+    public void setup()
+    {
+        FunctionAndTypeManager functionAndTypeManager = MetadataManager.createTestMetadataManager().getFunctionAndTypeManager();
+        maxFunction = functionAndTypeManager.getJavaAggregateFunctionImplementation(
+                functionAndTypeManager.lookupFunction(TestDoubleMaxAggregation.FUNCTION_NAME, fromTypes(DOUBLE)));
+    }
+
+    @Test
+    public void max()
```
rschlussel, line 51: call this testMax()

presto-main/src/test/java/com/facebook/presto/operator/aggregation/TestDoubleMaxAggregation.java
```diff
@@ -16,15 +16,50 @@
...
+import static com.facebook.presto.operator.aggregation.AggregationTestUtils.assertAggregation;
+import static com.facebook.presto.sql.analyzer.TypeSignatureProvider.fromTypes;
+import static java.lang.Double.NEGATIVE_INFINITY;
+import static java.lang.Double.NaN;
+import static java.lang.Double.POSITIVE_INFINITY;
 
 public class TestDoubleMaxAggregation
         extends AbstractTestAggregationFunction
 {
+    private static final String FUNCTION_NAME = "max";
```
rschlussel, line 39: we shouldn't need this. this class already has a get_function_name() method.

presto-main/src/test/java/com/facebook/presto/operator/aggregation/TestDoubleMaxAggregation.java
```diff
@@ -16,15 +16,50 @@
...
+    private JavaAggregationFunctionImplementation maxFunction;
+
+    @BeforeClass
+    public void setup()
+    {
+        FunctionAndTypeManager functionAndTypeManager = MetadataManager.createTestMetadataManager().getFunctionAndTypeManager();
```
rschlussel, line 45: AbstractTestAggregationFunction (which this extends) already has a function and type manager and a method getFunction() for getting the function implementation using that. Can we use that and just inline the call to getFunction() in the test instead of adding a setup?

rschlussel (commented):
same comments about TestDoubleMinAggregation as I put for TestDoubleMaxAggregation.

tdcmeehan:
@rschlussel should this be closed in favor of #22386?

rschlussel:
yes, let's close this one

Issue #20572, opened 2023-08-15
Global Row Level Security Support for Presto


hmadison:
## Expected Behavior or Use Case
Presto should have an interface which allows end users to inject predicates to provide "Row Level Security".

## Presto Component, Service, or Connector
To the best of my understanding, this would be a coordinator change.

## Possible Implementation
We could update the ConnectionManager to install additional logical plan optimizers which would be provided by a new SPI.

```java
    interface SecurityPlanProvider {
    /**
     * Provides a {@link ConnectorPlanOptimizer} which is able to modify the query on the coordinator
     * before it is submitted to any worker node. This allows for any additional filtering based on the user
     * session to be added.
     * 
     * @param targetConnectorId The identifier of the connector which the optimizer would be applied to.
     * @return A logical plan optimizer which.
     */
    Optional<ConnectorPlanOptimizer> getSecurityPlanOptimizer(ConnectorId targetConnectorId);
}
```

The provider optimizers would be installed before any connector provider optimizers in the ConnectionManager allowing the connector provider optimizers the ability to work on any additional input that the “security optimizer” would provide. 

## Context

“Row level security” has a number of open, stale or closed issues tracking its implementation inside of Presto:
- https://github.com/prestodb/presto/pull/15519
- https://github.com/prestodb/presto/issues/13626
- https://github.com/prestodb/presto/pull/12701
- https://github.com/prestodb/presto/issues/12491
- https://github.com/prestodb/presto/issues/8980
- https://github.com/prestodb/presto/pull/15519

While there has been work in this area, there isn’t a wholistic approach which covers an entire deployment. 

Unless you are using Hive and are able to take advantage of the Apache Ranger support, the “best” interface to cover a “Row Level Security” use case is ConnectorPlanOptimizer. The downside of this approach is that the only way to add new instances of ConnectorPlanOptimizer to inject the additional clauses to subset the data being queried is to provide a completely new adapter to Presto. 

While this approach would work for new connectors, it doesn’t offer the best user experience when attempting to make use of this interface for existing connectors. The only three methods which I’m aware of would be to use “ByteBuddy” style class rewrites, wrap the existing connector or maintain a fork of the connector code.

tdcmeehan:
Is there an example of any current system that could be leveraged to implement this row level security?

imjalpreet:
There is another PR that was based on how Trino has implemented Row Filtering: https://github.com/prestodb/presto/pull/16955

[...]

PR #21913, opened 2024-02-13
Add support for row filtering

JaneHang9305:
Cherry-pick of https://github.com/trinodb/trino/commit/7efb49ca76782ef83b347b8d51ea422057bca1e0
Co-authored-by: [Martin Traverso](https://github.com/martint)

## Description

SystemAccessControl/AccessControl to provide filters/projections on the base table given an authorization identifier (user/role) that the query planner injects into the plan.

## Motivation and Context

Presto should support row-level filtering and data masking for more granular access controls to data.
https://github.com/prestodb/presto/issues/20572
https://github.com/trinodb/trino/issues/1480

## Impact

There is no performance impact on users unless onboarding the rules.

## Test Plan

Test on local.

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* ...
* ...

Hive Changes
* ...
* ...
```

```
== NO RELEASE NOTE ==
```


jaystarshot:
@JaneHang9305  You need to add a coauthor in the commit ref- https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#attribution.
@tdcmeehan @imjalpreet @agrawalreetika This has been a stale issue since a while now. Should we just discuss this trino approach in a doc/wg meeting and proceed with merging/path forward?

tdcmeehan:
It seems like this filter is specified on a per-connector level.  If that's the case, why not just use a connector rewrite to provide the filter?

JaneHang9305:
> It seems like this filter is specified on a per-connector level. If that's the case, why not just use a connector rewrite to provide the filter?

This is a unified solution for all connectors, and each connector can have its own rules for filtering.

tdcmeehan:
Could we achieve a unified solution through a library, which provides a connector rewrite rule that the connector can add in to add support for row level filtering?

Originally I thought an approach like this might be a good idea, however I think while there are costs to the connector rewrite framework, it also has enormous benefits in the flexibility it allows and the ability to make changes that are just cumbersome to do purely in the engine.  So I'm wondering, if the primary motivation is to use one unified approach, can this be accomplished through composition--connectors compose a library which adds a rewrite rule which provides a rule level filter?  That way they can be shared among Hive, Hudi, Iceberg and other connectors easily.

tdcmeehan:
The same argument from a different angle: in what way is our existing SPI footprint insufficient to allow for connector-specified row level filtering?  Row level filtering is [listed as one of the motivations for connector rewrites](https://prestodb.io/blog/2019/12/23/improve-presto-planner/)--in what way does this improve on that?  If it's a narrower API footprint, can this be achieved through composition as mentioned above?

jaystarshot:
I do agree that it would be better if we have a smaller SPI footprint that would be better. However, I don't think the addition of the filter itself is not connector specific but rather the rule level filters that connectors will provide which are needed connector specific. It's like ConnectorIdentity which is connector specific. 
Furthermore, we already have a package for security `com.facebook.presto.spi.security` and it might be better to centralize all security-connector updates from one location.

tdcmeehan:
Let me know if I'm understanding this correctly.

For a table scan, during analysis, we we check the table we want to filter, and call an SPI hook on the Connector to ask for its row filters.  Then we'll process and validate these filters, and put them back into the analysis.  Then, during planning, we'll check if any row level filters were specified, and add them to as filters to the table scan node.  Filtering accomplished.

It seems like one argument is this is the place where it ought to go, because it's security related, and security related things ought to be centralized in one place.  But I would argue that connector plan rewrites are a conscious choice about splitting where things ought to go: certain things are global, and ought to be centralized in the engine, and certain things can only be inferred by the context of the connector, and can be pushed down entirely to the connector.  An example is advanced filter pushdown: while this is common in Hive-like systems, other systems treat this all very differently, and there's not a whole lot of benefit for centralizing this in Presto.  Likewise, because the rules which specify what rows to filter out are connector-specific, I just don't see the benefit to centralizing this and adding a larger SPI footprint when we already have the ability to do this entirely in the connector.

jaystarshot:
Yes that's correct, however this is not an advanced filter pushdown where we need any connector-specific info regarding the tableScans. I think this pushdown could be handled in presto-main itself. Replicating the pushdown across all connectors might simply lead to code duplication. Additionally, it might make it challenging for administrators to implement central access policies across all connectors/catalogs

tdcmeehan:
I don't understand how this PR solves the problem of making central access policies less challenging.  The policies must come from the connector, correct?

jaystarshot:
Yes but they follow the same interface I guess? Also since the filter logic is in presto-main now (lets say to avoid duplication) we would need something in the spi for the connector to communicate with the engine?

jaystarshot:
So i guess the question is whether this filter pushdown deserves its own connector specific optimizers or it this perfectly fine to avoid duplication

tdcmeehan:
Let's consider the approaches by analogy.  By analogy, the approach in this PR is like object-oriented [inheritance](https://en.wikipedia.org/wiki/Inheritance_(object-oriented_programming)).  There is base functionality, which is applying the filters, but there are abstract methods, which are the SPI hooks.  Similar to object oriented inheritance, you must know how the parent class uses the abstract methods in order to implement them effectively.

Let us consider the connector rewrite rule approach to be more like [composition](https://en.wikipedia.org/wiki/Object_composition).  In this approach, you need not know the parent interface rules.  Instead, you compose a new behavior, which defines privacy filters based on arbitrary criteria, and pushes them to table scan nodes.

The outcome seems the same between both.  Because we've embarked on the composition path, I am wary of doing this through the inheritance path, since we already have the ability to do this with composition.  The code reuse argument seems to be able to be solved by just making the rewrite a library that connectors mix-in/compose to existing rewrites.

jaystarshot:
Sounds good, yeah connector optimizers will be the better approach for presto. If they need to be "improved" like for eg: _connector optimizers_ for relational plan, that functionality can be added i guess

Issue #21731, opened 2024-01-19
Query 20240119_063836_00856_pugph failed: Unsupported column type: timestamp


JxKim:
I was trying to use MySql connector and inserting datas which are results selected from other tables into the mysql table.
There is a column with datetime type in the dest table. When I executed the 'insert into select' query, an exception occurred.
Now I can only create a tmp table in MySql and use it as a bridge. Why timestamp not supported yet???

* Presto version used:0.280-A95C1B4
* Storage (HDFS/S3/GCS..): 
* Data source and connector used: mysql connector

## Steps to Reproduce

1.
2.
3.
4.

## Screenshots (if appropriate)
![Uploading image.png…]()

yhwang:
Can you provide the stack trace and SQL you are using? I guess you tried to upload a screenshot but the link just points back to this issue.

Akanksha-kedia:
@JxKim According to me, what i understood from mentioned above "The issue you're encountering is likely due to the way Presto handles TIMESTAMP data types. Presto has its own internal representation for TIMESTAMP, and it may not always map perfectly to the TIMESTAMP or DATETIME types in MySQL.

As of now, the Presto MySQL connector only supports a subset of MySQL's data types, and the support for TIMESTAMP is limited. This is a known limitation and is documented in the Presto's official documentation.

As a workaround, you could convert the TIMESTAMP columns to a string format that Presto can handle, perform your operations, and then convert them back to TIMESTAMP in MySQL if necessary. You could also consider using a different connector that fully supports TIMESTAMP, like the PostgreSQL connector, if switching databases is an option for you."

JxKim:
> @JxKim According to me, what i understood from mentioned above "The issue you're encountering is likely due to the way Presto handles TIMESTAMP data types. Presto has its own internal representation for TIMESTAMP, and it may not always map perfectly to the TIMESTAMP or DATETIME types in MySQL.
> 
> As of now, the Presto MySQL connector only supports a subset of MySQL's data types, and the support for TIMESTAMP is limited. This is a known limitation and is documented in the Presto's official documentation.
> 
> As a workaround, you could convert the TIMESTAMP columns to a string format that Presto can handle, perform your operations, and then convert them back to TIMESTAMP in MySQL if necessary. You could also consider using a different connector that fully supports TIMESTAMP, like the PostgreSQL connector, if switching databases is an option for you."

why is mysql connector not supporting datetime mapping yet? Are there any technical reasons for this?

[...]

PR #21937, opened 2024-02-15
Support timestamp type in mysql jdbc connector

hantangwangd:
## Description

Fix issue #21731. Support column type: timestamp in mysql connector.

## Test Plan

 - Newly added test case in TestMySqlIntegrationSmokeTest.testMysqlTimestamp()

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

Mysql Changes
* Support timestamp column type
```

### DIFF
--- a/presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/JdbcPageSink.java
+++ b/presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/JdbcPageSink.java
@@ -17,6 +17,7 @@ import com.facebook.airlift.log.Logger;
 import com.facebook.presto.common.Page;
 import com.facebook.presto.common.block.Block;
 import com.facebook.presto.common.type.DecimalType;
+import com.facebook.presto.common.type.TimestampType;
 import com.facebook.presto.common.type.Type;
 import com.facebook.presto.common.type.UuidType;
 import com.facebook.presto.spi.ConnectorPageSink;
@@ -33,6 +34,8 @@ import java.sql.Date;
 import java.sql.PreparedStatement;
 import java.sql.SQLException;
 import java.sql.SQLNonTransientException;
+import java.sql.Timestamp;
+import java.time.Instant;
 import java.util.Collection;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
@@ -164,6 +167,13 @@ public class JdbcPageSink
             long utcMillis = DAYS.toMillis(type.getLong(block, position));
             long localMillis = getInstanceUTC().getZone().getMillisKeepLocal(DateTimeZone.getDefault(), utcMillis);
             statement.setDate(parameter, new Date(localMillis));
+        }
+        else if (type instanceof TimestampType) {
+            long timestampValue = type.getLong(block, position);
+            statement.setTimestamp(parameter,
+                    Timestamp.from(Instant.ofEpochSecond(
+                            ((TimestampType) type).getEpochSecond(timestampValue),
+                            ((TimestampType) type).getNanos(timestampValue))));
         }
         else if (UuidType.UUID.equals(type)) {
             Slice slice = type.getSlice(block, position);
--- a/presto-common/src/main/java/com/facebook/presto/common/type/TimestampType.java
+++ b/presto-common/src/main/java/com/facebook/presto/common/type/TimestampType.java
@@ -82,6 +82,30 @@ public final class TimestampType
         return Objects.hash(getClass(), precision);
     }
 
+    /**
+     * Gets the timestamp's number of total seconds.
+     * The epoch second count is a simple incrementing count of seconds where second 0 is 1970-01-01T00:00:00Z.
+     *
+     * Returns:
+     * the total seconds in timestamp
+     */
+    public long getEpochSecond(long timestamp)
+    {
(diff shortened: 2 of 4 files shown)


presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/JdbcPageSink.java
```diff
@@ -165,6 +169,22 @@ else if (DATE.equals(type)) {
             long localMillis = getInstanceUTC().getZone().getMillisKeepLocal(DateTimeZone.getDefault(), utcMillis);
             statement.setDate(parameter, new Date(localMillis));
         }
+        else if (type instanceof TimestampType) {
+            long timestampValue = type.getLong(block, position);
+            int nanos = ((TimestampType) type).getPrecision() == MILLISECONDS ?
```
yhwang, line 174: Just thinking would it make sense to move `get nanos` and `get millis` to `com.facebook.presto.common.type.TimestampType` although this would be the only place to use them.

presto-common/src/main/java/com/facebook/presto/common/type/TimestampType.java
```diff
@@ -82,6 +82,33 @@ public int hashCode()
...
+                MICROSECONDS.toSeconds(timestamp);
+    }
+
+    /**
+     * Gets the timestamp's nanosecond portion.
+     *
+     * Returns:
+     * this timestamp's fractional seconds component
+     */
+    public int getNanos(long timestamp)
```
yhwang, line 105: just curious that any reason you return `int` here instead of `long`. I saw the API for `Instant.ofEpochSecond` is `(long, long)`.

Another thing I found is that `java.util.concurrent.TimeUnit` provides `MICROSECONDS.toNanos(long)` and `MILLISECONDS.toNanos(long)` not sure if that's relevant.

hantangwangd: > just curious that any reason you return `int` here instead of `long`. I saw the API for `Instant.ofEpochSecond` is `(long, long)`.

I consulted the API definition of `java.time.Instant`, `int` type would be enough as the return value of `getNanos` would always less than 1000,000,000. What's your opinion?

> Another thing I found is that `java.util.concurrent.TimeUnit` provides `MICROSECONDS.toNanos(long)` and `MILLISECONDS.toNanos(long)` not sure if that's relevant.

Thanks, that could be useful. Consulting the implementation in `DecodeTimestampOptions`, I have modified the code to use them. Please take a look!

hantangwangd: By the way, seems the second parameter of `Instant.ofEpochSecond(long, long)` could be nanos representing longer than one second, so its type was declared as `long`.

presto-common/src/main/java/com/facebook/presto/common/type/TimestampType.java
```diff
@@ -82,6 +82,32 @@ public int hashCode()
...
+     * The epoch second count is a simple incrementing count of seconds where second 0 is 1970-01-01T00:00:00Z.
+     *
+     * Returns:
+     * the total seconds from the epoch of 1970-01-01T00:00:00Z in timestamp
+     */
+    public long getEpochSecond(long timestamp)
+    {
+        return this.precision == MILLISECONDS ?
+                MILLISECONDS.toSeconds(timestamp) :
+                MICROSECONDS.toSeconds(timestamp);
```
yhwang, line 96: How about `this.precision.toSeconds(timestamp)`?
If that's the case, I start to wonder do we need this function. sorry for the hassle.

presto-common/src/main/java/com/facebook/presto/common/type/TimestampType.java
yhwang: Thanks for the explanation. I believe for the current usage, it only reduces the stack size but not the heap because `java.sql.Timestamp` is still using `long` and consumes the same amount of heap size.

However, if this method is used by other code and the constructed object is using int, then it does reduce the heap usage. You can leave it as `int`.

presto-common/src/main/java/com/facebook/presto/common/type/TimestampType.java
hantangwangd: > How about `this.precision.toSeconds(timestamp)`?

Oh, it's better, I have fixed it. Thanks a lot for your suggestion that lead to the continuous improvement in our implementation.

I think this function still has some value to provide the utility logic and corresponding comment, for the usage `TimestampType.getPrecision().toSeconds(long)` outside is not so straightforward. What do you think?

yhwang: sure! sold.

yhwang (approved):

kiersten-stokes (approved):
LGTM!

tdcmeehan (approved):

hantangwangd:
> It looks like this PR only supports milliseconds precision for timestamp type. I will raise a github issue and link with this PR.

@pratyakshsharma Sure, thank you for pointing out the problems if there is any. I think this change can internally support microseconds precision as well. So please describe the problematic scenario in the new issue.

PR #21941, opened 2024-02-15
[native] Add missing task runtime stats 

mbasmanova:
* Some operator runtime stats were missing from task runtime stats. These include numSplits, inputBatches, outputBatches and numMemoryAllocations .
* createTime task runtime stat was reported only after task has finished.
* elapsedTimeInNanos stat was reported only after task has finished.

```
== NO RELEASE NOTE ==
```


majetideepak (approved):
Thanks for the cleanup, Masha!

PR #10259, opened 2018-03-26
Expose "$partitions" system tables for partitioned Hive tables

losipiuk:
@electrum, @findepi 
**Please just do high level review to validate if approach is generally OK**

This is an try to workaround the problem with listing Hive partitions when there are more of them than configuration set hard limit.
We decided that want to try approach when for each partitioned Hive table we expose a counterpart system table (suffixed with `$partitions`) which 
will expose all the partitioning keys.

This PR achieves that by exposing new `Optional<SystemTable> ConnectorMetadata.getSystemTable()` method.
The method is then plugged to per-connector SystemMemory connector via `SystemTablesProvider` (which used to be just `List<SystemTable>` before).
The `SystemTablesProvider` is responsible for listing and resolving system tables.
The used implementation is provided with `List<SystemTable>` (for sake of static system tables) and `Metadata` which is used to delegate calls to `ConnectorMetadata.getSystemTable` for dynamic ones.

The approach generally seems to work but please let me know what you think about it.

One limitation I see with current implementation is that we are only listing the static system tables. This leads to some weirdnesses. E.g you will get empty table when query `SHOW COLUMNS FROM "table$partitions"`, because it is rewritten to `information_schema` query which depends on fact that queried tables are listed via `ConnectorMedata.listTables`. 
Potentially this can be extended by adding `ConnectorMetadata.listSystemTables` method. Yet for Hive we cannot cheaply test if table is partitioned or not (I think).
And without that check it gets problematic if the `$partitions` table is queried for table which is not actually partitioned as we do not have any columns to return. But generally it should be doable - we just need to think about what is exact expected behaviour.

Edit: addressed limitation with `Use SystemTablesProvider.getSystemTable in listTableColumns` commit.

PR #6751, opened 2016-12-02
Optimize Hive partition filtering

electrum:
Unfortunately, the partitions limit breaks `SHOW PARTITIONS` even with filters due to #6753.

PR #21947, opened 2024-02-15
Refactor code in HivePartitionManager

NikhilCollooru:
## Description
In HivePartitionManager, we are returning the iterator of the list of partitions only to iterate and create a list again. This can be costly when the list of partitions is huge. list --> iterator --> list conversion is unnecessary and we can just return the list directly.

In meta internal testing we observed that this conversion is taking ~300ms when the partitions list size is ~4k. 
20240215_185854_00006_cabkz

## Impact
None

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-hive/src/main/java/com/facebook/presto/hive/HivePartitionManager.java
+++ b/presto-hive/src/main/java/com/facebook/presto/hive/HivePartitionManager.java
@@ -50,7 +50,7 @@ import org.weakref.jmx.Nested;
 import javax.inject.Inject;
 
 import java.util.ArrayList;
-import java.util.Iterator;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -142,7 +142,7 @@ public class HivePartitionManager
         this.executorServiceMBean = new ThreadPoolExecutorMBean((ThreadPoolExecutor) threadPoolExecutor);
     }
 
-    public Iterable<HivePartition> getPartitionsIterator(
+    public List<HivePartition> getPartitionsList(
             SemiTransactionalHiveMetastore metastore,
             ConnectorTableHandle tableHandle,
             Constraint<ColumnHandle> constraint,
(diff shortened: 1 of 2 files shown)


jaystarshot (approved):

presto-hive/src/main/java/com/facebook/presto/hive/HivePartitionManager.java
```diff
@@ -360,24 +364,6 @@ private boolean queryAccessesTooManyBuckets(HiveBucketHandle handle, Optional<Hi
...
-                        "Query over table '%s' can potentially read more than %s partitions",
-                        partition.getTableName(),
-                        maxPartitionsPerScan));
-            }
-            partitionList.add(partition);
-            count++;
-        }
-        return partitionList.build();
-    }
-
```
ajaygeorge, line 380: @rschlussel  was this change introduced to lazily get the next partition and keep track of the count and fail if it is above maxPartitionsPerScan ?.

presto-hive/src/main/java/com/facebook/presto/hive/HivePartitionManager.java
```diff
@@ -263,7 +261,13 @@ public HivePartitionResult getPartitions(SemiTransactionalHiveMetastore metastor
...
         List<HiveColumnHandle> partitionColumns = getPartitionKeyColumnHandles(table);
 
-        List<HivePartition> partitions = getPartitionsAsList(getPartitionsIterator(metastore, tableHandle, constraint, session).iterator());
+        List<HivePartition> partitions = getPartitionsList(metastore, tableHandle, constraint, session);
+        if (partitions.size() > maxPartitionsPerScan) {
+            throw new PrestoException(HIVE_EXCEEDED_PARTITION_LIMIT, format(
+                    "Query over table '%s' can potentially read more than %s partitions",
+                    hiveTableHandle.getSchemaTableName().toString(),
+                    maxPartitionsPerScan));
+        }
```
ajaygeorge, line 270: don't we want to fail fast if we exceed maxPartitionsPerScan .
can we have a counter (Atomic) to keep track of count from getPartitionListFromPartitionNames and fail if it is more than maxPartitionsPerScan

presto-hive/src/main/java/com/facebook/presto/hive/HivePartitionManager.java
rschlussel: so, i didn't introduce this code i just moved it (original changes came from here https://github.com/prestodb/presto/pull/6751/commits/47ae18c25350708eb5b9b6d4eb3fb8b50a66c359 and https://github.com/prestodb/presto/pull/10259/commits/59dcbd22706ed2eef798bf9dbb9baa08509f2b5e).  But, yeah, I think it's basically there for enforcing the maxPartitionsPerScan limit without having to enumerate all the partitions first in case it's really big.

ajaygeorge: Thanks Rebecca. since we are not really passing in a Iterator rather grabbing the Iterator from the Iterable List from line 191 we have the entire size already available to us and this PR refactoring makes sense.

ajaygeorge (approved):
LGTM

presto-hive/src/main/java/com/facebook/presto/hive/HivePartitionManager.java
NikhilCollooru: As per our offline discussion, the suggestion was to do the check in the threads itself when they are doing the work parallely. The maxPartitionPerScan limit check should still happen outside of the parallel value parsing threads because not all users will enable this parallel-parsing-of-values feature. Also the number of queries hitting this limit will be few. So hence we decided to proceed with the code changes in the pr.

Issue #21886, opened 2024-02-08
Consolidate LegacySqlQueryScheduler and SqlQueryScheduler


arhimondr:
The difference between SqlQueryScheduler and LegacySqlQueryScheduler is in the former ability to support full section retries in Presto Unlimited.  This was implemented before moving to Presto on Spark strategy that supports more granular retries through spark. The new scheduler has never been fully rolled out. 

The proposal is to remove the `SqlQueryScheduler` and rename `LegacySqlQueryScheduler` to `SqlQueryScheduler` dropping related properties and interfaces

arhimondr:
CC: @rschlussel @shrinidhijoshi

PR #14879, opened 2020-07-23
Fix infinite loop in non-legacy SqlQueryScheduler

rschlussel:
Queries that don't have table scan inputs could enter an infinite loop inthe scheduler if they hit the queryStateMachine.isDone() check and had any sections ready for execution. If the query is done (e.g. canceled or abandoned at this point), we abort the section.  Since "aborted" is a failed state, the section would continue to be considered "ready for execution"  in our next time through, and the loop would continue indefinitely. We need to explicitly exit the scheduling loop if the query is finished.

Queries with inputs didn't have this issue because the split scheduler would be closed when the query finished, so the scheduler would hit an error when it tried to create the scheduler for the scan stages.

I don't have a test because I was only able to reproduce the issue by explicitly calling queryStateMachine.transitionToFailed() here so we would enter the if statement.  Recommendations about how to test this are appreciated.

```
== RELEASE NOTES ==
General Changes
* Fix potential infinite loop when the setting ``use_legacy_scheduler`` is set to false.
```

PR #14264, opened 2020-03-18
Set use-legacy-scheduler true by default

rschlussel:
There was a regression with the refactored scheduler, that it could
cause high cpu usage on the coordinator.  Use legacy scheduler by
default until that's fixed.

```
== RELEASE NOTES ==

General Changes
* Change the default for configuration property ``use-legacy-scheduler`` to ``true`` in order to mitigate a regression in the new scheduler.

```

PR #21952, opened 2024-02-16
Remove the newer SqlQueryScheduler

rschlussel:
Remove the "new" SqlQueryScheduler, which was created to support stage retries, but was never rolled out and is no longer important because of Presto-on-Spark. Rename LegacySqlQueryScheduler to SqlQueryScheduler.

## Description
Remove the SqlQueryScheduler and rename LegacySqlQueryScheduler back to SqlQueryScheduler.  Remove corresponding session properties that are no longer relevant.

## Motivation and Context
 The new SqlQueryScheduler, which was built to enable stage retries when materializing exchanges has never been rolled out, and has a bug that causes it to use a lot of cpu that has not been resolved (was disabled here,  https://github.com/prestodb/presto/pull/14264 and an attempt to fix it here https://github.com/prestodb/presto/pull/14879 did not resolve the issue).

The new scheduler was deprioritized with the investment in presto-on-spark.  This PR removes the "new" scheduler, which has been tech debt for a while since we are not investing in fixing it and making it stable for production.

The LegacySqlQueryScheduler has been the default, so there is no user facing change except for removing the ability to try out the new scheduler.

Fixes https://github.com/prestodb/presto/issues/21886

## Impact
Removes ability to use the experimental SqlQueryScheduler that allowed for stage retries

## Test Plan
CI

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Remove the configuration property ``use-legacy-scheduler`` and the corresponding session property ``use_legacy_scheduler`.   The property previously defaulted to true, and the new scheduler, which was intended to replace it eventually, was never productionized and is no longer needed. The configuration property ``max-stage-retries`` and the session property ``max_stage_retries`` have also been removed.

```


kaikalur:
If it is a useful/functional piece we should keep it. Who knows maybe we will find use for it!

rschlussel:
> If it is a useful/functional piece we should keep it. Who knows maybe we will find use for it!

The problem is that it's buggy and it's not clear how much effort it would take to be confident enough to use it in production.  If the feature were important for us now, i would 100% say let's keep it and take it that last mile. But it's not really a priority, so in the meantime it's tech debt that makes it harder to add features to the scheduler (you need to add everything in two places in case one day someone decides we should actually productionize and use the new scheduler).

tdcmeehan:
If it comes in handy later, we can simply recover it from Git history.  I agree with @rschlussel that there's costs to maintaining this class, and while for this one class it might be small, dead code over the whole project it is a significant technical burden and I think we should err on the side of removing unused functionality.

presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
```diff
@@ -1667,16 +1655,16 @@ public SystemSessionProperties(
...
-                        "d. Print VLOG(0) and lower messages from elsewhere",
+                                "property is to control logging levels using glog flags. E,g, to enable verbose mode, add " +
+                                "'--v 1'. More advanced glog gflags usage can be found at " +
+                                "https://rpg.ifi.uzh.ch/docs/glog.html\n" +
+                                "e.g. --vmodule=mapreduce=2,file=1,gfs*=3 --v=0\n" +
+                                "will:\n" +
+                                "\n" +
+                                "a. Print VLOG(2) and lower messages from mapreduce.{h,cc}\n" +
+                                "b. Print VLOG(1) and lower messages from file.{h,cc}\n" +
+                                "c. Print VLOG(3) and lower messages from files prefixed with \"gfs\"\n" +
```
ajaygeorge, line 1666: nit. stray spaces?

ajaygeorge (dismissed):
LGTM. Please take a look at the failing tests as well.

steveburnett:
I checked the documentation and none of these properties appear anywhere in presto-docs other than the [0.232](http://prestodb.io/docs/current/release/release-0.232.html) and [0.239](http://prestodb.io/docs/current/release/release-0.239.html) release notes, so no doc updates needed.

ajaygeorge (approved):
LGTM

PR #21954, opened 2024-02-16
Fix variable width encoding description in serialized-page.rst

kevinwilfong:
The bullet points under "VARIABLE_WIDTH Encoding" list "total number of bytes in all values" as 4 bytes, but in the descriptive paragraph below it says 1 byte.  I checked the code and the correct value is 4 bytes.

I updated the text, but the image is wrong as well, I don't know how to regenerate it.

### DIFF
--- a/presto-docs/src/main/sphinx/develop/serialized-page.rst
+++ b/presto-docs/src/main/sphinx/develop/serialized-page.rst
@@ -152,7 +152,7 @@ Let’s again take the example from the Null Flags section and say that we
 have a string column with 10 rows with nulls in zero-based rows 1, 4, 6, 7, 9.
 The non-null rows will have values: 0 - Denali, 2 - Reinier, 3 - Whitney,
 5 - Bona, 8 - Bear. We’ll have 4 bytes storing the number of rows: 10, followed
-by 40 bytes of offsets, followed by 3 bytes of null flags, followed by 1 bytes
+by 40 bytes of offsets, followed by 3 bytes of null flags, followed by 4 bytes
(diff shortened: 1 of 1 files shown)


tdcmeehan (dismissed):

tdcmeehan:
@mbasmanova it would be nice to fix the image as well--can you help @kevinwilfong with how you generated it?

ajaygeorge (approved):

steveburnett (dismissed):
LGTM! (docs) 

Pull branch, local build, text changes look good. The graphic, as mentioned already, also needs to be updated, but the text changes are good now.

kevinwilfong:
thanks! I'll reach out to @mbasmanova

mbasmanova:
Usually, I create graphics using Power Point. It would be nice to come up with some way to check in these PowerPoint files along the documentation.

tdcmeehan:
Perhaps we can use Omnigraffle, which stores the graphic as code?  Here's an example for T-Digest: https://github.com/prestodb/presto/tree/b463cd0b9ae9e499515fe186bc3f1271b2644d5c/presto-main/src/main/java/com/facebook/presto/tdigest/docs

tdcmeehan:
Omnigraffle is not an open tool.  We can begin to use a new tool to generate these diagrams which is in an open format.  But I feel it's not acceptable to just change the text without updating the image.

elharo (approved):
I think we should merge this as is and file an issue for the image. I'd rather have correct text and incorrect image than incorrect text and incorrect image. Let's not let the perfect be the enemy of the good.

PR #21956, opened 2024-02-17
Add config for default view security mode

jainxrohit:
## Description

Presto has a default view creation security mode as 'DEFINER'. However, some administrators may prefer to use 'INVOKER' as the default.

To address this, a new configuration flag `default-view-security-mode` has been added that allows the default view security mode to be changed according to the administrator's preference.

```
== RELEASE NOTES ==

General Changes
* Add a config `default-view-security-mode` to choose the default security mode for view creation.

```

### DIFF
--- a/presto-docs/src/main/sphinx/sql/create-view.rst
+++ b/presto-docs/src/main/sphinx/sql/create-view.rst
@@ -38,6 +38,9 @@ In the ``INVOKER`` security mode, tables referenced in the view are
 accessed using the permissions of the query user (the *invoker* of the
 view). A view created in this mode is simply a stored query.
 
+The ``default-view-security-mode`` can be used to configure the default
+security mode for view creation.
+
 Examples
 --------
 
@@ -59,6 +62,10 @@ Create a view that replaces an existing view::
     CREATE OR REPLACE VIEW test AS
     SELECT orderkey, orderstatus, totalprice / 4 AS quarter
     FROM orders
+
+Set the default view security mode to ``INVOKER``::
+
+    SET SESSION default_view_security_mode='INVOKER'
 
 See Also
 --------
--- a/presto-main/src/main/java/com/facebook/presto/execution/CreateViewTask.java
+++ b/presto-main/src/main/java/com/facebook/presto/execution/CreateViewTask.java
@@ -37,6 +37,7 @@ import javax.inject.Inject;
 import java.util.List;
 import java.util.Optional;
 
+import static com.facebook.presto.SystemSessionProperties.getDefaultViewSecurityMode;
 import static com.facebook.presto.metadata.MetadataUtil.createQualifiedObjectName;
 import static com.facebook.presto.metadata.MetadataUtil.toSchemaTableName;
 import static com.facebook.presto.spi.analyzer.ViewDefinition.ViewColumn;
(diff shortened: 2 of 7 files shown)


steveburnett:
Consider adding documentation for this new configuration flag. Would [Properties Reference](http://prestodb.io/docs/current/admin/properties.html) be a good place for it?

rschlussel:
> Consider adding documentation for this new configuration flag. Would [Properties Reference](http://prestodb.io/docs/current/admin/properties.html) be a good place for it?

https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/sql/create-view.rst#security would be a good place also.

rschlussel (commented):
Not needed for this PR, but one other aspect of the security mode stuff that would be nice to improve is that when you run SHOW CREATE VIEW it doesn't show you the security mode that it was created with.  It makes it hard to validate that any particular view is using the right security mode.

presto-main/src/main/java/com/facebook/presto/execution/CreateViewTask.java
```diff
@@ -97,9 +98,10 @@ public ListenableFuture<?> execute(CreateView statement, TransactionManager tran
                 .collect(toImmutableList());
 
         ConnectorTableMetadata viewMetadata = new ConnectorTableMetadata(toSchemaTableName(name), columnMetadata);
-        // use DEFINER security by default
+
+        CreateView.Security defaultSecurityMode = getDefaultViewSecurityMode(session);
```
NikhilCollooru, line 102: nit: `defaultSecurityMode` -> `defaultViewSecurityMode`

NikhilCollooru (approved):

presto-docs/src/main/sphinx/sql/create-view.rst
```diff
@@ -38,6 +38,9 @@ In the ``INVOKER`` security mode, tables referenced in the view are
 accessed using the permissions of the query user (the *invoker* of the
 view). A view created in this mode is simply a stored query.
 
+The ``default-view-security-mode`` can be used to configure the default
+security mode for view creation.
+
```
steveburnett, line 43: Consider adding an example that shows how to use ``default-view-security-mode``.

steveburnett (changes requested):
Thanks for the documentation!

presto-docs/src/main/sphinx/sql/create-view.rst
jainxrohit: @steveburnett It can be used as any other session property. Do you have anything specific in mind?

presto-docs/src/main/sphinx/sql/create-view.rst
```diff
@@ -60,6 +63,10 @@ Create a view that replaces an existing view::
     SELECT orderkey, orderstatus, totalprice / 4 AS quarter
     FROM orders
 
+
+Set the default view security mode to ``INVOKER``
```
steveburnett, line 67:
```suggestion
Set the default view security mode to ``INVOKER``::

```
Formatting suggestion to make example a code block.

presto-docs/src/main/sphinx/sql/create-view.rst
steveburnett: The example you added works, thanks!

steveburnett (changes requested):
Thanks for the code example! Just one nit about formatting the code example. Here's what it looks like currently in a local build of the doc: 
Screenshot 2024-02-23 at 12 44 42 PM

rschlussel (approved):

Issue #7122, opened 2017-01-18
TIMESTAMP behaviour does not match sql standard


losipiuk:
It seems like meaning of `TIMESTAMP` and `TIMESTAMP WITH TIMEZONE` datatypes in Presto is totally not what is specified by SQL standard (and what other databases do).

This is my understanding of SQL 2003 standard (4.6.2 Datetimes):

`TIMESTAMP WITH TIMEZONE` represents absolute point in time. Typically databases store it internally as seconds since epoch in some fixed timezone (usually UTC). When querying `TIMESTAMP WITH TIMEZONE` data the values are presented to user in session timezone (yet session timezone is used just for presentation purposes).

`TIMESTAMP` does not represent specific point in time, but rather a reading of a wall clock+calendar. Selecting values from `TIMESTAMP` column should return same result set no matter what is the client's session timezone.

While Presto semantics is different:

`TIMESTAMP` seems to do what `TIMESTAMP WITH TIMEZONE` should.

`TIMESTAMP WITH TIMEZONE` encodes explicit timezone information to each value stored in table. The sql standard does not define a type like that. But it does not seem very practical. Assuming that values selected from `TIMESTAMP WITH TIMEZONE` are presented to user in session timezone anyway, the per-row timezone information can be stripped away and all values can be stored in some arbitrary fixed timezone (e.g. UTC).

Please comment on the semantics. It seems wrong. Why the choice - as it is hard to believe that it was not done intentionally.

cc: @dain, @martint, @fiedukow, @cawallin 

## Edit: roadmap: https://github.com/prestodb/presto/issues/10326

martint:
Indeed. We've been talking about this for a while in our team. It was due to misinterpretation of the spec.

We need to fix it, but haven't had the time to figure out what needs to be updated and the impact.

findepi:
> the per-row timezone information can be stripped away and all values can be stored in some arbitrary fixed timezone (e.g. UTC).

If there is/will be an `extract(hour from ts)` equivalent, it must be treat the `ts` using the right TZ.

losipiuk:
We are happy to do the work. We just have settle how to handle the backward incompatibility this change implies. 
Do we we want to have some transition-phase when old semantics can be enforced with configuration property (I guess it would be tricky and ugly to implement).
Or we can just change the semantics of types and do not care about backward compatibility.

martint:
Ideally, yes, but it depends on how ugly it gets.

losipiuk:
[...]

PR #21959, opened 2024-02-19
[Iceberg]Support year/month/day/hour transforms both on legacy and non-legacy TimestampType column

hantangwangd:
## Description

When we create partitions on TimestampType columns(using year/month/day/hour transform) in Iceberg, we will meet various problems because of the legacy timestamp logic in presto engine do not match the logic used in Iceberg transform calculation. See issue: #7122 .

This PR fix the problems, and correct the behaviors for both legacy and non-legacy timestamp.

## Test Plan

 - Test year/month/day/hour transform on TimestampType column for various time zones in legacy timestamp mode
 - Test year/month/day/hour transform on TimestampType column in non-legacy timestamp mode

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

Iceberg Changes
* Support year/month/day/hour transforms both on legacy and non-legacy TimestampType column
* Fix the bug that `CAST` from non-legacy timestamp to date rounding to future when the timestamp is prior than `1970-01-01 00:00:00.000`
```

### DIFF
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergPageSink.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergPageSink.java
@@ -16,6 +16,8 @@ package com.facebook.presto.iceberg;
 import com.facebook.airlift.json.JsonCodec;
 import com.facebook.presto.common.Page;
 import com.facebook.presto.common.block.Block;
+import com.facebook.presto.common.block.BlockBuilder;
+import com.facebook.presto.common.function.SqlFunctionProperties;
 import com.facebook.presto.common.type.BigintType;
 import com.facebook.presto.common.type.BooleanType;
 import com.facebook.presto.common.type.DateType;
@@ -46,8 +48,11 @@ import org.apache.iceberg.PartitionField;
 import org.apache.iceberg.PartitionSpec;
 import org.apache.iceberg.Schema;
 import org.apache.iceberg.io.LocationProvider;
-import org.apache.iceberg.transforms.Transform;
 
+import java.time.Instant;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.time.ZoneOffset;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
@@ -72,6 +77,8 @@ import static java.util.Objects.requireNonNull;
 import static java.util.UUID.randomUUID;
 import static java.util.concurrent.CompletableFuture.completedFuture;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static java.util.concurrent.TimeUnit.NANOSECONDS;
+import static java.util.concurrent.TimeUnit.SECONDS;
 
 public class IcebergPageSink
         implements ConnectorPageSink
@@ -124,7 +131,9 @@ public class IcebergPageSink
         this.session = requireNonNull(session, "session is null");
         this.fileFormat = requireNonNull(fileFormat, "fileFormat is null");
         this.maxOpenWriters = maxOpenWriters;
-        this.pagePartitioner = new PagePartitioner(pageIndexerFactory, toPartitionColumns(inputColumns, partitionSpec));
+        this.pagePartitioner = new PagePartitioner(pageIndexerFactory,
+                toPartitionColumns(inputColumns, partitionSpec),
+                session);
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergPageSink.java
```diff
@@ -376,6 +378,23 @@ public static Object getIcebergValue(Block block, int position, Type type)
         throw new UnsupportedOperationException("Type not supported as partition column: " + type.getDisplayName());
     }
 
+    private static Object adjustForTransformCalculationIfNecessary(SqlFunctionProperties functionProperties, Type type, Object value)
```
ZacBlanco, line 381: nit: Should we rename this to something like `adjustTimestampForPartitionTransform`? Just to make it clearer that this is for timestamp-based transforms

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergPageSink.java
```diff
@@ -442,11 +466,34 @@ public int[] partitionPage(Page page)
             for (int i = 0; i < columns.size(); i++) {
                 PartitionColumn column = columns.get(i);
                 Block block = page.getBlock(column.getSourceChannel());
+                // adjust legacy timestamp value to compatible with Iceberg non-identity transform calculation
+                if (column.sourceType instanceof TimestampType && session.getSqlFunctionProperties().isLegacyTimestamp() && !column.getField().transform().isIdentity()) {
+                    TimestampType timestampType = (TimestampType) column.sourceType;
+                    BlockBuilder bb = timestampType.createBlockBuilder(null, block.getPositionCount());
```
ZacBlanco, line 472: nit: use full names instead of abbreviations

```suggestion
                    BlockBuilder blockBuilder = timestampType.createBlockBuilder(null, block.getPositionCount());
```

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergPageSink.java
```diff
@@ -442,11 +466,34 @@ public int[] partitionPage(Page page)
             for (int i = 0; i < columns.size(); i++) {
                 PartitionColumn column = columns.get(i);
                 Block block = page.getBlock(column.getSourceChannel());
+                // adjust legacy timestamp value to compatible with Iceberg non-identity transform calculation
+                if (column.sourceType instanceof TimestampType && session.getSqlFunctionProperties().isLegacyTimestamp() && !column.getField().transform().isIdentity()) {
```
ZacBlanco, line 470: I think this conditional here makes the code harder to read. Can we try factoring this out into some other function? Or at least factor out the logic which adjusts the timestamp into a separate function.

ZacBlanco (changes requested):
I think the main logic looks ok. Just a few nits and one question

presto-main/src/main/java/com/facebook/presto/type/TimestampOperators.java
```diff
@@ -134,7 +137,7 @@ public static long castToDate(SqlFunctionProperties properties, @SqlType(Standar
             long millis = date + chronology.getZone().getOffset(date);
             return TimeUnit.MILLISECONDS.toDays(millis);
         }
-        return TimeUnit.MILLISECONDS.toDays(value);
+        return floorDiv(value, MILLISECONDS_PER_DAY);
```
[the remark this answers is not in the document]
hantangwangd: If the milliseconds value is negative(prior than 1970-01-01), and do not exactly at midnight, `TimeUnit.MILLISECONDS.toDays(value)` would round forwards and `floorDiv(value, MILLISECONDS_PER_DAY)` would round backwards.

Take test case in `TestTimestamp.testCastFromTimestampToDate()` for example, `cast(timestamp '1969-12-31 05:06:07.234' as date)` would return `date '1970-01-01'` in case we use `TimeUnit.MILLISECONDS.toDays(value)`, and would return `date '1969-12-31'` in case we use `floorDiv(value, MILLISECONDS_PER_DAY)`.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergPageSink.java
hantangwangd: OK, fixed!

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergPageSink.java
hantangwangd: Sure, thanks for the suggestion, I have extract a separate function `adjustBlockIfNecessary` to strip this part of logic out.

presto-main/src/main/java/com/facebook/presto/type/TimestampOperators.java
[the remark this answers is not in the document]
ZacBlanco: Ah, I see. Because the due to the division in java, the standard truncating divide rounds forward when numerator is negative. floorDiv makes sure that we round down. Makes sense now. Thanks!

hantangwangd:
> Can you add to the release note an explanation of the bug fix on Timestamp to Date conversion?

Sure, it has been added! Please take a look, thanks!

gupteaj:
Changes looks good to me. In general, I see many references to the legacy system property, can we create a common function for these cases ?

ZacBlanco (dismissed):

hantangwangd:
> Changes looks good to me. In general, I see many references to the legacy system property, can we create a common function for these cases ?

@gupteaj Sure, good suggestion! Fixed!

tdcmeehan (approved):

PR #21964, opened 2024-02-19
[Iceberg] Reduce testTableDropWithMissingMetadata run time

ZacBlanco:
## Description

testTableDropWithMissingMetadata was taking > 4 minutes to run in
most instances due to the retry configuration inside of
HiveTableOperations.

This adds a configuration visible to the test infrastructure in order
to update the retry time during tests.

## Motivation and Context

Reduce CI time

## Impact

N/A

## Test Plan

N/A

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/HiveTableOperations.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/HiveTableOperations.java
@@ -28,6 +28,7 @@ import com.facebook.presto.spi.PrestoException;
 import com.facebook.presto.spi.SchemaTableName;
 import com.facebook.presto.spi.TableNotFoundException;
 import com.facebook.presto.spi.security.PrestoPrincipal;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
@@ -48,6 +49,7 @@ import org.apache.iceberg.util.Tasks;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.NotThreadSafe;
 
+import java.time.Duration;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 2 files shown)


presto-iceberg/src/main/java/com/facebook/presto/iceberg/HiveTableOperations.java
```diff
@@ -88,6 +91,9 @@ public class HiveTableOperations
     public static final String PREVIOUS_METADATA_LOCATION = "previous_metadata_location";
     private static final String METADATA_FOLDER_NAME = "metadata";
 
+    @VisibleForTesting
+    protected static Supplier<RetryConfig> defaultRetrySupplier = RetryConfig::new;
```
hantangwangd, line 95: Should we make `defaultRetrySupplier` thread local to ensure thread safety? As the tests may run concurrently, and we might add other tests which set `RetryConfig` as well.

hantangwangd (commented):
It's a very useful fix, just one little point for discussion.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/HiveTableOperations.java
ZacBlanco: Even if tests run concurrently, setting this as a thread-local probably won't solve the issue because we don't actually know which threads is going to end up creating the `HiveTableOperations`, since the query runner executes across a large number of threads.

kiersten-stokes (approved):
Great idea - thank you!

presto-iceberg/src/main/java/com/facebook/presto/iceberg/HiveTableOperations.java
hantangwangd: I'm not so sure but as I understand `HiveTableOperations ` is not a global attribute in query runner, and its lifecycle would not exceed a single transaction. So that the creating and using of `HiveTableOperations` could be thread safe. Its not important anymore if we plan to add the configs to IcebergConfig as @tdcmeehan and your issue #21967 has mentioned.

hantangwangd (approved):
LGTM!

presto-iceberg/src/main/java/com/facebook/presto/iceberg/HiveTableOperations.java
hantangwangd: Yes, that's exactly what I meant.

tdcmeehan:
> @tdcmeehan I filed this earlier: #21967
> 
> I just want to fix the test because it was causing some issue in my local build

When I walked backward from the code, I found it to be relatively straightforward to add the configs (even though it would require passing config parameters from several layers of dependencies).  My vote is to just do it the proper way.

ZacBlanco:
Closing in favor of the proper approach via whoever picks up #21967

PR #21969, opened 2024-02-21
Propagate PrestoException directly

NikhilCollooru:
## Description
If a PrestoException is returned by the underlying storage system implementation then directly return it.

## Motivation and Context
The Storage connector of Presto in Meta deployment is capable to return Presto exceptions directly. This PR is to directly return them, so that we can have our own meta specific error codes.

## Impact
None

## Test Plan
Existing tests

```
== NO RELEASE NOTE ==
```

The corresponding Meta internal PR 
https://github.com/facebookexternal/presto-facebook/pull/2624

### DIFF
--- a/presto-hive/src/main/java/com/facebook/presto/hive/OrcFileWriter.java
+++ b/presto-hive/src/main/java/com/facebook/presto/hive/OrcFileWriter.java
@@ -53,6 +53,7 @@ import static com.facebook.presto.hive.HiveErrorCode.HIVE_WRITE_VALIDATION_FAILE
 import static com.facebook.presto.hive.HiveManifestUtils.createFileStatisticsPage;
 import static com.facebook.presto.spi.StandardErrorCode.NOT_SUPPORTED;
 import static com.google.common.base.MoreObjects.toStringHelper;
+import static com.google.common.base.Throwables.throwIfInstanceOf;
 import static java.util.Objects.requireNonNull;
 
 public class OrcFileWriter
@@ -245,13 +246,14 @@ public class OrcFileWriter
         try {
             orcWriter.close();
         }
-        catch (IOException | UncheckedIOException e) {
+        catch (IOException | UncheckedIOException | PrestoException e) {
             try {
                 rollbackAction.call();
             }
             catch (Exception ignored) {
                 // ignore
             }
+            throwIfInstanceOf(e, PrestoException.class);
             throw new PrestoException(HIVE_WRITER_CLOSE_ERROR, "Error committing write to Hive. " + e.getMessage(), e);
         }
 
@@ -281,6 +283,9 @@ public class OrcFileWriter
             finally {
                 rollbackAction.call();
             }
+        }
+        catch (PrestoException e) {
+            throw e;
         }
         catch (Exception e) {
             throw new PrestoException(HIVE_WRITER_CLOSE_ERROR, "Error rolling back write to Hive", e);
--- a/presto-hive/src/test/java/com/facebook/presto/hive/TestOrcFileWriter.java
+++ b/presto-hive/src/test/java/com/facebook/presto/hive/TestOrcFileWriter.java
@@ -1,0 +1,146 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.hive;
+
+import com.facebook.presto.common.ErrorCode;
+import com.facebook.presto.common.io.DataOutput;
+import com.facebook.presto.common.io.DataSink;
+import com.facebook.presto.orc.DefaultOrcWriterFlushPolicy;
+import com.facebook.presto.orc.OrcWriterOptions;
+import com.facebook.presto.spi.PrestoException;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import io.airlift.units.DataSize;
+import org.testng.annotations.Test;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Optional;
+
+import static com.facebook.presto.common.ErrorType.EXTERNAL;
+import static com.facebook.presto.common.type.VarcharType.VARCHAR;
+import static com.facebook.presto.hive.HiveErrorCode.HIVE_WRITER_CLOSE_ERROR;
+import static com.facebook.presto.hive.parquet.ParquetTester.HIVE_STORAGE_TIME_ZONE;
+import static com.facebook.presto.orc.DwrfEncryptionProvider.NO_ENCRYPTION;
+import static com.facebook.presto.orc.NoOpOrcWriterStats.NOOP_WRITER_STATS;
+import static com.facebook.presto.orc.OrcEncoding.ORC;
+import static com.facebook.presto.orc.metadata.CompressionKind.NONE;
+import static io.airlift.units.DataSize.Unit.MEGABYTE;
+import static org.testng.Assert.assertEquals;
+
+public class TestOrcFileWriter
+{
+    private static final ErrorCode STORAGE_ERROR_CODE = new ErrorCode(123, "STORAGE_ERROR_CODE", EXTERNAL);
+
+    @Test
+    public void testPrestoExceptionPropagation()
+    {
+        // This test is to verify that a PrestoException thrown by the underlying data sink implementation is propagated as is
+        OrcFileWriter orcFileWriter = createOrcFileWriter(true);
+
+        try {
+            // Throws PrestoException with STORAGE_ERROR error code
+            orcFileWriter.commit();
+        }
+        catch (Exception e) {
+            assertEquals(e.getClass(), PrestoException.class);
+            assertEquals(e.getMessage(), "Dummy PrestoException from mocked data sink instance");
+            assertEquals(((PrestoException) e).getErrorCode(), STORAGE_ERROR_CODE);
+        }
+    }
+
+    @Test
+    public void testIOExceptionPropagation()
+    {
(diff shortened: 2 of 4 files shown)


presto-hive/src/main/java/com/facebook/presto/hive/OrcFileWriter.java
```diff
@@ -245,14 +246,18 @@ public Optional<Page> commit()
...
             }
             catch (Exception ignored) {
                 // ignore
             }
-            throw new PrestoException(HIVE_WRITER_CLOSE_ERROR, "Error committing write to Hive. " + e.getMessage(), e);
+            ErrorCodeSupplier errorCodeSupplier = HIVE_WRITER_CLOSE_ERROR;
+            if (e instanceof PrestoException) {
+                errorCodeSupplier = () -> ((PrestoException) e).getErrorCode();
+            }
+            throw new PrestoException(errorCodeSupplier, "Error committing write to Hive. " + e.getMessage(), e);
```
ajaygeorge, line 260: wondering why we need to wrap a PrestoException inside another PrestoException in line 260.?
If it is a PrestoException do we need to even catch it, since looks like all we are doing is adding an additional message to the exception.

If you feel we need to catch it then can we keep that in a separate catch block above the current one so that you can get rid of the instanceOf check in line 257. 

Also the rollbackAction should move into the finally block since it is a cleanup similar to how it is done in  rollback() method below. 
Do we need the same error code handling done in orcWriter.close() call in rollback() as well.

swapsmagic:
Please link the reference PR from meta repo to make it clear how this will work end to end.

presto-hive/src/main/java/com/facebook/presto/hive/OrcFileWriterFactory.java
```diff
@@ -291,8 +292,12 @@ else if (com.facebook.hive.orc.OrcOutputFormat.class.getName().equals(storageFor
...
                     dwrfWriterEncryption));
         }
-        catch (IOException e) {
-            throw new PrestoException(HIVE_WRITER_OPEN_ERROR, "Error creating " + orcEncoding + " file. " + e.getMessage(), e);
+        catch (PrestoException | IOException e) {
+            ErrorCodeSupplier errorCodeSupplier = HIVE_WRITER_OPEN_ERROR;
+            if (e instanceof PrestoException) {
+                errorCodeSupplier = () -> ((PrestoException) e).getErrorCode();
+            }
+            throw new PrestoException(errorCodeSupplier, "Error creating " + orcEncoding + " file. " + e.getMessage(), e);
```
swapsmagic, line 300: Add some unit test to verify this as well.

presto-hive/src/main/java/com/facebook/presto/hive/OrcFileWriter.java
NikhilCollooru: Yeah after looking more ..the extra wrapping is not necessary. Let me streamline it and just propagate it if its a presto exception

NikhilCollooru: This is a commit call. we need to rollback only if there is an exception during writer.close() call.

ajaygeorge: I was talking about the rollback code a few lines below which also uses  orcWriter.close();
 ```
    @Override
    public void rollback()
    {
        try {
            try {
                orcWriter.close();
            }
```

File: presto-hive-metastore/src/main/java/com/facebook/presto/hive/metastore/MetastoreUtil.java [hunk 3]
```diff
@@ -502,6 +502,9 @@ public static FileSystem getFileSystem(HdfsEnvironment hdfsEnvironment, HdfsCont
         try {
             return hdfsEnvironment.getFileSystem(context, path);
         }
+        catch (PrestoException e) {
+            throw e;
+        }
```
ajaygeorge, line 507: these are redundant since you are catching and throwing back immediately.

NikhilCollooru: you are right . will remove this :neutral_face:

presto-hive/src/main/java/com/facebook/presto/hive/OrcFileWriter.java
NikhilCollooru: Discussed offline with Ajay, we need to rollback only when writer.close fails . So no need to put rollbackAction.call in final block.

presto-hive/src/main/java/com/facebook/presto/hive/OrcFileWriter.java
```diff
@@ -245,13 +246,14 @@ public Optional<Page> commit()
...
         }
-        catch (IOException | UncheckedIOException e) {
+        catch (IOException | UncheckedIOException | PrestoException e) {
             try {
                 rollbackAction.call();
             }
             catch (Exception ignored) {
                 // ignore
             }
+            propagateIfPossible(e.getCause(), PrestoException.class);
```
ajaygeorge, line 256: does this work?
propagateIfPossible will match the given exception or a RuntimeException / Error as per javadocs.
1. does `e.getCause` match `PrestoException`?
2. Also what if the actual exception is an UncheckedIOException , since it is a RuntimeException looks like it will match propagateIfPossible's rules and we will not throw a PrestoException back.

Also propagateIfPossible is a [deprecated](https://guava.dev/releases/23.0/api/docs/com/google/common/base/Throwables.html#propagateIfPossible-java.lang.Throwable-) method and scheduled for removal. (I know we use it at other places in presto)

NikhilCollooru: 1. good catch. it should just be `e` and not `e.getCause()`
2. `[propagateIfPossible](https://guava.dev/releases/23.0/api/docs/com/google/common/base/Throwables.html#propagateIfPossible-java.lang.Throwable-)([Throwable](http://docs.oracle.com/javase/8/docs/api/java/lang/Throwable.html?is-external=true) throwable)` is deprecated but `[propagateIfPossible](https://guava.dev/releases/23.0/api/docs/com/google/common/base/Throwables.html#propagateIfPossible-java.lang.Throwable-java.lang.Class-)([Throwable](http://docs.oracle.com/javase/8/docs/api/java/lang/Throwable.html?is-external=true) throwable, [Class](http://docs.oracle.com/javase/8/docs/api/java/lang/Class.html?is-external=true)<X> declaredType)` is NOT deprecated.

NikhilCollooru: I see the description of propagateIfPossible . It says "Propagates throwable exactly as-is, if and only if it is an instance of RuntimeException, Error, or declaredType."

May be i will use [throwIfInstanceOf](https://guava.dev/releases/23.0/api/docs/src-html/com/google/common/base/Throwables.html#line.72)

swapsmagic (approved):
Having some unit test would be good to make sure this behavior won't get broken in the future.

NikhilCollooru:
> Having some unit test would be good to make sure this behavior won't get broken in the future.

Done.

swapsmagic (approved):

jainxrohit (dismissed):

presto-hive/src/test/java/com/facebook/presto/hive/TestOrcFileWriter.java
```diff
@@ -0,0 +1,104 @@
...
+import static com.facebook.presto.hive.parquet.ParquetTester.HIVE_STORAGE_TIME_ZONE;
+import static com.facebook.presto.orc.DwrfEncryptionProvider.NO_ENCRYPTION;
+import static com.facebook.presto.orc.NoOpOrcWriterStats.NOOP_WRITER_STATS;
+import static com.facebook.presto.orc.OrcEncoding.ORC;
+import static com.facebook.presto.orc.metadata.CompressionKind.NONE;
+import static io.airlift.units.DataSize.Unit.MEGABYTE;
+
+public class TestOrcFileWriter
+{
+    @Test(expectedExceptions = PrestoException.class, expectedExceptionsMessageRegExp = "Dummy Presto Exception from mocked data sink instance")
```
ajaygeorge, line 41: can you add a test case for non PrestoExceptions as well so that we are testing that code path as well.

presto-hive/src/test/java/com/facebook/presto/hive/TestOrcFileWriter.java
```diff
@@ -0,0 +1,145 @@
...
+                        .withDictionaryMaxMemory(new DataSize(32, MEGABYTE))
+                        .build(),
+                new int[0],
+                ImmutableMap.of(),
+                HIVE_STORAGE_TIME_ZONE,
+                Optional.empty(),
+                null,
+                NOOP_WRITER_STATS,
+                NO_ENCRYPTION,
+                Optional.empty());
```
ajaygeorge, line 102: nit: instead of duplicating just have a method called createOrcFileWriter which takes in a data sink (or boolean) which you can flip between true and false to trigger the different paths.

presto-hive/src/test/java/com/facebook/presto/hive/TestOrcFileWriter.java
```diff
@@ -0,0 +1,145 @@
...
+
+        // Throws PrestoException with STORAGE_ERROR error code
+        orcFileWriter.commit();
+    }
+
+    @Test(expectedExceptions = PrestoException.class, expectedExceptionsMessageRegExp = "Error committing write to Hive.*")
```
ajaygeorge, line 75: instead of just asserting on the exception message and type alone, we should catch it and assert on the cause as well so as to check if the cause is indeed the IOException.

ajaygeorge (approved):
LGTM % a couple of nits.
Thanks for cleaning this up and adding tests !

ajaygeorge (approved):

NikhilCollooru:
@aaneja `TestQuickStatsProvider.quickStatsBuildTimeIsBounded` is failing after multiple retries. This seems a new test, can you check why its failing. I dont think its related to the code changes in this PR.
https://github.com/prestodb/presto/actions/runs/8056340779/job/22007021834?pr=21969

aaneja:
> @aaneja `TestQuickStatsProvider.quickStatsBuildTimeIsBounded` is failing after multiple retries. This seems a new test, can you check why its failing. I dont think its related to the code changes in this PR. https://github.com/prestodb/presto/actions/runs/8056340779/job/22007021834?pr=21969

@NikhilCollooru  A fix - https://github.com/prestodb/presto/pull/22020/
Will update this comment if/when CI passes for this

PR #21973, opened 2024-02-21
[native] Propagate task.max-partial-aggregation-memory to Velox.

spershin:
Propagate 'task.max-partial-aggregation-memory' config property to Velox's 'max_partial_aggregation_memory' query property.

```
== NO RELEASE NOTE ==
```


gggrace14 (approved):
Checked and it looks good to me!

PR #21983, opened 2024-02-21
Add partition verification for INSERT and CreateTableAsSelect  queries

kewang1024:
```
== RELEASE NOTES ==

Verifier Changes
* Add support to do extended partition verification for INSERT and CreateTableAsSelect queries. This can be enabled by the configuration property ``extended-verification``. It would verify each partition's data checksum if the inserted table is partitioned.

```


kewang1024:
### Test plan
1. Follow [Deploying Presto](https://prestodb.io/docs/current/installation/deployment.html#:~:text=Deploying%20Presto%201%20Installing%20Presto%20Download%20the%20Presto,...%205%20An%20Example%20Deployment%20with%20Docker%20) to set up 2 Presto servers.

steveburnett (dismissed):
LGTM! (docs)

Pull branch, local build of docs, looks good. Thanks!

presto-verifier/src/main/java/com/facebook/presto/verifier/framework/InsertExtendedVerification.java
```diff
@@ -0,0 +1,177 @@
...
+            DataMatchResult partitionMatchResult = match(
+                PARTITION_DATA,
+                checksumValidator,
+                controlDataColumns,
+                testDataColumns,
+                    controlPartitionChecksum.get(i),
+                    testPartitionChecksum.get(i));
+            if (!partitionMatchResult.isMatched()) {
+                return partitionMatchResult;
+            }
```
gggrace14, line 129: It looks you don't include the particular partition that has the mismatch, do you? If so, it might not be quite useful for users to debug.

Can you merge the DataMatchResult and show all that partitions that have mismatch? That'll be useful for debugging.

presto-verifier/src/main/java/com/facebook/presto/verifier/framework/InsertExtendedVerification.java
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    @Override
+    public DataMatchResult verify(
+            QueryObjectBundle control,
+            QueryObjectBundle test,
+            Optional<QueryResult<Void>> controlQueryResult,
+            Optional<QueryResult<Void>> testQueryResult,
+            ChecksumQueryContext controlChecksumQueryContext,
+            ChecksumQueryContext testChecksumQueryContext)
```
gggrace14, line 86: You might notice that the base class DataVerification.verify() supports several workflow branches
```
if (isControlEnabled()) {
    if (saveSnapshot) {
    }
}
else if (QUERY_BANK_MODE.equals(runningMode)) {
   if (snapshotQuery != null) {
   }
   else {
    }
}
```

The big thing I see here we need to respect all the workflow branches, and make sure no one of them breaks. Ex., some workflow branches disable control or disable checksum. So the user won't set a valid control cluster to run. So we cannot assume there's a control run before this point.

I agree with you we inherit the DataVerification and run this InsertExtendedVerification only for Insert and CAST

presto-docs/src/main/sphinx/admin/verifier.rst
```diff
@@ -286,6 +286,7 @@ Name                                        Description
 ``running-mode``                            Set to ``query-bank`` to make the Verifier run in ``query-bank`` mode. Supports
                                             ``query-bank`` and ``control-test``. Defaults to ``control-test``.
 ``save-snapshot``                           Set to ``true`` to save checksums to ``mysql`` database.
+``insert-extended-verification``            Set to ``true`` to run extended verification for INSERT queries.
```
gggrace14, line 289: It can also apply to ctas in addition to insert, right? In that case you can have a more general naming of the config.

mbasmanova:
@kewang1024 Ke, would you update PR description , commit message and documentation to provide more details about what kind of verification is being added in this PR. The existing one-liner in the PR description doesn't clarify that.

presto-docs/src/main/sphinx/admin/verifier.rst
```diff
@@ -286,6 +286,8 @@ Name                                        Description
 ``running-mode``                            Set to ``query-bank`` to make the Verifier run in ``query-bank`` mode. Supports
                                             ``query-bank`` and ``control-test``. Defaults to ``control-test``.
 ``save-snapshot``                           Set to ``true`` to save checksums to ``mysql`` database.
+``extended-verification``                   Set to ``true`` to run extended verification. It only applies to ``Insert``and ``CreateTableAsSelect`` queries.
```
steveburnett, line 289:
```suggestion
``extended-verification``                   Set to ``true`` to run extended verification. It only applies to ``Insert`` and ``CreateTableAsSelect`` queries.
```
Nit: missing space causes formatting error in local build of docs. 
Screenshot 2024-02-27 at 10 18 54 AM

steveburnett (dismissed):
LGTM! (docs)

Pull of updated branch, local build of docs, everything looks good. Thanks!

presto-docs/src/main/sphinx/admin/verifier.rst
```diff
@@ -286,6 +286,8 @@ Name                                        Description
 ``running-mode``                            Set to ``query-bank`` to make the Verifier run in ``query-bank`` mode. Supports
                                             ``query-bank`` and ``control-test``. Defaults to ``control-test``.
 ``save-snapshot``                           Set to ``true`` to save checksums to ``mysql`` database.
+``extended-verification``                   Set to ``true`` to run extended verification. It only applies to ``Insert`` and ``CreateTableAsSelect`` queries.
```
spershin, line 289: Would it be beet to have "writer" word in it? Or any other that would hint on the nature of the extended verification?

presto-verifier/src/main/java/com/facebook/presto/verifier/framework/VerificationFactory.java
```diff
@@ -100,18 +100,34 @@ public Verification get(SourceQuery sourceQuery,
...
 
         QueryRewriter queryRewriter = queryRewriterFactory.create(queryActions.getHelperAction());
+        DeterminismAnalyzer determinismAnalyzer = new DeterminismAnalyzer(
+                sourceQuery,
+                queryActions.getHelperAction(),
+                queryRewriter,
+                checksumValidator,
+                typeManager,
+                determinismAnalyzerConfig);
+        FailureResolverManager failureResolverManager = failureResolverManagerFactory.create(new FailureResolverFactoryContext(sqlParser, queryActions.getHelperAction()));
```
spershin, line 110: Before this change we only ran this for "QUERY" (meaning SELECT I assume).
Now, having it in front of the switch, we run it for CREATE_VIEW and CREATE_TABLE as well.
Is this intended?

spershin (approved):
@kewang1024 
Thanks for working on this!
It looks good with a couple of comments.

presto-verifier/src/main/java/com/facebook/presto/verifier/framework/VerificationFactory.java
kewang1024: Otherwise we need to create them for [CREATE_TABLE_AS_SELECT, INSERT] and [QUERY] separately and the code looks wordy and a bit messy.
Those three lines only creates object and has no external/expensive calls, and it makes the code much cleaner.

ericyuliu: previous logic falls thru all 3 query types CTAS, INSERT, QUERY, using same code path, which is ok.
but for below CREATE_VIEW and CREATE_TABLE, it seems not necessary.

ericyuliu (approved):
Nit: Since ExtendedVerification is inheriting from DataVerification and used only for CTAS and INSERT in switch case; it is no longer a generic extension for other query types.  Composition might be better than Inheritance here.

But overall looking good, thanks for adding this feature.

kewang1024:
> Nit: Since ExtendedVerification is inheriting from DataVerification and used only for CTAS and INSERT in switch case; it is no longer a generic extension for other query types. Composition might be better than Inheritance here.
> 
> But overall looking good, thanks for adding this feature.

Sure, I can try this change in next PR which would add verification for bucketed table

ajaygeorge (approved):
Approving since this is already reviewed by multiple folks.

PR #21984, opened 2024-02-21
Optimize repeat function by creating runlength block

feilong-liu:
## Description
Currently repeat function implementation is to write the result block multiple times. However this is not efficient, since the result block is a repeat of the same value multiple times, we can create a RunLengthBlock which is more efficient, both in memory and construction cost.

## Motivation and Context
To improve performance of repeat function.
Also the result block created by repeat function will have small memory size, and it will be evaluated during query compilation (otherwise it will be skipped as the result block is too big https://github.com/prestodb/presto/blob/6998a7362764a73186a13a713a651ec71314c5f7/presto-main/src/main/java/com/facebook/presto/sql/planner/RowExpressionInterpreter.java#L776)

## Impact
Improve performance for repeat function. repeat(1, 1000) shows 35% performance improvement.

Current:
```
Benchmark                          (repeatArgument)  Mode  Cnt      Score       Error  Units
BenchmarkRepeatFunction.benchmark              1000  avgt   20  60774.020 ± 14228.873  ns/op
```

After optimization
```
Benchmark                          (repeatArgument)  Mode  Cnt      Score      Error  Units
BenchmarkRepeatFunction.benchmark              1000  avgt   20  39639.608 ± 8549.023  ns/op
```

## Test Plan
Add unit tests

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Improve repeat function to create RunLengthEncodedBlock to improve performance
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/operator/scalar/RepeatFunction.java
+++ b/presto-main/src/main/java/com/facebook/presto/operator/scalar/RepeatFunction.java
@@ -14,7 +14,7 @@
 package com.facebook.presto.operator.scalar;
 
 import com.facebook.presto.common.block.Block;
-import com.facebook.presto.common.block.BlockBuilder;
+import com.facebook.presto.common.block.RunLengthEncodedBlock;
 import com.facebook.presto.common.type.StandardTypes;
 import com.facebook.presto.common.type.Type;
 import com.facebook.presto.spi.function.Description;
@@ -22,12 +22,10 @@ import com.facebook.presto.spi.function.ScalarFunction;
 import com.facebook.presto.spi.function.SqlNullable;
 import com.facebook.presto.spi.function.SqlType;
 import com.facebook.presto.spi.function.TypeParameter;
-import io.airlift.slice.Slice;
 
 import static com.facebook.presto.common.type.UnknownType.UNKNOWN;
 import static com.facebook.presto.spi.StandardErrorCode.INVALID_FUNCTION_ARGUMENT;
 import static com.facebook.presto.util.Failures.checkCondition;
-import static java.lang.Math.toIntExact;
 
 @ScalarFunction(value = "repeat", calledOnNullInput = true)
 @Description("Repeat an element for a given number of times")
@@ -43,9 +41,9 @@ public final class RepeatFunction
             @SqlNullable @SqlType("unknown") Boolean element,
             @SqlType(StandardTypes.INTEGER) long count)
     {
+        checkValidCount(count);
         checkCondition(element == null, INVALID_FUNCTION_ARGUMENT, "expect null values");
-        BlockBuilder blockBuilder = createBlockBuilder(UNKNOWN, count);
-        return repeatNullValues(blockBuilder, count);
+        return RunLengthEncodedBlock.create(UNKNOWN, null, (int) count);
     }
 
     @TypeParameter("T")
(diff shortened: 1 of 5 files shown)


presto-main/src/test/java/com/facebook/presto/type/TestArrayOperators.java
```diff
@@ -1705,8 +1718,6 @@ public void testRepeat()
         // illegal inputs
         assertInvalidFunction("REPEAT(2, -1)", INVALID_FUNCTION_ARGUMENT);
         assertInvalidFunction("REPEAT(1, 1000000)", INVALID_FUNCTION_ARGUMENT);
-        assertInvalidFunction("REPEAT('loooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooongvarchar', 9999)", INVALID_FUNCTION_ARGUMENT);
-        assertInvalidFunction("REPEAT(array[1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20], 9999)", INVALID_FUNCTION_ARGUMENT);
```
feilong-liu, line 1709: These two functions now succeeds because it will not exceed the memory limit after outputting RunLengthEncodedBlock.

kaikalur (approved):
Ship it!

kaikalur:
Also can we do some similar for SEQUENCE - using literal encoder?

mlyublena (approved):
Nice catch!

presto-main/src/main/java/com/facebook/presto/operator/scalar/RepeatFunction.java
```diff
@@ -55,112 +53,19 @@ public static Block repeat(
...
+        Block result = RunLengthEncodedBlock.create(type, element, (int) count);
+        checkCondition(result.getSizeInBytes() < MAX_SIZE_IN_BYTES, INVALID_FUNCTION_ARGUMENT,
+                "result of repeat function must not take more than 1000000 bytes");
+        return result;
     }
 
-    private static BlockBuilder createBlockBuilder(Type type, long count)
+    private static void checkArguments(long count)
     {
         checkCondition(count <= MAX_RESULT_ENTRIES, INVALID_FUNCTION_ARGUMENT, "count argument of repeat function must be less than or equal to 10000");
```
rschlussel, line 68: why does it need to be < 10K?

presto-main/src/main/java/com/facebook/presto/operator/scalar/RepeatFunction.java
```diff
@@ -55,112 +53,19 @@ public static Block repeat(
...
-        }
-        return blockBuilder.build();
+        Block result = RunLengthEncodedBlock.create(type, element, (int) count);
+        checkCondition(result.getSizeInBytes() < MAX_SIZE_IN_BYTES, INVALID_FUNCTION_ARGUMENT,
+                "result of repeat function must not take more than 1000000 bytes");
+        return result;
     }
 
-    private static BlockBuilder createBlockBuilder(Type type, long count)
+    private static void checkArguments(long count)
```
rschlussel, line 66: call this checkValidCount for clarity

rschlussel (dismissed):
This is nice!

presto-main/src/main/java/com/facebook/presto/operator/scalar/RepeatFunction.java
feilong-liu: I guess this is intentional to avoid generating too large array when this function was added. Since we never heard complaint about it, I think it's reasonable to keep this limit here.

presto-main/src/main/java/com/facebook/presto/operator/scalar/RepeatFunction.java
feilong-liu: Done

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java
```diff
@@ -7461,4 +7461,11 @@ public void testMergeKHyperLogLog()
                 "select cardinality(khyperloglog_agg(v1, v2)), uniqueness_distribution(khyperloglog_agg(v1, v2)) from (values (1, 1, 2, 3), (1, 1, 4, 0), (1, 2, 90, 20), (1, 2, 87, 1), (2, 1, 11, 30), (2, 1, 11, 11), " +
                         "(2, 2, 9, 1), (2, 2, 87, 2)) t(k1, k2, v1, v2)");
     }
+
+    @Test
+    public void testRepeat()
```
feilong-liu, line 7466: Add test for non constant function arguments

presto-main/src/test/java/com/facebook/presto/operator/scalar/BenchmarkRepeatFunction.java
```diff
@@ -0,0 +1,142 @@
...
+
+        @Setup
+        public void setup()
+        {
+            MetadataManager metadata = createTestMetadataManager();
+            ExpressionCompiler compiler = new ExpressionCompiler(metadata, new PageFunctionCompiler(metadata, 0));
+
+            Type inputType = INTEGER;
+            ConstantExpression inputValue = new ConstantExpression((long) 2, inputType);
+            ConstantExpression repeatNum = new ConstantExpression((long) repeatArgument, INTEGER);
```
kaikalur, line 116: Try non-const args here as well

steveburnett:
Nit: suggest changing the first word of the release note entry from "Optimize" to "Improve" following the [Order of Changes](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines#order-of-changes) in the [Release Note Guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines.

mlyublena (approved):

pranjalssh (commented):
Good stuff

pranjalssh (approved):

PR #21992, opened 2024-02-23
[Native] Add partitioning for TPC-DS tables in AbstractTestNativeTpcdsQueries

aditi-pandit:
The partitioning added is more close to TPC-DS schemas used in practice for benchmarking.

```
== NO RELEASE NOTE ==
```


yingsu00 (changes requested):
@aditi-pandit I think we need to keep the un-partitioned schema. We can add the partitioned schema as a separate test.

PR #21999, opened 2024-02-24
[Iceberg]Support non-identity partition predicate for thoroughly push down

hantangwangd:
## Description

This PR extends Iceberg capabilities to pushdown thoroughly predicates on partitioning columns. Besides pushdown thoroughly predicates on identity partitioning columns, it also pushdown thoroughly predicates if they align with partitioning boundaries.

It is also helpful for DELETE, as it allows to do metadata-only delete in more cases, where we don't really needing to do a row-level delete.

## Test Plan

 - Added test cases for pushing down predicates on columns of various partition transforms

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergUtil.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergUtil.java
@@ -14,11 +14,13 @@
 package com.facebook.presto.iceberg;
 
 import com.facebook.airlift.log.Logger;
+import com.facebook.presto.common.GenericInternalException;
 import com.facebook.presto.common.predicate.Domain;
 import com.facebook.presto.common.predicate.NullableValue;
 import com.facebook.presto.common.predicate.TupleDomain;
 import com.facebook.presto.common.type.DecimalType;
 import com.facebook.presto.common.type.Decimals;
+import com.facebook.presto.common.type.TimestampType;
 import com.facebook.presto.common.type.Type;
 import com.facebook.presto.common.type.TypeManager;
 import com.facebook.presto.common.type.VarbinaryType;
@@ -154,15 +156,20 @@ import static com.google.common.collect.Streams.mapWithIndex;
 import static com.google.common.collect.Streams.stream;
 import static io.airlift.slice.Slices.utf8Slice;
 import static io.airlift.slice.Slices.wrappedBuffer;
+import static java.lang.Double.doubleToRawLongBits;
+import static java.lang.Double.longBitsToDouble;
 import static java.lang.Double.parseDouble;
 import static java.lang.Float.floatToRawIntBits;
+import static java.lang.Float.intBitsToFloat;
 import static java.lang.Float.parseFloat;
 import static java.lang.Integer.parseInt;
 import static java.lang.Long.parseLong;
 import static java.lang.Math.toIntExact;
+import static java.lang.Math.ulp;
 import static java.lang.String.format;
 import static java.util.Collections.emptyIterator;
 import static java.util.Comparator.comparing;
+import static java.util.Objects.requireNonNull;
 import static org.apache.iceberg.BaseMetastoreTableOperations.ICEBERG_TABLE_TYPE_VALUE;
 import static org.apache.iceberg.BaseMetastoreTableOperations.TABLE_TYPE_PROP;
 import static org.apache.iceberg.CatalogProperties.IO_MANIFEST_CACHE_ENABLED;
@@ -188,6 +195,16 @@ public final class IcebergUtil
     private static final Pattern SIMPLE_NAME = Pattern.compile("[a-z][a-z0-9]*");
     private static final Logger log = Logger.get(IcebergUtil.class);
     public static final int MIN_FORMAT_VERSION_FOR_DELETE = 2;
+
+    public static final long DOUBLE_POSITIVE_ZERO = 0x0000000000000000L;
+    public static final long DOUBLE_POSITIVE_INFINITE = 0x7ff0000000000000L;
+    public static final long DOUBLE_NEGATIVE_ZERO = 0x8000000000000000L;
+    public static final long DOUBLE_NEGATIVE_INFINITE = 0xfff0000000000000L;
+
+    public static final int REAL_POSITIVE_ZERO = 0x00000000;
+    public static final int REAL_POSITIVE_INFINITE = 0x7f800000;
+    public static final int REAL_NEGATIVE_ZERO = 0x80000000;
+    public static final int REAL_NEGATIVE_INFINITE = 0xff800000;
 
     private IcebergUtil() {}
 
@@ -736,6 +753,47 @@ public final class IcebergUtil
         throw new PrestoException(GENERIC_INTERNAL_ERROR, "Invalid partition type " + type.toString());
     }
 
+    /**
+     * Returns the adjacent value that compares bigger than or less than {@code value} based on parameter {@code isPrevious}.
+     * <p>
+     * The type of the value must match {@code #type.getJavaType()}.
+     *
+     * @throws IllegalStateException if the type is not {@code #isOrderable()}
+     */
+    public static Optional<Object> getAdjacentValue(Type type, Object value, boolean isPrevious)
+    {
+        if (!type.isOrderable()) {
+            throw new IllegalStateException("Type is not orderable: " + type);
+        }
+        requireNonNull(value, "value is null");
+
+        if (type.equals(BIGINT) || type instanceof TimestampType) {
+            return getBigintAdjacentValue(value, isPrevious);
+        }
+
+        if (type.equals(INTEGER) || type.equals(DATE)) {
+            return getIntegerAdjacentValue(value, isPrevious);
+        }
+
+        if (type.equals(SMALLINT)) {
+            return getSmallIntAdjacentValue(value, isPrevious);
+        }
+
+        if (type.equals(TINYINT)) {
+            return getTinyIntAdjacentValue(value, isPrevious);
+        }
+
+        if (type.equals(DOUBLE)) {
+            return getDoubleAdjacentValue(value, isPrevious);
+        }
+
+        if (type.equals(REAL)) {
+            return getRealAdjacentValue(value, isPrevious);
+        }
+
+        return Optional.empty();
+    }
+
     public static Map<Integer, HivePartitionKey> getPartitionKeys(ContentScanTask<DataFile> scanTask)
     {
         StructLike partition = scanTask.file().partition();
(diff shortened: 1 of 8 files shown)


hantangwangd:
Convert to draft as this PR depending on #21959. So it should be reviewed after #21959 is merged.

hantangwangd:
> Is this a cherry pick? At first glance, I don't see substantial similarities with the referenced PR.

I referenced the implementation of trino, which may evolve gradually through a series of PRs. Is it suitable to be described as a "cherry-pick"?

hantangwangd:
> Based on my reading of the code, this wouldn't qualify as a cherry pick as it's substantially different, and even the way it's performed is different between the two systems. I think we can remove the cherry pick reference.

OK, fixed! Thanks for the suggestion.

presto-common/src/main/java/com/facebook/presto/common/type/Type.java
```diff
@@ -186,4 +189,36 @@ public interface Type
...
     int compareTo(Block leftBlock, int leftPosition, Block rightBlock, int rightPosition);
+
+    /**
+     * Returns the maximum value that compares less than {@code value}.
+     * <p>
+     * The type of the value must match {@link #getJavaType}.
+     *
+     * @throws IllegalStateException if this type is not {@link #isOrderable() orderable}
+     */
+    default Optional<Object> getPreviousValue(Object value)
```
tdcmeehan, line 200: Just for consistency, should we implement these for other orderable types such as Double, Real, etc?

tdcmeehan: Also, because Types may be registered through the SPI, I think instead of returning optional empty, by default it should throw.  All engine recognized types that are orderable should have implementations, and SPI types that don't implement this should throw clear exceptions that indicate to plugin authors what to do to fix the problem.

tdcmeehan (changes requested):
Just some initial comments, I'll need more time to review this.

presto-common/src/main/java/com/facebook/presto/common/type/Type.java
hantangwangd: > Just for consistency, should we implement these for other orderable types such as Double, Real, etc?

For now we only used this method for Iceberg's non-identity/bucket/void transform, so the implementation of double/real are useless at the moment, as they are not the supported types for truncate/year/month/day/hour. But as your suggestion, we could support double/real based on java provided method `Math.ulp()` for consistency and potential future usage. And I will figure out whether we can also support decimal.

hantangwangd: > Also, because Types may be registered through the SPI, I think instead of returning optional empty, by default it should throw. All engine recognized types that are orderable should have implementations, and SPI types that don't implement this should throw clear exceptions that indicate to plugin authors what to do to fix the problem.

Good suggestion, make the default methods in `Type` to throw and let individual types to determine whether it could figure out the previous value and next value. In that way, all engine recognized types that are orderable implement the two methods and return corresponding result or an empty optional to declare that  it could not figure out the value exactly adjacent a specified value. And the SPI type would be reminded clearly. I will fix it.

tdcmeehan (changes requested):
Instead of making changes directly to `Type`, can we just have a new utility method in the Iceberg connector which takes in the Type, the Object, and from those two things infers the primtive type of the Object from the SQL Type?  The helper methods on Type don't directly relate to Block usage, which is what we expect for Types.

hantangwangd:
> Instead of making changes directly to `Type`, can we just have a new utility method in the Iceberg connector which takes in the Type, the Object, and from those two things infers the primtive type of the Object from the SQL Type? The helper methods on Type don't directly relate to Block usage, which is what we expect for Types.

Very reasonable, I will do this. Should we place the utility methods in iceberg connector or common `TypeUtils`? If we place them in iceberg connector, then should we just need to support the types needed by iceberg's truncate/year/month/day/hour transform?

presto-common/src/main/java/com/facebook/presto/common/type/AbstractIntType.java
```diff
@@ -141,4 +143,14 @@ public static long hash(int value)
         // xxhash64 mix
         return rotateLeft(value * 0xC2B2AE3D27D4EB4FL, 31) * 0x9E3779B185EBCA87L;
     }
+
+    protected static void checkValueValid(long value)
```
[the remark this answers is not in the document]
hantangwangd: Good idea, will fix it.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergLogicalPlanner.java
```diff
@@ -240,6 +247,836 @@ public void testFiltersWithPushdownDisable()
...
+        if (legacyTimestamp) {
+            sessionBuilder.setTimeZoneKey(TimeZoneKey.getTimeZoneKey(zoneId));
+        }
+        Session session = sessionBuilder.build();
+        assertUpdate(session, "CREATE TABLE test_hour_transform (d TIMESTAMP, b BIGINT) WITH (partitioning = ARRAY['hour(d)'])");
+
+        assertPlan(session, "SELECT * FROM test_hour_transform WHERE d IS NOT NULL",
+                output(exchange(strictTableScan("test_hour_transform", identityMap("d", "b")))));
+        assertPlan(session, "SELECT * FROM test_hour_transform WHERE d IS NULL",
+                output(exchange(strictTableScan("test_hour_transform", identityMap("d", "b")))));
```
ZacBlanco, line 274: nit: If I'm parsing these tests right, I think you are really just checking that the table scan has the correct parameters. We don't care whether the output or exchanges or there. Should be able to replace those with `anyTree` instead. This goes for almost all of the added tests

```suggestion
anyTree(strictTableScan("test_hour_transform", identityMap("d", "b")));
```

presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergLogicalPlanner.java
```diff
@@ -240,6 +247,836 @@ public void testFiltersWithPushdownDisable()
...
+        assertPlan(session, "SELECT * FROM test_day_transform_date WHERE CAST(d AS date) >= DATE '2015-01-13'",
+                output(exchange(strictTableScan("test_day_transform_date", identityMap("d", "b")))));
+
+        //TODO: d comparison with TIMESTAMP should be unwrapped
+        assertPlan(session, "SELECT * FROM test_day_transform_date WHERE d >= cast(TIMESTAMP '2015-01-13 00:00:00' as date)",
+                output(exchange(strictTableScan("test_day_transform_date", identityMap("d", "b")))));
+        assertPlan(session, "SELECT * FROM test_day_transform_date WHERE d > cast(TIMESTAMP '2015-01-13 00:00:00.001' as date)",
+                output(exchange(strictTableScan("test_day_transform_date", identityMap("d", "b")))));
+
+        // TODO: the following test cases could work after optimizer implementing unwrap the left function
```
ZacBlanco, line 350: IMO I think we should file github issue for this, move these to a separate test function with an `@Ignore` and a comment linking to the github issue. Same for the rest of the left-unwrapped TODOs.

ZacBlanco (commented):
Great stuff! The core changes LGTM. Mainly a few nits on the tests

presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergLogicalPlanner.java
hantangwangd: I think the main purpose here is to check whether there exist a `FilterNode` after the pushdown. And `anyTreeI()` would bring in ambiguity.  what's your opinion, is there a better way to demonstrate something like `here do not exist FilterNode`?

presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergLogicalPlanner.java
hantangwangd: Great idea, I will do it.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergUtil.java
```diff
@@ -735,6 +752,160 @@ public static Object deserializePartitionValue(Type type, String valueString, St
...
     }
 
+    /**
+     * Returns the maximum value that compares less than {@code value}.
+     * <p>
+     * The type of the value must match {@code #type.getJavaType()}.
+     *
+     * @throws IllegalStateException if the type is not {@code #isOrderable()}
+     */
+    public static Optional<Object> getPreviousValue(Type type, Object value)
```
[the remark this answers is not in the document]
hantangwangd: OK, I will figure out and fix it.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergLogicalPlanner.java
ZacBlanco: You could try `anyNot` from `PlanMatchPattern`. I think this will also make the intentions more clear.

```suggestion
                anyNot(FilterNode.class, anyTree(strictTableScan("test_hour_transform", identityMap("d", "b"))));
```

hantangwangd: Thank you for the information, I will try it.

hantangwangd:
Hi @ZacBlanco, I have combined `getPreviousValue()` and `getNextValue()` to a single function `getAdjacentValue()`, and used `anyNot(...)` to extract functions `thoroughlyPushdown(...)` and `notThoroughlyPushdown(...)` which could have a more clearer meaning. Then refactored test cases based on these functions, and moved the left-unwrap TODOs to separate @Ignore test functions. Please take a look when convenient, thanks!

ZacBlanco (dismissed):

tdcmeehan (approved):

hantangwangd:
@tdcmeehan @ZacBlanco  just resolved the conflict in `IcebergPlanOptimizer`.

ZacBlanco (dismissed):

ZacBlanco:
Don't forget to squash your commits

tdcmeehan (approved):

hantangwangd:
> Don't forget to squash your commits

OK, done!

presto-iceberg/src/main/java/com/facebook/presto/iceberg/PartitionTransforms.java
```diff
@@ -172,11 +247,21 @@ private static Block bucketInteger(Block block, int count)
         return bucketBlock(block, count, position -> bucketHash(INTEGER.getLong(block, position)));
     }
 
+    private static int bucketValueInteger(Block block, int position, int count)
+    {
+        return bucketValue(block, count, position, pos -> bucketHash(INTEGER.getLong(block, pos)));
```
Marcono1234, line 252: Here and for some of the other newly added methods: Is it intended that this calls `bucketValue(..., count, position, ...)` even though the signature of the method is `bucketValue(..., int position, int count, ...)` (that is, the arguments are switched)?

Sorry for this drive-by and late review comment, hopefully it is useful though. I wasn't sure whether a separate issue or discussion was justified.

ZacBlanco: Thanks for pointing this out. I think in some cases the code will still work but bucketing will technically be incorrect. We should switch the arguments. I'll open a PR later today to fix this

hantangwangd: Nice catch @Marcono1234, thanks for pointing out this, seems we didn't have sufficient test cases to cover these codes. So perhaps we need to supplement some test cases that can cover these code logics in the future.

ZacBlanco: https://github.com/prestodb/presto/pull/24309 fixes this

PR #22043, opened 2024-02-28
Use indicative mood in the body

elharo:
The subject line above is imperative. "If applied, this commit will ___________" is indicative mood, not imperative.

## Description
The example given does not fit the instructions, but does fit common practice so change the instructions. 

## Test Plan
CI

## Release Notes

```
== NO RELEASE NOTE ==
```


tdcmeehan (approved):
Thanks!  These guidelines were probably mis-copied from https://cbea.ms/git-commit/#imperative, which actually states the imperative mood is not needed in the body.

PR #22044, opened 2024-02-28
[native] Introduce PrestoConnector Interface

arhimondr:
## Description

To allow adding support for more connectors in Prestissimo

Allow dynamic registration of Presto to Velox connector adapters to enable support
for third party connectors. An additional adapter can be configured and
registered dynamically based on the configuration properties.

## Motivation and Context

Currently hardcoded to support only Hive, Iceberg, Tpch connectors

## Test Plan

Automated tests

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```


arhimondr:
Ready for review

spershin (dismissed):
Looks good!
Thanks for the change.
It makes it easier to add support for new connectors and also have their implementations outside of PrestoDB.

presto-native-execution/presto_cpp/presto_protocol/special/ColumnHandle.cpp.inc
```diff
@@ -17,55 +17,21 @@ void to_json(json& j, const std::shared_ptr<ColumnHandle>& p) {
...
 void from_json(const json& j, std::shared_ptr<ColumnHandle>& p) {
   String type;
   try {
-    type = p->getSubclassKey(j);
+    if (j.is_array()) {
+      // enum is serialized as an array: ["type","instance"]
+      type = j[0];
+    } else {
+      type = p->getSubclassKey(j);
+    }
```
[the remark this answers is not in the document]
arhimondr: Sometimes connector entities are empty (do not contain any fields). In it common to define them as enums (e.g.: https://github.com/prestodb/presto/blob/master/presto-tpch/src/main/java/com/facebook/presto/tpch/TpchTransactionHandle.java, https://github.com/prestodb/presto/blob/master/presto-kudu/src/main/java/com/facebook/presto/kudu/KuduTransactionHandle.java).

Presto serializes enum's as a JSON array (e.g.: `["tpch","INSTANCE"]`).

Previously only `TransactionHandle` was allowed to be an `enum`: https://github.com/prestodb/presto/blob/master/presto-native-execution/presto_cpp/presto_protocol/presto_protocol.cpp#L1307 as this was the only instance when an enum was used (`TpchTransactionHandle`).

However other connectors may implement other entities similarly.

presto-native-execution/presto_cpp/main/types/PrestoToVeloxConnector.cpp
```diff
@@ -0,0 +1,1388 @@
...
+#include "velox/connectors/tpch/TpchConnector.h"
+#include "velox/connectors/tpch/TpchConnectorSplit.h"
+
+namespace facebook::presto {
+
+namespace {
+std::vector<std::unique_ptr<const PrestoToVeloxConnector>>& connectors() {
+  static std::vector<std::unique_ptr<const PrestoToVeloxConnector>> connectors;
+  return connectors;
+}
```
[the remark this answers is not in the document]
arhimondr: `getPrestoToVeloxConnectors` is declared to return a `const` vector to avoid potential modifications (to ensure it is only modified via `registerPrestoToVeloxConnector`)

aditi-pandit:
@arhimondr : This is a very good change. Thanks !

presto-native-execution/presto_cpp/presto_protocol/special/ConnectorTableLayoutHandle.cpp.inc
```diff
@@ -17,56 +17,23 @@ void to_json(json& j, const std::shared_ptr<ConnectorTableLayoutHandle>& p) {
...
-
-  throw TypeError(type + " no abstract type ConnectorTableLayoutHandle");
+  getConnectorProtocol(type).to_json(j, p);
 }
 
 void from_json(const json& j, std::shared_ptr<ConnectorTableLayoutHandle>& p) {
   String type;
   try {
-    type = p->getSubclassKey(j);
+    if (j.is_array()) {
```
aditi-pandit, line 26: This code repeats in multiple places. Can this be abstracted as a function someplace ?

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -233,8 +234,21 @@ void PrestoServer::run() {
   registerMemoryArbitrators();
   registerShuffleInterfaceFactories();
   registerCustomOperators();
-  protocol::registerHiveConnectors();
-  protocol::registerTpchConnector();
+
+  protocol::registerConnectorProtocol(
```
aditi-pandit, line 238: Do we have to ensure that there is a PrestoToVeloxConnector for each ConnectorProtocol type ? If yes, then maybe we should combine them in a single function.

presto-native-execution/presto_cpp/main/types/PrestoToVeloxSplit.cpp
```diff
@@ -174,20 +33,19 @@ velox::exec::Split toVeloxSplit(
...
-            tpchSplit->partNumber),
-        splitGroupId);
-  }
+
   if (std::dynamic_pointer_cast<const protocol::EmptySplit>(connectorSplit)) {
     // We return NULL for empty splits to signal to do nothing.
     return velox::exec::Split(nullptr, splitGroupId);
   }
 
+  for (auto& connector : getPrestoToVeloxConnectors()) {
```
aditi-pandit, line 42: Could the Split _type() be used to retrieve the PrestoToVeloxConnector that can be used for the split conversion ?

presto-native-execution/presto_cpp/main/types/PrestoToVeloxConnector.cpp
```diff
@@ -0,0 +1,1388 @@
...
+  for (int i = 0; i < bytesFilters.size(); ++i) {
+    bytesGeneric.emplace_back(std::unique_ptr<common::Filter>(
+        dynamic_cast<common::Filter*>(bytesFilters[i].release())));
+  }
+
+  return std::make_unique<common::MultiRange>(
+      std::move(bytesGeneric), nullAllowed, false);
+}
+
+std::unique_ptr<common::Filter> toFilter(
```
aditi-pandit, line 617: Maybe we could move the TupleDomainFilter conversion to a separate class. Though I too am a bit on the fence about it since its used only during SplitConversion.

aditi-pandit (commented):
Thanks @arhimondr for this change. It definitely helps clean the code a lot.

presto-native-execution/presto_cpp/main/types/PrestoToVeloxConnector.cpp
arhimondr: It might be easier to decide where to move once we have more call sites. Let's keep it here for now.

arhimondr:
Thank you for the review @aditi-pandit. Those are great suggestions. Addressed.

Could you please take an another look?

presto-native-execution/presto_cpp/main/types/tests/PlanConverterTest.cpp
```diff
@@ -99,13 +98,16 @@ class PlanConverterTest : public ::testing::Test {
  protected:
   static void SetUpTestCase() {
     memory::MemoryManager::testingSetInstance({});
+    registerPrestoToVeloxConnector(
```
aditi-pandit, line 101: These tests are a bit odd. They register the connectors at different points in the test. I expected some error logic to be tested between the unregister and register or before the register, but there is none.

@majetideepak : Any comments ?

aditi-pandit (approved):
Thanks @arhimondr for the changes. They look good.

Please get a review from @majetideepak as well. He has spent most time on this code.

presto-native-execution/presto_cpp/presto_protocol/presto_protocol-json-cpp.mustache
```diff
@@ -23,8 +23,8 @@
 #include <folly/Format.h>
 #include <iostream>
 
+#include "ConnectorProtocol.h"
```
majetideepak, line 26: Use full path starting from `presto_cpp`

presto-native-execution/presto_cpp/presto_protocol/ConnectorProtocol.h
```diff
@@ -0,0 +1,264 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include "presto_protocol.h"
```
majetideepak, line 16: full path starting from `presto_cpp`

presto-native-execution/presto_cpp/main/types/tests/PrestoToVeloxSplitTest.cpp
```diff
@@ -44,9 +45,15 @@ protocol::ScheduledSplit makeHiveScheduledSplit() {
 }
 } // namespace
 
-class PrestoToVeloxSplitTest : public ::testing::Test {};
+class PrestoToVeloxSplitTest : public ::testing::Test {
+ protected:
+  static void SetUpTestCase() {
```
majetideepak, line 50: Let's use `void SetUp() override` and `void TearDown() override`

presto-native-execution/presto_cpp/main/types/tests/PlanConverterTest.cpp
majetideepak: Probably, a relic from the past. The tests cover the plan conversion and not the connectors.

presto-native-execution/presto_cpp/main/types/tests/PlanConverterTest.cpp
```diff
@@ -99,13 +98,16 @@ class PlanConverterTest : public ::testing::Test {
...
+        std::make_unique<HivePrestoToVeloxConnector>("hive"));
+    registerPrestoToVeloxConnector(
+        std::make_unique<HivePrestoToVeloxConnector>("hive-plus"));
```
majetideepak, line 104: Is hive-plus even used?

presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp
```diff
@@ -15,20 +15,14 @@
...
-#include "velox/connectors/tpch/TpchConnector.h"
 #include "velox/core/QueryCtx.h"
 #include "velox/exec/HashPartitionFunction.h"
 #include "velox/exec/RoundRobinPartitionFunction.h"
 #include "velox/expression/Expr.h"
 #include "velox/vector/ComplexVector.h"
 #include "velox/vector/FlatVector.h"
 #include "velox/core/Expressions.h"
-#include "velox/common/compression/Compression.h"
+#include "PrestoToVeloxConnector.h"
```
majetideepak, line 25: Full path starting from `presto_cpp`

presto-native-execution/presto_cpp/main/types/PrestoToVeloxConnector.h
```diff
@@ -0,0 +1,211 @@
...
+    std::unique_ptr<const PrestoToVeloxConnector> connector);
+
+const PrestoToVeloxConnector& getPrestoToVeloxConnector(
+    const std::string& connectorId);
+
+class PrestoToVeloxConnector {
+ public:
+  virtual ~PrestoToVeloxConnector() = default;
+
+  [[nodiscard]] const std::string& connectorId() const {
```
majetideepak, line 39: let's not return a reference here.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -233,8 +233,15 @@ void PrestoServer::run() {
   registerMemoryArbitrators();
   registerShuffleInterfaceFactories();
   registerCustomOperators();
-  protocol::registerHiveConnectors();
-  protocol::registerTpchConnector();
+
+  registerPrestoToVeloxConnector(
+      std::make_unique<HivePrestoToVeloxConnector>("hive"));
```
majetideepak, line 238: I would expect a single instance of `HivePrestoToVeloxConnector` to be used for both 
`hive` and `hive-hadoop2`. The unique_ptr choice seems odd to me.
The ConnectorProtocol translates the protocol structures from Presto to Velox. There is no state.
I would expect a singleton ConnectorProtocol instance for each Velox connector and not for each Presto Connector.

presto-native-execution/presto_cpp/main/types/PrestoToVeloxConnector.cpp
```diff
@@ -0,0 +1,1429 @@
...
+  static std::
+      unordered_map<std::string, std::unique_ptr<const PrestoToVeloxConnector>>
+          connectors;
+  return connectors;
+}
+} // namespace
+
+void registerPrestoToVeloxConnector(
+    std::unique_ptr<const PrestoToVeloxConnector> connector) {
+  const std::string& connectorId = connector->connectorId();
```
majetideepak, line 40: auto

presto-native-execution/presto_cpp/main/types/tests/PlanConverterTest.cpp
arhimondr: It is used in `ScanAggCustomConnectorId.json`

presto-native-execution/presto_cpp/main/types/PrestoToVeloxConnector.h
arhimondr: Could you please explain why a copy is preferred? What problem is there with leaving the decision to copy or not to copy to the caller?

presto-native-execution/presto_cpp/main/PrestoServer.cpp
arhimondr: Here's how I'm thinking about it.

PrestoToVeloxConnector is a virtual interface and must be interacted with via a pointer. 

When making this choice I considered:

- Raw pointer

The method would expect the lifecycle of a `PrestoToVeloxConnector` managed elsewhere and could be error prone.

It is probably fine if implementations are expected to be statically allocated. However it is hard to say whether it will always be the case for every possible connectors. I can see a case when an implementation of a `PrestoToVeloxConnector` is configured based on runtime parameters / configuration.

- Shared pointer

It is quite unusual to have two exactly same instances of `PrestoToVeloxConnector`. The `hive-hadoop2` connector exists for purely legacy reasons. Wasn't sure if it's worth expressing in the interface.

With these considerations I decided to simply go with a unique pointer.

presto-native-execution/presto_cpp/main/types/tests/PrestoToVeloxSplitTest.cpp
majetideepak: If someone runs these tests in parallel using something like https://github.com/google/gtest-parallel, then the static function can cause problems.

presto-native-execution/presto_cpp/main/types/PrestoToVeloxConnector.h
majetideepak: From an API standpoint, I see two problems. 
A user could change the private connectorId_ value by removing the const.
The lifetime of this variable might exceed that of the connector.
I don't think a copy here hurts.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
majetideepak: >  It is probably fine if implementations are expected to be statically allocated. However it is hard to say whether it will always be the case for every possible connectors. I can see a case when an implementation of a PrestoToVeloxConnector is configured based on runtime parameters / configuration.

This is not possible in the current setup since the presto_protocol structures are static. Dynamic connector support would need a new design of the protocol as well.

majetideepak:
From an API standpoint, `registerPrestoToVeloxConnector std::make_unique<IcebergPrestoToVeloxConnector>("iceberg"));` is also misleading to me since there is no iceberg connector in Velox. We use the HiveConnector from Velox.
The older API made that a bit explicit `registerConnector("iceberg", "hive-iceberg");`, but is not ideal.

How about we make it clear by naming it `IcebergConnectorProtocolToVelox` instead?

presto-native-execution/presto_cpp/main/types/tests/PrestoToVeloxSplitTest.cpp
arhimondr: How does the gtest-parallel work? Does it start separate processes for different fixture with different test selected? If that's the case does it change anything?

I actually think there's an another problem and ideally the `registerPrestoToVeloxConnector` should be called not at the fixture (test suite) level, but on the binary level via a global setup (https://google.github.io/googletest/advanced.html#global-set-up-and-tear-down).

The `SetUpTestCase` is called once for every "fixture" (or test suite in other words) in a binary. A single test binary can be executing multiple fixtures. In theory if there's one more test added to the same binary that needs to register a HivePrestoToVeloxConnector  with the same name it may fail (as it may already be registered by an another fixture). But this is not the case today. The two tests that need the connector to be registered (`PrestoToVeloxSplitTest` and `PlanConverterTest`) are currently separated into two separate binaries. 

I didn't find us using global setup anywhere, and since the tests were already structured this way (`PrestoToVeloxSplitTest` and `PlanConverterTest` being in separate binaries) I decided not to change it. But it feels like ideally this initialization has to be moved even one level above.

presto-native-execution/presto_cpp/main/types/PrestoToVeloxConnector.h
arhimondr: > A user could change the private connectorId_ value by removing the const.

You mean via a const_cast? Are we generally trying to make codebase resilient to `const_cast` mistakes? 

> The lifetime of this variable might exceed that of the connector.

Should the caller be responsible for creating a copy if the ownership change may occur? It seems to be quite common to return a reference / const reference to the managed objects (e.g.: map, vector, etc ..) and it's usually a responsibility of the caller side to think about ownership, lifecycle and whether a copy is necessary.

> I don't think a copy here hurts.

I don't think it hurts either. Copying a couple of bytes shouldn't be a problem. I'm quite new to the project and would love to gain better understand why certain design decisions are preferred over the other.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
arhimondr: > This is not possible in the current setup since the presto_protocol structures are static. Dynamic connector support would need a new design of the protocol as well.

Could you please elaborate?

I can conceivably see someone adding a connector taking a configuration property. For example the same very line could've been like:

```
...
registerPrestoToVeloxConnector(
      std::make_unique<HivePrestoToVeloxConnector>("hive", systemConfig->ignoreUnsupportedPrestoFilters()));
registerPrestoToVeloxConnector(
      std::make_unique<HivePrestoToVeloxConnector>("hive-hadoop2", true));
...
```

Basically an instance of a `HivePrestoToVeloxConnector` is created with a certain configuration in mind.

The current connectors don't take any configuration options, but some other connectors potentially could.

arhimondr:
@majetideepak 

Initially I was going to call it `IcebergPrestoConnectorToVeloxConnector` but the multiple `Connector` sounded like a tautology, hence shortened to `IcebergPrestoToVeloxConnector`.

You are right. The relation doesn't alway have to be 1-1, and in some cases different Presto connectors can translate to a single Velox connector. Does it feel like the current naming precludes that?

`IcebergConnectorProtocolToVelox` may conflate with the `PrestoProtocol` component that is responsible for simply understanding the Presto protocol. The  `PrestoToVeloxConnector` responsibility is to initialize Velox connector based on entities returned by the `PrestoProtocol.

presto-native-execution/presto_cpp/main/types/PrestoToVeloxConnector.h
majetideepak: > It seems to be quite common to return a reference / const reference to the managed objects

Fair enough. Let's stick to the existing patterns. I see we return a reference in the current API as well 
https://github.com/prestodb/presto/blob/2ff43e238a94eb182629ba04882d26ad21c3c8cc/presto-native-execution/presto_cpp/presto_protocol/Connectors.h#L50

presto-native-execution/presto_cpp/main/types/tests/PrestoToVeloxSplitTest.cpp
majetideepak: I see that this works today. But in general, I feel it is cleaner to keep the initialization local to a test than at the global level and hope there are no global conflicts. The example conflict you gave of adding another protocol instance by a subsequent test shows this. Let's stick to what other tests do to conclude this.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
majetideepak: I misunderstood your initial statement.
If you meant dynamic loading of catalog properties,
Ddon't we already load config from the `.properties` file and also session level properties?
When would we need this?

majetideepak:
> The relation doesn't alway have to be 1-1, and in some cases different Presto connectors can translate to a single Velox connector. Does it feel like the current naming precludes that

The naming is fine. But the arguments don't clarify that. That is my concern. My thought is to remove VeloxConnector from the name so that the user look at the documentation or code comments for more details.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
majetideepak: Will this change the protocol translation?

arhimondr: Currently it is not needed for existing connectors. It may be needed in the future. I can see a use for this pattern for an internal connector I'm trying to build at Meta.

presto-native-execution/presto_cpp/main/types/tests/PrestoToVeloxSplitTest.cpp
arhimondr: Discussed offline. Going to add `unregister` as well as move it to `SetUp`  / `TearDown` due to: 

> Note that SetUpTestSuite() may be called multiple times for a test fixture class that has derived classes

https://google.github.io/googletest/advanced.html#global-set-up-and-tear-down

presto-native-execution/presto_cpp/presto_protocol/ConnectorProtocol.cpp
```diff
@@ -0,0 +1,49 @@
...
+      protocols;
+  return protocols;
+}
+} // namespace
+
+void registerConnectorProtocol(
+    const std::string& connectorName,
+    std::unique_ptr<ConnectorProtocol> protocol) {
+  bool ok = protocols().insert({connectorName, std::move(protocol)}).second;
+  VELOX_CHECK(
```
aditi-pandit, line 32: Nit : Can you inline `protocols().insert({connectorName, std::move(protocol)}).second ` in the VELOX_CHECK ?

presto-native-execution/presto_cpp/presto_protocol/ConnectorProtocol.h
```diff
@@ -0,0 +1,264 @@
...
+  }
+
+ private:
+  template <typename DERIVED, typename BASE>
+  static void to_json_template(
+      json& j,
+      const std::shared_ptr<BASE>& p,
+      typename std::enable_if<std::is_base_of<BASE, DERIVED>::value, BASE>::
+          type* = 0) {
+    j = *std::static_pointer_cast<DERIVED>(p);
```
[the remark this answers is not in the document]
arhimondr: It relies on the `nlohmann::json`: https://github.com/nlohmann/json?tab=readme-ov-file#basic-usage

The `nlohmann::json` library would try to resolve `to_json` / `from_json` methods statically for a given type.

The required methods are automatically generated and declared in the `presto_protocol.h`. Here's an example for `HiveTableHandle`: https://github.com/prestodb/presto/blob/master/presto-native-execution/presto_cpp/presto_protocol/presto_protocol.h#L3137

aditi-pandit (approved):
Thanks @arhimondr

PR #22050, opened 2024-02-29
[native] Fix 'Number of sorting keys must be greater than zero'

mbasmanova:
A query with row_number() window function using same partitioning and sorting
keys and a limit on row number values used to fail with 

```
Caused by: java.lang.RuntimeException: sortingKeys_.size() > 0 (0 vs. 0) Number of sorting keys must be greater than zero
```

This happened because Prestissimo dropped sorting keys from Presto's
TopNRowNumberNode and tried to create TopNRowNumberNode in Velox with no
sorting keys. The fix is to use Velox's RowNumberNode when all sorting keys are
present as partitioning keys.


amitkdutta (approved):
Looks great. Thanks @mbasmanova

aditi-pandit (approved):
Thanks @mbasmanova

xiaoxmeng (approved):
@mbasmanova thanks for the fix!

Issue #21911, opened 2024-02-12
Presto Hive Connector creates too many small splits


pranjalssh:
We have a table with wide columns(maps), and it takes a long time(3.6 min)
```
select count(1) from table
```
Meta internal query id: [20240212_231123_91020_7kizs](https://www.internalfb.com/intern/presto/query/?query_id=20240212_231123_91020_7kizs#op_perf)

Using session property `hive.file_splittable=false` makes the same query take 10s!

Meta internal query id: [20240212_230955_07056_7jgdp](https://www.internalfb.com/intern/presto/query/?query_id=20240212_230955_07056_7jgdp#op_perf)

This session property creates exactly 1 split from 1 file in table. If otherwise true, presto scheduler will create multiple splits from the same file.

This is because presto scheduler creates splits according to file sizes - but does not take into account if we read only selected columns from the file. This is an extreme case where we read no data, but just metadata. In general, we should be able to tune split sizes based on amount of data we select from the files - so we can have fewer splits and presto runs faster.

There is an orthogonal problem when Presto is slow to process a large number of splits - that needs to be looked at in future.

This issue focuses on increasing split sizes when possible so query can be run with fewer splits. There can be 2 ways to fix this:

1) Use column stats from metastore to find relative size of file that we will be reading - and make split sizes according to it, not the whole file size.

2) Make it adaptive. If workers report they didn't read too much from the files when processing splits - increase split sizes.

mbasmanova:
CC: @tdcmeehan @majetideepak @aditi-pandit @pettyjamesm

tdcmeehan:
For most table formats, I'll bet approach 1 is the way to go.  For example, Iceberg already stores column-level information on the size in bytes for each column per file, and it already knows what columns will be accessed from the file during planning time, so it should be possible to change Iceberg to avoid this edge case.  I imagine it's the same for Delta and Hudi.

For Hive, we'd need to build this inside of Presto.  I wonder if the partition stats in the Hive metastore are sufficient for us to consistently make a good determination on the split size (i.e. are we losing anything by not knowing it per file).

mbasmanova:
> Use column stats from metastore to find relative size of file that we will be reading - and make split sizes according to it, not the whole file size.

I think this solution can be effective and seems not too hard to implement.

kaikalur:
@pranjalssh this will be a great issue to fix as it helps current adhoc queries as well as prestissimo. So I say let's prioritize this

[...]

PR #22051, opened 2024-02-29
Use dynamic split sizes in hive connector

pranjalssh:
## Description
FIxes https://github.com/prestodb/presto/issues/21911

> presto scheduler creates splits according to file sizes - but does not take into account if we read only selected columns from the file. In general, we should be able to tune split sizes based on amount of data we select from the files - so we can have fewer splits and presto runs faster.
## Test Plan
Added unit test, gated changes behind config, and ran test queries on a cluster

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

Hive Changes
* Add session property ``hive.dynamic_split_sizes_enabled`` to use dynamic split sizes based on data selected by query.
```


mbasmanova (commented):
@pranjalssh Would you document the new configuration property?

https://prestodb.io/docs/current/connector/hive.html#hive-configuration-properties

CC: @steveburnett

presto-hive/src/main/java/com/facebook/presto/hive/HiveSplitManager.java
```diff
@@ -281,6 +291,49 @@ public ConnectorSplitSource getSplits(
...
+                readColumnHandles.addAll(layout.getRequestedColumns().get());
+            }
+            readColumnHandles.addAll(layout.getPredicateColumns().values());
+            readColumnHandles.removeIf(handle -> handle.isHidden());
+
+            TableStatistics tableStatistics = metadata.getHiveStatisticsProvider().getTableStatistics(session, tableName, allColumns, allColumnTypes, partitions);
+            double totalSize = tableStatistics.getTotalSize().getValue();
+            double requiredSize = 0;
+            double rowCount = tableStatistics.getRowCount().getValue();
+            for (Map.Entry<ColumnHandle, ColumnStatistics> entry : tableStatistics.getColumnStatistics().entrySet()) {
```
mbasmanova, line 315: Here we loop over all columns (can be thousands), but process only a set of readColumnHandles columns (can be a handful). Would it make sense to change this to loop over readColumnHandles instead?

presto-hive/src/main/java/com/facebook/presto/hive/HiveSplitManager.java
```diff
@@ -281,6 +291,49 @@ public ConnectorSplitSource getSplits(
...
+                            int size = ((FixedWidthType) type).getFixedSize();
+                            value = size * rowCount;
+                        }
+                    }
+                    requiredSize += value;
+                    readColumnHandles.remove(entry.getKey());
+                }
+            }
+
+            if (readColumnHandles.isEmpty() && totalSize > 0 && isFinite(totalSize) && isFinite(requiredSize)) {
```
mbasmanova, line 332: How could it happen that readColumnHandles.isEmpty() is false here?

Can we check totalSize before going over the loop?

presto-hive/src/main/java/com/facebook/presto/hive/HiveSplitManager.java
```diff
@@ -281,6 +291,49 @@ public ConnectorSplitSource getSplits(
...
+
+            TableStatistics tableStatistics = metadata.getHiveStatisticsProvider().getTableStatistics(session, tableName, allColumns, allColumnTypes, partitions);
+            double totalSize = tableStatistics.getTotalSize().getValue();
+            double requiredSize = 0;
+            double rowCount = tableStatistics.getRowCount().getValue();
+            for (Map.Entry<ColumnHandle, ColumnStatistics> entry : tableStatistics.getColumnStatistics().entrySet()) {
+                if (readColumnHandles.contains(entry.getKey())) {
+                    double value = entry.getValue().getDataSize().getValue();
+                    // We do not compute total size stats for fixed width types, so count them manually.
+                    if (!isFinite(value) && isFinite(rowCount)) {
```
mbasmanova, line 319: > isFinite(rowCount)

Can we move this check before the loop?

presto-hive/src/main/java/com/facebook/presto/hive/HiveSplitManager.java
```diff
@@ -281,6 +291,49 @@ public ConnectorSplitSource getSplits(
...
+            readColumnHandles.removeIf(handle -> handle.isHidden());
+
+            TableStatistics tableStatistics = metadata.getHiveStatisticsProvider().getTableStatistics(session, tableName, allColumns, allColumnTypes, partitions);
+            double totalSize = tableStatistics.getTotalSize().getValue();
+            double requiredSize = 0;
+            double rowCount = tableStatistics.getRowCount().getValue();
+            for (Map.Entry<ColumnHandle, ColumnStatistics> entry : tableStatistics.getColumnStatistics().entrySet()) {
+                if (readColumnHandles.contains(entry.getKey())) {
+                    double value = entry.getValue().getDataSize().getValue();
+                    // We do not compute total size stats for fixed width types, so count them manually.
```
mbasmanova, line 318: Does that mean 'totalSize' doesn't include fixed-width columns?

presto-hive/src/test/java/com/facebook/presto/hive/TestHiveSplitScheduling.java
```diff
@@ -0,0 +1,106 @@
...
+        try {
+            getQueryRunner().execute("CREATE TABLE test_orders WITH (partitioned_by = ARRAY['ds', 'ts']) AS " +
+                    "SELECT orderkey, orderpriority, comment, custkey, '2020-09-01' as ds, '00:01' as ts FROM orders WHERE orderkey < 1000 " +
+                    "UNION ALL " +
+                    "SELECT orderkey, orderpriority, comment, custkey, '2020-09-02' as ds, '00:02' as ts FROM orders WHERE orderkey >= 1000 AND orderkey < 2000");
+
+            TestHiveEventListenerPlugin.TestingHiveEventListener eventListener = getEventListener();
+
+            // Wait for previous events to finish
+            Thread.sleep(2000);
```
mbasmanova, line 60: Is there a way to avoid explicit sleep calls?

mbasmanova (commented):
@pranjalssh Thank you for working on this. What kind of speed up have you observed on production queries? It would be nice to update commit message to provide more details about this change. Is there a way to see whether this optimization kicked in and what was the 'ratio' applied after the query finished running?

kaikalur:
I have seen 90% reduction in number of splits for some queries with this option (380K vs 3M)! And corresponding reduction in latency. so this can help sometimes especially in things like count( * ) queries

steveburnett:
> @pranjalssh Would you document the new configuration property?
> 
> https://prestodb.io/docs/current/connector/hive.html#hive-configuration-properties
> 
> CC: @steveburnett

@pranjalssh, when you can, please add documentation as suggested by @mbasmanova and recommended in the [Documentation](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#documentation) topic of the [Review and Commit Guidelines](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines). 

When you do, you can use the request review feature to tag me so I'll be notified and able to respond in a timely way and not delay the PR. Thanks!

presto-hive/src/main/java/com/facebook/presto/hive/HiveSplitManager.java
pranjalssh: Refactored to only query for read columns. I previously incorrectly assumed totalSize for summed for just columns provided

presto-hive/src/test/java/com/facebook/presto/hive/TestHiveSplitScheduling.java
pranjalssh: I found that event manager is synchronous in these tests, so we can just remove all sleep calls

pranjalssh:
Addressed comments @steveburnett @mbasmanova

steveburnett (approved):
LGTM! (docs)

Pull updated branch, new local build, everything looks good.

presto-hive/src/main/java/com/facebook/presto/hive/HiveSplitManager.java
```diff
@@ -296,16 +305,65 @@ public ConnectorSplitSource getSplits(
...
+            SchemaTableName tableName,
+            HiveTableLayoutHandle layout,
+            TransactionalMetadata metadata,
+            List<HivePartition> partitions)
+    {
+        double ratio = 1.0;
+        if (!isDynamicSplitSizesEnabled(session)) {
+            return ratio;
+        }
+        HiveTableHandle hiveTableHandle = new HiveTableHandle(tableName.getSchemaName(), tableName.getTableName());
```
rschlussel, line 325: use the function mergeRequestedAndPredicateColumns().  It will also handle if the same struct columns appear in both sets with different subfields.

presto-hive/src/main/java/com/facebook/presto/hive/HiveSplitManager.java
```diff
@@ -296,16 +305,65 @@ public ConnectorSplitSource getSplits(
...
+            if (!isFinite(value) && isFinite(rowCount)) {
+                HiveColumnHandle columnHandle = (HiveColumnHandle) entry.getKey();
+                Type type = metadata.getColumnMetadata(session, hiveTableHandle, columnHandle).getType();
+                if (type instanceof FixedWidthType) {
+                    int size = ((FixedWidthType) type).getFixedSize();
+                    value = size * rowCount;
+                }
+            }
+            requiredSize += value;
+            readColumnHandles.remove(entry.getKey());
```
rschlussel, line 353: why are we removing a value from readColumnHandles? It doesn't look like we use it again.

presto-hive/src/main/java/com/facebook/presto/hive/HiveSplitManager.java
```diff
@@ -296,16 +305,65 @@ public ConnectorSplitSource getSplits(
...
                 layout.isPartialAggregationsPushedDown());
 
-        HiveSplitSource splitSource = computeSplitSource(splitSchedulingContext, table, session, hiveSplitLoader);
+        HiveSplitSource splitSource = computeSplitSource(splitSchedulingContext, table, session, hiveSplitLoader, ratio);
         hiveSplitLoader.start(splitSource);
 
         return splitSource;
     }
 
+    private double getSplitScanRatio(
```
rschlussel, line 314: Can you clarify what splitScanRatio means? The name is confusing to me.  It looks like it's the proportion of bytes we expect to read based on column stats of columns we actually use compared to the total data size.

presto-hive/src/main/java/com/facebook/presto/hive/HiveSplitManager.java
```diff
@@ -296,16 +305,65 @@ public ConnectorSplitSource getSplits(
...
+        }
+        readColumnHandles.addAll(layout.getPredicateColumns().values());
+        readColumnHandles.removeIf(handle -> handle.isHidden() || handle.isPartitionKey());
+
+        Map<String, Type> readColumnTypes = readColumnHandles.stream().collect(toImmutableMap(HiveColumnHandle::getName, handle -> metadata.getColumnMetadata(session, hiveTableHandle, handle).getType()));
+        Map<String, ColumnHandle> readColumns = readColumnHandles.stream().collect(toImmutableMap(HiveColumnHandle::getName, identity()));
+
+        TableStatistics tableStatistics = metadata.getHiveStatisticsProvider().getTableStatistics(session, tableName, readColumns, readColumnTypes, partitions);
+        double totalSize = tableStatistics.getTotalSize().getValue();
+        double requiredSize = 0;
```
rschlussel, line 339: what does requiredSize mean? it looks like we're getting the total size of all columns we're selecting.  What's "required" about it?

presto-hive/src/main/java/com/facebook/presto/hive/HiveSplitManager.java
```diff
@@ -296,16 +305,65 @@ public ConnectorSplitSource getSplits(
...
     }
 
+    private double getSplitScanRatio(
+            ConnectorSession session,
+            SchemaTableName tableName,
+            HiveTableLayoutHandle layout,
+            TransactionalMetadata metadata,
+            List<HivePartition> partitions)
+    {
+        double ratio = 1.0;
```
rschlussel, line 321: we don't really use this variable in the logic. We either return 1.0 for unsupported cases or we return some computation that doesn't use any previously set value.  I would get rid of this variable and inline all the returns instead.  It makes it easier to see t's not really changed.

presto-hive/src/main/java/com/facebook/presto/hive/HiveSplitManager.java
pranjalssh: Done

presto-hive/src/main/java/com/facebook/presto/hive/HiveSplitManager.java
pranjalssh: Updated to readSize

presto-hive/src/main/java/com/facebook/presto/hive/HiveSplitSource.java
```diff
@@ -126,6 +130,14 @@ private HiveSplitSource(
         this.useRewindableSplitSource = useRewindableSplitSource;
         this.remainingInitialSplits = new AtomicInteger(maxInitialSplits);
         this.splitWeightProvider = isSizeBasedSplitWeightsEnabled(session) ? new SizeBasedSplitWeightProvider(getMinimumAssignedSplitWeight(session), maxSplitSize) : HiveSplitWeightProvider.uniformStandardWeightProvider();
+        // Clamp value within [0.1, 1.0].
+        // This ratio will be used to increase split sizes. The range implies
+        // 1) We do not increase more than 10x(>= 0.1)
```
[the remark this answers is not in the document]
rschlussel: Can you answer this question @pranjalssh?

rschlussel (commented):
looks good

pranjalssh:
@rschlussel updated comment

```
        // We schedule only upto 10x larger splits - being conservative not to schedule splits with too many rows.
        // For default size of 64MB, this will keep split sizes sent within 1GB. Usually files are smaller than this.
        ````

rschlussel (approved):

PR #22058, opened 2024-03-01
Skip duplicate key check when upcasting key in map cast

feilong-liu:
## Description
In map cast function, it will check if there are duplicate keys after cast. However it's not necessary when it's upcasting, skip it for these cases.

## Motivation and Context
map cast is slow, skip unnecessary duplicate key check.

## Impact
Test with a production query, the cast itself has CPU cost reduced by 80%.

## Test Plan
Added end to end test queries.

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* ...
* ...

Hive Changes
* ...
* ...
```

```
== NO RELEASE NOTE ==
```


presto-main/src/main/java/com/facebook/presto/operator/scalar/MapToMapCast.java
```diff
@@ -239,7 +244,12 @@ public static Block mapCast(
             }
         }
         else {
-            TypedSet typedSet = new TypedSet(toKeyType, fromMap.getPositionCount() / 2, "map-to-map cast");
+            boolean skipDuplicateKeyCheck = fromKeyType.equals(INTEGER) && toKeyType.equals(BIGINT)
+                    || fromKeyType.equals(INTEGER) && toKeyType.equals(VARCHAR) || fromKeyType.equals(BIGINT) && toKeyType.equals(VARCHAR);
```
kaikalur, line 248: Put this in a supporting function to be super clear.

kaikalur: Also use nested if statements for better readability

kaikalur (changes requested):

kaikalur (approved):
Ship it!

presto-main/src/main/java/com/facebook/presto/operator/scalar/MapToMapCast.java
```diff
@@ -274,4 +283,15 @@ public static Block mapCast(
         mapBlockBuilder.closeEntry();
         return (Block) toMapType.getObject(mapBlockBuilder, mapBlockBuilder.getPositionCount() - 1);
     }
+
+    public static boolean canSkipDuplicateKeysCheck(Type castFromType, Type castToType)
```
mlyublena, line 287: instead of special-casing can we introduce the following utility functions if these aren't already in the code base:
```
isUpcast(fromType, toType)
isLosslessCast(fromType, toType) // lossless cast is either an upcast or some combination of types e.g. integer->string
```

mlyublena, line 287: also nit: perhaps call them fromType and toType to improve readability (otherwise the difference between the two names is in the middle)

presto-main/src/main/java/com/facebook/presto/operator/scalar/MapToMapCast.java
```diff
@@ -274,4 +283,15 @@ public static Block mapCast(
...
+    {
+        if (castFromType.equals(INTEGER)) {
+            return castToType.equals(BIGINT) || castToType.equals(VARCHAR);
+        }
+        else if (castFromType.equals(BIGINT)) {
```
mlyublena, line 292: nit: else is redundant here

presto-main/src/main/java/com/facebook/presto/operator/scalar/MapToMapCast.java
```diff
@@ -239,7 +244,11 @@ public static Block mapCast(
             }
         }
         else {
-            TypedSet typedSet = new TypedSet(toKeyType, fromMap.getPositionCount() / 2, "map-to-map cast");
+            boolean skipDuplicateKeyCheck = canSkipDuplicateKeysCheck(fromKeyType, toKeyType);
+            TypedSet typedSet = null;
+            if (!skipDuplicateKeyCheck) {
```
mlyublena, line 249: instead of calling the function skip and using it always negated, perhaps invert the logic and call it requiresDuplicateKeyCheck

presto-main/src/main/java/com/facebook/presto/operator/scalar/MapToMapCast.java
feilong-liu: Should the new isUpcast and isLosslessCast be comprehensive? If yes, it will be notrival to include all possible conversions and not worth the effort considering Presstisimo is coming soon.
I am specializing for the int -> bigint/varchar and bigint -> varchar here, which is the most common cases for map cast. 
A compromise is to add the conversion cases we are interested in these isUpcast and isLosslessCast util functions, however with comment saying that this is not a comprehensive list and returning false does not necessarily mean that it is not upcast/lossless cast, which I am not sure if it's really helpful.

feilong-liu: Done

presto-main/src/main/java/com/facebook/presto/operator/scalar/MapToMapCast.java
feilong-liu: The fromType can be types other than integer and bigint, for these cases we currently do not consider for skipping duplicate check

presto-main/src/main/java/com/facebook/presto/operator/scalar/MapToMapCast.java
feilong-liu: Done

kaikalur (approved):

presto-main/src/main/java/com/facebook/presto/operator/scalar/MapToMapCast.java
kaikalur: A better task would be to revisit/fix implicit parameter conversions/casts for map operations for all types in the analyzer.

PR #22059, opened 2024-03-01
Remove map cast for element access

feilong-liu:
Rewrite expression from cast(feature as map<bigint, float>)[key] -> feature[cast(key as integer)], where feature is of type map<integer, float> and key is of type bigint, and element_at(cast(feature as map<bigint, float>), key) -> element_at(feature, try_cast(key as integer)), where feature is of type map<integer, float> and key is of type bigint, so as to get rid of map cast function.

## Description
Map cast is expensive, this PR tries to get rid of map cast when possible, it currently targets the following cases:
- cast(feature as map<bigint, float>)[key] -> feature[cast(key as integer)], where feature is of type map<integer, float> and key is of type bigint
- element_at(cast(feature as map<bigint, float>), key) -> element_at(feature, try_cast(key as integer)), where feature is of type map<integer, float> and key is of type bigint
 
Notice that here when it's accessing the map using subscript function, we use CAST function in index, and when it's element_at function, we use TRY_CAST function, so that
- when the key is out of integer range, for feature[key] it will fail both with and without optimization, fail with map key not exists before optimization and with cast failure after optimization
- when the key is out of integer range, for element_at(feature, key) it will return NULL both before and after optimization

## Motivation and Context
Get rid of the expensive map cast expression.

## Impact
Get rid of map cast for cost for these expressions.

## Test Plan
Unit tests

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Add an optimizer rule to get rid of map cast in map access functions when possible
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
+++ b/presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
@@ -315,6 +315,7 @@ public final class SystemSessionProperties
     public static final String USE_PARTIAL_AGGREGATION_HISTORY = "use_partial_aggregation_history";
     public static final String TRACK_PARTIAL_AGGREGATION_HISTORY = "track_partial_aggregation_history";
     public static final String REMOVE_REDUNDANT_CAST_TO_VARCHAR_IN_JOIN = "remove_redundant_cast_to_varchar_in_join";
+    public static final String REMOVE_MAP_CAST = "remove_map_cast";
     public static final String HANDLE_COMPLEX_EQUI_JOINS = "handle_complex_equi_joins";
     public static final String SKIP_HASH_GENERATION_FOR_JOIN_WITH_TABLE_SCAN_INPUT = "skip_hash_generation_for_join_with_table_scan_input";
     public static final String GENERATE_DOMAIN_FILTERS = "generate_domain_filters";
@@ -1908,6 +1909,11 @@ public final class SystemSessionProperties
                         featuresConfig.isRemoveRedundantCastToVarcharInJoin(),
                         false),
                 booleanProperty(
+                        REMOVE_MAP_CAST,
+                        "Remove map cast when possible",
+                        false,
+                        false),
+                booleanProperty(
                         HANDLE_COMPLEX_EQUI_JOINS,
                         "Handle complex equi-join conditions to open up join space for join reordering",
                         featuresConfig.getHandleComplexEquiJoins(),
@@ -3180,6 +3186,11 @@ public final class SystemSessionProperties
     public static boolean isRemoveRedundantCastToVarcharInJoinEnabled(Session session)
     {
         return session.getSystemProperty(REMOVE_REDUNDANT_CAST_TO_VARCHAR_IN_JOIN, Boolean.class);
+    }
+
+    public static boolean isRemoveMapCastEnabled(Session session)
+    {
+        return session.getSystemProperty(REMOVE_MAP_CAST, Boolean.class);
     }
 
     public static boolean shouldHandleComplexEquiJoins(Session session)
--- a/presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/RemoveMapCastRule.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/RemoveMapCastRule.java
@@ -1,0 +1,133 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.sql.planner.iterative.rule;
+
+import com.facebook.presto.Session;
+import com.facebook.presto.common.type.MapType;
+import com.facebook.presto.common.type.Type;
+import com.facebook.presto.expressions.RowExpressionRewriter;
+import com.facebook.presto.expressions.RowExpressionTreeRewriter;
+import com.facebook.presto.metadata.FunctionAndTypeManager;
+import com.facebook.presto.spi.relation.CallExpression;
+import com.facebook.presto.spi.relation.RowExpression;
+import com.facebook.presto.sql.planner.iterative.Rule;
+import com.facebook.presto.sql.relational.FunctionResolution;
+import com.google.common.collect.ImmutableSet;
+
+import java.util.Set;
+
+import static com.facebook.presto.SystemSessionProperties.isRemoveMapCastEnabled;
+import static com.facebook.presto.common.function.OperatorType.SUBSCRIPT;
+import static com.facebook.presto.common.type.BigintType.BIGINT;
+import static com.facebook.presto.common.type.IntegerType.INTEGER;
+import static com.facebook.presto.sql.relational.Expressions.call;
+import static com.facebook.presto.sql.relational.Expressions.castToInteger;
+import static com.facebook.presto.sql.relational.Expressions.tryCast;
+import static java.util.Objects.requireNonNull;
+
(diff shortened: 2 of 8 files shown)


presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
```diff
@@ -1900,6 +1901,11 @@ public SystemSessionProperties(
                         "If both left and right side of join clause are varchar cast from int/bigint, remove the cast here",
                         featuresConfig.isRemoveRedundantCastToVarcharInJoin(),
                         false),
+                booleanProperty(
+                        REMOVE_MAP_CAST,
+                        "Remove map cast when possible",
+                        false,
+                        false),
```
jaystarshot, line 1908: You might need to add this to features config to make this into a config property

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/RemoveMapCastRule.java
```diff
@@ -0,0 +1,124 @@
...
+                    && functionResolution.isCastFunction(((CallExpression) node.getArguments().get(0)).getFunctionHandle())
+                    && ((CallExpression) node.getArguments().get(0)).getArguments().get(0).getType() instanceof MapType) {
+                CallExpression castExpression = (CallExpression) node.getArguments().get(0);
+                RowExpression castInput = castExpression.getArguments().get(0);
+                Type fromKeyType = ((MapType) castInput.getType()).getKeyType();
+                Type fromValueType = ((MapType) castInput.getType()).getValueType();
+                Type toKeyType = ((MapType) castExpression.getType()).getKeyType();
+                Type toValueType = ((MapType) castExpression.getType()).getValueType();
+
+                if (toValueType.equals(fromValueType) && fromKeyType.equals(INTEGER) && toKeyType.equals(BIGINT) && node.getArguments().get(1).getType().equals(BIGINT)) {
```
kaikalur, line 110: Put it in a separate function

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java
```diff
@@ -7470,4 +7471,20 @@ public void testRepeat()
...
+                "values 0.5, 0.1");
+        assertQuery(enableOptimization, "select element_at(feature, key) from (values (map(array[cast(1 as integer), 2, 3, 4], array[0.3, 0.5, 0.9, 0.1]), cast(2 as bigint)), (map(array[cast(1 as integer), 2, 3, 4], array[0.3, 0.5, 0.9, 0.1]), 4)) t(feature,  key)",
+                "values 0.5, 0.1");
+        assertQuery(enableOptimization, "select element_at(feature, key) from (values (map(array[cast(1 as integer), 2, 3, 4], array[0.3, 0.5, 0.9, 0.1]), cast(2 as bigint)), (map(array[cast(1 as integer), 2, 3, 4], array[0.3, 0.5, 0.9, 0.1]), 400000000000)) t(feature, key)",
+                "values 0.5, null");
+        assertQueryFails(enableOptimization, "select feature[key] from (values (map(array[cast(1 as integer), 2, 3, 4], array[0.3, 0.5, 0.9, 0.1]), cast(2 as bigint)), (map(array[cast(1 as integer), 2, 3, 4], array[0.3, 0.5, 0.9, 0.1]), 400000000000)) t(feature, key)",
+                ".*Out of range for integer.*");
```
kaikalur, line 7488: Try some unrelated ones where this rule should not trigger - like varchar, map<varchar,float> etc

kaikalur (dismissed):

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/RemoveMapCastRule.java
```diff
@@ -0,0 +1,124 @@
...
+import static com.facebook.presto.common.type.BigintType.BIGINT;
+import static com.facebook.presto.common.type.IntegerType.INTEGER;
+import static com.facebook.presto.sql.relational.Expressions.call;
+import static com.facebook.presto.sql.relational.Expressions.castToInteger;
+import static com.facebook.presto.sql.relational.Expressions.tryCast;
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Remove cast on map if possible. Currently it only supports subscript and element_at function, and only works when map key is of type integer and index is bigint. For example:
+ * cast(feature as map<bigint, float>)[key] -> feature[cast(key as integer)], where feature is of type map<integer, float> and key is of type bigint
```
mlyublena, line 41: nit: can you put the before and after on separate lines for readability? something like this:
```
For arguments feature: map<integer, float>, key: bigint
Input: cast(feature as map<bigint, float>)[key]

Output:
feature[cast(key as integer)]
```

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/RemoveMapCastRule.java
```diff
@@ -0,0 +1,124 @@
...
+ * element_at(cast(feature as map<bigint, float>), key) -> element_at(feature, try_cast(key as integer)), where feature is of type map<integer, float> and key is of type bigint
```
mlyublena, line 42: where is the cast added in the first place? parser/analyzer?

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/RemoveMapCastRule.java
mlyublena, line 110: why do we special case the int/bigint pair? shouldn't this work for any cast that is not lossy?

kaikalur: Non-exact types as keys is dubious (and also not common) so we don't want to do that.

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/RemoveMapCastRule.java
kaikalur: Type promotion - ideally we should fix the type promo. For example:

element_at(map(int, real), bigint)

promotes map to map(bigint, real) instead we should coerce bigint to int? But I think our function resolution logic maybe inadequate for that. 

So for now, we just decided to drop these unnecessary casts. This helps a lot in our current prod workloads as there is a lot of mixing of int and bigint in keys and map

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/RemoveMapCastRule.java
mlyublena: +1

mlyublena: sounds good, if we extract this in a function as suggested above we can also extend this in the future to handle other pairs if needed

presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
steveburnett: > You might need to add this to features config to make this into a config property

If this becomes a new feature property, please add documentation for it.

pranjalssh (approved):

presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
rschlussel: @feilong-liu can you add a configuration property for this as per @jaystarshot's suggestion?  Session properties should always have a corresponding config property.

kaikalur: > @feilong-liu can you add a configuration property for this as per @jaystarshot's suggestion? Session properties should always have a corresponding config property.

We are trying to move away from that model for optimizations that we know are safe (famous last words lol) to reduce noise. So feature config is only for features and not for optimizations.

rschlussel: Adding a configuration doesn't seem especially noisy to me, but it does allow people to disable the feature for all queries if there ends up being a bug or regression in the optimizer (not everyone has infrastructure set up to easily tack session properties on to all user sessions if needed).

PR #22061, opened 2024-03-01
[native] Fix bug in creation of HashNode(kLeftSemiProject).

spershin:
## Description
In Presto Native we create HashNode(kLeftSemiProject) with nullAware=false from toVeloxQueryPlan(FilterNode). 'nullAware' dictates how we treat nulls in the left keys.

This causes some queries in Presto Native to return more rows than Presto. Because Presto does not have any member in class SemiJoinNode that indicates how it processes nulls and is always 'null aware'.

Interesting, that toVeloxQueryPlan(SemiJoinNode) code path creates HashNode(kLeftSemiProject) with nullAware=true, which is correct.

To fix the issue we route the creation of HashNode(kLeftSemiProject) in toVeloxQueryPlan(FilterNode) through toVeloxQueryPlan(SemiJoinNode), thus having only one place where we create HashNode(kLeftSemiProject) with nullAware=true.

## Test Plan
Added the previously failing query to the integration test.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp
+++ b/presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp
@@ -1625,6 +1625,14 @@ core::PlanNodePtr VeloxQueryPlanConverterBase::toVeloxQueryPlan(
       }
     }
 
+    // No clear join type - fallback to the standard 'to velox expr'.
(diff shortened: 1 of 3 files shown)


xiaoxmeng (approved):
@spershin thanks for the fix!

amitkdutta:
Thanks @spershin

PR #22063, opened 2024-03-02
Adding necessary validation when writingBytes to VariableWidthBlockBuilder 

swapsmagic:
## Description

   VariableBlockWidthBuilder requires creation of Slice to write Strings. But when the Strings larger than the supported slice size, we ended up having IndexOutOfBoundException. The reason being not having proper validation when we are writing bytes.
 
    - As part of this change, we are adding proper validation and raise SliceTooLargeException in those scenario.
    - Assigning a new error code **SLICE_TOO_LARGE** for this errors instead of getting GENERIC_INTERNAL_ERROR

## Motivation and Context
 At Meta we see lot of queries failing with IndexOutOfBoundException with a negative index error when writing bytes to slice. This was due to the amount of data we are writing crosses the limit but the necessary checks are not in place to capture and raise this error early. 

## Impact
IndexOutOfBoundException due to Large slice writing will be correctly categorized and failed with SliceTooLargeException. And the error code for all SliceTooLargeException errors is going to be SLICE_TOO_LARGE to help categorize it appropriately. 

## Test Plan
unit test

Benchmark Data:
With Changes
————————
Result "com.facebook.presto.operator.aggregation.BenchmarkArrayAggregation.arrayAggregation":
  28.633 ±(99.9%) 0.584 ns/op [Average]
  (min, avg, max) = (27.431, 28.633, 29.705), stdev = 0.673
  CI (99.9%): [28.049, 29.217] (assumes normal distribution)

Run complete. Total time: 00:01:43

Benchmark                                    (type)  Mode  Cnt   Score   Error  Units
BenchmarkArrayAggregation.arrayAggregation  VARCHAR  avgt   20  28.633 ± 0.584  ns/op

Baseline Number:
——————————
Result "com.facebook.presto.operator.aggregation.BenchmarkArrayAggregation.arrayAggregation":
  27.114 ±(99.9%) 0.491 ns/op [Average]
  (min, avg, max) = (26.338, 27.114, 28.068), stdev = 0.565
  CI (99.9%): [26.623, 27.605] (assumes normal distribution)

Run complete. Total time: 00:01:40

Benchmark                                    (type)  Mode  Cnt   Score   Error  Units
BenchmarkArrayAggregation.arrayAggregation  VARCHAR  avgt   20  27.114 ± 0.491  ns/op

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Fix problem when writing large varchar values to throw a user error when it exceeds internal limits

### CODE REVIEW

File: presto-common/src/main/java/com/facebook/presto/common/block/VariableWidthBlockBuilder.java [hunk 1]
```diff
@@ -244,9 +245,12 @@ public BlockBuilder writeLong(long value)
     @Override
     public BlockBuilder writeBytes(Slice source, int sourceIndex, int length)
     {
+        checkValidSliceRange(sourceIndex, length);
```
kaikalur, line 248: This is probably too expensive. This is one of the most heavily used functions so make sure it doesn't impact common use cases.

Maybe TRY/CATCH will work better.

kaikalur: > For every write call, Slice already does similar [checks](https://github.com/prestodb/slice/blob/master/src/main/java/com/facebook/slice/DynamicSliceOutput.java#L136). Also with this approach the code is much cleaner compared to try/catch way.

We need to balance cleanliness with performance in backend code. So I suggest a writing a benchmark test and measure the impact.

presto-spi/src/main/java/com/facebook/presto/spi/StandardErrorCode.java
```diff
@@ -113,6 +113,7 @@ public enum StandardErrorCode
     QUERY_PLANNING_TIMEOUT(0x0001_0027, INTERNAL_ERROR),
     NATIVE_EXECUTION_TASK_ERROR(0x0001_0028, INTERNAL_ERROR),
     UNSUPPORTED_ANALYZER_TYPE(0x0001_0029, INTERNAL_ERROR),
+    SLICE_TOO_LARGE(0x0001_0030, INTERNAL_ERROR),
```
NikhilCollooru, line 116: the error code should be `0x0001_002A`

elharo (commented):
Does this have an issue in the tracker? It feels release note worthy bug fix. like a

presto-common/src/main/java/com/facebook/presto/common/block/VariableWidthBlockBuilder.java
```diff
@@ -255,6 +259,8 @@ public BlockBuilder writeBytes(Slice source, int sourceIndex, int length)
     @Override
     public AbstractVariableWidthBlockBuilder writeBytes(byte[] source, int sourceIndex, int length)
     {
+        checkValidSliceRange(sourceIndex, length);
```
elharo, line 262: I'm a little confused. Maybe the argument names are off, but if sourceIndex is an index into the source byte[] argument, which the name sounds like it is, then it shouldn't matter. What matters is where we are in the target slice.

sourceIndex could be quite large, e.g. a billion, but if it's only writing the next 10 bytes and we have room for that in sliceOutput, why do we care? Am I miusunderstanding how this works?

elharo (changes requested):

ajaygeorge (commented):
some comments and questions.

presto-common/src/main/java/com/facebook/presto/common/block/VariableWidthBlockBuilder.java
```diff
@@ -244,9 +245,12 @@ public BlockBuilder writeLong(long value)
     @Override
     public BlockBuilder writeBytes(Slice source, int sourceIndex, int length)
     {
+        checkValidSliceRange(sourceIndex, length);
+
         if (!initialized) {
             initializeCapacity();
         }
+
         sliceOutput.writeBytes(source, sourceIndex, length);
```
elharo, line 254: Looking through the airlift code it seems to me that writeBytes already does this when it's trying to allocate more space. This seems redundant. I suppose this check will get called before the one in writeBytes, but does that change anything?

elharo (changes requested):

feilong-liu (commented):
Good change to make the error code more clear. I see the error handling is done for the writeBytes functions, what about other functions like writeLong, writeShort etc.?

presto-common/src/main/java/com/facebook/presto/common/block/VariableWidthBlockBuilder.java
swapsmagic: @kaikalur added benchmark numers.

presto-common/src/main/java/com/facebook/presto/common/block/VariableWidthBlockBuilder.java
swapsmagic: The existing code doesn't consider the scenario when sourceIndex + length crosses integer limit. This leads to a negative index and ended up with IndexOutOfBoundException. I am addressing that scenario.

presto-common/src/main/java/com/facebook/presto/common/block/VariableWidthBlockBuilder.java
swapsmagic: The arguments are what is supposed to be. The `writeBytes` method transfers the bytes from the source buffer to the SliceOutput buffer. The validation was missing where if the source buffer is larger than the SliceOutput buffer and crossing the integer limit than it overflows and becomes negative. In either case it fails, but in the former case we throw IndexOutOfBoundException which we are fixing here by catching it early and throwing the right exception.

presto-common/src/main/java/com/facebook/presto/common/block/BlockUtil.java
```diff
@@ -76,6 +77,13 @@ static void checkValidPosition(int position, int positionCount)
         }
     }
 
+    static void checkValidSliceRange(int sourceIndex, int length)
+    {
+        if (sourceIndex > MAX_ARRAY_SIZE - length) {
```
ajaygeorge, line 82: nit. might be good to leave a comment as to why we are subtraction instead of addition. 
Something like
// souceIndex + length check can overflow . Hence we are subtracting length from MAX_ARRAY_SIZE .

presto-common/src/main/java/com/facebook/presto/common/block/VariableWidthBlockBuilder.java
elharo: OK, got it. I see what you're doing now. I suggest chanigng the method name to checkValidSourceRange or something like that, since the range bing checked is in the source, not the target

steveburnett:
Suggest rephrasing of release note entry following the [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines). Maybe something like this?

```
== RELEASE NOTES ==

General Changes
* Improve VariableWidthBlockBuilder to check that it does not write more bytes than the maximum size of a slice
* Add error code SLICE_TOO_LARGE for SliceTooLargeException failures 

```

presto-common/src/main/java/com/facebook/presto/common/block/VariableWidthBlockBuilder.java
kaikalur: > @kaikalur added benchmark numbers.

Hmm more than 15% overhead? I guess since a lot of other checks are happening it's fine. Make sure to run some verifier suites and the overhead is tolerable

ajaygeorge: @kaikalur wondering why this check is expensive. https://github.com/prestodb/presto/pull/22063#discussion_r1513261003

kaikalur: > @kaikalur wondering why this check is expensive. [#22063 (comment)](https://github.com/prestodb/presto/pull/22063#discussion_r1513261003)

Good question! Not immediately clear why.

@swapsmagic - also run with TRY/CATCH and see what overhead that has.

swapsmagic: Benchmark with Try/Catch:
------------------
Result "com.facebook.presto.operator.aggregation.BenchmarkArrayAggregation.arrayAggregation":
  28.105 ±(99.9%) 0.938 ns/op [Average]
  (min, avg, max) = (26.503, 28.105, 29.512), stdev = 1.080
  CI (99.9%): [27.167, 29.043] (assumes normal distribution)

Run complete. Total time: 00:01:41

Benchmark                                    (type)  Mode  Cnt   Score   Error  Units
BenchmarkArrayAggregation.arrayAggregation  VARCHAR  avgt   20  28.105 ± 0.938  ns/op

Process finished with exit code 0

kaikalur: > ## Benchmark with Try/Catch:
> Result "com.facebook.presto.operator.aggregation.BenchmarkArrayAggregation.arrayAggregation": 28.105 ±(99.9%) 0.938 ns/op [Average] (min, avg, max) = (26.503, 28.105, 29.512), stdev = 1.080 CI (99.9%): [27.167, 29.043] (assumes normal distribution)
> 
> Run complete. Total time: 00:01:41
> 
> Benchmark (type) Mode Cnt Score Error Units BenchmarkArrayAggregation.arrayAggregation VARCHAR avgt 20 28.105 ± 0.938 ns/op
> 
> Process finished with exit code 0

Hmm - looks fishy. Try/catch is supposed to be zero cost unless the catch block is executed. These benchmark runs may not be reliable if you are doing something else on the side that might skew the result.

swapsmagic:
> Good change to make the error code more clear. I see the error handling is done for the writeBytes functions, what about other functions like writeLong, writeShort etc.?

Those methods don't have this issue and existing validations are sufficient to catch the issue.

kaikalur (approved):
Do some verifier runs

presto-common/src/main/java/com/facebook/presto/common/block/VariableWidthBlockBuilder.java
swapsmagic: Synced up with @kaikalur. Closing this.

presto-common/src/main/java/com/facebook/presto/common/block/VariableWidthBlockBuilder.java
swapsmagic: This range we are checking for the Slice where we are writing to make sure it's within the range of destination slice range. So named it accordingly.

NikhilCollooru (approved):

feilong-liu (approved):

tdcmeehan:
For release notes, perhaps we can rephrase more like the following (given audience of release notes are operators not software developers)

```
== RELEASE NOTES ==

General Changes
* Fix problem when writing large varchar values to throw a user error when it exceeds internal limits

```

`VariableWidthBlockBuilder` and `*Exception` are programming constructs which are difficult to translate for Presto operators.

PR #20384, opened 2023-07-26
DDL for Table Constraints

ClarenceThreepwood:
Introduces table constraints to Presto (Primary Key and Unique)(#20034). Support for other types of constraints will follow.

This PR Implements 
1. creating tables with unique and PK constraints
2. Adding unique and PK constraints to existing tables
3. Dropping existing constraints from tables

== RELEASE NOTES ==
General Changes
* Add framework to add and drop table constraints (Primary Key and Unique)

Hive Connector Changes
* Add support for table constraints. Other systems that support table constraints may extend this framework in their respective connectors to add support.

PR #22064, opened 2024-03-02
Support for NOT NULL column constraints

ClarenceThreepwood:
Follow up to #20384 

Adds support for NOT NULL column constraints in the create table statement (for the Hive connector)
Adds support for ALTER TABLE ALTER COLUMN set/drop NOT NULL

== RELEASE NOTES ==
General Changes
* Add framework to add and drop not null column constraints

Hive Connector Changes
* Add support for not null constraints. Other systems that support table constraints may extend this framework in their respective connectors to add support.


presto-docs/src/main/sphinx/sql/create-table.rst
```diff
@@ -61,12 +61,13 @@ Create a new table ``orders``::
     )
     WITH (format = 'ORC')
 
-Create the table ``orders`` if it does not already exist, adding a table comment
-and a column comment and a primary key constraint on column ``orderkey``::
+Create the table ``orders`` if it does not already exist, adding a table comment,
+a column comment, a not null constraint on column ``orderstatus`` and a primary
```
steveburnett, line 65:
```suggestion
a column comment, a ``not null`` constraint on column ``orderstatus``, and a primary
```
Same formatting suggestion, also use of the serial (Oxford) comma following the [Gitlab documentation style guide Punctuation topic](https://docs.gitlab.com/ee/development/documentation/styleguide/#punctuation).

steveburnett (changes requested):
Thanks for the docs! Looks good: a few nits, nothing else.

presto-docs/src/main/sphinx/sql/alter-table.rst
```diff
@@ -76,6 +77,18 @@ Drop constraint ``pk`` from the ``users`` table if table ``users`` exists and co
 
     ALTER TABLE IF EXISTS users DROP CONSTRAINT IF EXISTS pk;
 
+Add not null constraint to column ``zip`` in the ``users`` table::
+
+    ALTER TABLE users ALTER COLUMN zip SET NOT NULL;
+
+Add not null constraint to column ``zip`` in the ``users`` table if  table ``users`` and column ``zip`` exist::
```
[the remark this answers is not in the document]
ClarenceThreepwood: It appears to me that database schema entities are quoted in this text e.g. table name "users", column name "zip", constraint name "pk", etc. However properties such as table comment, column comment, the phrase primary key constraint, etc. are not. I think "not null constraint"  falls under the latter category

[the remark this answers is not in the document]
steveburnett: Okay, that makes sense. Thanks!

presto-docs/src/main/sphinx/sql/alter-table.rst
```diff
@@ -76,6 +77,18 @@ Drop constraint ``pk`` from the ``users`` table if table ``users`` exists and co
...
+
+    ALTER TABLE IF EXISTS users ALTER COLUMN IF EXISTS zip SET NOT NULL;
+
+Drop not null constraint from column ``zip`` in the ``users``table::
```
steveburnett, line 88:
```suggestion
Drop not null constraint from column ``zip`` in the ``users`` table::
```
Missing space before table breaks the formatting of ``users``.

steveburnett (changes requested):
Thanks for the update! Found one final nit on formatting.

steveburnett (dismissed):
LGTM! (docs)

Pull updated branch, new local build, no concerns. Thanks for the quick revisions!

presto-hive/src/main/java/com/facebook/presto/hive/HiveMetadata.java
```diff
@@ -647,7 +649,12 @@ private ConnectorTableMetadata getTableMetadata(Optional<Table> table, SchemaTab
             throw new PrestoException(HIVE_UNSUPPORTED_FORMAT, format("Not a Hive table '%s'", tableName));
         }
 
-        Function<HiveColumnHandle, ColumnMetadata> metadataGetter = columnMetadataGetter(table.get(), typeManager, metastoreContext.getColumnConverter());
+        List<TableConstraint<String>> tableConstraints = metastore.getTableConstraints(metastoreContext, tableName.getSchemaName(), tableName.getTableName());
+        List<String> notNullColumns = tableConstraints.stream()
+                .filter(NotNullConstraint.class::isInstance)
+                .map(c -> c.getColumns().stream().findFirst().orElseThrow(() -> new PrestoException(GENERIC_INTERNAL_ERROR, "NOT NULL constraint with no column")))
```
ZacBlanco, line 655: nit: use a non-abbreviated name

presto-main/src/main/java/com/facebook/presto/sql/rewrite/ShowQueriesRewrite.java
```diff
@@ -522,42 +522,55 @@ protected Node visitShowCreate(ShowCreate node, Void context)
 
                 ConnectorTableMetadata connectorTableMetadata = metadata.getTableMetadata(session, tableHandle.get()).getMetadata();
 
+                Set<String> notNullColumns = connectorTableMetadata.getTableConstraintsHolder().getTableConstraints()
+                        .stream()
+                        .filter(constraint -> (constraint instanceof NotNullConstraint))
```
ZacBlanco, line 527: nit: extra parentheses are not necessary here?

presto-main/src/main/java/com/facebook/presto/sql/rewrite/ShowQueriesRewrite.java
```diff
@@ -522,42 +522,55 @@ protected Node visitShowCreate(ShowCreate node, Void context)
...
+                                        UNIQUE,
+                                        constraint.isEnabled(),
+                                        constraint.isRely(),
+                                        constraint.isEnforced()));
+                            }
+                            return Optional.empty();
+                        })
+                        .filter(Optional::isPresent)
+                        .map(Optional::get)
+                        .map(e -> (ConstraintSpecification) e)
```
ZacBlanco, line 572: I prefer this form, but up to you

```suggestion
                        .map(ConstraintSpecification.class::cast)
```

presto-parser/src/main/antlr4/com/facebook/presto/sql/parser/SqlBase.g4
```diff
@@ -62,6 +62,10 @@ statement
         ADD constraintSpecification                                    #addConstraint
     | ALTER TABLE (IF EXISTS)? tableName=qualifiedName
         DROP CONSTRAINT (IF EXISTS)? name=identifier                   #dropConstraint
+    | ALTER TABLE (IF EXISTS)? tableName=qualifiedName
+        ALTER COLUMN (IF EXISTS)? column=identifier SET NOT NULL       #alterColumnSetNotNull
+    | ALTER TABLE (IF EXISTS)? tableName=qualifiedName
+        ALTER COLUMN (IF EXISTS)? column=identifier DROP NOT NULL      #alterColumnDropNotNull
```
ZacBlanco, line 68: pedantic, but according to the SQL 2016 spec, `COLUMN` is optional: https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#_11_12_alter_column_definition

Also, it seems `IF EXISTS` is not in the spec for `ALTER TABLE` or `ALTER COLUMN`.. It's probably fine to leave it though.

presto-parser/src/main/java/com/facebook/presto/sql/tree/AlterColumnNotNull.java
```diff
@@ -0,0 +1,121 @@
...
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        AlterColumnNotNull that = (AlterColumnNotNull) o;
+        return Objects.equals(table, that.table) &&
+                Objects.equals(column, that.column) &&
+                Objects.equals(dropConstraint, that.dropConstraint);
```
ZacBlanco, line 103: Have you intentionally left out `tableExists` and `columnExists`?

presto-parser/src/main/java/com/facebook/presto/sql/tree/AlterColumnNotNull.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    @Override
+    public int hashCode()
+    {
+        return Objects.hash(table, column, dropConstraint);
```
ZacBlanco, line 109: same comment

ZacBlanco (commented):
Just a few questions on the API and the grammar

presto-hive-metastore/src/main/java/com/facebook/presto/hive/metastore/thrift/ThriftHiveMetastore.java
```diff
@@ -287,6 +290,34 @@ public List<UniqueConstraint<String>> getUniqueConstraints(MetastoreContext meta
...
+    @Override
+    public List<NotNullConstraint<String>> getNotNullConstraints(MetastoreContext metastoreContext, String dbName, String tableName)
+    {
+        try {
+            Optional<NotNullConstraintsResponse> notNullConstraintsResponse = retry()
+                    .stopOnIllegalExceptions()
+                    .run("getNotNullConstraints", stats.getGetNotNullConstraints().wrap(() ->
+                            getMetastoreClientThenCall(metastoreContext, client -> client.getNotNullConstraints("hive", dbName, tableName))));
+
+            if (!notNullConstraintsResponse.isPresent() || notNullConstraintsResponse.get().getNotNullConstraints().size() == 0) {
```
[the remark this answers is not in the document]
elharo: as someone reading this for the first time, I find the original code far easier to follow than the rewritten "functional" code. I strongly prefer keeping the original.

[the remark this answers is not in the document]
ZacBlanco: I think the readability argument is subjective to one's exposure with functional programming paradigms.

I am merely suggesting to use the idiomatic functional style with the `Optional`, because the point of having optional is to perform operations like this without having to write conditionals which tend to be subject to programmer-error. Using `Optional`'s API prevents relying on an IDEs advance syntax checking features to warn whether or not an `isPresent` check is performed before a `get`, etc. By using the API you guarantee that you never get a runtime exception due to a missing `Optional.isPresent` check.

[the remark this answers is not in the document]
elharo: I'm afraid I disagree. The problem is not one's exposure to functional programming paradigms. It is the baroque design of Java's pseudo-functional APIs bolted on to the language years too late. Pretending Java is a functional language like Haskell or XSLT is a recipe for obfuscation.

presto-hive-metastore/src/main/java/com/facebook/presto/hive/metastore/file/FileHiveMetastore.java
```diff
@@ -1086,13 +1087,17 @@ public MetastoreOperationResult addConstraint(MetastoreContext metastoreContext,
...
                 tableConstraint = new UniqueConstraint(Optional.of(randomUUID().toString()),
                         tableConstraint.getColumns(),
                         tableConstraint.isEnabled(),
                         tableConstraint.isRely(),
                         tableConstraint.isEnforced());
             }
+            else if (tableConstraint instanceof NotNullConstraint) {
+                tableConstraint = new NotNullConstraint(Optional.of(randomUUID().toString()),
+                        tableConstraint.getColumns());
+            }
```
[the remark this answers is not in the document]
ClarenceThreepwood: Added to be consistent with ThriftHiveMetastore

presto-hive/src/main/java/com/facebook/presto/hive/HiveMetadata.java
```diff
@@ -3724,11 +3741,31 @@ public TableLayoutFilterCoverage getTableLayoutFilterCoverage(ConnectorTableLayo
     }
 
     @Override
-    public void dropConstraint(ConnectorSession session, ConnectorTableHandle tableHandle, String constraintName)
+    public void dropConstraint(ConnectorSession session, ConnectorTableHandle tableHandle, Optional<String> constraintName, Optional<String> columnName)
```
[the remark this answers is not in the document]
ClarenceThreepwood: expanded the check below to only allow one argument

presto-parser/src/main/java/com/facebook/presto/sql/tree/AlterColumnNotNull.java
ClarenceThreepwood: Yes. I'm not quite sure where this comparison code is called, but we always seem to omit the exists flags. See RenameColumn, AddColumn, RenameTable, etc.

rschlussel: @ClarenceThreepwood  it looks like you added these exists fields to those classes https://github.com/prestodb/presto/pull/15003 (via cherry-picking a trino commit). Are you sure it wasn't just an oversight?

presto-parser/src/main/antlr4/com/facebook/presto/sql/parser/SqlBase.g4
ClarenceThreepwood: Thanks for catching this. Fixed

presto-parser/src/main/java/com/facebook/presto/sql/tree/AlterColumnNotNull.java
ClarenceThreepwood: That is true, and it was probably an oversight 4 years ago. Do you think these fields should be added? I can add these for all the relevant parse nodes if so

ClarenceThreepwood: Added

presto-hive-metastore/src/main/java/com/facebook/presto/hive/metastore/file/FileHiveMetastore.java
```diff
@@ -1087,13 +1088,20 @@ public MetastoreOperationResult addConstraint(MetastoreContext metastoreContext,
...
+            else {
+                throw new PrestoException(NOT_SUPPORTED, "This connector can only handle Unique/Primary Key constraints at this time");
```
aaneja, line 1103: nit : Fix the exeception message to include 'can only handle Primary, Unique key and Not Nulll constraints '

presto-hive-metastore/src/main/java/com/facebook/presto/hive/metastore/file/FileHiveMetastore.java
```diff
@@ -1102,10 +1110,25 @@ public MetastoreOperationResult addConstraint(MetastoreContext metastoreContext,
 
     public List<TableConstraint<String>> getTableConstraints(MetastoreContext metastoreContext, String schemaName, String tableName)
     {
-        return readConstraintsFile(schemaName, tableName).stream()
+        Set<TableConstraint> rawConstraints = readConstraintsFile(schemaName, tableName);
+
+        List<TableConstraint<String>> constraints = rawConstraints.stream()
```
aaneja, line 1115: Instead of streaming 3 times, how about we just forEach over the rawConstraints and check and add the constraints to a list builder? Or is it a hard requirement that we add the PK, then unique and then the Not Null constraints ?

ClarenceThreepwood: It leads to non-determinism in the tests if these are not in order. Since this is the file-based metastore anyway I am not too concerned about the performance overhead

aaneja (approved):

ZacBlanco (approved):

tdcmeehan (commented):
Mostly nits

rschlussel (commented):
I have a concern about the alter column support. NOT NULL constraints are being treated differently (with different syntax) compared to all other constraints, but I think we should be consistent across all our constraints.

nit: noticed some of your commit titles are too long.

presto-hive/src/main/java/com/facebook/presto/hive/HiveMetadata.java
```diff
@@ -649,7 +651,12 @@ private ConnectorTableMetadata getTableMetadata(Optional<Table> table, SchemaTab
             throw new PrestoException(HIVE_UNSUPPORTED_FORMAT, format("Not a Hive table '%s'", tableName));
         }
 
-        Function<HiveColumnHandle, ColumnMetadata> metadataGetter = columnMetadataGetter(table.get(), typeManager, metastoreContext.getColumnConverter());
+        List<TableConstraint<String>> tableConstraints = metastore.getTableConstraints(metastoreContext, tableName.getSchemaName(), tableName.getTableName());
+        List<String> notNullColumns = tableConstraints.stream()
+                .filter(NotNullConstraint.class::isInstance)
+                .map(constraint -> constraint.getColumns().stream().findFirst().orElseThrow(() -> new PrestoException(GENERIC_INTERNAL_ERROR, "NOT NULL constraint with no column")))
```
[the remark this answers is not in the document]
rschlussel: can you have a constraint without a column?  Should this be enforced when we create the TableConstraints? 

Also, related to Tim's question, would this indicate some kind of problem with the metastore (that it's giving us something invalid), and if so would it be an external error?

presto-spi/src/main/java/com/facebook/presto/spi/constraints/NotNullConstraint.java
```diff
@@ -0,0 +1,55 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.Collections;
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Optional;
+
+import static java.util.stream.Collectors.toCollection;
+
+public class NotNullConstraint<T>
```
rschlussel, line 26: Can you explain why this is parametrized?  When would it be anything other than a String.

presto-hive-metastore/src/main/java/com/facebook/presto/hive/metastore/file/FileHiveMetastore.java
```diff
@@ -1102,10 +1110,25 @@ public MetastoreOperationResult addConstraint(MetastoreContext metastoreContext,
...
                 .map(constraint -> (TableConstraint<String>) constraint)
+                .filter(constraint -> constraint instanceof PrimaryKeyConstraint)
+                .collect(toList());
+
+        constraints.addAll(rawConstraints.stream()
```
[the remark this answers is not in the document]
ClarenceThreepwood: Added a validate method to TableConstraintsHolder to perform this check in one place

presto-hive/src/main/java/com/facebook/presto/hive/HiveMetadata.java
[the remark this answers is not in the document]
ClarenceThreepwood: > can you have a constraint without a column? Should this be enforced when we create the TableConstraints?
> 
The grammar does not allow you to create a constraint without a specifying a set of columns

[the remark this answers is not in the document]
tdcmeehan: @ClarenceThreepwood for such things we use `ErrorType` `EXTERNAL`.

presto-parser/src/main/antlr4/com/facebook/presto/sql/parser/SqlBase.g4
```diff
@@ -62,6 +62,10 @@ statement
         ADD constraintSpecification                                    #addConstraint
     | ALTER TABLE (IF EXISTS)? tableName=qualifiedName
         DROP CONSTRAINT (IF EXISTS)? name=identifier                   #dropConstraint
+    | ALTER TABLE (IF EXISTS)? tableName=qualifiedName
```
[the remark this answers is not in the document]
rschlussel: I see, thanks for the explanation!

presto-spi/src/main/java/com/facebook/presto/spi/constraints/TableConstraintsHolder.java
```diff
@@ -14,25 +14,39 @@
...
 import java.util.Optional;
 
+import static com.facebook.presto.spi.StandardErrorCode.NOT_SUPPORTED;
+import static java.lang.String.format;
 import static java.util.Collections.emptyList;
 import static java.util.Objects.requireNonNull;
 
 public class TableConstraintsHolder
 {
+    public static void validateTableConstraints(Collection<TableConstraint<String>> constraints)
```
rschlussel, line 33: methods go after member variables.  See ordering in our codestyle: https://github.com/prestodb/presto/blob/master/CONTRIBUTING.md#code-style

presto-hive/src/main/java/com/facebook/presto/hive/HiveMetadata.java
```diff
@@ -649,7 +651,14 @@ private ConnectorTableMetadata getTableMetadata(Optional<Table> table, SchemaTab
             throw new PrestoException(HIVE_UNSUPPORTED_FORMAT, format("Not a Hive table '%s'", tableName));
         }
 
-        Function<HiveColumnHandle, ColumnMetadata> metadataGetter = columnMetadataGetter(table.get(), typeManager, metastoreContext.getColumnConverter());
+        List<TableConstraint<String>> tableConstraints = metastore.getTableConstraints(metastoreContext, tableName.getSchemaName(), tableName.getTableName());
+        List<String> notNullColumns = tableConstraints.stream()
+                .filter(NotNullConstraint.class::isInstance)
+                .map(constraint -> constraint.getColumns().stream()
+                        .findFirst()
+                        .orElseThrow(() -> new PrestoException(GENERIC_INTERNAL_ERROR, "NOT NULL constraint with no column")))
```
rschlussel, line 659: Let's throw an external error here since the error is caused by invalid values we're getting from the metastore.  Also let's add the table name to the error message to help with debugging for queries that access multiple tables.

presto-hive/src/main/java/com/facebook/presto/hive/HiveMetadata.java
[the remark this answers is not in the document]
ClarenceThreepwood: Thanks for the pointer - changed the errorcode to HIVE_METASTORE_ERROR of type EXTERNAL

presto-spi/src/main/java/com/facebook/presto/spi/constraints/TableConstraintsHolder.java
ClarenceThreepwood: Fixed. The doc is not clear about static methods. I will update the code-style to make it more obvious

presto-spi/src/main/java/com/facebook/presto/spi/constraints/TableConstraintsHolder.java
```diff
@@ -29,10 +33,20 @@ public class TableConstraintsHolder
     private final List<TableConstraint<String>> tableConstraints;
     private final Map<String, ColumnHandle> columnNameToHandleAssignments;
 
+    public static void validateTableConstraints(Collection<TableConstraint<String>> constraints)
```
rschlussel, line 36: I would move this below the constructor because 
1. we usually have constructors first 
2. we put method definitions below where they get called.

rschlussel (dismissed):

ClarenceThreepwood:
Thanks @rschlussel. I've also pushed the final change you suggested

rschlussel (approved):

PR #22095, opened 2024-03-05
Add function inlining in a later optimizer pass

### LINKED ISSUE #22085: Fix bug when KEY_BASED_SAMPLING_ENABLED=true and inline_sql_functions=true
[key_sampling_percent](https://github.com/prestodb/presto/blob/master/presto-main/src/main/java/com/facebook/presto/operator/scalar/sql/SimpleSamplingPercent.java) is a SQL function.
When we set `inline_sql_functions=true`, we expect the coordinator would always inline this function and we would not need an implementation in the worker eval (Prestissimo in this case).
However, KeyBasedSampler runs after sql functions have already been inlined, so the function doesn't get inlined. 

## Possible Solution
Fix is to add another pass of inline sql functions at the end.

## Steps to Reproduce

presto:tpch> set session inline_sql_functions=true;
presto:tpch> set session KEY_BASED_SAMPLING_ENABLED=true;
presto:tpch> select count(1) from orders join lineitem using(orderkey);

Query 20240304_044118_00007_4g98f failed:  Scalar function name not registered: presto.default.key_sampling_percent, called with arguments: (VARCHAR).

ZacBlanco:
## Description

Adds a function inlining pass to the optimizer after the KeyBasedSampling rule in order to make sure that any rewritten functions are inlined.

## Motivation and Context

Fixes #22085

## Impact

N/A

## Test Plan

Added a test case

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/sql/planner/PlanOptimizers.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/planner/PlanOptimizers.java
@@ -634,7 +634,15 @@ public class PlanOptimizers
                         ruleStats,
                         statsCalculator,
                         estimatedExchangesCostCalculator,
-                        ImmutableSet.of(new RemoveRedundantIdentityProjections())));
+                        ImmutableSet.of(new RemoveRedundantIdentityProjections())),
+                new IterativeOptimizer(
+                        metadata,
+                        ruleStats,
+                        statsCalculator,
+                        estimatedExchangesCostCalculator,
+                        ImmutableSet.<Rule<?>>builder()
+                                .addAll(new InlineSqlFunctions(metadata, sqlParser).rules())
+                                .build()));
 
         builder.add(
                 new IterativeOptimizer(
--- a/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeGeneralQueries.java
+++ b/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeGeneralQueries.java
@@ -28,7 +28,9 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.UUID;
 
+import static com.facebook.presto.SystemSessionProperties.INLINE_SQL_FUNCTIONS;
 import static com.facebook.presto.SystemSessionProperties.JOIN_DISTRIBUTION_TYPE;
+import static com.facebook.presto.SystemSessionProperties.KEY_BASED_SAMPLING_ENABLED;
 import static com.facebook.presto.common.type.BigintType.BIGINT;
 import static com.facebook.presto.common.type.VarcharType.VARCHAR;
 import static com.facebook.presto.hive.HiveStorageFormat.DWRF;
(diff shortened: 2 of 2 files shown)


presto-main/src/main/java/com/facebook/presto/sql/planner/PlanOptimizers.java
```diff
@@ -634,7 +634,15 @@ public PlanOptimizers(
...
                         estimatedExchangesCostCalculator,
-                        ImmutableSet.of(new RemoveRedundantIdentityProjections())));
+                        ImmutableSet.of(new RemoveRedundantIdentityProjections())),
+                new IterativeOptimizer(
+                        metadata,
+                        ruleStats,
+                        statsCalculator,
+                        estimatedExchangesCostCalculator,
+                        ImmutableSet.<Rule<?>>builder()
+                                .addAll(new InlineSqlFunctions(metadata, sqlParser).rules())
```
ClarenceThreepwood, line 644: Do we still need the earlier pass?

ZacBlanco: I wasn't sure at first but I briefly skimmed all the rules between the original and new location and I feel pretty confident it can be moved rather than copied

tdcmeehan: We might need to keep both--if we don't inline SQL functions at the beginning, then we might lose downstream opportunities to optimize them.

aditi-pandit: Agree that keeping the earlier pass is better.

majetideepak:
@ZacBlanco can you confirm the test you added reproduces the bug without your fix?

ZacBlanco:
I can confirm that the test does not pass without this change

presto-main/src/main/java/com/facebook/presto/sql/planner/PlanOptimizers.java
```diff
@@ -855,13 +863,13 @@ public PlanOptimizers(
         // Optimizers above this do not need to care about aggregations with the type other than SINGLE
         // This optimizer must be run after all exchange-related optimizers
         builder.add(new IterativeOptimizer(
-                metadata,
```
tdcmeehan, line 858: Unrelated change?

tdcmeehan (approved):

presto-main/src/main/java/com/facebook/presto/sql/planner/PlanOptimizers.java
ZacBlanco: IntelliJ formatter didn't like this indentation for some reason. Fixed

tdcmeehan (dismissed):

majetideepak (dismissed):
Thanks, @ZacBlanco

tdcmeehan (approved):

rschlussel (approved):

PR #21951, opened 2024-02-16
Add buffered bytes metadata to task results

tdcmeehan:
## Description
 
Add remaining buffered bytes to task results endpoints.  Add a new HEAD request to task results to allow for a lightweight way to retrieve header-only information.

## Motivation and Context
See #21926 for context and motivation.

## Impact
This will allow the design in #21926 to be more RESTful and reduce the need for special-case flags.

## Test Plan
Unit test coverage.

## Contributor checklist

- [ ] Please make sure your submission complies with our [development](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#development), [formatting](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#formatting), [commit message](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#commit-formatting-and-pull-requests), and [attribution guidelines](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#attribution).
- [ ] PR description addresses the issue accurately and concisely.  If the change is non-trivial, a GitHub Issue is referenced.
- [ ] Documented new properties (with its default value), SQL syntax, functions, or other functionality.
- [ ] If release notes are required, they follow the [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines).
- [ ] Adequate tests were added if applicable.
- [ ] CI passed.

## Release Notes

If release note is NOT required, use:

```
== NO RELEASE NOTE ==
```

PR #22096, opened 2024-03-05
[native] Allow single zero to be passed as X-Presto-Buffer-Remaining-Bytes

Yuhta:
To be compatible with https://github.com/prestodb/presto/pull/21951


mbasmanova (approved):
Thanks.

PR #22102, opened 2024-03-06
[native] Change shuffle maxWait unit

Yuhta:
We want to use 100ms for data fetch max wait, so need to change the unit here. All changes are functionally equivalent to previous version.

### DIFF
--- a/presto-native-execution/presto_cpp/main/PrestoExchangeSource.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoExchangeSource.cpp
@@ -132,7 +132,7 @@ bool PrestoExchangeSource::shouldRequestLocked() {
 
 folly::SemiFuture<PrestoExchangeSource::Response> PrestoExchangeSource::request(
     uint32_t maxBytes,
-    uint32_t maxWaitSeconds) {
+    std::chrono::microseconds maxWait) {
   // Before calling 'request', the caller should have called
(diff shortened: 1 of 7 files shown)


presto-native-execution/presto_cpp/main/PrestoExchangeSource.cpp
```diff
@@ -132,7 +132,7 @@ bool PrestoExchangeSource::shouldRequestLocked() {
 
 folly::SemiFuture<PrestoExchangeSource::Response> PrestoExchangeSource::request(
     uint32_t maxBytes,
-    uint32_t maxWaitSeconds) {
+    std::chrono::microseconds maxWait) {
```
aditi-pandit, line 135: Nit : Rename variable to maxWaitMcs (seems like other time variables in this code have shorthand for their unit in the name).

aditi-pandit (commented):
Thanks @Yuhta. Just had minor nit.

presto-native-execution/presto_cpp/main/PrestoExchangeSource.cpp
Yuhta: We had the unit because we were using integer, now we encode the unit in the type itself so it is no longer needed

aditi-pandit (approved):
Thanks @Yuhta. Looking forward to all the exchange improvements for Prestissimo.

Issue #22078, opened 2024-03-04
Row ID support


elharo:
Row ID is a binary key for a row that is globally unique across the warehouse. It identifies a specific row at a specific version of a partition’s data. A row_id will continue to locate the same row in the warehouse until there is a mutation to the underlying partition. 

## Expected Behavior or Use Case

Row IDs resolve to the same value across compute engines. In this way they can be used as a strong-equality comparison across compute engines.

## Possible Implementation

row_number:row_group_id:data_version:partition_id:tbl_id:table_guid.substring(0, 6):spec_version

Row IDs will be binary encoded. There is no delimiter in the actual row_id. It’s just binary compressed fixed width fields (except for the variable width row_group_id)

data_version:partition_id:tbl_id:table_guid.substring(0, 6):spec_version has the same value for all rows in a partition 

The XLDB reader libraries will supply  row_number:row_group_id as independent pieces  accessed alongside rows.

data_version:partition_id:tbl_id:table_guid.substring(0, 6):spec_version: This component will be exposed as a binary field in the Partition object. It currently is fixed width 26 bytes. 

In essence there are two pieces: one to locate the partition and one to locate the row within the partition. The partition ID is passed into the reader library from above while the reader supplies the row ID within the partition. The reader passes back the row ID alongside the other columns. 

## Context

A row_id would introduce an automatic skew mitigation, giving each row a unique key to be shuffled on.

tdcmeehan:
In order for a row ID to be interpreted the same between engines, there must be a spec or shared library that defines how to derive a row ID given the table and file formats used to read the data.  Hence, I think a row ID might be considered something that is connector-specific: each data source will likely have different conventions on how to define a row ID.  Some data sources may have no clear explicit convention, and should not expose a row ID.  Only connectors whose sources have a row ID convention defined in a spec or shared library should expose it.

For those that do, it seems that we should use the constraints framework to convey uniqueness of the column, as this particular column is just a special case of general uniqueness constraint.  CC: @ClarenceThreepwood

elharo:
There are some Meta-internal details I'm having to redact in this proposal, so my apologies if this isn't as clear as it should be. I do want this feature to be generally useful though, and certainly not to cause any problems for any other Presto users. Also, this is all still new so nothing is written in stone yet and changes are very much possible.

[...]

PR #22103, opened 2024-03-06
Support binary IDs for Hive partitions

elharo:
## Description
Part of the row ID plans for #22078

## Motivation and Context
Adds a place for the partition component of the row ID, should a warehouse choose to supply one.

## Impact
Hive partitions have an id field. 

## Test Plan
./mvnw test -pl :presto-hive-metastore

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-hive-metastore/src/main/java/com/facebook/presto/hive/metastore/Partition.java
+++ b/presto-hive-metastore/src/main/java/com/facebook/presto/hive/metastore/Partition.java
@@ -22,6 +22,7 @@ import com.google.common.collect.ImmutableMap;
 
 import javax.annotation.concurrent.Immutable;
 
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -45,6 +46,7 @@ public class Partition
     private final boolean sealedPartition;
     private final int createTime;
     private final long lastDataCommitTime;
+    private final Optional<byte[]> rowIdPartitionComponent;
 
     @JsonCreator
     public Partition(
@@ -58,7 +60,8 @@ public class Partition
             @JsonProperty("eligibleToIgnore") boolean eligibleToIgnore,
             @JsonProperty("sealedPartition") boolean sealedPartition,
             @JsonProperty("createTime") int createTime,
-            @JsonProperty("lastDataCommitTime") long lastDataCommitTime)
+            @JsonProperty("lastDataCommitTime") long lastDataCommitTime,
+            @JsonProperty("rowIdPartitionComponent") Optional<byte[]> rowIdPartitionComponent)
     {
         this.databaseName = requireNonNull(databaseName, "databaseName is null");
         this.tableName = requireNonNull(tableName, "tableName is null");
@@ -71,6 +74,7 @@ public class Partition
         this.sealedPartition = sealedPartition;
         this.createTime = createTime;
         this.lastDataCommitTime = lastDataCommitTime;
+        this.rowIdPartitionComponent = requireNonNull(rowIdPartitionComponent);
     }
 
     @JsonProperty
@@ -145,6 +149,20 @@ public class Partition
         return lastDataCommitTime;
     }
 
+    /**
+     * A unique identifier for a specific version of a
+     * specific partition in a specific table.
+     */
+    @JsonProperty
+    public Optional<byte[]> getRowIdPartitionComponent()
+    {
+        if (rowIdPartitionComponent.isPresent()) {
+            byte[] copy = Arrays.copyOf(rowIdPartitionComponent.get(), rowIdPartitionComponent.get().length);
+            return Optional.of(copy);
+        }
+        return Optional.empty();
+    }
+
     @Override
     public String toString()
     {
@@ -208,6 +226,7 @@ public class Partition
         private boolean isSealedPartition = true;
         private int createTime;
         private long lastDataCommitTime;
+        private Optional<byte[]> rowIdPartitionComponent = Optional.empty();
 
         private Builder()
         {
@@ -226,6 +245,7 @@ public class Partition
             this.isEligibleToIgnore = partition.isEligibleToIgnore();
             this.createTime = partition.getCreateTime();
             this.lastDataCommitTime = partition.getLastDataCommitTime();
+            this.rowIdPartitionComponent = partition.getRowIdPartitionComponent();
         }
 
         public Builder setDatabaseName(String databaseName)
@@ -299,9 +319,20 @@ public class Partition
             return this;
         }
 
+        /**
+         * Sets a unique identifier for a specific version of a
+         * specific partition in a specific table. This value will normally be
+         * supplied by Metastore, along with other metadata.
+         */
+        public Builder setRowIdPartitionComponent(Optional<byte[]> rowIdPartitionComponent)
+        {
+            this.rowIdPartitionComponent = rowIdPartitionComponent;
+            return this;
+        }
+
         public Partition build()
         {
-            return new Partition(databaseName, tableName, values, storageBuilder.build(), columns, parameters, partitionVersion, isEligibleToIgnore, isSealedPartition, createTime, lastDataCommitTime);
+            return new Partition(databaseName, tableName, values, storageBuilder.build(), columns, parameters, partitionVersion, isEligibleToIgnore, isSealedPartition, createTime, lastDataCommitTime, rowIdPartitionComponent);
         }
     }
 }
--- a/presto-hive-metastore/src/main/java/com/facebook/presto/hive/metastore/file/PartitionMetadata.java
+++ b/presto-hive-metastore/src/main/java/com/facebook/presto/hive/metastore/file/PartitionMetadata.java
@@ -56,6 +56,7 @@ public class PartitionMetadata
     private final Map<String, HiveColumnStatistics> columnStatistics;
     private final boolean eligibleToIgnore;
     private final boolean sealedPartition;
+    private final Optional<byte[]> rowIdPartitionComponent;
 
     @JsonCreator
     public PartitionMetadata(
@@ -69,7 +70,8 @@ public class PartitionMetadata
             @JsonProperty("externalLocation") Optional<String> externalLocation,
             @JsonProperty("columnStatistics") Map<String, HiveColumnStatistics> columnStatistics,
             @JsonProperty("eligibleToIgnore") boolean eligibleToIgnore,
-            @JsonProperty("sealedPartition") boolean sealedPartition)
+            @JsonProperty("sealedPartition") boolean sealedPartition,
+            @JsonProperty("rowIdPartitionComponent") Optional<byte[]> rowIdPartitionComponent)
     {
         this.columns = ImmutableList.copyOf(requireNonNull(columns, "columns is null"));
         this.parameters = ImmutableMap.copyOf(requireNonNull(parameters, "parameters is null"));
@@ -83,6 +85,7 @@ public class PartitionMetadata
         this.columnStatistics = ImmutableMap.copyOf(requireNonNull(columnStatistics, "columnStatistics is null"));
         this.eligibleToIgnore = eligibleToIgnore;
         this.sealedPartition = sealedPartition;
+        this.rowIdPartitionComponent = requireNonNull(rowIdPartitionComponent);
     }
 
     @Deprecated
@@ -96,7 +99,8 @@ public class PartitionMetadata
             Optional<String> externalLocation,
             Map<String, HiveColumnStatistics> columnStatistics,
             boolean eligibleToIgnore,
-            boolean sealedPartition)
+            boolean sealedPartition,
+            Optional<byte[]> rowIdPartitionComponent)
     {
         this(
                 columns,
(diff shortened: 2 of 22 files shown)


elharo:
2024-03-06T21:58:00.8591018Z [ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:testCompile (default-testCompile) on project presto-hive-metastore: Compilation failure
2024-03-06T21:58:00.8595815Z [ERROR] /home/runner/work/presto/presto/presto-hive-metastore/src/test/java/com/facebook/presto/hive/metastore/TestRecordingHiveMetastore.java:[108,48] constructor Partition in class com.facebook.presto.hive.metastore.Partition cannot be applied to given types;
2024-03-06T21:58:00.8601112Z [ERROR]   required: java.lang.String,java.lang.String,java.util.List<java.lang.String>,com.facebook.presto.hive.metastore.Storage,java.util.List<com.facebook.presto.hive.metastore.Column>,java.util.Map<java.lang.String,java.lang.String>,java.util.Optional<java.lang.Long>,boolean,boolean,int,long,byte[]
2024-03-06T21:58:00.8645732Z [ERROR]   found: java.lang.String,java.lang.String,com.google.common.collect.ImmutableList<java.lang.String>,com.facebook.presto.hive.metastore.Storage,com.google.common.collect.ImmutableList<com.facebook.presto.hive.metastore.Column>,com.google.common.collect.ImmutableMap<java.lang.String,java.lang.String>,java.util.Optional<java.lang.Object>,boolean,boolean,int,int

elharo:
Failrues starting to look flkay:


2024-03-08T20:30:51.5502068Z ERROR: for conf_hadoop-master_1  Cannot start service hadoop-master: driver failed programming external connectivity on endpoint conf_hadoop-master_1 (b89f75e51118e98a7f8b8021eafabdfa0b1018fe1e83f550d3813e3b4b976d62): Error starting userland proxy: listen tcp4 0.0.0.0:50070: bind: address already in use

tdcmeehan:
Can this PR also include the code that populates the binary ID?  And can there be some documentation on under what circumstances this ID is populated?

elharo:
I'm going to draft an RFC for all of this. I should have done that first, but I just discovered that we had an RFC process this morning. :-)

elharo:
> Can this PR also include the code that populates the binary ID? And can there be some documentation on under what circumstances this ID is populated?

I think the current answer, though I am still working this out and might have missed something, is that the code that populates the binary ID cannot yet be included. ATM it comes from the Meta-internal Metastore client which is not in this repo. And that leads to the question of under what circumstances is this ID populated? I think the answer to that is that if the library that is creating Partition objects by calling Metastore and using the results of that is getting a binary ID from Metastore and stuffing that in here via Partition.Builder#setId (or perhaps by a direct constructor call), then the binary ID will be set. However, nothing in the open source repo at this time is going to put anything in here.

This makes sense because different warehouses are going to need different formats of binary partition IDs (might rename that rowComponentId or something like that, by the way) that work for their warehouses. 

Let me see if I can work that into some Javadoc.

elharo:
OK, this one might actually be a non-flaky genuine failure in this PR:

```
2024-03-13T21:20:02.0668862Z [ERROR] Failures: 
2024-03-13T21:20:02.0671842Z [ERROR]   TestHiveSplit.testJsonRoundTrip:147 expected [Optional[[B@6b71d0bf]] but found [Optional[[B@1c778b61]]
2024-03-13T21:20:02.0672983Z [INFO] 
2024-03-13T21:20:02.0673526Z [ERROR] Tests run: 2943, Failures: 1, Errors: 0, Skipped: 90
2024-03-13T21:20:02.0674232Z [INFO] 
2024-03-13T21:20:02.0711728Z [INFO]
```

presto-hive-metastore/src/main/java/com/facebook/presto/hive/metastore/Partition.java
```diff
@@ -145,6 +179,20 @@ public long getLastDataCommitTime()
         return lastDataCommitTime;
     }
 
+    /**
+     * A warehouse-wide unique identifier for a specific version of a
+     * specific partition in a specific table.
+     */
+    @JsonProperty
```
sdruzkin, line 186: Does @JsonProperty work well with Optional?

sdruzkin (approved):
Code LGTM. Please update the PR description.

presto-hive-metastore/src/main/java/com/facebook/presto/hive/metastore/Partition.java
elharo: It should as long as the JDK8 module is installed in the mapper. Tests for this pass. We could use null instead though I'm told that Optional is preferred in Presto. Note that partitionVersion is already declared Optional and is an `@JsonProperty` in this class so I think this is safe.

kaikalur (dismissed):

sdruzkin (approved):
LGTM

presto-hive-metastore/src/main/java/com/facebook/presto/hive/metastore/Partition.java
```diff
@@ -68,9 +71,13 @@ public Partition(
         this.parameters = ImmutableMap.copyOf(requireNonNull(parameters, "parameters is null"));
         this.partitionVersion = requireNonNull(partitionVersion, "partitionVersion is null");
         this.eligibleToIgnore = eligibleToIgnore;
-        this.sealedPartition = sealedPartition;
+        this.sealedPartition = sealedPartition; // foo
```
sdruzkin, line 74: remove

elharo: done

presto-hive-metastore/src/main/java/com/facebook/presto/hive/metastore/Partition.java
```diff
@@ -145,6 +152,20 @@ public long getLastDataCommitTime()
         return lastDataCommitTime;
     }
 
+    /**
+     * A warehouse-wide unique identifier for a specific version of a
```
rschlussel, line 156:
```suggestion
     * A unique identifier for a specific version of a
```

presto-hive-metastore/src/main/java/com/facebook/presto/hive/metastore/Partition.java
```diff
@@ -299,9 +322,20 @@ public Builder setLastDataCommitTime(long lastDataCommitTime)
             return this;
         }
 
+        /**
+         * Sets a warehouse-wide unique identifier for a specific version of a
```
rschlussel, line 326:
```suggestion
         * Sets a unique identifier for a specific version of a
```

presto-hive-metastore/src/main/java/com/facebook/presto/hive/metastore/file/PartitionMetadata.java
```diff
@@ -83,6 +85,7 @@ public PartitionMetadata(
         this.columnStatistics = ImmutableMap.copyOf(requireNonNull(columnStatistics, "columnStatistics is null"));
         this.eligibleToIgnore = eligibleToIgnore;
         this.sealedPartition = sealedPartition;
+        this.rowIdPartitionComponent = rowIdPartitionComponent;
```
rschlussel, line 88: requireNonNull

presto-hive/src/main/java/com/facebook/presto/hive/HivePartitionMetadata.java
```diff
@@ -27,21 +27,24 @@ public class HivePartitionMetadata
...
-            Set<ColumnHandle> redundantColumnDomains)
+            Set<ColumnHandle> redundantColumnDomains,
+            Optional<byte[]> rowIdPartitionComponent)
     {
         this.partition = requireNonNull(partition, "partition is null");
         this.hivePartition = requireNonNull(hivePartition, "hivePartition is null");
         this.tableToPartitionMapping = requireNonNull(tableToPartitionMapping, "tableToPartitionMapping is null");
         this.encryptionInformation = requireNonNull(encryptionInformation, "encryptionInformation is null");
         this.redundantColumnDomains = requireNonNull(redundantColumnDomains, "redundantColumnDomains is null");
+        this.rowIdPartitionComponent = rowIdPartitionComponent == null ? Optional.empty() : rowIdPartitionComponent;
```
rschlussel, line 47: As best i can tell, we should never pass in a null value here.  enforce that in the code and call requireNonNull instead.

presto-hive-metastore/src/main/java/com/facebook/presto/hive/metastore/Partition.java
```diff
@@ -71,6 +74,10 @@ public Partition(
         this.sealedPartition = sealedPartition;
         this.createTime = createTime;
         this.lastDataCommitTime = lastDataCommitTime;
+        if (rowIdPartitionComponent == null) {
```
rschlussel, line 77: if the concern about null here is due to the serialized object, i believe jackson handles that already (if it's an optional type, serializes empty as null, and reads it back as an empty object).

rschlussel (commented):
a few small comments, but general approach seems fine.

rschlussel (commented):
looks good, just squash the commits and update the commit message to follow the commit guidelines (e.g. `Add binary id to Hive Partition` and then provide more context in the commit body) .  Also, you're welcome to not create those intermediate commits in the first place, github can show you the diff between version in the pr even if it's not in a separate commit.

elharo:
Are you suggesting amending every change as a PR is developed? Or is there another way I'm missing?

rschlussel (approved):

rschlussel:
> Are you suggesting amending every change as a PR is developed? Or is there another way I'm missing?

That's what I normally do.

presto-hive/src/test/java/com/facebook/presto/hive/TestHiveSplitManager.java
```diff
@@ -459,24 +459,13 @@ private void assertRedundantColumnDomains(Range predicateRange, PartitionStatist
...
-                                ImmutableMap.of(),
-                                ImmutableMap.of()),
-                        COLUMNS,
-                        ImmutableMap.of(),
-                        Optional.empty(),
-                        false,
-                        true,
-                        0,
-                        0),
+                new Partition("test_db", "test_table", ImmutableList.of(PARTITION_VALUE), new Storage(
```
tdcmeehan, line 462: Can you check if these test changes follow our guidelines on line width, and if so, could you please revert to the previous formatting? https://github.com/prestodb/presto/blob/master/CONTRIBUTING.md#code-style

180 characters is a general guideline, but when there's lots of arguments, one argument per line is more readable.

tdcmeehan (changes requested):
Just a nit on the formatting, core changes LGTM.

presto-hive/src/test/java/com/facebook/presto/hive/TestHiveSplitManager.java
elharo: I did an IntelliJ reformat on this fragment. It moved some lines a few spaces left, otherwise left them unchanged. It's useful to stick with what the autoformatter does to avoid a lot of churn. 

Ideally we should autoformat unconditionally without human intervention, but that's going to take a little time.

tdcmeehan: Definitely makes sense, but we do note to be careful about using the auto-formatting tool in the contributing guide, and we also lose history of the individual lines for these tests, so best to revert to the previous formatting.

rschlussel: often we'll put reformatting in a separate commit so it doesn't clutter up the real changes, but I disagree about leaving it behind entirely. if the new formatting is correct, I think it's good to commit it so people don't need to keep undoing autoreformats manually. I wouldn't worry about losing history (you can just go back farther in the git commit history if you need to).

elharo: anyway, undone for now to unblock this

rschlussel (approved):

presto-hive/src/test/java/com/facebook/presto/hive/TestHiveSplitManager.java
tdcmeehan: The trouble is whatever is auto-formatted by IntelliJ isn't necessarily guaranteed to be correct, which is the case here because we exceed line width guidelines linked above.  I also find the formatting confusing: `TestParquetQuickStatsBuilder` puts the remaining arguments after the `new Storage` argument on the same line, yet for some reason we put the arguments for `new Storage` all on individual lines.  I find this mix match of formatting confusing and think we should just fix it.

I also agree than in principle it would be great to format things consistently and automatically, and in such a way that it perfectly guaranteed alignment with our style guide.

elharo: There are several existing line length violations in this file, none of which seem to be on lines I touched. I don't see any line length violations in TestHIveSplitManager on lines I touched.

Yes, the formatting is confusing. That's caused by inlining way too many, too complex method arguments. Using more local variables would clarify this greatly. But this problem is not new with this PR, and shouldn't be blocking.

tdcmeehan: To be clear, I'm just referring to `TestParquetQuickStatsBuilder` and `TestQuickStatsProvider`, the rest I think can remain (although ideally these should be in a separate reformatting commit).

tdcmeehan (approved):

PR #22104, opened 2024-03-06
Re-enable TaskManagerTest.aggregationSpill

tanjialiang:
The test was disabled for preparing to deprecate spill pct. Now the spill pct is deprecated, we should bring this test back
```
== NO RELEASE NOTE ==
```


xiaoxmeng (approved):
@tanjialiang LGTM. Thanks!

Issue #22108, opened 2024-03-07
[Iceberg] Multiple entries with same value


wypb:
After https://github.com/prestodb/presto/pull/21605, Query may fail when executing on partitioned table:
```
SELECT * FROM partitioned_table
```
Stack Trace:
```
Caused by: java.lang.IllegalArgumentException: Multiple entries with same value: nationkey_19=1:nationkey:bigint and nationkey_17=1:nationkey:bigint
	at com.google.common.collect.ImmutableMap.conflictException(ImmutableMap.java:215)
	at com.google.common.collect.ImmutableMap.checkNoConflict(ImmutableMap.java:209)
	at com.google.common.collect.RegularImmutableBiMap.checkNoConflictInValueBucket(RegularImmutableBiMap.java:131)
	at com.google.common.collect.RegularImmutableBiMap.fromEntryArray(RegularImmutableBiMap.java:89)
	at com.google.common.collect.RegularImmutableBiMap.fromEntries(RegularImmutableBiMap.java:58)
	at com.google.common.collect.ImmutableBiMap.copyOf(ImmutableBiMap.java:368)
	at com.google.common.collect.ImmutableBiMap.copyOf(ImmutableBiMap.java:340)
	at com.facebook.presto.sql.planner.EffectivePredicateExtractor$Visitor.visitTableScan(EffectivePredicateExtractor.java:213)
	at com.facebook.presto.sql.planner.EffectivePredicateExtractor$Visitor.visitTableScan(EffectivePredicateExtractor.java:90)
	at com.facebook.presto.spi.plan.TableScanNode.accept(TableScanNode.java:203)
	at com.facebook.presto.sql.planner.EffectivePredicateExtractor.extract(EffectivePredicateExtractor.java:87)
	at com.facebook.presto.sql.planner.optimizations.PredicatePushDown$Rewriter.visitJoin(PredicatePushDown.java:450)
	at com.facebook.presto.sql.planner.optimizations.PredicatePushDown$Rewriter.visitJoin(PredicatePushDown.java:177)
	at com.facebook.presto.sql.planner.plan.JoinNode.accept(JoinNode.java:317)
	at com.facebook.presto.sql.planner.plan.InternalPlanNode.accept(InternalPlanNode.java:36)
	at com.facebook.presto.sql.planner.plan.SimplePlanRewriter$RewriteContext.rewrite(SimplePlanRewriter.java:92)
[...]

PR #22109, opened 2024-03-07
Fix multiple entries with same value when we enable delete as join

wypb:
## Description

Fixes https://github.com/prestodb/presto/issues/22108
## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

Iceberg Changes
* Fix multiple entries with same value when we enable delete as join
```

CC: @tdcmeehan @yingsu00

### DIFF
--- a/presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedTestBase.java
+++ b/presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedTestBase.java
@@ -978,6 +978,19 @@ public class IcebergDistributedTestBase
     }
 
     @Test(dataProvider = "equalityDeleteOptions")
+    public void testPartitionedTableWithEqualityDelete(String fileFormat, boolean joinRewriteEnabled)
+            throws Exception
+    {
+        Session session = deleteAsJoinEnabled(joinRewriteEnabled);
+        String tableName = "test_v2_equality_delete" + randomTableSuffix();
+        assertUpdate(session, "CREATE TABLE " + tableName + " WITH (partitioning = ARRAY['nationkey'], format = '" + fileFormat + "') " + " AS SELECT * FROM tpch.tiny.nation", 25);
+        Table icebergTable = updateTable(tableName);
+        writeEqualityDeleteToNationTable(icebergTable, ImmutableMap.of("regionkey", 1L, "nationkey", 1L), ImmutableMap.of("nationkey", 1L));
+        assertQuery(session, "SELECT * FROM " + tableName, "SELECT * FROM nation WHERE regionkey != 1 or nationkey != 1 ");
+        assertQuery(session, "SELECT nationkey, comment FROM " + tableName, "SELECT nationkey, comment FROM nation WHERE regionkey != 1 or nationkey != 1");
+    }
+
+    @Test(dataProvider = "equalityDeleteOptions")
     public void testEqualityDeletesWithPartitions(String fileFormat, boolean joinRewriteEnabled)
             throws Exception
     {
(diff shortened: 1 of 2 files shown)


presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedTestBase.java
```diff
@@ -956,6 +956,19 @@ public void testTableWithPositionDeleteAndEqualityDelete(String fileFormat, bool
...
 
+    @Test(dataProvider = "equalityDeleteOptions")
+    public void testPartitionedTableWithEqualityDelete(String fileFormat, boolean joinRewriteEnabled)
+            throws Exception
+    {
+        Session session = deleteAsJoinEnabled(joinRewriteEnabled);
+        String tableName = "test_v2_equality_delete" + randomTableSuffix();
+        assertUpdate(session, "CREATE TABLE " + tableName + " WITH (partitioning = ARRAY['nationkey'], format = '" + fileFormat + "') " + " AS SELECT * FROM tpch.tiny.nation", 25);
+        Table icebergTable = updateTable(tableName);
+        writeEqualityDeleteToNationTable(icebergTable, ImmutableMap.of("regionkey", 1L, "nationkey", 1L), ImmutableMap.of("nationkey", 1L));
```
jasonf20, line 967: There is a already a test that checks partitioned tables, it looks like this bug occurs when the equality_delete fields also contain the partition values, can you add a comment here mentioning that since it's important for the test

jasonf20 (commented):
@wypb Thanks for the fix! Overall LGTM. I don't think these delete files should contain the partition fields as part of the equality delete fields since the delete file is already associated with just 1 given partition. However, since this is something that could happen in the wild and might even make sense in some cases where partition evolution occurred it makes sense to support this.

hantangwangd:
Seems the spec indeed allows additional columns of the deleted row existing in equality delete files. If that's the case, then should we handle this problem when creating the `EQUALITY_DELETES` TableScanNode? As the same columnHandles in assignments might lead in problems in other places as well.

jasonf20:
The data file can contain more columns, but the only ones used here should be the ones mentioned listed in the equality delete field ids list.

wypb:
He @jasonf20 Thank you for your review. I also had another problem with duplicate keys in line 250 of IcebergEqualityDeleteAsJoin Class.  I fixed this internally, but I'm not sure how to build a test case yet.

jasonf20:
@wypb line 250 doesn't seem to map to anything related to map building on my end, are you on a modified version of the file? 
Also what is the use case that leads to this? Was it this same fix that resolved it for you or did you fix the table data?

wypb:
Oh, sorry, I did make changes to this file internally, but I used open source code to run internal SQL has the same problem. The corresponding open source code is the following:
```
Map<Integer, VariableReferenceExpression> reverseAssignmentsMap = updatedTableScan
                    .getAssignments()
                    .entrySet()
                    .stream()
                    .collect(Collectors.toMap(assignment -> ((IcebergColumnHandle) (assignment.getValue())).getId(), Map.Entry::getKey));
```

jasonf20:
Can you share the fix and what query / data led to the error?

wypb:
Yes, but I am still checking whether there is anything special about the table that caused the error.

hantangwangd:
> The data file can contain more columns, but the only ones used here should be the ones mentioned listed in the equality delete field ids list.

As I see, data files for normal tableScanNode would not meet this problem. Isn't this problem introduced by  `IcebergEqualityAsJoin.allFields()` which aims to build `deleteColumnAssignments` for a `EQUALITY_DELETES ` tableScanNode?

hantangwangd:
Seems we just did not handle the duplication from partition columns and equality columns  in `IcebergEqualityAsJoin.allFields()`.

tdcmeehan (approved):

tdcmeehan:
Failure was due to https://github.com/prestodb/presto/commit/7e5f88564ab35d0dec8521f45663ec507ebd7506, which is unrelated

wypb:
>> Can you share the fix and what query / data led to the error?

Hi @jasonf20 The fix for the Duplicate key problem can be found [here](https://github.com/wypb/presto/pull/1/files).

hantangwangd (approved):

jasonf20:
> > > Can you share the fix and what query / data led to the error?
> 
> Hi @jasonf20 The fix for the Duplicate key problem can be found [here](https://github.com/wypb/presto/pull/1/files).

Thanks, I'll have a look

PR #14506, opened 2020-05-09
Auto-resolve certain result mismatches

caithagoras:
The PR extends `FailureResolver` to not only be able to resolve query failures, but also result mismatches. Added 2 types of result mismatch resolvers.

## StructuredColumnMismatchResolver
If array element or map key/value contains floating point types,
column checksum is unlikely to match. Hence, we can allow Verifier
to auto resolve those cases.
* For an array column, resolve if the element type contains floating
  point types and the cardinality matches.
* For a map column, resolve if either key or value contains floating
  point types, the cardinality sum matches, and the checksum of the
  key or value that does not contains floating point type matches.
* Resolve a test case only when all columns are resolved.

## IgnoredFunctionsMismatchResolver
In the case of a results mismatch, if the query uses a function in a
specified list, the test case is marked as resolved.
This is useful to mark certain FB functions that is known to produce
different results on Verifier.

```
== RELEASE NOTES ==

Verifier Changes
* Add support to auto-resolve result mismatch for structured-type columns.
* Add support to auto-resolve result mismatch in case the query uses functions to be ignored.
```

PR #12489, opened 2019-03-18
Improve Presto Verifier

caithagoras:
Improve the existing verifier.

### Behavior differences
- Fix the issue where result-mismatched insert queries are not reran for non-determinism check.
- Fix the issue where checksum queries are treated as main queries while main queries are treated as part of the setup queries, leading to various mis-information on the UIs.
- Select is rewritten into `CreateTableAsSelect` and checksum queries are used for result comparison. This removes the row count limit for Select queries verification.
- Floating point columns are compared using relative errors.
- Orderable array columns are applied with `array_sort` before `checksum`.
- Verification happens along with rewrite, to avoid the super-long wait for the rewrite to complete before verification even starts.
- Query timeout are now enforced by session properties instead of by local timer.
- Add retries for query failures due to network errors and presto retryable errors.
- Add configuration properties `metadataTimeout` and `checksumTimeout`, the timeout duration for different Presto queries.
- Add configuration property `human-readable.log-file` to allow human readable events to be logged to file. Streamline human readable event logging.
- Add integration smoke test.
- Remove a bunch of unnecessary configuration properties.
- Make verifier easier to configure for community users.

### Changes to output event
  - Add `EventStatus status`, to indicate whether is verification is succeeded, failed, skipped or failed but auto-resolved.
  - Add `deterministic` to indicate whether result mismatch is due to non-determinism.
  - Add `originalControlQuery` and `originalTestQuery`, the control and test queries before the rewrite.
  - Add `controlChecksumQuery` and `testChecksumQuery`.
  - Add `errorCode`.
  - Add `mismatchedColumns`, a mapping from mismatched column names to their types.
  - Remove `source`.

PR #22143, opened 2024-03-08
[verifier] Add option to validate array(float) via error margin.

spershin:
## Description
Basically allow verifier to validate array(float) and array(double) columns in the same way as we validate float and double columns. Later we can extend this to the maps as well.

## Motivation and Context
We are planning to use this when comparing Presto Native with Presto Java to reduce the noise from array(double) columns. This will also strengthen the Presto verification as non-deterministic array(double) column could hide a valid correctness issue in another column by marking query SKIPPED in the verifier.

## Test Plan
Added new unit test and updated existing one.
Ran verifier on a query producing non-deterministic array(double) (artificial query).
Result is successful: the query passes correctness in the new verifier, while getting skipped in the old one.
Successfully run real use case verifier suite on the new verifier.

```
== NO RELEASE NOTE ==
```


spershin:
Submitted PR a bit too early - the validation part has not been implemented yet. But the PR is good to see where it is all going.

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnChecksum.java
```diff
@@ -23,12 +23,27 @@ public class ArrayColumnChecksum
...
+    private final long nanCount;
+    private final long positiveInfinityCount;
+    private final long negativeInfinityCount;
     private final Object cardinalityChecksum;
     private final long cardinalitySum;
 
-    public ArrayColumnChecksum(@Nullable Object checksum, @Nullable Object cardinalityChecksum, long cardinalitySum)
+    public ArrayColumnChecksum(
+            @Nullable Object checksum,
+            @Nullable Object sum,
```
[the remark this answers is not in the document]
spershin: @elharo 
I'm not sue, I was following the existing code from FloatingPointColumnValidator.
As I can see classes have long for counts and Object for everything else, why - not sure. :)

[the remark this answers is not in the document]
elharo: Let's watch out for Chesterton's Fence. Before proceeding we should figure out why checksum is an Object. Presumably there's a reason but maybe that reason applies to the new cardinality checksum and maybe it doesn't. It's also possible checksum shouldn't have been an Object in the first place, in which case we don't have to fix this now, but let's not repeat the mistake.  Any idea who could shed some light on this?

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnValidator.java
```diff
@@ -16,56 +16,86 @@
...
 import com.facebook.presto.common.type.RowType;
 import com.facebook.presto.common.type.Type;
-import com.facebook.presto.sql.tree.CoalesceExpression;
-import com.facebook.presto.sql.tree.Expression;
-import com.facebook.presto.sql.tree.FunctionCall;
-import com.facebook.presto.sql.tree.LongLiteral;
-import com.facebook.presto.sql.tree.QualifiedName;
-import com.facebook.presto.sql.tree.SingleColumn;
-import com.facebook.presto.sql.tree.TryExpression;
+import com.facebook.presto.sql.tree.*;
```
rschlussel, line 19: don't use * imports

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnChecksum.java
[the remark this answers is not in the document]
rschlussel: I can't see a good reason why sum is an Object in FloatingPointColumnChecksum.  the only places it's used, it's blindly cast to a double https://github.com/prestodb/presto/blob/master/presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/FloatingPointColumnValidator.java#L145-L146. For arrays, the sum is not necessarily a double, and I think it would be better not to assume it's a double here. (i mean you've only implemented it for double type, but that's more incidental.  You should be able to do sum validation on arrays of other numeric types)

Regarding why Objects are used for checksums throughout, I'm not sure. I assume it was most convenient, but I also suspect the code could be improved to use more specific types. In case it's helpful for gleaning more, the object usage seems to have been introduced here https://github.com/prestodb/presto/pull/12489/commits/04d45ea146cc858142e9fbd460bfe5e3ded82856 and here: https://github.com/prestodb/presto/pull/14506/commits/7c481036fb85f32fb7ff4c9fc0762f5df6389946 if that helps at all.

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnValidator.java
spershin: These are added automatically by the IJ, I think, will remove, but got to keep an eye on it.

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnChecksum.java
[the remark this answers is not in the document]
spershin: @elharo 

> Let's watch out for Chesterton's Fence. 

If we follow that rule you mentioned - we should not deviate of what already is in there. I suggest approach of 'as is'  and then follow up if we decide to change this. I would like this PR to stay focused on its task, if possible.

@rschlussel 

> Some additional thoughts - since we could imagine wanting to add special validation for other types.

Yes, we would likely go that way. Maps of doubles is the next candidate, especially for values part. I was thinking about it, however wanted to stay focused on just arrays now to unblock NGA use case quickly. We can adjust few things to fit Maps more as we go and introduce more unconventional support.

> Finally, I think we have the same problem you're solving here for map types, where it does checksums on the keys and values.

Not only that, but the structural types. like rows, but not sure if we want to really go there. Let's fo one step at a time? Harder to break existing stuff that way.

spershin:
The latest update includes:
1. Fixed failing tests.
2. Refactored validate() from FloatingPointColumnValidator class to FloatingPointColumnChecksum.
3. Fixed query generation for array(double) validator (infinity issue).
4. Added TestChecksumValidator.testFloatingPointArray() unit test.
5. Added support of array(double) and array(real) to TestChecksumValidator.testChecksumQuery() unit test.
6. Improved TestChecksumValidator.testChecksumQuery() to tell which line does not match.

spershin:
Fixed Checkstyle errors.

spershin:
More Checkstyle errors...

spershin:
Aaaand Checkstyle again...

spershin:
Fixed the checksum query.

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnChecksum.java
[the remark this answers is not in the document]
spershin: @elharo @rschlussel 
I believe objects are being used, because the aggregation can return NULL when there are no rows.
I'm currently debugging exactly such case.
To handle these nulls we use Object rather than a direct type.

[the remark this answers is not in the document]
rschlussel: Interesting. but then why not use a Double type (which can be null since it's an object) rather than a double primitive type?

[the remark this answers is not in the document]
spershin: Good question.
We didn't think about in the beginning?

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnValidator.java
```diff
@@ -16,82 +16,205 @@
...
+
+    private static Expression generateLambdaExpression(String functionName)
+    {
+        return new LambdaExpression(
+                ImmutableList.of(new LambdaArgumentDeclaration(identifier("x"))),
+                functionCall(functionName, new Identifier("x")));
+    }
+
+    @Override
+    public List<ColumnMatchResult<ArrayColumnChecksum>> validate(Column column, ChecksumResult controlResult, ChecksumResult testResult)
```
kewang1024, line 141: NIT: Can we put this function to its original place? (before function generateArrayChecksum). we should put public functions ahead then private function

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnValidator.java
```diff
@@ -16,82 +16,205 @@
...
+                testResult.getRowCount(),
+                controlResult.getRowCount());
+
+        boolean useFloatingPointPath = useFloatingPointPath(column);
+
+        ArrayColumnChecksum controlChecksum = toColumnChecksum(column, controlResult, useFloatingPointPath);
+        ArrayColumnChecksum testChecksum = toColumnChecksum(column, testResult, useFloatingPointPath);
+
+        // Not floating point case (we have '$checksum' column).
+        if (!useFloatingPointPath) {
```
kewang1024, line 155: We don't need `useFloatingPointPath ` here, we can infer this information from ArrayColumnChecksum structure

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ChecksumResult.java
```diff
@@ -75,6 +75,11 @@ public Map<String, Object> getChecksums()
         return checksums;
     }
 
+    public boolean hasChecksum(String columnName)
```
kewang1024, line 78: Looks like not used in this PR?

kewang1024 (commented):
Thanks @spershin for helping improve the verification of array(double)

presto-verifier/src/main/java/com/facebook/presto/verifier/framework/VerifierConfig.java
```diff
@@ -248,6 +249,19 @@ public VerifierConfig setAbsoluteErrorMargin(double absoluteErrorMargin)
...
+    }
+
+    @ConfigDescription("When set to true, arrays of floating point numbers are validated like floating point columns, using error margins. False by default.")
+    @Config("array-floating-point-error-margin")
+    public VerifierConfig setArrayFloatingPointErrorMargin(boolean arrayFloatingPointErrorMargin)
+    {
+        this.arrayFloatingPointErrorMargin = arrayFloatingPointErrorMargin;
+        return this;
+    }
+
```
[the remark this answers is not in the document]
spershin: Yes, for rollout, I plan to set it to 'false' by default and then enable and use for a while and then deprecate.
Now it is true, so I can test using the builds.

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnValidator.java
spershin: I placed them like this of better see the logic of the code:
public generateChecksumColumns() uses the following generateArrayChecksum(), generateInfinityLambdaExpression() and generateLambdaExpression().

public validate() uses the following toColumnChecksum().

I can bring them back, no issue.

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnValidator.java
spershin: How so?

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ChecksumResult.java
spershin: Ah, yes, I used 1st and then had to go different path because of the NULLs corner case. :(
Took me so much time with the slow testing (build pkg).

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/FloatingPointColumnChecksum.java
```diff
@@ -86,4 +101,63 @@ public String toString()
...
+        }
+
+        public ValidateResult(boolean matched, Optional<String> message)
+        {
+            this.matched = matched;
+            this.message = message;
+        }
+    }
+
+    public static ValidateResult validate(
```
kewang1024, line 124: Even changed to static function, I think it makes more sense to still put this function in [FloatingPointColumnValidator.java](https://github.com/prestodb/presto/pull/22143/files#diff-69d71b377216251d0931b7431965855d0a64922052e1167ad998c848f0b2cf5e)

FloatingPointColumnChecksum is only the data structure, FloatingPointColumnValidator will handle the validation logic

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnValidator.java
kewang1024: if `ArrayColumnChecksum `'s checksum is not null, then it's regular ArrayColumnChecksum
If it is null, then it's the float point check

spershin: > if `ArrayColumnChecksum `'s checksum is not null, then it's regular ArrayColumnChecksum If it is null, then it's the float point check

@kewang1024 

That was my 1st choice, but it all breaks on the corner case when we get checksums on empty tables (no rows) and then a bunch of checksum columns have nulls.

I realized after that if we are triggering floating point path checking the array element type - we should be consitent and always check using that approach. :)

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/FloatingPointColumnChecksum.java
spershin: @kewang1024 

Initially I kept it in the FloatingPointColumnValidator, but then I thought that this logic is better to be incapsulated in the structure that keeps the data. It already implements equals(), toString().
Make sense to put validate() here too, imho.

Let me know if you really prefer it back to the validator class.

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnValidator.java
spershin: Will be addressed in the next push.

spershin:
The last update:
1. Move ArrayColumnValidator.validate to be above private methods as per suggestion.
2. Remove unused ChecksumResult.hasChecksum()
3. Refactor ArrayColumnChecksum to contain FloatingPointColumnChecksum. Later we can use the same trick for Map keys and values.

spershin:
@kewang1024 , @rschlussel 
This change is close to the final change.
Code review is appreciated.
The previous change works successfully on the 10 NGA queries batch.
Will test on the real workload the final change (after comments are addressed) before merging.

kewang1024 (commented):
Thanks again! Overall looks good, some NIT

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnChecksum.java
```diff
@@ -25,12 +32,43 @@ public class ArrayColumnChecksum
     private final Object checksum;
     private final Object cardinalityChecksum;
     private final long cardinalitySum;
+    // For array(floating point) we have extra aggregations collected.
+    private final FloatingPointColumnChecksum floatingPointChecksum;
 
-    public ArrayColumnChecksum(@Nullable Object checksum, @Nullable Object cardinalityChecksum, long cardinalitySum)
+    // Constructor for array(non- floating point) types.
+    public ArrayColumnChecksum(
```
rschlussel, line 39: it would be better to have one constructor that does the assignment and the other constructors call that one and pass in whatever args are needed.

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnChecksum.java
```diff
@@ -63,19 +101,44 @@ public boolean equals(Object obj)
...
+        else {
+            return format(
+                    "%s, cardinality_checksum: %s, cardinality_sum: %s",
+                    floatingPointChecksum.toString(),
+                    cardinalityChecksum,
+                    cardinalitySum);
+        }
+    }
+
+    public static FloatingPointColumnChecksum.ValidateResult validateFloatingPoint(
```
rschlussel, line 130: it's weird to me that this logic is over here, but the rest of the validation logic is in ArrayColumnValidator.  I would do one or the other.

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnChecksum.java
```diff
@@ -63,19 +101,44 @@ public boolean equals(Object obj)
...
+            ArrayColumnChecksum controlChecksum,
+            ArrayColumnChecksum testChecksum,
+            double relativeErrorMargin,
+            double absoluteErrorMargin)
+    {
+        // Check the non floating point members first.
+        if (!Objects.equals(controlChecksum.getCardinalityChecksum(), testChecksum.getCardinalityChecksum()) ||
+                !Objects.equals(controlChecksum.getCardinalitySum(), testChecksum.getCardinalitySum())) {
+            return new FloatingPointColumnChecksum.ValidateResult(false, Optional.of("cardinality mismatch"));
```
rschlussel, line 139: Do we need this new class? Can we use ColumnMatchResult instead?

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnValidator.java
```diff
@@ -16,36 +16,87 @@
...
+import static com.facebook.presto.sql.QueryUtil.identifier;
 import static com.facebook.presto.verifier.framework.VerifierUtil.delimitedIdentifier;
 import static com.google.common.base.Preconditions.checkArgument;
 
 public class ArrayColumnValidator
         implements ColumnValidator
 {
+    private final double relativeErrorMargin;
+    private final double absoluteErrorMargin;
+    private final boolean arrayFloatingPointErrorMargin;
```
rschlussel, line 51: I find this name unclear.  Maybe something like "useErrorMarginForFloatArrays"?

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/FloatingPointColumnChecksum.java
rschlussel: I think either all validation in should be in the respective checksum classes or none of them(also, might fit better in the validate class if it were a non-static method like equals e.g. `floatingPointChecksum.validate(FloatingPointChecksum other)`). As things stand it seems a bit out of place, and I agree with @kewang1024 that it would be better in the validate class.

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnValidator.java
```diff
@@ -16,36 +16,87 @@
...
+
+    @Inject
+    public ArrayColumnValidator(VerifierConfig config)
+    {
+        this.relativeErrorMargin = config.getRelativeErrorMargin();
```
rschlussel, line 56: I wonder if we should inject FloatingPointColumnValidator instead, and then can just call floatingPointColumnValidator.validate() and leave these configs as implementation details of the floatingpointcolumnvalidator.

rschlussel (approved):

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/FloatingPointColumnChecksum.java
```diff
@@ -37,6 +42,16 @@ public FloatingPointColumnChecksum(@Nullable Object sum, long nanCount, long pos
         this.rowCount = rowCount;
     }
 
+    // Constructor to handle cases when the result table is empty (sum aggregations returns null).
```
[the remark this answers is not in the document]
spershin: Actually, I would prefer to leave it.
We use it, check the updated changes.

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnChecksum.java
spershin: Moved validate() logic to Validate classes.

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnChecksum.java
spershin: I did this for FloatingPointColumnChecksum, but for ArrayColumnChecksum I can only do this for empty ctor, as we need to initialize one member as optional in some cases.

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnChecksum.java
spershin: ColumnMatchResult is a templated class and we must return different instantiations of it from ArrayColumnChecksum and FloatingPointColumnChecksum.

I decide to have that small class to carry actual validation result, except all the columns info, which will be provided by concrete validators.

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnValidator.java
spershin: Replaced by useErrorMarginForFloatingPointArrays and "use-error-margin-for-floating-point-arrays".

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnValidator.java
spershin: @rschlussel 
Do you mean make FloatingPointColumnValidator a member of ArrayColumnValidator and pass the former to the constructor of the latter?
We can do this.

spershin:
@kewang1024 @rschlussel 

Thanks for the review.
I have addressed the comments so far.

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnChecksum.java
kewang1024: The `ValidateResult` still feel not needed, I don't see particular bad effect of using `ColumnMatchResult`

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnChecksum.java
kewang1024: I think providing two constructors here could be more straightforward

```
public ArrayColumnChecksum(@Nullable Object checksum, @Nullable Object cardinalityChecksum, long cardinalitySum)
{
this(checksum, cardinalityChecksum, cardinalitySum, Optional.empty());
}
```

```
public ArrayColumnChecksum(
@Nullable Object checksum,
@Nullable Object cardinalityChecksum,
long cardinalitySum,
Optional<FloatingPointColumnChecksum> floatingPointChecksum
)
{
// assignment
}
```

kewang1024 (approved):

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnChecksum.java
spershin: @kewang1024 
> I think providing two constructors here could be more straightforward

```
Hmm, not sure if looked at the new code or the old one?
We have 3 ctors now:

    // Constructor for array(non-floating point) types.
    public ArrayColumnChecksum(
            @Nullable Object checksum,
            @Nullable Object cardinalityChecksum,
            long cardinalitySum)

    // Constructor for array(floating point) types.
    public ArrayColumnChecksum(
            @Nullable Object sum,
            long nanCount,
            long positiveInfinityCount,
            long negativeInfinityCount,
            @Nullable Object cardinalityChecksum,
            long cardinalitySum)

    // Constructor to handle cases of array(floating point) when the result table is empty (sum aggregations returns null).
    public ArrayColumnChecksum()

```

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnChecksum.java
spershin: Ok, I guess, we can get rid of it.

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnChecksum.java
kewang1024: yeah, I saw that
combined with @rschlussel's suggestion, I think changing the constructors to the above two constructors can be more starightforward?

spershin:
Addressed ArrayColumnChecksum ctor comments.

presto-verifier/src/test/java/com/facebook/presto/verifier/resolver/TestStructuredColumnMismatchResolver.java
```diff
@@ -42,10 +44,10 @@ public TestStructuredColumnMismatchResolver()
     @Test
     public void testResolveArray()
```
rschlussel, line 45: Can you update the StructuredColumnsMismatchResolver and then add a test that we don't resolve column mismatches for float arrays if we're using error margin validation.

presto-verifier/src/test/java/com/facebook/presto/verifier/checksum/TestChecksumValidator.java
```diff
@@ -245,6 +275,75 @@ public void testFloatingPoint()
         assertEquals(mismatchedColumns.get(1).getMessage(), Optional.of("relative error: 1.0000500025000149E-4"));
     }
 
+    @Test
+    public void testFloatingPointArray()
```
rschlussel, line 279: can you add a test for the empty table case too?

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnChecksum.java
```diff
@@ -63,19 +78,29 @@ public boolean equals(Object obj)
...
     {
-        return Objects.hash(checksum, cardinalityChecksum, cardinalitySum);
+        return Objects.hash(checksum, floatingPointChecksum, cardinalityChecksum, cardinalitySum);
     }
 
     @Override
     public String toString()
     {
-        return format("checksum: %s, cardinality_checksum: %s, cardinality_sum: %s", checksum, cardinalityChecksum, cardinalitySum);
+        if (!floatingPointChecksum.isPresent()) {
```
rschlussel, line 95: nit: instead of repeating everything for constructing the string, you can use MoreObject.toStringHelper to construct a nicely printed string with all the fields. Something like:
```
MoreObjects.ToStringHelper toStringHelper = MoreObjects.toStringHelper(ArrayColumnchecksum)
    .add("cardinality_checksum", cardinalityChecksum)
    .add("cardinality_sum", cardinalitySum)

if(floatingPointChecksum.isPresent() {
    toStringHelper.add("floating_point_checksum", floatingPointchecksum.get());
} 
else {
    toStringHelper.add("checksum", checksum);
}

return toStringHelper.toString();
```

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnChecksum.java
```diff
@@ -63,19 +78,29 @@ public boolean equals(Object obj)
...
+            return format("checksum: %s, cardinality_checksum: %s, cardinality_sum: %s", checksum, cardinalityChecksum, cardinalitySum);
+        }
+        else {
+            return format(
+                    "%s, cardinality_checksum: %s, cardinality_sum: %s",
+                    floatingPointChecksum.get().toString(),
```
rschlussel, line 101: fyi, you don't need to call toString() here. The format function will do it if you pass in the object.

rschlussel (commented):
Thanks for the update. A few comments, but generally looks good!

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnChecksum.java
rschlussel: not sure if it's important that this toString matches the toStrings of the other checksums, so I'll add the caveat that the generated string will look slightly different than what you have, see the docs for example output https://guava.dev/releases/19.0/api/docs/com/google/common/base/MoreObjects.html#toStringHelper(java.lang.Object)

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnValidator.java
```diff
@@ -16,36 +16,85 @@
...
 public class ArrayColumnValidator
         implements ColumnValidator
 {
+    private final FloatingPointColumnValidator floatingPointValidator;
+    private final boolean useErrorMarginForFloatingPointArrays;
+
+    @Inject
+    public ArrayColumnValidator(VerifierConfig config)
+    {
+        this.floatingPointValidator = new FloatingPointColumnValidator(config);
```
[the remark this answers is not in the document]
spershin: @rschlussel 

I don't understand this. :)
Where do we take a floatingPointColumnValidator from to pass it into the ctor?
Are these Validator classes some kind of singletons and are being reused for multiple columns?
Thanks!

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnValidator.java
```diff
@@ -55,10 +104,31 @@ public List<SingleColumn> generateChecksumColumns(Column column)
     @Override
     public List<ColumnMatchResult<ArrayColumnChecksum>> validate(Column column, ChecksumResult controlResult, ChecksumResult testResult)
     {
-        ArrayColumnChecksum controlChecksum = toColumnChecksum(column, controlResult);
-        ArrayColumnChecksum testChecksum = toColumnChecksum(column, testResult);
+        checkArgument(
```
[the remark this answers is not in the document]
spershin: I'm not sure about it, honestly - repeating what FloatingPointColumnValidator is doing.

presto-verifier/src/test/java/com/facebook/presto/verifier/checksum/TestChecksumValidator.java
spershin: Yes, I was going to, but hit that snag with nulls in the Map.
I can see if I can use that HashMap to generate the nulls.

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnChecksum.java
spershin: @rschlussel 
I don't know. I don't feel that this change worth the trouble at the moment.
You are saying "to construct a nicely printed string with all the fields" - are they going to be more nicely printed than with the current code? How?

Maybe we can consider a small refactor followup to move all Checksums to such format?

rschlussel: not necessarily nicer than the current code. just helps you with not duplicating for the floating point vs. non-floating point case.  It would look something like `ArrayColumnChecksum{cardinality_checksum=123, cardinality_sum=345, ...}`. anyway, fine to leave it as is.

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnValidator.java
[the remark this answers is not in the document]
spershin: Yes, it looks like a sanity check to me.
I think we can remove it from ArrayColumnValidator and move it in FloatingPointColumnValidator from overridden validate() to the public validate(), which is called from the ArrayColumnValidator.

rschlussel (approved):

pranjalssh (approved):

PR #22151, opened 2024-03-11
[native] Do not add extra LocalExchange for native queries with join spill enabled

mbasmanova:
Unlike Java workers, native workers do not have any special requirements on the plan share when join spill is enabled. At the same, it is confusing that query plan changes based on whether join spilling is enabled or not.

For reference, here is a commit that added extra LocalExchange for Java workers: https://github.com/prestodb/presto/commit/ba98c10770a9b40687aae64c3612bdb2101d5603#diff-0d1dbdeb9958c20dabf2907e354679fdc10db52c4010f0e5d7fe38cbaa4872ee

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/AddLocalExchanges.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/AddLocalExchanges.java
@@ -72,6 +72,7 @@ import static com.facebook.presto.SystemSessionProperties.getTaskWriterCount;
 import static com.facebook.presto.SystemSessionProperties.isDistributedSortEnabled;
 import static com.facebook.presto.SystemSessionProperties.isEnforceFixedDistributionForOutputOperator;
 import static com.facebook.presto.SystemSessionProperties.isJoinSpillingEnabled;
+import static com.facebook.presto.SystemSessionProperties.isNativeExecutionEnabled;
 import static com.facebook.presto.SystemSessionProperties.isQuickDistinctLimitEnabled;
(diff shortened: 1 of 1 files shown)


ajaygeorge (approved):
LGTM

rschlussel (approved):

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/AddLocalExchanges.java
```diff
@@ -762,7 +763,7 @@ public PlanWithProperties visitUnion(UnionNode node, StreamPreferredProperties p
         public PlanWithProperties visitJoin(JoinNode node, StreamPreferredProperties parentPreferences)
         {
             PlanWithProperties probe;
-            if (isSpillEnabled(session) && isJoinSpillingEnabled(session)) {
+            if (isSpillEnabled(session) && isJoinSpillingEnabled(session) && !isNativeExecutionEnabled(session)) {
```
tdcmeehan, line 766: Could you add a comment explaining why this is the case?  It seems to make it more confusing, and the comment would be helpful to separate the first two conditions (this is for some reason needed whenever we spill joins, stemming from a limitation in the Java implementation) from the second (native engine is implemented in such a way to avoid this limitation).

mbasmanova: @tdcmeehan Just saw this comment. I'll send a follow-up PR to address this.

mbasmanova: Created #22154

PR #22156, opened 2024-03-11
Add Metastore cache

ajaygeorge:
- **Add Metastore cache interface**
- **Add Metastore In Memory cache implementation**

## Description
Add Metastore cache interface
Add Metastore In Memory cache implementation

## Motivation and Context
Provides a way to plug in different caching implementations. Default implementation of Loading Cache backed InMemory cache is provided.

## Test Plan
Unit Tests
Verifier testing.

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-hive-metastore/src/main/java/com/facebook/presto/hive/metastore/CachingHiveMetastore.java
+++ b/presto-hive-metastore/src/main/java/com/facebook/presto/hive/metastore/CachingHiveMetastore.java
@@ -18,61 +18,27 @@ import com.facebook.presto.common.type.Type;
 import com.facebook.presto.hive.ForCachingHiveMetastore;
 import com.facebook.presto.hive.HiveTableHandle;
 import com.facebook.presto.hive.HiveType;
-import com.facebook.presto.hive.MetastoreClientConfig;
-import com.facebook.presto.spi.PrestoException;
 import com.facebook.presto.spi.constraints.TableConstraint;
 import com.facebook.presto.spi.security.PrestoPrincipal;
 import com.facebook.presto.spi.security.RoleGrant;
 import com.facebook.presto.spi.statistics.ColumnStatisticType;
-import com.google.common.cache.CacheBuilder;
-import com.google.common.cache.CacheLoader;
-import com.google.common.cache.LoadingCache;
-import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Iterables;
-import com.google.common.collect.SetMultimap;
-import com.google.common.util.concurrent.UncheckedExecutionException;
 import io.airlift.units.Duration;
 import org.weakref.jmx.Managed;
 
 import javax.annotation.concurrent.ThreadSafe;
 import javax.inject.Inject;
 
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
-import java.util.Map.Entry;
-import java.util.Objects;
 import java.util.Optional;
 import java.util.OptionalLong;
 import java.util.Set;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.ThreadLocalRandom;
 import java.util.function.Function;
 
-import static com.facebook.presto.hive.HiveErrorCode.HIVE_CORRUPTED_PARTITION_CACHE;
-import static com.facebook.presto.hive.HiveErrorCode.HIVE_PARTITION_DROPPED_DURING_QUERY;
 import static com.facebook.presto.hive.metastore.CachingHiveMetastore.MetastoreCacheScope.ALL;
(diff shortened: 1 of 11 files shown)


jainxrohit:
I assume InMemoryMetastoreCache implementation is mostly same as CachingHiveMetastore?

ajaygeorge:
> I assume InMemoryMetastoreCache implementation is mostly same as CachingHiveMetastore?

@jainxrohit Mostly the same. Where it differs is how cache invalidation happens for calls such as drop/add Constraint etc.
The actual call is routed to the delegate and the invalidation happens on the cache implementation.

NikhilCollooru:
CachingHiveMetastore implements ExtendedHiveMetastore. ExtendedHiveMetastore and the new MetastoreCache interface are 90+% similar. A different caching implementation can implement ExtendedHiveMetastore and bind that instead of CachingHiveMetastore. For that reason i am not convinced to introduce a yet another interface.

ajaygeorge:
@NikhilCollooru You are correct. They appear to be similar. Where they differ is the objective of the `MetastoreCache` to introduce a clean separation of responsibilities. Imagine if we want to plugin a different in-memory cache instead of the LoadingCache it would be as simple as wiring in a new `MetastoreCache` implementation instead of making changes to the CachingHiveMetastore. 
Also at present if we want to turn off caching we set the TTL as 0. Using this approach we can just plugin the NoopMetastoreCache which is cleaner and less hacky. Interestingly TTL of 0 in Memcache means quite the opposite i.e. cache entries will never expire.

jainxrohit:
@ajaygeorge I do feel there is a need some optimization in the interface design. Right now it feels CachingHiveMetastore and MetastoreCache are doing similar job. An example to follow would be CachingFileSystem imo. We can setup sometime to unblock on this quickly.

rschlussel:
I'm a bit conflicted because I agree that there are already so many layers of indirection with the hive metastore, but I think I like the approach here.  I think it makes sense to isolate the specific caching implementation from the rest of the logic (my only comment in that regard for the interface is that the names of the methods should all be getXXX vs. e.g. listRoles because it doesn't matter what metastore call puts them there, the cache is just for storing and retrieving information)

rschlussel:
>  Also at present if we want to turn off caching we set the TTL as 0. Using this PR approach we can just plugin the NoopMetastoreCache which is cleaner and less hacky. Interestingly TTL of 0 in Memcache means quite the opposite i.e. cache entries will never expire.

I do think we should separate these two issues, as we can fix the code not to use a CachingHiveMetastore if caching is disabled without also introducing this new interface.

presto-hive-metastore/src/main/java/com/facebook/presto/hive/metastore/CachingHiveMetastore.java
```diff
@@ -87,284 +53,51 @@ public enum MetastoreCacheScope
...
-                metastoreImpersonationEnabled,
-                OptionalLong.of(cacheTtl.toMillis()),
-                refreshInterval.toMillis() >= cacheTtl.toMillis() ? OptionalLong.empty() : OptionalLong.of(refreshInterval.toMillis()),
-                maximumSize,
-                partitionVersioningEnabled,
-                metastoreCacheScope,
-                partitionCacheValidationPercentage,
-                partitionCacheColumnCountLimit,
-                metastoreCacheStats);
+        this.delegate = requireNonNull(delegate, "delegate is null");
```
NikhilCollooru, line 63: do we still need this delegate ? given its abstracted by the MetastoreCache ?

presto-hive-metastore/src/main/java/com/facebook/presto/hive/metastore/NoopMetastoreCache.java
```diff
@@ -0,0 +1,230 @@
...
+import java.util.Optional;
+import java.util.Set;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Noop Metastore cache which does not cache any data
+ * All the calls are routed to the underlying delegate
+ */
+public class NoopMetastoreCache
```
jainxrohit, line 36: Not a fan of name Noop, its doing something, its calling the delegate.

rschlussel: Even better might be to avoid using the CachingHiveMetastore entirely when we aren't using the caching. Just use the delegate directly.

ajaygeorge: removed this since we can turn off by avoiding the CachingPath if needed.

jainxrohit:
@ajaygeorge  since you are working on a different PR now, can we close this? If the idea is to build in that?

ajaygeorge:
> @ajaygeorge since you are working on a different PR now, can we close this? If the idea is to build in that?

@jainxrohit We can close this since #22256 supersedes this PR.

PR #22158, opened 2024-03-11
[native] Deprecate native_execution_enabled session property

### LINKED ISSUE #20008: [native pos] Remove workaround of isNative flag passed to SqlToRowExpressionTranslator
isNative flag passed to SqlToRowExpressionTranslator from SqlFunctionUtils is hard coded to false https://github.com/vermapratyush/presto/blob/d5eed561c31e21cb0b6cfa5f94038d9d4f2b4b89/presto-main/src/main/java/com/facebook/presto/sql/relational/SqlFunctionUtils.java#L111
https://github.com/prestodb/presto/blob/d5eed561c31e21cb0b6cfa5f94038d9d4f2b4b89/presto-main/src/main/java/com/facebook/presto/sql/relational/SqlFunctionUtils.java#L111

Ideally, this should be derived from session value, hard coding this to false does not break any flow since we always disable inline sql function for native: 
https://github.com/prestodb/presto/blob/master/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/NativeQueryRunnerUtils.java#L42

### LINKED ISSUE #22153: Deprecate `native_execution_enabled`
Native execution is an invariant for the cluster that will never change--it should just be a system property.

We should not have session properties that represent invariants as it increases the footprint of session properties visible to the user and makes Presto harder to operate.

Engine

## Possible Implementation
Just replace all references of `SystemSessionProperties#isNativeExecutionEnabled` with a new system property that accomplishes this.

## Context
This came out of a discussion in #22047 which discussed how to expose this session property to connectors.  It was revealed that it might have been more optimal to originally expose it as a config property.

mbasmanova:
Fixes #22153
Fixes #20008

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
+++ b/presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
@@ -340,7 +340,7 @@ public final class SystemSessionProperties
     public static final String NATIVE_ROW_NUMBER_SPILL_ENABLED = "native_row_number_spill_enabled";
     public static final String NATIVE_TOPN_ROW_NUMBER_SPILL_ENABLED = "native_topn_row_number_spill_enabled";
     public static final String NATIVE_JOIN_SPILLER_PARTITION_BITS = "native_join_spiller_partition_bits";
-    public static final String NATIVE_EXECUTION_ENABLED = "native_execution_enabled";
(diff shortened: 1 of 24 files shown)


tdcmeehan (approved):
Looks great, thank you!

PR #22169, opened 2024-03-12
[Verifier] Support reuse of source query output table

gggrace14:
Add support to Presto Verifier to reuse the output tables of the source query for control and test. Add configs --control-reuse-table and --test-reuse-table to turn the feature on and off. If reuse-table is turned on, skip the execution of main query as well as the setup and teardown queries. Compute checksums on the output table of main query directly.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-docs/src/main/sphinx/admin/verifier.rst
+++ b/presto-docs/src/main/sphinx/admin/verifier.rst
@@ -276,6 +276,10 @@ Name                                        Description
                                             are emitted to ``stdout``.
 ``control.table-prefix``                    The table prefix to be appended to the control target table.
 ``test.table-prefix``                       The table prefix to be appended to the test target table.
+``control.reuse-table``                     If ``true``, reuse the output table of the control source Insert and CreateTableAsSelect
+                                            query. Otherwise, run the control source query and write to a temporary table.
+``test.reuse-table``                        If ``true``, reuse the output table of the test source Insert and CreateTableAsSelect
+                                            query. Otherwise, run the test source query and write to a temporary table.
 ``test-id``                                 A string to be attached to output events.
 ``max-concurrency``                         Maximum number of concurrent verifications.
 ``suite-repetition``                        How many times a suite is verified.
--- a/presto-verifier/src/main/java/com/facebook/presto/verifier/event/QueryInfo.java
+++ b/presto-verifier/src/main/java/com/facebook/presto/verifier/event/QueryInfo.java
@@ -53,12 +53,11 @@ public class QueryInfo
     private final String bucketChecksumQuery;
     private final String jsonPlan;
     private final String outputTableName;
-
+    private final boolean isReuseTable;
     private final Double cpuTimeSecs;
     private final Double wallTimeSecs;
     private final Long peakTotalMemoryBytes;
     private final Long peakTaskTotalMemoryBytes;
-
     private final String extraStats;
 
     public QueryInfo(
@@ -79,7 +78,8 @@ public class QueryInfo
             Optional<String> bucketChecksumQuery,
             Optional<String> jsonPlan,
             Optional<QueryActionStats> queryActionStats,
-            Optional<String> outputTableName)
+            Optional<String> outputTableName,
+            boolean isReuseTable)
     {
         Optional<QueryStats> stats = queryActionStats.flatMap(QueryActionStats::getQueryStats);
         this.catalog = requireNonNull(catalog, "catalog is null");
@@ -105,6 +105,7 @@ public class QueryInfo
         this.peakTaskTotalMemoryBytes = stats.map(QueryStats::getPeakTaskTotalMemoryBytes).orElse(null);
         this.extraStats = queryActionStats.flatMap(QueryActionStats::getExtraStats).orElse(null);
         this.outputTableName = outputTableName.orElse(null);
+        this.isReuseTable = isReuseTable;
     }
 
     private static double millisToSeconds(long millis)
@@ -221,6 +222,12 @@ public class QueryInfo
     }
 
     @EventField
+    public boolean getIsReuseTable()
+    {
+        return isReuseTable;
+    }
+
+    @EventField
     public Double getCpuTimeSecs()
     {
         return cpuTimeSecs;
@@ -279,6 +286,7 @@ public class QueryInfo
         private Optional<String> jsonPlan = Optional.empty();
         private Optional<QueryActionStats> queryActionStats = Optional.empty();
         private Optional<String> outputTableName = Optional.empty();
+        private boolean isReuseTable;
 
         private Builder(
                 String catalog,
@@ -376,6 +384,12 @@ public class QueryInfo
             return this;
         }
 
+        public Builder setIsReuseTable(boolean isReuseTable)
+        {
+            this.isReuseTable = isReuseTable;
+            return this;
+        }
+
         public QueryInfo build()
         {
             return new QueryInfo(
@@ -396,7 +410,8 @@ public class QueryInfo
                     bucketChecksumQuery,
                     jsonPlan,
                     queryActionStats,
-                    outputTableName);
+                    outputTableName,
+                    isReuseTable);
         }
     }
 }
--- a/presto-verifier/src/main/java/com/facebook/presto/verifier/framework/AbstractVerification.java
+++ b/presto-verifier/src/main/java/com/facebook/presto/verifier/framework/AbstractVerification.java
@@ -196,6 +196,8 @@ public abstract class AbstractVerification<B extends QueryBundle, R extends Matc
         ChecksumQueryContext testChecksumQueryContext = new ChecksumQueryContext();
         Optional<R> matchResult = Optional.empty();
         Optional<DeterminismAnalysisDetails> determinismAnalysisDetails = Optional.empty();
+        boolean controlReuseTable = false;
+        boolean testReuseTable = false;
 
         Optional<PartialVerificationResult> partialResult = Optional.empty();
         Optional<Throwable> throwable = Optional.empty();
@@ -204,11 +206,22 @@ public abstract class AbstractVerification<B extends QueryBundle, R extends Matc
             // Rewrite queries
             if (isControlEnabled()) {
                 control = Optional.of(getQueryRewrite(CONTROL));
+                controlReuseTable = control.isPresent() && control.get() instanceof QueryObjectBundle && ((QueryObjectBundle) control.get()).isReuseTable();
+                if (controlReuseTable) {
+                    controlQueryContext.setState(QueryState.REUSE);
+                    controlQueryContext.setMainQueryStats(QueryActionStats.queryIdStats(sourceQuery.getQueryId(CONTROL).get()));
+                }
             }
+
             test = Optional.of(getQueryRewrite(TEST));
+            testReuseTable = test.isPresent() && test.get() instanceof QueryObjectBundle && ((QueryObjectBundle) test.get()).isReuseTable();
+            if (testReuseTable) {
+                testQueryContext.setState(QueryState.REUSE);
+                testQueryContext.setMainQueryStats(QueryActionStats.queryIdStats(sourceQuery.getQueryId(TEST).get()));
+            }
 
             // First run setup queries
-            if (isControlEnabled()) {
+            if (isControlEnabled() && !controlReuseTable) {
                 QueryBundle controlQueryBundle = control.get();
                 QueryAction controlSetupAction = setupOnMainClusters ? queryActions.getControlAction() : queryActions.getHelperAction();
                 controlQueryBundle.getSetupQueries().forEach(query -> runAndConsume(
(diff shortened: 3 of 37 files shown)


mbasmanova:
@gggrace14 Ge, how much faster verifier runs with this optimization?

spershin:
> @gggrace14 Ge, how much faster verifier runs with this optimization?

@mbasmanova 

This needs to be thoroughly tested to get some numbers and also depends on workflow and number of queries.
We can run this for a while in Native and compare the difference.

But, obviously it is the time without optimization minus time it takes to run setup, main and teardown queries.
Note, that it looks like (I noticed) the verifier runs control query 1st and then runs the test one, so both main queries are SERIALIZED.
Not sure if there is a verifier option to run them in parallel.

gggrace14:
class QueryRewriter, class AbstractVerification and class VerifierDao have the main changes. Others are about passing the required info around.

presto-verifier/src/main/java/com/facebook/presto/verifier/source/VerifierDao.java
```diff
@@ -34,32 +35,40 @@ public interface VerifierDao
             "  control_catalog varchar(256) NOT NULL,\n" +
             "  control_schema varchar(256) NOT NULL,\n" +
             "  control_query text NOT NULL,\n" +
+            "  control_query_id mediumtext DEFAULT NULL,\n" +
```
spershin, line 38: Strange, didn't we put control and test query Ids into the table before this change?
The ones that you click on, like "Control Query ID".

gggrace14:
> Not sure if there is a verifier option to run them in parallel.

For control & test checksum queries, current implementation is to run them in sequence. For main queries, param --concurrent-control-and-test can run them in parallel.

presto-verifier/pom.xml
```diff
@@ -45,6 +45,11 @@
             <artifactId>presto-hive-common</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>com.facebook.presto</groupId>
+            <artifactId>presto-client</artifactId>
+        </dependency>
+
```
spershin, line 52: What do we need from the presto-client in the verifier?

spershin (approved):
@gggrace14
Thanks for this change (and others) to enabled re-use of the tables and reduce verification time and compute.

The change looks good to me, few questions.

Maybe worth @kewang1024, @ajaygeorge  or @aweisberg to take a quick look at it too.

presto-verifier/src/main/java/com/facebook/presto/verifier/source/VerifierDao.java
gggrace14: Yeah, the "Control Query ID" link on UI is populated with the query_id returned by the actual control query run triggered by Verifier today. If we reuse the shadow/prod query output and don't trigger the control/test query runs, we have to pass the query_ids to Verifer through xdb, so that the UI can display them.

Otherwise our debugging experience is bad. Found it out from testing.

presto-verifier/pom.xml
gggrace14: One place is the following lines of change. The change is about passing the query_id from xdb to the UI by setting them in the related data structures.

In presto-verifier
```
import com.facebook.presto.client.StatementStats;
...
public static QueryActionStats queryIdStats(String queryId)
    {
        return new QueryActionStats(Optional.of(QueryStats.create(queryId, StatementStats.builder().setState(QueryState.FINISHED.toString()).build())), Optional.empty());
    }
```

presto-verifier/src/main/java/com/facebook/presto/verifier/framework/AbstractVerification.java
```diff
@@ -202,52 +205,74 @@ public VerificationResult run()
             // Rewrite queries
             if (isControlEnabled()) {
                 control = Optional.of(getQueryRewrite(CONTROL));
+                controlReuseTable = control.isPresent() && control.get() instanceof QueryObjectBundle && ((QueryObjectBundle) control.get()).isReuseTable();
+                if (controlReuseTable) {
+                    controlQueryContext.setState(QueryState.REUSE);
+                    controlQueryContext.setMainQueryStats(QueryActionStats.queryIdStats(sourceQuery.getQueryId(CONTROL).get()));
+                }
```
kewang1024, line 212: It would be clearer to put the logic inside the getQueryRewrite()

presto-verifier/src/main/java/com/facebook/presto/verifier/framework/AbstractVerification.java
```diff
@@ -202,52 +205,74 @@ public VerificationResult run()
...
             }
+
             test = Optional.of(getQueryRewrite(TEST));
+            testReuseTable = test.isPresent() && test.get() instanceof QueryObjectBundle && ((QueryObjectBundle) test.get()).isReuseTable();
+            if (testReuseTable) {
+                testQueryContext.setState(QueryState.REUSE);
+                testQueryContext.setMainQueryStats(QueryActionStats.queryIdStats(sourceQuery.getQueryId(TEST).get()));
+            }
```
kewang1024, line 220: same as above

presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/QueryRewriteConfig.java
```diff
@@ -31,6 +31,21 @@ public class QueryRewriteConfig
...
     private Map<String, Object> tableProperties = ImmutableMap.of();
 
+    private boolean reuseTable;
+
+    public boolean isReuseTable()
+    {
+        return reuseTable;
+    }
+
+    @ConfigDescription("If true, reuse the output table of the source query. Otherwise, run the query and write to a temporary shadow table.")
```
kewang1024, line 41: 1.Is it the best effort?
if it is set to true, but the output table is deleted, does the verification fail or follow original routine?

2.Does it only apply to INSERT and CTAS queries? if so, we should specify in the config description

steveburnett:
Suggest adding 

* documentation for the new configuration properties. 

  A good location might be near the `control.table-prefix` and `test.table-prefix` entries (lines 275-276) in https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/admin/verifier.rst?plain=1. 

* a release note entry. Perhaps something like the following:

```
== RELEASE NOTES ==

Verifier Changes
* Add  `control-reuse-table` and `test-reuse-table` configuration properties for the Presto Verifier to reuse the output tables of the source query for control and test :pr:`22169`
```

presto-verifier/src/test/java/com/facebook/presto/verifier/rewrite/TestQueryRewriter.java
```diff
@@ -468,19 +475,53 @@ public void testRewriteFunctionCalls()
...
+    @Test
+    public void testReuseTableRewrite()
+    {
+        String query = "INSERT INTO dest_table SELECT * FROM test_table";
+        QueryConfiguration configuration = new QueryConfiguration(
+                CATALOG,
+                SCHEMA,
+                Optional.of("user"),
+                Optional.empty(),
+                Optional.empty(), true);
```
arhimondr, line 487: nit: new line

presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/QueryRewriter.java
```diff
@@ -150,11 +171,21 @@ public QueryObjectBundle rewriteQuery(@Language("SQL") String query, ClusterType
...
-                    clusterType);
+                    clusterType,
+                    false);
         }
         if (statement instanceof Insert) {
             Insert insert = (Insert) statement;
             QualifiedName originalTableName = insert.getTarget();
+            if (shouldReuseTable) {
+                return new QueryObjectBundle(
+                        originalTableName,
```
arhimondr, line 182: What if the original table contains some other data then the data inserted by the query being verifier. How do we know what partitions of the original table to compute the checksum over?

presto-verifier/src/main/java/com/facebook/presto/verifier/framework/AbstractVerification.java
```diff
@@ -280,14 +310,17 @@ else if ((isControlEnabled()) && !skipChecksum) {
...
+                }
                 return new VerificationResult(this, true, Optional.empty());
             }
             throwable = Optional.of(t);
             partialResult = Optional.of(concludeVerificationPartial(control, test, controlQueryContext, testQueryContext, matchResult, determinismAnalysisDetails, Optional.of(t)));
         }
         finally {
             if (!smartTeardown
-                    || testQueryContext.getState() != QueryState.SUCCEEDED
+                    || testQueryContext.getState() != QueryState.SUCCEEDED && testQueryContext.getState() != QueryState.REUSE
```
arhimondr, line 323: nit: maybe wrap `testQueryContext.getState() != QueryState.SUCCEEDED && testQueryContext.getState() != QueryState.REUSE` into `( .. )`

presto-verifier/src/main/java/com/facebook/presto/verifier/framework/AbstractVerification.java
```diff
@@ -79,7 +79,7 @@ public abstract class AbstractVerification<B extends QueryBundle, R extends Matc
     private static final String SNAPSHOT_DOES_NOT_EXIST = "SNAPSHOT_DOES_NOT_EXIST";
 
     private final QueryActions queryActions;
-    private final SourceQuery sourceQuery;
+    private SourceQuery sourceQuery;
```
arhimondr, line 82: nit: consider keeping it final and return a new instance in `VerificationResult` (instead of `this`)

presto-verifier/src/main/java/com/facebook/presto/verifier/event/QueryInfo.java
```diff
@@ -53,12 +53,11 @@ public class QueryInfo
     private final String bucketChecksumQuery;
     private final String jsonPlan;
     private final String outputTableName;
-
+    private final Boolean isReuseTable;
```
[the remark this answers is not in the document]
gggrace14: Ah, typo

presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/QueryRewriter.java
gggrace14: A good question! I have a next PR on top of this one that assembles a partition predicate and applies it to the checksum query.

presto-verifier/src/main/java/com/facebook/presto/verifier/framework/AbstractVerification.java
gggrace14: Revised

presto-verifier/src/main/java/com/facebook/presto/verifier/framework/VerificationManager.java
```diff
@@ -352,26 +355,25 @@ private void reportProgressUntilFinished()
                 VerificationResult result = completionService.take().get();
                 Optional<VerifierQueryEvent> event = result.getEvent();
                 completed++;
-                if (!event.isPresent()) {
-                    statusCount.compute(SKIPPED, (status, count) -> count == null ? 1 : count + 1);
```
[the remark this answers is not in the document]
gggrace14: !event.isPresent() is when the test case is to be resubmitted. From our experience, we find it confusing if we also add the resubmission count to the counter SKIPPED. This problem becomes more apparent when reuse-table is enabled, as we resubmit more to be safe.

The main condition of SKIPPED (line 361) is when the control query run fails, ex., due to syntax error, which is how ppl usually assume the counter SKIPPED. When SKIPPED includes resubmissions as today, we find it confusing that the counters SUCCEED, FAILED and SKIPPED don't add up to the total test cases, as one test case can be resubmitted multiple times. So I split it in this change and add one more counter RESUBMITTED (line 366).

presto-verifier/src/main/java/com/facebook/presto/verifier/source/PrestoQuerySourceQuerySupplier.java
```diff
@@ -41,20 +41,26 @@ public class PrestoQuerySourceQuerySupplier
                     resultSet.getString("name"),
                     resultSet.getString("control_query"),
                     resultSet.getString("test_query"),
+                    Optional.ofNullable(resultSet.getString("name")), // Assume column `name` has the source `control_query_id`.
```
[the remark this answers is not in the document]
gggrace14: Revised here.

presto-verifier/src/main/java/com/facebook/presto/verifier/framework/AbstractVerification.java
gggrace14: A good catch. Also thought it's better if I can make this class immutable. However, I can't figure out how the constructor can look like, esp. since this class AbstractVerification is an abstract class. Please let me know if you have a good suggestion.

I thought I could add an abstract method to this class
```
public abstract AbstractVerification withSourceQuery(SourceQuery sourceQuery);
```
Then its subclass DataVerification overrides this method, creates and returns a DataVerification instance. However, we can't access those private fields, ex. queryActions, of the base abstract class in the subclass when calling the DataVerification constructor.

presto-jdbc/pom.xml
```diff
@@ -229,6 +229,7 @@
                             <goal>shade</goal>
                         </goals>
                         <configuration>
+                            <shadedArtifactAttached>true</shadedArtifactAttached>
```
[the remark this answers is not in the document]
gggrace14: Figured out a way to avoid this attribute and changing the packaging.

gggrace14:
> Suggest adding
> 
> * documentation for the new configuration properties.
>   A good location might be near the `control.table-prefix` and `test.table-prefix` entries (lines 275-276) in https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/admin/verifier.rst?plain=1.
> * a release note entry. Perhaps something like the following:
> 
> ```
> == RELEASE NOTES ==
> 
> Verifier Changes
> * Add  `control-reuse-table` and `test-reuse-table` configuration properties for the Presto Verifier to reuse the output tables of the source query for control and test :pr:`22169`
> ```

Adding the corresponding documentation to the verifier.rst.

For the release note, I have a next PR <link> on top of this for the complete feature. I'd prefer to add release note when the complete feature is in, to expose it to the user.

presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/QueryRewriteConfig.java
gggrace14: 1. If the output table is deleted, the verification will fail.
2. Revised.

presto-docs/src/main/sphinx/admin/verifier.rst
```diff
@@ -276,6 +276,10 @@ Name                                        Description
                                             are emitted to ``stdout``.
 ``control.table-prefix``                    The table prefix to be appended to the control target table.
 ``test.table-prefix``                       The table prefix to be appended to the test target table.
+``control.reuse-table``                     If true, reuse the output table of the control source Insert and CreateTableAsSelect
```
steveburnett, line 279:
```suggestion
``control.reuse-table``                     If ``true``, reuse the output table of the control source Insert and CreateTableAsSelect
```
Nit for formatting, same as #22965.

presto-docs/src/main/sphinx/admin/verifier.rst
```diff
@@ -276,6 +276,10 @@ Name                                        Description
...
+                                            query. Otherwise, run the control source query and write to a temporary table.
+``test.reuse-table``                        If true, reuse the output table of the test source Insert and CreateTableAsSelect
```
steveburnett, line 281:
```suggestion
``test.reuse-table``                        If ``true``, reuse the output table of the test source Insert and CreateTableAsSelect
```

steveburnett (changes requested):
Thanks for the doc! Nit of formatting, no other concerns.

presto-docs/src/main/sphinx/admin/verifier.rst
gggrace14: Revised

presto-docs/src/main/sphinx/admin/verifier.rst
gggrace14: Revised

presto-verifier/src/main/java/com/facebook/presto/verifier/framework/AbstractVerification.java
gggrace14: Hi @arhimondr , I figured out a different way that doesn't need to make this sourceQuery field mutable. The original objective is to pass a signal so that the "resubmitted" verification workflow will not reuse output tables. Previously I constructed a new SourceQuery instance. A different way I figured is to check the `VerificationContext.getResubmissionCount()` as below. I figured I'm able to access VerificationContext when calling `queryRewriter.rewriteQuery()`
```
    @Override
    protected QueryObjectBundle getQueryRewrite(ClusterType clusterType)
    {
        return queryRewriter.rewriteQuery(getSourceQuery().getQuery(clusterType), getSourceQuery().getQueryConfiguration(clusterType), clusterType,
                getVerificationContext().getResubmissionCount() == 0);
    }
```

arhimondr (approved):

gggrace14:
> Thanks for the doc! Nit of formatting, no other concerns.

Hi @steveburnett , revised the format of verifier.rst document. Would you be able to help review again?

gggrace14:
> > Thanks for the doc! Nit of formatting, no other concerns.
> 
> Hi @steveburnett , revised the format of verifier.rst document. Would you be able to help review again?

Hi @steveburnett , thanks for the review! It looks to need your approval to be able to merge.

steveburnett:
Not a required test so not blocking merge, but the test 

`singlestore tests / singlestore-dockerized-tests (pull_request)` 

failed here. I believe I've seen other mentions of that test being flaky lately, wanted to mention it in case people are looking for examples of it failing.

gggrace14:
> LGTM! (docs)
> 
> Pull updated branch, new local build, looks good. Thanks!

Thank you, @steveburnett !

PR #22177, opened 2024-03-13
[native] Disable dynamic filter pushdown

mbasmanova:
Velox implements dynamic filter pushdown from join into collocated table scan adaptively without guidance from the planner.

```
== NO RELEASE NOTE ==
```


presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/PredicatePushDown.java
```diff
@@ -1524,6 +1529,11 @@ private PlanNode visitNonFilteringSemiJoin(SemiJoinNode node, RewriteContext<Row
             return output;
         }
 
+        private boolean isEnableDynamicFiltering()
+        {
+            return !nativeExecution && SystemSessionProperties.isEnableDynamicFiltering(session);
+        }
```
spershin, line 1535: @mbasmanova 
I wonder if that would be better to have that check in the SystemSessionProperties itself?

mbasmanova: That's a good question. SystemSessionProperties.isEnableDynamicFiltering is a static method, so doesn't have access to FeaturesConfig.

@tdcmeehan Tim, what do you think?

tdcmeehan: Practically, you could inject an instance of `SystemSessionProperties` and create a non-static method which has access to the feature config.  But IMO it's just simpler and more straightforward to do it as it's written here.

mbasmanova:
@tdcmeehan Tim, do you think you can accept this PR?

spershin (approved):
Thanks!

I believe it would be better to take featuresConfig.isNativeExecutionEnabled() into account in the SystemSessionProperties.isEnableDynamicFiltering(session).
If possible.
Thus it is guaranteed to be one spot and we don't need to change any code, except config's.

I'm pretty sure there would be other props that would need such behavior and having it consistent can be a bonus.

mbasmanova:
> I'm pretty sure there would be other props that would need such behavior 

Indeed. One other such property would be optimize_hash_generation. I didn't plan to disable it though.

> I believe it would be better to take featuresConfig.isNativeExecutionEnabled() into account in the SystemSessionProperties.isEnableDynamicFiltering(session).

I'm just not sure how to do that. SystemSessionProperties.isEnableDynamicFiltering is a static method and making it have access to FeaturesConfig would require making it non-static which will be inconsistent with how all other session properties are accessed.

xiaoxmeng (approved):
@mbasmanova thanks!

arhimondr (approved):

PR #22185, opened 2024-03-13
[native] Only detach worker when more than 1/2 of driver threads blocked.

spershin:
## Description
We were detaching workers when only one stuck operator detected. This makes it possible for one bad query to disable the cluster. This change makes the detachment happen only in more dramatic situation than just one bad query.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
+++ b/presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
@@ -121,6 +121,7 @@ void PeriodicTaskManager::start() {
   }
 
   if (server_ && server_->hasCoordinatorDiscoverer()) {
+    numDriverThreads_ = server_->numDriverThreads();
     addWatchdogTask();
   }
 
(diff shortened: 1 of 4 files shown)


presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
```diff
@@ -83,6 +83,10 @@ PeriodicTaskManager::PeriodicTaskManager(
       server_(server) {}
 
 void PeriodicTaskManager::start() {
+  numDriverThreads_ = size_t(
```
Yuhta, line 86: Could be an accessor on `PrestoServer` so we don't repeat the logic here

spershin: @Yuhta 
The code allows `server_` (`PrestoServer*`) to be null, so can't really use this for bullet-proof code.
I also thought just taking it from `driverCPUExecutor_` itself, but that could be null too.
So, doing this. :)

spershin: @Yuhta 

I can do
`x = driverCPUExecutor_ ? driverCPUExecutor_->threads() : size_t();
`
If that's better?

Yuhta (approved):
LGTM, you might need an owner stamp for it to be merged though

gggrace14 (approved):

PR #22186, opened 2024-03-13
Add docs for isRowPositionList

elharo:
## Description
Javadoc for a constructor argument with unclear purpose

## Motivation and Context
Save my work in the code, before I have to figure this out again in two weeks. 

## Impact
nona

## Test Plan
CI

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* ...
* ...

Hive Changes
* ...
* ...
```

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-hive/src/main/java/com/facebook/presto/hive/orc/OrcBatchPageSource.java
+++ b/presto-hive/src/main/java/com/facebook/presto/hive/orc/OrcBatchPageSource.java
@@ -83,6 +83,12 @@ public class OrcBatchPageSource
         this(recordReader, orcDataSource, columns, typeManager, systemMemoryContext, stats, runtimeStats, nCopies(columns.size(), false));
     }
 
+    /**
+     * @param columns an ordered list of the fields to read
+     * @param isRowPositionList list of indices of columns. If true, then the column with the matching
+     *     index is a row number. If false, it isn't. This should have the same length as {@code columns},
+     *     and the columns in each list should match up.
+     */
     public OrcBatchPageSource(
             OrcBatchRecordReader recordReader,
             OrcDataSource orcDataSource,
@@ -100,7 +106,7 @@ public class OrcBatchPageSource
 
         this.stats = requireNonNull(stats, "stats is null");
         this.runtimeStats = requireNonNull(runtimeStats, "runtimeStats is null");
-        this.isRowPositionList = requireNonNull(isRowPositionList, "rowPosIndices is null");
+        this.isRowPositionList = requireNonNull(isRowPositionList, "isRowPositionList is null");
 
         this.constantBlocks = new Block[size];
         this.hiveColumnIndexes = new int[size];


presto-hive/src/main/java/com/facebook/presto/hive/orc/OrcBatchPageSource.java
```diff
@@ -100,7 +106,7 @@ public OrcBatchPageSource(
 
         this.stats = requireNonNull(stats, "stats is null");
         this.runtimeStats = requireNonNull(runtimeStats, "runtimeStats is null");
-        this.isRowPositionList = requireNonNull(isRowPositionList, "rowPosIndices is null");
+        this.isRowPositionList = requireNonNull(isRowPositionList, "isRowPositionList is null");
```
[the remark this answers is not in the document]
elharo: Good news is we will have a HiveColumnHandle for the row number. I suppose the question is whether we want to refactor this existing code along the way. The current approach with a list of booleans to tell which columns are row IDs is really strange.

presto-hive/src/main/java/com/facebook/presto/hive/orc/OrcBatchPageSource.java
```diff
@@ -83,6 +83,12 @@ public OrcBatchPageSource(
         this(recordReader, orcDataSource, columns, typeManager, systemMemoryContext, stats, runtimeStats, nCopies(columns.size(), false));
     }
 
+    /**
+     * @param columns an ordered list of the fields to read
+     * @param isRowPositionList list of indices of columns. If true, then the column with the matching
```
mbasmanova, line 88: > list of indices of columns

This is a list of booleans, not indices, right?

This design is indeed strange.

mbasmanova: Can there be > 1 'row number' column? What does that mean?

tdcmeehan:
I believe it was coded this way to allow the ORC read to infer position columns without having to introduce a type for this.  Reason being, Hive is rather large and can be thought of as two things: a connector for accessing the Hive table format, and a library for distributed files.  The latter is just because Hive was the first to do this, but over time, other table formats have evolved, but when it comes to the data layer they are all mostly similar at that point (they differ significantly in the metadata layer).  So, to avoid having to introduce a type in Hive, where there is no formal specification of how to define the position, this was added so that Iceberg could access its own definition of position, without having to introduce this undefined concept in Hive.

With this context in mind, if we feel there is a refactor to remove this list of booleans without necessarily introducing a new convention in Hive which may not be well defined, please let's create an issue for it.

tdcmeehan:
IMO a sensible refactor is to just provide an optional integer input.  If present, this integer is validated to be within the indices of the types for the reader.  We use this to infer which column is the row position list column.  I think it's safe to just go ahead and do this refactor.

elharo (commented):
Issue #22231 filed to clean this up, but until someone gets around to doing that, we should merge this PR to update the current docs.

tdcmeehan (approved):

PR #22203, opened 2024-03-14
Fix aggregation metadata optimization when existing Limit/TopN

hantangwangd:
## Description

When LimitNode/TopNNode exists below the aggregation node, we should not do metadata optimization for the aggregation. As `LimitNode` and `TopNNode` are cardinality-sensitive, so they may lead in some problems. For example, on a hive table:

```
CREATE TABLE test_metadata_query_optimization_with_limit(a varchar, b int, c int) WITH (partitioned_by = ARRAY['b', 'c']);
INSERT INTO test_metadata_query_optimization_with_limit VALUES('1001', 1, 1), ('1002', 1, 1), ('1003', 1, 1), ('1004', 1, 2), ('1005', 1, 2), ('1006', 1, 2), ('1007', 2, 1), ('1008', 2, 1), ('1009', 2, 1);
```

It would not lead in wrong result when do metadata optimization in the following statement, as limit node exists above aggregation in the plan:

```
select distinct b, c from test_metadata_query_optimization_with_limit order by c desc limit 3;
 b | c 
---+---
 1 | 2 
 1 | 1 
 2 | 1 
(3 rows)
```

But when we try to do metadata optimization in the following statement:

```
with tt as (select b, c from test_metadata_query_optimization_with_limit order by c desc limit 3) select b, min(c), max(c) from tt group by b;
```

We would get a wrong result as follows, this is because the limit node exists below aggregation in the plan:

```
 b | min(c) | max(c) 
---+-------+-------
 1 |     1 |     2
 2 |     1 |     1
```

However, the correct result should be:

```
 b | min(c) | max(c) 
---+-------+-------
 1 |     2 |     2 
```

The PR fix this problem.

## Test Plan

 - Newly added tests for the cases described above

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-hive/src/test/java/com/facebook/presto/hive/TestHiveLogicalPlanner.java
+++ b/presto-hive/src/test/java/com/facebook/presto/hive/TestHiveLogicalPlanner.java
@@ -165,6 +165,50 @@ public class TestHiveLogicalPlanner
     }
 
     @Test
+    public void testMetadataQueryOptimizationWithLimit()
+    {
+        QueryRunner queryRunner = getQueryRunner();
+        Session sessionWithOptimizeMetadataQueries = getSessionWithOptimizeMetadataQueries();
+        Session defaultSession = queryRunner.getDefaultSession();
+        try {
+            queryRunner.execute("CREATE TABLE test_metadata_query_optimization_with_limit(a varchar, b int, c int) WITH (partitioned_by = ARRAY['b', 'c'])");
+            queryRunner.execute("INSERT INTO test_metadata_query_optimization_with_limit VALUES" +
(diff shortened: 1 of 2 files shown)


feilong-liu (approved):
Nice fix

jaystarshot (approved):

ClarenceThreepwood (approved):
lgtm!

Follow on question - should we be turning this on by default? Is that how you unearthed this bug?

tdcmeehan:
@ClarenceThreepwood I believe the problem is it depends on one's knowledge regarding accurate statistics, which in some connectors or deployments is not guaranteed.

tdcmeehan (approved):

PR #22205, opened 2024-03-14
Improve cte scheduling

### LINKED ISSUE #21639: Improve scheduling of independent ctes in CTE Materialization
Currently when multiple ctes are materialized, they will be executed one by one. This is necessary for  dependent CTEs but independent ctes can be scheduled for execution parallelly.
This will improve end to end latency 

Relavant code link  - [here](https://github.com/prestodb/presto/blob/77dfcc805ec1ce06d0cc95ebf3f10d7b7838aad8/presto-main/src/main/java/com/facebook/presto/sql/planner/BasePlanFragmenter.java#L208)

Enhance https://github.com/prestodb/presto/issues/19744

jaystarshot:
## Description

Fixes #21639 

Previously each materialized cte was scheduled one by one for safe dependent cte handling. This PR reduces latency improves cte scheduling by scheduling multiple subgraphs of dependent ctes independently.
This PR has these major changes

Commit 1 - [link](https://github.com/prestodb/presto/pull/22205/commits/76a0c2de9521007791aaa682f1ee6fc1e1177fab)
1. Adding dependency graph in SequenceNode
2. Using the dependency graph to schedule subgraphs in basePlanFragmenter
3. Updating optimizers especially [SimplifyPlanWithEmptyInput.java](https://github.com/prestodb/presto/pull/22205/commits/97fb98e7673f587fe98a0ef3e9b1e1c0128439ba#diff-2341227ca888731a72c9e28c05277e64930ce13aee3ca6af0d744c1e6923c035) to properly process this depenent graph

Commit 2  [link](https://github.com/prestodb/presto/pull/22205/commits/edd771f778d3655e4287b35f0b06bdb5164fbb4f) 
1. Moving SequenceNode to presto-main since the graph dependency is not available in presto-spi 

## Test Plan

Tested with prod queries and confirmed from the splits page that the behavior was as expected for independent ctes

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Improved latency of materialized ctes by scheduling multiple dependent subgraphs independently.

### CODE REVIEW

**Comment by jaystarshot** [comment]
This works, but going for the ultimate latency where only the cte consumers would be blocked till producers complete

**Comment by rschlussel** [comment]
I'm having trouble understanding why we need to keep track of the whole cte graph vs. just the direct dependencies of the sequence node. 
I think an example of the plan for dependent ctes would help me.  e.g. what would the plan look like after logical optimization for a query like
`with cte_1 as (select * from table), cte2 as (SELECT * from cte1 where a > b), SELECT * From cte1 join cte2 on cte1.x = cte2.x`?


(also, for the commits - the commit that moves sequence node out of spi should be first since the other one depends on it.  And also the other commit, take out the uber specific references and maybe add more information to the body).

**Comment by jaystarshot** [comment]
After logical planning, the sequence node stores the cte producers in a topologically sorted list.
In your example, the graph would be **cte2->cte1** (since cte1 must be executed before cte2).
So the plan would be 

```
Sequence 
   CteProducers:     1. cte2
                     2. cte1
   PrimarySource: (SELECT * FROM cte2 JOIN cte1 ON..)
```

We need to keep a track of the whole graph since in this example cte1 must be executed before cte2 since cte1 is a dependency of cte2

Also consider cases like  Graph (cte3->cte4, cte2->cte1, cte1). Here the two subgraphs are independent and can be scheduled independently. same for (cte1, cte2, cte3) where all are independent.

**Comment by rschlussel** [comment]
> After logical planning, the sequence node stores the cte producers in a topologically sorted list. In your example, the graph would be **cte2->cte1** (since write of cte1 must be executed before write of cte2). So the plan would be
> 
> ```
> Sequence 
>    CteProducers:     1. cte2
>                      2. cte1
>    PrimarySource: (SELECT * FROM cte2 JOIN cte1 ON..)
> ```
> 
> We need to keep a track of the whole graph since in this example cte1 must be executed before cte2 since cte1 is a dependency of cte2
> 
> Also consider cases like Graph (cte3->cte4, cte2->cte1, cte5). Here the three subgraphs are independent and can be scheduled independently. same for (cte1, cte2, cte3) where all are independent.

But does cte2 have its own sequence node where it depends on cte1?   Something like below:
```
Sequence 
   CteProducers:     1. cte1
                                2. sequence (cte2)
                                        CteProducers:
                                             1. cte1
                                            PrimarySource: `SELECT * from cte1 where a < b`
                                  
   PrimarySource: (SELECT * FROM cte2 JOIN cte1 ON..)I'm basically trying 
```
I'm basically trying to understand why the graph needs to be specified explicitly vs. derived from the plan structure.

**Comment by jaystarshot** [comment]
Ahh I see, nope currently there is only one sequence node per query and it is responsible for maintaining all these dependencies.  Reasoning behind this was ease of other optimizations and cases where its difficult to represent in the plan structure since there is a common node `eg: graph (a->b, c->b) and graph(a->b, a->c)`

File: presto-main/src/main/java/com/facebook/presto/sql/planner/BasePlanFragmenter.java [hunk 1]
```diff
@@ -204,25 +204,33 @@ public PlanNode visitTableFinish(TableFinishNode node, RewriteContext<FragmentPr
...
-        SubPlan lastSubPlan = buildSubPlan(source, childProperties, context);
-
-        for (int sourceIndex = cteProducerCount - 2; sourceIndex >= 0; sourceIndex--) {
-            source = node.getCteProducers().get(sourceIndex);
-            childProperties = new FragmentProperties(new PartitioningScheme(
+        // To ensure that the execution order is maintained, we use an independent dependency graph.
+        // This approach creates subgraphs sequentially, enhancing control over the execution flow. However, there are optimization opportunities:
+        // 1. Can consider blocking only the CTEConsumer stages that are in a reading state.
+        //    This approach sounds good on paper may not be ideal as it can block the entire query, leading to resource wastage since no progress can be made until the writing operations are complete.
+        // 2. ToDo: Another improvement will be to schedule the execution of subgraphs based on their order in the overall execution plan instead of a topological sorting done here
```
rschlussel, line 211: general question again, but still trying to understand the basics. can you draw out what a fragmented plan with dependent ctes would look like before and after this change?

jaystarshot: Sure, lets take an example case of **Graph(a, b)**  or sql 
`WITH t as (SELECT * FROM orders), t1 as (Select * FROM customer) SELECT * FROM t JOIN t1 ON true`

Lets say that the Plan would be the below for simplicity.

```
...
     Exchange
        Sequence:
                           cteProducer1
                           cteProducer2
       PrimarySource:
                      Exchange
                                  Join 
                                     ....

```

Lets say the primary source creates subplan (subPrimary1->subPrimary2) 
and creates a list of subplans for the cteProducers as (subCteProducerT1 and subCteProducerT2)
The code is a bit tricker here but in the visitSequence we add the the cteproducers to the context created by the parent exchange. ([here](https://github.com/prestodb/presto/blob/1367d3caa79a3a9b62e3821f0cb27760a0d39e4c/presto-main/src/main/java/com/facebook/presto/sql/planner/BasePlanFragmenter.java#L300) is actually where the parent exchange creates new properties or context)

so before the subplans look like


```
SubParent -> subPrimary1 -> subPrimary2
          -> subCteProducer1 -> subCteProducer2
```
[code](https://github.com/prestodb/presto/blob/master/presto-main/src/main/java/com/facebook/presto/sql/planner/BasePlanFragmenter.java#L222) is where we were adding the subCteProducer2 as a child of subCteProducer1

Now we would add the subCteProducer to the same level as the subCteProducer1

```
SubParent ->  subPrimary1 -> subPrimary2
           -> subCteProducer1 
           -> subCteProducer2
```
                                     
In both cases similar structure would be represented in the sectionedPlan

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/LogicalCteOptimizer.java
```diff
@@ -319,9 +325,10 @@ public Optional<String> peekActiveCte()
 
         public void addDependency(String currentCte)
         {
-            graph.addNode(currentCte);
+            cteDependencyGraph.addNode(currentCte);
             Optional<String> parentCte = peekActiveCte();
-            parentCte.ifPresent(s -> graph.putEdge(currentCte, s));
+            // (current -> parentCte) this indicates that currentCte must be processed first
```
jaystarshot, line 330: Fixed wrong comments about the dependency graph and added explanations

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/LogicalCteOptimizer.java
```diff
@@ -342,9 +349,29 @@ public boolean isComplexCte(String cteId)
         public List<PlanNode> getTopologicalOrdering()
         {
             ImmutableList.Builder<PlanNode> topSortedCteProducerListBuilder = ImmutableList.builder();
-            Traverser.forGraph(graph).depthFirstPostOrder(graph.nodes())
+            Traverser.forGraph(cteDependencyGraph).depthFirstPostOrder(cteDependencyGraph.nodes())
                     .forEach(cteName -> topSortedCteProducerListBuilder.add(cteProducerMap.get(cteName)));
```
jaystarshot, line 353: Here we are not reversing the list and hence the topological ordering is correct, if we reverse the graph, we should reverse the list here and in SequenceNode

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/LogicalCteOptimizer.java
jaystarshot: I could reverse the graph to be according to the comments and then add a reverse in topological ordering, but then i would have to update some tests (since topological ordering is not unique). Its easier to update the comments

presto-main/src/test/java/com/facebook/presto/cost/TestCostCalculator.java
```diff
@@ -531,7 +532,7 @@ public void testSequence()
                 Optional.empty(),
                 new PlanNodeId("sequence"),
                 ImmutableList.of(cteProducerNode1, cteProducerNode2),
-                joinNode);
+                joinNode, GraphBuilder.directed().build());
```
rschlussel, line 535: put this on its own line

rschlussel (approved):

PR #22207, opened 2024-03-14
[Iceberg] Skip data size stat on fixed-width types

### LINKED ISSUE #22201: Iceberg column stats - Do not store data size stats for numeric types
In [22161](https://github.com/prestodb/presto/issues/22161#issuecomment-1993509424) we observed that when ANALYZE-ing Iceberg tables, we were storing data size stats for all column types. This differs from the default behavior of how we store stats in Hive tables, and causes the CBO to underestimate the size of the build side of a join

When data size stats tracking [was introduced](https://github.com/prestodb/presto/commit/9b31a8c6394ad940363989adee9c3f53ddd8bcb3#diff-0c0891a866a8fb3f6d18febf76fca447cb99fc0a22941d31b22142fff47b3acfR658-R661), we made an implicit assumption that this would be stored only for VARCHAR, CHAR and other variable width data types. This [assumption](https://github.com/prestodb/presto/blob/27c6d9e60618c3607bef6b7732c24b6e739b957d/presto-main/src/main/java/com/facebook/presto/cost/PlanNodeStatsEstimate.java#L203-L205) still holds, and we enforce this via the [`getSupportedColumnStatistics`](https://github.com/prestodb/presto/blob/08c49debf27a9668af0db15c163bfa74406bf0d2/presto-hive-metastore/src/main/java/com/facebook/presto/hive/metastore/thrift/ThriftHiveMetastore.java#L406-L411) method

## Expected Behavior & Possible solution
Instead of simply converting the [`Collection<ComputedStatistics>`](https://github.com/prestodb/presto/blob/08c49debf27a9668af0db15c163bfa74406bf0d2/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergHiveMetadata.java#L538) to `PartitionStatistics`, we need to convert/translate the stats to only those supported via `getSupportedColumnStatistics`

ZacBlanco:
## Description

This change will now skip any columns which are fixed-width types when calculating the column statistics

## Motivation and Context

The optimizer utilizes the data size stats during join enumeration to determine whether to do a partitioned or broadcast join. The stats collection code before this change used the manifest file's data size to estimate the total size. This value is actually incorrect most of the time and may change depending on the file format, compression, etc.

Since fixed-width type sizes are known prior, we can skip collecting them and let the optimizer infer the data size instead. This is the same behavior as in the hive connector.

Fixes #22201  

## Impact

`SHOW STATS` will now return `NULL` for any fixed-width primitive types.

## Test Plan

New test added

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/TableStatisticsMaker.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/TableStatisticsMaker.java
@@ -15,6 +15,8 @@ package com.facebook.presto.iceberg;
 
 import com.facebook.airlift.log.Logger;
 import com.facebook.presto.common.predicate.TupleDomain;
+import com.facebook.presto.common.type.FixedWidthType;
+import com.facebook.presto.common.type.TypeManager;
 import com.facebook.presto.hive.NodeVersion;
 import com.facebook.presto.spi.ConnectorSession;
 import com.facebook.presto.spi.Constraint;
@@ -57,6 +59,7 @@ import java.nio.ByteOrder;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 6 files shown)


presto-iceberg/src/main/java/com/facebook/presto/iceberg/TableStatisticsMaker.java
```diff
@@ -322,11 +329,16 @@ public void updateColumnSizes(Partition summary, Map<Integer, Long> addedColumnS
         }
         for (Types.NestedField column : summary.getNonPartitionPrimitiveColumns()) {
             int id = column.fieldId();
-
-            Long addedSize = addedColumnSizes.get(id);
-            if (addedSize != null) {
-                columnSizes.put(id, addedSize + columnSizes.getOrDefault(id, 0L));
+            com.facebook.presto.common.type.Type type = toPrestoType(column.type(), typeManager);
+            if (type instanceof FixedWidthType) {
+                continue;
```
aaneja, line 334: Can you add a note explaining why we're doing this

aaneja (approved):

presto-iceberg/src/main/java/com/facebook/presto/iceberg/TableStatisticsMaker.java
ZacBlanco: fixed

aaneja (approved):

ClarenceThreepwood (approved):
lgtm!

tdcmeehan (approved):

PR #22230, opened 2024-03-18
Fix long overflow for OperatorStats

kewang1024:
In OperatorStats, summing long types leads to overflow.  Use double type to do calculation. 
If final result is larger than Long.MAX_VALUE, return Long.MAX_VALUE.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/operator/OperatorStats.java
+++ b/presto-main/src/main/java/com/facebook/presto/operator/OperatorStats.java
@@ -659,21 +659,21 @@ public class OperatorStats
         long addInputCalls = this.addInputCalls;
         long addInputWall = this.addInputWall.roundTo(NANOSECONDS);
         long addInputCpu = this.addInputCpu.roundTo(NANOSECONDS);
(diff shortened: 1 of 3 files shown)


presto-main/src/main/java/com/facebook/presto/operator/OperatorStats.java
```diff
@@ -817,6 +817,11 @@ public OperatorStats add(Iterable<OperatorStats> operators)
                 joinProbeKeyCount);
     }
 
+    public static long convertDoubleToLong(double dataSize)
```
tdcmeehan, line 820: I think we could remove this and just use a simple cast.  See https://docs.oracle.com/javase/specs/jls/se8/html/jls-5.html#jls-5.1.3

tdcmeehan (changes requested):
Can you add a test case for this in `TestOperatorStats`?

kewang1024:
Thanks @tdcmeehan for your good suggestions, just updated accordingly

PR #22233, opened 2024-03-18
Fix pull lambda expression optimizer to skip special form expression with long arguments list

feilong-liu:
## Description
This optimizer will visit the expressions to find validate expression to pull from body of lambda expression. However, we see that this optimizer can have significant runtime when the number of arguments in expression is extremely large. This PR will skip the check when this is the case.

## Motivation and Context
Queries with large IN clause see extreme latency for this optimizer.

## Impact
Avoid query planning timeout for these queries.

## Test Plan
Test locally to make sure the failed queries succeed now.

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Fix the lambda expression pull up optimizer for large queries to avoid query planning timeout
```


presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/PullUpExpressionInLambdaRules.java
```diff
@@ -295,12 +296,13 @@ private RowExpression getArgumentForRegexTypeExpression(RowExpression rowExpress
         @Override
         public Boolean visitSpecialForm(SpecialFormExpression specialForm, Boolean context)
         {
-            if (UNSUPPORTED_TYPES.contains(specialForm.getForm())) {
+            if (UNSUPPORTED_TYPES.contains(specialForm.getForm()) || specialForm.getArguments().size() > MAX_NUM_OF_ARGUMENTS) {
```
kaikalur, line 299: There are many other visitors that seem to work fine with this kind of expression. So I suggest looking closer into what you are doing here that's causing the slowdown. Maybe some inefficient data structures? Let's root cause this.

kaikalur (changes requested):

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/PullUpExpressionInLambdaRules.java
```diff
@@ -295,12 +297,13 @@ private RowExpression getArgumentForRegexTypeExpression(RowExpression rowExpress
...
                 return false;
             }
-            List<RowExpression> validArguments = getValidArguments(specialForm);
-            Map<RowExpression, Boolean> validRowExpressionMap = specialForm.getArguments().stream().distinct().collect(toImmutableMap(identity(), x -> validArguments.contains(x) ? x.accept(this, context) : false));
```
feilong-liu, line 302: This is the problem, where we call `validArguments.contains(x)` for every argument. When validArguments is a list, it will be a O(N) operation for the contains, hence this line will be O(N^2). 
I verified by changing the List to Set, the latency of this optimizer is reduced significantly. 
[Using list](https://www.internalfb.com/intern/presto/query/?query_id=20240319_211452_00011_frxfm#runtime_stats) vs. [Using set ](https://www.internalfb.com/intern/presto/query/?query_id=20240319_211109_00005_frxfm#runtime_stats)

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/PullUpExpressionInLambdaRules.java
feilong-liu: Root cause that the main contribution of the latency is the List contains check commented below. Fix this part to use a Set instead.

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/PullUpExpressionInLambdaRules.java
kaikalur: Add a synthetic test with 100k IN list to verify this?

feilong-liu: In addition to change the List to Set, I also skip check when the number of arguments exceed 1000. This makes the synthetic test not working

kaikalur:
Why set the limit? 100k element map should be ok?

Get Outlook for Android<https://aka.ms/AAb9ysg>

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/PullUpExpressionInLambdaRules.java
pranjalssh: @feilong-liu is the limit still needed

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/PullUpExpressionInLambdaRules.java
```diff
@@ -254,18 +255,18 @@ public Boolean visitCall(CallExpression call, Boolean context)
         }
 
         // For the conditional expressions, not all arguments will be evaluated, we only try to extract from the arguments which will always be executed
-        private static List<RowExpression> getValidArguments(SpecialFormExpression specialForm)
+        private static Set<RowExpression> getValidArguments(SpecialFormExpression specialForm)
         {
-            List<RowExpression> validArgument;
+            Set<RowExpression> validArgument;
```
mlyublena, line 260: nit: variable name: validArguments

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/PullUpExpressionInLambdaRules.java
```diff
@@ -213,6 +213,7 @@ private static class ValidExpressionExtractor
         // Bind expression will complicate the lambda expression, we apply this optimization before DesugarLambdaRule. And if there are bind expression, skip
         private static final List<SpecialFormExpression.Form> UNSUPPORTED_TYPES = ImmutableList.of(SpecialFormExpression.Form.BIND);
         private static final List<Class<?>> SUPPORTED_JAVA_TYPES = ImmutableList.of(boolean.class, long.class, double.class, Slice.class, Block.class);
+        private static final int MAX_NUM_OF_ARGUMENTS = 1_000;
```
mlyublena, line 216: is that a realistic upper limit? should we have something smaller?

mlyublena (dismissed):

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/PullUpExpressionInLambdaRules.java
feilong-liu: Removed, I do not think it's needed now.

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/PullUpExpressionInLambdaRules.java
feilong-liu: Removed the limit, I think it's not needed now.

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java
```diff
@@ -7325,6 +7325,24 @@ public void testLambdaExpressionPullUp()
         assertQuery(session, "SELECT ANY_MATCH(a, x -> (x LIKE CONCAT('a', b))) AS rejected_by_control_model, FILTER(a, x -> x LIKE CONCAT('a', b) ESCAPE '#') FROM ( SELECT ARRAY[CAST(random() AS VARCHAR)] a, CAST(random() AS VARCHAR) AS b )", "values (false, array[])");
     }
 
+    @Test(timeOut = 60_000)
```
feilong-liu, line 7328: Verified that it will timeout without change in this PR

jaystarshot (approved):

mlyublena (approved):

PR #22243, opened 2024-03-20
[native] Let API report correct query memory reservations

tanjialiang:
Current code reports wrong memory to coordinator. Report the sum of all query root pool consumptions instead. This can be further improved by reporting all managed pool usages in memory manager but that will require additional API changes to memory manager
```
== NO RELEASE NOTE ==
```


xiaoxmeng (approved):
@tanjialiang thanks!

PR #22253, opened 2024-03-20
[native]Call cache shrink from Periodic Task manager

mohsaka:
## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* ...
* ...

Hive Changes
* ...
* ...
```

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
+++ b/presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
@@ -12,6 +12,20 @@
  * limitations under the License.
  */
 
+#if defined(__APPLE__)
+#include <mach/mach_host.h>
+#include <mach/mach_init.h>
+#include <mach/mach_types.h>
+#include <mach/vm_statistics.h>
+#elif defined(__linux__)
+#include <folly/File.h>
+#include <folly/FileUtil.h>
+#include <sys/sysctl.h>
+#include <sys/sysinfo.h>
+#include <sys/types.h>
+#include <cinttypes>
+#endif
+
 #include "presto_cpp/main/PeriodicTaskManager.h"
 #include <folly/executors/CPUThreadPoolExecutor.h>
 #include <folly/stop_watch.h>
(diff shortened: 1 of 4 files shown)


mohsaka:
@xiaoxmeng We are currently trying to automatically shrink the cache utilizing the periodic task manager however we are experiencing a large amount of these spikes. I believe this is due to the system-memory-gb memory usage being larger than <system memory limit> - <non-system-memory-gb memory usage>. As a result, whenever we clear the cache, it just fills up again. This is with a SF-10k TPC-DS dataset on a 8 Node cluster with 122GiB container memory limit. 

Screenshot 2024-03-19 at 12 20 41 PM

Our condition for the shrink was 5GiB from the system memory limit. Which is at 117GiB. With a polling interval of 100ms. 

A few questions.
1. From the memory documentation it seems like Meta is doing 80% system-memory-gb, with a 98% of system memory limit as the condition. Is this correct?
2. What is the rate that you check the memory usage? I tried 2 and 1 seconds however it was too infrequent and resulted in OOMKills. Deepak is concerned that at 100ms its running too often and is taking up a thread.
3. Do you experience this up and down spiking as well?

Thanks!


cc @majetideepak

presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
```diff
@@ -420,6 +437,83 @@ void PeriodicTaskManager::addCacheStatsUpdateTask() {
...
+    LOG(INFO) << "System memory limit: " << containerMemoryLimit;
+  }
+  inUseMemory = (memInfo.totalram - memInfo.freeram) * memInfo.mem_unit;
+  return (containerMemoryLimit - inUseMemory) < (minReserveGB << 30);
+}
+#endif
+
+void PeriodicTaskManager::addCacheShrinkTask() {
+#if defined(__APPLE__)
+  addTask(
```
czentgr, line 495: If we set this to 0 to disable (and not have the default) - should we skip creating the task?
If it is 0 currently then it looks like it will run without wait.

I think we should have some way to disable it.

mohsaka: Sure

mohsaka: Done

majetideepak (commented):
@mohsaka Great addition! Let's make this pluggable as different platforms have different shrink conditions and values.
1) Let's follow the `registerTaskListener` model where we register a `class CacheShrink`.
2) This `class CacheShrink` will be a pure virtual class similar to `class TaskListener` and has a method, `bool runCacheShrink() = 0`.

The class implementation will add another method `shrinkConditionFunc()` if need be.
Let's move the example shrink functions here to a test and add a test.

Issue #21841, opened 2024-02-02
Improve Redis HBO Provider default configs and error messaging


jaystarshot:
The redis hbo provider currently has no default config values. Also the error messaging is not wrapped in a custom error package

PR #22282, opened 2024-03-21
Fix Flaky RedisHboProvider test

### LINKED ISSUE #22261: TestRedisProviderPlugin.testStartup is flaky
Seen here: https://github.com/prestodb/presto/actions/runs/8364278644/job/22899225390?pr=22258

jaystarshot:
## Description
Fixes #22261 
and #21841 except default values

```
Error:  Tests run: 22, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 200.24 s <<< FAILURE! - in TestSuite
Error:  com.facebook.presto.statistic.TestRedisProviderPlugin.testStartup  Time elapsed: 0.004 s  <<< FAILURE!
java.net.BindException: Address already in use (Bind failed)
```
Error message in one of the runs. This could happen if a test Redis server was initiated on a port that was already in use by another test's server. To prevent this, we now ensure the test creates the Redis server using a random port.

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```


redis-hbo-provider/src/test/java/com/facebook/presto/statistic/TestRedisProviderPlugin.java
```diff
@@ -15,21 +15,37 @@
...
     @Test
     public void testStartup()
             throws Exception
     {
-        RedisServer server = RedisServer.newRedisServer(57872);
+        RedisServer server = RedisServer.newRedisServer(); // bind to a random port
         server.start();
-        RedisProviderPlugin plugin = new RedisProviderPlugin(redisHboPropertyFilePath);
+        RedisProviderPlugin plugin = new RedisProviderPlugin(
+                createConfigMap(RedisURI.create(server.getHost(), server.getBindPort()).toString()));
```
jaystarshot, line 48: Removed the resources file because the plugin expects the resources file to contain the uri but since we are now generating the port randomly, we would not know beforehand. This was the easiest modification

redis-hbo-provider/src/test/java/com/facebook/presto/statistic/TestRedisProviderPlugin.java
```diff
@@ -15,21 +15,37 @@
...
+import java.util.Map;
+
 import static com.facebook.airlift.testing.Assertions.assertInstanceOf;
 
 public class TestRedisProviderPlugin
 {
-    private final String redisHboPropertyFilePath = "src/test/java/resources/test-redis-hbo-provider.properties";
+    private Map<String, String> createConfigMap(String serverUri)
+    {
+        Map<String, String> configMap = new HashMap<>();
```
tdcmeehan, line 30: Please use an immutable map

redis-hbo-provider/src/main/java/com/facebook/presto/statistic/RedisProviderPlugin.java
```diff
@@ -35,6 +35,10 @@ public class RedisProviderPlugin
     private Injector injector;
 
     @VisibleForTesting
+    public RedisProviderPlugin(Map<String, String> configs)
```
tdcmeehan, line 38: Can we have the constructor below delegate to this constructor?

redis-hbo-provider/src/main/java/com/facebook/presto/statistic/RedisProviderPlugin.java
```diff
@@ -49,6 +48,12 @@ public RedisProviderPlugin(String configPath)
         this.propertyMap = propertyMap;
     }
 
+    @VisibleForTesting
+    public RedisProviderPlugin(Map<String, String> configs)
+    {
+        this.propertyMap = configs;
+    }
+
     public RedisProviderPlugin()
```
rschlussel, line 57: i would move this constructor first in the file

redis-hbo-provider/src/main/java/com/facebook/presto/statistic/RedisProviderPlugin.java
```diff
@@ -34,7 +34,6 @@ public class RedisProviderPlugin
 
     private Injector injector;
 
-    @VisibleForTesting
     public RedisProviderPlugin(String configPath)
```
rschlussel, line 37: can we inline this code in the no argument constructor now.  It's the only thing that calls this.

redis-hbo-provider/src/main/java/com/facebook/presto/statistic/RedisProviderPlugin.java
```diff
@@ -49,6 +48,12 @@ public RedisProviderPlugin(String configPath)
         this.propertyMap = propertyMap;
```
rschlussel, line 48: call the below constructor instead of setting it directly

redis-hbo-provider/src/main/java/com/facebook/presto/statistic/RedisProviderPlugin.java
```diff
@@ -49,6 +48,12 @@ public RedisProviderPlugin(String configPath)
...
     }
 
+    @VisibleForTesting
+    public RedisProviderPlugin(Map<String, String> configs)
+    {
+        this.propertyMap = configs;
```
rschlussel, line 54: if it's not null, make an immutable copy of the property map for safety.

redis-hbo-provider/src/main/java/com/facebook/presto/statistic/RedisProviderPlugin.java
rschlussel, line 37: also, can't comment on the line and it's not related to this fix, but why do we set the property map to null if we can't load the properties?  I'd expect that if someone set the properties, they would be surprised to discover they don't really have an hbo provider set up, so I think it would be better to throw an exception.

jaystarshot: Agreed

redis-hbo-provider/src/main/java/com/facebook/presto/statistic/RedisProviderPlugin.java
jaystarshot: ` Call to 'this()' must be first statement in constructor body` Java doesn't allow that

rschlussel: oh yup. wasn't thinking about all the stuff happening above

redis-hbo-provider/src/main/java/com/facebook/presto/statistic/RedisPlanStatisticsProvider.java
```diff
@@ -93,11 +93,13 @@ public Map<PlanNodeWithHash, HistoricalPlanStatistics> getStats(List<PlanNodeWit
                         output.put(redisFutureEntry.getKey(), redisFutureEntry.getValue().get());
                     }
                     catch (Exception e) {
+                        // This exception will be handled by the RedisProviderApiStats which will increment metrics and not fail the query if stats not found
                         log.error(String.format("Error reading done Redis future fore key %s", redisFutureEntry.getKey().toString()));
```
rschlussel, line 97: not part of this change, but typo here fore -> for.  but also, this exception and the below get logged in RedisProviderApiStats, so no need to log them here too.

redis-hbo-provider/src/main/java/com/facebook/presto/statistic/RedisProviderSerdeException.java
```diff
@@ -0,0 +1,32 @@
...
+        extends PrestoException
+{
+    public RedisProviderSerdeException(String message)
+    {
+        this(message, null);
+    }
+
+    public RedisProviderSerdeException(String message, Throwable cause)
+    {
+        super(GENERIC_INTERNAL_ERROR, "RedisProviderPlugin Exception: " + message, cause);
```
rschlussel, line 30: This seems worth having its own error code.  If the exceptions are specific to the Redis HBO provider, then maybe a separate error code provider/error code for this.  Otherwise, maybe an HBO failure error code?

rschlussel (approved):
Comments can be addressed as a follow up if you prefer.  But would be good to add more information to the body of the commit messages here to provide context (e.g. describing that the test would fail because the port was already in use, so fixed it by binding to any available port)

redis-hbo-provider/src/main/java/com/facebook/presto/statistic/RedisProviderSerdeException.java
jaystarshot: Its just one error code for now so will create a new HBO_FAILURE, this can be used in other places as well in future.

rschlussel (approved):

Issue #21315, opened 2023-11-06
CAST(varchar as <integral type>) doesn't trim leading and trailing whitespaces, but fails


gggrace14:
Current behavior of casting varchar to integral types is to throw exception if the varchar is valid but contains leading or trailing whitespaces. For example,
```
SELECT CAST('100 ' AS BIGINT)
```
throws the following exception
```
com.facebook.presto.spi.PrestoException: Cannot cast '100 ' to BIGINT
	at com.facebook.presto.type.VarcharOperators.castToBigint(VarcharOperators.java:185)
```
This is different from the behavior described in ANSI SQL

> ... The declared type of the result of the \<cast specification> is TD... let SD be the declared type of the \<value expression>... let SV be its value...
> If SD is character string, then SV is replaced by SV with any leading or trailing \<space>s removed.
> Case:
> i) If SV does not comprise a \<signed numeric literal>, then an exception condition is raised.
> ii) Otherwise, let LT be that \<signed numeric literal>. The \<cast specification> is equivalent to CAST ( LT AS TD )

Accordingly, TRY(CAST(varchar AS <integral type>) and TRY_CAST(varchar AS <integral type>) return NULL rather than the actual valid integral number.

Should Presto change the above CAST behavior to accept and trim leading and trailing whitespaces and be compliant with ANSI SQL?

The exception is thrown because the implementation of casting varchar to integral types directly delegates to Java, ex.
https://github.com/prestodb/presto/blob/04138bc26a4fc801630fe936463d89d071d00c0a/presto-main/src/main/java/com/facebook/presto/type/VarcharOperators.java#L176-L187
And Java does not accept whitespaces or any character other than decimal digits and sign character.
https://docs.oracle.com/javase/8/docs/api/java/lang/Long.html#parseLong-java.lang.String-

## Your Environment
* Presto version used:

## Expected Behavior
Casting of varchar trims leading and trailing whitespaces.

## Current Behavior
Casting of varchar to integral types throws exception if the varchar contains leading or trailing whitespaces

## Possible Solution
Casting of varchar trims leading and trailing whitespaces before delegating to Java `Long.parseLong()`.

## Context
Throwing exception makes Presto have a different semantics of above CAST from ANSI SQL and other mainstream engines or libraries that compliant with ANSI SQL, ex., MySQL, folly, etc.

mbasmanova:
CC: @aditi-pandit @tdcmeehan @majetideepak

mbasmanova:
[...]

PR #22284, opened 2024-03-21
Allow leading and trailing spaces in CAST(str as INTEGER)

spershin:
## Description
Allow leading and trailing spaces in CAST(str as INTEGER). Also as BIGINT, SMALLINT and TINYINT.
This is according to SQL standard.
Fixes https://github.com/prestodb/presto/issues/21315

## Motivation and Context
This is according to SQL standard.

## Impact
This is query behavior change.

## Test Plan
A new unit test that fails on the old code.
Also maybe need to update documentation to reflect this.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* CAST(str as INTEGER), CAST(str as BIGINT), CAST(str as SMALLINT), CAST(str as TINYINT) now allow leading and trailing spaces in the string.

```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/type/VarcharOperators.java
+++ b/presto-main/src/main/java/com/facebook/presto/type/VarcharOperators.java
@@ -179,7 +179,7 @@ public final class VarcharOperators
     public static long castToBigint(@SqlType("varchar(x)") Slice slice)
     {
         try {
-            return Long.parseLong(slice.toStringUtf8());
+            return Long.parseLong(slice.toStringUtf8().trim());
         }
         catch (Exception e) {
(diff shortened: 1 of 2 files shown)


mbasmanova:
CC: @tdcmeehan @steveburnett

elharo (changes requested):
This needs a test that fails with the old rules and passes with the new ones

spershin:
@elharo 

> This needs a test that fails with the old rules and passes with the new ones

Of course. We will have them.
Thing is, I haven't found the existing tests and wasn't sure if we don't have them or I just can't find them.
So decided to see if any checks/tests will fail.
See the comment in the "Test Plan" section of the PR Description.

spershin:
@elharo 

I have added a unit test as requested.

kaikalur (approved):

tdcmeehan (approved):
Thanks for fixing this.

PR #22286, opened 2024-03-21
Fix flaky test testRunawayRegexAnalyzerTimeout

### LINKED ISSUE #20411: Test TestHivePushdownDistributedQueries testRunawayRegexAnalyzerTimeout fails with a timeout
https://github.com/prestodb/presto/actions/runs/5685313042/job/15409924223?pr=20409

Error:  Failures: 
Error:    TestHivePushdownDistributedQueries.testRunawayRegexAnalyzerTimeout » Test 
The exception was thrown with the wrong message: expected "Regexp matching interrupted" but got "The query optimizer exceeded the timeout of 1.00s."

ClarenceThreepwood:
Fixes #20411 

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestDistributedQueries.java
+++ b/presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestDistributedQueries.java
@@ -373,7 +373,9 @@ public abstract class AbstractTestDistributedQueries
         computeActual("EXPLAIN ANALYZE DROP TABLE orders");
     }
 
-    @Test(expectedExceptions = RuntimeException.class, expectedExceptionsMessageRegExp = "Regexp matching interrupted", timeOut = 30_000)
+    @Test(expectedExceptions = RuntimeException.class,
+            expectedExceptionsMessageRegExp = "Regexp matching interrupted|The query optimizer exceeded the timeout of .*",
+            timeOut = 30_000)
     public void testRunawayRegexAnalyzerTimeout()
     {
         Session session = Session.builder(getSession())


tdcmeehan (dismissed):

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestDistributedQueries.java
```diff
@@ -373,12 +374,24 @@ public void testExplainAnalyzeDDL()
         computeActual("EXPLAIN ANALYZE DROP TABLE orders");
     }
 
-    // Flaky test: https://github.com/prestodb/presto/issues/20764
-    @Test(expectedExceptions = RuntimeException.class, expectedExceptionsMessageRegExp = "Regexp matching interrupted", timeOut = 30_000, enabled = false)
-    public void testRunawayRegexAnalyzerTimeout()
+    @DataProvider
+    public static Object[][] timeouts()
```
rschlussel, line 378: curious why you parametrized the test.

rschlussel (approved):
i apologize. i only saw this after i merged my change.  but adding the exception to the expected regex i think is probably a more reliable fix than what i had.  My only question is why test different timeout settings.

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestDistributedQueries.java
ClarenceThreepwood: With an optimizer timeout of 0.01s - you are much more likely to hit the other error message. But it might be overkill

rschlussel: yeah, i think the point of this test is to make sure that a long running regex gets interrupted.  It's just sometimes possible that the timeout will be hit elsewhere.

ClarenceThreepwood: ok - I'll remove the parametrization and keep the regex

rschlussel (approved):

PR #22292, opened 2024-03-22
[native] rename task memory reclaim runtime stats

xiaoxmeng:


amitkdutta (approved):
Looks good. Thanks @xiaoxmeng

presto-native-execution/presto_cpp/main/PrestoTask.cpp
```diff
@@ -485,10 +485,10 @@ protocol::TaskInfo PrestoTask::updateInfoLocked() {
   }
 
   if (taskStats.memoryReclaimCount > 0) {
-    taskRuntimeStats["memoryReclaimCount"].addValue(
+    taskRuntimeStats["taskMemoryReclaimCount"].addValue(
```
mbasmanova, line 488: Should we use constants for these and add comments explaining what these stats mean? Would be nice to also start documenting these. Thanks.

mbasmanova (approved):

presto-native-execution/presto_cpp/main/PrestoTask.cpp
xiaoxmeng: Yeah, using constant will ease management. Will do in the followup with documentation. Thanks!

PR #22294, opened 2024-03-22
Test case for Iceberg Equality delete bug returning incorrect number …

nmahadevuni:
…of rows

## Description
This is a test case for reproducing an issue with Iceberg equality delete. Not a fix. After equality deletes are applied to the table, it is returning incorrect number of rows. The equality delete is simulated by calling the Iceberg API to add a parquet file with deleted rows and modifying the table's metadata. Querying the table after applying equality deletes is compared against the same query with a NOT IN query to eliminate the deleted rows manually. These two do not match.

## Test Plan
TestIcebergEqualityDelete.java is the one to reproduce the issue.

```
== NO RELEASE NOTE ==
```


nmahadevuni:
Please see the linked issue for explanation.

PR #20687, opened 2023-08-28
Fix #20447: Wait for queries to start running and add retry on failure

imjalpreet:
## Description
Fix Flakiness in com.facebook.presto.tests.TestQueryManager#testQueryCountMetrics  
Also, introduce retry mechanism to the above test

## Motivation and Context

Fix #20447

## Contributor checklist

- [x] Please make sure your submission complies with our [development](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#development), [formatting](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#formatting), [commit message](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#commit-formatting-and-pull-requests), and [attribution guidelines](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#attribution).
- [x] PR description addresses the issue accurately and concisely.  If the change is non-trivial, a GitHub Issue is referenced.
- [x] Documented new properties (with its default value), SQL syntax, functions, or other functionality.
- [x] If release notes are required, they follow the [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines).
- [x] Adequate tests were added if applicable.
- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

PR #22308, opened 2024-03-22
Fix and enable flaky test TestQueryManager.testQueryCountMetrics 

ClarenceThreepwood:
Fixes ##20447

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-tests/src/test/java/com/facebook/presto/tests/TestQueryManager.java
+++ b/presto-tests/src/test/java/com/facebook/presto/tests/TestQueryManager.java
@@ -18,12 +18,12 @@ import com.facebook.presto.execution.QueryInfo;
 import com.facebook.presto.execution.QueryManager;
 import com.facebook.presto.execution.QueryState;
 import com.facebook.presto.execution.TestingSessionContext;
+import com.facebook.presto.plugin.blackhole.BlackHolePlugin;
 import com.facebook.presto.resourceGroups.FileResourceGroupConfigurationManagerFactory;
 import com.facebook.presto.server.BasicQueryInfo;
 import com.facebook.presto.server.testing.TestingPrestoServer;
 import com.facebook.presto.spi.PrestoException;
 import com.facebook.presto.spi.QueryId;
-import com.facebook.presto.tests.tpch.TpchQueryRunnerBuilder;
 import com.google.common.base.Stopwatch;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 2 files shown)


presto-tests/src/test/java/com/facebook/presto/tests/TestQueryManager.java
```diff
@@ -237,28 +237,29 @@ public void testQueryOutputSizeExceeded()
...
+    @Test(timeOut = 120_000L)
     public void testQueryCountMetrics()
             throws Exception
     {
         DispatchManager dispatchManager = queryRunner.getCoordinator().getDispatchManager();
-        // Create a total of 10 queries to test concurrency limit and
+        // Create a total of 6 queries to test concurrency limit and
         // ensure that some queries are queued as concurrency limit is 3
-        createQueries(dispatchManager, 10);
+        createQueries(dispatchManager, 6);
```
ClarenceThreepwood, line 247: 2-3 out of 10 queries were consistently failing for me on my local mac complaining of insufficient resources

ClarenceThreepwood:
I'm seeing good robust results after this fix. I'm inclined to think the retry mechanism in #20687 may not be necessary 
a) reduces the number of concurrent queries 
b) introduces a synchronization point for the queries in flight before checking state

tdcmeehan:
@ClarenceThreepwood can you modify this test to use the black hole connector (see #22307 and #22298)?  The idea is to use that connector to force queries to hang.  It might require a change to the test, which will improve the sequencing.

presto-tests/src/test/java/com/facebook/presto/tests/TestQueryManager.java
```diff
@@ -57,12 +58,20 @@
...
     public void setUp()
             throws Exception
     {
-        queryRunner = TpchQueryRunnerBuilder.builder().build();
+        queryRunner = getSimpleQueryRunner();
+        queryRunner.installPlugin(new BlackHolePlugin());
+        queryRunner.createCatalog("blackhole", "blackhole");
+        // Black hole connectors do not have external metadata, which means that table metadata is local to each coordinator.
+        // So for the purposes of the test setup, create a local table per connector.
+        for (int i = 0; i < queryRunner.getCoordinators().size(); i++) {
```
tdcmeehan, line 72: The referenced tests use multiple coordinators, however this one just uses a single coordinator, so we can remove the comment and the for loop and just use `queryRunner.getCoordinator()`.

tdcmeehan (approved):

PR #22327, opened 2024-03-25
[Iceberg] Collect data size statistics for Iceberg tables

### LINKED ISSUE #22208: [Iceberg] variable-width column data sizes are generally wrong
## Your Environment

Any iceberg table

## Expected Behavior

Data size statistics should accurately reflect the size in memory when operating in Presto.

## Current Behavior

Iceberg's `TableStatisticsMaker` tries to use the Iceberg manifest file information to calculate the data size for each column. This information is used in the optimizer for things like determining the join distribution type based on row size. However, in the iceberg spec, this data is actually the on-disk data size, not necessarily the in-memory size which is what we care about.

It turns out that most, if not all of the data size statistics iceberg reports is incorrect by an order of 3-5x. This amount could change depending on disk storage format, compression, encryption, etc.

## Possible Solution

There are two dimensions to the solution

1. Do we want to allow ANALYZE on iceberg tables to generate stats for data size?
2. Can we still utilize the manifest information, even when stats don't exist?

| | without ANALYZE | with ANALYZE |
|-|---|-|
| don't change manifest info | don't change the code at all | just add data size collection support to iceberg |
| do change manifest info | add some heuristics to TableStatisticsMaker | add support for correct data size with ANALYZE _and_ try to improve TableStatisticsMaker with heuristics| 

1. Don't return data size stats at all except after an `ANALYZE`
3. Return incorrect stats, but allow `ANALYZE` to overwrite and improve them.
4. Keep incorrect stats, but apply some adjustment factor based on file type/configuration

## Steps to Reproduce

Use the `IcebergQueryRunner` and execute `SHOW STATS` on any tpch/ds table.

```sql
SHOW STATS FOR (select comment from orders);
 column_name | data_size | distinct_values_count | nulls_fraction | row_count | low_value | high_value
-------------+-----------+-----------------------+----------------+-----------+-----------+------------
 comment     |  167815.0 | NULL                  |            0.0 | NULL      | NULL      | NULL
 NULL        | NULL      | NULL                  | NULL           |   15000.0 | NULL      | NULL
(2 rows)

Query 20240314_192652_00012_nmth9, FINISHED, 1 node
Splits: 1 total, 1 done (100.00%)
[Latency: client-side: 124ms, server-side: 83ms] [0 rows, 0B] [0 rows/s, 0B/s]
```

check the value returned by the aggregation function used to calculate data size (`sum_data_size_for_stats`)

```sql
presto:tpch> select sum_data_size_for_stats(comment) from orders;
 _col0
--------
 727364
(1 row)

Query 20240314_192726_00013_nmth9, FINISHED, 4 nodes
Splits: 9 total, 9 done (100.00%)
[Latency: client-side: 295ms, server-side: 244ms] [15K rows, 228KB] [61.5K rows/s, 934KB/s]
```
 

727364 > 167815 by a factor of about 4.5.

## Context

Can cause query slowdowns if incorrect join distribution type is chosen

ZacBlanco:
## Description

Previously, the data size statistic was computed by using the Iceberg data manifests data size field. This is value is misleading for Presto because it represents the compressed on-disk size rather than in-memory size.

This change allows ANALYZE to read and write data size statistic values to puffin files.

There are some other minor changes included in this PR

- Updates the `iceberg.hive-statistics-merge-strategy` configuration to pass a comma-separated list of overrides
- Updates the Iceberg changelog test so it doesn't always use port 8080

## Motivation and Context

Previous statistics reported by data size are wrong

Closes #22208 

## Impact

Presto-generated puffin files will now include a blob for data size after running `ANALYZE`

## Test Plan

New tests added to read+write data size.

## Contributor checklist

- [x] CI passed.

## Release Notes

```
General Changes
* The `iceberg.hive-statistics-merge-strategy` flag has been updated to accept a comma-separated list of the following values: NUMBER_OF_DISTINCT_VALUES, TOTAL_SIZE_IN_BYTES
```


presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -1172,7 +1173,7 @@ each Iceberg data type to the corresponding Presto data type, and from each Pres
 The following tables detail the specific type maps between PrestoDB and Iceberg. 
 
 Iceberg to PrestoDB type mapping
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
steveburnett, line 1176: Thanks for catching this, and also in line 1219.

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -225,10 +225,11 @@ Property Name                                           Description
 
 ``iceberg.enable-parquet-dereference-pushdown``         Enable parquet dereference pushdown.                          ``true``
 
-``iceberg.hive-statistics-merge-strategy``              Determines how to merge statistics that are stored in the     ``NONE``
-                                                        Hive Metastore. The available values are ``NONE``,
-                                                        ``USE_NULLS_FRACTION_AND_NDV``, ``USE_NULLS_FRACTIONS``
-                                                        and, ``USE_NDV``
+``iceberg.hive-statistics-merge-strategy``              Comma separated list of statistics to use from the
+                                                        hive metastore to override Iceberg table statistics. Only
```
steveburnett, line 229:
```suggestion
                                                        Hive Metastore to override Iceberg table statistics.
```
The sentence "Only valid when the Iceberg connector is configured with Hive." could be moved out of the middle of the paragraph to highlight it, which improves readability by making it easier for the user to quickly read, and dismiss this property if it's irrelevant to their situation. (I'm including a screenshot only because I'd already mocked this up in my local build to see how it looks.)
Screenshot 2024-03-26 at 1 08 17 PM

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -225,10 +225,11 @@ Property Name                                           Description
...
+                                                        valid when the Iceberg connector is configured with Hive.
```
steveburnett, line 230:
```suggestion
```

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -225,10 +225,11 @@ Property Name                                           Description
...
+                                                        The available values are ``NUMBER_OF_DISTINCT_VALUES``,
+                                                        ``NUMBER_OF_NON_NULL_VALUES``, ``TOTAL_SIZE_IN_BYTES``.
```
steveburnett, line 232:
```suggestion
                                                        ``NUMBER_OF_NON_NULL_VALUES``, ``TOTAL_SIZE_IN_BYTES``.
                                                        
                                                        Note: Only valid when the Iceberg connector is configured with Hive.
```

steveburnett (changes requested):
Thanks for the doc! One suggestion about moving a sentence to improve readiability. Let me know what you think, and if you have a better idea.

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -225,10 +225,13 @@ Property Name                                           Description
...
-                                                        Hive Metastore. The available values are ``NONE``,
-                                                        ``USE_NULLS_FRACTION_AND_NDV``, ``USE_NULLS_FRACTIONS``
-                                                        and, ``USE_NDV``
+``iceberg.hive-statistics-merge-strategy``              Comma separated list of statistics to use from the
+                                                        Hive Metastore to override Iceberg table statistics.
+                                                        The available values are ``NUMBER_OF_DISTINCT_VALUES``,
+                                                        ``NUMBER_OF_NON_NULL_VALUES``, ``TOTAL_SIZE_IN_BYTES``.
+
+                                                        **Note**: Only valid when the Iceberg connector is
+                                                        configured with Hive
```
steveburnett, line 234:
```suggestion
                                                        configured with Hive.
```
I think I left off the period of the sentence.

steveburnett (changes requested):
Thanks for the quick response, looks good. One tiny nit that I think I overlooked previously.

steveburnett (approved):
LGTM! (docs)

Pull updated branch, new local build, reviewed  and noticed the new content and formatting fixes, everything looks good. Thanks!

presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedTestBase.java
```diff
@@ -633,49 +634,92 @@ public void testStringFilters()
     }
 
     @Test
-    public void testReadWriteNDVs()
+    public void testReadWriteStats()
```
hantangwangd, line 637: Should we add test case to show the behavior for iceberg configured with hive but do not config `hive_statistics_merge_strategy`?

hantangwangd (commented):
Overall looks good to me, some little things for discussion.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/TableStatisticsMaker.java
```diff
@@ -164,20 +184,14 @@ private TableStatistics makeTableStatistics(IcebergTableHandle tableHandle, Cons
...
             Long nullCount = summary.getNullCounts().get(fieldId);
             if (nullCount != null) {
                 columnBuilder.setNullsFraction(Estimate.of(nullCount / recordCount));
             }
-            if (summary.getColumnSizes() != null) {
-                Long columnSize = summary.getColumnSizes().get(fieldId);
-                if (columnSize != null) {
-                    columnBuilder.setDataSize(Estimate.of(columnSize));
-                }
-            }
```
hantangwangd, line 180: Since the column size is useless now in summary, should we as well remove the invocation of `updateColumnSizes(...)` in `getSummaryFromFiles(...)` when getting the summary?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergHiveMetadata.java
```diff
@@ -455,7 +456,7 @@ public TableStatistics getTableStatistics(ConnectorSession session, ConnectorTab
             });
             RowExpression translatedPredicate = rowExpressionService.getDomainTranslator().toPredicate(predicate);
             PartitionStatistics hiveStatistics = metastore.getTableStatistics(getMetastoreContext(session), handle.getSchemaName(), handle.getIcebergTableName().getTableName());
-            TableStatistics mergedStatistics = mergeHiveStatistics(icebergStatistics, hiveStatistics, mergeStrategy, icebergTable.spec());
+            TableStatistics mergedStatistics = mergeHiveStatistics(icebergStatistics, hiveStatistics, mergeFlags, icebergTable.spec());
```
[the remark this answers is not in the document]
ZacBlanco: Good point. Fixed. Thanks for bringing it up

presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedTestBase.java
ZacBlanco: Added at https://github.com/prestodb/presto/blob/c449f368e438d518e0df21048244b20cd15d8cba/presto-iceberg/src/test/java/com/facebook/presto/iceberg/hive/TestIcebergHiveStatistics.java#L284-L300

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergHiveMetadata.java
```diff
@@ -481,9 +487,9 @@ public TableStatistics getTableStatistics(ConnectorSession session, ConnectorTab
                             IcebergColumnHandle::getName,
                             IcebergColumnHandle::getType)));
         }).orElseGet(() -> {
-            if (!mergeStrategy.equals(HiveStatisticsMergeStrategy.NONE)) {
-                PartitionStatistics hiveStats = metastore.getTableStatistics(getMetastoreContext(session), handle.getSchemaName(), handle.getIcebergTableName().getTableName());
-                return mergeHiveStatistics(icebergStatistics, hiveStats, mergeStrategy, icebergTable.spec());
+            PartitionStatistics hiveStats = metastore.getTableStatistics(getMetastoreContext(session), handle.getSchemaName(), handle.getIcebergTableName().getTableName());
+            if (!mergeFlags.isEmpty()) {
+                return mergeHiveStatistics(icebergStatistics, hiveStats, mergeFlags, icebergTable.spec());
```
hantangwangd, line 492: Should we get hive stats inside the `if` clause?

hantangwangd (commented):
LGTM! Only one little thing. Thanks for the important improvement.

@tdcmeehan Would you like to take a final look?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergHiveMetadata.java
ZacBlanco: Yup. Thanks for catching it

hantangwangd (approved):

steveburnett (dismissed):
LGTM! (docs) 

Pull updated branch, new local build.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergHiveMetadata.java
```diff
@@ -500,9 +506,16 @@ public TableStatisticsMetadata getStatisticsCollectionMetadata(ConnectorSession
...
-                .flatMap(meta -> metastore.getSupportedColumnStatistics(getMetastoreContext(session), meta.getType())
-                        .stream()
-                        .map(statType -> statType.getColumnStatisticMetadata(meta.getName())))
+                .flatMap(meta -> {
+                    try {
+                        return metastore.getSupportedColumnStatistics(getMetastoreContext(session), meta.getType())
+                                .stream()
+                                .map(statType -> statType.getColumnStatisticMetadata(meta.getName()));
+                    }
+                    catch (IllegalArgumentException e) {
```
aaneja, line 515: Under what conditions is this exception thrown ?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/TableStatisticsMaker.java
```diff
@@ -459,6 +522,10 @@ public static List<ColumnStatisticMetadata> getSupportedColumnStatistics(String
             supportedStatistics.add(NUMBER_OF_DISTINCT_VALUES.getColumnStatisticMetadataWithCustomFunction(columnName, "sketch_theta"));
         }
 
+        if (!(type instanceof FixedWidthType)) {
```
aaneja, line 525: Nice, I think this is clearer than before now about what types we track the size for

aaneja (commented):
Still reviewing tests, but changes LGTM overall

presto-iceberg/src/main/java/com/facebook/presto/iceberg/util/StatisticsUtil.java
```diff
@@ -57,29 +67,33 @@ public static TableStatistics mergeHiveStatistics(TableStatistics icebergStatist
...
-                            }
-                        }
-                        else {
-                            nullsFraction = Estimate.of((double) nullCount / hiveStatistics.getBasicStatistics().getRowCount().getAsLong());
-                        }
-                        mergedStats.setNullsFraction(nullsFraction);
-                    });
+                // data size
+                if (mergeFlags.contains(ColumnStatisticType.TOTAL_SIZE_IN_BYTES)) {
+                    hiveColumnStats.getTotalSizeInBytes().ifPresent(size -> mergedStats.setDataSize(Estimate.of(size)));
```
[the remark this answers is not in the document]
ZacBlanco: Yes, the Iceberg counts are always going to be more accurate than the Hive ones If the manifests are correct

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergHiveMetadata.java
ZacBlanco: In `ThriftHiveMetastore#getSupportedColumnStatistics`: https://github.com/prestodb/presto/blob/244d203952c4a283891ac6c817a8f4a732aa89c4/presto-hive-metastore/src/main/java/com/facebook/presto/hive/metastore/thrift/ThriftHiveMetastore.java#L416

Since the set of column statistics in Iceberg could differ from Hive, I didn't want to fail the operation if one of the types isn't supported by Hive.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergHiveMetadata.java
```diff
@@ -469,6 +474,7 @@ public TableStatistics getTableStatistics(ConnectorSession session, ConnectorTab
                     }
                 }
             }
+            RowExpression translatedPredicate = rowExpressionService.getDomainTranslator().toPredicate(predicate);
```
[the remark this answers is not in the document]
ZacBlanco: moved back to it's original location. Probably just got moved around while refactoring

presto-iceberg/src/main/java/com/facebook/presto/iceberg/util/StatisticsUtil.java
```diff
@@ -16,27 +16,37 @@
...
     {
     }
 
-    public static TableStatistics mergeHiveStatistics(TableStatistics icebergStatistics, PartitionStatistics hiveStatistics, HiveStatisticsMergeStrategy mergeStrategy, PartitionSpec spec)
+    /**
+     * Attempts to merge statistics from Iceberg and hive tables.
+     * <br>
+     * If a statistic is unknown on the Iceberg table, but known in Hive, the Hive statistic
+     * will always be used. Otherwise, hive statistics are only used if specified in the
+     * hive-statistic-merge-strategy
```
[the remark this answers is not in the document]
ZacBlanco: Thanks for pointing this out. I updated the comments in this file to ensure they reflect the current implementation

ZacBlanco:
@aaneja @hantangwangd FYI while testing some of this code again I also noticed that the total table size result was incorrect so I also added a function to calculate it properly:

https://github.com/prestodb/presto/blob/ba65e5156474643ba56798161cd55cf0a6e13f79/presto-iceberg/src/main/java/com/facebook/presto/iceberg/util/StatisticsUtil.java#L101-L115

presto-iceberg/src/main/java/com/facebook/presto/iceberg/util/StatisticsUtil.java
```diff
@@ -57,29 +68,49 @@ public static TableStatistics mergeHiveStatistics(TableStatistics icebergStatist
...
+            ColumnStatistics stats = entry.getValue();
+            return stats.getDataSize().or(() -> {
+                if (columnHandle.getType() instanceof FixedWidthType) {
+                    return stats.getNullsFraction().map(nulls -> rowCount * (1 - nulls) * ((FixedWidthType) columnHandle.getType()).getFixedSize());
+                }
+                else {
+                    return Estimate.unknown();
+                }
+            });
+        }).reduce(Estimate.of(0.0), (currentSize, newSize) -> currentSize.flatMap(current -> newSize.map(add -> current + add)))));
```
hantangwangd, line 114: Should we filter out metadata columns here or some place before here? Or we would always get an unknown estimate for the table's total size, as data size statistics for metadata columns of `VariableWidthType` would be unknown.

In addition, should we set total size to the row for whole table statistics in `ShowStatsRewrite.createTableStatsRow(..)`, so that we can read and get it through `show stats for...`?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/util/StatisticsUtil.java
```diff
@@ -13,36 +13,47 @@
...
+    /**
+     * Attempts to merge statistics from Iceberg and Hive tables.
+     * <br>
+     * Statistics from Hive are only merged if the corresponding flag exists in
+     * the {@code mergeFlags}.
+     */
+    public static TableStatistics mergeHiveStatistics(TableStatistics icebergStatistics, PartitionStatistics hiveStatistics, EnumSet<ColumnStatisticType> mergeFlags, PartitionSpec spec)
     {
-        if (mergeStrategy.equals(NONE) || spec.isPartitioned()) {
+        if (spec.isPartitioned()) {
```
hantangwangd, line 49: Seems currently we could not get valid values of `data_size` and `distinct_values_count` for a partitioned iceberg table configured with hive. Because we do not write and use iceberg statistics when configured with hive, and do not use hive statistics when iceberg table is partitioned.

Maybe we should write and use iceberg statistics when table is partitioned?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/util/StatisticsUtil.java
ZacBlanco: > Should we filter out metadata columns here or some place before here?

The calculation here only uses column statistics if they are added to the `TableStatistics.Builder`. As far as I know, the metadata columns (`SYNTHESIZED` from `BaseHiveColumnHandle`) are always filtered out before being added to the `TableStatistics.Builder`.

If data size of a `VariableWidthType` is unknown and it has column statistics added to `TableStatistics.Builder`, then it should follow that the total size of the table is unknown.


> In addition, should we set total size to the row for whole table statistics in ShowStatsRewrite.createTableStatsRow(..), so that we can read and get it through show stats for...?

I think this can be left to a separate PR? It would apply to all connectors and I feel is outside the scope of this PR since it's not Iceberg related. I think it should be added though. I will create a follow-up PR later.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/util/StatisticsUtil.java
ZacBlanco: I have another draft on my local machine which adds support for table statistics in some cases of partitioned tables. I will try to get it submitted early next week, but I think changing that behavior is out of scope for this PR.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/util/StatisticsUtil.java
hantangwangd: > The calculation here only uses column statistics if they are added to the `TableStatistics.Builder`. As far as I know, the metadata columns (`SYNTHESIZED` from `BaseHiveColumnHandle`) are always filtered out before being added to the `TableStatistics.Builder`.

Can you make sure of this? I think I just met this problem by simply executing a `show stats for <table_name>`. I see the metadata columns are passed in as `selectedColumns` for `TableStatisticsMaker.getTableStatistics()`.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/util/StatisticsUtil.java
hantangwangd: Sure, looking forward to see it.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/util/StatisticsUtil.java
ZacBlanco: > > The calculation here only uses column statistics if they are added to the `TableStatistics.Builder`. As far as I know, the metadata columns (`SYNTHESIZED` from `BaseHiveColumnHandle`) are always filtered out before being added to the `TableStatistics.Builder`.
> 
> Can you make sure of this? I think I just met this problem by simply executing a `show stats for <table_name>`. I see the metadata columns are passed in as `selectedColumns` for `TableStatisticsMaker.getTableStatistics()`.

Ah you're right. It seems `SHOW STATS` passes the hidden columns. However, when running queries, unless the query uses that column it won't be passed. I think for the purpose of `SHOW STATS` that we can omit passing hidden columns to the `selectedColumns` list when running `getTableStatistics`, since they are filtered out later anyways.

hantangwangd (approved):
LGTM, thanks for the fix!

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergHiveMetadata.java
```diff
@@ -500,9 +501,17 @@ public TableStatisticsMetadata getStatisticsCollectionMetadata(ConnectorSession
...
-                        .map(statType -> statType.getColumnStatisticMetadata(meta.getName())))
+                .flatMap(meta -> {
+                    try {
+                        return metastore.getSupportedColumnStatistics(getMetastoreContext(session), meta.getType())
+                                .stream()
+                                .map(statType -> statType.getColumnStatisticMetadata(meta.getName()));
+                    }
+                    // thrown in the case the type isn't supported by HMS statistics
+                    catch (IllegalArgumentException e) {
+                        return ImmutableSet.<ColumnStatisticMetadata>of().stream();
```
aaneja, line 512: Just return a [`Stream.empty()`](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#empty--) ?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/util/StatisticsUtil.java
```diff
@@ -57,29 +68,49 @@ public static TableStatistics mergeHiveStatistics(TableStatistics icebergStatist
...
+                .map(EnumSet::copyOf)
+                .orElse(EnumSet.noneOf(ColumnStatisticType.class));
+    }
+
+    public static String encodeMergeFlags(EnumSet<ColumnStatisticType> flags)
+    {
+        return Joiner.on(",").join(flags.stream().map(Enum::name).iterator());
+    }
+
+    public static TableStatistics.Builder calculateAndSetTableSize(TableStatistics.Builder builder)
```
aaneja, line 101: Let's move this static utility to the TableStatistics.Builder class itself ?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/util/StatisticsUtil.java
```diff
@@ -57,29 +68,49 @@ public static TableStatistics mergeHiveStatistics(TableStatistics icebergStatist
...
         });
-        return statsBuilder.build();
+        return calculateAndSetTableSize(statsBuilder).build();
+    }
+
+    public static EnumSet<ColumnStatisticType> decodeMergeFlags(String input)
+    {
+        return Optional.of(Arrays.stream((input).trim().split(","))
+                        .filter(value -> !value.isEmpty())
+                        .map(ColumnStatisticType::valueOf)
```
aaneja, line 89: Let's also add a check here that the mapped ColumnStatisticType is one of {NUMBER_OF_DISTINCT_VALUES, TOTAL_SIZE_IN_BYTES} since that is all we support at the moment ?

aaneja (commented):
LGTM except a few small things

presto-iceberg/src/main/java/com/facebook/presto/iceberg/util/StatisticsUtil.java
ZacBlanco: I can't because it relies on casting to `IcebergColumnHandle` to check if the type is a `FixedWidthType`. The SPI for `ColumnHandle` has no methods which would allow this

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergHiveMetadata.java
ZacBlanco: fixed

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergSessionProperties.java
```diff
@@ -156,14 +161,14 @@ public IcebergSessionProperties(
                         false),
                 new PropertyMetadata<>(
                         HIVE_METASTORE_STATISTICS_MERGE_STRATEGY,
-                        "choose how to include statistics from the Hive Metastore when calculating table stats. Valid values are: "
-                                + Joiner.on(", ").join(HiveStatisticsMergeStrategy.values()),
+                        "Flags to choose which statistics from the Hive Metastore are used when calculating table stats. Valid values are: "
+                                + Joiner.on(", ").join(NUMBER_OF_DISTINCT_VALUES, TOTAL_SIZE_IN_BYTES),
```
aaneja, line 165: nitpick : Can we reference the `supportedMergeFlags` from StatisticsUtil  to build this csv list ?

aaneja (approved):

hantangwangd (dismissed):

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergSessionProperties.java
ZacBlanco: Fixed

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -225,10 +225,13 @@ Property Name                                           Description
 
 ``iceberg.enable-parquet-dereference-pushdown``         Enable parquet dereference pushdown.                          ``true``
 
-``iceberg.hive-statistics-merge-strategy``              Determines how to merge statistics that are stored in the     ``NONE``
-                                                        Hive Metastore. The available values are ``NONE``,
-                                                        ``USE_NULLS_FRACTION_AND_NDV``, ``USE_NULLS_FRACTIONS``
-                                                        and, ``USE_NDV``
+``iceberg.hive-statistics-merge-strategy``              Comma separated list of statistics to use from the
```
tdcmeehan, line 228: If the behavior of this flag is changing from release to release, I think we'd want to introduce a new flag for this behavior, and deprecate the old flag.

ZacBlanco: Is there a process for deprecation? I felt this was safe, because very few (if any) users consume these flags except for our internal benchmarks. I can add a release note for it.

Also, since Presto doesn't have a 1.X release, According to SemVer rules, this should be OK to change between releases.

tdcmeehan: @ZacBlanco that is definitely not the case.  Presto is over a decade old, and people rely on features not breaking release to release.

I think in this case though, it's fine, because it hasn't made it to our public documentation yet.  Please add a release note.

ZacBlanco: The release notes have been added to the PR description

tdcmeehan (approved):

PR #22328, opened 2024-03-25
[native] Fix cumulative memory usage calculation issue

xiaoxmeng:
We calculate sinceLastPeriodMs without initializing lastTaskStatsUpdateMs so it caused
very large cumulative memory usage report compared with Java. Also change the interval
from int64_t to double to be aligned with Java and it is used in a multiply operation.

### DIFF
--- a/presto-native-execution/presto_cpp/main/PrestoTask.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoTask.cpp
@@ -417,8 +417,6 @@ protocol::TaskInfo PrestoTask::updateInfoLocked() {
   prestoTaskStats.endTime = util::toISOTimestamp(taskStats.executionEndTimeMs);
 
   const uint64_t currentTimeMs = velox::getCurrentTimeMs();
-  const uint64_t sinceLastPeriodMs = currentTimeMs - lastTaskStatsUpdateMs;
-
   if (taskStats.executionEndTimeMs > taskStats.executionStartTimeMs) {
     prestoTaskStats.elapsedTimeInNanos =
         (taskStats.executionEndTimeMs - taskStats.executionStartTimeMs) *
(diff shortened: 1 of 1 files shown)


presto-native-execution/presto_cpp/main/PrestoTask.cpp
```diff
@@ -417,7 +417,7 @@ protocol::TaskInfo PrestoTask::updateInfoLocked() {
   prestoTaskStats.endTime = util::toISOTimestamp(taskStats.executionEndTimeMs);
 
   const uint64_t currentTimeMs = velox::getCurrentTimeMs();
-  const uint64_t sinceLastPeriodMs = currentTimeMs - lastTaskStatsUpdateMs;
+  const double sinceLastPeriodMs = currentTimeMs - lastTaskStatsUpdateMs;
```
mbasmanova, line 420: Would you explain what was happening before this change?

Wondering if we can add some tests to prevent this from breaking in the future.

xiaoxmeng: Prestissimo started to see a special value on cumulative memory usage report which was a double in Java but uint64_t in cpp. I suspect this is an overflow.

spershin: How can we overflow 'ms'?
This is not possible, I believe.
If we see large numbers - it is likely due to a bug.

xiaoxmeng: It is time interval * average memory usage

xiaoxmeng: Ok, this is a regression at Feb22 when we started to see bad stats.

presto-native-execution/presto_cpp/main/PrestoTask.cpp
```diff
@@ -437,14 +442,9 @@ protocol::TaskInfo PrestoTask::updateInfoLocked() {
...
-    lastTaskStatsUpdateMs = taskStats.executionStartTimeMs;
-  }
-
   const int64_t currentBytes = stats.currentBytes;
 
-  int64_t averageMemoryForLastPeriod =
+  const int64_t averageMemoryForLastPeriod =
       (currentBytes + lastMemoryReservation) / 2;
 
   prestoTaskStats.cumulativeUserMemory +=
```
spershin, line 450: Can we move 
```
const double sinceLastPeriodMs = currentTimeMs - lastTaskStatsUpdateMs;
```
close to the code that actually uses "sinceLastPeriodMs" (here)?
Otherwise it is detached and not clear why we need it, hence I was asking "why double". 🙂

spershin (approved):
Thanks!

presto-native-execution/presto_cpp/main/PrestoTask.cpp
```diff
@@ -437,16 +435,17 @@ protocol::TaskInfo PrestoTask::updateInfoLocked() {
...
   if (lastTaskStatsUpdateMs == 0) {
     lastTaskStatsUpdateMs = taskStats.executionStartTimeMs;
   }
 
-  const int64_t currentBytes = stats.currentBytes;
-
-  int64_t averageMemoryForLastPeriod =
-      (currentBytes + lastMemoryReservation) / 2;
-
+  const double sinceLastPeriodMs = currentTimeMs - lastTaskStatsUpdateMs;
```
mbasmanova, line 448: @xiaoxmeng This code is rather fragile. Lots of things happening in a single very long method. Let's consider refactoring.

PR #22333, opened 2024-03-26
Add presto worker type to hive metastore

kewang1024:
```
== NO RELEASE NOTE ==
```
For debuggability, add presto worker type to hive metastore to identify which engine wrote the files

### DIFF
--- a/presto-hive-hadoop2/src/test/java/com/facebook/presto/hive/s3select/S3SelectTestHelper.java
+++ b/presto-hive-hadoop2/src/test/java/com/facebook/presto/hive/s3select/S3SelectTestHelper.java
@@ -170,7 +170,7 @@ public class S3SelectTestHelper
                 new HiveTypeTranslator(),
                 new HiveStagingFileCommitter(hdfsEnvironment, listeningDecorator(executor)),
                 new HiveZeroRowFileCreator(hdfsEnvironment, new OutputStreamDataSinkFactory(), listeningDecorator(executor)),
-                new NodeVersion("test_version"),
+                new NodeVersion("test_version", NodeVersion.PrestoWorkerType.JAVA),
                 new HivePartitionObjectBuilder(),
                 new HiveEncryptionInformationProvider(ImmutableSet.of()),
                 new HivePartitionStats(),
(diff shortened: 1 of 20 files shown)


tdcmeehan:
Instead of adding this concept in the code, can you add it to the `node.version` in your deployment?  CPP/Java is just one of many things that may affect the output, and I'd avoid adding this to unrelated connectors and dependencies of Hive.

tdcmeehan (changes requested):

kewang1024:
Another way to add: https://github.com/prestodb/presto/pull/22370

kewang1024:
Discussed offline with @tdcmeehan who thinks this is not needed yet for open source, we can revisit this when needed.

Only make changes meta internally: https://github.com/facebookexternal/presto-facebook/pull/2693

mbasmanova:
> Discussed offline with @tdcmeehan who thinks this is not needed yet for open source, we can revisit this when needed.

@kewang1024 @tdcmeehan Ke, Tim, would you share the summary of this discussion?

tdcmeehan:
@mbasmanova my main concern is persisting names like "CPP" or "Java".  While Prestissimo may be considered to be beta (and even production worthy if you have the resources to support it), many things have yet to be working properly, and I worry about persisting things like "CPP"/"Java" until we settle on the eventual abstractions that will ultimately differentiate older Java eval clusters from Prestissimo clusters.  In simple terms, when we think about names, I don't know if "CPP-Presto" is the name we want to start persisting in the metastore, ORC footers, Parquet files, etc.  Until we settle on all the work items that make Prestissimo the default, I would rather wait to persist this information--for example, what if Presto with Prestissimo workers becomes like a 2.0 release, I would want to wait until we know what that will be called and how the connector SPI will change until we start persisting the information that differentiates Java from CPP (as a hypothetical example, perhaps we instead persist the system and version, and as part of our move to C++ we make a 1.0 or 2.0 release signifying a major change in the system, and we instead persist something like Presto1.x or Presto2.x).  I just don't want to persist this until we have a clearer picture of how we will get to the C++-worker future in deployments outside of big tech companies, because I anticipate it changing until we get there.  In the meantime, if it's important to the deployment of Presto, one is free to alter their version to add this information.

mbasmanova:
@tdcmeehan Tim, thank you for explaining.

> I would want to wait until we know what that will be called

I was under impression that we settled on Prestissimo name. If that's not the case, curious if you have any timeline in mind for making the decision.

tdcmeehan:
@mbasmanova I considered it a project name, but when we're done, we're not going to update the website and docs etc. In terms of the timeline of when we announce that this is the new Presto (and how that's versioned and documented and how the APIs change), I would like to set that milestone and put out a rough timeline in the near future as it doesn't exist right now.

PR #22341, opened 2024-03-26
Add a dev page for viewing query plan

yhwang:
## Description
Add a dev page under the URI: `/ui/dev/index.html`
to parse a JSON file of SQL query and plot the query plan.

## Motivation and Context
It would be handy to view the query plan of an old query which you saved its JSON file locally.

## Impact
Create a new page and it can be accessed via a hidden URI: `/ui/dev/index.html`. It's not seeable from the Presto UI only if you
know the URL.

## Test Plan
Manually verified the new page.

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```


yhwang:
Its usage is:
- Go to the Presto UI and use the secret path, i.e: http://localhost:8080/ui/dev/index.html
  Screenshot 2024-03-26 at 4 56 04 PM
  you will see the page like the screenshot above
- Click on the `Browse` button and select a JSON file that is from an old SQL query you had earlier. After you select the file. Its plan will show up in the canvas below.
  Screenshot 2024-03-26 at 4 56 35 PM

yhwang:
cc @aditi-pandit @tdcmeehan 

One thing worth mentioning is that I wanted to use a path like: `http://localhost:8080/ui/dev/`. But somehow Presto's HTTP server can't render the `index.html` under the `dev` directory. Therefore, users have to explicitly specify the `index.html` in the path.

tdcmeehan:
@elharo most of the `.js` code is autogenerated from the `.jsx` code.  @yhwang will be adding some docs around this to make it clear.

aaneja:
FYI, I have a standalone web app - https://github.com/aaneja/presto-standalone-webapp, that can display the query landing page, query plan, and other tabs using the info JSON of a previously executed query 

It uses the same assets as the existing web app and functions because the webapp is mostly client side except the calls to `/v1/query`

We could build on this idea and have the `/dev` route serve these (pseudo) static assets

tdcmeehan:
@aaneja is the proposal to broaden the scope of these changes to modify the Presto UI to serve all tabs of the existing query lookup page?  If so, I think that's a good idea, but I'd recommend that be a followup, since the most inscrutable information from `QueryInfo` seems to be the query plan.

yhwang:
> Where does all this opaque JavaScript come from? I'm not ure I could fix a bug in this if I needed to. Is there a more natural form?

@elharo like @tdcmeehan  mentioned, those `.js` files under the `dist` directory are generated from those `.jsx` files under the `src` directory. Those `.js` files changed because I updated the `PageTitle.jsx` which is used by most of the pages. FYI, I do have a TODO item to make those `.js` files being generated on build time [here](https://github.com/prestodb/presto/issues/21062) (item 6)


> @yhwang try to add this line to line 136 in `CoordinatorModule.java` to make it so `/dev` is routed to `index.html`:
> 
> `httpServerBinder(binder).bindResource("/dev", "webapp/dev").withWelcomeFile("index.html");`

Thanks @tdcmeehan let me add that

tdcmeehan:
@aaneja got it.  My thought was people who will be using this are already using a query runner, but I believe @yhwang has a technique that avoids that and just loads a lightweight web server.

yhwang:
There are two options to implement this:
- Use the webpack dev server to run a dev version of Presto UI and proxy the `/v1` API calls to a running Presto cluster
- Add dev page into the Presto UI directly

For the first one, it can be done by tweaking the webpack dev server config and we already have a proxy setting for that. You can try that out by running this command:
```
yarn --cwd presto-main/src/main/resources/webapp/src serve --env=apiHost=<the running presto coordinator's IP> --env=apiPort=<port number>
```
But the hurdle for this is that you need to spin up another web server alongside the Presto cluster

For the second one, this PR, it runs with the Presto Server as another static HTML page. It loads the JSON from your local filesystem which means it only consumes your local resources since the page loads the JSON file and displays the query plan locally without any network traffic to the Presto server. If you have an up-and-running Presto cluster, you can leverage this page too.

Therefore, I believe the second has more benefits than the first one. That's why I have this implementation. (Actually, I had the implementation for the first one too)

yhwang:
@tdcmeehan I tried out your suggestion about the `index.html` page. However, based on our current file structure under `webapp` directory, that will expose `dev` in two paths:
- /dev/
- /ui/dev/

I guess we need to restructure the folders under the `webapp` to avoid this. I will add this as another working item for UI. I won't touch that in this PR, is that okay for you?

jaystarshot:
Apologies for the ignorance, but what do you mean by **JSON file** of a query plan? Is it just sql?

tdcmeehan:
> @tdcmeehan I tried out your suggestion about the `index.html` page. However, based on our current file structure under `webapp` directory, that will expose `dev` in two paths:
> 
> * /dev/
> * /ui/dev/
> 
> I guess we need to restructure the folders under the `webapp` to avoid this. I will add this as another working item for UI. I won't touch that in this PR, is that okay for you? So the full path for now will be `/ui/dev/index.html`. Let me also update the description.

Sounds good

yhwang:
@jaystarshot the latter one:
> Or is this the json returned in `8080/v1/query/<<QueryId>>?pretty`

Sometimes you may save the JSON output and the server gets restarted. In this case, you need a tool to check out the query plan, otherwise, you need to traverse the JSON manually.

yhwang:
@tdcmeehan is this what you want to see? The plan viewer works exactly the same as the existing `Live Plan` viewer. I directly re-use the code from there. But I disabled the stage links which open the stage page. I guess the main purpose of this PR is to provide a plan viewer. BTW, I didn't edit the clip. My JSON file is not complicated, so it renders the plan quickly.

https://github.com/prestodb/presto/assets/1147320/05ad9ed1-7690-47e5-b25d-d951bede7163

tdcmeehan (dismissed):
LGTM

tdcmeehan:
@yhwang : @steveburnett recently added https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/clients/presto-console.rst -- what do you think about adding documentation noting this secret URL and noting that it's for developers?  That way we don't introduce more tribal knowledge about this URL.

yhwang:
@tdcmeehan 
> what do you think about adding documentation noting this secret URL and noting that it's for developers? That way we don't introduce more tribal knowledge about this URL.

That's a good idea. One thing that will change is the URL if the folders under the `webapp` directory are restructured. I can point to `http://localhost:8080/ui/dev/index.html` for now.

yhwang:
Have to do a rebase to get @steveburnett 's change and update the `presto-console.rst` to include a `Plan Viewer` section per @tdcmeehan suggestion.

The new section looks like this:
Screenshot 2024-03-29 at 12 16 35 PM

presto-docs/src/main/sphinx/clients/presto-console.rst
```diff
@@ -35,3 +35,23 @@ After starting Presto, you can access the web UI at the default port
 
 .. figure:: ../images/presto_console.png
    :align: center
+
+Plan Viewer
+=======================
+
+Plan Viewer is a tool to generate the graph of a query plan on the Presto Console.
```
steveburnett, line 42:
```suggestion
Plan Viewer generates the graph of a query plan on the Presto Console.
```

presto-docs/src/main/sphinx/clients/presto-console.rst
```diff
@@ -35,3 +35,23 @@ After starting Presto, you can access the web UI at the default port
...
+You can access it using this path: ``/ui/dev/index.html`` along with your server IP
+and port number. For example: ``http://localhost:8080/ui/dev/index.html``. When a
```
steveburnett, line 44:
```suggestion
and port number. For example: 

.. code-block:: none
    http://localhost:8080/ui/dev/index.html

```
Suggest changing the formatting of the URI to a code block, like the one earlier on the page, so that a user can copy to Clipboard easier.

presto-docs/src/main/sphinx/clients/presto-console.rst
```diff
@@ -35,3 +35,23 @@ After starting Presto, you can access the web UI at the default port
 
 .. figure:: ../images/presto_console.png
    :align: center
+
+Plan Viewer
+=======================
+
+Plan Viewer is a tool to generate the graph of a query plan on the Presto Console.
+You can access it using this path: ``/ui/dev/index.html`` along with your server IP
```
steveburnett, line 43:
```suggestion
To access the Plan Viewer using a web browser, add the path ``/ui/dev/index.html`` to your server IP
```

presto-docs/src/main/sphinx/clients/presto-console.rst
```diff
@@ -35,3 +35,23 @@ After starting Presto, you can access the web UI at the default port
...
+and port number. For example: ``http://localhost:8080/ui/dev/index.html``. When a
+Presto server restarts, all the query information is reset. However, the
```
steveburnett, line 45:
```suggestion
Use the ``JSON`` tab on the ``Query Details`` page to export and save a query in JSON format. 
```

presto-docs/src/main/sphinx/clients/presto-console.rst
```diff
@@ -35,3 +35,23 @@ After starting Presto, you can access the web UI at the default port
...
+information of a query can be exported in JSON format using the ``JSON`` tab on
```
steveburnett, line 46:
```suggestion
```

presto-docs/src/main/sphinx/clients/presto-console.rst
```diff
@@ -35,3 +35,23 @@ After starting Presto, you can access the web UI at the default port
...
+the Query Details page and saved locally. Here is where the ``JSON`` tab is located.
```
steveburnett, line 47:
```suggestion
```

presto-docs/src/main/sphinx/clients/presto-console.rst
```diff
@@ -35,3 +35,23 @@ After starting Presto, you can access the web UI at the default port
...
+
+.. figure:: ../images/query_details_json.png
+   :align: center
+
```
steveburnett, line 51:
```suggestion

*Note:* When a Presto server restarts, all the query information is reset. 
```
Suggest moving this to here, after you explain how to export the current query.

presto-docs/src/main/sphinx/clients/presto-console.rst
```diff
@@ -35,3 +35,23 @@ After starting Presto, you can access the web UI at the default port
...
+With the JSON files, you can use the Plan Viewer to view the graph of query plans.
+
+1. Access the Plan Viewer page. For example: ``http://localhost:8080/ui/dev/index.html``
+2. Click on the ``Browse`` button and select the JSON file you want to examine.
```
steveburnett, line 55:
```suggestion
2. Select the ``Browse`` button and select the JSON file you want to examine. The query plan graph is displayed. 
```
See the [Gitlab documentation style guide recommended word list entry for click](https://docs.gitlab.com/ee/development/documentation/styleguide/word_list.html#click) for discussion and suggestions. 

Moved "The query plan graph is displayed." to here, as it's the result or output of the user's action in step 2.

presto-docs/src/main/sphinx/clients/presto-console.rst
```diff
@@ -35,3 +35,23 @@ After starting Presto, you can access the web UI at the default port
...
+3. The query plan graph will show up.
```
steveburnett, line 56:
```suggestion
```

presto-docs/src/main/sphinx/clients/presto-console.rst
```diff
@@ -35,3 +35,23 @@ After starting Presto, you can access the web UI at the default port
...
+4. Click on the ``Browse`` button to view other JSON file if needed.
```
steveburnett, line 57:
```suggestion
4. Select the ``Browse`` button to view other JSON files if needed.
```

steveburnett (changes requested):
Thanks for the quick documentation update! Great content and instructions, including the screenshot. 

I made some suggestions for conciseness, readability, and flow, and a couple of style suggestions. Let me know what you think!

presto-docs/src/main/sphinx/clients/presto-console.rst
yhwang: fixed

presto-docs/src/main/sphinx/clients/presto-console.rst
yhwang: lesson learned, thanks for the info

yhwang:
@steveburnett thanks for the comments. I believe I addressed them all. 
Here is the updated doc:
Screenshot 2024-03-29 at 1 25 51 PM

steveburnett (approved):
LGTM! (docs)

Pull updated branch, new local build of docs, everything looks good. Thanks!

tdcmeehan (approved):

aaneja:
>There are two options to implement this:
> - Use the webpack dev server to run a dev version of Presto UI and proxy the /v1 API calls to a running Presto cluster
> - Add dev page into the Presto UI directly

@yhwang IMO, neither of these is a great option for a user (not a Presto dev) to solely examine a JSON offline since they need to bring up a Presto server (with at least a single catalog) online. What I had in mind was having a JettyServer that just serves pre-canned user defined JSONs. That way users could use existing presto jar(s) from a release and only bring this up instead of the full presto-server

I think what we have right now is a good starting point though !

yhwang:
@aaneja what you described is option 1. Just bring up the web dev server and go straight to http://localhost:8080/dev/index.html

jaystarshot:
Does this web dev server need to be user specific or can we create a new website which can allow people to view all plans?

jaystarshot:
Ahh yes i meany have one publicly available

yhwang:
I plan to create a single-page application for this. Then users can directly download, open the HTML file locally, and display the query details of a saved query. After that, we could figure out if that page could be hosted somewhere publicly.

PR #22343, opened 2024-03-27
[native] presto task info update code refactor and add cumulative memory test

xiaoxmeng:
Refactor Presto task info update code logic by splitting the update logic into a couple of methods
with one per each type info:
- updateOutputBufferInfoLocked: update output buffer related stats which is mostly used by scale
   writer by Presto coordinator.
- updateTimeInfoLocked: update task execution timing related stats such as task start time, end time etc.
- updateMemoryInfoLocked: update memory usage stats such as cumulative memory usage, peak memory
   usage etc.
- updateExecutionInfoLocked: update task internal execution stats such as operator and pipeline stats.

Add unit test to verify lastTaskStatsUpdateMs is initialized properly on the first call to get
task info. If not initialized, we might get very large cumulative memory usage if the task memory
usage is not zero.

### DIFF
--- a/presto-native-execution/presto_cpp/main/PrestoTask.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoTask.cpp
@@ -27,7 +27,7 @@ namespace facebook::presto {
 
 namespace {
 
-#define TASK_STATS_SUM(taskStats, taskStatusSum, statsName)      \
+#define TASK_STATS_SUM(taskStats, statsName, taskStatusSum)      \
   do {                                                           \
     for (int i = 0; i < taskStats.pipelineStats.size(); ++i) {   \
       auto& pipeline = taskStats.pipelineStats[i];               \
@@ -49,6 +49,7 @@ protocol::TaskState toPrestoTaskState(exec::TaskState state) {
     case exec::kFailed:
       return protocol::TaskState::FAILED;
     case exec::kAborted:
+      [[fallthrough]];
     default:
       return protocol::TaskState::ABORTED;
   }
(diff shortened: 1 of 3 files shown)


presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp
```diff
@@ -1253,42 +1255,94 @@ TEST_F(TaskManagerTest, getResultsFromFailedTask) {
   ASSERT_EQ(results->data->capacity(), 0);
 }
 
-TEST_F(TaskManagerTest, testCumulativeMemory) {
-  auto filePaths = makeFilePaths(10);
-  auto vectors = makeVectors(10, 1'000);
-  for (int i = 0; i < 10; ++i) {
+DEBUG_ONLY_TEST_F(TaskManagerTest, testCumulativeMemory) {
```
mbasmanova, line 1258: It took me quite some time to figure out what was happening and how #22328 is fixing the problem. It feels like the code in PrestoTask is rather fragile and can break easily. This test feels quite heavy weight. It uses TestValue, which makes it debug-only, and performs a lot of moves that are hard to understand. Can we do something to simplify the code to make it less error prone and more easily testable?

https://github.com/prestodb/presto/pull/22328/files#r1540631632

presto-native-execution/presto_cpp/main/PrestoTask.cpp
```diff
@@ -209,28 +210,29 @@ void addSpillingOperatorMetrics(OperatorStatsCollector& collector) {
 
 // Process the runtime stats of individual protocol::OperatorStats. Do not add
```
mbasmanova, line 211: I wonder what does "process" mean here.

presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp
xiaoxmeng: Simplify the test per our offline discussion. Thanks!

presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp
```diff
@@ -24,6 +24,7 @@
 #include "velox/common/base/Fs.h"
 #include "velox/common/base/tests/GTestUtils.h"
 #include "velox/common/file/FileSystems.h"
+#include "velox/common/testutil/TestValue.h"
```
mbasmanova, line 27: not needed anymore?

presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp
```diff
@@ -48,6 +49,7 @@ static const std::string kHiveConnectorId = "test-hive";
 using namespace facebook::velox;
 using namespace facebook::velox::exec;
 using namespace facebook::velox::exec::test;
+using namespace facebook::velox::common::testutil;
```
mbasmanova, line 52: not needed anymore?

mbasmanova (approved):
Thanks.

presto-native-execution/presto_cpp/main/PrestoTask.cpp
xiaoxmeng: Updated the code comment. Thanks!

mbasmanova (approved):

PR #15716, opened 2021-02-11
Allow session property rules to override

aweisberg:
This is to allow rules setting resource limits to override values set as session properties. For the sneaky "I want to bypass resource limits case"

Test plan - There are included unit tests

depends on https://github.com/facebookexternal/presto-facebook/pull/1413

```
== RELEASE NOTES ==

General Changes
* Add support for overriding session properties using session property managers :doc:`/admin/session-property-managers`. Setting 'overrideSessionProperties` to true will cause the property to be overridden and remain overridden even if subsequent rules match the property but don't have `overrideSessionProperties` set.
```

PR #22345, opened 2024-03-27
[native] Disable exchange_materialization_strategy

mbasmanova:
exchange_materialization_strategy = 'ALL' is used to execute the query in Large Batch Mode (LBM). This mode is deprecated and is not supported by Prestissimo. This change is to ignore this session property when using Prestissimo.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
+++ b/presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
@@ -2025,8 +2025,11 @@ public final class SystemSessionProperties
         return !isStrategyNone && hasMaterializedCTE;
     }
 
-    public static ExchangeMaterializationStrategy getExchangeMaterializationStrategy(Session session)
+    public static ExchangeMaterializationStrategy getExchangeMaterializationStrategy(Session session, boolean nativeExecution)
     {
+        if (nativeExecution) {
+            return ExchangeMaterializationStrategy.NONE;
+        }
         return session.getSystemProperty(EXCHANGE_MATERIALIZATION_STRATEGY, ExchangeMaterializationStrategy.class);
     }
 
--- a/presto-main/src/main/java/com/facebook/presto/sql/planner/PlanFragmenter.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/planner/PlanFragmenter.java
@@ -52,6 +52,7 @@ public class PlanFragmenter
     private final SqlParser sqlParser;
     private final PlanChecker distributedPlanChecker;
     private final PlanChecker singleNodePlanChecker;
+    private final boolean nativeExecution;
 
     @Inject
     public PlanFragmenter(Metadata metadata, NodePartitioningManager nodePartitioningManager, QueryManagerConfig queryManagerConfig, SqlParser sqlParser, FeaturesConfig featuresConfig)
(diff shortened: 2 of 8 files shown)


arhimondr (approved):

tdcmeehan:
Hey folks--I understand we've deprecated LBM, however, I wonder if it's wise to override session properties in this way.

If the goal is to assist migration, we do have a way to override session properties (even ones users have supplied) with a session property override: #15716

I'm wondering if we should instead have the code validate that in native, this should never happen (i.e. it should throw an error).  That reflects the reality, which is that Prestissimo is not, and will not, be designed to support LBM.  We could do this for other properties that should also throw, such as `optimize_hash_generation`.  In the future, it would be great to make it so that Prestissimo clusters don't even allow these properties as an option (i.e. they don't show up in `SHOW SESSION`, and `SET SESSION` will fail).

To support migration, we can use a session property override and just set it to override to `NONE`.  That is to support Meta's cluster-by-cluster migration effort.

Thoughts?

amitkdutta:
@tdcmeehan Thanks Tim for all the comments.

> 
In the future, it would be great to make it so that Prestissimo clusters don't even allow these properties as an option (i.e. they don't show up in SHOW SESSION, and SET SESSION will fail).

This would be perfect and ideally the queries should fail with `session property not supported` error for Prestissimo clusters (like regular cases where unsupported session property causes query to fail). Also `show session` will not show it and `set session` will fail.

Also I think worker should throw and I think it does today (I need to double check)

Now regarding the config override. We are actually overriding with the session override you suggested. There are a few issues with override though:

- Override is for (as I understand) to replace a supported session property with a defined value. In this case, we have a few unsupported properties, overriding them to run the query seems very hacky workaround.
- As we move with Prestissimo, we have found/will find a handful of configs that are totally unsupported. We can definitely have config override to aid migration, but this will become internal to Meta what configs are not supported. It will be great to have it codified in coordinator (even not in the ideal state initially) so that everyone in the community will be aware with it.
- For migration or configuring cluster its also valuable to disable unsupported properties from the get go - this reduces config maintenance hassle.

tdcmeehan:
Thanks for your reply @amitkdutta.  I have a few follow up questions:

> Override is for (as I understand) to replace a supported session property with a defined value. In this case, we have a few unsupported properties, overriding them to run the query seems very hacky workaround.

I am a little confused.  Doesn't it seem like we're doing the same sort of override in this PR (because it's changing the value to `NONE`), except the override is not configurable, and it's happening without a Presto operators knowledge?  Wouldn't it be better to have the session property throw an explicit error to alert users that this session property simply does not work?

> As we move with Prestissimo, we have found/will find a handful of configs that are totally unsupported. We can definitely have config override to aid migration, but this will become internal to Meta what configs are not supported. It will be great to have it codified in coordinator (even not in the ideal state initially) so that everyone in the community will be aware with it.

I definitely agree with the intention--but again, wouldn't it be better to make such session properties throw, so when people initially try to use Prestissimo, they know that such session properties are not supported at all?  My worry is people continue to use such session properties, thinking they work.  Even more confusiong, when they `SET SESSION` trying to get it to work, the only way they'll discover it's not taking effect is if they `SHOW SESSION`.  Isn't an error a better way to signal lack of compatibility?

PR #22352, opened 2024-03-27
[native] Allow config values to have =

tanjialiang:
```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/common/ConfigReader.cpp
+++ b/presto-native-execution/presto_cpp/main/common/ConfigReader.cpp
@@ -13,12 +13,12 @@
  */
 
 #include "presto_cpp/main/common/ConfigReader.h"
+#include <fmt/format.h>
 #include <fstream>
 #include "velox/common/base/Exceptions.h"
 #include "velox/core/Config.h"
 
 namespace facebook::presto::util {
-
 std::unordered_map<std::string, std::string> readConfig(
     const std::string& filePath) {
   // https://teradata.github.io/presto/docs/141t/configuration/configuration.html
@@ -36,10 +36,16 @@ std::unordered_map<std::string, std::string> readConfig(
       continue;
     }
 
-    std::vector<std::string> configParts;
-    folly::split('=', line, configParts);
-    VELOX_USER_CHECK_EQ(configParts.size(), 2, "Malformed config: {}", line);
-    properties.emplace(configParts[0], configParts[1]);
+    const auto delimiterPos = line.find('=');
+    VELOX_CHECK_NE(
+        delimiterPos,
+        std::string::npos,
+        "No '=' sign found for property pair '{}'",
+        line);
+    const auto name = line.substr(0, delimiterPos);
+    VELOX_CHECK(!name.empty(), "property pair '{}' has empty key", line);
+    const auto value = line.substr(delimiterPos + 1);
+    properties.emplace(name, value);
   }
 
   return properties;
--- a/presto-native-execution/presto_cpp/main/common/tests/ConfigTest.cpp
+++ b/presto-native-execution/presto_cpp/main/common/tests/ConfigTest.cpp
@@ -12,8 +12,10 @@
  * limitations under the License.
  */
 #include <gtest/gtest.h>
+#include "presto_cpp/main/common/ConfigReader.h"
 #include "presto_cpp/main/common/Configs.h"
 #include "velox/common/base/Exceptions.h"
+#include "velox/common/base/tests/GTestUtils.h"
 #include "velox/common/file/File.h"
 #include "velox/common/file/FileSystems.h"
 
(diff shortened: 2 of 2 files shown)


mbasmanova (commented):
@tanjialiang Please, add a test to make sure this doesn't break in the future.

presto-native-execution/presto_cpp/main/common/ConfigReader.cpp
```diff
@@ -36,10 +35,10 @@ std::unordered_map<std::string, std::string> readConfig(
       continue;
     }
 
-    std::vector<std::string> configParts;
-    folly::split('=', line, configParts);
-    VELOX_USER_CHECK_EQ(configParts.size(), 2, "Malformed config: {}", line);
-    properties.emplace(configParts[0], configParts[1]);
+    auto delimiterPos = line.find('=');
```
mbasmanova, line 38: Should we check that delimiterPos != std::string::npos? Perhaps, add a test for invalid config.

presto-native-execution/presto_cpp/main/common/tests/ConfigTest.cpp
```diff
@@ -191,4 +203,42 @@ TEST_F(ConfigTest, remoteFunctionServer) {
...
+      {"key", "value"},
+      {"emptyvaluekey", ""}};
+  ASSERT_EQ(configMap, expected);
+}
+
+TEST_F(ConfigTest, parseInvalid) {
+  auto testInvalid = [this](const std::string& fileContent) {
+    setUpConfigFilePath();
+    writeConfigFile(fileContent);
+    ASSERT_THROW(presto::util::readConfig(configFilePath), VeloxException);
```
mbasmanova, line 238: VELOX_ASSERT_THROW

presto-native-execution/presto_cpp/main/common/ConfigReader.cpp
```diff
@@ -36,10 +35,12 @@ std::unordered_map<std::string, std::string> readConfig(
...
+    VELOX_CHECK_NE(delimiterPos, std::string::npos);
```
mbasmanova, line 39: Let's add an error message with enough context to help troubleshooting.

presto-native-execution/presto_cpp/main/common/ConfigReader.cpp
```diff
@@ -36,10 +35,12 @@ std::unordered_map<std::string, std::string> readConfig(
...
+    auto name = line.substr(0, delimiterPos);
+    VELOX_CHECK(!name.empty());
```
mbasmanova, line 41: Should we trim the 'name'?

tanjialiang: it's trimmed at the beginning of the loop

mbasmanova: Hmm... Let's say we have "   xxx = yyy  ". At the start of the loop we trim this to "xxx = yyy", then we get name="xxx " with space at the end, no?

presto-native-execution/presto_cpp/main/common/ConfigReader.cpp
```diff
@@ -36,10 +36,16 @@ std::unordered_map<std::string, std::string> readConfig(
       continue;
     }
 
-    std::vector<std::string> configParts;
-    folly::split('=', line, configParts);
-    VELOX_USER_CHECK_EQ(configParts.size(), 2, "Malformed config: {}", line);
-    properties.emplace(configParts[0], configParts[1]);
+    auto delimiterPos = line.find('=');
```
xiaoxmeng, line 39: nit: const auto delimiterPos =

presto-native-execution/presto_cpp/main/common/ConfigReader.cpp
```diff
@@ -36,10 +36,16 @@ std::unordered_map<std::string, std::string> readConfig(
...
+    VELOX_CHECK_NE(
+        delimiterPos,
+        std::string::npos,
+        "No '=' sign found for property pair '{}'",
+        line);
+    auto name = line.substr(0, delimiterPos);
```
xiaoxmeng, line 45: const nits.

xiaoxmeng (commented):
@tanjialiang thanks!

xiaoxmeng (approved):
@tanjialiang LGTM. Thanks for the update!

presto-native-execution/presto_cpp/main/common/ConfigReader.cpp
tanjialiang: at the start of the loop we get xxx=yyy so we get name = "xxx"

tanjialiang: The original logic essentially takes out all spaces in the line before proceeding to parse the key and values. I assume the assumption was made that no spaces are allowed in keys and values.

mbasmanova: Got it. Thank you for clarifying. It would be nice to document this behavior.

mbasmanova: Thinking about it again, this logic seems dangerous. Looks like it allows

```
foo=bar
f o o = b  ar
```

Does it make sense to allow spaces in config names? CC: @amitkdutta @xiaoxmeng

PR #22367, opened 2024-03-29
Skip IcebergPlanOptimizer for non-DATA tables

ZacBlanco:
The rule was being incorrectly applied to the $changelog system table. It would assume certain columns were partition columns resulting in incorrect query output.

## Description

Added a conditional to skip the optimization if a non-DATA table is in use.

## Motivation and Context

Incorrect query results with the following example:

```sql
create table test_v2 (a int, b int) WITH (format_version = '2', delete_mode = 'copy-on-write', partitioning = Array['a']);
insert into test_v2 values (1,1);
insert into test_v2 values (2,2);
delete from test_v2 where a = 1;

presto:test_demo> select * from "test_v2$changelog";
 operation | ordinal |     snapshotid      |  rowdata   
-----------+---------+---------------------+------------
 INSERT    |       0 | 4301417906194409386 | {a=2, b=2} 
 DELETE    |       1 | 8675841727593397135 | {a=1, b=1} 
(2 rows)

presto:test_demo> select * from "test_v2$changelog" where ordinal = 0;     // correct 
 operation | ordinal |     snapshotid      |  rowdata   
-----------+---------+---------------------+------------
 INSERT    |       0 | 4301417906194409386 | {a=2, b=2} 
(1 row)

presto:test_demo> select * from "test_v2$changelog" where operation = 'INSERT'; // wrong result 
 operation | ordinal |     snapshotid      |  rowdata   
-----------+---------+---------------------+------------
 INSERT    |       0 | 4301417906194409386 | {a=2, b=2} 
 INSERT    |       1 | 8675841727593397135 | {a=1, b=1} 
(2 rows)

presto:test_demo> select * from "test_v2$changelog" where operation = 'BAD'; // wrong,  
 operation | ordinal |     snapshotid      |  rowdata   
-----------+---------+---------------------+------------
 BAD       |       1 | 8675841727593397135 | {a=1, b=1} 
 BAD       |       0 | 4301417906194409386 | {a=2, b=2} 
```

## Impact

N/A

## Test Plan

New test for the changelog table ensures query result output is correct.

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergPlanOptimizer.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergPlanOptimizer.java
@@ -54,6 +54,7 @@ import static com.facebook.presto.expressions.LogicalRowExpressions.FALSE_CONSTA
 import static com.facebook.presto.expressions.LogicalRowExpressions.TRUE_CONSTANT;
 import static com.facebook.presto.hive.BaseHiveColumnHandle.ColumnType.SYNTHESIZED;
 import static com.facebook.presto.iceberg.IcebergSessionProperties.isPushdownFilterEnabled;
+import static com.facebook.presto.iceberg.IcebergTableType.DATA;
 import static com.facebook.presto.iceberg.IcebergUtil.getIcebergTable;
 import static com.facebook.presto.spi.ConnectorPlanRewriter.rewriteWith;
 import static com.google.common.base.Preconditions.checkArgument;
@@ -125,6 +126,9 @@ public class IcebergPlanOptimizer
             }
 
             TableScanNode tableScan = (TableScanNode) filter.getSource();
+            if (((IcebergTableHandle) tableScan.getTable().getConnectorHandle()).getIcebergTableName().getTableType() != DATA) {
+                return visitPlan(filter, context);
+            }
 
             Map<String, IcebergColumnHandle> nameToColumnHandlesMapping = tableScan.getAssignments().entrySet().stream()
                     .collect(Collectors.toMap(e -> e.getKey().getName(), e -> (IcebergColumnHandle) e.getValue()));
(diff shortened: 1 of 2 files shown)


presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergPlanOptimizer.java
```diff
@@ -125,6 +126,9 @@ public PlanNode visitFilter(FilterNode filter, RewriteContext<Void> context)
             }
 
             TableScanNode tableScan = (TableScanNode) filter.getSource();
+            if (((IcebergTableHandle) tableScan.getTable().getConnectorHandle()).getIcebergTableName().getTableType() != DATA) {
+                return visitPlan(filter, context);
+            }
```
hantangwangd, line 131: Should we reserve the table type of `EQUALITY_DELETES` and `DATA_WITHOUT_EQUALITY_DELETES`? Intuitively they are both partitioning aware, so might fit this optimization.

ZacBlanco: Just to confirm, you mean that this rule should also run for the `EQUALITY_DELETES` and `DATA_WITHOUT_EQUALITY_DELETES`?

hantangwangd: Yes, but I'm not so sure, need to further confirm, I just bring it up for discussion. So, we can filter out them now and add them back when we confirm they are suitable, or we can just reserve them now and discard them when we find they are not suitable here. Which one do you prefer?

ZacBlanco: I think we should confirm that they are suitable for this rule before allowing them. If we haven't confirmed yet, then let's exclude it until it is confirmed

hantangwangd: Sure, that's no problem.

hantangwangd (approved):
LGTM.

PR #22378, opened 2024-03-31
[native] Centralize sslcontext creation.

amitkdutta:
Currently SSL context creation code is duplicated between PrestoServer and tests. This can led to a situation where tests create context in a different way than server, and thus skipping server's sslcontext creation code path. Here we centralize sslcontext creation, so that server and tests use exactly same code path.


xiaoxmeng (approved):
@amitkdutta LGTM. Thanks!

Issue #22040, opened 2024-02-28
map_top_n returns wrong results if NaN appears in the input


rschlussel:
This issue was discovered as part of an audit of all the comparison and ordering behaviors for NaN across Presto functions (related to https://github.com/prestodb/presto/issues/21936 and https://github.com/prestodb/presto/issues/21877). 

While there are a lot of inconsistencies in how NaN are handled that need to be addressed,  `map_top_n` can produce definite wrong results when NaN values shows up in the map.

According to the documentaion, map_top_n "Truncates map items. Keeps only the top N elements by value. n must be a non-negative integer"

In the presence of NaN values, NaN seems to "reset"  the search for topn entries
```
 
select map_top_n(map(array['a', 'b', 'c'], array[nan(), 3, 2]),1);
  _col0               
---------
 {b=3.0} 
(1 row)

-- BUG! regardless of interpretation of NaN, 2 is always less than 3. So this result is definitely incorrect
 select map_top_n(map(array['a', 'b', 'c'], array[3, nan(), 2]),1);
  _col0               
---------
 {c=2.0} 
(1 row)

select map_top_n(map(array['a', 'b', 'c'], array[3, 2, nan()]),1);
  _col0               
---------
 {c=NaN} 
(1 row)
```

Issue #16851, opened 2021-10-08
NaN IS DISTINCT FROM NaN returns different results depending on the NaN's representation


tomvanbussel:
The query
```sql
SELECT nan() IS DISTINCT FROM nan_column FROM parquet_table
```
where `nan_column` is a column containing only NaNs, can return different results depending on the bit representation of the NaNs. This will return false if the NaN has the same representation as `nan()` and true otherwise.

Issue #21936, opened 2024-02-15
Clarifications on comparisons with NaN


kgpai:
## Current Behavior
Currently in Presto NaN() seems to behave like its > infinity,

``` 
select ARRAY_SORT(a) from  (values array[NAN(), INFINITY(), -INFINITY(), 1]) as T(a); 
—  [ "-Infinity", 1,  "Infinity",  "NaN" ]

select ARRAY_SORT_Desc(a) from  (values array[NAN(), INFINITY(), -INFINITY(), 1]) as T(a); 
—  ["NaN", "Infinity", 1, "-Infinity"]
```

This is not just limited to array_sort and you can see this behavior in order by too: 
``` 
SELECT x, y from (values (1, 1.0), (2, NAN()), (3, -NAN()), (4, INFINITY()), (5, -INFINITY())) as T(x, y) order by y;

— 
5	-Infinity
1	1
4	Infinity
2	NaN
3	NaN
```

However running 
``` 
SELECT NAN() > INFINITY()

-- False
```

gives a contradictory answer. 

## Expected Behavior
Comparison with NaN is not defined so we should confirm to https://en.wikipedia.org/wiki/NaN#Comparison_with_NaN. 

## Context
Velox is trying to ensure correctness with Presto Java but we are seeing inconsistencies in behavior and wanted clarifications on these.

kgpai:
cc: @mbasmanova @spershin @tdcmeehan @Yuhta @kagamiori

mbasmanova:
CC: @kaikalur @feilong-liu

tdcmeehan:
As discussed in #21877, I am not sure if it's most appropriate to follow the IEEE standard, or to follow the behavior of most engines.  It appears that behavior of NaN is not defined in the spec, so we are actually free to follow whatever behavior is most intuitive to users--in my opinion, this should be consistent with that of other engines.  The real requirement is to make them consistent, and document the behavior well for functions.

rschlussel:
I've been looking into the NaN inconsistencies, and wanted to give an update/summary of my thinking. If we can get alignment on the solution we can start tackling the problem

reference document of presto’s treatment of NaN for comparison and ordering across all functions and operators: https://docs.google.com/spreadsheets/d/1KclsskH88CLHMiu_Q2P25S3QfPc9H3Tu7h_9MXP10Dk/edit#gid=128250498

## What does IEEE-754 say about NaN 

[...]

PR #22386, opened 2024-04-01
Implement new NaN behavior

rschlussel:
## Description
This PR contains all the changes to overhaul the NaN operators to conform with the new definition proposed in https://github.com/prestodb/rfcs/blob/main/RFC-0001-nan-definition.md.

According to the new nan definition, Nan is larger than all other numbers and is equal to itself.  This PR also changes +0 and -0 to always be considered equal/not distinct, whereas previously there was inconsistency here as well.  

I recommend reviewing commit by commit as it is divided into logically distinct pieces that should be easier to review. If it would be helpful I can also split it up into smaller PRs.

Fixes the following issues:
https://github.com/prestodb/presto/issues/22040
https://github.com/prestodb/presto/issues/21936
https://github.com/prestodb/presto/issues/21877
https://github.com/prestodb/presto/issues/22679
https://github.com/prestodb/presto/issues/13807
https://github.com/prestodb/presto/issues/21065
https://github.com/prestodb/presto/issues/22716
https://github.com/facebookincubator/velox/pull/9511

It should also fix should also fix https://github.com/prestodb/presto/issues/16851, though I'm not sure how to create the file to test it.  

## Motivation and Context
The motivation for this change is to provide consistent NaN semantics across all of our functions and operators.  It is also to ensure that these semantics are consistent with velox as we move to native workers. For more details see the RFC: https://github.com/prestodb/rfcs/blob/main/RFC-0001-nan-definition.md.

## Impact
nan will now be treated as greater than all other numbers for all functions and as equal to itself for all functions. This changes the behavior of many existing functions and operators. These differences include (but are not limited to) =, <, >, joins, various distincting functions like set_agg and array_distinct,  array_min.

It also fixes https://github.com/prestodb/presto/issues/22040, a wrong results bug with map_top_n in the presence of nans.

This PR also changes joins and aggregations to treat +0 and -0 as equal/not distinct.

## Test Plan
Added tests for all affected functions.

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Change handling of floating point numbers in Presto to consistently treat NaNs as larger than any other number and equal to itself.  It also changes the handling of positive and negative zero to always be considered equal to each other.  Read more here: https://github.com/prestodb/rfcs/blob/main/RFC-0001-nan-definition.md. The new nan behavior can be disabled by setting the configuration property ``use-new-nan-definition`` to ``false``. This configuration property is intended to be temporary to ease migration in the short term, and will be removed in a future release.
* Fix a bug where map_top_n could return wrong results if there is any NaN input
* Fix a bug with array_min/array_max where it would return NaN rather than null when there was both NaN and null input. 
```

### DIFF
--- a/presto-common/src/main/java/com/facebook/presto/common/predicate/TupleDomainFilterUtils.java
+++ b/presto-common/src/main/java/com/facebook/presto/common/predicate/TupleDomainFilterUtils.java
@@ -148,7 +148,7 @@ public class TupleDomainFilterUtils
 
     /**
      * Returns true is ranges represent != or NOT IN filter for double, float or string column.
-     *
+     * <p>
      * The logic is to return true if ranges are next to each other, but don't include the touch value.
      */
     private static boolean isNotIn(List<Range> ranges)
@@ -161,7 +161,7 @@ public class TupleDomainFilterUtils
         Marker previousHigh = firstRange.getHigh();
 
         Type type = previousHigh.getType();
-        if (type != DOUBLE && type != REAL && !isVarcharType(type) && !(type instanceof CharType)) {
+        if (!type.equals(DOUBLE) && !type.equals(REAL) && !isVarcharType(type) && !(type instanceof CharType)) {
             return false;
         }
 
@@ -219,10 +219,10 @@ public class TupleDomainFilterUtils
             checkArgument(range.isSingleValue(), "Unexpected range of boolean values");
             return BooleanValue.of(((Boolean) range.getSingleValue()).booleanValue(), nullAllowed);
         }
-        if (type == DOUBLE) {
+        if (type.equals(DOUBLE)) {
             return doubleRangeToFilter(range, nullAllowed);
         }
-        if (type == REAL) {
+        if (type.equals(REAL)) {
             return floatRangeToFilter(range, nullAllowed);
         }
         if (type instanceof DecimalType) {
--- a/presto-common/src/main/java/com/facebook/presto/common/predicate/TupleDomainFilterUtils.java
+++ b/presto-common/src/main/java/com/facebook/presto/common/predicate/TupleDomainFilterUtils.java
@@ -25,6 +25,8 @@ import com.facebook.presto.common.predicate.TupleDomainFilter.DoubleRange;
 import com.facebook.presto.common.predicate.TupleDomainFilter.FloatRange;
 import com.facebook.presto.common.predicate.TupleDomainFilter.LongDecimalRange;
 import com.facebook.presto.common.predicate.TupleDomainFilter.MultiRange;
+import com.facebook.presto.common.predicate.TupleDomainFilter.OldDoubleRange;
+import com.facebook.presto.common.predicate.TupleDomainFilter.OldFloatRange;
 import com.facebook.presto.common.type.CharType;
 import com.facebook.presto.common.type.DecimalType;
 import com.facebook.presto.common.type.Type;
@@ -136,7 +138,9 @@ public class TupleDomainFilterUtils
             }
         }
 
-        if (firstRangeFilter instanceof DoubleRange || firstRangeFilter instanceof FloatRange) {
+        // if we are dealing with OldDoubleRange or OldFloatRange, we are using the old nan definition
+        // and need special handling for NaNs
+        if (firstRangeFilter instanceof OldDoubleRange || firstRangeFilter instanceof OldFloatRange) {
             // != and NOT IN filters should return true when applied to NaN
             // E.g. NaN != 1.0 as well as NaN NOT IN (1.0, 2.5, 3.6) should return true; otherwise false.
             boolean nanAllowed = isNotIn(ranges);
@@ -220,10 +224,10 @@ public class TupleDomainFilterUtils
             return BooleanValue.of(((Boolean) range.getSingleValue()).booleanValue(), nullAllowed);
         }
         if (type.equals(DOUBLE)) {
-            return doubleRangeToFilter(range, nullAllowed);
+            return doubleRangeToFilter(range, nullAllowed, type == DOUBLE);
         }
         if (type.equals(REAL)) {
-            return floatRangeToFilter(range, nullAllowed);
+            return floatRangeToFilter(range, nullAllowed, type == REAL);
         }
         if (type instanceof DecimalType) {
             if (((DecimalType) type).isShort()) {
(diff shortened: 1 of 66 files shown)


elharo (commented):
A lot (all?) of these type comparisons look like they should be .equals for everything, not just the real and double types. Type is not an enum. I guess they're supposed to be singletons? But then why doesn't this work for real and double?

rschlussel:
It doesn't work for real and double in this PR because i added a flag to it for the nan migration, so depending on the configuration property, the flag could be true or false.  (Also, while you're certainly welcome to review while it's in this state, I'm planning to clean up the commits to be logically independent once I've finished fixing the tests/adding tests and covering all the cases).

presto-docs/src/main/sphinx/language/types.rst
```diff
@@ -65,12 +65,22 @@ Floating-Point
 A real is a 32-bit inexact, variable-precision implementing the
 IEEE Standard 754 for Binary Floating-Point Arithmetic.
 
+It strays from the IEEE standard when it comes to handling of nans.
```
steveburnett, line 68:
```suggestion
Presto strays from the IEEE standard when handling NaNs.
```
Is "Presto" correct? I found "It" vague. Also, consider this minor rephrasing suggestion for conciseness.

presto-docs/src/main/sphinx/language/types.rst
```diff
@@ -65,12 +65,22 @@ Floating-Point
...
+In Presto, NaN is considered larger than any other value for
+all comparison and sorting operations. Additionally, NaN=NaN will
+be true for all equality and distinctness purposes
```
steveburnett, line 71:
```suggestion
be true for all equality and distinctness purposes.
```

presto-docs/src/main/sphinx/language/types.rst
```diff
@@ -65,12 +65,22 @@ Floating-Point
...
+
 ``DOUBLE``
 ^^^^^^^^^^
 
 A double is a 64-bit inexact, variable-precision implementing the
 IEEE Standard 754 for Binary Floating-Point Arithmetic.
 
+It strays from the IEEE standard when it comes to handling of nans.
```
steveburnett, line 79:
```suggestion
Presto strays from the IEEE standard when handling NaNs.
```

presto-docs/src/main/sphinx/language/types.rst
```diff
@@ -65,12 +65,22 @@ Floating-Point
...
+In Presto, NaN is considered larger than any other value for
+all comparison and sorting operations. Additionally, NaN=NaN will
+be true for all equality and distinctness purposes
```
steveburnett, line 82:
```suggestion
be true for all equality and distinctness purposes.
```

steveburnett (changes requested):
Thanks for the doc! Only a few nits for you to consider.

rschlussel:
test failure will be addressed by https://github.com/prestodb/presto/pull/22888

rschlussel:
Thanks @steveburnett. I've addressed your comments

rschlussel:
FYI @bikramSingh91

presto-common/src/main/java/com/facebook/presto/common/predicate/TupleDomainFilterUtils.java
```diff
@@ -148,7 +152,7 @@ public static TupleDomainFilter toFilter(Domain domain)
 
     /**
      * Returns true is ranges represent != or NOT IN filter for double, float or string column.
```
elharo, line 154: is --> if

elharo (approved):

presto-docs/src/main/sphinx/language/types.rst
```diff
@@ -65,12 +65,22 @@ Floating-Point
 A real is a 32-bit inexact, variable-precision implementing the
 IEEE Standard 754 for Binary Floating-Point Arithmetic.
 
+Presto strays from the IEEE standard when handling nans.
```
steveburnett, line 68:
```suggestion
Presto strays from the IEEE standard when handling NaNs.
```
Nit

presto-docs/src/main/sphinx/language/types.rst
```diff
@@ -65,12 +65,22 @@ Floating-Point
...
+all comparison and sorting operations. Additionally, NaN=NaN will
+be true for all equality and distinctness purposes.
+
 ``DOUBLE``
 ^^^^^^^^^^
 
 A double is a 64-bit inexact, variable-precision implementing the
 IEEE Standard 754 for Binary Floating-Point Arithmetic.
 
+Presto strays from the IEEE standard when handling nans.
```
steveburnett, line 79:
```suggestion
Presto strays from the IEEE standard when handling NaNs.
```

steveburnett (changes requested):
Two nits to consider, nothing else. Thanks!

steveburnett (approved):
LGTM! (docs)

Pull updated branch, new docs build, everything looks good. Thanks!

presto-common/src/main/java/com/facebook/presto/common/predicate/TupleDomainFilterUtils.java
rschlussel: did it separately https://github.com/prestodb/presto/pull/22905

sdruzkin (commented):
presto-orc ✅

sdruzkin (dismissed):

jaystarshot (approved):

presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java
```diff
@@ -3103,4 +3105,17 @@ public FeaturesConfig setPrintEstimatedStatsFromCache(boolean printEstimatedStat
         this.printEstimatedStatsFromCache = printEstimatedStatsFromCache;
         return this;
     }
+
+    public boolean getUseNewNanDefinition()
+    {
+        return useNewNanDefinition;
+    }
+
+    @Config("use-new-nan-definition")
```
tdcmeehan, line 3114: Should we call this the new NaN definition, and by default this is false, or should we refer to the old behavior as the deprecated NaN definition, and by default that is true?

tdcmeehan (commented):
`Add property for new nan behavior`

presto-common/src/main/java/com/facebook/presto/common/type/TypeUtils.java
```diff
@@ -212,4 +213,52 @@ static <V> Map<String, V> normalizeEnumMap(Map<String, V> entries)
...
+        // this ensures that doubleCompare(NaN, NaN) will return 0
+        long aBits = doubleToLongBits(a);
+        long bBits = doubleToLongBits(b);
+        if (aBits < bBits) {
+            return -1;
+        }
+        if (aBits > bBits) {
+            return 1;
+        }
+        return 0;
```
tdcmeehan, line 263: In theory, isn't it possible for multiple NaNs to have separate values when represented as bits?  Supposing this came from a datasource and not from Presto itself.  If so, would this work?

tdcmeehan (commented):
`Add operator double support for new NaN...`

tdcmeehan (commented):
`Add NaN definition to DoubleType ...`

presto-common/src/main/java/com/facebook/presto/common/type/DoubleType.java
```diff
@@ -149,7 +153,11 @@ public final BlockBuilder createFixedSizeBlockBuilder(int positionCount)
     @SuppressWarnings("EqualsWhichDoesntCheckParameterClass")
     public boolean equals(Object other)
     {
-        return other == DOUBLE;
+        // this is safe because we will only have one type of double in the system at
+        // a time. .equals() comparison is always used against the static DOUBLE
+        // instance to check if something is double type. this hack is temporary
+        // and will be removed when we full remove the old nan behavior
+        return other == DOUBLE || other == OLD_NAN_DOUBLE;
```
tdcmeehan, line 165: Should we just do an instanceof check?  It wouldn't require a lengthy explanation.  Likewise for float.

presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java
tdcmeehan: Oh, I see the later commit that enables it.  I would have recommended this if it were disabled by default, but I think it's fine as it is.

presto-common/src/main/java/com/facebook/presto/common/type/TypeUtils.java
rschlussel: doubleToLongBits coerces all the nans to the same representation.  There's a different function Double.doubelToRawLongBits() that doesn't do that.

rschlussel: I added a comment in the code and some tests for this.

tdcmeehan (dismissed):
This looks awesome, thanks for the great work @rschlussel.

PR #22412, opened 2024-04-03
Improve error for bad row cast

rschlussel:
## Description

Selecting from a row with two columns with the same name (or same name when all lower case) results in an ambiguous field reference.  Throw a semantic exception instead of a GENERIC_INTERNAL_ERRROR even though this exception gets caught after we've finished analysis.

For example
`SELECT CAST(ROW(1, 2) AS ROW(a BIGINT, A DOUBLE)).a` will now return a semantic exception.

This also improves the test for this case to check for a specific exception and removes a deprecated method for testing invalid functions that catches all runtime exceptions.

## Motivation and Context
Motivation is to improve an error to correctly reflect that it is a semantic error from the user.

## Impact
New exception type and error code for these queries

## Test Plan
unit test

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/sql/planner/ExpressionInterpreter.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/planner/ExpressionInterpreter.java
@@ -650,11 +650,11 @@ public class ExpressionInterpreter
                 Type type = type(node.getValue());
                 List<Expression> expressionValues = toExpressions(values, types);
                 List<Expression> simplifiedExpressionValues = Stream.concat(
-                        expressionValues.stream()
-                                .filter(ExpressionDeterminismEvaluator::isDeterministic)
-                                .distinct(),
-                        expressionValues.stream()
-                                .filter((expression -> !isDeterministic(expression))))
+                                expressionValues.stream()
+                                        .filter(ExpressionDeterminismEvaluator::isDeterministic)
+                                        .distinct(),
+                                expressionValues.stream()
+                                        .filter((expression -> !isDeterministic(expression))))
                         .collect(toImmutableList());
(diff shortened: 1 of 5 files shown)


pranjalssh (approved):
Does this need a release note?

hantangwangd:
Currently, there are several different statements related to `row cast` that have different behaviors. Examples are as follows:

```
presto> select typeof(cast(row(1, 2) as ROW(KEY1 VARCHAR, "key1" VARCHAR)));
                _col0
-------------------------------------
 row("key1" varchar, "key1" varchar)
(1 row)

presto> select cast(row(1, 2) as ROW(KEY1 VARCHAR, "key1" VARCHAR)).key1;
Query 20240209_085621_09358_9e2qz failed: Ambiguous field com.facebook.presto.common.type.RowType$Field@7b7cd13 in type row("key1" varchar, "key1" varchar)

presto> select cast(row(1, 2) as ROW(KEY1 VARCHAR, "key1" VARCHAR));
  _col0
----------
 {key1=2}
(1 row)
```

So, should we handle this problem in analyzing stage so that we can unify the behaviors for these different statements? PR #21869 happens to handle this problem as part of a bigger problem about how to treat the case for quoted field in row type when casting. It handled the field names duplication in `TypeSignature.parseRowTypeSignature(..)`so that all statements in above examples would fail universally in analyzing stage, please take a look when convenient. Any thought would be appreciated!

rschlussel:
Thanks @hantangwangd! i didn't realize you were already working on this.  This was an opportunistic fix on my part because I wanted to get rid of the deprecated test methods.  I agree that fixing in the analyzer stage is better.  I'll close this in favor of your PR.

PR #22424, opened 2024-04-04
Use multipart uploads for large writes in PrestoS3FileSystem

ZacBlanco:
CTAS without partitioning will, for each node, basically write one huge file to the local disk before uploading it when the write it complete. This change attempts to upload as the write it occurring to prevent excessive disk usage.

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* ...
* ...

Hive Connector Changes
* ...
* ...
```

```
== NO RELEASE NOTE ==
```


presto-hive/src/test/java/com/facebook/presto/hive/s3/TestPrestoS3FileSystem.java
```diff
@@ -793,6 +797,44 @@ public ObjectListing listObjects(ListObjectsRequest listObjectsRequest)
...
+            assertEquals(s3.putObjectRequests, 2);
+
+            try (FSDataOutputStream out = fs.create(new Path("test"))) {
+                out.write(new byte[600]);
+                out.write(new byte[600]);
+            }
+            assertEquals(s3.initiateMultipartRequests, 1);
+            assertEquals(s3.uploadPartRequests, 2);
+            assertEquals(s3.completePartUploadRequests, 1);
+        }
```
ClarenceThreepwood, line 835: Maybe add a test that simulates error and subsequent cleanup?

presto-hive-common/src/main/java/com/facebook/presto/hive/HiveCommonSessionProperties.java
```diff
@@ -176,6 +179,11 @@ public HiveCommonSessionProperties(HiveCommonClientConfig hiveCommonClientConfig
                         READ_MASKED_VALUE_ENABLED,
                         "Return null when access is denied for an encrypted parquet column",
                         hiveCommonClientConfig.getReadNullMaskedParquetEncryptedValue(),
+                        false),
+                dataSizeProperty(
+                        S3_MIN_MULTIPART_PART_SIZE,
+                        "minimum upload size for multipart uploads to s3",
+                        io.airlift.units.DataSize.succinctDataSize(100, DataSize.Unit.MEGABYTE),
```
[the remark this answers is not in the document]
ZacBlanco: It's already configurable within the hive connector. I just exposed it as a session parameter for benchmarking purposes. I am fine removing it if you don't think it should be a session parameter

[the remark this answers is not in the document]
ClarenceThreepwood: NM - I was looking in the wrong place. This is in HiveS3Config

However the default there appears to be 5MB. Should we make these consistent?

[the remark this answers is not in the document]
ZacBlanco: I agree on making them consistent. However, S3 also enforces a 5MB minimum part limit with a maximum of 10k parts. This gives a default maximum file size of ~50G. Given that with this new implementation we don't know the total file size ahead of time, 50G seems like a relatively low limit, at least on the scale of Presto :)

I did intentionally set this to a larger value to allow for larger file sizes - 100MiB gives a 20x increase to the 50G limit for a max upload of ~1T, with little downsides IMO. Larger part sizes should generally decrease overhead due to having to make lest S3 requests.

I think it is better to make the default minimum part size larger than the current 5MiB. 100MiB was mostly an arbitrary pick to lift the file limit, but I am open to other suggestions.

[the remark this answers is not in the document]
majetideepak: For reference, we use 10MiB in Velox. That would give a 100GiB file size.
Velox also did not make the part size configurable. It helps to minimize the configs. If required, 
we can make the part size adaptive as we upload parts. Arrow does this.
Each presto writer will consume 100MiB. For CTAS queries with many writers, will there be a higher change of OOM?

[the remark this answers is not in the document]
ZacBlanco: The 10MiB reference from Velox is useful. I am okay aligning on that value for now just for consistency's sake. I also like the idea of using an adaptive approach. There are probably a few ways to do it - I think that's better left for a follow-up PR though.

As for the memory portion. 100MiB isn't necessarily used by each writer. When writing, the data first goes to a temp file on disk through a `BufferedInputStream` which I believe uses an 8K buffer by default. Then, when uploading through the S3 client, you specify the temp file on disk. The S3 client takes care of streaming it through it's own input stream. I dug through the S3 client code and didn't see anywhere that loads the part file all at once to be uploaded. It seems to be entirely streaming, so I think that memory usage/OOM is a concern regardless of part size.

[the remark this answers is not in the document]
majetideepak: If the upload is streaming from a file, then the memory usage should be low. The overhead should be similar to the TransferManager path?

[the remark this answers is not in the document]
ZacBlanco: Sorry - typo in my original comment. I meant to say it _is not_ a concern regardless of part size.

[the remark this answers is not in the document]
ZacBlanco: I've updated the configs in both places to default to 10MiB.

presto-hive/src/main/java/com/facebook/presto/hive/s3/PrestoS3FileSystem.java
```diff
@@ -1288,25 +1350,97 @@ public void close()
...
+                }
+
+                partList.add(client.uploadPart(request));
+                STATS.uploadSuccessful();
+                Files.deleteIfExists(partFile);
+                if (!last) {
+                    newTempFileBuffer();
+                }
+            }
+            catch (SdkClientException | IOException e) {
```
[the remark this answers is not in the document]
ZacBlanco: Good point. Originally, I was trying to mimic the error handling from the previous code, however I have given it more thought I think it does make sense to have a catch-all block here and abort regardless of the exception. This has been fixed

presto-hive/src/test/java/com/facebook/presto/hive/s3/TestPrestoS3FileSystem.java
ZacBlanco: Added

ClarenceThreepwood (dismissed):
lgtm

presto-hive/src/main/java/com/facebook/presto/hive/s3/PrestoS3FileSystem.java
```diff
@@ -1288,25 +1349,97 @@ public void close()
...
+                    }
                 }
-                // close transfer manager but keep underlying S3 client open
-                transferManager.shutdownNow(false);
             }
         }
 
-        private void uploadObject()
-                throws IOException
+        private void flushPartUpload(boolean last)
```
aaneja, line 1377: With this implementation, we are not utilizing the disk as a buffer adequately IMO. The uploads follow a create-part->upload-part-> delete-part flow, which means that our overall write speed is bounded by the network speed/bw (or the disk speed).
We could address this by creating a pipelined writer where the upload and delete of a part happens parallelly in the background and while the part-writer produces parts in a different thread. We can cap the total number of on-disk parts so that we don't fill up the disk.

aaneja:
>The additional performance gain is due to the fact that the original implementation relied on extending FilterOutputStream. In the default implementation of the FilterOutputStream, all writes end up occurring through the write(byte b) interface

Could you measure what the impact would be be if we fixed this problem only without replacing the TransferManager ? I'm also curious if you played around with [`task.writer-count`](https://github.com/prestodb/presto/blob/641c2693879270996ce463368b0d6afe0b636338/presto-main/src/main/java/com/facebook/presto/execution/TaskManagerConfig.java#L437-L458) (and `task.partitioned-writer-count`) to see if we can utilize the disk/network better. I found this [old thread](https://groups.google.com/g/presto-users/c/SooEm5B203c) which discusses the impact of these configs

presto-hive/src/test/java/com/facebook/presto/hive/s3/TestPrestoS3FileSystem.java
```diff
@@ -793,6 +800,109 @@ public ObjectListing listObjects(ListObjectsRequest listObjectsRequest)
         }
     }
 
+    @Test
+    public void testMultipart()
```
aaneja, line 804: Can you also add a test that writes a complete pre-canned file using the PrestoS3FileSystem and reads it back and asserts on its contents ? Testing around boundary cases (under min part size, just above min part size, large file size) will prove robustness

presto-hive/src/main/java/com/facebook/presto/hive/s3/PrestoS3FileSystem.java
```diff
@@ -1288,25 +1349,97 @@ public void close()
...
+            catch (Exception e) {
+                STATS.uploadFailed();
+                upload.ifPresent(this::cleanUpFailedMultipartUpload);
+                throw new RuntimeException(e);
+            }
+        }
+
+        private String uploadSingleFile()
+        {
+            TransferManager transfer = TransferManagerBuilder.standard()
```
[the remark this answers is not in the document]
ZacBlanco: It could also be that the min part size was configured to a very large size. I didn't dig into what optimizations the transfer manager might make for larger-sized files. I just deferred to that implementation since it seems robust enough to handle large file uploads

presto-hive/src/main/java/com/facebook/presto/hive/s3/PrestoS3FileSystem.java
ZacBlanco: I had considered something like this as an option. However, I felt the pipelined version was too complex to pursue. The main goal for this PR was just to lift the disk-size limitation. Additional performance gain was a non-goal, but we happened to get some by properly overriding the `FilterOutputStream`

ZacBlanco:
@aaneja I re-ran the benchmarks for creating the `tpch.orders` orders table and built a branch which only overrides the outputstream methods. Additionally, I tested with `task.writer-count` set to the default (1) and 8. I also tested with a larger min part size for one of the runs on my PR. Here are. the results:

image

presto-hive/src/main/java/com/facebook/presto/hive/s3/PrestoS3FileSystem.java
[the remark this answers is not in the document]
ZacBlanco: With the results [here](https://github.com/prestodb/presto/pull/22424#issuecomment-2052150813), I am of the opinion that we should let this PR merge since the performance is always better than master in the benchmark configurations. If we want to improve perf further, we can iterate in another PR on a proper streaming implementation. Trino also has a proper streaming implementation.

When implementing the pipelining I think there are more design decisions that we may wish to make (or further configuration) which is not really in scope for the goal of the PR, which was lifting the disk space limitation.

presto-hive/src/main/java/com/facebook/presto/hive/s3/PrestoS3FileSystem.java
ZacBlanco: With the results [here](https://github.com/prestodb/presto/pull/22424#issuecomment-2052150813), I am of the opinion that we should let this PR merge since the performance is always better than master in the benchmark configurations. If we want to improve perf further, we can iterate in another PR on a proper streaming implementation. Trino also has a proper streaming implementation.

When implementing the pipelining I think there are more design decisions that we may wish to make (or further configuration) which is not really in scope for the goal of the PR, which was lifting the disk space limitation.

presto-hive/src/test/java/com/facebook/presto/hive/s3/TestPrestoS3FileSystem.java
ZacBlanco: Added here: https://github.com/prestodb/presto/blob/ec82b5ea7861e87960288fae99956e3ffde98787/presto-hive/src/test/java/com/facebook/presto/hive/s3/TestPrestoS3FileSystem.java#L910

presto-hive/src/main/java/com/facebook/presto/hive/s3/PrestoS3FileSystem.java
aaneja: Sounds good. Can you make an issue and link as a TODO for the streaming implementation ?

presto-hive/src/test/java/com/facebook/presto/hive/s3/TestPrestoS3FileSystem.java
```diff
@@ -793,6 +803,179 @@ public ObjectListing listObjects(ListObjectsRequest listObjectsRequest)
...
+
+    private static void assertRoundTripWrite(IOSupplier<Path> writerExpected, FileSystem fs, byte[] actual)
+            throws IOException
+    {
+        Path writtenFile = writerExpected.get();
+        byte[] input = new byte[actual.length];
+        try (FSDataInputStream in = fs.open(writtenFile)) {
+            assertEquals(actual.length, in.read(input));
+        }
+        assertEquals(actual, input, "failed on length " + actual.length);
```
aaneja, line 969: The 'actual' here is the expected data

presto-hive/src/test/java/com/facebook/presto/hive/s3/TestPrestoS3FileSystem.java
```diff
@@ -793,6 +803,179 @@ public ObjectListing listObjects(ListObjectsRequest listObjectsRequest)
...
+        }
+    }
+
+    private static void assertRoundTripWrite(IOSupplier<Path> writerExpected, FileSystem fs, byte[] actual)
+            throws IOException
+    {
+        Path writtenFile = writerExpected.get();
+        byte[] input = new byte[actual.length];
+        try (FSDataInputStream in = fs.open(writtenFile)) {
+            assertEquals(actual.length, in.read(input));
```
aaneja, line 967: We're missing the case where the actually written data is larger than expected. How about we use this for the method -
```
private static void assertRoundTripWrite(IOSupplier<Path> writerExpected, FileSystem fs, byte[] expected)
            throws IOException
    {
        Path writtenFile = writerExpected.get();
        byte[] actual;
        try (FSDataInputStream in = fs.open(writtenFile)) {
            actual = ByteStreams.toByteArray(in);
        }
        assertEquals(actual, expected);
    }
```

presto-hive/src/test/java/com/facebook/presto/hive/s3/MockAmazonS3.java
```diff
@@ -150,4 +216,81 @@ public ObjectListing listObjects(ListObjectsRequest listObjectsRequest)
     public void shutdown()
     {
     }
+
+    @Override
```
aaneja, line 220: Nice work with getting these added to the mock !

aaneja:
Aside : Your [experiment](https://github.com/prestodb/presto/pull/22424#issuecomment-2052150813) showed that the default `task.writer-count` may not be appropriate. Do you think we should set this higher ? Any suggestions on how we could set this dynamically for a better OOTB experience ?

presto-hive/src/main/java/com/facebook/presto/hive/s3/PrestoS3FileSystem.java
ZacBlanco: Created https://github.com/prestodb/presto/issues/22536

presto-hive/src/test/java/com/facebook/presto/hive/s3/TestPrestoS3FileSystem.java
ZacBlanco: Thanks - good catch. I've updated the method

ZacBlanco:
I did some digging into `task.writer-count`. I think the ideal value will depend on a number of hardware factors: the writer's core count, disk metrics (capacity, bandwidth, media type), network bandwidth. Additionally, the rate at which data can be fed into a `TableWriter` operator will determine the write speed. That rate could depend on a variety of non-hardware factors too such as the file format and compression algorithm. There could be more. Those are just the factors I could think of off the top of my head.

In these few experiments I ran without compression and with NVMe disks to try to achieve the highest data rate possible to stress the new implementation. We found that the performance increased, but for other workloads without compression I can't say that we would see gains as large. I could also see a situation which, if deployed using traditional spinning platter disks, concurrent writes would lead to abysmal performance.

I don't think there's a one-size fits all where we could guarantee better performance, but I think most "standard" use cases ("standard" referring to our standard cluster types used for benchmarking) would probably benefit from a writer count at least equal to the number of threads on the machine. Most deployments nowadays are probably using SSD and have multiple cores, so I think it's probably safe to increase the parallelism by a bit since it seems to give a measurable benefit. This does affect table layouts though (more files), so that's another thing to consider.

aaneja (approved):

aaneja:
I think we need the task writer to be adaptive to scale out/in based on utilization levels. I'm not sure if Prestissimo does this either, I will check. We can suggest this improvement if it does not

ZacBlanco:
@tdcmeehan do you want to review this? Or know anyone else who might want to?

jaystarshot:
Nice! But can this increase aws operation cost if the buffer size is too small?

ZacBlanco:
@jaystarshot You're correct that it can increase the costs by a fair margin. The minimum part size for S3 is 5MB. I put together a spreadsheet with the cost calculations for uploading 1PB to S3 with a fixed file size of 256MB and varying min part sizes:

image

When using a small part size, the cost for uploading to S3 can be significantly higher than standard PUT requests. However, you also need to consider that previously, we deferred to the TransferManager class, which makes the decision on whether or not an upload should be a PutObject or Multipart, so I'm not 100% sure how often we already do multipart uploads with the existing implementation.

Additionally, if you're uploading data at this scale, the monthly storage cost is going to far outstrip the cost of uploading the data in the first place. If you were to store 1PB in S3 and use the minimum part size of 5MB, the one-time upload cost is just 5.13% of the monthly storage cost, linearly decreasing with increases in part size.

The data from the chart is in the following spreadsheet: https://docs.google.com/spreadsheets/d/1DNNpfbahHAiwcrBnhxaGlgexAhraVV9_yu9VUpxez3E/edit?usp=sharing

jaystarshot:
Sounds good, but for temporary tables it might make more sense to have a bigger buffer size.

tdcmeehan:
I think this warrants a release note, since there are consequences to one's S3 costs.  We haven't included a flag, so there's no way to back out of this change--is that OK?

ZacBlanco:
Users could "back out" of the increased costs by increasing the min-multipart-upload size configuration. If you'd prefer a boolean flag to flip to the old behavior I can do so.

I will add in a release note for it

tdcmeehan:
I think it's safer to introduce a schedule, where we introduce the change with a release note, then follow up with a change in flag default value, then deprecate the old functionality.  I'm just not confident that this won't have outsized impact on certain environments, and better visibility can help to mitigate that.

ZacBlanco:
I've updated the PR to add a config `presto.s3.use-streaming-uploads` which uses the new implementation. I split out the Input/OutputStream classes from the `PrestoS3FileSystem` file to make the class smaller and easier to read. The new implementation lives in `PrestoS3StreamingOutputStream`. Also, the "old" PrestoS3OutputStream implementation now overrides the `write` methods properly without changing any other behavior to get some performance wins by default.

presto-docs/src/main/sphinx/connector/hive.rst
```diff
@@ -411,7 +411,13 @@ Property Name                         Description
 
 ``hive.s3.multipart.min-file-size``   Minimum file size before multi-part upload to S3 is used.   ``16 MB``
 
-``hive.s3.multipart.min-part-size``   Minimum multi-part upload part size.                        ``5 MB``
+``hive.s3.multipart.min-part-size``   Minimum multi-part upload part size. Can also be set via    ``10MB``
```
steveburnett, line 414:
```suggestion
``hive.s3.multipart.min-part-size``   Minimum multi-part upload part size. Can also be set using  ``10MB``
```
Do not use Latin abbreviations. For context and suggestions, see the GitLab [recommended word list entry for via](https://docs.gitlab.com/ee/development/documentation/styleguide/word_list.html#via). 

(I tested this, and discovered I needed to delete two blank spaces between "using" and "``10MB``" or the extra characters in "using" threw off the format for the Default value column.)

presto-docs/src/main/sphinx/connector/hive.rst
```diff
@@ -411,7 +411,13 @@ Property Name                         Description
...
+                                      the session property ``s3_multipart_min_part_size``
```
steveburnett, line 415:
```suggestion
                                      the session property ``s3_multipart_min_part_size``.
```
Nit, punctuation.

steveburnett (changes requested):
Looks good! Two nits, nothing major.

presto-docs/src/main/sphinx/connector/hive.rst
ZacBlanco: fixed, thanks!

presto-docs/src/main/sphinx/connector/hive.rst
ZacBlanco: fixed

steveburnett (approved):
LGTM! (docs)

Pull updated branch, new docs build, docs look good. Thanks!

presto-hive/src/main/java/com/facebook/presto/hive/s3/PrestoS3InputStream.java
```diff
@@ -0,0 +1,349 @@
...
+                    .run("getS3Object", () -> {
+                        InputStream stream;
+                        try {
+                            GetObjectRequest request = new GetObjectRequest(host, keyFromPath(path))
+                                    .withRange(position, (position + length) - 1);
+                            stream = s3.getObject(request).getObjectContent();
+                        }
+                        catch (RuntimeException e) {
+                            stats.newGetObjectError();
+                            if (e instanceof AmazonS3Exception) {
```
tdcmeehan, line 117: Please move this into a dedicated catch clause?

presto-hive/src/main/java/com/facebook/presto/hive/s3/PrestoS3InputStream.java
```diff
@@ -0,0 +1,349 @@
...
+            return 0;
+        }
+
+        try {
+            return retry()
+                    .maxAttempts(maxAttempts)
+                    .exponentialBackoff(BACKOFF_MIN_SLEEP, maxBackoffTime, maxRetryTime, 2.0)
+                    .stopOn(InterruptedException.class, PrestoS3FileSystem.UnrecoverableS3OperationException.class, EOFException.class, FileNotFoundException.class, AbortedException.class)
+                    .onRetry(stats::newGetObjectRetry)
+                    .run("getS3Object", () -> {
```
tdcmeehan, line 108: I would move this lambda into a new method and use a method reference.

tdcmeehan (commented):
LGTM just a couple nits

presto-hive/src/main/java/com/facebook/presto/hive/s3/PrestoS3InputStream.java
ZacBlanco: I think it was originally written this way to only have to call `stats.newGetObjectError()` in only one catch branch to prevent mis-collecting stats. But I can put it in both catch clauses

presto-hive/src/main/java/com/facebook/presto/hive/s3/PrestoS3InputStream.java
ZacBlanco: updated

presto-hive/src/main/java/com/facebook/presto/hive/s3/PrestoS3InputStream.java
ZacBlanco: updated

presto-hive/src/main/java/com/facebook/presto/hive/s3/PrestoS3InputStream.java
```diff
@@ -0,0 +1,363 @@
...
+    @Override
+    public long getPos()
+    {
+        return nextReadPosition;
+    }
+
+    @Override
+    public int read()
+    {
+        // This stream is wrapped with BufferedInputStream, so this method should never be called
```
elharo, line 191: This is a public class so this is not guaranteed.

presto-hive/src/test/java/com/facebook/presto/hive/s3/TestPrestoS3FileSystem.java
```diff
@@ -794,6 +806,191 @@ public ObjectListing listObjects(ListObjectsRequest listObjectsRequest)
...
+            assertEquals(s3.uploadPartRequests, streaming ? 2 : 0);
+            assertEquals(s3.completePartUploadRequests, streaming ? 1 : 0);
+            assertEquals(s3.putObjectRequests, streaming ? 2 : 3);
+        }
+    }
+
+    private void assertDirEmpty(java.nio.file.Path path)
+            throws IOException
+    {
+        try (Stream<java.nio.file.Path> dirIter = Files.walk(path)) {
```
elharo, line 852: This is more complex than it has to be. You can just do path.toFile().list()

presto-hive/src/main/java/com/facebook/presto/hive/s3/PrestoS3InputStream.java
ZacBlanco: This a direct refactor from the original class which was private. I've removed the comment

presto-hive/src/test/java/com/facebook/presto/hive/s3/TestPrestoS3FileSystem.java
ZacBlanco: Thanks for the pointer, fixed

PR #22435, opened 2024-04-06
Document and test HiveFileSplit

elharo:
## Description
Added tests and API doc for HiveFileSplit

## Motivation and Context
"split" does not seem to be a common term of the art in computer science. The only references I could find to it referred to splitting ML data sets into training and verification data, which is not what this is. Also "split" is very hard to google for, and zero is not a positive number.

## Impact
More accurate precondition violation messages.

## Test Plan
CI

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-hive/src/main/java/com/facebook/presto/hive/HiveFileSplit.java
+++ b/presto-hive/src/main/java/com/facebook/presto/hive/HiveFileSplit.java
@@ -24,6 +24,11 @@ import java.util.Optional;
 import static com.google.common.base.Preconditions.checkArgument;
 import static java.util.Objects.requireNonNull;
 
+/**
+ * Identifies a piece of a file at a given path starting at a given position and continuing for length bytes.
+ * This class does not check whether the file exists, or, if it exists, that it is long enough to
+ * contain all the indicated bytes.
+ */
 public class HiveFileSplit
 {
     private final String path;
@@ -34,6 +39,15 @@ public class HiveFileSplit
     private final Optional<byte[]> extraFileInfo;
     private final Map<String, String> customSplitInfo;
 
+    /**
+     * @param path the absolute path to the file that contains the split
+     * @param start the position in the containing file of the first byte of the split
+     * @param length the number of bytes in the split
+     * @param fileSize the total length of the file that contains the split
+     * @param fileModifiedTime the most recent time when the file containing the split was modified
+     * @throws NullPointerException if any object parameter is null
+     * @throws IllegalArgumentException if any numeric parameter is less than zero
+     */
     @JsonCreator
     public HiveFileSplit(
             @JsonProperty("path") String path,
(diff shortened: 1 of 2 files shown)


steveburnett (approved):
LGTM! (docs)

presto-hive/src/main/java/com/facebook/presto/hive/HiveFileSplit.java
```diff
@@ -24,6 +24,11 @@
 import static com.google.common.base.Preconditions.checkArgument;
 import static java.util.Objects.requireNonNull;
 
+/**
+ * Represents a piece of a file at a given path starting at a given position and continuing for length bytes.
+ * All numeric values must be zero or greater. Object values must be non-null.
```
rschlussel, line 29: Suggest removing this line. I found it hard to understand what you meant here (i thought you were saying all numeric values in the file had to be non-negative), it can easily get out of date if someone happens to add some other field that can be negative, and this information is pretty easy to find in the constructor.  If you don't want to remove it I suggest specifying which fields can't be negative.  
```suggestion
```

presto-hive/src/main/java/com/facebook/presto/hive/HiveFileSplit.java
```diff
@@ -24,6 +24,11 @@
...
+ * There is no guarantee the file exists, or, if it exists, that it is large enough to provide length bytes.
```
rschlussel, line 30: I would remove this as well. I guess technically, there's no guarantee about these things because this is just a data structure storing some information, but we do get the path from the metastore and then get metadata for the file from the file system to be able to populate this, so if we exclude concurrent updates, then we do have guarantees about this.

rschlussel (commented):
Would be nice to add documentation to the ConnectorSplit interface as well to explain that a split represents a chunk of work for each worker to process.

presto-hive/src/main/java/com/facebook/presto/hive/HiveFileSplit.java
elharo: The API doc for a public class should not assume the class will only be used in a particular way. Unless we can make this class non-public (which maybe we can do? I've noticed Presto tends to have more public API than it needs.) This class provides no such guarantees.

presto-hive/src/main/java/com/facebook/presto/hive/HiveFileSplit.java
elharo: done

elharo:
> Would be nice to add documentation to the ConnectorSplit interface as well to explain that a split represents a chunk of work for each worker to process.

https://github.com/prestodb/presto/issues/22458

presto-hive/src/main/java/com/facebook/presto/hive/HiveFileSplit.java
rschlussel: maybe rephrase it to say that the creator of a HiveFileSplit is responsible for ensuring the existence of the file, start and length, and this class does not do any such validation itself.

elharo: rephrased

rschlussel (approved):

PR #19145, opened 2023-03-07
Add support for Filelist caching for symlink tables

agrawalreetika:
Test plan - (Please fill in how you tested your changes)

Using the hive catalog access the symlink tables by adding following properties:

```
hive.file-status-cache-expire-time=24h
hive.file-status-cache-size=100000000
hive.file-status-cache-tables=*
```

Metrics Queries -
```
presto> SELECT * FROM jmx.current."com.facebook.presto.hive:name=<catalog>,type=cachingdirectorylister";
presto> SELECT "listlocatedstatuscalls.totalcount", "listobjectscalls.totalcount", "liststatuscalls.totalcount" from jmx.current."com.facebook.presto.hive.s3:name=<catalog>,type=prestos3filesystem";
```

== RELEASE NOTES ==
```
Hive Changes
* Add support for Filelist caching for symlink tables
```

Issue #22425, opened 2024-04-05
TestDirectoryListCacheInvalidation.testDirectoryListCacheInvalidation flakes


elharo:
part of [product-tests-basic-environment](https://github.com/prestodb/presto/actions/runs/8563059094/job/23486572111?pr=22008#logs)

```
2024-04-05T12:44:45.2986763Z 2024-04-05 18:29:45 INFO: FAILURE     /    com.facebook.presto.tests.hive.TestDirectoryListCacheInvalidation.testDirectoryListCacheInvalidation (Groups: hive_list_caching) took 1.1 seconds
2024-04-05T12:44:45.2992676Z 2024-04-05 18:29:45 SEVERE: Failure cause:
2024-04-05T12:44:45.2993582Z java.lang.AssertionError: expected [0] but found [1]
2024-04-05T12:44:45.2994337Z 	at org.testng.Assert.fail(Assert.java:110)
2024-04-05T12:44:45.2995197Z 	at org.testng.Assert.failNotEquals(Assert.java:1413)
2024-04-05T12:44:45.2996010Z 	at org.testng.Assert.assertEqualsImpl(Assert.java:149)
2024-04-05T12:44:45.2996800Z 	at org.testng.Assert.assertEquals(Assert.java:131)
2024-04-05T12:44:45.2997583Z 	at org.testng.Assert.assertEquals(Assert.java:911)
2024-04-05T12:44:45.2998343Z 	at org.testng.Assert.assertEquals(Assert.java:945)
2024-04-05T12:44:45.3001854Z 	at com.facebook.presto.tests.hive.TestDirectoryListCacheInvalidation.testDirectoryListCacheInvalidation(TestDirectoryListCacheInvalidation.java:44)
2024-04-05T12:44:45.3003719Z 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
2024-04-05T12:44:45.3004881Z 	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
2024-04-05T12:44:45.3006245Z 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
2024-04-05T12:44:45.3007602Z 	at java.lang.reflect.Method.invoke(Method.java:498)
2024-04-05T12:44:45.3008836Z 	at org.testng.internal.invokers.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:135)
[...]

PR #22443, opened 2024-04-08
Revert "Add support for Filelist caching for symlink tables"

elharo:
Reverts prestodb/presto#19145 due to flakiness. This PR depends on the PR that introduced the flakiness so roll this one back first.


tdcmeehan:
Why not just disable the product test?

tdcmeehan:
@elharo based on the stack trace in #22425, it seems to be failing during the setup portion of the test case.  To me it seems like a testing issue, and it would be quicker and more straightforward to just disable the test.  Even if you're correct, the procedure that was added is opt-in, and no one will be using this until 0.287 is released, and that release hasn't even been cut yet, so there's actually only a risk if someone is using recent unstable releases (which are not widely used).

agrawalreetika:
Hi @elharo @tdcmeehan,
I have been traveling, but I doubt if the feature is actually affected. I can take out some time and verify the feature if that works and we can disable the tests for now. As I checked the last some recent changes caused this failing and affecting JMX metrics in here.
LMK what do you guys think?

elharo:
I'm not comfortable with an in-between state where we leave the feature in untested. The safe options are fix or revert. If you can find and revert some recent change that caused this to start failing, that would be great. Otherwise, let's revert now and fix it at our leisure later.

agrawalreetika:
@elharo Could you please confirm if we are seeing test failure for https://github.com/prestodb/presto/blob/master/presto-product-tests/src/main/java/com/facebook/presto/tests/hive/TestSymlinkTableListCaching.java as well? I didn't find any CI example with failure on this.

elharo:
No, I don't think TestSymlinkTableListCaching is flaky but it depends on the PR that is. 

Are you proposing to block the release until this is fixed? That feels risky. It seems much safer to simply roll everything back, and then fix at our leisure. If the fix is done before the release, great. If not, nothing else needs to wait for it.

agrawalreetika:
@elharo Sure thanks for confirmation. These changes shouldn't be flaky.

Also for other one, I have found the RCA for flakiness for TestDirectoryListCacheInvalidation and proposed the test modification  https://github.com/prestodb/presto/pull/22455

tdcmeehan:
Superseded by #22455

PR #21929, opened 2024-02-14
[native] Add dockerfiles for dependencies and runtime and remove existing dockerfiles

majetideepak:
## Description

Remove the existing dockerfiles for Prestissimo as they are out of date.
We split the release-centos-dockerfile into dependency image dockerfile
and runtime image dockerfile.
This split is helpful since the dependencies are rarely updated. Prestissimo can
now build faster from the dependency image.
The orchestration will be moved to the helm project.
The release team will also use the new dockerfiles to release official dependency
and runtime Prestissimo images.

## Motivation and Context

There are no working dockerfiles for Prestissimo. This PR adds these files.

## Impact

Enable official Prestissimo dependency and runtime images.

## Test Plan

Approval from the release team.

## Contributor checklist

- [ ] Please make sure your submission complies with our [development](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#development), [formatting](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#formatting), [commit message](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#commit-formatting-and-pull-requests), and [attribution guidelines](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#attribution).
- [ ] PR description addresses the issue accurately and concisely.  If the change is non-trivial, a GitHub Issue is referenced.
- [ ] Documented new properties (with its default value), SQL syntax, functions, or other functionality.
- [ ] If release notes are required, they follow the [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines).
- [ ] Adequate tests were added if applicable.
- [ ] CI passed.

## Release Notes
Please follow [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines) and fill in the release notes below.

```
== RELEASE NOTES ==

Prestissimo (Native Execution)
* Add dockerfiles for dependencies and runtime

```

PR #22448, opened 2024-04-08
[native ] Fix setup scripts to align with Velox setup scripts

majetideepak:
## Description
Velox setup scripts have been modified to not install the build dependencies by default.
Fix setup scripts to adopt that behavior.
Fix setup scripts to encapsulate presto packages into a function.
Fix setup-centos8.sh script to include Velox dependencies.

## Motivation and Context
Above

## Contributor checklist

- [ ] CI passed.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/Makefile
+++ b/presto-native-execution/Makefile
@@ -113,7 +113,8 @@ centos-container:			#: Build the linux container for CircleCi
 linux-container:
 	rm -rf /tmp/docker && \
 	mkdir -p /tmp/docker && \
-	cp scripts/setup-$(CONTAINER_NAME).sh scripts/$(CONTAINER_NAME)-container.dockfile velox/scripts/setup-helper-functions.sh /tmp/docker && \
+	$(eval VELOX_SETUP_SCRIPT=$(shell [ $(CONTAINER_NAME) == centos ] && echo setup-centos8.sh; [ $(CONTAINER_NAME) != centos ] && echo setup-$(CONTAINER_NAME).sh;)) \
+	cp scripts/setup-$(CONTAINER_NAME).sh scripts/$(CONTAINER_NAME)-container.dockfile velox/scripts/${VELOX_SETUP_SCRIPT} velox/scripts/setup-helper-functions.sh /tmp/docker && \
 	cd /tmp/docker && \
 	docker build --build-arg cpu_target=$(CPU_TARGET) --tag "prestocpp/prestocpp-$(CPU_TARGET)-$(CONTAINER_NAME):$(USER)-$(shell date +%Y%m%d)" -f $(CONTAINER_NAME)-container.dockfile .
 
--- a/presto-native-execution/scripts/setup-centos.sh
+++ b/presto-native-execution/scripts/setup-centos.sh
@@ -17,42 +17,72 @@ set -x
 export FB_OS_VERSION=v2024.04.01.00
 export RE2_VERSION=2021-04-01
 export nproc=$(getconf _NPROCESSORS_ONLN)
-
-dnf install -y maven java python3-devel clang-tools-extra jq perl-XML-XPath
-
-python3 -m pip install regex pyyaml chevron black
-
 export CC=/opt/rh/gcc-toolset-9/root/bin/gcc
 export CXX=/opt/rh/gcc-toolset-9/root/bin/g++
 
 CPU_TARGET="${CPU_TARGET:-avx}"
 SCRIPT_DIR=$(readlink -f "$(dirname "${BASH_SOURCE[0]}")")
-if [ -f "${SCRIPT_DIR}/setup-helper-functions.sh" ]
+if [ -f "${SCRIPT_DIR}/setup-centos8.sh" ]
 then
-  source "${SCRIPT_DIR}/setup-helper-functions.sh"
+  source "${SCRIPT_DIR}/setup-centos8.sh"
 else
-  source "${SCRIPT_DIR}/../velox/scripts/setup-helper-functions.sh"
+  source "${SCRIPT_DIR}/../velox/scripts/setup-centos8.sh"
 fi
 
-export COMPILER_FLAGS=$(echo -n $(get_cxx_flags $CPU_TARGET))
+function install_presto_deps_from_package_managers {
+  dnf install -y maven java clang-tools-extra jq perl-XML-XPath
+  # This python version is installed by the Velox setup scripts
+  pip3.9 install regex pyyaml chevron black
+}
 
-(
+function install_gperf {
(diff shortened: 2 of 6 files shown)


presto-native-execution/scripts/setup-macos.sh
```diff
@@ -16,7 +16,7 @@ set -eufx -o pipefail
 # Run the velox setup script first.
 source "$(dirname "${BASH_SOURCE}")/../velox/scripts/setup-macos.sh"
 
-MACOS_DEPS="${MACOS_DEPS} bison gperf"
+MACOS_VELOX_DEPS="${MACOS_VELOX_DEPS} gperf"
```
czentgr, line 19: Do we need to include `MACOS_VELOX_DEPS`?
The sourcing of the velox setup-macos script would set this but is the idea that brew will install both velox and prestissimo deps. This would then alleviate the need to run the velox script separately (prior). I found that running only the prestissimo script will not work on the Linux because dnf or apt isn't run to install all the deps (e.g. compiler and such) when run for the first time.

Perhaps we should do this similarly for Linux to make sure such deps are installed?

majetideepak: I fixed the macos script.
The new setup scripts by default do not install anything.
We source the Velox setup scripts and install the build deps, deps from package manager, and then the packages via download.

majetideepak: > I found that running only the prestissimo script will not work on the Linux because dnf or apt isn't run to install all the deps

Can you try now?

presto-native-execution/scripts/setup-centos.sh
```diff
@@ -17,42 +17,57 @@ set -x
 export FB_OS_VERSION=v2024.04.01.00
 export RE2_VERSION=2021-04-01
 export nproc=$(getconf _NPROCESSORS_ONLN)
+source "$(dirname "${BASH_SOURCE}")/../velox/scripts/setup-centos8.sh"
 
-dnf install -y maven java python3-devel clang-tools-extra jq perl-XML-XPath
+function install_presto_build_prerequisites {
```
majetideepak, line 22: These are not build prereqs. Fix them!

majetideepak:
I verified these changes locally.

majetideepak:
@czentgr can you make another pass? Thanks!

presto-native-execution/Makefile
```diff
@@ -113,7 +113,7 @@ centos-container:			#: Build the linux container for CircleCi
 linux-container:
 	rm -rf /tmp/docker && \
 	mkdir -p /tmp/docker && \
-	cp scripts/setup-$(CONTAINER_NAME).sh scripts/$(CONTAINER_NAME)-container.dockfile velox/scripts/setup-helper-functions.sh /tmp/docker && \
+	cp scripts/setup-$(CONTAINER_NAME).sh scripts/$(CONTAINER_NAME)-container.dockfile velox/scripts/setup-centos8.sh velox/scripts/setup-helper-functions.sh /tmp/docker && \
```
kgpai, line 116: should this be velox/scripts/setup-${CONTAINER_NAME} ? ..

kgpai (commented):
Looks good one comment.

presto-native-execution/Makefile
czentgr: In prestissimo the setup script is called `setup-centos.sh` but in velox it is `setup-centos8.sh`. So it shouldn't be CONTAINER_NAME because that would be `centos` and there would be no equivalent script.
However, what if this is `ubuntu`? But also `setup-${CONTAINER_NAME}8` wouldn't work....

presto-native-execution/scripts/setup-centos.sh
```diff
@@ -17,42 +17,68 @@ set -x
...
 
-(
+function install_gperf {
   wget http://ftp.gnu.org/pub/gnu/gperf/gperf-3.1.tar.gz &&
   tar xvfz gperf-3.1.tar.gz &&
   cd gperf-3.1 &&
   ./configure --prefix=/usr/local/gperf/3_1 &&
   make "-j$(nproc)" &&
   make install &&
   ln -s /usr/local/gperf/3_1/bin/gperf /usr/local/bin/
```
czentgr, line 45: Can we include a line about removing the symlink if it exists?
I've run into the problem where the symlink exists and line 45 will fail aborting the processing.

presto-native-execution/scripts/setup-macos.sh
```diff
@@ -16,25 +16,31 @@ set -eufx -o pipefail
 # Run the velox setup script first.
 source "$(dirname "${BASH_SOURCE}")/../velox/scripts/setup-macos.sh"
 
-MACOS_DEPS="${MACOS_DEPS} bison gperf"
 export FB_OS_VERSION=v2024.04.01.00
 
-export PATH=$(brew --prefix bison)/bin:$PATH
```
czentgr, line 22: None of the prestissimo/velox deps require bison? 
On Silicon Macs this will be a problem going forward but is likely set later by the users. But removing this we don't expect issues?

presto-native-execution/scripts/setup-centos.sh
majetideepak: Good point. I will add a check if the target exists.

presto-native-execution/Makefile
majetideepak: Will fix!

presto-native-execution/scripts/setup-macos.sh
majetideepak: This is a temporary assignment on the current shell. I am thinking there should not be any issues since we don't run this script on every build and things build fine.

presto-native-execution/Makefile
czentgr: I made a comment about this but I deleted a duplikcate comment so I guess it deleted both?
Anyway I said that in prestissimo the script is called `setup-centos.sh` while in Velox it is `setup-centos8.sh`.

The CONTAINER_NAME is set to `centos`. So we'd need `setup-${CONTAINER_NAME}8` here. But this doesn't work if the container is `ubuntu`. If it is only CentOS then we are ok.

majetideepak: I am removing this target as part of the https://github.com/prestodb/presto/pull/21929.
I am thinking we can use `if-else` for now :).

majetideepak:
@czentgr, @kgpai addressed your comments.

kgpai (approved):
LGTM.

czentgr (approved):
Thank you!

presto-native-execution/scripts/setup-macos.sh
majetideepak: Also, bison is installed in the Velox setup script.

PR #22471, opened 2024-04-10
[native] Advance velox version

xiaoxmeng:
Also deprecate presto_cpp.query_memory_pool_init_capacity which has been supported in velox


tanjialiang:
@xiaoxmeng  Thanks for the cleaning up. Can we also delete the histogram counter definition in Counters.cpp?

tanjialiang (approved):

PR #22473, opened 2024-04-10
[native] Remove redundant move in PrestoToVeloxConnectorTest

amitkdutta:
With redundant move statement, Meta internal build fails with "moving a temporary object prevents copy elision" error. This compiler error is enabled by following flags:
[-Werror,-Wpessimizing-move]. Fixing here by removing the move statements.


xiaoxmeng (approved):
@amitkdutta thanks!

majetideepak (approved):

PR #22477, opened 2024-04-10
Fix duplicate key problem when querying iceberg equality deletes table

hantangwangd:
## Description

When querying an Iceberg table with equality delete files, we would meet duplicate key error in some cases when enable `iceberg.delete_as_join_rewrite_enabled`. See the test cases in `IcebergDistributedTestBase.testEqualityDeleteWithPartitionColumnMissingInSelect()`. 

The reason is we did not properly handle the duplication between identity partition column id and it's source id  when collecting `deleteSchemas`, so when constructing the `DATA_WITHOUT_EQUALITY_DELETES` tableScan node we would get duplicate column reference in it's assignments. The PR fix this bug.

## Test Plan

 - Newly added test case `IcebergDistributedTestBase.testEqualityDeleteWithPartitionColumnMissingInSelect()`

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```


jasonf20 (approved):
Nice catch and fix! This makes sense to me. 

Perhaps the dedup should happen in `createAssignmentsForUnselectedFields` since that way it's a little safer and doesn't depend as much on it's input + it keeps the delete information more in sync with the source?

However, that is a nitpick and this fix LGTM as is.

hantangwangd:
@jasonf20 thanks for the review. 

> Perhaps the dedup should happen in `createAssignmentsForUnselectedFields` since that way it's a little safer and doesn't depend as much on it's input + it keeps the delete information more in sync with the source?

The composition of the key in `deleteSchemas` involving how we think about the equality delete files with different `equalityFieldIds` on the identity partition columns in a single partition. For example in partition (a = 1) we have two equality delete files, one contains (a = 1, c = 2) and the other contains (c = 2), should they be considered to have the same delete schema?

In my opinion, I prefer to consider that they have the same delete schema, it would make things more simple as we needn't keep in mind all the time the different between a `equalityFieldId` and it's corresponding identity partition field id,  meanwhile it could also reduce the level of the join tree. What do you think?

jasonf20:
@hantangwangd My initial instinct was to keep the delete schemas separate as they are declared differently by the user/producer. However, as you mentioned, if we merge them it will reduce the number of joins so for performance reasons it's probably better that way. I agree with you 👍

hantangwangd:
@jasonf20 I have refactored the construction and file scan logic for `EQUALITY_DELETES` table to support the `delete-schema-merging` as we discussed. And the test case has been modified to add such situations. Please take a look when convenient, any thoughts or problems please let me know, thanks!

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergUtil.java
```diff
@@ -887,12 +888,17 @@ public DeleteFile next()
             return result;
         }
 
-        private boolean shouldIncludeFile(DeleteFile file)
+        private boolean shouldIncludeFile(DeleteFile file, PartitionSpec partitionSpec)
         {
             boolean matchesPartition = !requestedPartitionSpec.isPresent() ||
                     requestedPartitionSpec.get().equals(partitionSpecsById.get(file.specId()).fields().stream().map(PartitionField::fieldId).collect(Collectors.toSet()));
```
jasonf20, line 894: nit: no need for another lookup in `partitionSpecsById` since it is now passed in as an arg

jasonf20 (approved):
LGTM.

small nit: I think it would be good to explain this in the code in a short comment when filtering out the identity columns, stating we are merging the schemas within the partition to reduce the join nesting.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergUtil.java
hantangwangd: Nice point, fixed!

hantangwangd:
> small nit: I think it would be good to explain this in the code in a short comment when filtering out the identity columns, stating we are merging the schemas within the partition to reduce the join nesting.

OK, done!

tdcmeehan (approved):

PR #22483, opened 2024-04-10
[native] Update two tests in QueryContextManagerTest.

spershin:
## Description
This is preparation for default value of Velox QueryConfig property 'max_spill_level' to become 1.
We now test default values based not on values hardcoded in presto_cpp, but on the default values of the QueryConfig itself.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/tests/QueryContextManagerTest.cpp
+++ b/presto-native-execution/presto_cpp/main/tests/QueryContextManagerTest.cpp
@@ -47,7 +47,7 @@ TEST_F(QueryContextManagerTest, nativeSessionProperties) {
   protocol::TaskId taskId = "scan.0.0.1.0";
   protocol::SessionRepresentation session{
       .systemProperties = {
-          {"native_max_spill_level", "1"},
+          {"native_max_spill_level", "2"},
           {"native_spill_compression_codec", "NONE"},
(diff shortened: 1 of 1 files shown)


presto-native-execution/presto_cpp/main/tests/QueryContextManagerTest.cpp
```diff
@@ -47,31 +47,39 @@ TEST_F(QueryContextManagerTest, nativeSessionProperties) {
   protocol::TaskId taskId = "scan.0.0.1.0";
   protocol::SessionRepresentation session{
       .systemProperties = {
-          {"native_max_spill_level", "1"},
+          {"native_max_spill_level", "2"},
```
kewang1024, line 50: Curious: Why did we change this value?

kewang1024 (approved):

gggrace14 (approved):
Looks good to me.

btw, I think in the description you mean "this change" tests default values based not on values hardcoded in presto_cpp, but on the default values of the QueryConfig itself.

presto-native-execution/presto_cpp/main/tests/QueryContextManagerTest.cpp
spershin: Default will be 1, so we need something different here than default.

PR #22486, opened 2024-04-11
[native]Fix flaky test UnsafeRowShuffleTest

tanjialiang:
Test is flaky because lingering task caused memory leak. We need to make sure task is deleted before cleanup the test environment.
```
== NO RELEASE NOTE ==
```


xiaoxmeng (approved):
@tanjialiang LGTM. Thanks!

PR #22426, opened 2024-04-05
Add UDAF noisy_approx_set_sfm_from_index_and_zeros

fazalmajid:
## Description
`noisy_approx_set_sfm()` allows the creation of a Sketch-Flip-Merge sketch, essentially a differentially private alternative to the HyperLogLog sketch. The current implementation of `noisy_approx_set_sfm()` requires a key to be passed to it, which then gets hashed. Internally it calculates a 64-bit hash using Murmur3, takes the first log2(buckets) of the hash as a bucket index and a count of trailing zeroes in the remaining bits of the hash, capped to the precision of the SFM

The new UDAF noisy_approx_set_sfm_from_index_and_zeros() added in this PR allows bypassing this and supplying the bucket index and number of zeros directly.

## Motivation and Context
The new function variant helps with data minimization in privacy-sensitive applications.

## Impact
This does not affect any existing functionality.

Added two new UDAFs `noisy_approx_set_sfm_from_index_and_zeros()`, variants of `noisy_approx_set_sfm()`:

* `noisy_approx_set_sfm_from_index_and_zeros(index bigint, zeros bigint, epsilon double, buckets int, precision int) SfmSketch`
* `noisy_approx_set_sfm_from_index_and_zeros(index bigint, zeros bigint, epsilon double, buckets int) SfmSketch` (same as above with precision defaulting to 24)

## Test Plan
Added unit test cases for the new functions and verifying the SfmSketch structures returned are the same as the existing functions.

Unit tested using:
```
mvn -Dsurefire.failIfNoSpecifiedTests=false -Dtest='*Sfm*' test -pl presto-main -am -T2C
```

## Contributor checklist

- [X] Please make sure your submission complies with our [development](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#development), [formatting](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#formatting), [commit message](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#commit-formatting-and-pull-requests), and [attribution guidelines](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#attribution).
- [X] PR description addresses the issue accurately and concisely.  If the change is non-trivial, a GitHub Issue is referenced.
- [X] Documented new properties (with its default value), SQL syntax, functions, or other functionality.
- [X] If release notes are required, they follow the [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines).
[...]

PR #20526, opened 2023-08-09
Add array_top_n function

setyven:
## Description
Add array_top_n function to return an array of top n elements, given an array.

## Test Plan

Test ran sucessfully in local machine.
```
./mvnw clean install -Dtest=TestArraySqlFunctions -Dmaven.javadoc.skip=true -T1C -fn -pl presto-main
```

## Contributor checklist

- [x] Please make sure your submission complies with our [development](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#development), [formatting](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#formatting), [commit message](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#commit-formatting-and-pull-requests), and [attribution guidelines](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#attribution).
- [ ] PR description addresses the issue accurately and concisely.  If the change is non-trivial, a GitHub Issue is referenced.
- [x] Documented new properties (with its default value), SQL syntax, functions, or other functionality.
- [x] If release notes are required, they follow the [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines).
- [x] Adequate tests were added if applicable.
- [ ] CI passed.

## Release Notes
```
== RELEASE NOTES ==

General Changes
* add :func:`array_top_n` to return an array of top N elements of a given array.

```

PR #22489, opened 2024-04-11
Add scalar UDF bitwise_trailing_zeros

fazalmajid:
## Description
Add a scalar function `bitwise_trailing_zeros(bigint) bigint` that counts the number of trailing zero least significant bits in a bigint in twos-complement notation, basically what the Java standard method [java.lang.Long.numTrainilingZeros](https://download.java.net/java/early_access/panama/docs/api/java.base/java/lang/Long.html#numberOfTrailingZeros(long)) does.

## Motivation and Context
Used in some bit-twiddling operations, and used internally by HyperLogLog (GHLL) and Sketch-Flip-Merge (SFM), also exposed in #22426 
## Impact
New function `bitwise_trailing_zeros(bigint) bitwise_trailing_zeros` similar to `bitwise_count(bigint, bigint) bigint`

## Test Plan
Added unit test to `presto-main/src/test/java/com/facebook/presto/operator/scalar/TestBitwiseFunctions.java` that incidentally also revealed a Presto parser bug (cannot parse `-9223372036854775808` as a bigint, even though it is a valid value).
## Contributor checklist

- [X] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```


mlyublena:
Naming: perhaps bitwise_trailing_zeros_count for consistency with bitwise_count (and to avoid confusing it with "add zeros")

kaikalur:
I also think we should be a bit careful and not add gratuitous udfs that are extremely limited in scope - especially as native implementation (because of velox). SQL UDFs are fine though. See if you can make it a SQL udf

steveburnett (changes requested):
I see the checkbox is checked for "Documented new properties (with its default value), SQL syntax, functions, or other functionality.", but I don't see documentation for the new function.

yingsu00 (commented):
Code looks good, please add documentations. Thank you.

fazalmajid:
@ yingsu00 I added basic documentation of the function to the Sphinx docs for bitwise functions.

Regarding the other comments (no optimized implementation in Velox, utility of yet another bitwise function), this function is primarily useful for testing purposes, as the number of trailing zeros is used internally by the HyperLogLog and Sketch-Flip-merge approximate unique count algorithms, and is really hard to calculate otherwise than a UDF in Presto. I perfectly understand if this is deemed non-essential by the reviewers.

presto-docs/src/main/sphinx/functions/bitwise.rst
```diff
@@ -51,6 +51,16 @@ Bitwise Functions
         SELECT bitwise_arithmetic_shift_right(-8, 2); -- -2
         SELECT bitwise_arithmetic_shift_right(7, 2); -- 1
 
+.. function:: bitwise_trailing_zeros(x) -> bigint
+
+    Count the number of trailing zero bits set in ``x`` (treated as
+    ``bits``-bit signed integer) in 2's complement big-endian representation::
```
steveburnett, line 57: Where is ``bits`` from here?

fazalmajid: Sorry, that was a copy-paste leftover from `bit_count(x, bits)`. This function uses a Java `long` and bits is always 64.

steveburnett (approved):
LGTM! (docs)

elharo (commented):
Is there a reason this is a UDF and not just a function?

fazalmajid:
> Is there a reason this is a UDF and not just a function?

Isn't this just a question of terminology, that if the function is incorporated in the Presto core itself, it's no longer a UDF but just a function?

elharo:
I think it's a little more than terminology. As I understand it, there is a real difference between fucntions and UDFs, and according to that what's proposed here is a function, not a UDF. (I could be wrong about that, in which case please do correct me.) If so, let's get the terminology right.

rschlussel:
> I think it's a little more than terminology. As I understand it, there is a real difference between fucntions and UDFs, and according to that what's proposed here is a function, not a UDF. (I could be wrong about that, in which case please do correct me.) If so, let's get the terminology right.

I've seen people often use "udf" colloquially to mean any kind of function.  You are right that this is a regular function and not a user defined function.

@fazalmajid this needs a release note for the added function in the pr description.  see https://github.com/prestodb/presto/pull/20526 for an example of a release note for a new function.

kaikalur:
> > I think it's a little more than terminology. As I understand it, there is a real difference between fucntions and UDFs, and according to that what's proposed here is a function, not a UDF. (I could be wrong about that, in which case please do correct me.) If so, let's get the terminology right.
> 
> I've seen people often use "udf" colloquially to mean any kind of function. You are right that this is a regular function and not a user defined function.
> 

I disagree - if by "regular function" you mean it has to be in Java/CPP. We are using SQL UDF loosely and a lot of java SQL UDF are like built-ins

> @fazalmajid this needs a release note for the added function in the pr description. see #20526 for an example of a release note for a new function.

rschlussel:
I only meant that it's a function that's shipped with the product (built in) and not a function that is defined by the user separately (e.g. using the CREATE FUNCTION syntax).  Usually when I hear the word udf, people are using it as a synonym for function, and they mean basically any kind of function that can be used with Presto, but sometimes people use it to distinguish built in functions from ones defined by users outside the core product.  I prefer to use "function" rather than "udf" when talking about functions that are part of the Presto code base to remove that ambiguity.

aditi-pandit:
@fazalmajid : Thanks for this function. Please can you create an issue to implement this function on the Velox side as well.

fazalmajid:
I am tempted to abandon this PR. I originally created this function to test the equivalence of [noisy_approx_set_sfm()](https://prestodb.io/docs/current/functions/noisy.html#noisy_approx_set_sfm-col-epsilon-buckets-precision-SfmSketch) and [noisy_approx_set_sfm_from_index_and_zeros()](https://prestodb.io/docs/current/functions/noisy.html#noisy_approx_set_sfm_from_index_and_zeros-col_index-col_zeros-epsilon-buckets-precision-SfmSketch) but it is otherwise fairly niche and primarily of interest to people who work with HyperLogLog and SfmSketch.

elharo:
That sounds about right. We've probably been a bit too lax in accepting functions into the core without considering the long-term complexity burden every new function adds. It is possible to add extension functions that do not live in this git repository and are not installed in every deployment.

PR #22509, opened 2024-04-12
Fix flaky TaskManagerTest.aggregationSpill

### LINKED ISSUE #22381: TaskManagerTest.aggregationSpill flaked with a velox runtime error
linux-build-and-unit-test failed on a pure documentation change in https://github.com/prestodb/presto/pull/22356

logs attached

[aggregationspill.txt](https://github.com/prestodb/presto/files/14818144/aggregationspill.txt)

xiaoxmeng:
Deflake this test by adding additional check if spill injection is triggered or not

Fixes #22381

### DIFF
--- a/presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp
+++ b/presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp
@@ -545,7 +545,7 @@ class TaskManagerTest : public testing::Test {
                   finalAggTaskId, false, std::nullopt, std::nullopt, cbState)
               .get();
       const int64_t spilledBytes = sumOpSpillBytes("Aggregation", *taskInfo);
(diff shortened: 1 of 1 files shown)


amitkdutta (approved):
Thanks a lot @xiaoxmeng

amitkdutta:
CC: @rschlussel @elharo

rschlussel:
Thanks for fixing this!  I'm curious, why is it non-deterministic whether spill injection is triggered?

xiaoxmeng:
> Thanks for fixing this! I'm curious, why is it non-deterministic whether spill injection is triggered?

@rschlussel the spilling will only trigger if there are more than one input vector to the final aggregation. If there is only one to the final aggregation operator, then there will be no spill even though we setup with spill injection.

PR #22520, opened 2024-04-15
Write full metrics of Delete files

agrawalreetika:
## Description
In case of Positional Deletes "file_path" in lower bound & upper bound do not have full file_path Buffer

## Motivation and Context
https://github.com/apache/iceberg/issues/10064

Also checked from Spark, it doesn't truncate Delete files (Positional) metrics.
image

## Test Plan
CURRENT -
image

image

AFTER -
image

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergFileWriterFactory.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergFileWriterFactory.java
@@ -36,6 +36,7 @@ import com.google.common.collect.ImmutableMap;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.mapred.JobConf;
+import org.apache.iceberg.MetricsConfig;
 import org.apache.iceberg.Schema;
 import org.apache.iceberg.types.Types;
 
@@ -110,11 +111,12 @@ public class IcebergFileWriterFactory
             JobConf jobConf,
             ConnectorSession session,
             HdfsContext hdfsContext,
-            FileFormat fileFormat)
+            FileFormat fileFormat,
+            MetricsConfig metricsConfig)
     {
         switch (fileFormat) {
             case PARQUET:
-                return createParquetWriter(outputPath, icebergSchema, jobConf, session, hdfsContext);
+                return createParquetWriter(outputPath, icebergSchema, jobConf, session, hdfsContext, metricsConfig);
             case ORC:
                 return createOrcWriter(outputPath, icebergSchema, jobConf, session);
         }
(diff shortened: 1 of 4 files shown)


tdcmeehan:
Do we need to do the same for ORC?

hantangwangd:
> Do we need to do the same for ORC?

Seems we do not use Iceberg's metrics for ORC file. Referring to https://github.com/prestodb/presto/blob/master/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergOrcFileWriter.java#L92 . So the metrics for columns of `String` type is enforced by `OrcWriterOptions.maxStringStatisticsLimit` whose default value is `64B`. Once this threshold is exceeded, the min/max value would not be saved.

hantangwangd:
Does the absent of `min/max` metrics value about `file_path` in ORC file impact the `Delete File pruning based on the file_path` functionality you mentioned in https://github.com/apache/iceberg/issues/10064 ? If so, should we do some special treatment for deleteFile writers in orc? For example set a more larger `OrcWriterOptions.maxStringStatisticsLimit` compare to a file path. The default threshold `64B` is a little small for a file path, so it's very likely that we couldn't get the metrics.

hantangwangd:
@agrawalreetika OK, got it. As I understand, the pruning your plan to do is an optional optimization when the metrics is available, and would just be skipped when they are absent, right?

As for ORC, maybe it's better to use Iceberg's metrics when writing just as Parquet does, so that their behaviors would be more consistent. If it's feasible, the modification would be in a subsequent PR.

So, this change looks good to me. Thanks for the fix!

hantangwangd (approved):

tdcmeehan (approved):

yingsu00 (approved):

agrawalreetika:
To add additional details here, 
I did some more experiments and explored the Iceberg code. So after writing full path metrics now, we started getting the advantage of mapping/pruning deleted files to data files based on `file_path` in upper bound and lower bound which is present in Iceberg [here](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/util/ContentFileUtil.java#L52)
So we should also include these `file_path` metrics in the ORC format as well.

Currently, Presto writes one positional delete file for each base data file that is included in the DELETE operation, but Spark has [rewrite_position_deletes](https://iceberg.apache.org/docs/latest/spark-procedures/#rewrite_position_delete_files) procedure which allows combining DELETE files. So in this case single DELETE file could have entries for multiple data files.
But currently, I found this issue that after running this procedure it's not writing lowerbound & upperbound metrics for DELETE Object so I have created an issue [10146](https://github.com/apache/iceberg/issues/10146)

Once we start getting metrics in case of DELETE file having multiple data files, we could have logic of delete file pruning for this case since it's not handled in Iceberg [here](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/util/ContentFileUtil.java#L52)

PR #22524, opened 2024-04-15
Use Arrays.hashCode to hash array fields

elharo:
## Description
Use Arrays.hashCode to hash array fields

## Motivation and Context
Objects.hash does not account for arrays

## Impact
Hash codes are more reproducible.

## Test Plan
ci

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-common/src/main/java/com/facebook/presto/common/predicate/TupleDomainFilter.java
+++ b/presto-common/src/main/java/com/facebook/presto/common/predicate/TupleDomainFilter.java
@@ -604,7 +604,7 @@ public interface TupleDomainFilter
         @Override
         public int hashCode()
         {
-            return Objects.hash(min, max, size, containsEmptyMarker, hashTable, nullAllowed);
+            return Objects.hash(min, max, size, containsEmptyMarker, Arrays.hashCode(hashTable), nullAllowed);
         }
 
         @Override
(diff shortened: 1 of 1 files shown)


presto-common/src/main/java/com/facebook/presto/common/predicate/TupleDomainFilter.java
```diff
@@ -604,7 +604,7 @@ public boolean equals(Object o)
         @Override
         public int hashCode()
         {
-            return Objects.hash(min, max, size, containsEmptyMarker, hashTable, nullAllowed);
+            return Objects.hash(min, max, size, containsEmptyMarker, Arrays.hashCode(hashTable), nullAllowed);
```
tdcmeehan, line 607: This might be intentional, as it might be safe to simply use the identity hash code for the array, and there might be performance implications for hashing the entire array.  If this is the case, I think it would be better to use `identityHashCode` here to explicitly call out that we are intentionally using the reference as the hash code, not the contents of the array.  Can you check how these are hashed?  Also tagging @mbasmanova who wrote the code--do you remember if this was intentional?

rschlussel (approved):

presto-common/src/main/java/com/facebook/presto/common/predicate/TupleDomainFilter.java
elharo: Note that a few lines above in `equals` we do

 Arrays.equals(hashTable, that.hashTable)

so consistency with equals requires doing element by element hash here.

elharo: I'm not sure if we ever usage. Working though the code I only see one non-test usage of the BigintValuesUsingHashTable class.

tdcmeehan (approved):

PR #22533, opened 2024-04-16
[native] Drop const from methods in PeriodicMemoryChecker

tanjialiang:
Some system memory fetching methods like SIGAR require to keep some instance open and reuse the local connection to avoid initialization overhead. We need to drop the const to allow this to happen.
```
== NO RELEASE NOTE ==
```


xiaoxmeng (approved):
@tanjialiang thanks!

PR #22539, opened 2024-04-16
Fix analyzer for lambda in aggregation

feilong-liu:
## Description
Fix issue https://github.com/prestodb/presto/issues/22558
Fix query compilation error for lambda function in aggregation.
Sample query
```
SELECT
    id,
    REDUCE_AGG(value, 0, (a, b) -> a + b + 0, (a, b) -> a + b)
FROM (
    VALUES
        (1, 2),
        (1, 3),
        (1, 4),
        (2, 20),
        (2, 30),
        (2, 40)
) AS t(id, value)
GROUP BY
    id
```
It will throw error `(GENERIC_INTERNAL_ERROR) no type found for symbol 'expr_10'`
This is because the 0 inside the lambda was rewritten to symbol 'expr_10'. The fix is to skip rewrite for constants within lambda function.

## Motivation and Context
Fix bug.

## Impact
Fix compilation error for query

## Test Plan
Unit tests

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Fix compilation error for queries with lambda in aggregation function
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/sql/analyzer/ExpressionTreeUtils.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/analyzer/ExpressionTreeUtils.java
@@ -199,8 +199,12 @@ public final class ExpressionTreeUtils
             tempExpression = ((Cast) tempExpression).getExpression();
         }
 
-        if (tempExpression instanceof Literal || tempExpression instanceof ArrayConstructor) {
+        if (tempExpression instanceof Literal) {
             return true;
+        }
+
+        if (tempExpression instanceof ArrayConstructor) {
+            return ((ArrayConstructor) tempExpression).getValues().stream().allMatch(ExpressionTreeUtils::isConstant);
         }
 
         // ROW an MAP are special so we explicitly do that here.
--- a/presto-main/src/main/java/com/facebook/presto/sql/planner/TranslationMap.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/planner/TranslationMap.java
@@ -40,6 +40,7 @@ import java.util.Optional;
 import static com.facebook.presto.common.type.TypeUtils.isEnumType;
 import static com.facebook.presto.sql.analyzer.ExpressionTreeUtils.createSymbolReference;
 import static com.facebook.presto.sql.analyzer.ExpressionTreeUtils.getNodeLocation;
+import static com.facebook.presto.sql.analyzer.ExpressionTreeUtils.isConstant;
 import static com.facebook.presto.sql.analyzer.ExpressionTreeUtils.resolveEnumLiteral;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkState;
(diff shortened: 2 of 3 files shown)


presto-main/src/main/java/com/facebook/presto/sql/analyzer/ExpressionTreeUtils.java
```diff
@@ -199,10 +199,14 @@ public static boolean isConstant(Expression expression)
             tempExpression = ((Cast) tempExpression).getExpression();
         }
 
-        if (tempExpression instanceof Literal || tempExpression instanceof ArrayConstructor) {
```
feilong-liu, line 202: The elements in array constructor should also be Literals

presto-main/src/main/java/com/facebook/presto/sql/planner/TranslationMap.java
```diff
@@ -119,19 +120,27 @@ public Expression rewrite(Expression expression)
...
 
                 Expression translated = expressionToExpressions.getOrDefault(node, node);
                 return treeRewriter.defaultRewrite(translated, context);
             }
-        }, mapped);
+
+            @Override
+            public Expression rewriteLambdaExpression(LambdaExpression node, Boolean context, ExpressionTreeRewriter<Boolean> treeRewriter)
+            {
+                Expression result = super.rewriteLambdaExpression(node, true, treeRewriter);
```
feilong-liu, line 140: Context set to true for lambda expression

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java
```diff
@@ -7514,4 +7514,12 @@ public void testGuardConstraintFramework()
         assertQuery("select orderkey from (select * from (select * from orders where 1=0)) group by rollup(orderkey)",
                 "values (null)");
     }
+
+    @Test
+    public void testLambdaInAggregation()
```
feilong-liu, line 7519: These two will fail without fix here

presto-main/src/main/java/com/facebook/presto/sql/analyzer/ExpressionTreeUtils.java
rschlussel: Can you add a test for this?

feilong-liu: Added one more test for this in abstracttestqueries

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java
```diff
@@ -7514,4 +7514,14 @@ public void testGuardConstraintFramework()
...
+    {
+        assertQuery("SELECT id, reduce_agg(value, 0, (a, b) -> a + b+0, (a, b) -> a + b) FROM ( VALUES (1, 2), (1, 3), (1, 4), (2, 20), (2, 30), (2, 40) ) AS t(id, value) GROUP BY id", "values (1, 9), (2, 90)");
+        assertQuery("SELECT id, reduce_agg(value, 's', (a, b) -> concat(a, b, 's'), (a, b) -> concat(a, b, 's')) FROM ( VALUES (1, '2'), (1, '3'), (1, '4'), (2, '20'), (2, '30'), (2, '40') ) AS t(id, value) GROUP BY id",
+                "values (1, 's2s3s4s'), (2, 's20s30s40s')");
+        assertQueryFails("SELECT id, reduce_agg(value, array[id, value], (a, b) -> a || b, (a, b) -> a || b) FROM ( VALUES (1, 2), (1, 3), (1, 4), (2, 20), (2, 30), (2, 40) ) AS t(id, value) GROUP BY id",
```
feilong-liu, line 7524: This is for the change of isConstant(Expression expression) function. Before change "array[id, value]" is considered constant and this query pass. Now it will throw exception.

presto-main/src/main/java/com/facebook/presto/sql/analyzer/ExpressionTreeUtils.java
feilong-liu: This util is used for the reduce_agg function, and it will change the behaviour for the reduce_agg function.
For example, currently the query 
`SELECT id, reduce_agg(value, array[id, value], (a, b) -> a || b, (a, b) -> a || b) FROM ( VALUES (1, 2), (1, 3), (1, 4), (2, 20), (2, 30), (2, 40) ) AS t(id, value) GROUP BY id` will succeed as it considers expression `array[id, value]` as constant.
However, after this fix, it will fail with error `REDUCE_AGG only supports non-NULL literal as the initial value` as `array[id, value]` is not considered as constant now.
cc @kaikalur

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java
rschlussel: @feilong-liu explained to me that the reason we want this to fail is because semantically it doesn't make sense for the initial state of the lambda function to depend on the value of a column (because which row of the column is it even talking about? It wouldn't even be consistent within a query because on any given worker it would depend on which row it happened to read first)

PR #22560, opened 2024-04-18
Remove redundant argument check

elharo:
## Description
target can't be null

## Motivation and Context
simplification

## Impact
none

## Test Plan
CI

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/sql/planner/plan/TableFinishNode.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/planner/plan/TableFinishNode.java
@@ -67,7 +67,6 @@ public class TableFinishNode
     {
         super(sourceLocation, id, statsEquivalentPlanNode);
 
-        checkArgument(target != null || source instanceof TableWriterNode);
         this.source = requireNonNull(source, "source is null");
         this.target = requireNonNull(target, "target is null");
         this.rowCountVariable = requireNonNull(rowCountVariable, "rowCountVariable is null");


presto-main/src/main/java/com/facebook/presto/sql/planner/plan/TableFinishNode.java
```diff
@@ -67,7 +67,6 @@ public TableFinishNode(
     {
         super(sourceLocation, id, statsEquivalentPlanNode);
 
-        checkArgument(target != null || source instanceof TableWriterNode);
```
tdcmeehan, line 70: I don't think you should be removing the second part of this assertion.

tdcmeehan (changes requested):

presto-main/src/main/java/com/facebook/presto/sql/planner/plan/TableFinishNode.java
elharo: This is a tricky one. Two cases:

1. If target is null, it throws a NullPointerException on line 72
2. If target is not null, the `||` means that source does not have to be an instance of TableWriterNode, so we don't want to make that check.

I.e. removing the first but not the second condition would be a behavior change that makes this method throw where it previously did not.

elharo: It is possible the existing code is buggy. In particular, maybe it shouldn't be making the null check on line 72. What do you think?

tdcmeehan: @pranjalssh can you help us understand what the intention behind this check was?

tdcmeehan (approved):
I'll approve, but I think you should check with @pranjalssh to double check what the intention was on this.

PR #22562, opened 2024-04-18
Make number of preferred nodes configurable

arhimondr:
## Description

Allow changing number of preferred nodes when soft affinity scheduling is enabled

## Motivation and Context

The default number of `2` may not always be sufficient. We've seen cases when there are too many fallbacks to "random" nodes.

## Test Plan

CI

```
== RELEASE NOTES ==

General Changes
* Add `node-scheduler.max-preferred-nodes` configuration property to allow changing number of preferred nodes when soft affinity scheduling is enabled. :pr:`22562`
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/execution/scheduler/ConsistentHashingNodeProvider.java
+++ b/presto-main/src/main/java/com/facebook/presto/execution/scheduler/ConsistentHashingNodeProvider.java
@@ -15,7 +15,6 @@ package com.facebook.presto.execution.scheduler;
 
 import com.facebook.presto.metadata.InternalNode;
 import com.facebook.presto.spi.HostAddress;
-import com.facebook.presto.spi.NodeProvider;
 import com.google.common.collect.ImmutableList;
 import com.google.common.hash.HashFunction;
 
@@ -34,7 +33,6 @@ import static java.lang.String.format;
 import static java.util.Objects.requireNonNull;
 
 public class ConsistentHashingNodeProvider
-        implements NodeProvider
 {
     private static final HashFunction HASH_FUNCTION = murmur3_32();
     private final NavigableMap<Integer, InternalNode> candidates;
@@ -57,7 +55,6 @@ public class ConsistentHashingNodeProvider
         this.nodeCount = nodeCount;
     }
 
-    @Override
     public List<HostAddress> get(String key, int count)
     {
         if (count > nodeCount) {
--- a/presto-main/src/main/java/com/facebook/presto/execution/scheduler/NodeMap.java
+++ b/presto-main/src/main/java/com/facebook/presto/execution/scheduler/NodeMap.java
@@ -24,8 +24,6 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 
-import static java.lang.String.format;
-
 public class NodeMap
 {
     private final Map<String, InternalNode> activeNodesByNodeId;
@@ -92,15 +90,12 @@ public class NodeMap
         return allNodesByHostAndPort;
     }
 
-    public NodeProvider getActiveNodeProvider(NodeSelectionHashStrategy nodeSelectionHashStrategy)
+    public NodeProvider getNodeProvider(int nodeCount)
     {
-        switch (nodeSelectionHashStrategy) {
-            case MODULAR_HASHING:
-                return new ModularHashingNodeProvider(activeNodes);
-            case CONSISTENT_HASHING:
-                return consistentHashingNodeProvider.get();
-            default:
-                throw new IllegalArgumentException(format("Unknown NodeSelectionHashStrategy: %s", nodeSelectionHashStrategy));
+        if (consistentHashingNodeProvider.isPresent()) {
+            return (key) -> consistentHashingNodeProvider.get().get(key, nodeCount);
         }
+        ModularHashingNodeProvider modularHashingNodeProvider = new ModularHashingNodeProvider(allNodes);
+        return (key) -> modularHashingNodeProvider.get(key, nodeCount);
     }
 }
--- a/presto-main/src/main/java/com/facebook/presto/execution/scheduler/nodeSelection/SimpleNodeSelector.java
+++ b/presto-main/src/main/java/com/facebook/presto/execution/scheduler/nodeSelection/SimpleNodeSelector.java
@@ -18,10 +18,8 @@ import com.facebook.presto.execution.NodeTaskMap;
 import com.facebook.presto.execution.RemoteTask;
 import com.facebook.presto.execution.scheduler.BucketNodeMap;
 import com.facebook.presto.execution.scheduler.InternalNodeInfo;
-import com.facebook.presto.execution.scheduler.ModularHashingNodeProvider;
 import com.facebook.presto.execution.scheduler.NodeAssignmentStats;
 import com.facebook.presto.execution.scheduler.NodeMap;
-import com.facebook.presto.execution.scheduler.NodeSelectionHashStrategy;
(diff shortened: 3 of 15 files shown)


presto-main/src/main/java/com/facebook/presto/execution/scheduler/nodeSelection/SimpleNodeSelector.java
```diff
@@ -163,10 +159,6 @@ public SplitPlacementResult computeAssignments(Set<Split> splits, List<RemoteTas
                     preferredNodeCount = OptionalInt.of(candidateNodes.size());
                     break;
                 case SOFT_AFFINITY:
-                    // Using all nodes for soft affinity scheduling with modular hashing because otherwise temporarily down nodes would trigger too much rehashing
-                    if (nodeSelectionHashStrategy == MODULAR_HASHING) {
-                        nodeProvider = new ModularHashingNodeProvider(nodeMap.getAllNodes());
-                    }
```
hantangwangd, line 169: Is it intentional to use `nodeMap.allNodes` to build `ModularHashingNodeProvider` for all split nodeSelectionStrategies including `HARD_AFFINITY` and `NO_PREFERENCE` besides `SOFT_AFFINITY`? As I understand the whole change would affect them as well, is that OK?

hantangwangd (commented):
Overall looks good to me. Just a little question for discussing. Besides, should we add docs and release not for the config parameter `node-scheduler.max-preferred-nodes`?

presto-main/src/main/java/com/facebook/presto/execution/scheduler/nodeSelection/SimpleNodeSelector.java
arhimondr: Great question.

`nodeProvider` is only used when there is an affinity preference (`HARD_AFFINITY`, `SOFT_AFFINITY`). When there is no affinity preference (`NO_PREFERENCE`) the `nodeProvider` is not used and random nodes are selected via `randomNodeSelection.pickNodes(split)`.

For `SOFT_AFFINITY` when the `MODULAR_HASHING` strategy is selected nothing changes.

The only place when a potential change is possible is when `HARD_AFFINITY` is enabled, however I don't think it is the case in practice.

When `HARD_AFFINITY` is enabled the connector should return a set of "hard" allocated nodes. Meaning that a split is only allowed to be scheduled on that very specific set of nodes. Hence in theory a connector should never consult a `nodeProvider` when `HARD_AFFINITY` is required, as the `nodeProvider` does not guarantee any specific nodes to be selected. Based on what I checked it's always the case. All connectors that call `NodeProvider#get` only do so when `SOFT_AFFINITY` is preferred:

- https://github.com/prestodb/presto/blob/ff5282e221b461d4e9f9662c0f264cba09211f84/presto-hudi/src/main/java/com/facebook/presto/hudi/HudiSplit.java#L112
- https://github.com/prestodb/presto/blob/ff5282e221b461d4e9f9662c0f264cba09211f84/presto-delta/src/main/java/com/facebook/presto/delta/DeltaSplit.java#L133
- https://github.com/prestodb/presto/blob/ff5282e221b461d4e9f9662c0f264cba09211f84/presto-hive/src/main/java/com/facebook/presto/hive/HiveSplit.java#L170
- https://github.com/prestodb/presto/blob/ff5282e221b461d4e9f9662c0f264cba09211f84/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergSplit.java#L146

However I agree, the interface today makes it very obscure. I guess a better interface could've been a simple `ConnectorSplit#getSoftAffinityNodeSelectionKey`. We might want to consider a follow up refactor changing the interface to make it more explicit and easier to follow.

arhimondr:
Thank you @steveburnett . Let me also work on a change to the docs.

presto-main/src/main/java/com/facebook/presto/execution/scheduler/ConsistentHashingNodeProvider.java
```diff
@@ -34,7 +33,6 @@
 import static java.util.Objects.requireNonNull;
 
 public class ConsistentHashingNodeProvider
-        implements NodeProvider
```
NikhilCollooru, line 37: why are we removing this ? so that we can plugin the nodeCount preference dynamically .. seems hacky

presto-main/src/main/java/com/facebook/presto/execution/scheduler/NodeMap.java
```diff
@@ -92,15 +90,12 @@ public SetMultimap<HostAddress, InternalNode> getAllNodesByHostAndPort()
...
     {
-        switch (nodeSelectionHashStrategy) {
-            case MODULAR_HASHING:
-                return new ModularHashingNodeProvider(activeNodes);
-            case CONSISTENT_HASHING:
-                return consistentHashingNodeProvider.get();
-            default:
-                throw new IllegalArgumentException(format("Unknown NodeSelectionHashStrategy: %s", nodeSelectionHashStrategy));
+        if (consistentHashingNodeProvider.isPresent()) {
+            return (key) -> consistentHashingNodeProvider.get().get(key, nodeCount);
```
NikhilCollooru, line 96: This feels little hacky ..to plugin the node count dynamically.

NikhilCollooru (approved):
Apart from the comment i made earlier, the changes look okay to me.

presto-main/src/main/java/com/facebook/presto/execution/scheduler/nodeSelection/SimpleNodeSelector.java
hantangwangd: Thank you for the detailed message. Now I get that in practice, the situation is indeed as you said. And I agree with you that a follow-up refactor to make the interface more explicit would be great. LGTM, thanks!

presto-main/src/main/java/com/facebook/presto/execution/scheduler/ConsistentHashingNodeProvider.java
arhimondr: I'm actually not sure if `NodeProvider` interface is necessary. This interface seems a little counterintuitive to me. A `NodeProvide` is passed to the split just to be called with a key. I wonder if a better interface would be to have something like `ConnectorSplit#getSoftAffinityNodeSelectionKey` that returns a String (or an object) similar to what is done for fragment level caching with `ConnectorSplit#getSplitIdentifier`. Then a scheduler can make a decision on it's own of how many preferred nodes are required. Thoughts?

arhimondr: Another thing what this PR changes fundamentally, it moves the number of preferred nodes selection from a connector to scheduler. Previously it was a connector responsibility to request a certain number of preferred nodes. Now the scheduler will decide. 

From a practicality standpoint it feels like it's easier to make this selection on coordinator (a single property for every connector). But I wonder if there was some deeper thought process behind this that motivated to implement it this way.

@NikhilCollooru can you think of a situation when we think it would actually be preferred for a connector to chose of how many soft affinity nodes to chose for a certain split?

NikhilCollooru: yeah i dont see a usecase where you configure the node count based on connector. 
so moving the logic to scheduler seems the better way to do it.

arhimondr: Sounds good. Let me follow up with a refactor PR.

PR #22570, opened 2024-04-19
[codemod] Remove unused variables in github/presto-trunk/presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp

r-barnes:
Summary:
LLVM-15 has a warning `-Wunused-but-set-variable` which we treat as an error because it's so often diagnostic of a code issue. Unused variables can compromise readability or, worse, performance.

This diff either (a) removes an unused variable and, possibly, it's associated code, or (b) qualifies the variable with `[[maybe_unused]]`, mostly in cases where the variable _is_ used, but, eg, in an `assert` statement that isn't present in production code.

 - If you approve of this diff, please use the "Accept & Ship" button :-)

Reviewed By: palmje

Differential Revision: D56253787

### DIFF
--- a/presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp
+++ b/presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp
@@ -158,8 +158,6 @@ void setAggregationSpillConfig(
   queryConfigs.emplace(core::QueryConfig::kAggregationSpillEnabled, "true");
 }
 
-static const uint64_t kGB = 1024 * 1024 * 1024ULL;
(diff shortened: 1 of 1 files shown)


amitkdutta (changes requested):
@r-barnes Lets follow [PrestoDB commit guidelines](https://github.com/prestodb/presto/blob/master/CONTRIBUTING.md) and remove any meta specific message (e.g. `D56253787`, `Commit and Ship`) button. Please create a PR in this repo, instead of exporting directly from Meta internal tools.

PR #22587, opened 2024-04-22
Delay Presto map hash table build when lookups are few

pranjalssh:
Add session property `map_lookups_without_hashtable`, which denotes how many map lookups will be done using linear search, before resorting to hash table build of the entire map. Default is 0.

Usually hashing cost is high enough that its best to not build the hash table for map objects. Only when number of lookups exceed some threshold, we should resort to build the hashtable. This change controls that using a session prop.

```
== RELEASE NOTES ==

General Changes
* Add session property `map_lookups_without_hashtable`, which denotes how many map lookups will be done using linear search, before resorting to hash table build of the entire map. Default is 0.

### DIFF
--- a/presto-common/src/main/java/com/facebook/presto/common/block/SingleMapBlock.java
+++ b/presto-common/src/main/java/com/facebook/presto/common/block/SingleMapBlock.java
@@ -40,6 +40,7 @@ public class SingleMapBlock
     private final int offset;
     private final int positionCount;  // The number of keys in this single map * 2
     private final AbstractMapBlock mapBlock;
+    private int mapLookups;
 
     SingleMapBlock(int positionInMap, int offset, int positionCount, AbstractMapBlock mapBlock)
     {
@@ -171,9 +172,24 @@ public class SingleMapBlock
     /**
      * @return position of the value under {@code nativeValue} key. -1 when key is not found.
      */
-    public int seekKey(Object nativeValue, MethodHandle keyNativeHashCode, MethodHandle keyBlockNativeEquals, MethodHandle keyBlockHashCode)
+    public int seekKey(Object nativeValue, MethodHandle keyNativeHashCode, MethodHandle keyBlockNativeEquals, MethodHandle keyBlockHashCode, int mapLookupsWithoutHashTable)
     {
         if (positionCount == 0) {
+            return -1;
+        }
+
+        mapLookups += 1;
+        if (mapLookups <= mapLookupsWithoutHashTable) {
+            for (int i = 0; i < positionCount / 2; i++) {
+                try {
+                    if ((Boolean) keyBlockNativeEquals.invoke(mapBlock.getRawKeyBlock(), offset / 2 + i, nativeValue)) {
+                        return 2 * i + 1;
+                    }
+                }
+                catch (Throwable throwable) {
+                    throw handleThrowable(throwable);
+                }
+            }
             return -1;
         }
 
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: presto-common/src/main/java/com/facebook/presto/common/block/SingleMapBlock.java [hunk 1]
```diff
@@ -171,12 +172,27 @@ int[] getHashTable()
...
 
+        mapLookups += 1;
+        if (mapLookups <= mapLookupsWithoutHashTable) {
+            for (int i = 0; i < positionCount / 2; i++) {
+                try {
+                    if ((Boolean) keyBlockNativeEquals.invoke(mapBlock.getRawKeyBlock(), offset / 2 + i, nativeValue)) {
+                        return 2 * i + 1;
+                    }
+                }
+                catch (Throwable throwable) {
```
elharo, line 189: Per Effective Java, you should almost never catch Throwable. What's the more specific exception here?

presto-common/src/main/java/com/facebook/presto/common/block/SingleMapBlock.java
```diff
@@ -218,12 +234,27 @@ public int seekKey(Object nativeValue, MethodHandle keyNativeHashCode, MethodHan
...
 
+        mapLookups += 1;
+        if (mapLookups <= mapLookupsWithoutHashTable) {
+            for (int i = 0; i < positionCount / 2; i++) {
+                try {
+                    if ((Boolean) keyBlockNativeEquals.invokeExact(mapBlock.getRawKeyBlock(), offset / 2 + i, nativeValue)) {
+                        return 2 * i + 1;
+                    }
+                }
+                catch (Throwable throwable) {
```
elharo, line 251: ditto

presto-common/src/main/java/com/facebook/presto/common/block/SingleMapBlock.java
```diff
@@ -306,12 +352,27 @@ public int seekKeyExact(boolean nativeValue, MethodHandle keyNativeHashCode, Met
...
```
elharo, line 369: ditto

presto-common/src/main/java/com/facebook/presto/common/block/SingleMapBlock.java
```diff
@@ -350,12 +411,27 @@ public int seekKeyExact(double nativeValue, MethodHandle keyNativeHashCode, Meth
...
```
elharo, line 428: ditto

presto-common/src/main/java/com/facebook/presto/common/block/SingleMapBlock.java
```diff
@@ -401,12 +477,27 @@ public int seekKeyExact(Slice nativeValue, MethodHandle keyNativeHashCode, Metho
...
```
elharo, line 494: ditto

elharo (changes requested):

steveburnett:
* Please add documentation for the session property, and then I'll be happy to review this PR. 

* Thanks for writing a release note entry! Consider the following small suggestion adding your PR# in the entry to make easier the work of assembling the release note for the next release:

```
== RELEASE NOTES ==

General Changes
* Add session property `map_lookups_without_hashtable`, which denotes how many map lookups will be done using linear search, before resorting to hash table build of the entire map. Default is 0. :pr:`22587`
```

aditi-pandit:
@pranjalssh, @mbasmanova : Is there an equivalent for this in Prestissimo engine ? My intuition is that because of runtime optimizations in HashTable build we might not need this. But just thought to ask anyways.

Issue #22543, opened 2024-04-17
Be able to query Delta tables generated with the new Delta protocols


mblanco-denodo:
With the current state of the delta connector, you cannot query tables created against delta data that have been created with recent protocol versions (reading 3, writing 7). Currently it only allows querying delta tables that have been created with support for version 1 of the reading protocol and version 2 of the writing protocol.

The error that presto reports while reading the table is:

> Delta protocol version (3,7) is too new for this version of Delta Standalone Reader/Writer (1,2). Please upgrade to a newer release.' Received exception with message 'Delta protocol version (3,7) is too new for this version of Delta Standalone Reader/Writer (1,2). Please upgrade to a newer release.'

## Expected Behavior or Use Case

The query should return data

## Presto Component, Service, or Connector

Delta connector

## Possible Implementation

The connector has to use the new API Delta-Kernel, as the Delta-Standalone is being left back:
https://github.com/delta-io/delta/issues/2243
https://github.com/delta-io/delta/issues/2254
However, there are certain features that will be released on the next 3.2.0 version of the Delta-Kernel library, such as the ability to move to a snapshot by ID or by timestamp (currently the library only allows you to obtain data in the latest snapshot):
https://github.com/delta-io/delta/milestone/26?closed=1

## Context

As the tools that generate the delta tables (such as databricks) are updated, the new tables are already generated with the new versions of the protocols, so it would be important to have delta support that could handle this.

Issue #21828, opened 2024-01-31
Be able to query uppercase partitioned columns on delta tables


mblanco-denodo:
In Delta tables, NULL values are returned in partitioned uppercase columns. For example, for the following Delta table in S3:
![image](https://github.com/prestodb/presto/assets/136591152/c25f95a7-8764-4dcf-a5ac-69d31d3de78a)
a query like 
`select * from delta_uppercase_partitioned_test`
returns NULL values on the partitioned field BIRTH_YEAR
![image](https://github.com/prestodb/presto/assets/136591152/c819363a-6ca5-48cf-b050-6f6ebb37bb90)

## Expected Behavior or Use Case

Presto should be able to retrieve data from uppercase partitioned columns of delta tables, as this is a very common case.
## Presto Component, Service, or Connector

Delta Connector
## Possible Implementation

Through some debugging in the Delta connector source code, this behavior could be achieved by avoiding String.toLowerCase() conversions in the DeltaClient.getSchema() method. Additionally, we could make it configurable, via a new configuration property to allow old and new behavior as desired.
`delta.case-sensitive-partitions=true`

This solution was successfully tested on my local environment.

mblanco-denodo:
As the changes are tested in my local environment, I could have a fix PR soon

imjalpreet:
Fixed with https://github.com/prestodb/presto/pull/22596

PR #22596, opened 2024-04-24
Add support to be able to query new Delta protocols

mblanco-denodo:
## Description

Use of delta-kernel-api instead of delta-standalone-api in order to be able to query delta tables in newer protocols (delta reader 3, delta writer 7)
Added new configurable parameter (``delta.case-sensitive-partitions-enabled``) to allow case sensitive partition matching, allowing to query tables with partition names in uppercase. By default it is set to ``true``.

## Motivation and Context

Solves https://github.com/prestodb/presto/issues/22543 and https://github.com/prestodb/presto/issues/21828.

## Impact

Based on some query executions over some data in S3 (times in seconds):
| Table                               |  master        |  branch |
|----------------------------|:-------------:|--------:|
| nation   (tpch100)           |  63.06          | 18.92    |
| region   (tpch100)           |    50.25        |   16.05  |
| deltatbl_partition_prune | 33.63           |    4.98   |

Also experienced huge improvement on some queries during testing. Here are the HTML reports of the tests suites in master and in the branch: 
[master_vs_branch_test_execution_performance.zip](https://github.com/user-attachments/files/16027277/master_vs_branch_test_execution_performance.zip)

## Test Plan

Test refactor so that the current tests are all executed on older and newer versions of the same test data.

## Contributor checklist

- [x] CI passed.

## Release Notes

== RELEASE NOTES ==

Delta Connector Changes
* Added support to read tables up to Delta Protocol Version 3,7 through the use of the new Delta-Kernel api. :pr:`22596`
* Added new boolean configuration parameter ``delta.case-sensitive-partitions-enabled`` to be able to query data with partitioned columns with column names in uppercase. This property is set to ``true`` by default. :pr:`22596`


mblanco-denodo:
PR set to draft while waiting for the delta kernel api to release v3.2.0.  https://github.com/delta-io/delta/milestone/26
Current limitations that will be solved on 3.2.0:
 - Move to an snapshot different from the latest.
 - Move to an snapshot by timestamp
 - Reading partition columns of TimestampType is unsupported

mblanco-denodo:
After the release of Delta Kernel 3.2.0, all limitations are resolved. Also, it seems that there is an improvement in performance reading both delta v1 and delta v3 (perfomance local tests attached in the PR description). PR is ready for review.

presto-delta/src/main/java/com/facebook/presto/delta/DeltaClient.java
```diff
@@ -70,72 +77,90 @@ public DeltaClient(HdfsEnvironment hdfsEnvironment)
...
             SchemaTableName schemaTableName,
             String tableLocation,
             Optional<Long> snapshotId,
             Optional<Long> snapshotAsOfTimestampMillis)
     {
-        Optional<DeltaLog> deltaLog = loadDeltaTableLog(session, new Path(tableLocation), schemaTableName);
-        if (!deltaLog.isPresent()) {
+        Path location = new Path(tableLocation);
+        Optional<Engine> deltaEngine = loadDeltaTableClient(session, location,
+                schemaTableName);
```
tdcmeehan, line 89:
```suggestion
        Optional<Engine> deltaEngine = loadDeltaTableClient(session, location, schemaTableName);
```

presto-delta/src/main/java/com/facebook/presto/delta/DeltaClient.java
```diff
@@ -184,27 +218,53 @@ private static Snapshot getSnapshotAsOfTimestamp(DeltaLog deltaLog, long snapsho
...
+            while (columnBatches.hasNext()) {
+                CloseableIterator<Row> rows = columnBatches.next().getRows();
+                if (rows.hasNext()) {
+                    row = rows.next();
+                    break;
+                }
+            }
+            Map<String, String> partitionValues = row != null ?
+                    InternalScanFileUtils.getPartitionValues(row) : new HashMap<>(0);
+            columnBatches.close();
```
tdcmeehan, line 248:
```suggestion
        try (CloseableIterator<FilteredColumnarBatch> columnBatches = snapshot.getScanBuilder(deltaEngine).build().getScanFiles(deltaEngine)) {
            Row row = null;
            while (columnBatches.hasNext()) {
                CloseableIterator<Row> rows = columnBatches.next().getRows();
                if (rows.hasNext()) {
                    row = rows.next();
                    break;
                }
            }
            Map<String, String> partitionValues = row != null ?
                    InternalScanFileUtils.getPartitionValues(row) : new HashMap<>(0);
            columnBatches.close();
```

tdcmeehan (changes requested):
Just some initial minor comments, great work!

steveburnett (dismissed):
A single nit of punctuation, otherwise LGTM (docs). Thanks!

agrawalreetika (changes requested):
Thanks for the PR. Had few tables created with new Protocol earlier but wasn't accessible before this new Protocol changes support. Now able to access those tables.
Dropped in few minor comments, please check!  

You mentioned about performance improvement, Just curious which part of the new changes are contributing to performance improvement while reading the table?

mblanco-denodo:
@agrawalreetika performance improvement is due to the new Kernel API.

steveburnett (dismissed):
LGTM! (docs)

presto-delta/pom.xml
```diff
@@ -297,6 +325,12 @@
             <artifactId>javax.servlet-api</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.jetbrains</groupId>
```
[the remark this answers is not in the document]
mblanco-denodo: It was used to annotate sql queries in testing. I removed the annotations and the dependency.

presto-delta/src/main/java/com/facebook/presto/delta/DeltaClient.java
mblanco-denodo: Done

presto-delta/src/main/java/com/facebook/presto/delta/DeltaClient.java
mblanco-denodo: Done

presto-delta/src/main/java/com/facebook/presto/delta/DeltaExpressionUtils.java
```diff
@@ -64,109 +70,47 @@ public static List<TupleDomain<ColumnHandle>> splitPredicate(
...
-                    }
-                }
-
-                return nextItem != null;
-            }
+        Optional<List<TupleDomain.ColumnDomain<DeltaColumnHandle>>> columnDomains = predicate.getColumnDomains();
+        List<DeltaColumnHandle> partitionColumns = columnDomains.map(domains -> domains.stream()
+                .filter(entry -> entry.getColumn().getColumnType() == PARTITION)
+                .map(TupleDomain.ColumnDomain::getColumn)
+                .collect(Collectors.toList())).orElse(Collections.emptyList());
```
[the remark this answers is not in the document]
mblanco-denodo: Done

presto-delta/src/main/java/com/facebook/presto/delta/DeltaExpressionUtils.java
```diff
@@ -180,76 +124,261 @@ private static TupleDomain<String> extractPartitionColumnsPredicate(TupleDomain<
...
+            if (!rowsRemaining) {
+                if (!inputIterator.hasNext()) {
+                    return false;
+                }
+                FilteredColumnarBatch nextFile = inputIterator.next();
+                logger.debug("Advancing input iterator");
+                row = nextFile.getRows();
+            }
+            else {
+                logger.debug("There are still rows remaining int the iterator, not advancing the iterator");
```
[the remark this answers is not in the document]
mblanco-denodo: Done

presto-docs/src/main/sphinx/connector/deltalake.rst
```diff
@@ -41,6 +41,8 @@ Property Name                                   Description
                                                 In order for this option to work, also set
                                                 ``experimental.pushdown-dereference-enabled`` to
                                                 ``true``.
+``delta.case-sensitive-partitions-enabled``     Allows matching the names of partitioned columns in a     ``true``
+                                                case-sensitive manner
```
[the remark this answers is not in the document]
mblanco-denodo: Done

presto-delta/src/main/java/com/facebook/presto/delta/DeltaClient.java
```diff
@@ -70,72 +77,90 @@ public DeltaClient(HdfsEnvironment hdfsEnvironment)
...
-            snapshot = deltaLog.get().snapshot(); // get the latest snapshot
+            try {
+                snapshot = deltaTable.getLatestSnapshot(deltaEngine.get()); // get the latest snapshot
+            }
+            catch (TableNotFoundException e) {
+                throw new PrestoException(GENERIC_INTERNAL_ERROR,
+                        format("Could not move to latest snapshot on table '%s.%s'", schemaTableName.getSchemaName(),
+                                schemaTableName.getTableName()));
+            }
         }
```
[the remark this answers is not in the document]
mblanco-denodo: Done

presto-delta/src/main/java/com/facebook/presto/delta/DeltaClient.java
```diff
@@ -70,72 +77,90 @@ public DeltaClient(HdfsEnvironment hdfsEnvironment)
...
     }
 
     /**
      * Get the list of files corresponding to the given Delta table.
      *
      * @return Closeable iterator of files. It is responsibility of the caller to close the iterator.
      */
-    public CloseableIterator<AddFile> listFiles(ConnectorSession session, DeltaTable deltaTable)
+    public CloseableIterator<FilteredColumnarBatch> listFiles(ConnectorSession session, DeltaTable deltaTable)
     {
```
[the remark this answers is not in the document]
mblanco-denodo: Done

presto-delta/src/main/java/com/facebook/presto/delta/DeltaExpressionUtils.java
```diff
@@ -180,76 +124,261 @@ private static TupleDomain<String> extractPartitionColumnsPredicate(TupleDomain<
...
+                nextRow = row.next();
+                nextItem = nextRow;
+                rowsRemaining = true;
+            }
+            if (!rowsRemaining) {
+                try {
+                    row.close();
+                }
+                catch (IOException e) {
+                    throw new GenericInternalException("Cloud not close row batch", e);
```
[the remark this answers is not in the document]
mblanco-denodo: Done

presto-delta/src/test/java/com/facebook/presto/delta/AbstractDeltaDistributedQueryTestBase.java
```diff
@@ -138,7 +186,7 @@ private static DistributedQueryRunner createDeltaQueryRunner(Map<String, String>
      * @param deltaTableName Name of the delta table which is on the classpath.
      * @param hiveTableName Name of the Hive table that the Delta table is to be registered as in HMS
      */
-    private static void registerDeltaTableInHMS(QueryRunner queryRunner, String deltaTableName, String hiveTableName)
+    protected static void registerDeltaTableInHMS(QueryRunner queryRunner, String deltaTableName, String hiveTableName)
```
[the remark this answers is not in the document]
mblanco-denodo: I intended in use it in other test class but at the end I didn't and I forgot to change the method visibility back. Reverted to `private`.

agrawalreetika:
@mblanco-denodo Could you please squash the commits?
I think here 2 commits would be good. One for changes for `delta.case-sensitive-partitions-enabled` and other for the changes in Delta protocols

mblanco-denodo:
@agrawalreetika commits squashed into two as requested:
 - changes in delta protocol from standalone to kernel
 - changes with the new property for uppercase column processing

agrawalreetika (dismissed):

presto-delta/src/main/java/com/facebook/presto/delta/DeltaClient.java
```diff
@@ -70,72 +77,101 @@ public DeltaClient(HdfsEnvironment hdfsEnvironment)
...
+        requireNonNull(deltaTable, "deltaTable is null");
         checkArgument(deltaTable.getSnapshotId().isPresent(), "Snapshot id is missing from the Delta table");
-        Optional<DeltaLog> deltaLog = loadDeltaTableLog(
-                session,
+        Optional<Engine> deltaEngine = loadDeltaTableClient(session,
                 new Path(deltaTable.getTableLocation()),
                 new SchemaTableName(deltaTable.getSchemaName(), deltaTable.getTableName()));
+        if (!deltaEngine.isPresent()) {
```
tdcmeehan, line 156: Do you know under what circumstances this error could happen?

mblanco-denodo: Right now it should not happen, as the method `loadDeltaTableClient` will return a non null `Engine` or throw an exception. The usage of Optional in that method is inherited from the previous code, and it is a good practice to check if `deltaEngine.isPresent()` as in future iterations the logic of `loadDeltaTableClient` could change (note that all the Delta Kernel API is annotated with the `@Evolving` annotation, and they are making big API changes even on minor versions).

presto-delta/src/main/java/com/facebook/presto/delta/DeltaConfig.java
```diff
@@ -46,4 +47,16 @@ public DeltaConfig setMaxSplitsBatchSize(int maxSplitsBatchSize)
         this.maxSplitsBatchSize = maxSplitsBatchSize;
         return this;
     }
+
+    public boolean isCaseSensitivePartitionsEnabled()
+    {
+        return this.caseSensitivePartitionsEnabled;
+    }
+
+    @Config("delta.case-sensitive-partitions-enabled")
```
imjalpreet, line 56: To get a little clarity, was the issue of not being able to query mixed-case/upper-case columns only happening if the column is a partitioned column? Can we also have regular columns as mixed case or upper case, and if yes, does it work as expected?

mblanco-denodo: We've experienced the problem only on partitioned columns, as the non partitioned columns are in parquet data and reading parquet the case issue does not happen. However the fix should work on both cases.

imjalpreet: The reason I wanted to get an idea is we have removed the toLowerCase conversion for all columns not just the partitioned columns. That's why wanted to make sure it does not have any side effects for regular columns and if it's required even for regular columns, should we rename the config to `delta.case-sensitive-columns-enabled`?

mblanco-denodo: I can add a test on regular columns with and without the property to be sure

mblanco-denodo: I confirm that with `delta.case-sensitive-partitions-enabled=false` uppercase non partitioned columns are properly returning the data, so I think we should leave the property name as it is because it only changes the behavior to partitioned columns. I added the new test that compares the output between a table with a non partitioned column in lowercase and other table with the same column in uppercase to assure that there is no side effect and both tables return the same data.

steveburnett (dismissed):
LGTM! (docs)
Pull updated branch, local doc build, looks good. Thanks!

presto-delta/src/main/java/com/facebook/presto/delta/DeltaConfig.java
imjalpreet: Thanks for adding the tests and confirming.

imjalpreet (approved):
LGTM 👍🏼 

@mblanco-denodo Can you please rebase on master to fix the failing test suite?

presto-delta/src/main/java/com/facebook/presto/delta/DeltaClient.java
```diff
@@ -70,72 +77,101 @@ public DeltaClient(HdfsEnvironment hdfsEnvironment)
...
+            snapshot = getSnapshotAsOfTimestamp(deltaTable, deltaEngine,
+                    snapshotAsOfTimestampMillis.get(), schemaTableName);
         }
         else {
-            snapshot = deltaLog.get().snapshot(); // get the latest snapshot
+            try {
+                snapshot = deltaTable.getLatestSnapshot(deltaEngine); // get the latest snapshot
+            }
+            catch (TableNotFoundException e) {
+                throw new PrestoException(GENERIC_INTERNAL_ERROR,
```
elharo, line 128: embed the original exception as the cause here

presto-delta/src/main/java/com/facebook/presto/delta/DeltaClient.java
```diff
@@ -70,72 +77,101 @@ public DeltaClient(HdfsEnvironment hdfsEnvironment)
...
-        Metadata metadata = snapshot.getMetadata();
-        String format = metadata.getFormat().getProvider();
-        if (!PARQUET.name().equalsIgnoreCase(format)) {
-            throw new PrestoException(DELTA_UNSUPPORTED_DATA_FORMAT,
-                    format("Delta table %s has unsupported data format: %s. Currently only Parquet data format is supported", schemaTableName, format));
+        if (snapshot instanceof SnapshotImpl) {
+            String format = ((SnapshotImpl) snapshot).getMetadata().getFormat().getProvider();
+            if (!PARQUET.name().equalsIgnoreCase(format)) {
+                throw new PrestoException(DELTA_UNSUPPORTED_DATA_FORMAT,
+                        format("Delta table %s has unsupported data format: %s. Currently only Parquet data format is supported", schemaTableName, format));
```
elharo, line 138: Currently only --> Only the

presto-delta/src/main/java/com/facebook/presto/delta/DeltaClient.java
```diff
@@ -70,72 +77,101 @@ public DeltaClient(HdfsEnvironment hdfsEnvironment)
...
-        if (!deltaLog.isPresent()) {
+        try {
+            return sourceTable.getSnapshotAsOfVersion(deltaEngine.get(),
+                            deltaTable.getSnapshotId().get()).getScanBuilder(deltaEngine.get()).build()
+                    .getScanFiles(deltaEngine.get());
+        }
+        catch (TableNotFoundException e) {
             throw new PrestoException(NOT_FOUND,
-                    format("Delta table (%s.%s) no longer exists.", deltaTable.getSchemaName(), deltaTable.getTableName()));
+                    format("Delta table not found in '%s'", deltaTable.getTableLocation()));
```
elharo, line 169: nest exception

presto-delta/src/test/java/com/facebook/presto/delta/TestDeltaScanOptimizations.java
```diff
@@ -184,15 +189,18 @@ private void assertDeltaQueryOptimized(
             Map<String, Domain> expectedConstraint,
             Map<String, Domain> expectedEnforcedConstraint)
     {
+        System.out.println(tableScanWithConstraints(tableName, expectedConstraint, expectedEnforcedConstraint).toString());
```
elharo, line 192: remove this line

presto-delta/src/main/java/com/facebook/presto/delta/DeltaClient.java
```diff
@@ -70,72 +77,101 @@ public DeltaClient(HdfsEnvironment hdfsEnvironment)
...
     {
+        requireNonNull(deltaTable, "deltaTable is null");
         checkArgument(deltaTable.getSnapshotId().isPresent(), "Snapshot id is missing from the Delta table");
-        Optional<DeltaLog> deltaLog = loadDeltaTableLog(
-                session,
+        Optional<Engine> deltaEngine = loadDeltaTableClient(session,
                 new Path(deltaTable.getTableLocation()),
                 new SchemaTableName(deltaTable.getSchemaName(), deltaTable.getTableName()));
+        if (!deltaEngine.isPresent()) {
+            throw new PrestoException(GENERIC_INTERNAL_ERROR,
```
tdcmeehan, line 157: Please throw a system error here.  GENERIC_INTERNAL_ERROR is an error for anything that is not categorized, so we shouldn't be intentionally throwing it.

presto-delta/src/main/java/com/facebook/presto/delta/DeltaClient.java
```diff
@@ -184,27 +229,51 @@ private static Snapshot getSnapshotAsOfTimestamp(DeltaLog deltaLog, long snapsho
...
+                                field.isNullable(),
+                                partitionValues.containsKey(columnName));
+                    }).collect(Collectors.toList());
+        }
+        catch (TableNotFoundException e) {
+            throw new PrestoException(NOT_FOUND,
+                    format(TABLE_NOT_FOUND_ERROR_TEMPLATE, tableName.getSchemaName(), tableName.getTableName()));
+        }
+        catch (IOException e) {
+            throw new GenericInternalException("Could not close columnar batch row", e);
```
tdcmeehan, line 276: Please consider rethrowing an UncheckedIOException

presto-delta/src/main/java/com/facebook/presto/delta/DeltaClient.java
mblanco-denodo: Done. Added new DeltaErrorCodes to candle these cases

presto-delta/src/main/java/com/facebook/presto/delta/DeltaClient.java
mblanco-denodo: Done

presto-delta/src/test/java/com/facebook/presto/delta/TestDeltaScanOptimizations.java
mblanco-denodo: Done

presto-delta/src/main/java/com/facebook/presto/delta/DeltaClient.java
mblanco-denodo: Done

presto-delta/src/main/java/com/facebook/presto/delta/DeltaClient.java
mblanco-denodo: Done

tdcmeehan (approved):

abhiseksaikia:
@mblanco-denodo  The directory structure for the tests are pretty long and its breaking our build due to internal limit on path length (<490). 
image

Can  we have a fix for this?  cc : @tdcmeehan

tdcmeehan:
This refactoring only made the directory structure slightly larger, by introducing the `delta_v1`/`delta_v3` prefix.  Otherwise it's the same directory structure, which is what is specified by the Delta format.  While we could collapse this to something smaller like `v3-partition-values` and remove that prefix directory, long term I'd recommend finding a way to increase the limit on the directory structure so you don't encounter this again in the future @abhiseksaikia.

PR #22604, opened 2024-04-24
Revert "Preserve case for RowType's field name and JSON content when ``CAST``"

neeradsomanchi:
Reverts 5b7cda2

## Description
This reverts the commit 5b7cda2, which was causing issues with user queries.

There is no flag covering the code introduced to [check for duplicate field names](https://github.com/prestodb/presto/commit/5b7cda2e8ff94fff3c96514a90f9ee5f34cbf3f0#diff-9f6a39fd0d0af9512744d070a9734ce6e7e47eb24112345890187a6016ae7e61R622-R628). This causes some query failures because it breaks legacy behavior. 

## Test Plan
Unit Tests

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```


neeradsomanchi:
@hantangwangd Can you please cover the relevant logic (in the PR description) under a flag as well so that legacy behavior can be supported?

steveburnett (approved):
LGTM! (doc)

Pulled the branch and reviewed docs changes in a local docs build.

hantangwangd:
@neeradsomanchi Sure. By the way, is this all the problems caused by the change? And could you provide some examples of the failed cases, so that I can add them to test cases for legacy behavior.

hantangwangd:
@neeradsomanchi After recheck the code, I found it a little complex to cover the code in `TypeSignature` under a config flag. It may need a significant modification to pass the config flag. And I also wonder whether we should do this. Are the failed queries is something like follows and we want it to pass when the legacy support flag is true?

```
select CAST(ROW(1, 2) AS ROW(a BIGINT, A DOUBLE)).a;
select CAST(ROW(1, 2) AS ROW(KEY1 VARCHAR, "key1" VARCHAR));
select TYPEOF(CAST(row(1, 2) AS ROW(A VARCHAR, "a" VARCHAR)));
select TYPEOF(CAST(row(1, 2) AS ROW(a VARCHAR, A VARCHAR)));
```

If so, I think we should figure out firstly whether it is a incorrect behavior that need to be fixed even in legacy semantics. Any detailed messages would be appreciated!

cc: @tdcmeehan @rschlussel @mbasmanova

rschlussel:
@hantangwangd, I checked the failures, and they were really user errors (seemed like typos) that we should have been failing all along. It was something like `CAST (row(1, 'a') AS ROW(subfield BIGINT, subfield VARCHAR)) column`.  Before your change we only failed if you tried to use the duplicated field (e.g. if you then did `SELECT column.subfield`), with your change we fail even if you don't use the field. 

Failing is the right thing to do here, and there isn't really a reasonable use case where users would use the old behavior on purpose, so it's fine if a few queries fail after this and users need to fix them.  This should be good to add back as it was, but with a release note noting this fix.

hantangwangd:
@rschlussel, thank you very much for the checking on failed user cases and the detailed informations. And so glad to know that this change do not affect normal existing user cases.

So, to add back this change, should we resubmit a PR the same as #21869 or revert the revert PR #22604? Which one do you think is the better way?

rschlussel:
either is fine.

PR #22606, opened 2024-04-24
Add HBO support for CTE materialization

feilong-liu:
## Description
The three nodes, Sequence node, cte producer node, and cte consumer node are not handled in Canonicalization for HBO. It will make HBO not working for the query. This PR adds plan canonicalization for it.

## Motivation and Context
To have HBO work for CTE materialized query

## Impact
HBO work for CTE materialized query

## Test Plan
Unit test

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Enable HBO for CTE materialized query :pr:`22606`
```

### DIFF
--- a/presto-hive/src/test/java/com/facebook/presto/hive/TestHiveHistoryBasedStatsTracking.java
+++ b/presto-hive/src/test/java/com/facebook/presto/hive/TestHiveHistoryBasedStatsTracking.java
@@ -37,6 +37,8 @@ import org.testng.annotations.Test;
 
 import java.util.Map;
 
+import static com.facebook.presto.SystemSessionProperties.CTE_MATERIALIZATION_STRATEGY;
+import static com.facebook.presto.SystemSessionProperties.CTE_PARTITIONING_PROVIDER_CATALOG;
 import static com.facebook.presto.SystemSessionProperties.HISTORY_BASED_OPTIMIZATION_PLAN_CANONICALIZATION_STRATEGY;
 import static com.facebook.presto.SystemSessionProperties.JOIN_DISTRIBUTION_TYPE;
 import static com.facebook.presto.SystemSessionProperties.PARTIAL_AGGREGATION_STRATEGY;
@@ -256,6 +258,23 @@ public class TestHiveHistoryBasedStatsTracking
         finally {
             getQueryRunner().execute("DROP TABLE IF EXISTS test_orders");
         }
+    }
+
+    @Test
+    public void testHistoryBasedStatsCalculatorCTE()
+    {
+        String sql = "with t1 as (select orderkey, orderstatus from orders where totalprice > 100), t2 as (select orderkey, totalprice from orders where custkey > 100) " +
+                "select orderstatus, sum(totalprice) from t1 join t2 on t1.orderkey=t2.orderkey group by orderstatus";
+        Session cteMaterialization = Session.builder(defaultSession())
+                .setSystemProperty(CTE_MATERIALIZATION_STRATEGY, "ALL")
+                .setSystemProperty(CTE_PARTITIONING_PROVIDER_CATALOG, "hive")
+                .build();
+        // CBO Statistics
+        assertPlan(cteMaterialization, sql, anyTree(node(ProjectNode.class, anyTree(any())).withOutputRowCount(Double.NaN)));
+
+        // HBO Statistics
+        executeAndTrackHistory(sql, cteMaterialization);
+        assertPlan(cteMaterialization, sql, anyTree(node(ProjectNode.class, anyTree(any())).withOutputRowCount(3)));
     }
 
     @Override
--- a/presto-main/src/main/java/com/facebook/presto/sql/planner/CanonicalPlanGenerator.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/planner/CanonicalPlanGenerator.java
@@ -24,6 +24,8 @@ import com.facebook.presto.spi.plan.AggregationNode;
 import com.facebook.presto.spi.plan.AggregationNode.Aggregation;
 import com.facebook.presto.spi.plan.AggregationNode.GroupingSetDescriptor;
 import com.facebook.presto.spi.plan.Assignments;
+import com.facebook.presto.spi.plan.CteConsumerNode;
+import com.facebook.presto.spi.plan.CteProducerNode;
 import com.facebook.presto.spi.plan.DistinctLimitNode;
 import com.facebook.presto.spi.plan.EquiJoinClause;
 import com.facebook.presto.spi.plan.FilterNode;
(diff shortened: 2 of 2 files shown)


presto-main/src/main/java/com/facebook/presto/sql/planner/CanonicalPlanGenerator.java
```diff
@@ -830,6 +833,25 @@ public Optional<PlanNode> visitAggregation(AggregationNode node, Context context
         return Optional.of(canonicalPlan);
     }
 
+    @Override
+    public Optional<PlanNode> visitSequence(SequenceNode node, Context context)
+    {
+        node.getCteProducers().forEach(x -> x.accept(this, context));
+        return node.getPrimarySource().accept(this, context);
```
feilong-liu, line 840: For sequence node, just return the canonicalized primary source

presto-main/src/main/java/com/facebook/presto/sql/planner/CanonicalPlanGenerator.java
```diff
@@ -830,6 +833,25 @@ public Optional<PlanNode> visitAggregation(AggregationNode node, Context context
...
+    }
+
+    @Override
+    public Optional<PlanNode> visitCteProducer(CteProducerNode node, Context context)
+    {
+        return node.getSource().accept(this, context);
```
feilong-liu, line 846: For cte producer, canonicalize its source

presto-main/src/main/java/com/facebook/presto/sql/planner/CanonicalPlanGenerator.java
```diff
@@ -830,6 +833,25 @@ public Optional<PlanNode> visitAggregation(AggregationNode node, Context context
...
+    }
+
+    @Override
+    public Optional<PlanNode> visitCteConsumer(CteConsumerNode node, Context context)
+    {
+        return node.getOriginalSource().accept(this, context);
```
feilong-liu, line 852: Canonicalize original source for cte consumer

presto-main/src/main/java/com/facebook/presto/sql/planner/CanonicalPlanGenerator.java
```diff
@@ -830,6 +833,25 @@ public Optional<PlanNode> visitAggregation(AggregationNode node, Context context
         return Optional.of(canonicalPlan);
     }
 
+    @Override
+    public Optional<PlanNode> visitSequence(SequenceNode node, Context context)
+    {
+        node.getCteProducers().forEach(x -> x.accept(this, context));
```
jaystarshot, line 839: Do we need to visit producers?

feilong-liu: Yes, it's to trigger canonicalization of plan nodes under the cte producer, which will also need HBO to optimize the subquery.

jaystarshot (approved):
Looks good, just note that the stats equivalent node of the cteProducer and cteConsumer is the actual source too, I think this stats equivalent node is also passed to the table writers and table scan nodes in the physical optimizer and so tracking should also work but the flow is not entirely clear.

presto-main/src/main/java/com/facebook/presto/sql/planner/CanonicalPlanGenerator.java
```diff
@@ -830,6 +833,25 @@ public Optional<PlanNode> visitAggregation(AggregationNode node, Context context
         return Optional.of(canonicalPlan);
     }
 
+    @Override
+    public Optional<PlanNode> visitSequence(SequenceNode node, Context context)
```
ClarenceThreepwood, line 837: Add a test case where the sequence node is also canonicalized?

feilong-liu: The end to end test in testHistoryBasedStatsCalculatorCTE should suffice to make sure that it works.

presto-hive/src/test/java/com/facebook/presto/hive/TestHiveHistoryBasedStatsTracking.java
```diff
@@ -258,6 +260,23 @@ public void testPartialAggStatisticsGroupByPartKey()
...
+    public void testHistoryBasedStatsCalculatorCTE()
+    {
+        String sql = "with t1 as (select orderkey, orderstatus from orders where totalprice > 100), t2 as (select orderkey, totalprice from orders where custkey > 100) " +
+                "select orderstatus, sum(totalprice) from t1 join t2 on t1.orderkey=t2.orderkey group by orderstatus";
+        Session cteMaterialization = Session.builder(defaultSession())
+                .setSystemProperty(CTE_MATERIALIZATION_STRATEGY, "ALL")
+                .setSystemProperty(CTE_PARTITIONING_PROVIDER_CATALOG, "hive")
+                .build();
+        // CBO Statistics
+        assertPlan(cteMaterialization, sql, anyTree(node(ProjectNode.class, anyTree(any())).withOutputRowCount(Double.NaN)));
```
jaystarshot, line 273: @feilong-liu will this also work if the cte materialization was off in the first execution and then on in the second?

jaystarshot: Also the opposite case, if cte materialization was on for the first run and then off again. I am a little confused and not sure it will work as expected in all these cases.

steveburnett:
Suggest minor change in release note entry to include the PR number:

```
== RELEASE NOTES ==

General Changes
* Enable HBO for CTE materialized query :pr:`22606`
```

presto-hive/src/test/java/com/facebook/presto/hive/TestHiveHistoryBasedStatsTracking.java
feilong-liu: No, both cases are not supported. History of CTE materialization runs only apply to CTE materialization runs and non cte materialization runs only apply to non cte materialization runs.

The challenge of support history share between CTE and non CTE runs is that, for non CTE materialization runs, the CTE will be inlined and expanded in query plan, which will have different variable name and plan node ids, and may also go through different optimizations which results in different plans. In order to match the plan hash of CTEs between CTE and non CTE runs, we need to carefully design and change the HBO related plan caononicalization strategy and code, which is non-trival.

So I prefer to first enable HBO for history to be used within CTE matrilizations queries. Otherwise HBO will be barely available (as the CTE related node is not canonicalized and will return empty in hash, which will lead to all other operators which has these CTE nodes as child also return empty in hash)

rschlussel (approved):

presto-hive/src/test/java/com/facebook/presto/hive/TestHiveHistoryBasedStatsTracking.java
jaystarshot: Sg,  I have fixed the planNodeIds and now they are canonicalized as expected in an earlier commit.
For the different variable names, I believe we can still use the underlying plan statistics available in cteConsumer/producer (original node HBO stats). These statistics will be accessible if the previous query was run in non-CTE materialized mode. However linking with the execution stats is unclear

PR #22607, opened 2024-04-24
[native] Improve Zombie Tasks logging.

spershin:
## Description
1. Only get task info (toString) if really needed.
2. Log also the number of extra references.
3. Log the type of Task for each task.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/TaskManager.cpp
+++ b/presto-native-execution/presto_cpp/main/TaskManager.cpp
@@ -207,15 +207,16 @@ void checkSplitsForBatchTask(
 }
 
 struct ZombieTaskStats {
-  const std::string taskId;
-  const std::string taskInfo;
+  const std::string info;
+  const long numExtraReferences;
 
-  explicit ZombieTaskStats(const std::shared_ptr<exec::Task>& task)
-      : taskId(task->taskId()), taskInfo(task->toString()) {}
-
-  std::string toString() const {
-    return SystemConfig::instance()->logZombieTaskInfo() ? taskInfo : taskId;
-  }
+  ZombieTaskStats(
+      const std::shared_ptr<exec::Task>& task,
+      long _numExtraReferences)
+      : info(
+            SystemConfig::instance()->logZombieTaskInfo() ? task->toString()
+                                                          : task->taskId()),
+        numExtraReferences(_numExtraReferences) {}
 };
 
 // Helper structure holding stats for 'zombie' tasks.
@@ -235,7 +236,9 @@ struct ZombieTaskStatsSet {
     tasks.reserve(numSampleTasks);
   }
 
-  void updateCounts(std::shared_ptr<exec::Task>& task) {
+  void updateCounts(
+      std::shared_ptr<exec::Task>& task,
+      long numExtraReferences) {
     switch (task->state()) {
       case exec::TaskState::kRunning:
         ++numRunning;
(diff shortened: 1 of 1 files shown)


presto-native-execution/presto_cpp/main/TaskManager.cpp
```diff
@@ -235,7 +236,9 @@ struct ZombieTaskStatsSet {
     tasks.reserve(numSampleTasks);
   }
 
-  void updateCounts(std::shared_ptr<exec::Task>& task) {
+  void updateCounts(
+      std::shared_ptr<exec::Task>& task,
+      const long numExtraReferences) {
```
xiaoxmeng, line 241: drop const here as it is scalar function input. Thanks!

presto-native-execution/presto_cpp/main/TaskManager.cpp
```diff
@@ -207,15 +207,16 @@ void checkSplitsForBatchTask(
...
-
-  std::string toString() const {
-    return SystemConfig::instance()->logZombieTaskInfo() ? taskInfo : taskId;
-  }
+  const std::string info;
+  const long numExtraReferences;
+
+  ZombieTaskStats(
+      const std::shared_ptr<exec::Task>& task,
+      long numExtraReferences_)
```
xiaoxmeng, line 215: Comment say the extra references could be either from velox or presto? Thanks!

xiaoxmeng (dismissed):
@spershin thanks for the improvement % nits.

presto-native-execution/presto_cpp/main/TaskManager.cpp
```diff
@@ -207,15 +207,16 @@ void checkSplitsForBatchTask(
...
-  explicit ZombieTaskStats(const std::shared_ptr<exec::Task>& task)
-      : taskId(task->taskId()), taskInfo(task->toString()) {}
-
-  std::string toString() const {
-    return SystemConfig::instance()->logZombieTaskInfo() ? taskInfo : taskId;
-  }
+  const std::string info;
+  const long numExtraReferences;
+
+  ZombieTaskStats(
```
czentgr, line 213: Nit: Why was `explicit` removed? This should still be used for the construction during the emplace call like before?

presto-native-execution/presto_cpp/main/TaskManager.cpp
spershin: This is conveyed on another level.
The structure has no knowledge about it.

presto-native-execution/presto_cpp/main/TaskManager.cpp
spershin: explicit is was used for single-argument constructors, so there would be no implicit type conversion.

czentgr (dismissed):

presto-native-execution/presto_cpp/main/TaskManager.cpp
```diff
@@ -207,15 +207,16 @@ void checkSplitsForBatchTask(
...
+      const std::shared_ptr<exec::Task>& task,
+      long _numExtraReferences)
```
xiaoxmeng, line 215:
```
s/long/uint32_t/ why long? Thanks!
```

xiaoxmeng (approved):

presto-native-execution/presto_cpp/main/TaskManager.cpp
spershin: @xiaoxmeng
Simply based on where we get them from:
`const auto prestoTaskRefCount = prestoTask.use_count();`
which comes from
`long use_count() const _NOEXCEPT`

PR #22609, opened 2024-04-25
Support procedure expire_snapshots for iceberg

hantangwangd:
## Description

This PR support the procedure `expire_snapshots` for iceberg. It can be used to remove older snapshots or specified snapshots and their files which are no longer needed.

See examples as follow:

 - Remove snapshots older than specific day and time, but retain the last 10 snapshots::

```
CALL iceberg.system.expire_snapshots('schema_name', 'table_name', TIMESTAMP '2023-08-31 00:00:00.000', 10);
```

 - Remove snapshots with snapshot ID 10001 and 10002 (note that this snapshot ID should not be the current snapshot)::

```
CALL iceberg.system.expire_snapshots(schema => 'schema_name', table_name => 'table_name', snapshot_ids => ARRAY[10001, 10002]);
```

## Motivation and Context

Support expiring snapshots for iceberg

## Test Plan

 - Newly added test cases in `TestExpireSnapshotProcedure`

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==


Iceberg Connector Changes
* Add procedure `expire_snapshots` to remove old snapshots in Iceberg. :pr:`22609`
```


presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -912,6 +912,40 @@ procedure on the catalog's ``system`` schema::
     ``unregister_table`` only when using the Hive catalog. This is similar to
     the behavior listed above for the ``DROP TABLE`` command.
 
+Expire snapshots
+^^^^^^^^^^^^^^^^
+
+Each dml action in Iceberg produces a new snapshot while keeping the old data and metadata around for snapshot isolation and time travel. The expire_snapshots procedure can be used to remove older snapshots and their files which are no longer needed.
```
steveburnett, line 918:
```suggestion
Each DML (Data Manipulation Language) action in Iceberg produces a new snapshot while keeping the old data and metadata for snapshot isolation and time travel. Use `expire_snapshots` to remove older snapshots and their files.
```
Suggest capitalizing DML and adding the meaning of the acronym, other suggestions for active voice and conciseness.

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -912,6 +912,40 @@ procedure on the catalog's ``system`` schema::
...
+
+This procedure will remove old snapshots and the corresponding snapshot files, and it will never remove files which are still required by a non-expired snapshot.
```
steveburnett, line 920:
```suggestion
This procedure removes old snapshots and their corresponding files, and never removes files which are required by a non-expired snapshot.
```

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -912,6 +912,40 @@ procedure on the catalog's ``system`` schema::
...
+``schema``            ✔️         string          Schema of the table to update
+
+``table_name``        ✔️         string          Name of the table to update
+
+``older_than``                   timestamp       Timestamp before which snapshots will be removed (Default: 5 days ago)
+
+``retain_last``                  int             Number of ancestor snapshots to preserve regardless of older_than
+                                                 (defaults to 1)
+
+``snapshot_ids``                 array of long   Array of snapshot IDs to expire
```
steveburnett, line 936: Is a word missing after "array of long"?

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -912,6 +912,40 @@ procedure on the catalog's ``system`` schema::
...
+===================== ========== =============== =======================================================================
+
+Examples:
+
+* Remove snapshots older than specific day and time, but retain the last 10 snapshots::
+
+    CALL iceberg.system.expire_snapshots('schema_name', 'table_name', TIMESTAMP '2023-08-31 00:00:00.000', 10);
+
+* Remove snapshots with snapshot ID 10001 and 10002 (note that this snapshot ID should not be the current snapshot)::
```
steveburnett, line 945:
```suggestion
* Remove snapshots with snapshot ID 10001 and 10002 (note that these snapshot IDs should not be the current snapshot)::
```

steveburnett (changes requested):
Thanks for the doc! Looks good, I made a few suggestions. Let me know what you think.

presto-docs/src/main/sphinx/connector/iceberg.rst
hantangwangd: Thanks for the suggestion, fixed!

presto-docs/src/main/sphinx/connector/iceberg.rst
hantangwangd: No, what I want to express is that the type is `array of long`. Is it OK with this expression in your opinion?

steveburnett: That's fine! I was unfamiliar with the phrase and I should have looked it up first. This is good as is. Thanks!

steveburnett (dismissed):
LGTM! (docs)

Pull updated branch, new local docs build, everything looks good. Thanks!

steveburnett:
Suggest minor changes to the release note entry, following the [Order of Changes](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines#order-of-changes) and [Section naming](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines#section-naming) in Release Notes Guidelines. 

```
== RELEASE NOTES ==

Iceberg Connector Changes
* Add procedure `expire_snapshots` to remove old snapshots in Iceberg. :pr:`22609`
```

hantangwangd:
@steveburnett Thanks for the standardization of the release note, fixed!

kiersten-stokes (approved):
LGTM! All the expiry options seem at first like they're going to require a lot of explanation, but it's very clear and concise 😄

presto-iceberg/src/test/java/com/facebook/presto/iceberg/procedure/TestExpireSnapshotProcedure.java
```diff
@@ -0,0 +1,289 @@
...
+        Instant instant = Instant.ofEpochMilli(timeMillsUtc);
+        LocalDateTime localDateTime = instant
+                .atZone(ZoneId.of(zoneId))
+                .toLocalDateTime();
+        DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.SSS");
+        formatter = formatter.withZone(ZoneId.of(zoneId));
+        return localDateTime.format(formatter);
+    }
+
+    protected Table loadTable(String tableName)
```
tdcmeehan, line 246: Can these be private?

presto-iceberg/src/test/java/com/facebook/presto/iceberg/procedure/TestExpireSnapshotProcedure.java
```diff
@@ -0,0 +1,289 @@
...
+
+    protected File getCatalogDirectory()
+    {
+        Path dataDirectory = getDistributedQueryRunner().getCoordinator().getDataDirectory();
+        return dataDirectory.toFile();
+    }
+
+    protected long waitUntilAfter(long timestampMillis)
+    {
+        long current = System.currentTimeMillis();
```
tdcmeehan, line 266: This is run in a hot loop.  I don't see where we actually wait though--this seems to always wait until after the timestamp of the snapshot.  Do we need it?

tdcmeehan (commented):
Nice work!  Just a couple of questions.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/procedure/TestExpireSnapshotProcedure.java
hantangwangd: I think we need this. Iceberg's `ExpireSnapshots.expireOlderThan(timestamp)` will expires all snapshots older than the given milliseconds timestamp, that means the snapshot with the same timestamp would not be expired. So for the sake of stability in testing, we'd better wait for this millisecond to pass, in case that all things happen in the same millisecond.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/procedure/TestExpireSnapshotProcedure.java
hantangwangd: Nice catch, fixed!

presto-iceberg/src/test/java/com/facebook/presto/iceberg/procedure/TestExpireSnapshotProcedure.java
```diff
@@ -0,0 +1,289 @@
...
+            table.refresh();
+
+            waitUntilAfter(table.currentSnapshot().timestampMillis());
+
+            String currentTimestamp = getTimestampString(System.currentTimeMillis(), zoneId);
+
+            assertHasSize(table.snapshots(), 3);
+
+            // expire with retainLast param
+            assertUpdate(session, format("CALL system.expire_snapshots('%s', '%s', TIMESTAMP '%s')", TEST_SCHEMA, tableName, currentTimestamp));
```
ZacBlanco, line 142: Am I missing something? It seems like the `retainLast` param isn't specified here even though the comment indicates you should be using it.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/procedure/TestExpireSnapshotProcedure.java
ZacBlanco: Can we not just use `Thread.sleep(1)` in place of this?

ZacBlanco (commented):
a few minor things, otherwise LGTM

presto-iceberg/src/test/java/com/facebook/presto/iceberg/procedure/TestExpireSnapshotProcedure.java
hantangwangd: Oh, I referred to Iceberg's test cases which use this spinning wait. In my opinion, when the wait time is very short, for example shorter than 1ms in this case, a spinning wait may be better than a block wait. It could avoid thread scheduling and context switching which may be triggered by `Thread.sleep(ts)`, that might cause the waiting time more longer than expected.

Besides, should we obey the test standards to avoid using `Thread.sleep`? Referring to https://github.com/prestodb/presto/blob/master/CONTRIBUTING.md?plain=1#L374

presto-iceberg/src/test/java/com/facebook/presto/iceberg/procedure/TestExpireSnapshotProcedure.java
hantangwangd: Thanks for the pointer. My fault for losing the retainLast parameter, fixed!

presto-iceberg/src/test/java/com/facebook/presto/iceberg/procedure/TestExpireSnapshotProcedure.java
tdcmeehan: If it's shorter than 1ms, let's add a precondition which asserts that, so that the test fails rather than busy waits?

hantangwangd: Sure, that makes sense! Fixed.

ZacBlanco (dismissed):

presto-iceberg/src/test/java/com/facebook/presto/iceberg/procedure/TestExpireSnapshotProcedure.java
```diff
@@ -0,0 +1,295 @@
...
+        Path dataDirectory = getDistributedQueryRunner().getCoordinator().getDataDirectory();
+        return dataDirectory.toFile();
+    }
+
+    private long waitUntilAfter(long timestampMillis)
+    {
+        long current = System.currentTimeMillis();
+        if (timestampMillis - current > 10) {
+            throw new PrestoException(StandardErrorCode.INVALID_ARGUMENTS, "Do not allow to spinning wait for longer than 10ms");
+        }
```
tdcmeehan, line 271: Perhaps rename argument to be `snapshotTimeMillis`, and call current time `currentTimeMillis`.

We can just use a simple assert for this.

```suggestion
        assertTrue(snapshotTimeMillis - currentTimeMillis <= 10, format("Snapshot time %s is greater than the current time %s by more than 10ms"));
```

tdcmeehan (commented):
Just one nit, otherwise LGTM

presto-iceberg/src/test/java/com/facebook/presto/iceberg/procedure/TestExpireSnapshotProcedure.java
hantangwangd: Sure, thanks for the suggestion, fixed!

PR #22619, opened 2024-04-26
Improve KLL Sketch perf for non-grouped queries

ZacBlanco:
## Description

Defers calculations of memory sizes for Kll sketches unless using a grouped query.

## Motivation and Context

The current method for calculating memory usage has a hidden cost. Within `getEstimatedKllInMemorySize` we call `getSerializedSizeBytes`. The code for the serialized bytes size actually serializes the entire internal state to a byte array first before returning the length. This is expensive and should be avoided.

I am working on a PR to the upstream library to add a less-costly method but until released, I would like to fix this as non-grouped execution doesn't need the memory accounting for every sketch input.

## Impact

N/A

## Test Plan

N/A

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```


aaneja:
Can you paste the call stack or method in datasketches-java that ends up serializing the internal state for size calculation ?

ZacBlanco:
[You can find the relevant section here](https://github.com/apache/datasketches-java/blob/0bff44b13ccf2eb27d2519942016d9d83327af5f/src/main/java/org/apache/datasketches/kll/KllHeapItemsSketch.java#L219-L236)

However, for the `KllDoublesSketch`, the copying doesn't occur. The call chain eventually gets to [here](https://github.com/apache/datasketches-java/blob/0bff44b13ccf2eb27d2519942016d9d83327af5f/src/main/java/org/apache/datasketches/kll/KllDoublesSketch.java#L495) where `getNumRetained()` is just a few array index lookups. I have a local branch where I'm trying to replace the implementation for doubles with this version of the sketch.

They don't have a sketch implementation for raw `long`s yet, but I am planning on contributing that to the library so that for the native numeric types we can get better performance creating the sketches. You can see some of the perf numbers I was getting in [this comment](https://github.com/apache/datasketches-java/pull/554#issuecomment-2082973601)

tdcmeehan (approved):

PR #22646, opened 2024-05-01
[native] Temporarily disable two unit tests.

spershin:
## Description
For upcoming change in Velox these tests need to be disabled because they will start to fail.
After the Velox change is landed and used, we will re-enable the tests with the fix.

Velox PR: https://github.com/facebookincubator/velox/pull/9663

```
== NO RELEASE NOTE ==
```


aditi-pandit (approved):
Thanks @spershin

aditi-pandit:
@spershin : Do you have the Velox PR that would cause these tests to fail ? Would be great to link it here as well.

PR #22652, opened 2024-05-01
array_join adds an extra delimeter if the last element is null

kevinwilfong:
## Description
array_join adds an extra trailing delimeter to the result string if the last element in the array is null

## Motivation and Context
this seems to be clearly incorrect, if the null replacement argument isn't set nulls are otherwise ignored by array_join

## Impact
this will change the results of array_join for cases where the last element is null

## Test Plan
added a testcase where the null replacement argument is not set and the last element is null, this was previously untested

## Contributor checklist

- [X] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Fix array_join to not add a trailing delimeter when the last element in the array is NULL. :pr:`22652`

```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/operator/scalar/ArrayJoin.java
+++ b/presto-main/src/main/java/com/facebook/presto/operator/scalar/ArrayJoin.java
@@ -315,15 +315,19 @@ public final class ArrayJoin
             Slice nullReplacement)
     {
         int numElements = arrayBlock.getPositionCount();
+        boolean hasOutput = false;
 
         for (int i = 0; i < numElements; i++) {
+            if (arrayBlock.isNull(i) && nullReplacement == null) {
+                continue;
+            }
+
+            if (hasOutput) {
+                blockBuilder.writeBytes(delimiter, 0, delimiter.length());
+            }
+
             if (arrayBlock.isNull(i)) {
-                if (nullReplacement != null) {
-                    blockBuilder.writeBytes(nullReplacement, 0, nullReplacement.length());
-                }
-                else {
-                    continue;
-                }
+                blockBuilder.writeBytes(nullReplacement, 0, nullReplacement.length());
             }
             else {
                 try {
@@ -337,9 +341,7 @@ public final class ArrayJoin
                 }
             }
 
-            if (i != numElements - 1) {
-                blockBuilder.writeBytes(delimiter, 0, delimiter.length());
-            }
+            hasOutput = true;
         }
 
         blockBuilder.closeEntry();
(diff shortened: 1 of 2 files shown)


mbasmanova:
CC: @tdcmeehan @aditi-pandit @steveburnett

steveburnett:
Nit: suggest adding the PR to the release note entry. 

```
== RELEASE NOTES ==

General Changes
* Fix array_join to not add a trailing delimeter when the last element in the array is NULL. :pr:`22652`
```

presto-main/src/main/java/com/facebook/presto/operator/scalar/ArrayJoin.java
```diff
@@ -315,15 +315,24 @@ public static void arrayJoinProvidedBlock(
             Slice nullReplacement)
     {
         int numElements = arrayBlock.getPositionCount();
+        boolean firstElement = true;
 
         for (int i = 0; i < numElements; i++) {
+            boolean elementIsNull = arrayBlock.isNull(i);
+
+            if (elementIsNull && nullReplacement == null) {
```
tdcmeehan, line 323:
```suggestion
            if (arrayBlock.isNull(i) && nullReplacement == null) {
```

presto-main/src/main/java/com/facebook/presto/operator/scalar/ArrayJoin.java
```diff
@@ -315,15 +315,24 @@ public static void arrayJoinProvidedBlock(
             Slice nullReplacement)
     {
         int numElements = arrayBlock.getPositionCount();
+        boolean firstElement = true;
```
tdcmeehan, line 318:
```suggestion
        boolean hasOutput = false;
```

I think this name is a bit misleading.  It seems this is actually referring to whether or not this function has encountered any non-null positions in the block, or if there has been a null position, that it hasn't been replaced with the replacement string, and in either case has this value been written to the block builder.  Hence I'd rename this to something like the above.

presto-main/src/main/java/com/facebook/presto/operator/scalar/ArrayJoin.java
```diff
@@ -315,15 +315,24 @@ public static void arrayJoinProvidedBlock(
...
+            if (elementIsNull && nullReplacement == null) {
+                continue;
+            }
+
+            if (firstElement) {
+                firstElement = false;
+            }
+            else {
+                blockBuilder.writeBytes(delimiter, 0, delimiter.length());
+            }
```
tdcmeehan, line 332:
```suggestion
            if (hasOutput) {
                blockBuilder.writeBytes(delimiter, 0, delimiter.length());
            }
            else {
                hasOutput = true;
            }
```

kevinwilfong:
Thanks for the review!

tdcmeehan (approved):

PR #22661, opened 2024-05-03
Revert "Add histograms for optimizer cost calculation"

feilong-liu:
This reverts commit e8c465902a6b4025721d9fd4bf7936a155ba6a79. This reverts PR https://github.com/prestodb/presto/pull/21236

As there are other PRs referring to the changes in this PR, hence also reverting https://github.com/prestodb/presto/pull/22327 and https://github.com/prestodb/presto/pull/22395 here

## Description
This reverted commit will make query with large IN clause to be super slow. In the test query I ran, the latency increase from 7 seconds to 4 and half minutes!

If not reverting, it will take 4 and half minutes to execute this query.

```
presto:tpch> select orderpriority, sum(totalprice) from lineitem join orders on lineitem.orderkey = orders.orderkey where orders.orderkey in ( 0 , 1 , 2 , 3 , 4 , 5 , 6 , 7 , 8 , 9 , ..., 9995 , 9996 , 9997 , 9998 , 9999 ) group by 1;  -- the IN clause has element from 0 to 9999
  orderpriority  |        _col1         NG] i[0 0B 0B] o[0 0B 0B] splits[65/29/0]B] splits[0/0/0]
-----------------+----------------------
 2-HIGH          | 3.3029310959000003E8 
 5-LOW           | 3.4520414947999984E8 
 3-MEDIUM        | 3.5743611102999985E8 
 1-URGENT        | 3.6284532178999996E8 
 4-NOT SPECIFIED |       3.6984263411E8 
(5 rows)

Query 20240503_165715_00066_i9y9u, FINISHED, 4 nodes
http://localhost:8080/ui/query.html?20240503_165715_00066_i9y9u
Splits: 94 total, 94 done (100.00%)
[Latency: client-side: 4:30, server-side: 4:30] [53.7K rows, 1.74MB] [199 rows/s, 6.6KB/s]
```

After revert the change, it will take only 7 seconds!
```
presto:tpch> select orderpriority, sum(totalprice) from lineitem join orders on lineitem.orderkey = orders.orderkey where orders.orderkey in ( 0 , 1 , 2 , 3 , 4 , 5 , 6 , 7 , 8 , 9 , ..., 9995 , 9996 , 9997 , 9998 , 9999 ) group by 1;  -- the IN clause has element from 0 to 9999
  orderpriority  |        _col1         NG_FOR_PREREQUISITES] i[0 0B 0B] o[0 0B 0B] splits[0/0/0]
-----------------+----------------------
[...]

PR #22664, opened 2024-05-03
Fix optimizer performance regression for large IN

ZacBlanco:
## Description

This bug occurs due to large IN lists. When computing statistics for queries with this clause, we generate an instance of VariableStatisticEstimate for each expression in the list. Fixes the bug found in #22661 

For example, a query with

```sql
SELECT ... IN (1, 2,  ... 10_000);
```

generates 10k VariableStatisticEstimate instances. Then, for each instance, we sum the statistics and distinct values to get a final result to determine the probability of a value occurring. Creating these VariableStatisticEstimate instances were not the root cause of the issue.

The main problem is that when the
DIsjointRangeDomainHistogram was introduced it was designed as immutable. When your fold all of the instances together through the `reduce()` call at FilterStatsCalculator#L755 re-creates the internal TreeRangeSet. The set of disjoint ranges is the size of the IN list, so we were re-creating the set 10k times, the first with range a set size of 1, then 2, ... etc up to 10k.

The following change updates the DisjointRangeDomain histogram to be mutable. This prevents re-creating the range set for every new addition to the IN list. This does not seem to affect the correctness of the histogram cost calculations.

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* ...
* ...

Hive Connector Changes
* ...
* ...
```

```
== NO RELEASE NOTE ==
```


elharo:
"The main problem is that when the DIsjointRangeDomainHistogram was introduced it was designed as immutable. When folding all of the instances together via the reduce() call at FilterStatsCalculator#L755 it re-creates the internal TreeRangeSet. The set of disjoint ranges is the size of the IN list, so we were re-creating the set 10k times. The first with range a set size of 1, then 2, ... etc up to 10k."

Tangential note: this is exactly the pattern that led to the handshoe DoS attack on protobuf. Immutable data structures are not as risk-free as they seem.

elharo (dismissed):

jaystarshot:
I missed some context, but why is this a problem when the property (use_histogram) is turned off?

ZacBlanco:
This happened because the histograms were still computed in `PlanNodeStatsEstimateMath` here:

https://github.com/ZacBlanco/prestodb/blob/b377b41532c4607dd79ff83fa56d9da374995a76/presto-main/src/main/java/com/facebook/presto/cost/PlanNodeStatsEstimateMath.java#L289

presto-main/src/main/java/com/facebook/presto/util/Lazy.java
```diff
@@ -0,0 +1,64 @@
...
+        this.init = requireNonNull(init, "init is null");
+        return this;
+    }
+
+    public synchronized Lazy<T> reset(T value)
+    {
+        this.value = value;
+        this.init = null;
+        return this;
+    }
```
tdcmeehan, line 44: Do we use these methods anywhere outside of the constructor?

presto-main/src/main/java/com/facebook/presto/util/Lazy.java
```diff
@@ -0,0 +1,64 @@
...
+
+import java.util.Optional;
+import java.util.function.Supplier;
+
+import static java.util.Objects.requireNonNull;
+
+public class Lazy<T>
+        implements Supplier<T>
+{
+    private T value;
```
tdcmeehan, line 25: If the API of this class is to be able to reset the supplier, and supply a new value, shouldn't this be volatile as well?

ZacBlanco: Good point. I think the class can be simplified. I will make it so that volatile is not required here

presto-main/src/main/java/com/facebook/presto/util/Lazy.java
ZacBlanco: currently no, I will simplify the implementation

presto-main/src/main/java/com/facebook/presto/cost/VariableStatsEstimate.java
```diff
@@ -79,6 +84,16 @@ public VariableStatsEstimate(
...
+        this.histogram = requireNonNull(histogram, "histogram is null");
+    }
+
+    public VariableStatsEstimate(double lowValue,
+            double highValue,
+            double nullsFraction,
+            double averageRowSize,
+            double distinctValuesCount)
+    {
+        this(lowValue, highValue, nullsFraction, averageRowSize, distinctValuesCount, new UniformDistributionHistogram(lowValue, highValue));
```
feilong-liu, line 96: Instead of assign a UniformDistributionHistogram, can you make the histogram to be an optional field and assign empty() when histogram is turned off?

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java
```diff
@@ -2359,23 +2361,40 @@ public void testLargeIn()
...
+    @Test
+    public void testLargeInWithHistograms()
+    {
+        String longValues = range(0, 10_000)
+                .mapToObj(Integer::toString)
+                .collect(joining(", "));
+        Session session = Session.builder(getSession())
+                .setSystemProperty(ITERATIVE_OPTIMIZER_TIMEOUT, "30000ms")
+                .setSystemProperty(OPTIMIZER_USE_HISTOGRAMS, "true")
+                .build();
```
feilong-liu, line 2395: Can you also test with histogram to be false, and set a smaller timeout value.

presto-main/src/main/java/com/facebook/presto/cost/ComparisonStatsCalculator.java
```diff
@@ -31,26 +36,31 @@
...
     private ComparisonStatsCalculator() {}
 
     public static PlanNodeStatsEstimate estimateExpressionToLiteralComparison(
             PlanNodeStatsEstimate inputStatistics,
             VariableStatsEstimate expressionStatistics,
             Optional<VariableReferenceExpression> expressionVariable,
             OptionalDouble literalValue,
-            ComparisonExpression.Operator operator)
+            ComparisonExpression.Operator operator,
+            Optional<Session> session)
```
feilong-liu, line 49: If only histogram field in the session is used, can we pass a boolean instead of the whole session around?

ZacBlanco: Do you see a specific downside to passing the session? If we intend to make changes in the future which reference more feature flags, I feel it would be better to pass the session rather than adding a new argument for each potentially new flag

presto-main/src/main/java/com/facebook/presto/cost/VariableStatsEstimate.java
ZacBlanco: fixed

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java
ZacBlanco: added

presto-main/src/main/java/com/facebook/presto/cost/ComparisonStatsCalculator.java
feilong-liu: I was thinking about keeping it consistent as I see in some other cases it's passing a boolean flag. But also fine if you think passing session is better.

presto-main/src/main/java/com/facebook/presto/cost/FilterStatsCalculator.java
```diff
@@ -135,18 +135,19 @@ public PlanNodeStatsEstimate filterStats(
     public PlanNodeStatsEstimate filterStats(
             PlanNodeStatsEstimate statsEstimate,
             RowExpression predicate,
-            ConnectorSession session)
+            ConnectorSession session,
+            Optional<Session> systemSession)
```
rschlussel, line 138: when would the session not be present? My main concern would be if there are cases where the config won't be checked.

presto-main/src/main/java/com/facebook/presto/cost/PlanNodeStatsEstimateMath.java
```diff
@@ -259,11 +290,17 @@ private static VariableStatsEstimate addColumnStats(
...
-
-        return VariableStatsEstimate.builder()
+        VariableStatsEstimate.Builder statistics = VariableStatsEstimate.builder()
                 .setStatisticsRange(sum)
                 .setAverageRowSize(newAverageRowSize)
-                .setNullsFraction(newNullsFraction)
-                .build();
+                .setNullsFraction(newNullsFraction);
+        if (shouldUseHistograms) {
+            Optional<ConnectorHistogram> newHistogram = RangeAdditionStrategy.INTERSECT.equals(strategy) ?
```
rschlussel, line 298: use == comparison for enums so we get compiler errors if we're accidentally comparing with the wrong type of thing.

presto-main/src/main/java/com/facebook/presto/cost/FilterStatsCalculator.java
ZacBlanco: There was a discussion in the original PR on this: https://github.com/prestodb/presto/pull/21236#discussion_r1434243541

Mainly an issue due to the implementation of `ConnectorFilterStatsCalculatorService`. Most of the time when we call `filterStats, we will have a `Session` object, but when the stats filtering is called from `ConnectorFilterStatsCalculatorService`, we only get a `ConnectorSession`. ConnectorSession and Session are two separate classes with different methods, and you can't check for a system session property through a connector session. 

There is a piece of coded added here which is where the session is populated: https://github.com/prestodb/presto/blob/18d2e7aab4ae020e9abf380ac87f8c08225ab625/presto-main/src/main/java/com/facebook/presto/cost/ConnectorFilterStatsCalculatorService.java#L60-L62

It should usually be there as there are only two implementations of ConnectorSession (`FullConnectorSession` and `TestingConnectorSession`), but I didn't want to re-design the session APIs in the histogram PR for this, so I introduced the "hack" here to get the underlying session instance

presto-main/src/main/java/com/facebook/presto/cost/PlanNodeStatsEstimateMath.java
ZacBlanco: thanks, fixed

presto-main/src/main/java/com/facebook/presto/cost/FilterStatsCalculator.java
rschlussel: yeah, the whole connector session abstraction is kind of weird.  How much value do you think we get from being able to set this on a per-session basis.  because a nicer alternative for this problem is to only have a configuration property for this and inject it directly into filterStatsCalculator, Then we don't have to worry about passing the session around or expecting the session to probably be there even though the interface doesn't guarantee it.

ZacBlanco: Personally, I think it's been pretty useful during some of my manual testing to be able to set the session property within the same query runner/cluster and not have to change config+restart. Additionally, with our current infra it's a little bit of a headache to set the file-based configs flags. Session properties are nice because it doesn't require us to restart the whole cluster when I just want to see how a histogram affects a query plan's estimates and shape.

I could take a stab at shuffling some fields to see if we could get the underlying Session from the ConnectorsSession instance, but I'd probably punt it to another PR

rschlussel: can you add a comment then in the meantime explaining why it's optional and the plan to fix it.

presto-main/src/test/java/com/facebook/presto/operator/scalar/queryplan/TestJsonPrestoQueryPlanFunctionUtils.java
```diff
@@ -46,21 +46,36 @@ private TestJsonPrestoQueryPlanFunctionUtils() {}
                     "         \"highValue\" : \"NaN\",\n" +
                     "         \"nullsFraction\" : 1.0,\n" +
                     "         \"averageRowSize\" : 0.0,\n" +
-                    "         \"distinctValuesCount\" : 0.0\n" +
+                    "         \"distinctValuesCount\" : 0.0,\n" +
+                    "         \"histogram\" : {\n" +
+                    "           \"@class\" : \"com.facebook.presto.cost.UniformDistributionHistogram\",\n" +
```
jaystarshot, line 51: If the default is false, why do we see a plan change here?

ZacBlanco: This was changed in the original PR before making the histogram field optional. So, when the histogram _wasn't_ optional, this test failed because it couldn't deserialize the JSON plan into the correct types. However, with the commit making the type optional, this can probably be reverted.

ZacBlanco: I've reverted this in the commit that makes the field optional

presto-main/src/main/java/com/facebook/presto/cost/FilterStatsCalculator.java
ZacBlanco: Comment added

presto-main/src/main/java/com/facebook/presto/cost/VariableStatsEstimate.java
```diff
@@ -99,6 +115,15 @@ public double getNullsFraction()
         return nullsFraction;
     }
 
+    // We ignore the histogram during serialization because histograms can be
+    // quite large. Histograms are not used outside the coordinator, so there
+    // isn't a need to serialize them
+    @JsonIgnore
```
ZacBlanco, line 121: @rschlussel This is the change to prevent serializing histograms. An associated test is added in [TestVariableStatsEstimate](https://github.com/prestodb/presto/blob/43ef3e7ba04324eb2a08d15d3d6bd9fb4be78e7f/presto-main/src/test/java/com/facebook/presto/cost/TestVariableStatsEstimate.java)

rschlussel (approved):
just fix the json property and should be good.

rschlussel (dismissed):

aaneja (dismissed):

hantangwangd (commented):
The description in the first commit has a little inaccuracy,`ConnectorHistogram` have 2 methods, not 4 :-).

presto-main/src/main/java/com/facebook/presto/cost/DisjointRangeDomainHistogram.java
```diff
@@ -0,0 +1,359 @@
...
+ * "source" histogram.
+ * <br>
+ * For example, assume a source histogram represents a uniform distribution
+ * over the range [0, 100]. Next, assume a query with multiple filters such as
+ * <code>x < 10 OR x > 75</code>. This translates to two disjoint ranges over
+ * the histogram of [0, 10) and (75, 100], representing roughly 35% of the
+ * values in the original dataset. Using the example above, a cumulative
+ * probability for value 5 represents 5% of the original dataset, but 20% (1/5)
+ * of the range of constrained dataset. Similarly, all values in [10, 75] should
+ * compute their cumulative probability as 40% (2/5).
```
hantangwangd, line 59: Are the numbers in this example inaccurate?

presto-main/src/main/java/com/facebook/presto/cost/HistogramCalculator.java
```diff
@@ -0,0 +1,167 @@
...
+        boolean openHigh = range.getOpenHigh();
+        boolean openLow = range.getOpenLow();
+        Estimate min = histogram.inverseCumulativeProbability(0.0);
+        Estimate max = histogram.inverseCumulativeProbability(1.0);
+
+        // range is either above or below histogram
+        if ((!max.isUnknown() && max.getValue() < range.getLow())
+                || (!min.isUnknown() && min.getValue() > range.getHigh())) {
+            return Estimate.of(0.0);
+        }
```
hantangwangd, line 59: Should we consider the value of `openHigh` and `openLow` to make the judgement more refined? For example, if `openLow == true`, then we can judge `max.getValue() <= range.getLow()`.

presto-main/src/main/java/com/facebook/presto/cost/HistogramCalculator.java
```diff
@@ -0,0 +1,167 @@
...
+            // in the case the histogram has no values
+            if (totalDistinctValues.equals(Estimate.zero()) || range.getDistinctValuesCount() == 0.0) {
+                return Estimate.of(0.0);
+            }
+
+            // in the case only one is unknown
+            if (((lowPercentile.isUnknown() && !highPercentile.isUnknown()) ||
+                    (!lowPercentile.isUnknown() && highPercentile.isUnknown())) &&
+                    isFinite(range.length())) {
+                return useHeuristics ? Estimate.of(StatisticRange.INFINITE_TO_FINITE_RANGE_INTERSECT_OVERLAP_HEURISTIC_FACTOR) : Estimate.unknown();
```
hantangwangd, line 104: Here should we directly return `Estimate.of(StatisticRange.INFINITE_TO_FINITE_RANGE_INTERSECT_OVERLAP_HEURISTIC_FACTOR)` as `useHeuristics` is always true?

presto-main/src/main/java/com/facebook/presto/cost/HistogramCalculator.java
```diff
@@ -0,0 +1,167 @@
...
+
+        // in the case the range is a single value, this can occur if the input
+        // filter range is a single value (low == high) OR in the case that the
+        // bounds of the filter or this histogram are infinite.
+        // in the case of infinite bounds, we should return an estimate that
+        // correlates to the overlapping distinct values.
+        if (lowPercentile.equals(highPercentile)) {
+            if (!useHeuristics) {
+                return Estimate.zero();
+            }
```
hantangwangd, line 126: Should we return a `Estimate.unknown()` or a `Estimate.zero()` here? As the comment mentioned, there may exists cases that the bounds are infinite. I'm not quite so sure, what's your thought about it?

presto-main/src/main/java/com/facebook/presto/cost/HistogramCalculator.java
```diff
@@ -0,0 +1,167 @@
...
+            // If one of the bounds is unknown, but both percentiles are equal,
+            // it's likely that a heuristic value was returned
+            if (max.isUnknown() || min.isUnknown()) {
+                return totalDistinctValues.flatMap(distinct -> lowPercentile.map(lowPercent -> distinct * lowPercent));
```
hantangwangd, line 130: Is this calculation ok? I might be wrong, but as I understand, it may return a result bigger than `1.0`.

presto-main/src/main/java/com/facebook/presto/cost/HistogramCalculator.java
```diff
@@ -0,0 +1,167 @@
...
+                return Estimate.unknown();
+            }
+
+            if (totalDistinctValues.equals(Estimate.zero())) {
+                return Estimate.of(1.0);
+            }
+            return totalDistinctValues.flatMap(totalDistinct -> {
+                if (DoubleMath.fuzzyEquals(totalDistinct, 0.0, 1E-6)) {
+                    return Estimate.unknown();
+                }
```
hantangwangd, line 149: I'm a little confused here. I may miss some context message, but why return a `Estimate.of(1.0)` when `totalDistinctValues` equals 0, then return a `Estimate.unknown()` when `totalDistinctValues` fuzzy equals 0?

presto-main/src/main/java/com/facebook/presto/cost/HistogramCalculator.java
```diff
@@ -0,0 +1,167 @@
...
+
+        return Optional.of(lowPercentile)
+                .filter(lowPercent -> !lowPercent.isUnknown())
+                .map(Estimate::getValue)
+                .map(lowPercent -> Optional.of(highPercentile)
+                        .filter(highPercent -> !highPercent.isUnknown())
+                        .map(Estimate::getValue)
+                        .map(highPercent -> highPercent - lowPercent)
+                        .map(Estimate::of)
+                        .orElseGet(() -> Estimate.of(1.0)))
```
hantangwangd, line 164: Should we return `Estimate.of(1.0 - lowPercent)` here?

hantangwangd (commented):
Thanks for the great work, overall LGTM except some little things.

presto-main/src/main/java/com/facebook/presto/cost/HistogramCalculator.java
ZacBlanco: This case is required to return zero, or else it has adverse affects on plans for queries without statistics

presto-main/src/main/java/com/facebook/presto/cost/HistogramCalculator.java
ZacBlanco: I think this is a good idea. I will update it

presto-main/src/main/java/com/facebook/presto/cost/HistogramCalculator.java
ZacBlanco: I took a closer look again since I think this code was written before adding `map` to the `Estimate` type. Both `lowPercentile` and `highPercentile` should be available at this time, so I think we should just return `highPercent - lowPercent` always, and in the case of any unknown, just leave it to unknown as it shouldn't occur. Any other value for unknowns would just be a heuristic anyways.

The new version would just be: 

```java
return lowPercentile.flatMap(lowPercent -> highPercentile.map(highPercent -> highPercent - lowPercent));
```

presto-main/src/main/java/com/facebook/presto/cost/HistogramCalculator.java
ZacBlanco: This is a good point. I think originally I was defending against having a divisor of 0. I will update this to

```java
return totalDistinctValues.flatMap(totalDistinct -> {
                if (DoubleMath.fuzzyEquals(totalDistinct, 0.0, 1E-6)) {
                    return Estimate.of(1.0);
                }
                return Estimate.of(min(1.0, range.getDistinctValuesCount() / totalDistinct));
            })
            // in the case totalDistinct is NaN or 0
            .or(() -> Estimate.of(StatisticRange.INFINITE_TO_INFINITE_RANGE_INTERSECT_OVERLAP_HEURISTIC_FACTOR));
```

presto-main/src/main/java/com/facebook/presto/cost/HistogramCalculator.java
ZacBlanco: You are correct it could return something larger than 1.0. I checked all of our test cases and found that this conditional was not actually hit in any of them, so I have opted to remove it.

presto-main/src/main/java/com/facebook/presto/cost/DisjointRangeDomainHistogram.java
ZacBlanco: These numbers are accurate

hantangwangd: I'm a little confused that value 5 got a cumulative probability 20% (1/5) on dataset [0, 10) and (75, 100], which spans 35. Shouldn't it got 1/7 on a dataset which spans 35, or got 1/5 on a dataset which spans 25? Any misunderstandings please let me know, thanks.

presto-main/src/main/java/com/facebook/presto/cost/HistogramCalculator.java
hantangwangd: Thanks for the explanation, it's more clearer.

presto-main/src/main/java/com/facebook/presto/cost/HistogramCalculator.java
hantangwangd: OK, got it.

presto-main/src/main/java/com/facebook/presto/cost/DisjointRangeDomainHistogram.java
ZacBlanco: Ah you are correct. It should have been 85. Updated

hantangwangd (dismissed):
LGTM. Thanks for the fix and the great work!

feilong-liu (approved):

presto-main/src/main/java/com/facebook/presto/cost/PlanNodeStatsEstimateMath.java
```diff
@@ -22,14 +27,17 @@
 
 public class PlanNodeStatsEstimateMath
 {
-    private PlanNodeStatsEstimateMath()
-    {}
+    final boolean shouldUseHistograms;
```
tdcmeehan, line 30:
```suggestion
    private final boolean shouldUseHistograms;
```

presto-main/src/main/java/com/facebook/presto/cost/StatisticRange.java
```diff
@@ -28,22 +38,40 @@
...
 {
-    private static final double INFINITE_TO_FINITE_RANGE_INTERSECT_OVERLAP_HEURISTIC_FACTOR = 0.25;
-    private static final double INFINITE_TO_INFINITE_RANGE_INTERSECT_OVERLAP_HEURISTIC_FACTOR = 0.5;
+    protected static final double INFINITE_TO_FINITE_RANGE_INTERSECT_OVERLAP_HEURISTIC_FACTOR = 0.25;
+    protected static final double INFINITE_TO_INFINITE_RANGE_INTERSECT_OVERLAP_HEURISTIC_FACTOR = 0.5;
+
+    /**
+     * Serialized length in bytes. 3 doubles + 2 booleans
+     */
+    private static final int SERIALIZED_LENGTH = 3 * Double.BYTES + 2;
```
tdcmeehan, line 47:
```suggestion
    private static final int SERIALIZED_LENGTH_IN_BYTES = 3 * Double.BYTES + 2;
```

presto-main/src/main/java/com/facebook/presto/cost/ConnectorFilterStatsCalculatorService.java
```diff
@@ -53,7 +55,12 @@ public TableStatistics filterStats(
             Map<String, Type> columnTypes)
     {
         PlanNodeStatsEstimate tableStats = toPlanNodeStats(tableStatistics, columnNames, columnTypes);
-        PlanNodeStatsEstimate filteredStats = filterStatsCalculator.filterStats(tableStats, predicate, session);
+        // TODO: Consider re-designing the filter calculator APIs so that a proper Session instance
```
tdcmeehan, line 58: It seems like everywhere else, we just blindly cast to `FullConnectorSession`.  Do you know under what circumstances we won't be able to cast?  If not, can we try to just validate that it is a `FullConnectorSession` and just get the session?  I am trying to get rid of these references to `Optional<Session>`, which seem strange, as a session comes with every query.

ZacBlanco: I think currently, it should work all the time since there is only 1 implementation used outside of testing. I wrote this defensively in the case we ever came up with some separate implementation where we couldn't cast it properly. If you're okay with the behavior of failing the query in that case, then I can remove all the `Optional`'s

tdcmeehan: My preference is to fail loudly whenever this presumption no longer holds, rather than to silently default to false.

presto-main/src/main/java/com/facebook/presto/cost/StatisticRange.java
ZacBlanco: Given the changes to histograms to not serialize them when sending over plan fragments and potentially to not serialize them at all to workers, I've opted to revert the change to the serialization format of the the `StatisticRange` class.

jaystarshot (approved):

presto-main/src/main/java/com/facebook/presto/cost/ConnectorFilterStatsCalculatorService.java
ZacBlanco: This has been fixed

tdcmeehan (approved):
Thanks @ZacBlanco

rschlussel:
I saw a test failure today from testLargeInWithHistograms test https://github.com/prestodb/presto/issues/22944

PR #22667, opened 2024-05-03
Prefilter join build side when it's too large

kaikalur:
## Description
Optimize the build side of join using the distinct keys from left when the right (and left too) are large. 

## Motivation and Context

SELECT .. FROM T1 JOIN T2 USING(x)

can be very slow/memory intense when T2 is big (and T1 is also big). So the idea is to do something like dynamic filter except on the build side! So the above query becomes:

SELECT ... FROM T1 LEFT JOIN (SELECT * FROM T2 WHERE x IN (SELECT DISTINCT x FROM T1)) T2 USING(x)

This has helped us tremendously in some of our production workloads. So making it an optimization.

## Test Plan
Added tests

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
Added a new optimization for prefiltering the build side of a join with distinct keys from the probe side.  This can be enabled with the ``join_prefilter_build_side `` session property. :pr:`22667`

     join_prefilter_build_side

```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
+++ b/presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
@@ -345,6 +345,7 @@ public final class SystemSessionProperties
     public static final String NATIVE_EXECUTION_PROCESS_REUSE_ENABLED = "native_execution_process_reuse_enabled";
     public static final String NATIVE_DEBUG_VALIDATE_OUTPUT_FROM_OPERATORS = "native_debug_validate_output_from_operators";
     public static final String DEFAULT_VIEW_SECURITY_MODE = "default_view_security_mode";
+    public static final String JOIN_PREFILTER_BUILD_SIDE = "join_prefilter_build_side";
 
     private final List<PropertyMetadata<?>> sessionProperties;
 
@@ -1925,7 +1926,12 @@ public final class SystemSessionProperties
                         featuresConfig.getDefaultViewSecurityMode(),
                         false,
                         value -> CreateView.Security.valueOf(((String) value).toUpperCase()),
-                        CreateView.Security::name));
+                        CreateView.Security::name),
+                booleanProperty(
+                        JOIN_PREFILTER_BUILD_SIDE,
+                        "Prefiltering the build/inner side of a join with keys from the other side",
+                        false,
+                        false));
     }
 
     public static boolean isSpoolingOutputBufferEnabled(Session session)
@@ -3206,5 +3212,10 @@ public final class SystemSessionProperties
     public static CreateView.Security getDefaultViewSecurityMode(Session session)
     {
         return session.getSystemProperty(DEFAULT_VIEW_SECURITY_MODE, CreateView.Security.class);
+    }
+
+    public static boolean isJoinPrefilterEnabled(Session session)
+    {
+        return session.getSystemProperty(JOIN_PREFILTER_BUILD_SIDE, Boolean.class);
     }
 }
--- a/presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/JoinPrefilter.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/JoinPrefilter.java
@@ -1,0 +1,171 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.sql.planner.optimizations;
+
+import com.facebook.presto.Session;
+import com.facebook.presto.metadata.Metadata;
+import com.facebook.presto.spi.VariableAllocator;
+import com.facebook.presto.spi.WarningCollector;
+import com.facebook.presto.spi.plan.AggregationNode;
+import com.facebook.presto.spi.plan.EquiJoinClause;
+import com.facebook.presto.spi.plan.FilterNode;
+import com.facebook.presto.spi.plan.PlanNode;
+import com.facebook.presto.spi.plan.PlanNodeIdAllocator;
+import com.facebook.presto.spi.relation.VariableReferenceExpression;
+import com.facebook.presto.sql.planner.TypeProvider;
+import com.facebook.presto.sql.planner.plan.JoinNode;
+import com.facebook.presto.sql.planner.plan.SemiJoinNode;
+import com.facebook.presto.sql.planner.plan.SimplePlanRewriter;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+import static com.facebook.presto.SystemSessionProperties.isJoinPrefilterEnabled;
+import static com.facebook.presto.common.type.BooleanType.BOOLEAN;
+import static com.facebook.presto.spi.plan.AggregationNode.singleGroupingSet;
+import static com.facebook.presto.spi.plan.JoinType.INNER;
+import static com.facebook.presto.spi.plan.JoinType.LEFT;
+import static com.facebook.presto.sql.planner.PlannerUtils.clonePlanNode;
+import static com.facebook.presto.sql.planner.PlannerUtils.isScanFilterProject;
+import static com.facebook.presto.sql.planner.PlannerUtils.projectExpressions;
+import static com.facebook.presto.sql.planner.plan.ChildReplacer.replaceChildren;
+import static java.util.Objects.requireNonNull;
+
+public class JoinPrefilter
+        implements PlanOptimizer
+{
+    private final Metadata metadata;
+    private boolean isEnabledForTesting;
+
+    public JoinPrefilter(Metadata metadata)
+    {
+        this.metadata = requireNonNull(metadata, "metadata is null");
+    }
+
+    @Override
+    public void setEnabledForTesting(boolean isSet)
+    {
+        isEnabledForTesting = isSet;
+    }
+
+    @Override
+    public boolean isEnabled(Session session)
+    {
+        return isEnabledForTesting || isJoinPrefilterEnabled(session);
+    }
+
+    @Override
+    public PlanOptimizerResult optimize(PlanNode plan, Session session, TypeProvider types, VariableAllocator variableAllocator, PlanNodeIdAllocator idAllocator, WarningCollector warningCollector)
+    {
+        if (isEnabled(session)) {
+            Rewriter rewriter = new Rewriter(session, metadata, idAllocator, variableAllocator);
+            PlanNode rewritten = SimplePlanRewriter.rewriteWith(rewriter, plan, null);
+            return PlanOptimizerResult.optimizerResult(rewritten, rewriter.isPlanChanged());
+        }
+
+        return PlanOptimizerResult.optimizerResult(plan, false);
+    }
+
+    private static class Rewriter
+            extends SimplePlanRewriter<Void>
+    {
+        private final Session session;
+        private final Metadata metadata;
+        private final PlanNodeIdAllocator idAllocator;
+        private final VariableAllocator variableAllocator;
+        private boolean planChanged;
+
+        private Rewriter(Session session, Metadata metadata, PlanNodeIdAllocator idAllocator, VariableAllocator variableAllocator)
+        {
+            this.session = requireNonNull(session, "session is null");
+            this.metadata = requireNonNull(metadata, "functionAndTypeManager is null");
+            this.idAllocator = requireNonNull(idAllocator, "idAllocator is null");
+            this.variableAllocator = requireNonNull(variableAllocator, "idAllocator is null");
+        }
+
+        @Override
+        public PlanNode visitJoin(JoinNode node, RewriteContext<Void> context)
+        {
+            PlanNode left = node.getLeft();
+            PlanNode right = node.getRight();
+
+            PlanNode rewrittenLeft = rewriteWith(this, left);
+            PlanNode rewrittenRight = rewriteWith(this, right);
+            List<EquiJoinClause> equiJoinClause = node.getCriteria();
+
+            // We apply this for only left and inner join and the left side of the join is a simple scan and the join is on one key
+            if (equiJoinClause.size() == 1 &&
+                    (node.getType() == LEFT || node.getType() == INNER) &&
+                    isScanFilterProject(rewrittenLeft)) {
+                VariableReferenceExpression leftKey = equiJoinClause.stream().map(x -> x.getLeft()).findFirst().get();
+                VariableReferenceExpression rightKey = equiJoinClause.stream().map(x -> x.getRight()).findFirst().get();
+
+                // First create a SELECT DISTINCT leftKey FROM left
+                Map<VariableReferenceExpression, VariableReferenceExpression> leftVarMap = new HashMap();
+                PlanNode leftKeys = clonePlanNode(rewrittenLeft, session, metadata, idAllocator, ImmutableList.of(leftKey), leftVarMap);
+                PlanNode projectNode = projectExpressions(leftKeys, idAllocator, variableAllocator, ImmutableList.of(leftVarMap.get(leftKey)), ImmutableList.of());
+
+                // DISTINCT on the leftkey
+                PlanNode filteringSource = new AggregationNode(
+                        leftKey.getSourceLocation(),
+                        idAllocator.getNextId(),
+                        projectNode,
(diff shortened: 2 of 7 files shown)


agrawaldevesh:
Awesome ! Is this strictly opt in or can it be hbo or cbo'd ? 

Is there also a way to fail the select distinct early if its cardinality is too big ?

Finally, I did not follow: can it be applied to multiple equi join keys too ?

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java
```diff
@@ -7527,4 +7528,24 @@ public void testLambdaInAggregation()
...
+
+    @Test
+    public void testJoinPrefilter()
+    {
+        // Orig
+        String testQuery = "SELECT 1 from region join nation using(regionkey)";
+        MaterializedResult result = computeActual("explain(type distributed) " + testQuery);
+        assertTrue(((String) result.getMaterializedRows().get(0).getField(0)).indexOf("SemiJoin") == -1);
+        result = computeActual(testQuery);
+        assertTrue(result.getRowCount() == 25);
```
elharo, line 7540: assertEquals

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java
```diff
@@ -7527,4 +7528,24 @@ public void testLambdaInAggregation()
...
+
+        // With feature
+        Session session = Session.builder(getSession())
+                .setSystemProperty(JOIN_PREFILTER_ENABLED, String.valueOf(true))
+                .build();
+        result = computeActual(session, "explain(type distributed) " + testQuery);
+        assertTrue(((String) result.getMaterializedRows().get(0).getField(0)).indexOf("SemiJoin") != -1);
+        result = computeActual(session, testQuery);
+        assertTrue(result.getRowCount() == 25);
```
elharo, line 7549: assertEquals

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java
```diff
@@ -7527,4 +7528,24 @@ public void testLambdaInAggregation()
...
     }
+
+    @Test
+    public void testJoinPrefilter()
+    {
+        // Orig
+        String testQuery = "SELECT 1 from region join nation using(regionkey)";
+        MaterializedResult result = computeActual("explain(type distributed) " + testQuery);
+        assertTrue(((String) result.getMaterializedRows().get(0).getField(0)).indexOf("SemiJoin") == -1);
+        result = computeActual(testQuery);
```
elharo, line 7539: try to avoid reusing local variables

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java
```diff
@@ -7527,4 +7528,24 @@ public void testLambdaInAggregation()
...
+        assertTrue(result.getRowCount() == 25);
+
+        // With feature
```
elharo, line 7542: Split this out into a separate test method

elharo (commented):
Great idea!

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/JoinPrefilter.java
```diff
@@ -0,0 +1,171 @@
...
+import static com.facebook.presto.spi.plan.JoinType.INNER;
+import static com.facebook.presto.spi.plan.JoinType.LEFT;
+import static com.facebook.presto.sql.planner.PlannerUtils.clonePlanNode;
+import static com.facebook.presto.sql.planner.PlannerUtils.isScanFilterProject;
+import static com.facebook.presto.sql.planner.PlannerUtils.projectExpressions;
+import static com.facebook.presto.sql.planner.plan.ChildReplacer.replaceChildren;
+import static java.util.Objects.requireNonNull;
+
+public class JoinPrefilter
+        implements PlanOptimizer
```
ClarenceThreepwood, line 50: Any reason this cannot be implemented as an IterativeOptimizer?

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/JoinPrefilter.java
```diff
@@ -0,0 +1,171 @@
...
+                VariableReferenceExpression leftKey = equiJoinClause.stream().map(x -> x.getLeft()).findFirst().get();
+                VariableReferenceExpression rightKey = equiJoinClause.stream().map(x -> x.getRight()).findFirst().get();
+
+                // First create a SELECT DISTINCT leftKey FROM left
+                Map<VariableReferenceExpression, VariableReferenceExpression> leftVarMap = new HashMap();
+                PlanNode leftKeys = clonePlanNode(rewrittenLeft, session, metadata, idAllocator, ImmutableList.of(leftKey), leftVarMap);
+                PlanNode projectNode = projectExpressions(leftKeys, idAllocator, variableAllocator, ImmutableList.of(leftVarMap.get(leftKey)), ImmutableList.of());
+
+                // DISTINCT on the leftkey
+                PlanNode filteringSource = new AggregationNode(
```
ClarenceThreepwood, line 124: Isn't this redundant since the semijoin filters out duplicates?

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/JoinPrefilter.java
```diff
@@ -0,0 +1,171 @@
...
+            PlanNode right = node.getRight();
+
+            PlanNode rewrittenLeft = rewriteWith(this, left);
+            PlanNode rewrittenRight = rewriteWith(this, right);
+            List<EquiJoinClause> equiJoinClause = node.getCriteria();
+
+            // We apply this for only left and inner join and the right side of the join is a simple scan and the join is on one key
+            if (equiJoinClause.size() == 1 &&
+                    (node.getType() == LEFT || node.getType() == INNER) &&
+                    isScanFilterProject(rewrittenLeft)) {
```
ClarenceThreepwood, line 114: The comment above says "right side of the join is a simple scan"

ClarenceThreepwood (changes requested):
Can you share some performance numbers that you see in your workloads? Maybe even add a SqlBenchmark that showcases this optimization?

IIUC - this optimization reduces the size of the hash table that is built out of T2. In order to do this it adds a second table scan on T1 and then builds a second hash table to compute the distinct join key from T1. I'm curious where the benefit comes from? Is it just the improved performance of the semijoin?

Any thoughts on how this can be used in practice? I ask because this is not a cost based decision and can degrade performance in many usecases

presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
```diff
@@ -1925,7 +1926,12 @@ public SystemSessionProperties(
                         featuresConfig.getDefaultViewSecurityMode(),
                         false,
                         value -> CreateView.Security.valueOf(((String) value).toUpperCase()),
-                        CreateView.Security::name));
+                        CreateView.Security::name),
+                booleanProperty(
+                        JOIN_PREFILTER_ENABLED,
```
feilong-liu, line 1931: Since the prefilter is on build side, maybe the session property to reflect this?

yingsu00 (commented):
I'm curious, if we already know the distinct keys in T1, why not just make it as the build side? No need to calculate the hash values, just use the distinct values as the hash values. This way there is no need to scan T1 twice.

steveburnett:
suggest minor revision of the release notes entry 

```
== RELEASE NOTES ==

General Changes
* Add optimization for prefiltering the build side of a join with distinct keys from the probe side. This can be enabled with the ``join_prefilter_enabled`` session property. :pr:`22667`
```

presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
kaikalur: Done

kaikalur:
> Can you share some performance numbers that you see in your workloads? Maybe even add a SqlBenchmark that showcases this optimization?
> 
> IIUC - this optimization reduces the size of the hash table that is built out of T2. In order to do this it adds a second table scan on T1 and then builds a second hash table to compute the distinct join key from T1. I'm curious where the benefit comes from? Is it just the improved performance of the semijoin?
> 
> Any thoughts on how this can be used in practice? I ask because this is not a cost based decision and can degrade performance in many usecases

Two potential cases - a) build side is very big and only a few keys actually match so we shuffle a lot less right side, b) after the semijoin, the build side becomes small enough to broadcast which can eliminate shuffling the full left side which could have a lot of payload.

kaikalur:
> I'm curious, if we already know the distinct keys in T1, why not just make it as the build side? No need to calculate the hash values, just use the distinct values as the hash values. This way there is no need to scan T1 twice.

You need to get the rest of the fields!

kaikalur:
> > Can you share some performance numbers that you see in your workloads? Maybe even add a SqlBenchmark that showcases this optimization?
> > IIUC - this optimization reduces the size of the hash table that is built out of T2. In order to do this it adds a second table scan on T1 and then builds a second hash table to compute the distinct join key from T1. I'm curious where the benefit comes from? Is it just the improved performance of the semijoin?
> > Any thoughts on how this can be used in practice? I ask because this is not a cost based decision and can degrade performance in many usecases
> 
> Two potential cases - a) build side is very big and only a few keys actually match so we shuffle a lot less right side, b) after the semijoin, the build side becomes small enough to broadcast which can eliminate shuffling the full left side which could have a lot of payload.

Added test:

select count(1) from part join lineitem using (partkey) where part.name like '%x%'

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/JoinPrefilter.java
kaikalur: It's a pain to get the left and right sides in the iterative - left as an exercise lol

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/JoinPrefilter.java
kaikalur: But the Bag datastructure ooms if you have giant right side

kaikalur:
Add a task for making it cost based - needs tracking some new stats in hbo:

https://fburl.com/2aui8j1i

presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
```diff
@@ -3207,4 +3213,9 @@ public static CreateView.Security getDefaultViewSecurityMode(Session session)
     {
         return session.getSystemProperty(DEFAULT_VIEW_SECURITY_MODE, CreateView.Security.class);
     }
+
+    public static boolean isJoinPrefilterEnabled(Session session)
```
jaystarshot, line 3217: I think this must be added in Featuresconfig as a config property

kaikalur: No - this is not a feature of the cluster/deployment. This is a query level optimization so I don't want to add it there

jaystarshot: I see anyways its false now, but in future if we make this cost based, it might become a cluster level feature.

kaikalur: I'm generally against these feature config stuff for query level optimizations because HBO can be controlled at query level

kaikalur: Separately we need HBO configuration that can be used to control what features it supports. @feilong-liu  and I were talking about it

feilong-liu (approved):

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/JoinPrefilter.java
kaikalur: Also there could be things like shuffles etc. This is originally intended as a SQL rewrite which is what our PlanOptimizer is closest to

kaikalur:
@ClarenceThreepwood - can you take a look when you get a chance?

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/JoinPrefilter.java
ClarenceThreepwood: In that case wouldn't it be better to include this additional distinct as a generic cost-based optimization to all semijoins with a large right side?

ClarenceThreepwood (commented):
Please update the release note with the new name of the session property

kaikalur:
> > Add a task for making it cost based - needs tracking some new stats in hbo:
> > https://fburl.com/2aui8j1i
> 
> This is meta internal only?

Oops. Sorry. here correct link:

https://github.com/prestodb/presto/issues/22706

kaikalur:
addressed comments

ClarenceThreepwood:
> > Please update the release note with the new name of the session property
> 
> Done

It still has the old name here
Added a new optimization for prefiltering the build side of a join with distinct keys from the probe side.  This can be enabled with the ``join_prefilter_enabled`` session property. :pr:`22667`

kaikalur:
> > > Please update the release note with the new name of the session property
> > 
> > 
> > Done
> 
> It still has the old name here Added a new optimization for prefiltering the build side of a join with distinct keys from the probe side. This can be enabled with the `join_prefilter_enabled` session property. :pr:`22667`



> > > Please update the release note with the new name of the session property
> > 
> > 
> > Done
> 
> It still has the old name here Added a new optimization for prefiltering the build side of a join with distinct keys from the probe side. This can be enabled with the `join_prefilter_enabled` session property. :pr:`22667`

OK for real this time lol - damn scrollbar!

kaikalur:
OK all comments addressed (again)

ClarenceThreepwood (approved):

pranjalssh (approved):

yingsu00:
>> I'm curious, if we already know the distinct keys in T1, why not just make it as the build side? No need to calculate the hash values, just use the distinct values as the hash values. This way there is no need to scan T1 twice.

> You need to get the rest of the fields!

@kaikalur Could you explain a bit more? If we just make T1 the build side, we can still get the rest of the fields, can't we? For example, `select part.partkey, lineitem.quantity from part join lineitem using (partkey) where part.name like '%x%'` if we make part the build side, the dynamic filter on partkey can be propergated to lineitem, and part.partkey, lineitem.quantity can be output as well. It's still inner equijoin so the semantics should not change. Does the SQL standard say T2 must be the build side in `T1 join T2`?

kaikalur:
> > > I'm curious, if we already know the distinct keys in T1, why not just make it as the build side? No need to calculate the hash values, just use the distinct values as the hash values. This way there is no need to scan T1 twice.
> 
> > You need to get the rest of the fields!
> 
> @kaikalur Could you explain a bit more? If we just make T1 the build side, we can still get the rest of the fields, can't we? For example, `select part.partkey, lineitem.quantity from part join lineitem using (partkey) where part.name like '%x%'` if we make part the build side, the dynamic filter on partkey can be propergated to lineitem, and part.partkey, lineitem.quantity can be output as well. It's still inner equijoin so the semantics should not change. Does the SQL standard say T2 must be the build side in `T1 join T2`?

No but that's a join reordering problem?

yingsu00:
> No but that's a join reordering problem?

Yes, that's what I was asking. Can't we just reorder the join? Or does the SQL standard say the table on the right side of the JOIN keyword has to be the build side? It seems to me the join order is a decision of the engine, isn't it?

feilong-liu:
> > No but that's a join reordering problem?
> 
> Yes, that's what I was asking. Can't we just reorder the join? Or does the SQL standard say the table on the right side of the JOIN keyword has to be the build side? It seems to me the join order is a decision of the engine, isn't it?

My understanding is that it works when the probe side has large payload to output. For example:
select T1.k, T1.map, T1.row, T1.array, T2.k, T2.v from T1 Join T2 using(k)
here T1.map, T1.row, T1.array are large columns, if we put T1 on build side, we will pay a large memory cost to store these payloads, which can OOM.

yingsu00:
> My understanding is that it works when the probe side has large payload to output. For example:
select T1.k, T1.map, T1.row, T1.array, T2.k, T2.v from T1 Join T2 using(k)
here T1.map, T1.row, T1.array are large columns, if we put T1 on build side, we will pay a large memory cost to store these payloads, which can OOM.

@feilong-liu Thanks for your message. It makes sense. Maybe we can keep the pointers instead of copying all these fields to the hash table in that case.

feilong-liu:
> > My understanding is that it works when the probe side has large payload to output. For example:
> > select T1.k, T1.map, T1.row, T1.array, T2.k, T2.v from T1 Join T2 using(k)
> > here T1.map, T1.row, T1.array are large columns, if we put T1 on build side, we will pay a large memory cost to store these payloads, which can OOM.
> 
> @feilong-liu Thanks for your message. It makes sense. Maybe we can keep the pointers instead of copying all these fields to the hash table in that case.

Keep pointers sounds an interesting idea. Can you elaborate more? One challenge I can think of is that, even we keep pointers, we still need to read these payload and attach to the result later, which still need to load the payload from memory.

yingsu00:
The payload is already in memory from the incoming pages. But you're right, if both sides are too big to be held in memory, then the side that produces smaller hash table (including the output payload) should be the build side.

PR #21962, opened 2024-02-19
[native] Add support for optional velox.properties file

majetideepak:
## Description

Velox config has no required fields and therefore velox.properties should be optional.
From a Presto user perspective, existing config files (config.properties, node.properties)
for Java should work for Prestissimo as well.

## Motivation and Context

Make Prestissimo config aligned to Presto Java

## Impact

velox.properties is now optional.

## Test Plan

Tested locally.

## Contributor checklist

- [x] Please make sure your submission complies with our [development](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#development), [formatting](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#formatting), [commit message](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#commit-formatting-and-pull-requests), and [attribution guidelines](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#attribution).
- [x] PR description addresses the issue accurately and concisely.  If the change is non-trivial, a GitHub Issue is referenced.
- [x] Documented new properties (with its default value), SQL syntax, functions, or other functionality.
- [x] If release notes are required, they follow the [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines).
- [x] Adequate tests were added if applicable.
- [x] CI passed.

```
== NO RELEASE NOTE ==
```

PR #22669, opened 2024-05-04
[native] Remove velox.properties file from e2e tests.

amitkdutta:
## Description
Removing velox properties file which is not used with e2e tests. Since no config is passed, the file is not necessary.

## Motivation and Context
In https://github.com/prestodb/presto/pull/21962, we made velox properties file optional. Also no e2e tests utilizes this. It was added because the file was required to start the server. However as described in https://github.com/prestodb/presto/pull/21962, its optional now and hence removing from the tests as there isn't any usage in the test

## Impact
No impact

## Test Plan
Ran all e2e tests locally and in CI

```
== NO RELEASE NOTE ==
```


xiaoxmeng (approved):
@amitkdutta thanks!

NikhilCollooru (approved):

PR #22671, opened 2024-05-05
Optimize Metastore Code

konjac-h:
## Description

Optimize the Code by reducing unnecessary for loop. Reducing time complexity from O(2n) to O(n)

## Motivation and Context

See Description Above  

## Impact

Improve time complexity of Metastore 

## Test Plan

Simple Code Optimization.
Ran Verifier test to ensure nothing breaks 

## Contributor checklist

- [x] CI passed.

== NO RELEASE NOTE ==

### DIFF
--- a/presto-hive-metastore/src/main/java/com/facebook/presto/hive/metastore/InMemoryCachingHiveMetastore.java
+++ b/presto-hive-metastore/src/main/java/com/facebook/presto/hive/metastore/InMemoryCachingHiveMetastore.java
@@ -37,6 +37,7 @@ import javax.annotation.concurrent.ThreadSafe;
 import javax.inject.Inject;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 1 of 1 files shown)


elharo (changes requested):
This needs a test

konjac-h:
@elharo Hey thanks for reviewing. I totally agree that we should have a regression test. I filed a separate task, [T188102626](https://www.internalfb.com/intern/tasks/?t=188102626), to track it but I don't think the test will be the blocker of this PR given that this is a really straightforward optimization which improves time complexity from O(2n) to O(n)

steveburnett:
The checkbox for "If release notes are required, they follow the [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines)." is checked but there is no release note. 

Please add a release note following the [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines), even if it's only 
```
== NO RELEASE NOTE ==
```

elharo:
Was this an outright bug (valid query failed, incorrect results reported) or a performance problem? If the former I would expect a test that exposes the bug to be written first before a fix is attempted. Otherwise it's likely to regress again even if it's manually verified this time

jainxrohit:
@konjac-h can you please add in the PR description on what exactly was the issue?

jainxrohit (approved):

rschlussel:
A few comments:
1. it would be great to add more details to the issue description about what kind of issue we were seeing (e.g latency regression when fetching xxx from the metastore, failure when xxx, etc.), what the root cause of the problem was, and how this change addresses it.
2. it's great that you provided queryids for Meta employees, but you should make sure to provide an example query or category of query for the general public outside of meta to be able to see what kind of queries were causing problems.
3. If you are fixing a bug you should add a test.  The fact that a fix was needed indicates that there was something not working properly in the code, and even if your fix is simple, adding a test ensures that we don't regress.

konjac-h:
@rschlussel Sounds good. thanks for comments/review. I will fill out the information you mentioned.

rschlussel:
> this is a really straightforward optimization improving the time complexity from O(2n) to O(n)

I'm surprised that iterating through a list 2x instead of 1x is enough to cause query failures.  That seems like a very sensitive threshold, and I would expect in that case we would have already been seeing similar failures for something using 2x as many partitions. Also the error for the failing query occurs the metastore, but it seems like the fix only touches the cache, and not even the method that was in the other stack trace. Are you sure this is the root cause of the issue?

konjac-h:
@rschlussel These two PRs were originally suspected to be the problem so I optimized the only one that could do better. After this, we found another PR that actually contributes to the majority of the delay.

rschlussel:
Got it, thanks for the explanation! So i would update the PR description and commit message, as this isn't really related to the metastore regression described above.  Instead i would describe it as an opportunistic code improvement that removes an extra iteration through a list.  You also shouldn't need to add a test then.

konjac-h:
@rschlussel do you mind taking a look and stamp it when you get a chance? Thanks

rschlussel (approved):

PR #22681, opened 2024-05-07
Broadcast join if build estimation is small and from HBO

feilong-liu:
## Description
Previously I added this session property to do broadcast join when build side is small and probe side is unknown. However, in production we found that it will cause problem when the build side estimation is wrong. However, HBO is a confident source and we should be able to enable it when the estimation is from HBO.

## Motivation and Context
To improve performance for queries with unknown probe and small build.

## Impact
To improve performance for queries with unknown probe and small build.

## Test Plan
It's an easy change, and also tested locally end to end.

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Improve session property ``property-use_broadcast_when_buildsize_small_probeside_unknown`` to do broadcast join when probe side size is unknown and build side estimation from HBO is small
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/DetermineJoinDistributionType.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/DetermineJoinDistributionType.java
@@ -26,6 +26,7 @@ import com.facebook.presto.spi.plan.CteConsumerNode;
 import com.facebook.presto.spi.plan.PlanNode;
 import com.facebook.presto.spi.plan.TableScanNode;
 import com.facebook.presto.spi.plan.ValuesNode;
+import com.facebook.presto.spi.statistics.HistoryBasedSourceInfo;
 import com.facebook.presto.sql.analyzer.FeaturesConfig.JoinDistributionType;
 import com.facebook.presto.sql.planner.iterative.Lookup;
 import com.facebook.presto.sql.planner.iterative.Rule;
@@ -126,7 +127,10 @@ public class DetermineJoinDistributionType
         if (possibleJoinNodes.stream().anyMatch(result -> result.getCost().hasUnknownComponents()) || possibleJoinNodes.isEmpty()) {
             // TODO: currently this session parameter is added so as to roll out the plan change gradually, after proved to be a better choice, make it default and get rid of the session parameter here.
             if (isUseBroadcastJoinWhenBuildSizeSmallProbeSizeUnknownEnabled(context.getSession()) && possibleJoinNodes.stream().anyMatch(result -> ((JoinNode) result.getPlanNode()).getDistributionType().get().equals(REPLICATED))) {
-                return getOnlyElement(possibleJoinNodes.stream().filter(result -> ((JoinNode) result.getPlanNode()).getDistributionType().get().equals(REPLICATED)).map(x -> x.getPlanNode()).collect(toImmutableList()));
+                JoinNode broadcastJoin = (JoinNode) getOnlyElement(possibleJoinNodes.stream().filter(result -> ((JoinNode) result.getPlanNode()).getDistributionType().get().equals(REPLICATED)).map(x -> x.getPlanNode()).collect(toImmutableList()));
+                if (context.getStatsProvider().getStats(broadcastJoin.getBuild()).getSourceInfo() instanceof HistoryBasedSourceInfo) {
+                    return broadcastJoin;
+                }
             }
             if (isSizeBasedJoinDistributionTypeEnabled(context.getSession())) {
                 return getSizeBasedJoin(joinNode, context);
--- a/presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/DetermineSemiJoinDistributionType.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/DetermineSemiJoinDistributionType.java
@@ -36,6 +36,7 @@ import com.facebook.presto.cost.TaskCountEstimator;
 import com.facebook.presto.matching.Captures;
 import com.facebook.presto.matching.Pattern;
 import com.facebook.presto.spi.plan.PlanNode;
+import com.facebook.presto.spi.statistics.HistoryBasedSourceInfo;
 import com.facebook.presto.sql.analyzer.FeaturesConfig.JoinDistributionType;
 import com.facebook.presto.sql.planner.iterative.Rule;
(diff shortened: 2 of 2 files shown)


presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/DetermineSemiJoinDistributionType.java
```diff
@@ -124,6 +128,12 @@ private PlanNode getCostBasedDistributionType(SemiJoinNode node, Context context
         possibleJoinNodes.add(getSemiJoinNodeWithCost(node.withDistributionType(PARTITIONED), context));
 
         if (possibleJoinNodes.stream().anyMatch(result -> result.getCost().hasUnknownComponents())) {
+            if (isUseBroadcastJoinWhenBuildSizeSmallProbeSizeUnknownEnabled(context.getSession()) && possibleJoinNodes.stream().anyMatch(result -> ((SemiJoinNode) result.getPlanNode()).getDistributionType().get().equals(REPLICATED))) {
```
feilong-liu, line 131: Also enable it for semi join

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/DetermineJoinDistributionType.java
```diff
@@ -126,7 +127,10 @@ private PlanNode getCostBasedJoin(JoinNode joinNode, Context context)
         if (possibleJoinNodes.stream().anyMatch(result -> result.getCost().hasUnknownComponents()) || possibleJoinNodes.isEmpty()) {
             // TODO: currently this session parameter is added so as to roll out the plan change gradually, after proved to be a better choice, make it default and get rid of the session parameter here.
             if (isUseBroadcastJoinWhenBuildSizeSmallProbeSizeUnknownEnabled(context.getSession()) && possibleJoinNodes.stream().anyMatch(result -> ((JoinNode) result.getPlanNode()).getDistributionType().get().equals(REPLICATED))) {
-                return getOnlyElement(possibleJoinNodes.stream().filter(result -> ((JoinNode) result.getPlanNode()).getDistributionType().get().equals(REPLICATED)).map(x -> x.getPlanNode()).collect(toImmutableList()));
+                JoinNode broadcastJoin = (JoinNode) getOnlyElement(possibleJoinNodes.stream().filter(result -> ((JoinNode) result.getPlanNode()).getDistributionType().get().equals(REPLICATED)).map(x -> x.getPlanNode()).collect(toImmutableList()));
```
kaikalur, line 130: We should do it always - we trust history and if HBO says build side is small, we just broadcast it.

jaystarshot:
This might be a good time to fix and use the correct confidence level in planNodeStatsEstimate

jaystarshot:
But that might involve additional testing and could be an overkill for this case so will leave you to decide

steveburnett:
Suggest revising the release note entry to follow the [Order of changes](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines#order-of-changes) in the Release Notes Guidelines. Maybe something like this? Edit this as needed, please.

```
== RELEASE NOTES ==

General Changes
* Improve session property ``property-name`` (to?) broadcast join when probe side size is unknown and build side estimation from HBO is small
```

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/DetermineJoinDistributionType.java
ClarenceThreepwood, line 130: nit: limit lines to [180 chars](https://github.com/prestodb/presto/blob/master/CONTRIBUTING.md#code-style:~:text=Lines%20should%20be%20no%20more%20than%20180%20chars.%20Intellij%20%E2%80%9CReformat%20code%E2%80%9D%20does%20NOT%20enforce%20the%20line%20width%2C%20so%20you%20must%20adjust%20it%20yourself.)

ClarenceThreepwood (commented):
I have a couple of related questions

1. It is well known that incorrect cost estimates can negatively influence optimizer planning. With this change, if the cost were correct, then we lose the benefit of the optimization - do we not? 

2. I agree that the HBO is a better source of truth. I wonder if we should introduce a config parameter (or a set of configs) that govern whether some optimizations should apply based on stats sources? Maintenance becomes hard if we pepper the code-base with hard-coded special cases for HBO optimizations.

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/DetermineJoinDistributionType.java
feilong-liu: This depends on whether the other side of the join is small or not, i.e. if both sides have estimations showing data size is small, whether we should do partitioned join or broadcast join. I think we can open a separate PR for this.

feilong-liu:
> But that might involve additional testing and could be an overkill for this case so will leave you to decide

Agree that we need to have different confidence level for estimations. I will consider this PR as a simple fix for existing features and not including other major changes.

feilong-liu:
> I have a couple of related questions
> 
> 1. It is well known that incorrect cost estimates can negatively influence optimizer planning. With this change, if the cost were correct, then we lose the benefit of the optimization - do we not?

Yes, we will. But fail queries which ran successfully is a larger pain than failing to optimize queries already running successful/failing. And this is the reason why we still cannot turn on this feature internally.

> 2. I agree that the HBO is a better source of truth. I wonder if we should introduce a config parameter (or a set of configs) that govern whether some optimizations should apply based on stats sources? Maintenance becomes hard if we pepper the code-base with hard-coded special cases for HBO optimizations.

I think this is a good idea and not only applicable to this optimization, but also other optimizations. But it also involves larger changes. I think we can open an issue for this.

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/DetermineJoinDistributionType.java
```diff
@@ -126,7 +127,10 @@ private PlanNode getCostBasedJoin(JoinNode joinNode, Context context)
...
+                if (context.getStatsProvider().getStats(broadcastJoin.getBuild()).getSourceInfo() instanceof HistoryBasedSourceInfo) {
+                    return broadcastJoin;
```
jaystarshot, line 132: I think, in future whether the previous join was broadcast/hash and if it was efficient/or if it failed etc could also be stored using HBO

jaystarshot (approved):

PR #22694, opened 2024-05-08
Parallelize maven builds by default

ZacBlanco:
## Description

Adding these parameters defaults all maven builds from this directory to using a parallelized mode if running on a multi-core machine. It can significantly improve build performance when compiling multiple modules.

## Motivation and Context

Faster builds without having to write `-T 1C` for every maven command.

## Impact

Potentially faster builds. Console output may be intertwined which can make following errors more difficult

## Test Plan

N/A

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```


ZacBlanco:
cc: @aaneja

elharo (commented):
The build has several warnings about non-parallelizable plugins. Anything we should do about that?

ZacBlanco:
Thanks for bringing that up. As far as I'm aware, the plugins don't seem to fail when executing. I've personally never seen a build failure that would have been due to that issue.

I combed through the build and this is the list of plugins currently marked as "not parallelizable".

- presto-maven-plugin
  - a custom plugin found at github.com/prestodb/presto-maven-plugin . I didn't find any unsafe code browsing the source. I think it would require a new metadata update + plugin release to fix
- provisio-maven-plugin
  - a dependency of the presto-maven-plugin. We would need to upgrade the presto-maven-plugin to remove this warning
- antlr4-maven-plugin
  - A dependency upgrade fixes this, but the latest versions require Java 11+. We don't officially build with Java 11 yet
- build-helper-maven-plugin
  - recent versions are marked as thread safe. I think we should be able to just bump the version, but I don't see the declarations for it in our pom.xml.
- drift-maven-plugin
  - I couldn't find the source for the current plugin version.
- fmpp-maven-plugin
  - An old and unmainted plugin that is probably thread-safe based on the source, but would require a new patch version+release.


Overall, the main effort with removing these warnings would be contacting the maintainers and getting them to issue a patch release with the plugin metadata updated to mark it as thread safe. I think many developers already set the `-T` argument on their machines and I know for a fact our internal build system already uses the parallel builds. I think it should be safe to ignore the plugin warnings.

presto-proxy/src/test/java/com/facebook/presto/proxy/TestProxyServer.java
```diff
@@ -161,7 +161,7 @@ public void testSetSession()
         }
     }
 
-    @Test(timeOut = 10_000)
+    @Test(timeOut = 15_000)
```
aaneja, line 164: Is this a related change ?

ZacBlanco: Yes, I found this test was timing out in a few of the builds with `-T 1C`.

aaneja (dismissed):

presto-proxy/src/test/java/com/facebook/presto/proxy/TestProxyServer.java
tdcmeehan: Why do you think this could be?  Do you think it's because the overall load is higher with this change?  In which case, do you think we could see some CI failures for existing test cases that rely on timing/sleeping?

tdcmeehan: Can you make this change as part of this PR?

ZacBlanco: Updated the CI command for `test-other-modules` and reverted the change to this test

tdcmeehan (approved):

PR #22699, opened 2024-05-08
[Native] Remove duplicate call to registerStatsCounters

karteekmurthys:
## Description
The startup flow registers the stats twice. I have retained the registerStatsCounter call in the context of `enableRuntimeMetrics` check, refer: https://github.com/prestodb/presto/blob/master/presto-native-execution/presto_cpp/main/PrestoServer.cpp#L1104.

Also removed double registration of this metric `kCounterMemoryCacheTotalTinyPaddingBytes`.


amitkdutta (approved):
Great change @karteekmurthys. Thanks.

PR #22701, opened 2024-05-08
Also check for Presto Native operator names in QueryStats.

spershin:
Presto Native operator names are different.

## Description
To properly report 'shuffled' and 'raw input' query stats when running Presto Native, we also check Presto Native operator names.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/execution/QueryStats.java
+++ b/presto-main/src/main/java/com/facebook/presto/execution/QueryStats.java
@@ -49,6 +49,11 @@ import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
 public class QueryStats
 {
+    // Native worker uses different operator names.
+    private static final String NATIVE_OPERATOR_MERGE = "MergeExchange";
+    private static final String NATIVE_OPERATOR_EXCHANGE = "Exchange";
+    private static final String NATIVE_OPERATOR_TABLE_SCAN = "TableScan";
+
     private final DateTime createTime;
 
     private final DateTime executionStartTime;
@@ -293,7 +298,8 @@ public class QueryStats
             DataSize peakTaskUserMemory,
             DataSize peakTaskTotalMemory,
             DataSize peakNodeTotalMemory,
-            RuntimeStats runtimeStats)
+            RuntimeStats runtimeStats,
+            boolean isNativeExecutionEnabled)
     {
         int totalTasks = 0;
         int runningTasks = 0;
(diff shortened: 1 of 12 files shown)


presto-main/src/main/java/com/facebook/presto/execution/QueryStats.java
```diff
@@ -373,12 +373,18 @@ public static QueryStats create(
                 PlanFragment plan = stageInfo.getPlan().get();
                 for (OperatorStats operatorStats : stageExecutionStats.getOperatorSummaries()) {
                     // NOTE: we need to literally check each operator type to tell if the source is from table input or shuffled input. A stage can have input from both types of source.
+                    // NOTE: also check for operator names from presto native execution.
                     String operatorType = operatorStats.getOperatorType();
-                    if (operatorType.equals(ExchangeOperator.class.getSimpleName()) || operatorType.equals(MergeOperator.class.getSimpleName())) {
+                    if (operatorType.equals(ExchangeOperator.class.getSimpleName()) ||
+                            operatorType.equals(MergeOperator.class.getSimpleName()) ||
+                            operatorType.equals("Exchange") ||
```
spershin, line 380: Not sure if it would be good to declare some string constants for Native operator names.
If that's the case, then where?

presto-main/src/main/java/com/facebook/presto/execution/QueryStats.java
```diff
@@ -373,12 +373,18 @@ public static QueryStats create(
...
+                            operatorType.equals("MergeExchange")) {
                         shuffledPositions += operatorStats.getRawInputPositions();
                         shuffledDataSize += operatorStats.getRawInputDataSize().toBytes();
                     }
-                    else if (operatorType.equals(TableScanOperator.class.getSimpleName()) || operatorType.equals(ScanFilterAndProjectOperator.class.getSimpleName())) {
+                    else if (operatorType.equals(TableScanOperator.class.getSimpleName()) ||
```
xiaoxmeng, line 385: Do we want to add the native execution check here for the case without Operator suffix? Thanks!

xiaoxmeng (commented):
@spershin thanks!

xiaoxmeng (dismissed):

presto-main/src/main/java/com/facebook/presto/execution/QueryStats.java
xiaoxmeng: Can we pass the second check if it is for native execution? I think we have a utility for this?

presto-main/src/main/java/com/facebook/presto/execution/QueryStats.java
spershin: If we do, then for a while Native won't be checked properly here, because for now the names are being changed to Java names.
IMHO, this is fine as it is unlikely that any operator would have these names.
We can wait for other reviewers to chip in.

presto-main/src/main/java/com/facebook/presto/execution/QueryStats.java
```diff
@@ -373,12 +373,18 @@ public static QueryStats create(
                 PlanFragment plan = stageInfo.getPlan().get();
                 for (OperatorStats operatorStats : stageExecutionStats.getOperatorSummaries()) {
                     // NOTE: we need to literally check each operator type to tell if the source is from table input or shuffled input. A stage can have input from both types of source.
+                    // NOTE: also check for operator names from presto native execution.
                     String operatorType = operatorStats.getOperatorType();
-                    if (operatorType.equals(ExchangeOperator.class.getSimpleName()) || operatorType.equals(MergeOperator.class.getSimpleName())) {
+                    if (operatorType.equals(ExchangeOperator.class.getSimpleName()) ||
```
mbasmanova, line 378: It would be nice to declare a constant set of operator names and use it here. It would also be helpful to declare 2 separate sets: one for Java and another for Native execution and choose which one to use based on isNativeExecution flag.

spershin: @mbasmanova 
Few Qs regarding the comment.
1. Where to declae such constants? The same java file?
2. Should we declare constants for java names too? It seems using the class names and does not really need constants.
3. If we use isNativeExecution flag, then the code won't work until we change presto_cpp. But I guess this would be short term and is ok.

mbasmanova: We can define constants in this java file.

We can initially include both Java names and Velox names in the map of operators used when isNativeExecution == true.

tdcmeehan:
Wondering if we can just add a boolean flag on the `OperatorStats` indicating that this was for an exchange?  This actually would remove the need to hardcode operator names or check if we're using native execution altogether.  I took a brief look and it looks like, at least from the Java side, it should be pretty straightforward.

xiaoxmeng (approved):
@spershin Thanks for the update!

spershin:
@tdcmeehan 

So, does this mean that we want two bool flags?
One for Exchange/Merge and another one for Scan?

tdcmeehan:
@spershin yes, that works.

spershin:
@tdcmeehan 

I might need some pointer to code site(s) where we populate the OperatorStats.
Members like operatorType, operatorId, planNodeId and such.
It is not trivial to find for some reason.

spershin:
NVM, found it.

spershin:
I don't think it is a good idea to add two struct fields just for that.
Closing the PR.

PR #22708, opened 2024-05-09
[native]Add io throttling configs to tune remote io

xiaoxmeng:
This used to control backoff on remote storage io congestion. We will use it for Meta internally.


amitkdutta (approved):
Looks good. Thanks @xiaoxmeng

amitkdutta (changes requested):
Actually one question. These are meta specific config. Do we want to add them in oss? We can add it inside Meta instead of exposing in OSS

xiaoxmeng:
@amitkdutta close per our offline discuss to add in Meta internal code.

PR #22718, opened 2024-05-11
[native] Introduce lastCoordinatorHeartbeatMs to PrestoTask.

spershin:
## Description
Right now we use 'lastHeartbeatMs' to determine if a Task has been abandoned by
the Coordinator (Coordinator restarted or crashed or anyhow gone).

This does not work too well, as 'lastHeartbeatMs' is updated on every message to the Task.
Including 'getResults', which workers send to each other and refreshing it.
As the result, the Tasks get abandoned in waves, starting from stage 0, then stage 1, etc.
A large number of stages is possible and with the timeout is 1-3 minutes, that can cause
a large waiting time after the Coordinator long gone.

The new 'lastCoordinatorHeartbeatMs' is only updated by the messages sent by the
Coordinator, which results in all Tasks being cancelled at the same time after just single
timeout period.
It is used to determine if Coordinator has abandoned the Task.

## Motivation and Context
Get rid of the Tasks, which were abandoned by the Coordinator, instead of leaving them
running forever.

## Test Plan
Orchestrated the restart in the cluster with a long running query.
Observed prompt restart of the workers compared with the previous version.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/PrestoTask.h
+++ b/presto-native-execution/presto_cpp/main/PrestoTask.h
@@ -95,8 +95,15 @@ struct PrestoTask {
   /// has not been started, until the actual 'create task' message comes.
   bool taskStarted{false};
 
+  /// Time point (in ms) when the last message (any) came for this task.
+  // TODO (spershin): Deprecate it, use only the 'lastCoordinatorHeartbeatMs'.
   uint64_t lastHeartbeatMs{0};
+  /// Time point (in ms) when the last message came for this task from the
+  /// Coordinator. Used to determine if the Task has been abandoned.
+  uint64_t lastCoordinatorHeartbeatMs{0};
+  /// Time point (in ms) when the time we updated Task stats.
   uint64_t lastTaskStatsUpdateMs = {0};
+
   uint64_t lastMemoryReservation = {0};
   uint64_t createTimeMs{0};
   uint64_t firstSplitStartTimeMs{0};
@@ -132,9 +139,20 @@ struct PrestoTask {
   /// Updates when this task was touched last time.
   void updateHeartbeatLocked();
 
+  /// Updates time point (ms) when this task was touched last time by a message
+  /// from the Coordinator.
+  void updateCoordinatorHeartbeat();
+  void updateCoordinatorHeartbeatLocked();
+
   /// Returns time (ms) since the task was touched last time (last heartbeat).
   /// Returns zero, if never (shouldn't happen).
   uint64_t timeSinceLastHeartbeatMs() const;
+
+  /// Returns time (ms) since the task was touched last time by a message from
+  /// the Coordinator.
+  /// If above never happened, returns time since the task start or zero, if
+  /// task never started.
+  uint64_t timeSinceLastCoordinatorHeartbeatMs() const;
 
   protocol::TaskStatus updateStatus() {
     std::lock_guard<std::mutex> l(mutex);
--- a/presto-native-execution/presto_cpp/main/TaskManager.cpp
+++ b/presto-native-execution/presto_cpp/main/TaskManager.cpp
@@ -43,7 +43,7 @@ void cancelAbandonedTasksInternal(const TaskMap& taskMap, int32_t abandonedMs) {
   for (const auto& [id, prestoTask] : taskMap) {
     if (prestoTask->task != nullptr) {
       if (prestoTask->task->isRunning()) {
-        if (prestoTask->timeSinceLastHeartbeatMs() >= abandonedMs) {
+        if (prestoTask->timeSinceLastCoordinatorHeartbeatMs() >= abandonedMs) {
           LOG(INFO) << "Cancelling abandoned task '" << id << "'.";
           prestoTask->task->requestCancel();
         }
@@ -364,6 +364,7 @@ std::unique_ptr<TaskInfo> TaskManager::createOrUpdateErrorTask(
   auto prestoTask = findOrCreateTask(taskId, startProcessCpuTime);
   {
     std::lock_guard<std::mutex> l(prestoTask->mutex);
+    prestoTask->updateCoordinatorHeartbeatLocked();
     prestoTask->updateHeartbeatLocked();
     if (prestoTask->error == nullptr) {
       prestoTask->error = exception;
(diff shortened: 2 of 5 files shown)


presto-native-execution/presto_cpp/main/TaskManager.h
```diff
@@ -171,7 +171,7 @@ class TaskManager {
   // coordinator for a considerable time.
   void cancelAbandonedTasks();
 
-  std::unique_ptr<protocol::TaskInfo> createOrUpdateTask(
+  std::unique_ptr<protocol::TaskInfo> createOrUpdateTaskImpl(
```
xiaoxmeng, line 174: Why rename to createOrUpdateTaskImpl? Shall we rename findOrCreateTask as well? Thanks!

presto-native-execution/presto_cpp/main/TaskManager.cpp
```diff
@@ -780,6 +784,7 @@ folly::Future<std::unique_ptr<protocol::TaskInfo>> TaskManager::getTaskInfo(
   {
     std::lock_guard<std::mutex> l(prestoTask->mutex);
     prestoTask->updateHeartbeatLocked();
+    prestoTask->updateCoordinatorHeartbeatLocked();
```
xiaoxmeng, line 787: Can we pass a flag to updateHeartbeatLocked(fromCoordinator)?

presto-native-execution/presto_cpp/main/PrestoTask.h
```diff
@@ -95,8 +95,14 @@ struct PrestoTask {
   /// has not been started, until the actual 'create task' message comes.
   bool taskStarted{false};
 
+  /// Time point (in ms) when the last message (any) came for this task.
```
xiaoxmeng, line 98: Can we add unit test in PrestoTaskTest.cpp to cover this heartbeat timestamp? Thanks!

presto-native-execution/presto_cpp/main/PrestoTask.h
```diff
@@ -95,8 +95,14 @@ struct PrestoTask {
...
   uint64_t lastHeartbeatMs{0};
```
xiaoxmeng, line 99:
```
s/lastHeartbeatMs/lastWorkerHeartbeatMs/
```
Do we need to track the last heartbeat or message received from worker? Or shall we rename it to lastWorkerMsgTimeMs?

presto-native-execution/presto_cpp/main/TaskManager.cpp
```diff
@@ -468,6 +469,7 @@ std::unique_ptr<TaskInfo> TaskManager::createOrUpdateTask(
   auto prestoTask = findOrCreateTask(taskId, startProcessCpuTime);
   {
     std::lock_guard<std::mutex> l(prestoTask->mutex);
+    prestoTask->updateCoordinatorHeartbeatLocked();
```
xiaoxmeng, line 472: Why we don't update updateCoordinatorHeartbeatLocked in findOrCreateTask? thanks!

xiaoxmeng (commented):
@spershin thanks for fixing this % comments.

presto-native-execution/presto_cpp/main/TaskManager.h
spershin: Because we had two createOrUpdateTask() and w=one of them is internal.

presto-native-execution/presto_cpp/main/TaskManager.cpp
spershin: No, we cannot.
The sites we call these two differ.

presto-native-execution/presto_cpp/main/PrestoTask.h
spershin: I don't believe we can easily organize a unit test for it.
This will require orchestrating TaskManager and re-implementing some sort of TaskResource to emulate HTTP messages.

presto-native-execution/presto_cpp/main/PrestoTask.h
spershin: I don't think we should do this renaming in this change.
It is unnecessary and is actually unclear why you are suggesting it.

> Do we need to track the last heartbeat or message received from worker? Or shall we rename it to lastWorkerMsgTimeMs?

If we need this - it would be a separate PR.

presto-native-execution/presto_cpp/main/TaskManager.cpp
spershin: Because findOrCreateTask() is called from literally everywhere and we only need to cover the Coord's endpoints.

presto-native-execution/presto_cpp/main/PrestoTask.h
xiaoxmeng: We can create a dummy PrestoTask and verify the newly added API works as expected? Thanks!

presto-native-execution/presto_cpp/main/PrestoTask.h
xiaoxmeng: I am wondering if we need to update lastHeartbeatMs when we receive message from worker? Or we only need to record the heartbeat timestamp from coordinator? Thanks!

presto-native-execution/presto_cpp/main/PrestoTask.h
spershin: That would a useless unit test, IMHO.
The functions are extremely simple - no space for error. I'm reluctant to add test for this not because I don't want it, but because we have so many already and this one would be near useless.
Like they are literally one-liners:
`lastCoordinatorHeartbeatMs = velox::getCurrentTimeMs();`
or
```
  if (lastCoordinatorHeartbeatMs == 0UL) {
    return 0UL;
  }
  return getCurrentTimeMs() - lastCoordinatorHeartbeatMs;
```
The bugs can sneak in only from the workflow in TaskManager, which would be really hard to re-create in a test.

presto-native-execution/presto_cpp/main/PrestoTask.h
spershin: > I am wondering if we need to update lastHeartbeatMs when we receive message from worker? Or we only need to record the heartbeat timestamp from coordinator? Thanks!

I don't know that - just keeping the existing heartbeat as a status quo in order not to break the current behavior.

What I can do is to follow up by looking at the java code and see if it has something similar and how it is updated.
Not sure what I will have time for it, because it is low pri.

xiaoxmeng (approved):
@spershin thanks for the offline discussion and clear code comments. Please consider to deprecate the old heartbeat ts in followup per our discussion.

xiaoxmeng (approved):
@spershin thanks for the test!

PR #22734, opened 2024-05-13
[native]Add native dynamic filter stats to operator stats for hbo

xiaoxmeng:
Exports the dynamic filter stats from native worker to coordinator:
1. add DynamicFilterStats in OperatorStats and update presto protocol. 
2. Presto task collects the dynamic filter stats from velox.


steveburnett:
I don't see documentation for me to review, perhaps I don't need to review this. Let me know if there's something I can help with! The description is empty, does this PR need a release note entry?

xiaoxmeng:
> I don't see documentation for me to review, perhaps I don't need to review this. Let me know if there's something I can help with! The description is empty, does this PR need a release note entry?

@steveburnett Hi, the pr description is updated.

presto-main/src/main/java/com/facebook/presto/operator/DynamicFilterStats.java
```diff
@@ -0,0 +1,73 @@
...
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Records the dynamic filter execution stats for an operator on native worker.
+ */
+@ThriftStruct
+public class DynamicFilterStats
+{
+    private final Set<PlanNodeId> producerNodeIds;
```
feilong-liu, line 33: These PlanNodeIds will be the PlanNodeId of the joins which has dynamic filter pushdown?
How do we know which nodes these filters are pushed down to?

xiaoxmeng: DynamicFilterStats is set in operator stats of those nodes that has dynamic filters.

feilong-liu: For example, if we have a query plan like:
join <- (probe side) project <- filter <- scan
where the dynamic filter pushdown is enabled, we should have project, filter and scan all in the producerNodeIds, right?

feilong-liu: Synced offline, the plan node ids will be the where the dynamic filter is from, which is the join node in the example I showed above.

feilong-liu (dismissed):
Thanks for the change!

presto-native-execution/presto_cpp/main/PrestoTask.cpp
```diff
@@ -552,6 +552,17 @@ void PrestoTask::updateMemoryInfoLocked(
...
+namespace {
+presto::protocol::DynamicFilterStats toPrestoDynamicFilterStats(
+    const velox::exec::OperatorStats& veloxOpStats) {
+  presto::protocol::DynamicFilterStats dynamicFilterStats;
+  for (const auto& nodeId : veloxOpStats.dynamicFilterStats.producerNodeIds) {
+    dynamicFilterStats.producerNodeIds.emplace_back(nodeId);
+  }
+  return dynamicFilterStats;
+}
+} // namespace
```
tanjialiang, line 564: Move to merge with top anonymous namespace for all helpers?

tanjialiang (approved):
Thanks %minor

feilong-liu (approved):

tanjialiang (approved):

Issue #22019, opened 2024-02-27
Presto Alluxio sdk Cache issue for file changes of the same s3 URI


diablo47:
we have used presto SDK cache for some time in version 0.275 with alluxio 2.9.3, cache might become invalid and can't be queried once in about 1-2 months and everything will be fine after manually clearing all the caches. So we decided to upgrade Presto + Alluxio to the latest release: presto 0.285.1 + Alluxio 304 for new features and bug fixes. But things seem to be worse:

we have some hive tables with no partition, the content of the table might be updated hourly or daily as we only care about the latest data. Queries and cache work fine for the first version of the files, after file content changes for the same file S3 URI, the table can't be queried anymore with exceptions. Queries can be resumed after manually emptying the cached files. 

The error type might be different, seems related to different files read
first seen: ` don't know what type: 15`
and then `Not valid Parquet file` 
and sometimes: `java.lang.ArrayIndexOutOfBoundsException`

our previous presto version 0.275 with Alluxio 2.9.3 doesn't have this issue, and changed files can be read successfully most of the time. 

We currently disabled cache for our 0.285.1 deployment.

## Your Environment

* Presto version used: 0.285.1 with alluxio version 304
* Storage (HDFS/S3/GCS..):S3
* Data source and connector used: hive + parquet 
* Deployment (Cloud or On-prem): native deploy on AWS ec2, java version: 1.8.0_181
* complete debug logs: see files attached

error_1: `Not valid Parquet file` 
[coordinator_error_1.log](https://github.com/prestodb/presto/files/14419068/coordinator_error_1.log)
[worker_error_1.log](https://github.com/prestodb/presto/files/14419071/worker_error_1.log)

error 2: `java.lang.ArrayIndexOutOfBoundsException`
[coordinator_error_2.log](https://github.com/prestodb/presto/files/14419084/coordinator_error_2.log)
[worker_error_2.log](https://github.com/prestodb/presto/files/14419085/worker_error_2.log)

error 3: ` don't know what type: 15`
[coordinator_error_3.log](https://github.com/prestodb/presto/files/14419365/coordinator_error_3.log)
[worker_error_3.log](https://github.com/prestodb/presto/files/14419366/worker_error_3.log)

## Expected Behavior
changed files should be read successfully and the cache should be updated accordingly. 

## Current Behavior
after file content changes for the same file S3 URI, the table can't be queried anymore with exceptions. Queries can be resumed after manually emptying the cached files. 

## Steps to Reproduce

[...]

PR #22750, opened 2024-05-15
Implement cache refresh after file changed

beinan:
fix issue https://github.com/prestodb/presto/issues/22019

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-cache/src/main/java/com/facebook/presto/cache/CacheFactory.java
+++ b/presto-cache/src/main/java/com/facebook/presto/cache/CacheFactory.java
@@ -31,30 +31,30 @@ public class CacheFactory
             ExtendedFileSystem fileSystem,
             CacheManager cacheManager,
             boolean cachingEnabled,
-            CacheType cacheType,
-            boolean validationEnabled)
+            CacheConfig cacheConfig)
             throws IOException
     {
         if (!cachingEnabled) {
             return fileSystem;
         }
 
-        checkState(cacheType != null);
+        checkState(cacheConfig != null && cacheConfig.getCacheType() != null);
 
-        switch (cacheType) {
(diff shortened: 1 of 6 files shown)


presto-cache/src/main/java/com/facebook/presto/cache/CacheFactory.java
```diff
@@ -31,30 +31,30 @@ public ExtendedFileSystem createCachingFileSystem(
...
-            boolean validationEnabled)
+            CacheConfig cacheConfig)
             throws IOException
     {
         if (!cachingEnabled) {
             return fileSystem;
         }
 
-        checkState(cacheType != null);
+        checkState(cacheConfig.getCacheType() != null);
```
ChunxuTang, line 41: nit: Do you also want to check whether `cacheConfig` is null or not?

ChunxuTang (approved):
LGTM.
@beinan Thanks for the fix!

ChunxuTang:
Re-running the failed test.

nmahadevuni:
@beinan Why don't we do this by default? So, we do not encounter such staleless problems at all.

presto-cache/src/main/java/com/facebook/presto/cache/CacheFactory.java
beinan: good catch,  just fixed it

beinan:
@zacw7 do you also want to take a look at this PR?  The change is behind a newly added flag which is disabled by default, so I think there should be no impact on Meta's production.

ChunxuTang (approved):

zacw7 (approved):
LGTM. Thanks for the fix!

beinan:
force push to rerun the test

PR #22751, opened 2024-05-15
[native] Remove spill stats reporting from presto native

tanjialiang:
Removes spill related stats reporting from presto native. Migrated the reporting to velox's PeriodicStatsReporter.
```
== RELEASE NOTES ==
Since there are metrics removed from reporting, dashboards might be affected. Migrate the dashboard to point to the velox reported metrics instead.
The removed spill related metrics are going to be replaced by velox reported metrics in the following way:
original: presto_cpp.spill_metrics_name
new: velox.spill_metrics_name
```


xiaoxmeng (approved):
@tanjialiang thanks!

PR #22765, opened 2024-05-15
Log detail name for optimizers of RowExpressionRewriteRuleSet and StatsRecordingPlanOptimizer

feilong-liu:
## Description
Currently optimizer log will record the class name of the optimizer, however this can be a problem for row expression optimizers, which extends RowExpressionRewriteRuleSet. All such optimizers will have the same name in the log, like "FilterRowExpressionRewrite", "ValuesRowExpressionRewrite" etc. These give no meaningful information on which optimizers are triggered.

This PR will log more verbose information for these optimizers, by including the rewriter class name.
For example instead of "ProjectRowExpressionRewrite" we will log RemoveMapCastRule$Rewriter:ProjectRowExpressionRewrite" now.

Similar for StatsRecordingPlanOptimizer, it's a wrapping optimizer, and it's the delegate optimizer within it doing the work. After the change, it will show as "StatsRecordingPlanOptimizer:HistoricalStatisticsEquivalentPlanMarkingOptimizer" instead of "StatsRecordingPlanOptimizer".

## Motivation and Context
To make optimizer log more meaningful.

## Impact
Improve optimizer log

## Test Plan
Test locally end to end

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Improve logging for RowExpressionRewriteRuleSet and StatsRecordingPlanOptimizer optimizers to include more information :pr:`22765`
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/IterativeOptimizer.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/IterativeOptimizer.java
@@ -34,6 +34,7 @@ import com.facebook.presto.spi.plan.PlanNodeIdAllocator;
 import com.facebook.presto.sql.planner.PlannerUtils;
 import com.facebook.presto.sql.planner.RuleStatsRecorder;
 import com.facebook.presto.sql.planner.TypeProvider;
+import com.facebook.presto.sql.planner.iterative.rule.RowExpressionRewriteRuleSet;
 import com.facebook.presto.sql.planner.optimizations.PlanOptimizer;
 import com.facebook.presto.sql.planner.optimizations.PlanOptimizerResult;
 import com.google.common.base.Splitter;
@@ -179,7 +180,7 @@ public class IterativeOptimizer
 
                 if (!rule.isEnabled(context.session)) {
                     if (isVerboseOptimizerInfoEnabled(context.session) && isApplicable(node, rule, matcher, context)) {
-                        context.addRulesApplicable(rule.getClass().getSimpleName());
+                        context.addRulesApplicable(getNameOfOptimizerRule(rule));
                     }
                     continue;
                 }
@@ -198,7 +199,7 @@ public class IterativeOptimizer
                             transformedNode = transformedNode.assignStatsEquivalentPlanNode(node.getStatsEquivalentPlanNode());
                         }
                     }
-                    context.addRulesTriggered(rule.getClass().getSimpleName(), node, transformedNode, rule.isCostBased(context.session), rule.getStatsSource());
+                    context.addRulesTriggered(getNameOfOptimizerRule(rule), node, transformedNode, rule.isCostBased(context.session), rule.getStatsSource());
                     node = context.memo.replace(group, transformedNode, rule.getClass().getName());
 
                     done = false;
@@ -208,6 +209,15 @@ public class IterativeOptimizer
         }
 
         return progress;
+    }
+
+    private String getNameOfOptimizerRule(Rule<?> rule)
+    {
+        String ruleName = rule.getClass().getSimpleName();
+        if (rule instanceof RowExpressionRewriteRuleSet.RowExpressionRewriteRule) {
+            ruleName = ((RowExpressionRewriteRuleSet.RowExpressionRewriteRule) rule).getOptimizerNameForLog();
+        }
+        return ruleName;
     }
 
     private <T> Rule.Result transform(PlanNode node, Rule<T> rule, Matcher matcher, Context context)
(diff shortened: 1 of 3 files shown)


jaystarshot:
This also happens for StatsRecordingPlanOptimizer, wondering if we could fix that too

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/IterativeOptimizer.java
```diff
@@ -210,6 +211,45 @@ private boolean exploreNode(int group, Context context, Matcher matcher)
...
+        else if (rule instanceof RowExpressionRewriteRuleSet.ValuesRowExpressionRewrite) {
+            ruleName = ((RowExpressionRewriteRuleSet.ValuesRowExpressionRewrite) rule).getOptimizerNameForLog();
+        }
+        else if (rule instanceof RowExpressionRewriteRuleSet.AggregationRowExpressionRewrite) {
+            ruleName = ((RowExpressionRewriteRuleSet.AggregationRowExpressionRewrite) rule).getOptimizerNameForLog();
+        }
+        else if (rule instanceof RowExpressionRewriteRuleSet.TableFinishRowExpressionRewrite) {
+            ruleName = ((RowExpressionRewriteRuleSet.TableFinishRowExpressionRewrite) rule).getOptimizerNameForLog();
+        }
+        else if (rule instanceof RowExpressionRewriteRuleSet.TableWriterRowExpressionRewrite) {
```
jaystarshot, line 247: On adding a new rule you will have to extend this, wondering if there was a better way

feilong-liu: Yeah, I was trying to find a better way to achieve this but cannot think of one. Open to suggestions if there are better ways.

feilong-liu:
> This also happens for `StatsRecordingPlanOptimizer` i.e optimizers which extend that show up as StatsRecordingPlanOptimizer, wondering if we could fix that too but that would be a different PR

This is not a big change, added in this PR too.

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/IterativeOptimizer.java
```diff
@@ -210,6 +211,45 @@ private boolean exploreNode(int group, Context context, Matcher matcher)
...
+        if (rule instanceof RowExpressionRewriteRuleSet.ProjectRowExpressionRewrite) {
+            ruleName = ((RowExpressionRewriteRuleSet.ProjectRowExpressionRewrite) rule).getOptimizerNameForLog();
+        }
+        else if (rule instanceof RowExpressionRewriteRuleSet.SpatialJoinRowExpressionRewrite) {
+            ruleName = ((RowExpressionRewriteRuleSet.SpatialJoinRowExpressionRewrite) rule).getOptimizerNameForLog();
+        }
+        else if (rule instanceof RowExpressionRewriteRuleSet.JoinRowExpressionRewrite) {
+            ruleName = ((RowExpressionRewriteRuleSet.JoinRowExpressionRewrite) rule).getOptimizerNameForLog();
+        }
+        else if (rule instanceof RowExpressionRewriteRuleSet.JoinRowExpressionRewrite) {
```
jaystarshot, line 226: This else is repeated with above

jaystarshot (dismissed):
Don't see an immediate solution to better log rules, maybe rules need to have a field with info about the optimizer which is using but this can be refactored if needed in future.

jaystarshot:
@ClarenceThreepwood  Do you have any other ideas?

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/IterativeOptimizer.java
ZacBlanco: Would the following patch work for you @feilong-liu ? I think you can just extend the rule interface to add a method which does this, and extend all of the rules using the class with the method instead, and check if it is an instance of the abstract class e.g.

```java
    public abstract class RowExpressionRewriteRule<T> implements Rule<T> {
        public String getOptimizerNameForLog() {
            String rewriterName = rewriter.getClass().getName();
            return format("%s:%s", rewriterName.substring(rewriterName.lastIndexOf('.') + 1), this.getClass().getSimpleName());
        }
    }

// rule definition
public final class ProjectRowExpressionRewrite
            extends RowExpressionRewriteRule<ProjectNode>
...
// name check 
        if (rule instanceof RowExpressionRewriteRuleSet.RowExpressionRewriteRule) {
            ruleName = ((RowExpressionRewriteRuleSet.RowExpressionRewriteRule) rule).getOptimizerNameForLog();
        }
```

[simplify-rulename.patch](https://github.com/prestodb/presto/files/15343575/simplify-rulename.patch)

feilong-liu: Good suggestion. Applied the change.

jaystarshot (approved):

PR #22785, opened 2024-05-19
Enhance FunctionCallRewriter

gggrace14:
Enhance FunctionCallRewriter to support function substitution for multi-signature functions. Change to use Multimap to support one function name mapped to multiple function substitutes. Select the first function substitute where the original function declaration matches.

Inherit the filter field and window fields of the original function by the aggregate or window functions in the substitute. Introduce the FunctionAndTypeManager to FunctionCallRewriter to identify aggregate and window functions.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/FunctionCallRewriter.java
+++ b/presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/FunctionCallRewriter.java
@@ -13,51 +13,129 @@
  */
 package com.facebook.presto.verifier.rewrite;
 
+import com.facebook.presto.common.type.TypeManager;
+import com.facebook.presto.metadata.FunctionAndTypeManager;
+import com.facebook.presto.spi.function.FunctionKind;
+import com.facebook.presto.spi.function.Signature;
+import com.facebook.presto.spi.function.SqlFunction;
+import com.facebook.presto.sql.ExpressionFormatter;
+import com.facebook.presto.sql.parser.ParsingException;
+import com.facebook.presto.sql.parser.SqlParser;
+import com.facebook.presto.sql.tree.ArrayConstructor;
+import com.facebook.presto.sql.tree.CurrentTime;
 import com.facebook.presto.sql.tree.Expression;
 import com.facebook.presto.sql.tree.ExpressionRewriter;
 import com.facebook.presto.sql.tree.ExpressionTreeRewriter;
 import com.facebook.presto.sql.tree.FunctionCall;
+import com.facebook.presto.sql.tree.Identifier;
+import com.facebook.presto.sql.tree.IfExpression;
+import com.facebook.presto.sql.tree.Literal;
 import com.facebook.presto.sql.tree.Node;
-import com.facebook.presto.sql.tree.QualifiedName;
+import com.facebook.presto.sql.tree.OrderBy;
 import com.facebook.presto.sql.tree.Query;
+import com.facebook.presto.sql.tree.SearchedCaseExpression;
+import com.facebook.presto.sql.tree.SimpleCaseExpression;
+import com.facebook.presto.sql.tree.SortItem;
+import com.facebook.presto.sql.tree.Statement;
 import com.facebook.presto.sql.tree.SubqueryExpression;
+import com.facebook.presto.sql.tree.Window;
+import com.google.common.base.Splitter;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableMultimap;
+import com.google.common.collect.Multimap;
 
 import java.util.ArrayList;
-import java.util.Arrays;
+import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
+import java.util.Optional;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
-import static com.google.common.collect.ImmutableList.toImmutableList;
+import static com.facebook.presto.verifier.framework.VerifierUtil.PARSING_OPTIONS;
+import static com.google.common.base.Preconditions.checkState;
 import static java.util.Objects.requireNonNull;
 
 public class FunctionCallRewriter
-        extends DefaultTreeRewriter<FunctionCallRewriter.Context>
 {
+    private static final List<Class<?>> SUPPORTED_ORIGINAL_FUNCTIONS = ImmutableList.of(FunctionCall.class, CurrentTime.class);
+    private static final List<Class<?>> SUPPORTED_SUBSTITUTE_EXPRESSIONS = ImmutableList.of(FunctionCall.class, IfExpression.class, SimpleCaseExpression.class,
+            SearchedCaseExpression.class, Identifier.class, Literal.class, ArrayConstructor.class);
+    private static final String OMIT_IDENTIFIER = "_";
+
+    private final Multimap<String, FunctionCallSubstitute> functionCallSubstituteMap;
+    private final FunctionAndTypeManager functionAndTypeManager;
+
+    private FunctionCallRewriter(Multimap<String, FunctionCallSubstitute> functionCallSubstituteMap, FunctionAndTypeManager functionAndTypeManager)
+    {
+        this.functionCallSubstituteMap = requireNonNull(functionCallSubstituteMap, "functionCallSubstituteMap is null.");
+        this.functionAndTypeManager = requireNonNull(functionAndTypeManager, "functionAndTypeManager is null");
+    }
+
+    public static Optional<FunctionCallRewriter> getInstance(Multimap<String, FunctionCallSubstitute> functionCallSubstitutes, TypeManager typeManager)
+    {
+        if (functionCallSubstitutes.isEmpty()) {
+            return Optional.empty();
+        }
+        checkState(typeManager instanceof FunctionAndTypeManager, "FunctionAndTypeManager is required for FunctionCallRewriter.");
+        return Optional.of(new FunctionCallRewriter(functionCallSubstitutes, (FunctionAndTypeManager) typeManager));
+    }
+
+    public static Multimap<String, FunctionCallSubstitute> validateAndConstructFunctionCallSubstituteMap(String functionCallSubstitutes)
+    {
+        ImmutableMultimap.Builder<String, FunctionCallSubstitute> map = ImmutableMultimap.builder();
+        if (functionCallSubstitutes == null) {
+            return map.build();
+        }
+
+        Splitter commaSplitter = Splitter.on("/,/").omitEmptyStrings().trimResults();
+        Splitter slashSplitter = Splitter.on('/').omitEmptyStrings().trimResults();
+        for (String substitute : commaSplitter.split(functionCallSubstitutes)) {
+            List<String> specs = slashSplitter.splitToList(substitute);
+            if (specs.size() != 2) {
+                throw new IllegalArgumentException(String.format("Original function call and substitute must both be specified, %s.", substitute));
+            }
+            Expression originalExpression = parseOriginalFunctionCall(specs.get(0));
+            Expression substituteExpression = parseSubstituteExpression(specs.get(1));
+
+            if (originalExpression instanceof FunctionCall) {
+                FunctionCall originalFunction = (FunctionCall) originalExpression;
+                map.put(originalFunction.getName().getSuffix(), new FunctionCallSubstitute(originalExpression, substituteExpression));
+            }
+            else if (originalExpression instanceof CurrentTime) {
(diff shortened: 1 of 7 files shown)


presto-verifier/src/main/java/com/facebook/presto/verifier/framework/VerifierConfig.java
```diff
@@ -407,4 +410,21 @@ public VerifierConfig setRunningMode(String runningMode)
...
+
+    @ConfigDescription("Specification of function substitutions, in the format of" +
+            " /foo(c0,_)/bar(c0)/,/fred(c0,c1)/baz(qux(c1,c0))/,/foobar(c0)/if(qux(c1),bar(c0),baz(c1))/,...," +
+            " where foo(c0, _) would be substituted by bar(c0), with the declared arguments applied" +
+            " to the corresponding positions. Concatenate function substitutions with comma.")
+    @Config("function-substitutes")
+    public VerifierConfig setFunctionSubstitutes(String functionSubstitutes)
+    {
+        this.functionSubstitutes = Optional.ofNullable(functionSubstitutes);
+        this.functionSubstitutes.ifPresent(FunctionCallRewriter::validateAndConstructFunctionCallSubstituteMap);
```
arhimondr, line 427: Since the pattern is parsed and validated here I wonder if it would be cleaner to to set the `functionSubstitutes` type to `Multimap<String, FunctionCallSubstitute>` and return it from the `getFunctionSubstitutes`. The `setFunctionSubstitutes` could still take `String`, it should be fine. With `Multimap` you also don't need optional. You can set it to an empty multimap by default.

presto-verifier/src/main/java/com/facebook/presto/verifier/framework/VerifierConfig.java
```diff
@@ -407,4 +410,21 @@ public VerifierConfig setRunningMode(String runningMode)
...
+    {
+        return functionSubstitutes;
+    }
+
+    @ConfigDescription("Specification of function substitutions, in the format of" +
+            " /foo(c0,_)/bar(c0)/,/fred(c0,c1)/baz(qux(c1,c0))/,/foobar(c0)/if(qux(c1),bar(c0),baz(c1))/,...," +
+            " where foo(c0, _) would be substituted by bar(c0), with the declared arguments applied" +
+            " to the corresponding positions. Concatenate function substitutions with comma.")
+    @Config("function-substitutes")
+    public VerifierConfig setFunctionSubstitutes(String functionSubstitutes)
```
arhimondr, line 424: You also need to update `TestVerifierConfig` with this property

presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/QueryRewriter.java
```diff
@@ -113,15 +113,15 @@ public QueryRewriter(
             PrestoAction prestoAction,
             Map<ClusterType, QualifiedName> tablePrefixes,
             Map<ClusterType, List<Property>> tableProperties,
-            Optional<String> nonDeterministicFunctionSubstitutes)
+            Optional<String> functionSubstitutes)
```
arhimondr, line 116: Similar comment here. Should the parameter be a `Multimap`?

presto-verifier/src/test/java/com/facebook/presto/verifier/rewrite/TestQueryRewriter.java
```diff
@@ -466,6 +512,71 @@ public void testRewriteFunctionCalls()
...
+                        "        PARTITION BY b\n" +
+                        "        ORDER BY b DESC\n" +
+                        "    )\n" +
+                        "FROM test_table");
+
+        // Test mapping of multiple substitutions and matching of literal arguments.
+        assertCreateTableAs(
+                queryRewriter.rewriteQuery(
+                        "SELECT\n" +
+                                "    APPROX_PERCENTILE(a, 0.95)\n" +
```
arhimondr, line 563: Maybe add a test with both `APPROX_PERCENTILE(v)` and `APPROX_PERCENTILE(v, e)` present in a single query

presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/FunctionCallRewriter.java
```diff
@@ -13,51 +13,140 @@
...
-    public static class FunctionCallSubstitute
+    private static final List<Class<?>> SUPPORTED_ORIGINAL_FUNCTIONS = ImmutableList.of(FunctionCall.class, CurrentTime.class);
+    private static final List<Class<?>> SUPPORTED_SUBSTITUTE_EXPRESSIONS = ImmutableList.of(FunctionCall.class, IfExpression.class, SimpleCaseExpression.class,
+            SearchedCaseExpression.class, Identifier.class, Literal.class, ArrayConstructor.class);
+    private static final String OMIT_IDENTIFIER = "_";
+
+    private final Multimap<String, FunctionCallSubstitute> functionCallSubstituteMap;
+    private final FunctionAndTypeManager functionAndTypeManager;
+
+    private RewriterContext context = new RewriterContext();
```
arhimondr, line 70: nit: final

presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/FunctionCallRewriter.java
```diff
@@ -13,51 +13,140 @@
...
 
-        public FunctionCallSubstitute(QualifiedName name, List<Integer> originalArgumentIndices)
-        {
-            this.name = requireNonNull(name, "name is null");
-            this.originalArgumentIndices = ImmutableList.copyOf(requireNonNull(originalArgumentIndices, "originalArgumentIndices is null"));
+    public static FunctionCallRewriter getInstance(String functionCallSubstitutes, TypeManager typeManager)
+    {
+        Multimap<String, FunctionCallSubstitute> functionCallSubstituteMap = validateAndConstructFunctionCallSubstituteMap(functionCallSubstitutes);
+        if (functionCallSubstituteMap.isEmpty()) {
+            return null;
```
arhimondr, line 82: nit: Optional (+ flatmap in `QueryRewriter` instead of `map`)

arhimondr (commented):
LGTM % a couple of nits and questions

presto-verifier/src/main/java/com/facebook/presto/verifier/framework/VerifierConfig.java
gggrace14: Change to this class actually belongs to the new PR. I will move it there and address the comment there. Had put it in the wrong PR when I split the change.

presto-verifier/src/main/java/com/facebook/presto/verifier/framework/VerifierConfig.java
gggrace14: Yes, I think it works. Revised

presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/FunctionCallRewriter.java
gggrace14: Oh, I have code like below
```java
class FunctionCallRewriter
{
    private RewriterContext context = new RewriterContext();
    ....
    public Node rewrite(Node root)
    {
        context = new RewriterContext();
        return new Rewriter(functionCallSubstituteMap, functionAndTypeManager).process(root, context);
    }
}
```
Would like each call of `functionCallRewriter.rewrite()` to have a new RewriterContext instance. RewriteContext at the moment is just used for keeping track of the actual function texts that are substituted and what they are substituted with.

presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/FunctionCallRewriter.java
gggrace14: Revised

presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/FunctionCallRewriter.java
```diff
@@ -13,51 +13,140 @@
...
+        return new FunctionCallRewriter(functionCallSubstituteMap, (FunctionAndTypeManager) typeManager);
+    }
 
-        public QualifiedName name()
-        {
-            return name;
+    public static Multimap<String, FunctionCallSubstitute> validateAndConstructFunctionCallSubstituteMap(String functionCallSubstitutes)
+    {
+        ImmutableMultimap.Builder<String, FunctionCallSubstitute> map = ImmutableMultimap.builder();
+        if (functionCallSubstitutes == null) {
```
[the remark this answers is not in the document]
gggrace14: The way we'll call is `validateAndConstructFunctionCallSubstituteMap()` is as below
```java
    @Config("function-substitutes")
    public VerifierConfig setFunctionSubstitutes(String functionSubstitutes)
    {
        this.functionSubstitutes = FunctionCallRewriter.validateAndConstructFunctionCallSubstituteMap(functionSubstitutes);
        return this;
    }
```
I guess String functionSubstitutes has value null if we don't pass the param --function-substitutes. And we don't want to throw exception in this case. Suppose I'm not missing anything.

presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/FunctionCallRewriter.java
gggrace14: No need to keep it as a field anymore after taking your suggestion from another comment.

presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/FunctionCallRewriter.java
```diff
@@ -13,51 +13,140 @@
...
+            }
+            else if (originalExpression instanceof CurrentTime) {
+                CurrentTime originalFunction = (CurrentTime) originalExpression;
+                map.put(originalFunction.getFunction().getName(), new FunctionCallSubstitute(originalExpression, substituteExpression));
+            }
+        }
+
+        Multimap<String, FunctionCallSubstitute> result = map.build();
+        if (result.isEmpty()) {
+            throw new IllegalArgumentException(String.format("%s has no valid function call substitution spec.", functionCallSubstitutes));
```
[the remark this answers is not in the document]
gggrace14: Revised

presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/FunctionCallRewriter.java
```diff
@@ -13,51 +13,140 @@
...
+        }
+
+        return result;
+    }
+
+    public Node rewrite(Node root)
+    {
+        context = new RewriterContext();
```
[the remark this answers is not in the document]
gggrace14: Take your suggestion and revised!

For this suggested approach of returning RewriteResult directly by the rewrite() method, I need to add one more field, processedFunctionCallSubstitutes, to the QueryObjectBundle class. The consumer of QueryObjectBundle is AbstractVerification.updateQueryInfoWithQueryBundle(). So I need to include more classes in this PR that calls QueryObjectBundle constructor.

Yeah, I also identified this implicit assumption is not a good pattern. Previously didn't figure out a good way to handle it.

presto-verifier/src/test/java/com/facebook/presto/verifier/rewrite/TestQueryRewriter.java
gggrace14: Good idea! Revised.

presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/QueryRewriter.java
gggrace14: Revised

presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/FunctionCallRewriter.java
```diff
@@ -70,133 +159,340 @@ public boolean equals(Object o)
...
+                }
 
-    public static Map<QualifiedName, FunctionCallSubstitute> constructFunctionCallSubstituteMap(String functionCallSubstitutes)
-    {
-        ImmutableMap.Builder<QualifiedName, FunctionCallSubstitute> map = ImmutableMap.builder();
-        if (functionCallSubstitutes == null) {
-            return map.build();
+                List<Expression> originalArguments = ((FunctionCall) substitution.originalExpression).getArguments();
+                List<Expression> instanceArguments = instance.getArguments();
+                if (originalArguments.size() > instanceArguments.size()) {
```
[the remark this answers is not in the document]
gggrace14: I think `=` means a match of the arg # of the original function pattern with the actual function instance that we're visiting. We should not skip this substitution in this case. This condition is to skip a substitution we're traversing.

arhimondr (dismissed):

arhimondr (approved):

PR #22803, opened 2024-05-21
Abstract client info for internal and external use

woshizhouyihao:
## Description
since we have client info for different use, abstract client info and verifier module

## Motivation and Context
More fields for client info needed to be added for internal; thus need to refactor client info and verifier module

## Impact
no impact on public api and user change

## Test Plan
Run integration tests to ensure no change made on OSS

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-verifier/src/main/java/com/facebook/presto/verifier/PrestoVerifyCommand.java
+++ b/presto-verifier/src/main/java/com/facebook/presto/verifier/PrestoVerifyCommand.java
@@ -16,6 +16,8 @@ package com.facebook.presto.verifier;
 import com.facebook.presto.sql.parser.SqlParserOptions;
 import com.facebook.presto.verifier.framework.AbstractVerifyCommand;
 import com.facebook.presto.verifier.framework.SourceQuery;
+import com.facebook.presto.verifier.prestoaction.ClientInfoFactory;
+import com.facebook.presto.verifier.prestoaction.DefaultClientInfoFactory;
 import com.facebook.presto.verifier.prestoaction.PrestoExceptionClassifier;
 import com.facebook.presto.verifier.prestoaction.SqlExceptionClassifier;
 import com.google.common.collect.ImmutableList;
@@ -71,5 +73,11 @@ public class PrestoVerifyCommand
     public SqlExceptionClassifier getSqlExceptionClassifier()
     {
         return PrestoExceptionClassifier.defaultBuilder().build();
+    }
+
+    @Override
+    public Class<? extends ClientInfoFactory> getClientInfoFactory()
+    {
+        return DefaultClientInfoFactory.class;
     }
 }
--- a/presto-verifier/src/main/java/com/facebook/presto/verifier/framework/AbstractVerifyCommand.java
+++ b/presto-verifier/src/main/java/com/facebook/presto/verifier/framework/AbstractVerifyCommand.java
@@ -44,7 +44,8 @@ public abstract class AbstractVerifyCommand
         Bootstrap app = new Bootstrap(ImmutableList.<Module>builder()
                 .add(new VerifierModule(
                         getSqlParserOptions(),
-                        getCustomQueryFilterClasses()))
+                        getCustomQueryFilterClasses(),
+                        getClientInfoFactory()))
                 .add(new SourceQueryModule(getCustomSourceQuerySupplierTypes()))
                 .add(new EventClientModule(getCustomEventClientTypes()))
                 .add(new QueryActionsModule(getSqlExceptionClassifier(), getCustomQueryActionTypes()))
--- a/presto-verifier/src/main/java/com/facebook/presto/verifier/prestoaction/ClientInfoFactory.java
+++ b/presto-verifier/src/main/java/com/facebook/presto/verifier/prestoaction/ClientInfoFactory.java
@@ -1,0 +1,22 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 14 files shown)


ericyuliu (approved):
LGTM, thanks for adding this feature!

arhimondr (dismissed):

presto-verifier/src/main/java/com/facebook/presto/verifier/PrestoVerifyCommand.java
```diff
@@ -31,6 +32,13 @@
 public class PrestoVerifyCommand
         extends AbstractVerifyCommand
 {
+    @Override
+    public Module getVerifierModule(SqlParserOptions sqlParserOptions,
```
arhimondr, line 36: Do you need an another extension point? Have you considered using an existing one? (for example `getAdditionalModules`)?

Or maybe simply add an another method `getClientInfo`.

presto-verifier/src/main/java/com/facebook/presto/verifier/prestoaction/ClientInfo.java
```diff
@@ -28,17 +28,14 @@ public class ClientInfo
 
     private final String testId;
     private final Optional<String> testName;
-    private final String sourceQueryName;
-    private final String suite;
-    private final String queryStage;
+    private String sourceQueryName;
```
arhimondr, line 31: Would it be possible to restructure code to avoid removing `final`? Generally we are trying to avoid using mutable data structures if not strongly necessary. 
Ha  you considered introducing an additional class to store the properties you want to provide via an extension?

presto-verifier/src/main/java/com/facebook/presto/verifier/prestoaction/JdbcPrestoAction.java
```diff
@@ -131,14 +128,13 @@ private <T> T execute(Statement statement, QueryStage queryStage, StatementExecu
...
-        String clientInfo = new ClientInfo(
-                testId,
-                testName,
-                verificationContext.getSourceQueryName(),
-                verificationContext.getSuite(),
-                queryStage.name()).serialize();
-
-        try (PrestoConnection connection = getConnection(queryStage, clientInfo)) {
+        String clientInfoJsonString = clientInfo
+                .setSourceQueryName(verificationContext.getSourceQueryName())
```
arhimondr, line 132: I'm not sure whether this is thread safe. From what I understand the `ClientInfo` instance is a singleton instance being injected.

woshizhouyihao: Checked in debug mode, each thread will have a new clientinfo instance. 
From code, each query will get a new instance of PrestoAction. So i think it is thread safe here.

presto-verifier/src/main/java/com/facebook/presto/verifier/prestoaction/ClientInfo.java
woshizhouyihao: it makes sense. introduced clientinfoFactory and keep final keyword

presto-verifier/src/main/java/com/facebook/presto/verifier/PrestoVerifyCommand.java
woshizhouyihao: as for naming and functionality, i feel it is better to have a separate module.

fields in clientinfo are retrieved from config.properties and VerifierConfig class has not initialized yet during bootstrap. Is there a way we can get those info at bootstrap step?

presto-verifier/src/main/java/com/facebook/presto/verifier/framework/AbstractVerifyCommand.java
```diff
@@ -44,7 +44,7 @@ public void run()
         Bootstrap app = new Bootstrap(ImmutableList.<Module>builder()
                 .add(new VerifierModule(
                         getSqlParserOptions(),
-                        getCustomQueryFilterClasses()))
+                        getCustomQueryFilterClasses(), getClientInfoFactory()))
```
arhimondr, line 47: nit: new line

presto-verifier/src/main/java/com/facebook/presto/verifier/prestoaction/ClientInfoFactory.java
```diff
@@ -0,0 +1,38 @@
...
+import com.google.inject.Injector;
+
+public class ClientInfoFactory
+{
+    private Injector injector;
+
+    @Inject
+    public void setInjector(Injector injector)
+    {
+        this.injector = injector;
```
arhimondr, line 29: It is usually discouraged to use the `Injector` directly as it may cause various initialization order related problems. From what I remember If the Injector is used directly the `Guice`  framework doesn't know statically what dependency needs to be satisfied via `injector.getInstance` and may initialize instances in a wrong order. For example the framework might decide to initialize `ClientInfoFactory` before the `VerifierConfig` is initialized. 

Currently it is more of a theoretical problem, because `ClientInfoFactory#create` is not called during initialization, but there's a chance someone might decide to use it during initialization causing an execution order dependent failure.

I would consider introducing an `ClientInfoFactory` interface with a single `ClientInfo create(VerificationContext verificationContext, QueryStage queryStage)` method. Then you can return a `Class<? extends ClientInfoFactory>` from the `VerifyCommand#getClientInfoFactory` with the default implementation (e.g.: `DefaultClientInfoFactory.class`). Then have the `VerifierConfig` injected directly into the `DefaultClientInfoFactory` constructor.

woshizhouyihao: Thanks for the review, yeh it makes sense to use @inject feature instead of using injector.getInstance() in create method to ensure initializing order. 
i also make clientinfoFactory as a singleton

arhimondr (approved):

arhimondr (approved):

PR #22823, opened 2024-05-24
Allow disabling UI module dependency

ZacBlanco:
## Description

For those who wish to disable the UI locally, deactivate this profile in order to prevent building the module. Add `-P '!ui'`to your maven command.

## Motivation and Context

Allow users to skip building the UI module

## Impact

N/A

## Test Plan

Local tests to confirm this work both ways with clean builds.

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```


ZacBlanco:
@rschlussel Can you try this out and see if it works for you?

presto-main/pom.xml
```diff
@@ -579,6 +574,21 @@
     </build>
 
     <profiles>
+        <profile>
+            <!-- Requires the UI as a dependency by default. If the UI is   -->
```
elharo, line 578: This feels wonky to me. If the code isn't dependent, why ever mark it as so? If presto-ui and presto-everything-else are really completely separate projects the build should reflect that.

ZacBlanco: It's done in order to make sure that the jar's files are included on the classpath of the `presto-main` jar. While the code itself isn't dependent, `presto-main` does depend on the resources being available at their expected location on the classpath in order to serve the UI.

The constraints I was working with were: (1) The UI must still work when launched in a dev environment e.g. via the Hive/Iceberg query runners and with the `PrestoServer`. (2) The UI module should be included/built by default by default. I want to make sure the original behavior is preserved.

If it feels wonky to you, maybe it might also make sense, according to the maven documentation, to define the dependency as `runtime` scope.

> runtime
This scope indicates that the dependency is not required for compilation, but is for execution. Maven includes a dependency with this scope in the runtime and test classpaths, but not the compile classpath.
test

I would argue that the resource jar here falls under the definition of a `runtime` dependency. However, I had trouble getting this to work in the way I expected with the `runtime` scope applied. I am more than happy to try a different approach if you have suggestions

elharo: If it's needed at runtime and not compile time or test time, then yes, runtime is the natural scope for the dependency. 

You say "It's done in order to make sure that the jar's files are included on the classpath of the presto-main jar." Does presto-ui produce jars? There doesn't seem to be any Java code there. Is it packaged into a jar file somewhere from which the resources are loaded?

It's possible that we're getting tripped up by the cross-language build. And perhaps presto-main is more tightly coupled to presto-ui than it should or needs to be. Or that we're stretching Maven dependency management into areas it wasn't meant to go. Maven was designed to build and test Java source code. It's not as extensive as, for example, bazel; and it's not really suited for runtime configuration.

ZacBlanco: `presto-ui` does produce a jar. If you inspect the output jar in the target directory after building the module you'll see:

```console
$ jar -tf ./presto-ui/target/presto-ui-0.288-SNAPSHOT.jar         
META-INF/
META-INF/MANIFEST.MF
webapp/
webapp/tableau/
webapp/dev/
webapp/assets/
webapp/vendor/
webapp/vendor/tableau/
webapp/vendor/bootstrap/
webapp/vendor/bootstrap/css/
webapp/vendor/bootstrap/js/
webapp/vendor/bootstrap/fonts/
...
```

The `presto-ui` module uses a plugins which bind to the `compile` phase to do the compilation of the UI assets (HTML, javascript, CSS files) into maven's expected output directory structure. Those assets are included into a jar using maven's standard resource descriptor tags. The final jar contains just the UI assets. To serve the UI at runtime, the presto HTTP server searches for resources at a pre-defined location for the UI assets. As long as the jar with the assets is included on the classpath, the server will be able to find them and serve them back to the user.

Having resources within a jar is common practice and not something specific to maven. The Java 8 SE docs have a [dedicated section on it](https://docs.oracle.com/javase/8/docs/technotes/guides/lang/resources.html). I don't think this falls out of the purview of maven. It's very common to include some additional assets into a JAR file.

elharo: OK, Thanks for the explanation.

Given that, I don't see any reason this shouldn't work with only a runtime dependency. What failures do you see if the dependency has runtime scope?

ZacBlanco: The original failures I saw were just that in some cases that the UI assets wouldn't be found after the maven compilation. I did test a few clean builds this morning with the added `runtime` scope and it seems to work fine. I may have mixed up some of my previous tests with `provider` and `optional` scopes. It seems like the `runtime` scope is working fine though

steveburnett:
Should the `-P '!ui'` option be documented, maybe in the [Building Presto](https://github.com/prestodb/presto/blob/master/CONTRIBUTING.md#building-presto) topic of CONTRIBUTING.md?

steveburnett (approved):
LGTM! (docs) 

View udpated file in GitHub, everything looks good. Thanks!

ZacBlanco:
Thanks for pointing that out @steveburnett! I added a new commit on this PR which updates the `CONTRIBUTING.md` instructions for the UI as well as `presto-console.rst`.

steveburnett:
> Thanks for pointing that out @steveburnett! I added a new commit on this PR which updates the `CONTRIBUTING.md` instructions for the UI as well as `presto-console.rst`.

Reviewed and approved! Thanks for the documentation: it seemed a pity to add a new feature but then not tell anyone about it.

tdcmeehan (approved):
Thank you!

PR #22857, opened 2024-05-29
[Native] Modify setup-adapters.sh to install prometheus-cpp

karteekmurthys:
## Description
Adds promtheus-cpp library dependency required by this PR: https://github.com/prestodb/presto/pull/22360/.

## Motivation and Context
This change is required to setup Circle CI images so that we can run the unit test defined in https://github.com/prestodb/presto/pull/22360/.


majetideepak (commented):
You have to include `install_prometheus_cpp=1` in the 43-46 line block with comment
`# Install all adapters by default`

majetideepak (approved):

PR #22858, opened 2024-05-29
Add parquet compression codec in Iceberg table metadata

agrawalreetika:
## Description
Add parquet compression codec in Iceberg table metadata

## Motivation and Context
Currently Iceberg metadata property `write.parquet.compression-codec` is being set to `zstd` from Iceberg library here https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/TableMetadata.java#L96 irrespective of what's being set from Presto while Iceberg table creation.
This change is to make `write.parquet.compression-codec` metadata update consistent in Iceberg metadata file same as Presto engine. 

## Impact
Make the parquet compression codec similar to what is set in Presto when the Iceberg table is created via Presto

## Test Plan
When Iceberg table is created from Presto Iceberg connector by setting - 
`set session iceberg.compression_codec='NONE';`

Iceberg metadata after the change -
image

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergUtil.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergUtil.java
@@ -129,6 +129,7 @@ import static com.facebook.presto.iceberg.IcebergErrorCode.ICEBERG_INVALID_PARTI
 import static com.facebook.presto.iceberg.IcebergErrorCode.ICEBERG_INVALID_SNAPSHOT_ID;
 import static com.facebook.presto.iceberg.IcebergErrorCode.ICEBERG_INVALID_TABLE_TIMESTAMP;
 import static com.facebook.presto.iceberg.IcebergPartitionType.IDENTITY;
+import static com.facebook.presto.iceberg.IcebergSessionProperties.getCompressionCodec;
 import static com.facebook.presto.iceberg.IcebergSessionProperties.isMergeOnReadModeEnabled;
 import static com.facebook.presto.iceberg.IcebergTableProperties.getCommitRetries;
 import static com.facebook.presto.iceberg.IcebergTableProperties.getFormatVersion;
@@ -177,6 +178,8 @@ import static org.apache.iceberg.TableProperties.DELETE_MODE;
 import static org.apache.iceberg.TableProperties.DELETE_MODE_DEFAULT;
 import static org.apache.iceberg.TableProperties.FORMAT_VERSION;
 import static org.apache.iceberg.TableProperties.MERGE_MODE;
+import static org.apache.iceberg.TableProperties.ORC_COMPRESSION;
+import static org.apache.iceberg.TableProperties.PARQUET_COMPRESSION;
 import static org.apache.iceberg.TableProperties.UPDATE_MODE;
 import static org.apache.iceberg.TableProperties.WRITE_LOCATION_PROVIDER_IMPL;
 import static org.apache.iceberg.types.Type.TypeID.BINARY;
@@ -1037,12 +1040,20 @@ public final class IcebergUtil
         }
     }
 
-    public static Map<String, String> populateTableProperties(ConnectorTableMetadata tableMetadata, FileFormat fileFormat)
+    public static Map<String, String> populateTableProperties(ConnectorTableMetadata tableMetadata, FileFormat fileFormat, ConnectorSession session)
     {
         ImmutableMap.Builder<String, String> propertiesBuilder = ImmutableMap.builderWithExpectedSize(5);
         Integer commitRetries = getCommitRetries(tableMetadata.getProperties());
         propertiesBuilder.put(DEFAULT_FILE_FORMAT, fileFormat.toString());
         propertiesBuilder.put(COMMIT_NUM_RETRIES, String.valueOf(commitRetries));
+        switch (fileFormat) {
+            case PARQUET:
+                propertiesBuilder.put(PARQUET_COMPRESSION, getCompressionCodec(session).getParquetCompressionCodec().get().toString());
+                break;
+            case ORC:
+                propertiesBuilder.put(ORC_COMPRESSION, getCompressionCodec(session).getOrcCompressionKind().name());
+                break;
+        }
         if (tableMetadata.getComment().isPresent()) {
             propertiesBuilder.put(TABLE_COMMENT, tableMetadata.getComment().get());
         }
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergHiveMetadata.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergHiveMetadata.java
@@ -301,7 +301,7 @@ public class IcebergHiveMetadata
         }
 
         FileFormat fileFormat = getFileFormat(tableMetadata.getProperties());
-        TableMetadata metadata = newTableMetadata(schema, partitionSpec, targetPath, populateTableProperties(tableMetadata, fileFormat));
+        TableMetadata metadata = newTableMetadata(schema, partitionSpec, targetPath, populateTableProperties(tableMetadata, fileFormat, session));
         transaction = createTableTransaction(tableName, operations, metadata);
 
         return new IcebergWritableTableHandle(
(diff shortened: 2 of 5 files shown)


presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergUtil.java
```diff
@@ -1037,12 +1039,13 @@ public void close()
...
     }
 
-    public static Map<String, String> populateTableProperties(ConnectorTableMetadata tableMetadata, FileFormat fileFormat)
+    public static Map<String, String> populateTableProperties(ConnectorTableMetadata tableMetadata, FileFormat fileFormat, ConnectorSession session)
     {
         ImmutableMap.Builder<String, String> propertiesBuilder = ImmutableMap.builderWithExpectedSize(5);
         Integer commitRetries = getCommitRetries(tableMetadata.getProperties());
         propertiesBuilder.put(DEFAULT_FILE_FORMAT, fileFormat.toString());
         propertiesBuilder.put(COMMIT_NUM_RETRIES, String.valueOf(commitRetries));
+        propertiesBuilder.put(PARQUET_COMPRESSION, getCompressionCodec(session).getParquetCompressionCodec().get().toString());
```
imjalpreet, line 1048: We should add this property only if the file format is Parquet.

I can also see table properties for ORC(`write.orc.compression-codec`) and AVRO(`write.avro.compression-codec`). We should set those as well based on the file format selected.

imjalpreet (changes requested):
Thanks for the fix, minor change to set the table property based on file format.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergUtil.java
hantangwangd: It seems the iceberg community don't have consensus about whether this property should appear when the file format isn't parquet. Currently it would always be set regardless the underlying file formats. Referring to: https://github.com/apache/iceberg/pull/9503#issuecomment-1919889634

And I wonder if this is the ground truth, then should ORC_COMPRESSION and AVRO_COMPRESSION be set as well?

imjalpreet: From what I understand from the discussion in https://github.com/apache/iceberg/pull/9503, iceberg sets the default value for that property regardless of the file format. But since we are overriding the default value here, should we not update it only if someone is creating/writing the iceberg table in parquet format from Presto? If the format set from Presto is not parquet, we let the property be as-is.

> And I wonder if this is the ground truth, then should ORC_COMPRESSION and AVRO_COMPRESSION be set as well?

I think they are still setting the ORC_COMPRESSION and AVRO_COMPRESSION when the formats are being used: https://github.com/apache/iceberg/blob/f9cdde2516daea6d503798ef62e5869b5e46c31d/spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/SparkWriteConf.java#L490

hantangwangd: Yes, you are right, and I agree with you, set the corresponding property base on file format is concise and easy to understand. I just a little confuse that Iceberg itself always set `PARQUET_COMPRESSION` for a new table regardless the underlying file formats, but do not set `ORC_COMPRESSION` and `AVRO_COMPRESSION`.

hantangwangd (commented):
One question, currently in presto, the actual write operation is based on the compression encoding set in session. So should we figure out which compression encoding we should follow when writing data into an Iceberg table?

If we still follow the compression encoding set in session, then the property `PARQUET_COMPRESSION` just means a default value rather than a mandatory attribute. It seems that Iceberg consider the property as this way.

If we strictly follow the table property `PARQUET_COMPRESSION`, then that may means a big change, and maybe need to introduce something like `NOT_SPECIFIED`, so that we do not specify any one and can follow the compression encoding in session, which could be changed now and then.

What's your thought? @agrawalreetika @tdcmeehan @imjalpreet @ZacBlanco

ZacBlanco:
It seems a little odd to me that the iceberg table has a property specifying the compression. The type of compression on a parquet file is specified in the parquet footer. The reader can just scan the parquet footer metadata to determine how to decompress it. I don't see much value in having the table specify the write format and compression level other than keeping multiple engines consistent in the format they write their data in.

The [commit which added it](https://github.com/apache/iceberg/commit/5c3699e377271deda1d7049a50306a62cdfdeb2c) also doesn't seem to have any discussion around how it is used nor does the Iceberg documentation. I checked the code and it doesn't really seem to be used for anything useful.

My opinion though is that on table creation we should write out the value of the session property. Then, for later writes respect the table property rather than the session property.

In the case of conflicting values between the session property and table property, maybe we should throw an error. Also, if the compression scheme or write format described by the table is not supported by Presto, we should probably also throw an error? This has downsides of making Presto seem like it fails/doesn't work when it could just be a user configuration error. I don't particularly like failing in those cases, but to me it seems like the correct thing to do if Iceberg specifies a particular value in the table configuration. Why have the property if the engine is going to ignore it?

imjalpreet:
If we have a look at what happens in the case of spark here: https://github.com/apache/iceberg/blob/f9cdde2516daea6d503798ef62e5869b5e46c31d/spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/SparkWriteConf.java#L556 and https://github.com/apache/iceberg/blob/f9cdde2516daea6d503798ef62e5869b5e46c31d/spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/SparkConfParser.java#L267 they are following the below priority order to decide what compression to use: Spark Write Options > Spark Session Config > Iceberg Table Property > Default Value (`zstd` since iceberg 1.4). I have only glanced till now, please correct me if I missed anything.

In our case, we only have session property as an option. I think we can keep using session property as the codec we want to use while writing. We can decide whether we want to update the table property for tables created via Presto so that if data is inserted later from spark or other engines they can use the same compression specified via table property if its not set via Spark write options or spark session. Otherwise, if a user uses `GZIP` from Presto while creating the table and inserts via spark later, `zstd` will be used by default.

ZacBlanco:
Thanks for investigating the Spark behavior @imjalpreet . I think if Spark does not respect it in all cases, then I am fine to mimic their behavior

agrawalreetika:
Thank you all for your input. My thought was that maintaining this table property same as what Presto offers as part of the session property (be it default or any specific one that is set) when the table is being created from the Presto engine. Even if Presto is writing ORC, Or AVRO data in the table as per my understanding `write.parquet.compression-codec` table property should not matter since it should be specific to PARQUET, but my thought was when the table is created via Presto I thought we can respect default of Presto instead.

Please let me know what we all agree on to maintain this property, and I can make the necessary changes.

hantangwangd:
> My thought was that maintaining this table property same as what Presto offers as part of the session property (be it default or any specific one that is set) when the table is being created from the Presto engine. 

Agree that maintaining this table property the same as what Presto offers when creating a table.

As @imjalpreet investigated and described, SPARK would set the corresponding property base on file format, referring https://github.com/apache/iceberg/blob/f9cdde2516daea6d503798ef62e5869b5e46c31d/spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/SparkWriteConf.java#L490 . 

So maybe it's better to follow SPARK's behavior, setting the corresponding property base on file format. Or there is no reason to only set the default compression property for PARQUET, but not for ORC. What's your opinion? @agrawalreetika

agrawalreetika:
> So maybe it's better to follow SPARK's behavior, setting the corresponding property base on file format. Or there is no reason to only set the default compression property for PARQUET, but not for ORC. What's your opinion? @agrawalreetika

Sounds fair to me. 
So to confirm we are saying, that if Presto creates an Iceberg table with ORC file format we can add `write.orc.compression-codec` to the compression present in the Presto session property. For Avro, we should add `write.avro.compression-codec`. Please let me know if I am missing anything here?

And since `write.parquet.compression-codec` is being set to zstd by deafult irrespective of file format so in case of ORC and Avro file format creation from Presto, do we want to leave this as it is to `zstd` or update it to compression present in the Presto session property?

hantangwangd (commented):
Thanks for your fix, change looks good to me, could you add some test cases to show that different session property and file_format would affect the corresponding Iceberg table property on creation?

hantangwangd (approved):

imjalpreet (approved):
LGTM

PR #22866, opened 2024-05-30
Support Cast expression in FunctionCallRewriter

gggrace14:
Add support to FunctionCallRewriter for Cast expression substitution.

```
== NO RELEASE NOTE ==
```


presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/FunctionCallRewriter.java
```diff
@@ -97,17 +98,10 @@ public static Multimap<String, FunctionCallSubstitute> validateAndConstructFunct
             if (specs.size() != 2) {
                 throw new IllegalArgumentException(String.format("Original function call and substitute must both be specified, %s.", substitute));
             }
-            Expression originalExpression = parseOriginalFunctionCall(specs.get(0));
+            Expression originalExpression = parseOriginalExpression(specs.get(0));
```
elharo, line 101: Why rename this method? I'm guessing because cast is an expression but not a function call, but then why is this class FunctionCallRewriter and not ExpressionRewriter? and SUPPORTED_ORIGINAL_FUNCTIONS instead of SUPPORTED_ORIGINAL_EXPRESSIONS?

presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/FunctionCallRewriter.java
```diff
@@ -475,9 +517,18 @@ private static Expression parseOriginalFunctionCall(String functionCallSpec)
...
                     }
                 }
-                throw new IllegalArgumentException(String.format("Argument of type %s from %s is not supported.", argument.getClass().getSimpleName(), functionCallSpec));
+                throw new IllegalArgumentException(String.format("Argument of type %s from %s is not supported.", argument.getClass().getSimpleName(), expressionSpec));
             });
         }
+        else if (expression instanceof Cast) {
+            Cast cast = (Cast) expression;
+            if (!(cast.getExpression() instanceof Identifier)) {
+                throw new IllegalArgumentException(String.format("Argument of type %s from %s is not supported.", cast.getExpression().getClass().getSimpleName(), expressionSpec));
```
elharo, line 526: In general simple string concatenation is less prone to errors from misplaced arguments than format strings.

presto-verifier/src/test/java/com/facebook/presto/verifier/rewrite/TestQueryRewriter.java
```diff
@@ -568,6 +569,17 @@ public void testRewriteFunctionCalls()
                         "    AVG(a),\n" +
                         "    REPEAT(AVG(a), CAST(CARDINALITY(ARRAY[0.5, 0.9]) AS INTEGER))\n" +
                         "FROM test_table");
+
```
elharo, line 572: This should be a separate and new test method that can pass or fail independently of the existing tests.

arhimondr (approved):

presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/FunctionCallRewriter.java
gggrace14: Renaming SUPPORTED_ORIGINAL_FUNCTIONS to SUPPORTED_ORIGINAL_FUNCTIONS.

Yes, renaming parseOriginalFunctionCall to parseOriginalExpression to be more accurate in code, as we will support rewriting more than function calls. Calling the class FunctionCallRewriter so that the user can tell what this class generally does intuitively.

presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/FunctionCallRewriter.java
gggrace14: Okay, changed the occurrences of String.format() to connected strings. However, I seem to see the instance of String.format()  all over the places in the repo. There might be perf consideration to not have string concatenation which creates multiple string instances.

elharo: This is a common example of cargo cult programming. Devs use format because they see others use it, and they think it must be somehow newer and better. It really isn't, though it does introduce a whole new class of bugs that don't appear with simple concatenation. There's no performance issue when the string is concatenated/formatted unconditionally as here in line 526. Either the code reaches line 526 and the string is constructed, or it doesn't get to that line and it isn't. The only case where format might be minimally faster is if the string is constructed before the VM knows if it will be used; e.g. in the argument list of Preconditions.checkArgument.

presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/FunctionCallRewriter.java
rschlussel: can we name the class expression rewriter if it's expanded beyond function calls? And similarly rename other things from FunctionCall to Expression (e.g. functionCallSubstitute -> ExpressionSubstitute, makeFunctionCallSubstituteMapKey -> makeExpressionSubstituteKey).

presto-verifier/src/test/java/com/facebook/presto/verifier/rewrite/TestQueryRewriter.java
```diff
@@ -571,6 +571,25 @@ public void testRewriteFunctionCalls()
                         "FROM test_table");
     }
 
+    @Test
+    public void testRewriteCasts()
+    {
+        VerifierConfig verifierConfig = new VerifierConfig().setFunctionSubstitutes(
+                "/cast(x as varchar)/cast(floor(x) as varchar)/");
```
rschlussel, line 578: I know this is just testing the technical rewrite functionality, but this wouldn't necessarily be valid sql depending on the type. do you plan to use typeof() for the real substitution to make this conditional (e.g. `if(typeof(x) is DOUBLE OR typeof(x) is float, cast(floor(x) as varchar), cast(x as varchar)`)?

gggrace14: Hi @rschlussel , yes, with real substitution in prod, we need the `typeof`, ex. `if(typeof(a) = 'double', cast(floor(cast(cast(a as varchar) as double)) as varchar), cast(a as varchar))`. Here I just put a very short pattern for this unit test.

presto-verifier/src/main/java/com/facebook/presto/verifier/rewrite/FunctionCallRewriter.java
gggrace14: Hi @rschlussel , I thought it over, and still feel it a bit strange if we call it Expression.* but only support two types of expression. We can call it FunctionCall.* and take Cast as an extension.

However, I'm not quite sure, as both of you point it out. What do you think? I can certainly refactor.

rschlussel: I would prefer to rename. It's okay that currently there are only 2 kinds of expressions it supports rewriting.

PR #22878, opened 2024-05-31
Add config metrics-reporting-enabled and report metrics by default

gggrace14:
Add a new config metrics-reporting-enabled with default true. It will then
registerStatsCounters() by default.

[#22699](https://github.com/prestodb/presto/pull/22699) removes the call of registerStatsCounters() in PrestoServer.cpp.
To restore the behavior, we should enable registerStatsCounters()
by default. We have an existing config runtime-metrics-collection-enabled
for registerStatsCounters(), with default false.
This change adds a new config name to be more accurate.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/common/Configs.h
+++ b/presto-native-execution/presto_cpp/main/common/Configs.h
@@ -335,8 +335,12 @@ class SystemConfig : public ConfigBase {
       "cache.velox.ttl-check-interval"};
   static constexpr std::string_view kUseMmapAllocator{"use-mmap-allocator"};
 
+  /// TODO: Deprecate. Please use metrics-reporting-enabled instead.
   static constexpr std::string_view kEnableRuntimeMetricsCollection{
       "runtime-metrics-collection-enabled"};
+
+  static constexpr std::string_view kEnableMetricsReporting{
+      "metrics-reporting-enabled"};
 
   /// Specifies the memory arbitrator kind. If it is empty, then there is no
   /// memory arbitration.
@@ -710,6 +714,8 @@ class SystemConfig : public ConfigBase {
   std::chrono::duration<double> cacheVeloxTtlCheckInterval() const;
 
   bool enableRuntimeMetricsCollection() const;
+
+  bool enableMetricsReporting() const;
 };
 
 /// Provides access to node properties defined in node.properties file.
--- a/presto-native-execution/presto_cpp/main/PrestoServer.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoServer.cpp
@@ -158,8 +158,9 @@ void PrestoServer::run() {
     baseVeloxQueryConfig->initialize(
         fmt::format("{}/velox.properties", configDirectoryPath_), true);
 
-    if (systemConfig->enableRuntimeMetricsCollection()) {
-      enableRuntimeMetricReporting();
+    if (systemConfig->enableMetricsReporting() ||
+        systemConfig->enableRuntimeMetricsCollection()) {
+      enableMetricsReporting();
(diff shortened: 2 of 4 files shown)


xiaoxmeng (approved):
@gggrace14 LGTM. Thanks!

amitkdutta:
@xiaoxmeng @gggrace14 Not sure if I understand it properly. Looks like we are adding another config and making default to `true`. What's the issue with previous metric which was false by default and passed via deployment scripts?

presto-native-execution/presto_cpp/main/common/Configs.h
```diff
@@ -335,9 +335,13 @@ class SystemConfig : public ConfigBase {
       "cache.velox.ttl-check-interval"};
   static constexpr std::string_view kUseMmapAllocator{"use-mmap-allocator"};
 
+  /// TODO: Deprecate. Please use metrics-reporting-enabled instead.
   static constexpr std::string_view kEnableRuntimeMetricsCollection{
       "runtime-metrics-collection-enabled"};
 
+  static constexpr std::string_view kEnableMetricsReporting{
+      "metrics-reporting-enabled"};
```
spershin, line 343: I wonder if we should name the new config prop something like stats-counters-reporting-enabled?

Otherwise it might still confuse people if we are talking about runtime metrics we report for the individual tasks.

spershin (commented):
Thanks for fixing this!

majetideepak:
>  Looks like we are adding another config and making default to true. What's the issue with previous metric which was false by default and passed via deployment scripts?

Same question as Amit. Why do we need a new config?

spershin:
> > Looks like we are adding another config and making default to true. What's the issue with previous metric which was false by default and passed via deployment scripts?
> 
> Same question as Amit. Why do we need a new config?

@majetideepak 

Because the previous one could be confused with the runtime metrics we return for tasks.
My proposal is to name the new parameter as close to what we enable/disable as possible, for example 'stats-counters-reporting-enabled'.

amitkdutta:
> [#22699](https://github.com/prestodb/presto/pull/22699) removes the call of registerStatsCounters() in PrestoServer.cpp.
> To restore the behavior, we should enable registerStatsCounters()
> by default.

The PR description says `https://github.com/prestodb/presto/pull/22699 removes the call of registerStatsCounters() in PrestoServer.cpp. To restore the behavior, we should enable registerStatsCounters() by default.` meaning something already broken. How renaming fixing the issue is confusing here.

amitkdutta (changes requested):
Lets wait until we realize what's broken. Also please add [native] prefix.

majetideepak:
> Because the previous one could be confused with the runtime metrics we return for tasks.

@spershin I did not follow this. The current change simply adds the new config to the same (renamed) function `enableMetricsReporting()`. How do the old and new configs differ?

```
if (systemConfig->enableMetricsReporting() ||
        systemConfig->enableRuntimeMetricsCollection()) {
      enableMetricsReporting();
    }
```

spershin:
@majetideepak 
There might have been some misunderstanding.
We found that our counters stopped reporting and found [this](https://github.com/prestodb/presto/pull/22699) change.
It might be that something else is amiss.

I'm not aware of the enableMetricsReporting() renaming and hence was proposing the rename the config property to something close to the old registerStatsCounters() name.

We have metric and especially runtime metrics reported for operators and tasks.
And calling this metrics / runtime metrics creates confusion.

Wanted to avoid such confusion.

amitkdutta:
Metrics are reported, but the name of the metric changed recently in velox. So this PR isn't required.

PR #22885, opened 2024-05-31
[docs] Add Presto C++ config properties doc

### LINKED ISSUE #22877: Clarify Presto C++ documentation to include required coordinator configuration properties
The following properties should be enabled in a Presto coordinator when running with native workers.  We should update our documentation, and ideally provide examples, to clarify this so Prestissimo is easier to pick up and use.

```
native-execution-enabled=true
optimizer.optimize-hash-generation=false
regex-library=RE2J
use-alternative-function-signatures=true
experimental.table-writer-merge-operator-enabled=false
```

# Significant feature configs

## `native-execution-enabled`
This property is required when running native workers to account for underlying differences in behavior from Java workers.

## `optimizer.optimize-hash-generation`
We must set this property to false because Velox does not support optimized hash generation, instead preferring to perform this optimization adaptively.

## `regex-library`
Only RE2J is currently supported by Velox.

## `use-alternative-function-signatures`
Certain aggregation functions use intermediate types which are generic and not compatible with Velox aggregation function intermediate types (such as approx_distinct, whose intermediate type is `VARBINARY`).  Dummy stubs have been developed in the Presto codebase that provides alternative stub implementations which return the correct intermediate type.

## `experimental.table-writer-merge-operator-enabled`
I am not familiar with the reason for this property being disabled.

# Optional feature configs (left here fore discussion, however will not be added to the documentation)

## `parse-decimal-literals-as-double`
This is set to true in `NativeQueryRunnerUtils`, however, I believe this is no longer necessary, as Velox now supports the decimal type.

## `offset-clause-enabled`
This is set to true in `NativeQueryRunnerUtils`, however, I don’t think this should make a difference in native clusters.

## `inline-sql-functions`

Although native implementations of SQL functions have been developed in the Velox project, it is not required to run Presto C++ with this flag enabled.  Enabling this flag will render [SQL functions](https://prestodb.io/docs/current/sql/create-function.html) inoperable, as these functions will not be resolved into their underlying expressions by the coordinator.

Clear documentation for Presto C++

Docs

## Possible Implementation
Docs

steveburnett:
## Description
Add documentation of Presto C++ configuration properties for coordinator and worker to [existing Presto C++ documentation in Presto](https://prestodb.io/docs/current/presto-cpp.html).

## Motivation and Context
Fixes #22877. 

## Impact
Documentation.

## Test Plan
Local docs build. Also CI. 

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Add :doc:`/presto_cpp/properties` documentation :pr:`22885`
```

### DIFF
--- a/presto-docs/src/main/sphinx/presto_cpp/properties.rst
+++ b/presto-docs/src/main/sphinx/presto_cpp/properties.rst
@@ -1,0 +1,147 @@
+===============================
+Presto C++ Properties Reference
+===============================
+
+This section describes Presto C++ configuration properties.
+
+The following is not a complete list of all configuration and
+session properties, and does not include any connector-specific
+catalog configuration properties. For information on catalog 
+configuration properties, see :doc:`Connectors </connector/>`.
+
+.. contents::
+    :local:
+    :backlinks: none
+    :depth: 1
+
+Coordinator Properties
+----------------------
+
+Set the following configuration properties for the Presto coordinator exactly 
+as they are shown in this code block to enable the Presto coordinator's use of 
+Presto C++ workers. 
+
+.. code-block:: none
+
+    native-execution-enabled=true
+    optimizer.optimize-hash-generation=false
+    regex-library=RE2J
+    use-alternative-function-signatures=true
+    experimental.table-writer-merge-operator-enabled=false
+
+These Presto coordinator configuration properties are described here, in 
+alphabetical order. 
+
+``experimental.table-writer-merge-operator-enabled``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``boolean``
+* **Default value:** ``true``
+
+  Merge TableWriter output before sending to TableFinishOperator. This property must be set to 
+  ``false``. 
+
+``native-execution-enabled``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``boolean``
+* **Default value:** ``false``
+
+  This property is required when running Presto C++ workers because of 
+  underlying differences in behavior from Java workers.
+
+``optimizer.optimize-hash-generation``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``boolean``
+* **Default value:** ``true``
+
+  Set this property to ``false`` when running Presto C++ workers.  
+  Velox does not support optimized hash generation, instead using a HashTable 
+  with adaptive runtime optimizations that does not use extra hash fields. 
+
+``regex-library``
+^^^^^^^^^^^^^^^^^
+
+* **Type:** ``type``
+* **Allowed values:** ``RE2J``
+* **Default value:** ``JONI``
+
+  Only `RE2J <https://github.com/google/re2j>`_ is currently supported by Velox.
+
+``use-alternative-function-signatures``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``boolean``
+* **Default value:** ``false``
+
+  Some aggregation functions use generic intermediate types which are 
+  not compatible with Velox aggregation function intermediate types. One  
+  example function is ``approx_distinct``, whose intermediate type is 
+  ``VARBINARY``. 
+  This property provides function signatures for built-in aggregation 
+  functions which are compatible with Velox.
+
+Worker Properties
+-----------------
+
+The configuration properties of Presto C++ workers are described here, in alphabetical order. 
+
+``async-data-cache-enabled``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``boolean``
+* **Default value:** ``true``
+
+  In-memory cache.
+
+``query.max-memory-per-node``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``integer``
+* **Default value:** ``4GB``
+
+  Max memory usage for each query.
+
+``query-memory-gb``
+^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``integer``
+* **Default value:** ``38``
+
+  The total memory capacity that can be used across all query executions. 
+  Memory for system usage such as disk spilling and cache prefetch which 
+  are not counted in query memory usage.
+
+``query-reserved-memory-gb``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``integer``
+* **Default value:** ``4``
+
+  Specifies the amount of query memory capacity reserved 
+  to ensure that each query has the minimal memory capacity to run. A query can 
+  only allocate from the reserved query memory if its current capacity is less 
+  than the minimal memory capacity as specified by 
+  ``memory-pool-reserved-capacity``. 
+  
+  The exceeding capacity must allocate from the non-reserved query memory.
+
+``system-memory-gb``
+^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``integer``
+* **Default value:** ``40``
+
+  Memory allocation limit enforced via internal memory allocator. 
+
+  Set ``system-memory-gb`` to the available machine memory of the deployment.
+
+``task.max-drivers-per-task``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``integer``
+* **Default value:** ``number of hardware CPUs``
+
+  Number of drivers to use per task. Defaults to hardware CPUs.
+
--- a/presto-docs/src/main/sphinx/presto-cpp.rst
+++ b/presto-docs/src/main/sphinx/presto-cpp.rst
@@ -9,6 +9,7 @@ Note: Presto C++ is in active development. See :doc:`Limitations </presto_cpp/li
 
     presto_cpp/features
     presto_cpp/limitations
+    presto_cpp/properties
 
 Overview
 ========


steveburnett:
Reviewers, please help me with the content - especially anything currently marked `Default` or `Definition` as placeholders. Thank you!

presto-docs/src/main/sphinx/presto_cpp/properties.rst
```diff
@@ -0,0 +1,140 @@
...
+* **Type:** ``boolean``
+* **Default value:** ``default``
+
+  Definition.
+
+``native-execution-enabled``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``boolean``
+* **Default value:** ``default``
```
imjalpreet, line 42: Default value is `false`

presto-docs/src/main/sphinx/presto_cpp/properties.rst
```diff
@@ -0,0 +1,140 @@
...
+    optimizer.optimize-hash-generation=false
+    regex-library=RE2J
+    use-alternative-function-signatures=true
+    experimental.table-writer-merge-operator-enabled=false
+
+``experimental.table-writer-merge-operator-enabled``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``boolean``
+* **Default value:** ``default``
```
imjalpreet, line 34: Default Value is `true`

presto-docs/src/main/sphinx/presto_cpp/properties.rst
```diff
@@ -0,0 +1,140 @@
...
+
+  Definition.
```
imjalpreet, line 36: Maybe something like - `Merge TableWriter output before sending to TableFinishOperator`

presto-docs/src/main/sphinx/presto_cpp/properties.rst
```diff
@@ -0,0 +1,140 @@
...
+* **Default value:** ``default``
+
+  This property is required when running Presto C++ workers because of 
+  underlying differences in behavior from Java workers.
+
+``optimizer.optimize-hash-generation``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``boolean``
+* **Default value:** ``default``
```
imjalpreet, line 51: Default Value is `true`

presto-docs/src/main/sphinx/presto_cpp/properties.rst
```diff
@@ -0,0 +1,140 @@
...
+  Set this property to ``false`` when running Presto C++ workers 
+  because Velox does not support optimized hash generation, 
+  instead performing this optimization adaptively.
+
+``regex-library``
+^^^^^^^^^^^^^^^^^
+
+* **Type:** ``type``
+* **Allowed values:** ``RE2J``
+* **Default value:** ``default``
```
imjalpreet, line 62: Default Value is `JONI`

presto-docs/src/main/sphinx/presto_cpp/properties.rst
```diff
@@ -0,0 +1,140 @@
...
+
+  Only `RE2J <https://github.com/google/re2j>`_ is currently supported by Velox.
+
+``use-alternative-function-signatures``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``boolean``
+* **Default value:** ``default``
```
imjalpreet, line 70: Default Value is `false`

presto-docs/src/main/sphinx/presto_cpp/properties.rst
```diff
@@ -0,0 +1,140 @@
...
+* **Type:** ``boolean``
+* **Default value:** ``true``
+
+  In-memory cache.
+
+``query.max-memory-per-node``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``integer``
+* **Default value:** ``default``
```
imjalpreet, line 96: Default Value is `4GB` https://github.com/prestodb/presto/blob/d728296e0963193821ba0264a0662fff1160c913/presto-native-execution/presto_cpp/main/common/Configs.cpp#L200

presto-docs/src/main/sphinx/presto_cpp/properties.rst
```diff
@@ -0,0 +1,140 @@
...
+
+  Max memory usage for each query.
+
+``query-memory-gb``
+^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``integer``
+* **Default value:** ``default``
```
imjalpreet, line 104: Default Value is `38`

presto-docs/src/main/sphinx/presto_cpp/properties.rst
```diff
@@ -0,0 +1,140 @@
...
+  than the minimal memory capacity as specified by 
+  ``memory-pool-reserved-capacity``. 
+  
+  The exceeding capacity must allocate from the non-reserved query memory.
+
+``system-memory-gb``
+^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``integer``
+* **Default value:** ``default``
```
imjalpreet, line 129: Default Value is `40`

steveburnett:
Thank you @imjalpreet for the review! I have updated with a new (squashed) commit, and I believe I have addressed all your comments. Let me know if I missed anything.

presto-docs/src/main/sphinx/presto_cpp/properties.rst
```diff
@@ -0,0 +1,140 @@
...
+
+.. code-block:: none
+
+    native-execution-enabled=true
+    optimizer.optimize-hash-generation=false
+    regex-library=RE2J
+    use-alternative-function-signatures=true
+    experimental.table-writer-merge-operator-enabled=false
+
+``experimental.table-writer-merge-operator-enabled``
```
agrawalreetika, line 30: This also has session property available `table_writer_merge_operator_enabled`, should we document that as well here? 
https://github.com/prestodb/presto/blob/master/presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java#L1146

presto-docs/src/main/sphinx/presto_cpp/properties.rst
```diff
@@ -0,0 +1,140 @@
...
+``native-execution-enabled``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``boolean``
+* **Default value:** ``false``
+
+  This property is required when running Presto C++ workers because of 
+  underlying differences in behavior from Java workers.
+
+``optimizer.optimize-hash-generation``
```
agrawalreetika, line 47: Session property for this is `optimize_hash_generation` - https://github.com/prestodb/presto/blob/master/presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java#L390

agrawalreetika (changes requested):

presto-docs/src/main/sphinx/presto_cpp/properties.rst
```diff
@@ -0,0 +1,140 @@
...
+session properties, and does not include any connector-specific
+catalog configuration properties. For information on catalog 
+configuration properties, see :doc:`Connectors </connector/>`.
+
+.. contents::
+    :local:
+    :backlinks: none
+    :depth: 1
+
+Coordinator Properties
```
aditi-pandit, line 17: @steveburnett : The co-ordinator properties have to be set exactly as specified. But the worker configs can be altered from the defaults. Do you feel we should document them separately so that the difference is made clear to the users ?

majetideepak: I think the statement below makes it clear that the coordinator configs have to be set as specified?

presto-docs/src/main/sphinx/presto_cpp/properties.rst
majetideepak: I don't think `experimental.table-writer-merge-operator-enabled` is supported. Hence it has to be set to `false` explicitly. `true` is not an option. We have to clarify that below.

presto-docs/src/main/sphinx/presto_cpp/properties.rst
majetideepak: I don't think we should encourage/document sessions properties for these. Users are better off setting them in the properties file as they are fixed. Advanced users will know the session option :).

presto-docs/src/main/sphinx/presto_cpp/properties.rst
```diff
@@ -0,0 +1,140 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``boolean``
+* **Default value:** ``false``
+
+  Some aggregation functions use generic intermediate types which are 
+  not compatible with Velox aggregation function intermediate types. One  
+  example function is ``approx_distinct``, whose intermediate type is 
+  ``VARBINARY``. 
+  Stubs in the Presto codebase provide alternative stub implementations
```
majetideepak, line 76: I would prefer to no use the word stub as it has other uses such as a placeholders.
Let's say `Presto codebase provides alternate function signatures optimized for Velox.`

presto-docs/src/main/sphinx/presto_cpp/properties.rst
```diff
@@ -0,0 +1,140 @@
...
+
+  The total memory capacity that can be used across all query executions. 
+  Memory for system usage such as disk spilling and cache prefetch which 
+  are not counted in query memory usage.
+
+``query-reserved-memory-gb``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``integer``
+* **Allowed values:** ``0``
```
majetideepak, line 114: Let's remove allowed values to be consistent with other configs.

presto-docs/src/main/sphinx/presto_cpp/properties.rst
```diff
@@ -0,0 +1,140 @@
...
+* **Default value:** ``4``
+
+  Value must be set to ``0``. Specifies the amount of query memory capacity reserved
```
majetideepak, line 117: There is no need for this value to be set to 0. Let's remove it.

presto-docs/src/main/sphinx/presto_cpp/properties.rst
```diff
@@ -0,0 +1,140 @@
...
+
+  Value must be set to ``0``. Specifies the amount of query memory capacity reserved 
+  to ensure that each query has the minimal memory capacity to run. A query can 
+  only allocate from the reserved query memory if its current capacity is less 
+  than the minimal memory capacity as specified by 
+  ``memory-pool-reserved-capacity``. 
+  
+  The exceeding capacity must allocate from the non-reserved query memory.
+
+``system-memory-gb``
```
majetideepak, line 125: This is very important to be configured right. New users often overlook this. They must set this to the available machine memory.
I see that the configs are ordered alphabetically. Can we add a note indicating that users must set this according to their deployment?

majetideepak: I will see if we can automate the inference of this value similar to `task.max-drivers-per-task` below.

presto-docs/src/main/sphinx/presto_cpp/properties.rst
tdcmeehan: I don't think `optimized for Velox` is accurate.  This is required to work, so I think `compatible with Velox` is more accurate.

`This property provides function signatures for built-in aggregation functions which are compatible with Velox.`

presto-docs/src/main/sphinx/presto_cpp/properties.rst
steveburnett: @aditi-pandit, the worker configs are documented separately in the Worker Properties topic. But I think I can make it more clear that the coordinator properties have to be set exactly as specified. Thanks!

presto-docs/src/main/sphinx/presto_cpp/properties.rst
steveburnett: > This is very important to be configured right. New users often overlook this. They must set this to the available machine memory. I see that the configs are ordered alphabetically. Can we add a note indicating that users must set this according to their deployment?

Of course we can add text to address this.

steveburnett: > I will see if we can automate the inference of this value similar to `task.max-drivers-per-task` below.

@majetideepak if you do automate the inference of this value, let me know if this property entry should then be removed from the documentation.

presto-docs/src/main/sphinx/presto_cpp/properties.rst
steveburnett: @agrawalreetika, @majetideepak, because these properties are fixed I understand @majetideepak's recommendation.

However: if these are ever likely to be capable of being set to something else, I'd like to consider adding the session properties to the doc now to reduce the work for future writers. 

Let me know what you think, please.

steveburnett:
@agrawalreetika, @majetideepak, @aditi-pandit, @imjalpreet, @tdcmeehan: Thanks for your quick reviews! I have addressed the feedback from everyone and pushed (and squashed) an update. 

I ask that you all take another look and re-review as you feel appropriate, please.

presto-docs/src/main/sphinx/presto_cpp/properties.rst
```diff
@@ -0,0 +1,146 @@
...
+as they are shown in this code block to enable the Presto coordinator's use of 
+Presto C++ workers. 
+
+.. code-block:: none
+
+    native-execution-enabled=true
+    optimizer.optimize-hash-generation=false
+    regex-library=RE2J
+    use-alternative-function-signatures=true
+    experimental.table-writer-merge-operator-enabled=false
```
aditi-pandit, line 30: There is also inline-sql-functions=false. We should add this as well.

presto-docs/src/main/sphinx/presto_cpp/properties.rst
majetideepak: Makes sense!

presto-docs/src/main/sphinx/presto_cpp/properties.rst
aditi-pandit: @tdcmeehan : As I research this further, seems like the SqlFunctions are internal developer concepts and not visible to the user.

Do you feel this should be documented as well ?

tdcmeehan: This property applies to both SQL functions created through `CREATE FUNCTION` and to functions defined through the function SPI via e.g. `@SqlInvokedScalarFunction`.  In both cases they are **not** internal developer concepts:

1) Functions defined through `CREATE FUNCTION` are user-defined.
2) Functions annotated as `@SqlInvokedScalarFunction` are part of the SPI.  One can add a new SPI module that adds more SQL functions.  Hence, it is not an internal developer concept either.

presto-docs/src/main/sphinx/presto_cpp/properties.rst
agrawalreetika: If available, we usually document the session property along with the config properties.
ref - https://prestodb.io/docs/current/admin/properties.html#experimental-spill-enabled 

And I think having a Note along with the property would be better if it has to be something specific for Prestissimo.  WDYT @majetideepak @steveburnett ?

steveburnett: Good point @agrawalreetika! We have an established pattern of documenting session properties as part of the configuration property in Presto, it makes sense that the Presto C++ docs should follow the same pattern to help readers. 

Also, while as @majetideepak says "Advanced users will know the session option", I would argue that we shouldn't be writing the doc for only advanced readers. Let's help readers become advanced, by making them aware of the options.

Here, we have a special context of config properties that must (at least currently) be set to a fixed value. To address this concern, we could document the session property by adding something like the following sentence to the definition part of ``optimizer.optimize_hash_generation``:

"Optionally, use the session property ``optimize_hash_generation = false``." 

(I wouldn't suggest following the exact phrasing in https://prestodb.io/docs/current/admin/properties.html#experimental-spill-enabled of "This config property can be overridden by the [x] property." because at this time, the value is fixed, and we don't want them to override the correctly set config property. But if, for some reason, they don't want to or can't set the config property the way we want them to, they can use the session property.)


tl;dr: I think we should document the session properties, and I think the way I'm suggesting addresses the raised issues. Let me know what you think, please.

presto-docs/src/main/sphinx/presto_cpp/properties.rst
steveburnett: Based on these comments, I am leaving inline-sql-functions out.

steveburnett:
@agrawalreetika, @majetideepak, @aditi-pandit, @imjalpreet, @tdcmeehan - I have addressed the feedback from everyone as best I understand the consensus, and I have pushed (and squashed) an update.

Please review the latest version and let me know what changes are needed.

presto-docs/src/main/sphinx/presto_cpp/properties.rst
```diff
@@ -0,0 +1,157 @@
...
+
+``optimizer.optimize-hash-generation``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``boolean``
+* **Default value:** ``true``
+
+  Set this property to ``false`` when running Presto C++ workers 
+  because Velox does not support optimized hash generation, 
+  instead performing this optimization adaptively.
```
majetideepak, line 65: I don't think we perform the hash generation optimization adaptively yet.

presto-docs/src/main/sphinx/presto_cpp/properties.rst
```diff
@@ -0,0 +1,157 @@
...
+alphabetical order. 
+
+``experimental.table-writer-merge-operator-enabled``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``boolean``
+* **Default value:** ``true``
+
+  Merge TableWriter output before sending to TableFinishOperator. This property must be set to 
+  ``false`` when used.
```
majetideepak, line 42: `when used` to me implies that setting this is optional. But this is one of the properties that must be set since the default is not supported by Prestissimo. Is my interpretation wrong?

majetideepak (commented):
@steveburnett 2 comments.

presto-docs/src/main/sphinx/presto_cpp/properties.rst
aditi-pandit: The wording is partially correct since HashTable is adaptive and so the extra hash fields are not needed. Maybe say 
"instead uses a HashTable with adaptive runtime optimizations"

majetideepak: I am not clear how the HashTable adaptive runtime optimizations make the extra hash fields redundant.
Can you explain?

presto-docs/src/main/sphinx/presto_cpp/properties.rst
aditi-pandit: Bit thrown off by this since this seems supported by Prestissimo but likely not PrestoOnSpark https://github.com/prestodb/presto/blob/master/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestWriter.java#L402

majetideepak: We should probably add a test to confirm.

presto-docs/src/main/sphinx/presto_cpp/properties.rst
aditi-pandit: In java, the hash function was expensive. The extra hash fields were added so that between say a join followed by an agg where the grouping key is the same as the join key, the hash of the keys could be reused.

Since Velox HashTable has adaptive runtime optimizations, the extra hash fields are not used.

majetideepak: It is not clear to me if these adaptive runtime optimizations `don't need` or `don't use` the extra hash fields. We should clarify in the doc here.

steveburnett: Using the comments here, I have revised to the following: 
 
"Set this property to ``false`` when running Presto C++ workers.  
  Velox does not support optimized hash generation, instead using a HashTable 
  with adaptive runtime optimizations that does not use extra hash fields. "

presto-docs/src/main/sphinx/presto_cpp/properties.rst
steveburnett: removed "when used".

steveburnett:
@majetideepak and @aditi-pandit, I've revised addressing your most recent discussions. Let me know if you'd like any more changes, please.

majetideepak (dismissed):
thanks, @steveburnett

steveburnett:
@agrawalreetika, I believe I addressed all of your requested changes. Would you review this again and let me know if you want anything else?

agrawalreetika (approved):

aditi-pandit (dismissed):
Thanks @steveburnett for the rewrites.

imjalpreet (dismissed):

presto-docs/src/main/sphinx/presto_cpp/properties.rst
```diff
@@ -0,0 +1,157 @@
...
+    :depth: 1
+
+Coordinator Properties
+----------------------
+
+Set the following configuration properties for the Presto coordinator exactly 
+as they are shown in this code block to enable the Presto coordinator's use of 
+Presto C++ workers. 
+
+.. code-block:: none
```
tdcmeehan, line 24: In my opinion, I think we should remove all references to session properties below.  

* Firstly, some of them are incorrect (e.g. whereas the _config property_ `experimental.table-writer-merge-operator-enabled` does exist and is valid, the _session property_ is `table_writer_merge_operator_enabled`, not `experimental.table-writer-merge-operator-enabled`).
* Secondly, session properties are set per query/session, and these are all invariants, so it's best to never toggle this.  Hence, they should be set as config properties only.

tdcmeehan (changes requested):

presto-docs/src/main/sphinx/presto_cpp/properties.rst
majetideepak: I proposed to remove references to session properties as well here https://github.com/prestodb/presto/pull/22885#discussion_r1626334112.
I agree that since these are invariants, they should be set only as config properties.
There was an argument made against this that existing doc norm is to document session properties if available.

tdcmeehan: For the purposes of this document, it's just irrelevant that a corresponding session property exists in my opinion.  In the configuration properties of our documentation we may mention that it can be toggled.  What do you think @steveburnett ?  I think just telling people that they can override this per-query is too much information, since it is **never** desirable to toggle it per-query.

steveburnett: I originally made the argument to document the session property for these reasons: 

* @agrawalreetika's point "If available, we usually document the session property along with the config properties." So my first argument for including session properties was consistency with the existing doc. 
* My assumption, which may be wrong, that one or more of these configuration properties could in future be useful for a user to be able to toggle with a session property - this is a big assumption and I'll come back to it in a moment. 
If these session properties were ever to be useful to the user in future, documenting them now would improve the stability of the documentation. 

If we currently expect that the session properties for these configuration properties are never to be useful to the user, I agree they should be removed from this draft, and I can easily do that today. 
(If that current expectation changes in the future, we can always document them then.)

tdcmeehan: For the purposes of this document, it won't ever be useful to toggle the session properties, so let's remove them.  I think it's OK to break from convention here.

steveburnett: Makes sense, I'll update the PR.

steveburnett:
@majetideepak and @tdcmeehan, I have pushed a squashed commit update that removes the session properties from the doc. Take another look and let me know what you think, please.

tdcmeehan (approved):
Thanks @steveburnett !

steveburnett:
Thank you @agrawalreetika, @majetideepak, @aditi-pandit, @imjalpreet, and @tdcmeehan for your help!

PR #22886, opened 2024-05-31
Fix boolean precedence

elharo:
## Description
&& is evaluated before ||. The opposite was needed, so use parentheses. 

## Motivation and Context
I'm not sure where this code is used, but it looks like a real bug. However, it probably didn't bite because we don't have any coercers from byte or short.

## Impact
??

## Test Plan
CI

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTES ==
```

### DIFF
--- a/presto-hive/src/main/java/com/facebook/presto/hive/HiveCoercer.java
+++ b/presto-hive/src/main/java/com/facebook/presto/hive/HiveCoercer.java
@@ -81,10 +81,10 @@ public interface HiveCoercer
         else if (fromType instanceof VarcharType && (toHiveType.equals(HIVE_BYTE) || toHiveType.equals(HIVE_SHORT) || toHiveType.equals(HIVE_INT) || toHiveType.equals(HIVE_LONG))) {
             return new VarcharToIntegerNumberCoercer(fromType, toType);
         }
-        else if (fromHiveType.equals(HIVE_BYTE) && toHiveType.equals(HIVE_SHORT) || toHiveType.equals(HIVE_INT) || toHiveType.equals(HIVE_LONG)) {
+        else if (fromHiveType.equals(HIVE_BYTE) && (toHiveType.equals(HIVE_SHORT) || toHiveType.equals(HIVE_INT) || toHiveType.equals(HIVE_LONG))) {
             return new IntegerNumberUpscaleCoercer(fromType, toType);
         }
-        else if (fromHiveType.equals(HIVE_SHORT) && toHiveType.equals(HIVE_INT) || toHiveType.equals(HIVE_LONG)) {
+        else if (fromHiveType.equals(HIVE_SHORT) && (toHiveType.equals(HIVE_INT) || toHiveType.equals(HIVE_LONG))) {
             return new IntegerNumberUpscaleCoercer(fromType, toType);
         }
         else if (fromHiveType.equals(HIVE_INT) && toHiveType.equals(HIVE_LONG)) {
--- a/presto-hive/src/test/java/com/facebook/presto/hive/TestCoercingFilters.java
+++ b/presto-hive/src/test/java/com/facebook/presto/hive/TestCoercingFilters.java
@@ -17,17 +17,31 @@ import com.facebook.presto.common.Subfield;
 import com.facebook.presto.common.predicate.TupleDomainFilter;
 import com.facebook.presto.common.predicate.TupleDomainFilter.BigintRange;
 import com.facebook.presto.common.predicate.TupleDomainFilter.BytesRange;
+import com.facebook.presto.common.type.TypeManager;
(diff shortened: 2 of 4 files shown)


presto-hive/src/main/java/com/facebook/presto/hive/HiveCoercer.java
```diff
@@ -81,10 +81,10 @@ static HiveCoercer createCoercer(TypeManager typeManager, HiveType fromHiveType,
         else if (fromType instanceof VarcharType && (toHiveType.equals(HIVE_BYTE) || toHiveType.equals(HIVE_SHORT) || toHiveType.equals(HIVE_INT) || toHiveType.equals(HIVE_LONG))) {
             return new VarcharToIntegerNumberCoercer(fromType, toType);
         }
-        else if (fromHiveType.equals(HIVE_BYTE) && toHiveType.equals(HIVE_SHORT) || toHiveType.equals(HIVE_INT) || toHiveType.equals(HIVE_LONG)) {
+        else if (fromHiveType.equals(HIVE_BYTE) && (toHiveType.equals(HIVE_SHORT) || toHiveType.equals(HIVE_INT) || toHiveType.equals(HIVE_LONG))) {
```
yingsu00, line 84: @elharo Thanks for the fix. It does seem the original code has a bug. Will you be able to add some tests in com/facebook/presto/hive/TestCoercingFilters.java?

elharo: done

elharo (commented):
PTAL @yingsu00

presto-hive/src/test/java/com/facebook/presto/hive/TestCoercingFilters.java
```diff
@@ -70,4 +79,41 @@ public void testVarcharToInteger()
 
         assertFalse(coercingFilter.testNull());
     }
+
+    private final TypeManager typeManager = MetadataManager.createTestMetadataManager().getFunctionAndTypeManager();
```
yingsu00, line 83: In Presto, the data fields are defined at the beginning of the class. Could you please move this to line 40?

presto-hive/src/test/java/com/facebook/presto/hive/TestCoercingFilters.java
```diff
@@ -70,4 +79,41 @@ public void testVarcharToInteger()
...
+
+    @Test
+    public void testShortToInteger()
```
yingsu00, line 86: I think it's ok to put all these test cases into one test, but it's up to you.

presto-hive/src/test/java/com/facebook/presto/hive/TestCoercingFilters.java
```diff
@@ -70,4 +79,41 @@ public void testVarcharToInteger()
...
+    {
+        HiveCoercer coercer = HiveCoercer.createCoercer(typeManager, HIVE_SHORT, HIVE_INT);
```
yingsu00, line 88: Can we add some negative tests that the original code would fail? E.g. fromType is HIVE_DOUBLE and toType is HIVE_SHORT. Such case would create the VarcharToIntegerNumberCoercer before this fix, because the condition `toHiveType.equals(HIVE_SHORT)` on line 288 is true, but this is actually incorrect. In the test this shall throw an error and we need to verify the exception is thrown as expected.

elharo: done

presto-hive/src/test/java/com/facebook/presto/hive/TestCoercingFilters.java
elharo: Putting them in separate tests decouples them so they can pass or fail and be debugged independently of each other.

elharo (commented):
PTAL

presto-hive/src/test/java/com/facebook/presto/hive/TestCoercingFilters.java
```diff
@@ -17,17 +17,31 @@
...
+import static com.facebook.presto.hive.HiveType.HIVE_INT;
+import static com.facebook.presto.hive.HiveType.HIVE_LONG;
+import static com.facebook.presto.hive.HiveType.HIVE_SHORT;
+import static org.testng.Assert.assertEquals;
 import static org.testng.Assert.assertFalse;
 import static org.testng.Assert.assertTrue;
 
 public class TestCoercingFilters
 {
+    private final TypeManager typeManager = MetadataManager.createTestMetadataManager().getFunctionAndTypeManager();
```
arhimondr, line 43: nit: You can create it in `@BeforeClass` and destroy it in `@AfterClass(alwaysRun=true)` to save memory

arhimondr (approved):

PR #22916, opened 2024-06-04
Rewrote array_sum function in Java from SQL

abhinavmuk04:
## Description
Natively implement array_sum function similar to ArrayContains and other written functions

## Motivation and Context
Array_sum is currently written as a SQL Function, but it is not efficient. Implementing it as a native function will help improve performance

## Test Plan
Added large set of units tests for various data types, including various null and casting cases.

## Contributor checklist

- [ ] CI passed.

## Release Notes

== NO RELEASE NOTE ==


presto-main/src/main/java/com/facebook/presto/operator/scalar/ArraySum.java
```diff
@@ -0,0 +1,73 @@
...
+
+
+import static com.facebook.presto.common.function.OperatorType.ADD;
+import static com.facebook.presto.util.Failures.internalError;
+
+
+@SqlInvokedScalarFunction(value = "array_sum", deterministic = true, calledOnNullInput = false)
+@Description("Returns the sum of all array elements, or 0 if the array is empty. Ignores null elements.")
+@ScalarFunction("array_sum")
+public final class ArraySum
```
feilong-liu, line 27: Rename to ArraySumFunction like other function classes

presto-main/src/test/java/com/facebook/presto/type/TestArrayOperators.java
```diff
@@ -650,6 +650,30 @@ public void testArrayJoin()
         assertInvalidFunction("ARRAY_JOIN(ARRAY [cast(row(1, 2) AS row(col0 bigint, col1 bigint))], '-')", INVALID_FUNCTION_ARGUMENT);
     }
 
+    @Test
+    public void testArraySum()
```
feilong-liu, line 654: Did you test with empty array?

feilong-liu (commented):
Also change the tile and commit message to avoid using word native. In Presto it's usually used to refer to Presto CPP work. Reword it to indicate that it's implemented using Java code instead of SQL function etc.

presto-main/src/main/java/com/facebook/presto/operator/scalar/ArraySum.java
```diff
@@ -0,0 +1,73 @@
+package com.facebook.presto.operator.scalar;
```
elharo, line 1: license notice

presto-main/src/main/java/com/facebook/presto/operator/scalar/ArraySum.java
```diff
@@ -0,0 +1,73 @@
...
+                        sum = newValue;
+                    }
+                    else {
+                        sum = addFunction.invoke(sum, newValue);
+                    }
+                }
+                catch (Throwable throwable) {
+
+
+                    throw internalError(throwable);
```
elharo, line 64: Why not just let the exception bubble up?

presto-main/src/test/java/com/facebook/presto/type/TestArrayOperators.java
```diff
@@ -650,6 +650,30 @@ public void testArrayJoin()
...
+    {
+        assertFunction("ARRAY_SUM(ARRAY [1, 2, 3])",INTEGER,6);
```
elharo, line 656: Each of these should be a separate method that can pass or fail independently.

Yes, I know presto does this wrong in many existing tests. No reason to compound the problem.

elharo (dismissed):

presto-main/src/main/java/com/facebook/presto/operator/scalar/ArraySum.java
```diff
@@ -0,0 +1,73 @@
...
+
+                try {
+                    if (sum == null) {
+                        sum = newValue;
+                    }
+                    else {
+                        sum = addFunction.invoke(sum, newValue);
+                    }
+                }
+                catch (Throwable throwable) {
```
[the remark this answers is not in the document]
rschlussel: This is a pattern used in many other functions that are invoking another functions and there's good reason for it (i didn't notice this pattern before, but I just encountered it in some other changes i made recently).

It's expected that functions handle any exceptions themselves. invoke() throws Throwable, so we don't know anything about what kind of exceptions we might get. The underlying methods that we are invoking don't throw any checked exceptions, so we don't anticipate any particular exception getting thrown, so we just catch throwable here and wrap it in a runtime exception (PrestoException with a GENERIC_INTERNAL_ERROR error code).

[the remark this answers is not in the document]
elharo: Exceptions but not other Throwables. invoke throws Throwable but you don't need to catch it. It's enough to catch Exception. If you see a java.lang.Error or worse yet a Throwable that is neither an Exception or Error, the program is so borked the best you can hope for is a VM shutdown, and a full crash of the VM is likely.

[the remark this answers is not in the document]
rschlussel: java will yell at you about unhandled exceptions unless you catch Throwable.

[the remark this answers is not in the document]
rschlussel: looked inside that method, and it recasts Throwables of type error, so it's not actually swallowing them.

Issue #22910, opened 2024-06-03
Presto silently converts NaN to zero when casting it to an INTEGER but throws when casting to BIGINT.


bikramSingh91:
Since NaN cannot be represented as an integer value, it would make sense to instead return a user error similar to how it does when casting to BIGINT.

```
SELECT CAST(NAN() AS INTEGER) ;  ---- 0
SELECT CAST(NAN() AS BIGINT) ;  ---- Error:  Unable to cast NaN to bigint
```

rschlussel:
This needs to be fixed for all of the following casts
DOUBLE ->INTEGER
DOUBLE -> SMALLINT
DOUBLE  -> TINYINT.  
REAL -> BIGINT
REAL -> INTEGER
REAL -> SMALLINT
REAL -> TINYINT

PR #22917, opened 2024-06-04
Throw exception for cast of nan and infinity to int types

rschlussel:
## Description
Fix cast of nan and infinity from DOUBLE/REAL to
BIGINT/INT/SMALLINT/TINYTINT.  Previously all except double -> bigint would return zero.  Now they will all throw an INVALID_CAST_ARGUMENT exception.

Fix silent overflow for casting from REAL to BIGINT type. We now throw
an INVALID_CAST_ARGUMENT if the value is out of the BIGINT range.

Change error code from NUMERIC_VALUE_OUT_OF_RANGE to
INVALID_CAST_ARGUMENT for out of range values in casts from floating
point to integer types.  With the library we now use for the
implementaiton for these casts, we can't always tell by the exception
type that the value is out of range, so we return the more generic, but
still relevant INVALID_CAST_ARGUMENT instead.

## Motivation and Context
Fixes https://github.com/prestodb/presto/issues/22910.  
Nan and infinity can't be represented as an integer, so casting to an int type should throw an error.

This also brings Presto and velox behavior in alignment. We are choosing to modify Presto in this case because velox behavior is correct here (we would want to change Presto regardless of the native worker migration)

## Impact

CAST of nan and infinity to BIGINT, INTEGER, SMALLINT, and TINYINT types will now return an exception with the INVALID_CAST_ARGUMENT error code

Cast of out of range values for DOUBLE or FLOAT to BIGINT, INTEGER, SMALLINT, or TINYINT will now return an INVALID_CAST_ARGUMENT error code rather than NUMERIC_VALUE_OUT_OF_RANGE error code

CAST of real values outside of BIGINT range will now return a NUMERIC_VALUE_OUT_OF_RANGE error. Previously they would silently overflow.

## Test Plan
unit tests
## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Fix cast of NaN and Infinity from DOUBLE or REAL to  BIGINT, INTEGER, SMALLINT, and TINYINT. It will now return an exception with the INVALID_CAST_ARGUMENT error code. Previously it would return zero.
* Fix CAST of REAL values outside of BIGINT range to return an exception with an INVALID_CAST_ARGUMENT error code. Previously they would silently overflow.
* Change error code for cast from DOUBLE or REAL to BIGINT, INTEGER, SMALLINT or TINYINT for out of range values from ``NUMERIC_VALUE_OUT_OF_RANGE`` to ``INVALID_CAST_ARGUMENT``.

```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/type/DoubleOperators.java
+++ b/presto-main/src/main/java/com/facebook/presto/type/DoubleOperators.java
@@ -16,7 +16,6 @@ package com.facebook.presto.type;
 import com.facebook.presto.common.block.Block;
 import com.facebook.presto.common.type.AbstractLongType;
 import com.facebook.presto.common.type.StandardTypes;
-import com.facebook.presto.operator.scalar.MathFunctions;
 import com.facebook.presto.spi.PrestoException;
 import com.facebook.presto.spi.function.BlockIndex;
 import com.facebook.presto.spi.function.BlockPosition;
@@ -53,12 +52,10 @@ import static com.facebook.presto.common.function.OperatorType.XX_HASH_64;
 import static com.facebook.presto.common.type.DoubleType.DOUBLE;
 import static com.facebook.presto.spi.StandardErrorCode.DIVISION_BY_ZERO;
 import static com.facebook.presto.spi.StandardErrorCode.INVALID_CAST_ARGUMENT;
-import static com.facebook.presto.spi.StandardErrorCode.NUMERIC_VALUE_OUT_OF_RANGE;
 import static com.google.common.base.Preconditions.checkState;
 import static io.airlift.slice.Slices.utf8Slice;
 import static java.lang.Double.doubleToLongBits;
 import static java.lang.Float.floatToRawIntBits;
-import static java.lang.Math.toIntExact;
 import static java.lang.String.format;
 import static java.math.RoundingMode.FLOOR;
 import static java.math.RoundingMode.HALF_UP;
@@ -195,10 +192,10 @@ public final class DoubleOperators
     public static long castToInteger(@SqlType(StandardTypes.DOUBLE) double value)
     {
         try {
-            return toIntExact((long) MathFunctions.round(value));
+            return DoubleMath.roundToInt(value, HALF_UP);
         }
         catch (ArithmeticException e) {
-            throw new PrestoException(NUMERIC_VALUE_OUT_OF_RANGE, "Out of range for integer: " + value, e);
+            throw new PrestoException(INVALID_CAST_ARGUMENT, format("Unable to cast %s to integer", value), e);
         }
     }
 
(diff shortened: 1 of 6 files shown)


ClarenceThreepwood (dismissed):

ClarenceThreepwood:
> This also brings Presto and velox behavior in alignment. We are choosing to modify Presto in this case because velox behavior is correct here (we would want to change Presto regardless of the native worker migration)
> 
Another data point - this is also the approach that most other db engines take (e.g. Postgres)

presto-main/src/main/java/com/facebook/presto/type/DoubleOperators.java
```diff
@@ -195,10 +193,13 @@ public static boolean castToBoolean(@SqlType(StandardTypes.DOUBLE) double value)
     public static long castToInteger(@SqlType(StandardTypes.DOUBLE) double value)
     {
         try {
-            return toIntExact((long) MathFunctions.round(value));
+            return DoubleMath.roundToInt(value, HALF_UP);
         }
         catch (ArithmeticException e) {
-            throw new PrestoException(NUMERIC_VALUE_OUT_OF_RANGE, "Out of range for integer: " + value, e);
+            if (e.getMessage().equals("not in range")) {
```
elharo, line 199: How can this be done without relying on undocumented exception messages that are subject to change between JDKs and JDK versions?

If it can't, I'd drop the conditional and simply always use an INVALID_CAST_ARGUMENT

presto-main/src/main/java/com/facebook/presto/type/DoubleOperators.java
```diff
@@ -207,7 +208,13 @@ public static long castToInteger(@SqlType(StandardTypes.DOUBLE) double value)
     public static long castToSmallint(@SqlType(StandardTypes.DOUBLE) double value)
     {
         try {
-            return Shorts.checkedCast((long) MathFunctions.round(value));
+            return Shorts.checkedCast(DoubleMath.roundToInt(value, HALF_UP));
+        }
+        catch (ArithmeticException e) {
+            if (e.getMessage().equals("not in range")) {
```
elharo, line 214: ditto

presto-main/src/main/java/com/facebook/presto/type/RealOperators.java
```diff
@@ -194,18 +195,29 @@ public static Slice castToVarchar(@SqlType(StandardTypes.REAL) long value)
     @SqlType(StandardTypes.BIGINT)
     public static long castToLong(@SqlType(StandardTypes.REAL) long value)
     {
-        return (long) MathFunctions.round((double) intBitsToFloat((int) value));
+        try {
+            return DoubleMath.roundToLong(intBitsToFloat((int) value), HALF_UP);
+        }
+        catch (ArithmeticException e) {
+            if (e.getMessage().equals("not in range")) {
```
elharo, line 202: ditto

presto-main/src/main/java/com/facebook/presto/type/RealOperators.java
```diff
@@ -194,18 +195,29 @@ public static Slice castToVarchar(@SqlType(StandardTypes.REAL) long value)
...
     @SqlType(StandardTypes.INTEGER)
     public static long castToInteger(@SqlType(StandardTypes.REAL) long value)
     {
         try {
-            return toIntExact((long) MathFunctions.round((double) intBitsToFloat((int) value)));
+            return DoubleMath.roundToInt(intBitsToFloat((int) value), HALF_UP);
         }
         catch (ArithmeticException e) {
-            throw new PrestoException(NUMERIC_VALUE_OUT_OF_RANGE, "Out of range for integer: " + value, e);
+            if (e.getMessage().equals("not in range")) {
```
elharo, line 217: ditto

presto-main/src/main/java/com/facebook/presto/type/DoubleOperators.java
rschlussel: yeah, I didn't love it, but didn't see other options..  I'll switch to always using INVALID_CAST_ARGUMENT instead.

arhimondr (approved):

PR #22948, opened 2024-06-07
[native] Correct the cumulative memory usage calculation

xiaoxmeng:
```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/PrestoTask.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoTask.cpp
@@ -561,7 +561,7 @@ void PrestoTask::updateMemoryInfoLocked(
   const double sinceLastPeriodMs = currentTimeMs - lastTaskStatsUpdateMs;
 
   prestoTaskStats.cumulativeUserMemory +=
-      (averageMemoryForLastPeriod * sinceLastPeriodMs) / 1000;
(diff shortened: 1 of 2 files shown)


mbasmanova (approved):
Thank you for the fix. It would be nice to add a test.

tanjialiang (approved):

xiaoxmeng:
> Thank you for the fix. It would be nice to add a test.

We already have a test to cover that and the test has been changed accordingly.

presto-native-execution/presto_cpp/main/PrestoTask.cpp
```diff
@@ -561,7 +561,7 @@ void PrestoTask::updateMemoryInfoLocked(
   const double sinceLastPeriodMs = currentTimeMs - lastTaskStatsUpdateMs;
 
   prestoTaskStats.cumulativeUserMemory +=
-      (averageMemoryForLastPeriod * sinceLastPeriodMs) / 1000;
+      (averageMemoryForLastPeriod * sinceLastPeriodMs);
```
majetideepak, line 564: @xiaoxmeng can you clarify why this change is required? Thanks.

xiaoxmeng: IIRC, java use ms instead of us. The change is to align with java. without this change, native reports much less accumulative memory than java.

PR #22949, opened 2024-06-07
Fix serialization of aggregation node for HBO

feilong-liu:
## Description
When I was debugging why we see different hash for the same query when running Presto CPP and Presto Java, I found that the serialized Json string from Aggregation node is different between these two plans.
For Java, the "arguments" attribute is before "functionHandle".
For CPP, the "functionHandle" is before "arguments".

The order of attributes should respect the order defined in json constructor, however further debug shows that, this is due to the two get function which is annotated with JsonProperty but not present in the constructor.
https://github.com/prestodb/presto/blob/b67366892b06a3c9ba76730876178c85c6e9479a/presto-spi/src/main/java/com/facebook/presto/spi/plan/AggregationNode.java#L475,L485
Somehow this shows in different order in serialization of cpp and java plans.

I do not think these two get functions need to be annotated as Json property, so I removed them.
I also set the objectMapper used by HBO to use SORT_PROPERTIES_ALPHABETICALLY property, so that even we have other similar cases, we can still output consistent serialized string

## Motivation and Context
Described above

## Impact
It will bridge the gap between CPP and Java for HBO. And solve potential mismatch of HBO hash caused by this issue.

## Test Plan
Tested locally end to end in Java HiveQueryRunner, make sure 1) the two additional fields not appearing in serialized json string now 2) the new json string has fields sorted

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* * Fix serialization of aggregation node in HBO plan hash to output consistent hash :pr:`22949 `
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/cost/HistoryBasedPlanStatisticsManager.java
+++ b/presto-main/src/main/java/com/facebook/presto/cost/HistoryBasedPlanStatisticsManager.java
@@ -23,6 +23,7 @@ import com.facebook.presto.spi.statistics.HistoryBasedPlanStatisticsProvider;
 import com.facebook.presto.sql.analyzer.FeaturesConfig;
 import com.facebook.presto.sql.planner.CachingPlanCanonicalInfoProvider;
 import com.facebook.presto.sql.planner.PlanCanonicalInfoProvider;
+import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.SerializationFeature;
 import com.google.inject.Inject;
@@ -51,7 +52,7 @@ public class HistoryBasedPlanStatisticsManager
         requireNonNull(objectMapper, "objectMapper is null");
         this.sessionPropertyManager = requireNonNull(sessionPropertyManager, "sessionPropertyManager is null");
         this.historyBasedStatisticsCacheManager = new HistoryBasedStatisticsCacheManager();
-        ObjectMapper newObjectMapper = objectMapper.copy().configure(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS, true);
+        ObjectMapper newObjectMapper = objectMapper.copy().configure(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS, true).configure(MapperFeature.SORT_PROPERTIES_ALPHABETICALLY, true);
         this.planCanonicalInfoProvider = new CachingPlanCanonicalInfoProvider(historyBasedStatisticsCacheManager, newObjectMapper, metadata);
         this.config = requireNonNull(config, "config is null");
         this.isNativeExecution = featuresConfig.isNativeExecutionEnabled();


presto-main/src/main/java/com/facebook/presto/cost/HistoryBasedPlanStatisticsManager.java
```diff
@@ -51,7 +52,7 @@ public HistoryBasedPlanStatisticsManager(ObjectMapper objectMapper, SessionPrope
         requireNonNull(objectMapper, "objectMapper is null");
         this.sessionPropertyManager = requireNonNull(sessionPropertyManager, "sessionPropertyManager is null");
         this.historyBasedStatisticsCacheManager = new HistoryBasedStatisticsCacheManager();
-        ObjectMapper newObjectMapper = objectMapper.copy().configure(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS, true);
+        ObjectMapper newObjectMapper = objectMapper.copy().configure(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS, true).configure(MapperFeature.SORT_PROPERTIES_ALPHABETICALLY, true);
```
rschlussel, line 55: is this still needed once you remove the annotations that shouldn't have been there?

feilong-liu:
This specific problem with aggregation will not need this additional change. It’s for potential similar problem which is not discovered or may be introduced later.

Sent from my iPhone

On Jun 7, 2024, at 9:27 PM, Rebecca Schlussel ***@***.***> wrote:

﻿
@rschlussel commented on this pull request. In presto-main/src/main/java/com/facebook/presto/cost/HistoryBasedPlanStatisticsManager. java: > @@ -51,7 +52,7 @@ public HistoryBasedPlanStatisticsManager(ObjectMapper objectMapper, SessionPrope
ZjQcmQRYFpfptBannerStart
This Message Is From an External Sender

ZjQcmQRYFpfptBannerEnd

@rschlussel commented on this pull request.

In presto-main/src/main/java/com/facebook/presto/cost/HistoryBasedPlanStatisticsManager.java<https://github.com/prestodb/presto/pull/22949#discussion_r1631204105>:

> @@ -51,7 +52,7 @@ public HistoryBasedPlanStatisticsManager(ObjectMapper objectMapper, SessionPrope
         requireNonNull(objectMapper, "objectMapper is null");
         this.sessionPropertyManager = requireNonNull(sessionPropertyManager, "sessionPropertyManager is null");
         this.historyBasedStatisticsCacheManager = new HistoryBasedStatisticsCacheManager();
-        ObjectMapper newObjectMapper = objectMapper.copy().configure(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS, true);
+        ObjectMapper newObjectMapper = objectMapper.copy().configure(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS, true).configure(MapperFeature.SORT_PROPERTIES_ALPHABETICALLY, true);


is this still needed once you remove the annotations that shouldn't have been there?

presto-main/src/main/java/com/facebook/presto/cost/HistoryBasedPlanStatisticsManager.java
ClarenceThreepwood: If this is to future-proof serialization between java and native plans wdyt about adding a test (or test framework) that compares the serialized plans?

rschlussel (dismissed):
This change looks good, but I like @ClarenceThreepwood's idea of adding a Java/cpp serialization test to ensure the plans are the same. Can be done either in this PR or as a follow up.

tdcmeehan:
[JSON objects are by definition unordered](https://www.json.org/json-en.html).  Attempting to infer ordering in the JSON object is inherently fragile.  Isn't the problem in the hashing logic, rather than the serialization?

tdcmeehan:
I took a brief look at the code: https://github.com/prestodb/presto/blob/2c226fc1e0518b94f1faac01b3ee92ad0fd2c64c/presto-main/src/main/java/com/facebook/presto/sql/planner/CachingPlanCanonicalInfoProvider.java#L103-L109

What I'm not getting is, instead of converting the plan to a JSON object, and then hashing the serialized plan, why not just implement a visitor that recursively hashes the plan nodes (and then tack on the `PlanCanonicalizationStrategy` at the end of that)?  That way we're resilient to both the ordering of the fields, and to any extraneous fields which may be added.

feilong-liu:
> The order of attributes should respect the order defined in json constructor, however further debug shows that, this is due to the two get function which is annotated with JsonProperty but not present in the constructor.
https://github.com/prestodb/presto/blob/b67366892b06a3c9ba76730876178c85c6e9479a/presto-spi/src/main/java/com/facebook/presto/spi/plan/AggregationNode.java#L475,L485
Somehow this shows in different order in serialization of cpp and java plans.

> I do not think these two get functions need to be annotated as Json property, so I removed them.


Removing these annotations lead to native test failures https://app.circleci.com/pipelines/github/prestodb/presto/17132/workflows/dc57b848-9162-4ead-b157-402e6a33ce64/jobs/67557
I chose to not remove these annotations for now to unblock this fix for HBO, as the setting of "SORT_PROPERTIES_ALPHABETICALLY" should be enough to fix the problem. In the meantime, I am debugging the failure in native tests.

feilong-liu:
> This change looks good, but I like @ClarenceThreepwood's idea of adding a Java/cpp serialization test to ensure the plans are the same. Can be done either in this PR or as a follow up.

Yeah, will do it as a follow up.

feilong-liu:
> I took a brief look at the code:
> 
> https://github.com/prestodb/presto/blob/2c226fc1e0518b94f1faac01b3ee92ad0fd2c64c/presto-main/src/main/java/com/facebook/presto/sql/planner/CachingPlanCanonicalInfoProvider.java#L103-L109
> 
> What I'm not getting is, instead of converting the plan to a JSON object, and then hashing the serialized plan, why not just implement a visitor that recursively hashes the plan nodes (and then tack on the `PlanCanonicalizationStrategy` at the end of that)? That way we're resilient to both the ordering of the fields, and to any extraneous fields which may be added.

This means that we need to have a visitor to visit all elements of a class recursively, and output a unique hash in the end which is a big change. And it also needs to be updated whenever new fields are added to classes which could be referenced in a query plan.  Current approach already works well, so I prefer to stick to existing approaches here.

feilong-liu:
> [JSON objects are by definition unordered](https://www.json.org/json-en.html). Attempting to infer ordering in the JSON object is inherently fragile. Isn't the problem in the hashing logic, rather than the serialization?

The plan is first serialized to a json string, and this string is hashed at the end. For the problem I fixed here, it's the json string produced is different.

rschlussel (approved):

tdcmeehan:
> > [JSON objects are by definition unordered](https://www.json.org/json-en.html). Attempting to infer ordering in the JSON object is inherently fragile. Isn't the problem in the hashing logic, rather than the serialization?
> 
> The plan is first serialized to a json string, and this string is hashed at the end. For the problem I fixed here, it's the json string produced is different.

This assertion:

> The order of attributes should respect the order defined in json constructor

Is simply not true--it's at least not true according to the JSON spec.  Please consider a proper fix that corrects the hashing logic.

> This means that we need to have a visitor to visit all elements of a class recursively, and output a unique hash in the end which is a big change. And it also needs to be updated whenever new fields are added to classes which could be referenced in a query plan. Current approach already works well, so I prefer to stick to existing approaches here.

It is common to update equals and hashCode on Java objects where needed (as they are here), and it would be trivial to write a round trip unit test that proves serialized and deserialized output is identical to catch problems where they are incorrect.  It would seemingly be added anyway be added to address @ClarenceThreepwood's feedback.

feilong-liu:
> The order of attributes should respect the order defined in json constructor
> Is simply not true--it's at least not true according to the JSON spec.

After adding MapperFeature.SORT_PROPERTIES_ALPHABETICALLY in this PR, we will not rely on this assumption now.

tdcmeehan:
> > This change looks good, but I like @ClarenceThreepwood's idea of adding a Java/cpp serialization test to ensure the plans are the same. Can be done either in this PR or as a follow up.
> 
> Yeah, will do it as a follow up.

Was this test added?

PR #22957, opened 2024-06-07
Change confidenceLevel from boolean to Enum

abhinavmuk04:
## Description
Change confidenceLevel from boolean to Enum

## Motivation and Context
This will help us with future changes where we want to consider choosing the 'FACT' CBO over regular HBO

## Test Plan
Reran existing tests where previosuly 'isConfident' was used with new 'confidenceLevel' functions

## Contributor checklist

- [ ] CI passed.

## Release Notes

== NO RELEASE NOTE ==


jaystarshot:
Why not keep it a fraction and add a method to get high low etc? 
This can be used for eg-
Table scan - confidence 1 ,
 filter on top decreases confidence by 0.05
 so filter-tablescan can have confidence 0.95

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/PushPartialAggregationThroughExchange.java
```diff
@@ -337,17 +339,18 @@ private boolean partialAggregationNotUseful(AggregationNode aggregationNode, Exc
...
         if (!isUnknown(partialAggregationStatsEstimate) && usePartialAggregationHistory(context.getSession())) {
-            isConfident = aggregationStats.isConfident();
+            confidenceLevel = aggregationStats.confidenceLevel();
             // use rows instead of bytes when use_partial_aggregation_history flag is on
             inputSize = partialAggregationStatsEstimate.getInputRowCount();
             outputSize = partialAggregationStatsEstimate.getOutputRowCount();
         }
         double byteReductionThreshold = getPartialAggregationByteReductionThreshold(context.getSession());
 
+        boolean isConfident = confidenceLevel == LOW ? false : true;
```
rschlussel, line 353: you can inline this variable as confidenceLevel != LOW

presto-main/src/main/java/com/facebook/presto/cost/TableScanStatsRule.java
```diff
@@ -69,7 +70,7 @@ protected Optional<PlanNodeStatsEstimate> doCalculate(TableScanNode node, StatsP
         return Optional.of(PlanNodeStatsEstimate.builder()
                 .setOutputRowCount(tableStatistics.getRowCount().getValue())
                 .setTotalSize(tableStatistics.getTotalSize().getValue())
-                .setConfident(true)
+                .setConfidence(FACT)
```
rschlussel, line 73: add a TODO here to handle the confidence level properly when filters are pushed into the tablescan.

presto-main/src/test/java/com/facebook/presto/sql/planner/iterative/rule/TestPushPartialAggregationThroughExchange.java
```diff
@@ -197,7 +199,7 @@ public void testPartialAggregationEnabledWhenNotConfident()
                 .overrideStats("values", PlanNodeStatsEstimate.builder()
                         .setOutputRowCount(1000)
                         .addVariableStatistics(variable("b", DOUBLE), new VariableStatsEstimate(0, 100, 0, 8, 800))
-                        .setConfident(false)
+                        .setConfidence(FACT)
```
rschlussel, line 202: this should be LOW (and if the test passes without that then either the test or the code is broken).

rschlussel:
> Why not keep it a fraction and add a method to get high low etc? This can be used for eg- Table scan - confidence 1 , filter on top decreases confidence by 0.05 so filter-tablescan can have confidence 0.95

So that approach is nice because it allows for appreciating that compounding errors makes things worse, but my concern is that it's way more precision than we actually know and more than we really have a user for. For example, even though a filter on top of a join on top of a filter has a problem of compounding error, a filter on top of a tablescan is also basically a guess if there are no histograms, or even with histograms if your filter is a more complex function. Similarly if you have filters on multiple columns and you don't know the extent to which they are correlated, it's all a guess. I'm not sure it's worth teasing out of all these things the relative confidence levels because the benefit isn't clear to me.

And for similar reasons it would also be hard to keep all the relative confidences consistent (like is x actually less confident than y or did the particular paths the confidence levels came through derive their computation in different ways/were implemented at different times,etc.  but really we'd think y is less confident than x or their about the same)

I think until there's a clear use case for a continuous range of confidence levels, a few discrete options is simpler and does what we need.

presto-main/src/test/java/com/facebook/presto/sql/planner/iterative/rule/TestPushPartialAggregationThroughExchange.java
kaikalur: why? 1000 values row count is a fact for VALUES clause?

rschlussel: because this test is testing that partialaggregation is enabled when stats are not confident. So if the test passes when the stats *are* confident, then either the tests isn't testing what it's supposed to or something is wrong with the code.

rschlussel (commented):
there are many other stats rules that need to be updated. check for all classes that extend simple stats rule or all usages of PlanNodeStatsEstimate.Builder

presto-main/src/test/java/com/facebook/presto/cost/TestExchangeStatsRule.java
```diff
@@ -153,13 +154,13 @@ public void testExchangeConfidence()
                         .addSource(pb.values())))
                 .withSourceStats(0, PlanNodeStatsEstimate.builder()
                         .setOutputRowCount(100)
-                        .setConfident(true)
+                        .setConfidence(FACT)
                         .build())
                 .withSourceStats(1, PlanNodeStatsEstimate.builder()
                         .setOutputRowCount(100)
-                        .setConfident(false)
+                        .setConfidence(FACT)
```
rschlussel, line 161: this should be LOW

presto-main/src/test/java/com/facebook/presto/cost/TestExchangeStatsRule.java
```diff
@@ -153,13 +154,13 @@ public void testExchangeConfidence()
...
                         .build())
                 .check(check -> check
-                        .confident(false));
+                        .confident(FACT));
```
rschlussel, line 164: this should be low.  You should make sure to check what the tests are doing as you update them so you are not changing the meaning. 
This test was checking that the confidence would be set appropritely for an exchange node based on its input.  It had confidence set to true for some of its sources and false for one of them, and it expected that the overall confidence for the exchange node would be set to false.  We want to maintain both that behavior and the test for it with the enum change, so this test should several inputs with fact or high confidence and one with LOW confidence and the node confidence should be considered "LOW"

rschlussel (approved):
minor comment, but looks good.  Thanks!

rschlussel (approved):

jaystarshot (approved):

PR #22967, opened 2024-06-10
[native] Retrieve session properties supported by presto native

pramodsatya:
## Description
The mapping of Presto session property to the equivalent velox config currently exists in [`QueryContextManager.cpp : kPrestoToVeloxMapping`](https://github.com/prestodb/presto/blob/2855762422bfb3e84ec86dae7a101a978a5db3d3/presto-native-execution/presto_cpp/main/QueryContextManager.cpp#L35). This is used by the task manager to update the velox query config, based on session property values in the task create/update request. The Presto to velox config mapping is moved from `kPrestoToVeloxMapping` to a new class `SessionProperties`. An object of `SessionProperties` is now used by `QueryContextManager` to retrieve this mapping and update the [velox configs](https://github.com/prestodb/presto/blob/2855762422bfb3e84ec86dae7a101a978a5db3d3/presto-native-execution/presto_cpp/main/QueryContextManager.cpp#L92). 
[SPI v2 ](https://github.com/prestodb/rfcs/blob/main/RFC-0003-native-spi.md) motivates that `SHOW SESSIONS` in presto native should only return the list of session properties compatible with velox. In order to support this, the presto native sidecar process would return the list of session properties supported in presto native via proxygen endpoint `/v1/properties/session`.  Helper functions to serialize `SessionProperties` (and `SessionProperty`) is added to serialize the Presto native session properties as JSON, which is returned along this endpoint. The Presto coordinator can query this endpoint to retrieve session properties supported by the presto native worker.

## Motivation and Context
[SPI v2 ](https://github.com/prestodb/rfcs/blob/main/RFC-0003-native-spi.md) requires `SHOW SESSION` to return only the session properties that are compatible with its underlying evaluation engine. When a Presto native user issues `SHOW SESSION`, it should only include session properties compatible with velox that will actually be used. 

## Impact
Introduces proxygen endpoint `/v1/properties/session` for the presto native sidecar. This returns a map of session properties supported by the presto native worker.

## Release Notes
```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/PrestoServer.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoServer.cpp
@@ -458,6 +458,19 @@ void PrestoServer::run() {
   taskManager_ = std::make_unique<TaskManager>(
       driverExecutor_.get(), httpSrvCpuExecutor_.get(), spillerExecutor_.get());
 
+  if (systemConfig->prestoNativeSidecar()) {
+    httpServer_->registerGet(
+        "/v1/properties/session",
+        [this](
+            proxygen::HTTPMessage* /*message*/,
+            const std::vector<std::unique_ptr<folly::IOBuf>>& /*body*/,
+            proxygen::ResponseHandler* downstream) {
+          auto sessionProperties =
+              taskManager_->getQueryContextManager()->getSessionProperties();
+          http::sendOkResponse(downstream, sessionProperties.serialize());
+        });
+  }
+
   std::string taskUri;
   if (httpsPort.has_value()) {
     taskUri = fmt::format(kTaskUriFormat, kHttps, address_, httpsPort.value());
--- a/presto-native-execution/presto_cpp/main/QueryContextManager.cpp
+++ b/presto-native-execution/presto_cpp/main/QueryContextManager.cpp
@@ -28,36 +28,6 @@ using facebook::presto::protocol::TaskId;
 
 namespace facebook::presto {
 namespace {
-// Utility function to translate a config name in Presto to its equivalent in
-// Velox. Returns 'name' as is if there is no mapping.
-std::string toVeloxConfig(const std::string& name) {
-  using velox::core::QueryConfig;
-  static const folly::F14FastMap<std::string, std::string>
-      kPrestoToVeloxMapping = {
-          {"native_simplified_expression_evaluation_enabled",
-           QueryConfig::kExprEvalSimplified},
-          {"native_max_spill_level", QueryConfig::kMaxSpillLevel},
-          {"native_max_spill_file_size", QueryConfig::kMaxSpillFileSize},
-          {"native_spill_compression_codec",
-           QueryConfig::kSpillCompressionKind},
-          {"native_spill_write_buffer_size",
-           QueryConfig::kSpillWriteBufferSize},
-          {"native_spill_file_create_config",
-           QueryConfig::kSpillFileCreateConfig},
-          {"native_join_spill_enabled", QueryConfig::kJoinSpillEnabled},
-          {"native_window_spill_enabled", QueryConfig::kWindowSpillEnabled},
-          {"native_writer_spill_enabled", QueryConfig::kWriterSpillEnabled},
-          {"native_row_number_spill_enabled",
-           QueryConfig::kRowNumberSpillEnabled},
-          {"native_spiller_num_partition_bits",
-           QueryConfig::kSpillNumPartitionBits},
-          {"native_topn_row_number_spill_enabled",
-           QueryConfig::kTopNRowNumberSpillEnabled},
-          {"native_debug_validate_output_from_operators",
-           QueryConfig::kValidateOutputFromOperators}};
-  auto it = kPrestoToVeloxMapping.find(name);
-  return it == kPrestoToVeloxMapping.end() ? name : it->second;
-}
 
 // Update passed in query session configs with system configs. For any pairing
 // system/session configs if session config is present, it overrides system
@@ -87,26 +57,6 @@ void updateFromSystemConfigs(
       }
     }
   }
-}
-
-std::unordered_map<std::string, std::string> toVeloxConfigs(
-    const protocol::SessionRepresentation& session) {
-  // Use base velox query config as the starting point and add Presto session
-  // properties on top of it.
-  auto configs = BaseVeloxQueryConfig::instance()->values();
-  for (const auto& it : session.systemProperties) {
-    configs[toVeloxConfig(it.first)] = it.second;
-  }
-
-  // If there's a timeZoneKey, convert to timezone name and add to the
-  // configs. Throws if timeZoneKey can't be resolved.
-  if (session.timeZoneKey != 0) {
-    configs.emplace(
-        velox::core::QueryConfig::kSessionTimezone,
-        velox::tz::getTimeZoneName(session.timeZoneKey));
-  }
-  updateFromSystemConfigs(configs);
-  return configs;
 }
 
 std::unordered_map<std::string, std::unordered_map<std::string, std::string>>
@@ -167,7 +117,9 @@ void updateVeloxConnectorConfigs(
 QueryContextManager::QueryContextManager(
     folly::Executor* driverExecutor,
     folly::Executor* spillerExecutor)
-    : driverExecutor_(driverExecutor), spillerExecutor_(spillerExecutor) {}
+    : driverExecutor_(driverExecutor),
+      spillerExecutor_(spillerExecutor),
+      sessionProperties_(SessionProperties()) {}
 
 std::shared_ptr<velox::core::QueryCtx>
 QueryContextManager::findOrCreateQueryCtx(
@@ -242,6 +194,28 @@ void QueryContextManager::testingClearCache() {
 void QueryContextCache::testingClear() {
   queryCtxs_.clear();
   queryIds_.clear();
+}
+
+std::unordered_map<std::string, std::string>
+QueryContextManager::toVeloxConfigs(
+    const protocol::SessionRepresentation& session) {
+  // Use base velox query config as the starting point and add Presto session
+  // properties on top of it.
+  auto configs = BaseVeloxQueryConfig::instance()->values();
+  for (const auto& it : session.systemProperties) {
+    configs[sessionProperties_.toVeloxConfig(it.first)] = it.second;
+    sessionProperties_.updateVeloxConfig(it.first, it.second);
+  }
+
+  // If there's a timeZoneKey, convert to timezone name and add to the
+  // configs. Throws if timeZoneKey can't be resolved.
+  if (session.timeZoneKey != 0) {
+    configs.emplace(
+        velox::core::QueryConfig::kSessionTimezone,
+        velox::tz::getTimeZoneName(session.timeZoneKey));
+  }
+  updateFromSystemConfigs(configs);
+  return configs;
 }
 
 } // namespace facebook::presto
--- a/presto-native-execution/presto_cpp/main/QueryContextManager.h
+++ b/presto-native-execution/presto_cpp/main/QueryContextManager.h
@@ -19,6 +19,7 @@
 #include <memory>
 #include <unordered_map>
 
+#include "presto_cpp/main/SessionProperties.h"
 #include "presto_cpp/presto_protocol/presto_protocol.h"
 #include "velox/core/QueryCtx.h"
 
@@ -116,6 +117,10 @@ class QueryContextManager {
   /// Test method to clear the query context cache.
   void testingClearCache();
 
+  const SessionProperties& getSessionProperties() const {
+    return sessionProperties_;
+  }
+
  private:
   std::shared_ptr<velox::core::QueryCtx> findOrCreateQueryCtx(
       const protocol::TaskId& taskId,
@@ -125,10 +130,14 @@ class QueryContextManager {
           std::unordered_map<std::string, std::string>>&&
           connectorConfigStrings);
 
+  std::unordered_map<std::string, std::string> toVeloxConfigs(
+      const protocol::SessionRepresentation& session);
+
   folly::Executor* const driverExecutor_{nullptr};
   folly::Executor* const spillerExecutor_{nullptr};
 
   folly::Synchronized<QueryContextCache> queryContextCache_;
+  SessionProperties sessionProperties_;
 };
 
 } // namespace facebook::presto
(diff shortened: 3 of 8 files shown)


presto-native-execution/presto_cpp/main/QueryContextManager.cpp
```diff
@@ -27,37 +28,6 @@ using facebook::presto::protocol::TaskId;
 
 namespace facebook::presto {
 namespace {
-// Utility function to translate a config name in Presto to its equivalent in
```
aditi-pandit, line 30: @pramodsatya : You might want to separate just this change to move these properties to their own SystemSessionProperties file in a separate PR.

presto-native-execution/presto_cpp/main/SessionProperty.h
```diff
@@ -0,0 +1,133 @@
...
+  kBool,
+  kLong,
+  kString,
+  // Add more types as needed.
+  kUnknown
+};
+
+/// This is the interface of the session property.
+/// Note: This interface should align with java coordinator.
+class SessionProperty {
```
aditi-pandit, line 34: @pramodsatya : Didn't follow why we have separate SessionProperty and SessionPropertyData ? Are we planning to use SessionProperty separately from SessionPropertyData at all ?

pramodsatya: Thanks for the feedback @aditi-pandit, yes I agree both classes are not needed. I am working on refactoring this PR and the draft PR contains only the cherrypicked commits from fusionnext. I will finish refactoring and open it for review soon.

presto-native-execution/presto_cpp/main/QueryContextManager.cpp
pramodsatya: Thanks for the suggestion @aditi-pandit, the `SessionProperty` class already contains the presto session property to velox property mapping so it could be reused here. 
Could you please confirm if this should be moved to a separate PR? In that case, since `SessionProperty::veloxConfigName_` would be unused, could I remove this class variable?

presto-native-execution/presto_cpp/main/SessionProperties.cpp
```diff
@@ -0,0 +1,261 @@
...
+  addSessionProperty(
+      kDriverCpuTimeSliceLimitMs,
+      "Native Execution only. The cpu time slice limit in ms that a driver thread. "
+      "If not zero, can continuously run without yielding. If it is zero, then "
+      "there is no limit.",
+      INTEGER(),
+      false,
+      QueryConfig::kDriverCpuTimeSliceLimitMs,
+      // Overrides velox default value. Set it to 1 second to be aligned with
+      // Presto Java.
```
abevk2023, line 206: Two hard coded properties were added by updateVeloxConfigs() are already included and these properties will be visible in show session. Please remove updateVeloxConfigs() from QueryContextManager::findOrCreateQueryCtx().

presto-native-execution/presto_cpp/main/QueryContextManager.cpp
```diff
@@ -93,8 +64,9 @@ std::unordered_map<std::string, std::string> toVeloxConfigs(
   // Use base velox query config as the starting point and add Presto session
   // properties on top of it.
   auto configs = BaseVeloxQueryConfig::instance()->values();
+  SessionProperties sessionProperties;
   for (const auto& it : session.systemProperties) {
-    configs[toVeloxConfig(it.first)] = it.second;
+    configs[sessionProperties.toVeloxConfig(it.first)] = it.second;
```
abevk2023, line 69: Only configs defined by `BaseVeloxQueryConfig` are used.  Please keep `SessionProperties` as the source of truth, to ensure unified default values in the Prestissimo are used by Velox.

abevk2023 (commented):
Thank you for raising PR on session properties. I have a few comments.

presto-native-execution/presto_cpp/main/tests/SessionPropertiesTest.cpp
```diff
@@ -0,0 +1,92 @@
...
+          "spill_compression_codec",
+          "none")};
+
+  for (auto i = 0; i < names.size(); i++) {
+    auto property = sessionProperties.getSessionProperty(names[i]);
+    EXPECT_EQ(*expectedSessionProps[i], *property);
+  }
+}
+
+TEST_F(SessionPropertiesTest, sessionPropertyReporter) {
```
aditi-pandit, line 62: One of the tests you could write is iterate over the json from getSessionPropertiesMetadata(), and for each session property in it invoke getSessionPropertyMetadata(const std::string& name); and compare the result with the values in the original json from  getSessionPropertiesMetadata()

aditi-pandit (commented):
Thanks @pramodsatya. Have a first round of comments,

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -376,6 +376,18 @@ void PrestoServer::run() {
             .sendWithEOM();
       });
 
+  if (systemConfig->prestoNativeSidecar()) {
+    sessionPropertyReporter_ = std::make_unique<SessionPropertyReporter>();
+    httpServer_->registerGet(
+        "/v1/sessionProperties",
```
czentgr, line 382: This should likely be a different name. CamelCase isn't best practice for restful APIs. Maybe `v1/session-properties`? Or if we had other GET for properties - then we could do `v1/properties/session`.

pramodsatya (commented):
Thanks for the feedback @aditi-pandit @czentgr, addressed the comments. Could you please take another look?

presto-native-execution/presto_cpp/main/QueryContextManager.cpp
aditi-pandit: @pramodsatya : Post the refactoring there isn't much left for this PR. We can just keep in this single PR. Thanks for looking into this.

presto-native-execution/presto_cpp/main/QueryContextManager.cpp
```diff
@@ -89,12 +60,13 @@ void updateFromSystemConfigs(
 }
 
 std::unordered_map<std::string, std::string> toVeloxConfigs(
-    const protocol::SessionRepresentation& session) {
+    const protocol::SessionRepresentation& session,
+    const std::shared_ptr<SessionProperties> sessionProperties) {
```
aditi-pandit, line 64: Can this be a regular reference ? A shared_ptr needn't be used here.

presto-native-execution/presto_cpp/main/QueryContextManager.h
```diff
@@ -127,6 +128,7 @@ class QueryContextManager {
 
   folly::Executor* const driverExecutor_{nullptr};
   folly::Executor* const spillerExecutor_{nullptr};
+  const std::shared_ptr<SessionProperties> sessionProperties_;
```
aditi-pandit, line 131: This can simply be a SessionProperties object. A shared_ptr is not required. The lifetime of SessionProperties is the same as that of a QueryContextManager.

presto-native-execution/presto_cpp/main/SessionProperties.h
```diff
@@ -0,0 +1,192 @@
...
+  SessionPropertyReporter() {
+    sessionProperties_ = std::make_shared<SessionProperties>();
+  }
+
+  json getSessionPropertyMetadata(const std::string& name);
+
+  json getSessionPropertiesMetadata();
+
+ protected:
+  std::shared_ptr<SessionProperties> sessionProperties_;
```
aditi-pandit, line 189: Does this need to be a shared_ptr ? If this object is only used within the class, then it can be just an object, it doesn't need a shared_ptr.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -376,6 +376,18 @@ void PrestoServer::run() {
             .sendWithEOM();
       });
 
+  if (systemConfig->prestoNativeSidecar()) {
+    sessionPropertyReporter_ = std::make_unique<SessionPropertyReporter>();
```
aditi-pandit, line 380: This memory model is confusing. Seems you have one copy of SessionProperties in SessionPropertyReporter and another in QueryContextManager. There should be only one SessionProperties in PrestoServer. If the correct one is in QueryContextManager because many property translations happen there, then the SessionPropertyReporter should be passed the SessionProperties from the QueryContextManager.

Does this sound reasonable ?

presto-native-execution/presto_cpp/main/SessionProperties.cpp
```diff
@@ -0,0 +1,243 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "velox/core/QueryConfig.h"
+
+#include "presto_cpp/main/SessionProperties.h"
```
czentgr, line 16: This should be before the velox include. Also this should be first include as it is the header for this implementation file.

presto-native-execution/presto_cpp/main/SessionProperties.h
```diff
@@ -0,0 +1,191 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include "velox/type/Type.h"
+
+#include "presto_cpp/external/json/nlohmann/json.hpp"
```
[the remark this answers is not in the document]
czentgr: This shows resolved but the velox include is still before the presto_cpp one.

czentgr (commented):
Thanks! Just a few comments.

presto-native-execution/presto_cpp/main/SessionProperties.h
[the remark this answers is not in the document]
pramodsatya: Updated, apologies for missing this.

presto-native-execution/presto_cpp/main/QueryContextManager.h
```diff
@@ -116,6 +117,10 @@ class QueryContextManager {
   /// Test method to clear the query context cache.
   void testingClearCache();
 
+  SessionProperties getSessionProperties() {
```
aditi-pandit, line 120: Returned object should be const&. And this should be a const function as well.

presto-native-execution/presto_cpp/main/SessionProperties.cpp
```diff
@@ -0,0 +1,242 @@
...
+      isHidden,
+      veloxConfigName,
+      veloxDefault);
+}
+
+// List of native session properties is kept as the source of truth here.
+SessionProperties::SessionProperties() {
+  // Use empty instance to get default property values.
+  velox::core::QueryConfig c{{}};
+  using velox::core::QueryConfig;
```
aditi-pandit, line 58: Nit : Move this line above line 57, so that c{()} needn't use the fully qualified name.

presto-native-execution/presto_cpp/main/SessionProperties.cpp
```diff
@@ -0,0 +1,242 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "presto_cpp/main/SessionProperties.h"
+#include "velox/core/QueryConfig.h"
+
+using namespace facebook::presto;
```
aditi-pandit, line 17: Its a bit strange to have using namespace facebook::presto and then making it the namespace of this class as well. Is this using statement needed ?

aditi-pandit (commented):
@pramodsatya : Are you also planning to add an e2e test on theJava side that starts a side-car and validates all the session properties returned ?

presto-native-execution/presto_cpp/main/SessionProperties.cpp
pramodsatya: Removed it, thanks for pointing it out.

pramodsatya (commented):
Thanks @aditi-pandit, Pratik had informed me the e2e test will be added in the session property plugin PR subsequently. Could you please confirm this @pdabre12?

aditi-pandit:
> Thanks @aditi-pandit, Pratik had informed me the e2e test will be added in the session property plugin PR subsequently. Could you please confirm this @pdabre12?

Thanks @pramodsatya. That's alright. Please just add a link to his PR here later on.

aditi-pandit:
@pramodsatya : I'm satisfied with this code. 

@czentgr : Would be great to get a thumbs-up from you as well and I'll approve.

presto-native-execution/presto_cpp/main/tests/SessionPropertiesTest.cpp
```diff
@@ -0,0 +1,71 @@
...
+           "legacy_timestamp",
+           "Native Execution only. Use legacy TIME & TIMESTAMP semantics. Warning: "
+           "this will be removed",
+           "BOOLEAN",
+           false,
+           "adjust_timestamp_to_session_timezone",
+           "true")},
+      {"driver_cpu_time_slice_limit_ms",
+       std::make_shared<SessionProperty>(
+           "driver_cpu_time_slice_limit_ms",
```
spershin, line 35: Not sure what this test is supposed to test.
We copy initialization of few properties from the SessionProperties and then compare with the returned from SessionProperties.
When something changes in SessionProperties (say typo fixed or default value or deleted) we need to update this test.

Is it possible to use constants from presto-native-execution/presto_cpp/main/SessionProperties.h for session prop names rather than plain text in the test?

presto-native-execution/presto_cpp/main/tests/SessionPropertiesTest.cpp
```diff
@@ -0,0 +1,71 @@
...
+           "Native Execution only. The cpu time slice limit in ms that a driver thread. "
```
spershin, line 36: Description as well.

presto-native-execution/presto_cpp/main/tests/SessionPropertiesTest.cpp
```diff
@@ -0,0 +1,71 @@
...
+           "If not zero, can continuously run without yielding. If it is zero, then "
+           "there is no limit.",
+           "INTEGER",
+           false,
```
spershin, line 40: And types with default value.

pramodsatya: Removed validation of types, since there is no helper function in `SessionProperty` to retrieve property type.

czentgr (approved):
Thanks! Looks good.

aditi-pandit (approved):
Thanks @pramodsatya. Looks good.

PR #22980, opened 2024-06-11
Fix hash calculation for Timestamp in HiveBucketing to be Hive Compatible

kewang1024:
[Presto Java](https://github.com/prestodb/presto/blob/3d06a01ab54f287331975bd6a2f2cdf34aee9529/presto-hive/src/main/java/com/facebook/presto/hive/HiveBucketing.java#L175-L179) calculates hash for Timestamp type using formula:
hash(seconds << 30 + **milliseconds**)
[Hive](https://github.com/apache/hive/blob/master/serde/src/java/org/apache/hadoop/hive/serde2/io/TimestampWritable.java#L393-L397): 
hash(seconds << 30 + nanoseconds)
[Spark](https://github.com/apache/spark/blob/5a2f374a208f9580ea8d0183d75df6cd2bee8e1f/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/hash.scala#L966-L974): 
hash(seconds << 30 + nanoseconds)
[Velox](https://github.com/facebookincubator/velox/blob/102526b4444df4723448078141cf52034b339355/velox/connectors/hive/HivePartitionFunction.cpp#L44-L46): 
hash(seconds << 30 + nanoseconds)

To keep a consistent behavior with other engines, fix the hash calculation formula for Timestamp.

```
== RELEASE NOTES ==

Hive Connector Changes
* Fix hash calculation for Timestamp column to be hive compatible when writing to a table bucketed by Timestamp.  :pr:`22890`
* Add config `hive.legacy-timestamp-bucketing` and session property ``hive.legacy_timestamp_bucketing`` to use the original hash function for Timestamp column, which is not hive compatible. :pr:`22890`

```

### DIFF
--- a/presto-hive/src/main/java/com/facebook/presto/hive/HiveBucketing.java
+++ b/presto-hive/src/main/java/com/facebook/presto/hive/HiveBucketing.java
@@ -87,14 +87,14 @@ public final class HiveBucketing
         return (getHashCode(types, page, position) & Integer.MAX_VALUE) % bucketCount;
     }
 
-    public static int getHiveBucket(int bucketCount, List<TypeInfo> types, Page page, int position)
+    public static int getHiveBucket(int bucketCount, List<TypeInfo> types, Page page, int position, boolean useLegacyTimestampBucketing)
     {
-        return (getBucketHashCode(types, page, position) & Integer.MAX_VALUE) % bucketCount;
+        return (getBucketHashCode(types, page, position, useLegacyTimestampBucketing) & Integer.MAX_VALUE) % bucketCount;
     }
 
-    public static int getHiveBucket(int bucketCount, List<TypeInfo> types, Object[] values)
+    public static int getHiveBucket(int bucketCount, List<TypeInfo> types, Object[] values, boolean useLegacyTimestampBucketing)
     {
-        return (getBucketHashCode(types, values) & Integer.MAX_VALUE) % bucketCount;
+        return (getBucketHashCode(types, values, useLegacyTimestampBucketing) & Integer.MAX_VALUE) % bucketCount;
     }
 
     private static int getHashCode(List<Type> types, Page page, int position)
@@ -108,29 +108,29 @@ public final class HiveBucketing
         return result;
     }
 
-    private static int getBucketHashCode(List<TypeInfo> types, Page page, int position)
+    private static int getBucketHashCode(List<TypeInfo> types, Page page, int position, boolean useLegacyTimestampBucketing)
     {
         checkArgument(types.size() == page.getChannelCount());
         int result = 0;
         for (int i = 0; i < page.getChannelCount(); i++) {
-            int fieldHash = hash(types.get(i), page.getBlock(i), position);
+            int fieldHash = hash(types.get(i), page.getBlock(i), position, useLegacyTimestampBucketing);
             result = result * 31 + fieldHash;
         }
         return result;
     }
 
-    private static int getBucketHashCode(List<TypeInfo> types, Object[] values)
+    private static int getBucketHashCode(List<TypeInfo> types, Object[] values, boolean useLegacyTimestampBucketing)
     {
         checkArgument(types.size() == values.length);
         int result = 0;
         for (int i = 0; i < values.length; i++) {
-            int fieldHash = hash(types.get(i), values[i]);
+            int fieldHash = hash(types.get(i), values[i], useLegacyTimestampBucketing);
             result = result * 31 + fieldHash;
         }
         return result;
     }
 
-    private static int hash(TypeInfo type, Block block, int position)
+    private static int hash(TypeInfo type, Block block, int position, boolean useLegacyTimestampBucketing)
     {
         // This function mirrors the behavior of function hashCode in
         // HIVE-12025 ba83fd7bff serde/src/java/org/apache/hadoop/hive/serde2/objectinspector/ObjectInspectorUtils.java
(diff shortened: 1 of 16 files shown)


presto-hive/src/test/java/com/facebook/presto/hive/TestHiveBucketing.java
```diff
@@ -109,6 +109,10 @@ public void testHashingCompare()
         assertBucketEquals("timestamp", new Timestamp(1000 * LocalDateTime.of(1969, 12, 31, 23, 59, 59, 999_000_000).toEpochSecond(ZoneOffset.UTC)));
         assertBucketEquals("timestamp", new Timestamp(1000 * LocalDateTime.of(1950, 11, 19, 12, 34, 56, 789_000_000).toEpochSecond(ZoneOffset.UTC)));
         assertBucketEquals("timestamp", new Timestamp(1000 * LocalDateTime.of(2015, 11, 19, 7, 6, 5, 432_000_000).toEpochSecond(ZoneOffset.UTC)));
+        assertBucketEquals("timestamp", new Timestamp(10 + 1000 * LocalDateTime.of(1970, 1, 1, 0, 0, 0, 0).toEpochSecond(ZoneOffset.UTC)));
```
rschlussel, line 112: Can you split this into a separate test (e.g. testTimestampBucketWithNanoseconds)

rschlussel (commented):
This is a great find!  Can you add a release note for this bug fix.  

This could also cause problems with incompatibility between old partitions and new partitions using different bucketing algorithms, or even possibly wrong results reading old partitions and filtering out the wrong buckets.  I wonder if we need to add a configuration property for hive(e.g. called something like legacy-timestamp-bucketing) to help with the migration here.

presto-hive/src/main/java/com/facebook/presto/hive/HiveSessionProperties.java
```diff
@@ -641,6 +642,11 @@ public HiveSessionProperties(HiveClientConfig hiveClientConfig, OrcFileWriterCon
                         AFFINITY_SCHEDULING_FILE_SECTION_SIZE,
                         "Size of file section for affinity scheduling",
                         hiveClientConfig.getAffinitySchedulingFileSectionSize(),
+                        false),
+                booleanProperty(
+                        LEGACY_TIMESTAMP_BUCKETING,
+                        "Use legacy bucketing algorithm for timestamp type for bucketed table",
+                        false,
```
rschlussel, line 649: Make a configuration property for this in HiveClientConfig and set the default value for this session property from there.

presto-hive/src/test/java/com/facebook/presto/hive/TestHiveBucketing.java
kewang1024: Since all the data types are tested within `testHashingCompare`, I would keep them at the same place since testing millisecond doesn't sound like an edge case, it's just the original test somehow failed to include those. I added comment to make it more straightforward.

Let me know if you feel strongly against it.

rschlussel: my motivation is basically that when you have a zillion test cases in a single test it gets annoying to debug when it fails, and you lose some signal (it fails on case x, so the next case doesn't run, then you fix x and run the test again to discover there was also an issue with y).  I think ideally the original test is split up too, and since you're adding new tests, it's better at least not to exacerbate the  problem.

sdruzkin (dismissed):
ORC stuff LGTM

presto-hive-common/src/main/java/com/facebook/presto/hive/HiveCommonSessionProperties.java
```diff
@@ -176,6 +177,11 @@ public HiveCommonSessionProperties(HiveCommonClientConfig hiveCommonClientConfig
                         READ_MASKED_VALUE_ENABLED,
                         "Return null when access is denied for an encrypted parquet column",
                         hiveCommonClientConfig.getReadNullMaskedParquetEncryptedValue(),
+                        false),
+                booleanProperty(
+                        LEGACY_TIMESTAMP_BUCKETING,
+                        "Use legacy timestamp bucketing algorithm (which is not Hive compatible) for table bucketed by timestamp type.",
```
jaystarshot, line 183: Maybe explain what the legacy bucketing was

jaystarshot (dismissed):

kewang1024:
Other PRs are also failing on singlestore tests so it's unrelated to this change, @rschlussel

ClarenceThreepwood (dismissed):
This is a pretty insidious bug. Thanks for fixing!
nit - commit message is too long

presto-hive/src/main/java/com/facebook/presto/hive/HiveBucketing.java
```diff
@@ -233,42 +231,54 @@ private static int hash(TypeInfo type, Object value)
...
 
-    private static int hashOfMap(MapTypeInfo type, Block singleMapBlock)
+    private static int getHashForTimestamp(long millisSinceEpoch, boolean useLegacyTimestampBucketing)
+    {
+        if (useLegacyTimestampBucketing) {
+            // seconds << 30 + milliseconds
+            long secondsAndMillis = (Math.floorDiv(millisSinceEpoch, 1000L) << 30) + Math.floorMod(millisSinceEpoch, 1000L);
+            return (int) ((secondsAndMillis >>> 32) ^ secondsAndMillis);
+        }
+        else {
```
rschlussel, line 258: nit: this doesn't need to be inside an else block.

rschlussel (commented):
looks good. thanks for the fix!

rschlussel (approved):

steveburnett:
Suggest minor changes to release note entry

```
== RELEASE NOTES ==

Hive Connector Changes
* Fix hash calculation for Timestamp column to be hive compatible when writing to a table bucketed by Timestamp.  :pr:`22890`
* Add config `hive.legacy-timestamp-bucketing` and session property ``hive.legacy_timestamp_bucketing`` to use the original hash function for Timestamp column, which is not hive compatible. :pr:`22890`
```

rschlussel (approved):

PR #22989, opened 2024-06-11
Prioritize Facts over HBO stats

abhinavmuk04:
## Description
Create logic where CBO Facts will be chosen over HBO and other high confidence statistics

## Motivation and Context
Currently, we always prioritize HBO, however with the new Enum change we can work to return the statistics with the highest confidence level.

## Impact
When there are both HBO and CBO stats available, only return the statistics with the highest confidence. This will help improve optimizer.

## Test Plan
Created new tests where both HBO and Fact CBO were available ensuring that FACT is chosen in its respective cases.

## Contributor checklist

- [x] CI passed.

## Release Notes
== NO RELEASE NOTE ==

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/cost/HistoryBasedPlanStatisticsCalculator.java
+++ b/presto-main/src/main/java/com/facebook/presto/cost/HistoryBasedPlanStatisticsCalculator.java
@@ -44,6 +44,7 @@ import static com.facebook.presto.common.RuntimeMetricName.HISTORY_OPTIMIZER_QUE
 import static com.facebook.presto.common.RuntimeUnit.NANO;
 import static com.facebook.presto.cost.HistoricalPlanStatisticsUtil.getSelectedHistoricalPlanStatisticsEntry;
 import static com.facebook.presto.cost.HistoryBasedPlanStatisticsManager.historyBasedPlanCanonicalizationStrategyList;
+import static com.facebook.presto.spi.statistics.PlanStatistics.toConfidenceLevel;
 import static com.facebook.presto.sql.planner.iterative.Plans.resolveGroupReferences;
 import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.google.common.graph.Traverser.forTree;
@@ -201,7 +202,7 @@ public class HistoryBasedPlanStatisticsCalculator
                         Optional<HistoricalPlanStatisticsEntry> historicalPlanStatisticsEntry = getSelectedHistoricalPlanStatisticsEntry(entry.getValue(), inputTableStatistics.get(), historyMatchingThreshold);
                         if (historicalPlanStatisticsEntry.isPresent()) {
                             PlanStatistics predictedPlanStatistics = historicalPlanStatisticsEntry.get().getPlanStatistics();
-                            if (predictedPlanStatistics.getConfidence() > 0) {
+                            if ((toConfidenceLevel(predictedPlanStatistics.getConfidence()).getConfidenceOrdinal() >= delegateStats.confidenceLevel().getConfidenceOrdinal())) {
                                 return delegateStats.combineStats(
                                         predictedPlanStatistics,
                                         new HistoryBasedSourceInfo(entry.getKey().getHash(), inputTableStatistics, Optional.ofNullable(historicalPlanStatisticsEntry.get().getHistoricalPlanStatisticsEntryInfo())));
--- a/presto-main/src/test/java/com/facebook/presto/sql/planner/assertions/PlanMatchPattern.java
+++ b/presto-main/src/test/java/com/facebook/presto/sql/planner/assertions/PlanMatchPattern.java
@@ -38,6 +38,8 @@ import com.facebook.presto.spi.plan.SortNode;
 import com.facebook.presto.spi.plan.TopNNode;
 import com.facebook.presto.spi.plan.UnionNode;
 import com.facebook.presto.spi.plan.ValuesNode;
+import com.facebook.presto.spi.statistics.SourceInfo;
+import com.facebook.presto.spi.statistics.SourceInfo.ConfidenceLevel;
 import com.facebook.presto.sql.parser.ParsingOptions;
 import com.facebook.presto.sql.parser.ParsingOptions.DecimalLiteralTreatment;
 import com.facebook.presto.sql.parser.SqlParser;
@@ -781,6 +783,18 @@ public final class PlanMatchPattern
     public PlanMatchPattern withOutputRowCount(double expectedOutputRowCount)
     {
         matchers.add(new StatsOutputRowCountMatcher(expectedOutputRowCount));
+        return this;
+    }
+
+    public PlanMatchPattern withSourceInfo(SourceInfo sourceInfo)
+    {
+        matchers.add(new StatsSourceInfoMatcher(sourceInfo));
+        return this;
+    }
+
+    public PlanMatchPattern withConfidenceLevel(ConfidenceLevel confidenceLevel)
+    {
+        matchers.add(new StatsConfidenceLevelMatcher(confidenceLevel));
         return this;
     }
 
--- a/presto-tests/src/test/java/com/facebook/presto/execution/TestHistoryBasedStatsTracking.java
+++ b/presto-tests/src/test/java/com/facebook/presto/execution/TestHistoryBasedStatsTracking.java
@@ -27,10 +27,12 @@ import com.facebook.presto.spi.plan.PlanNode;
 import com.facebook.presto.spi.plan.ProjectNode;
 import com.facebook.presto.spi.plan.SortNode;
 import com.facebook.presto.spi.plan.TopNNode;
+import com.facebook.presto.spi.statistics.CostBasedSourceInfo;
 import com.facebook.presto.spi.statistics.HistoryBasedPlanStatisticsProvider;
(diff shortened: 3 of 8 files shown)


presto-main/src/main/java/com/facebook/presto/cost/HistoryBasedPlanStatisticsCalculator.java
```diff
@@ -201,7 +201,7 @@ private PlanNodeStatsEstimate getStatistics(PlanNode planNode, Session session,
                         Optional<HistoricalPlanStatisticsEntry> historicalPlanStatisticsEntry = getSelectedHistoricalPlanStatisticsEntry(entry.getValue(), inputTableStatistics.get(), historyMatchingThreshold);
                         if (historicalPlanStatisticsEntry.isPresent()) {
                             PlanStatistics predictedPlanStatistics = historicalPlanStatisticsEntry.get().getPlanStatistics();
-                            if (predictedPlanStatistics.getConfidence() > 0) {
+                            if (predictedPlanStatistics.getConfidence() > delegateStats.confidenceLevel().ordinal() - 1) {
```
rschlussel, line 204: would be clearer to convert the predictedPlanStatistics to a confidenceLevel and then compare the confidence levels directly.  This kind of logic is very error prone, both because it's complex and hard to reason about, and because it relies on assumptions about the relationship between PlanStatistics.getConfidence() and ConfidenceLevel.ordinal() that aren't enforced anywhere and could change.

You should also then add a test for the conversion between confidence and confidenceLevel.

presto-tests/src/test/java/com/facebook/presto/execution/TestHistoryBasedStatsTracking.java
```diff
@@ -471,6 +472,19 @@ public void testBroadcastJoin()
                         anyTree(any())));
     }
 
+    @Test
+    public void testFactPrioritization()
+    {
+        String query1 = "SELECT (SELECT nationkey FROM nation WHERE name = 'UNITED STATES') AS us_nationkey";
```
rschlussel, line 478: does these test fail without your change?  i assume both HBO and CBO will report the result as one
 You could add a PlanMatcher to check additionally that the source and confidence levels of  the statistics and from CBO and are FACT statistics.

You could also consider artificially registering historical stats with a different row count and expect that the results are still one.

presto-main/src/main/java/com/facebook/presto/cost/HistoryBasedPlanStatisticsCalculator.java
```diff
@@ -215,6 +218,16 @@ private PlanNodeStatsEstimate getStatistics(PlanNode planNode, Session session,
         return delegateStats;
     }
 
+    @VisibleForTesting
+    public ConfidenceLevel getEnumConfidenceLevel(PlanStatistics predictedPlanStatistics)
```
feilong-liu, line 222: This can be a static function?

presto-main/src/test/java/com/facebook/presto/cost/TestHistoryBasedPlanStatisticsCalculatorFunctions.java
```diff
@@ -0,0 +1,66 @@
...
+    @Test
+    public void testGetEnumConfidenceHigh()
+    {
+        HistoryBasedPlanStatisticsCalculator historyBasedPlanStatisticsCalculator;
+        Session session = testSessionBuilder()
+                .setSystemProperty(USE_HISTORY_BASED_PLAN_STATISTICS, "true")
+                .build();
+        LocalQueryRunner queryRunner = new LocalQueryRunner(session);
+
+        historyBasedPlanStatisticsCalculator = (HistoryBasedPlanStatisticsCalculator) queryRunner.getStatsCalculator();
```
feilong-liu, line 42: We do not need these if the function tested getEnumConfidenceLevel is static

presto-tests/src/test/java/com/facebook/presto/execution/TestHistoryBasedStatsTracking.java
```diff
@@ -471,6 +474,23 @@ public void testBroadcastJoin()
...
+        executeAndTrackHistory(query1);
+        assertPlan(query1, anyTree(node(EnforceSingleRowNode.class, anyTree(any()))
+                .withOutputRowCount(1)
+                .withSourceInfoAndConfidenceLevel(new CostBasedSourceInfo(FACT))));
+
+        Session session = Session.builder(createSession()).setSystemProperty("prefer_partial_aggregation", "false").build();
```
feilong-liu, line 486: Why we need this session property?

abhinavmuk04: This session property prevent getting a aggregation remote hash exchange error, which was a fix suggested by Sreeni

presto-main/src/main/java/com/facebook/presto/cost/HistoryBasedPlanStatisticsCalculator.java
rschlussel: I would move this to PlanStatistics.  

I would either make it non-static method on plan statistics like 
```
public ConfidenceLevel getConfidenceLevel()
```
or would make it static and as follows and take just a double
```
public static ConfidenceLevel toConfidenceLevel(double confidence)
```

presto-main/src/test/java/com/facebook/presto/sql/planner/assertions/PlanMatchPattern.java
```diff
@@ -784,6 +785,12 @@ public PlanMatchPattern withOutputRowCount(double expectedOutputRowCount)
         return this;
     }
 
+    public PlanMatchPattern withSourceInfoAndConfidenceLevel(SourceInfo sourceInfo)
```
rschlussel, line 788: I think as written, I would just call this `withSourceInfo()`. however, HistoryBasedSourceInfos are a bit hard to generate for tetsts, so it might be more useful to have two matchers that check what we are most interested in `withSourceInfoType(String expectedSourceInfoName)` that compares that the sourceInfo.getName() matches the expected and another `withConfidenceLevel(ConfidenceLevel expectedConfidenceLevel)` that checks that the confidence level matches the expected.

presto-main/src/main/java/com/facebook/presto/cost/HistoryBasedPlanStatisticsCalculator.java
```diff
@@ -201,7 +202,7 @@ private PlanNodeStatsEstimate getStatistics(PlanNode planNode, Session session,
                         Optional<HistoricalPlanStatisticsEntry> historicalPlanStatisticsEntry = getSelectedHistoricalPlanStatisticsEntry(entry.getValue(), inputTableStatistics.get(), historyMatchingThreshold);
                         if (historicalPlanStatisticsEntry.isPresent()) {
                             PlanStatistics predictedPlanStatistics = historicalPlanStatisticsEntry.get().getPlanStatistics();
-                            if (predictedPlanStatistics.getConfidence() > 0) {
+                            if (toConfidenceLevel(predictedPlanStatistics.getConfidence()).ordinal() >= delegateStats.confidenceLevel().ordinal()) {
```
jaystarshot, line 205: Why not add a comparator to the confidence level class? It will be useful in future comparisions

presto-spi/src/main/java/com/facebook/presto/spi/statistics/ConfidenceLevelComparator.java
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.spi.statistics;
+
+import com.facebook.presto.spi.statistics.SourceInfo.ConfidenceLevel;
+
+import java.util.Comparator;
+
+public class ConfidenceLevelComparator
```
feilong-liu, line 20: A separate class for comparison of confidence level seems too much here.

If needed, you can refer to https://github.com/prestodb/presto/blob/master/presto-common/src/main/java/com/facebook/presto/common/plan/PlanCanonicalizationStrategy.java on how to order enum.

feilong-liu (approved):

presto-main/src/test/java/com/facebook/presto/cost/TestHistoryBasedPlanStatisticsCalculatorFunctions.java
```diff
@@ -0,0 +1,47 @@
...
+import com.facebook.presto.spi.statistics.PlanStatistics;
+import com.facebook.presto.spi.statistics.TableWriterNodeStatistics;
+import org.testng.annotations.Test;
+
+import static com.facebook.presto.spi.statistics.PlanStatistics.toConfidenceLevel;
+import static com.facebook.presto.spi.statistics.SourceInfo.ConfidenceLevel.HIGH;
+import static com.facebook.presto.spi.statistics.SourceInfo.ConfidenceLevel.LOW;
+import static com.facebook.presto.testing.assertions.Assert.assertEquals;
+
+public class TestHistoryBasedPlanStatisticsCalculatorFunctions
```
rschlussel, line 28: call this class TestPlanStatistics because it's testing a function in PlanStatistics

rschlussel (commented):
minor comment, but generally looks good!

feilong-liu (approved):

rschlussel (approved):

PR #22991, opened 2024-06-12
Fix for previous correctness issue with TableScan Confidence

abhinavmuk04:
## Description
Fix for previous correctness issue with TableScan Confidence as there are a few cases where TableScan should be a low confidence

## Motivation and Context
TableScan Confidence should be set on whether or not there is filtering. If there is filtering it can be considered low and if not it can be considered high

## Test Plan
With amended code, all existing tests pass, ensuring that previous behavior is preserved.

## Contributor checklist

- [x] CI passed.

## Release Notes

== NO RELEASE NOTE ==

### DIFF
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/TableStatisticsMaker.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/TableStatisticsMaker.java
@@ -90,6 +90,7 @@ import static com.facebook.presto.iceberg.Partition.toMap;
 import static com.facebook.presto.iceberg.util.StatisticsUtil.calculateAndSetTableSize;
 import static com.facebook.presto.spi.statistics.ColumnStatisticType.NUMBER_OF_DISTINCT_VALUES;
 import static com.facebook.presto.spi.statistics.ColumnStatisticType.TOTAL_SIZE_IN_BYTES;
+import static com.facebook.presto.spi.statistics.SourceInfo.ConfidenceLevel.HIGH;
 import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.google.common.collect.Iterables.getOnlyElement;
 import static java.lang.Long.parseLong;
@@ -149,6 +150,7 @@ public class TableStatisticsMaker
         if (intersection.isNone()) {
             return TableStatistics.builder()
                     .setRowCount(Estimate.of(0))
+                    .setConfidenceLevel(HIGH)
                     .build();
         }
 
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/util/StatisticsUtil.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/util/StatisticsUtil.java
@@ -89,7 +89,7 @@ public final class StatisticsUtil
             }
             statsBuilder.setColumnStatistics(columnHandle, mergedStats.build());
         });
-        return calculateAndSetTableSize(statsBuilder).build();
+        return calculateAndSetTableSize(statsBuilder).setConfidenceLevel(icebergStatistics.getConfidence()).build();
     }
 
     public static EnumSet<ColumnStatisticType> decodeMergeFlags(String input)
--- a/presto-main/src/main/java/com/facebook/presto/cost/ConnectorFilterStatsCalculatorService.java
+++ b/presto-main/src/main/java/com/facebook/presto/cost/ConnectorFilterStatsCalculatorService.java
@@ -34,6 +34,7 @@ import java.util.Optional;
 
 import static com.facebook.presto.cost.FilterStatsCalculator.UNKNOWN_FILTER_COEFFICIENT;
 import static com.facebook.presto.cost.StatsUtil.toVariableStatsEstimate;
+import static com.facebook.presto.spi.statistics.SourceInfo.ConfidenceLevel.LOW;
 import static java.util.Objects.requireNonNull;
 
 public class ConnectorFilterStatsCalculatorService
@@ -80,7 +81,7 @@ public class ConnectorFilterStatsCalculatorService
             double totalSizeAfterFilter = filteredStatistics.getRowCount().getValue() / tableStatistics.getRowCount().getValue() * tableStatistics.getTotalSize().getValue();
             filteredStatsWithSize.setTotalSize(Estimate.of(totalSizeAfterFilter));
         }
-        return filteredStatsWithSize.build();
+        return filteredStatsWithSize.setConfidenceLevel(LOW).build();
     }
 
     private static PlanNodeStatsEstimate toPlanNodeStats(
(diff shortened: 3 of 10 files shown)


presto-main/src/main/java/com/facebook/presto/cost/TableScanStatsRule.java
```diff
@@ -71,8 +70,7 @@ protected Optional<PlanNodeStatsEstimate> doCalculate(TableScanNode node, StatsP
                 .setOutputRowCount(tableStatistics.getRowCount().getValue())
                 .setTotalSize(tableStatistics.getTotalSize().getValue())
 
-                // TODO Handle the confidence level properly when filters are pushed into the tablescan
-                .setConfidence(FACT)
+                // TODO Handle the confidence level as fact when there are no filters
```
kaikalur, line 73: It makes sense. We don't have reliable stats for scans even when there are no filters. So leaving it as default (low) is good.

kaikalur (approved):

presto-main/src/main/java/com/facebook/presto/cost/TableScanStatsRule.java
```diff
@@ -70,9 +69,6 @@ protected Optional<PlanNodeStatsEstimate> doCalculate(TableScanNode node, StatsP
         return Optional.of(PlanNodeStatsEstimate.builder()
                 .setOutputRowCount(tableStatistics.getRowCount().getValue())
                 .setTotalSize(tableStatistics.getTotalSize().getValue())
-
-                // TODO Handle the confidence level properly when filters are pushed into the tablescan
-                .setConfidence(FACT)
```
rschlussel, line 75: we do want it to be fact by default when there are no filters, as the stats on just the table without filters are generally reliable, and also the optimizations we run CBO are mostly before filters get pushed into the scan..  Removing this will significantly limit applications that rely on the isConfident() check, and can break exisiting queries.

presto-spi/src/main/java/com/facebook/presto/spi/statistics/TableStatistics.java
```diff
@@ -111,6 +121,7 @@ public static final class Builder
         private Estimate rowCount = Estimate.unknown();
         private Estimate totalSize = Estimate.unknown();
         private Map<ColumnHandle, ColumnStatistics> columnStatisticsMap = new LinkedHashMap<>();
+        private ConfidenceLevel confidenceLevel = FACT;
```
rschlussel, line 124: 1. @kaikalur pointed out that table scan statistics aren't facts like enforce single row is.  HIGH would be more appropriate.
2. I wonder if LOW is a better default, and then explicitly set something different when you know the confidence is higher (and probably most tablestatistics would set this to high).  I'm a bit torn about what the right default is here. @kaikalur what do you think?

presto-main/src/main/java/com/facebook/presto/cost/ConnectorFilterStatsCalculatorService.java
```diff
@@ -112,7 +114,7 @@ private static TableStatistics toTableStatistics(PlanNodeStatsEstimate planNodeS
         for (Map.Entry<VariableReferenceExpression, VariableStatsEstimate> entry : planNodeStats.getVariableStatistics().entrySet()) {
             builder.setColumnStatistics(columnByName.get(entry.getKey().getName()), toColumnStatistics(entry.getValue(), rowCount));
         }
-        return builder.build();
+        return builder.setConfidenceLevel(HIGH).build();
```
rschlussel, line 117: this should be set to the planNodeStats confidence level

presto-tpch/src/main/java/com/facebook/presto/tpch/TpchMetadata.java
```diff
@@ -392,7 +393,7 @@ private TableStatistics toTableStatistics(TableStatisticsData tableStatisticsDat
                 builder.setColumnStatistics(columnHandle, toColumnStatistics(stats, columnHandle.getType()));
             }
         });
-        return builder.build();
+        return builder.setConfidenceLevel(LOW).build();
```
rschlussel, line 396: this should be high. We have exact stats about tpch tables.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/TableStatisticsMaker.java
```diff
@@ -149,6 +150,7 @@ private TableStatistics makeTableStatistics(IcebergTableHandle tableHandle, Opti
         if (intersection.isNone()) {
             return TableStatistics.builder()
                     .setRowCount(Estimate.of(0))
+                    .setConfidenceLevel(LOW)
```
hantangwangd, line 153: Should we set `HIGH` or `FACT` confidence level for a definite `zero` table statistics?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/util/StatisticsUtil.java
```diff
@@ -89,7 +90,7 @@ public static TableStatistics mergeHiveStatistics(TableStatistics icebergStatist
             }
             statsBuilder.setColumnStatistics(columnHandle, mergedStats.build());
         });
-        return calculateAndSetTableSize(statsBuilder).build();
+        return calculateAndSetTableSize(statsBuilder).setConfidenceLevel(HIGH).build();
```
hantangwangd, line 93: Maybe the confidence level after merging shouldn't higher than any one of `icebergStatistics` and `hiveStatistics`, what do you think?

rschlussel:
the failure in com.facebook.presto.tpcds.TestTpcdsMetadataStatistics.testTableStatisticsSerialization is related to your changes.  Also, the actual results there show that the table statistics has a confidence of low, but I'd expect it to be high, so something should be fixed.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/util/StatisticsUtil.java
```diff
@@ -89,7 +89,7 @@ public static TableStatistics mergeHiveStatistics(TableStatistics icebergStatist
             }
             statsBuilder.setColumnStatistics(columnHandle, mergedStats.build());
         });
-        return calculateAndSetTableSize(statsBuilder).build();
+        return calculateAndSetTableSize(statsBuilder).setConfidenceLevel(statsBuilder.build().getConfidence()).build();
```
rschlussel, line 92: is this right? statsBuilder never sets the confidence, so seems like this will always be low. I think you first need to set it to the min confidence level as you are iterating through.  Also, would be good to add a test for this.

presto-spi/src/main/java/com/facebook/presto/spi/statistics/TableStatistics.java
```diff
@@ -111,6 +121,7 @@ public static final class Builder
         private Estimate rowCount = Estimate.unknown();
         private Estimate totalSize = Estimate.unknown();
         private Map<ColumnHandle, ColumnStatistics> columnStatisticsMap = new LinkedHashMap<>();
+        private ConfidenceLevel confidenceLevel = HIGH;
```
jaystarshot, line 124: Why is the default high? Should we keep this to unknown or something?

abhinavmuk04: @jaystarshot Previous code had it default as true, here we are keeping it at high which is lower than fact. In cases where there is filtering it is set to low. What do you think?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/util/StatisticsUtil.java
abhinavmuk04: @rschlussel only icebergStats have fine grained statistics, so only place where we can take confidenceLevel from is there.

rschlussel (dismissed):

hantangwangd (approved):

rschlussel (approved):

Issue #22585, opened 2024-04-22
[Native] - Correlated query returning incorrect results (Streaming agg assumes nested join produces pre-grouped results)


mknegi:
Query with correlated subquery in the exists subclause in where condition is returning incorrect results. Here is the testcase that fails due to this problem:
```
Testcase: presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java::testCorrelatedExistsSubqueries()
Run under derived class: TestTpchDistributedQueries
```

## Expected Behavior

java.lang.AssertionError: For query: 
 SELECT count(*) FROM orders o WHERE EXISTS(SELECT 1 FROM orders i WHERE o.orderkey < i.orderkey AND i.orderkey % 1000 = 0)
not equal
Actual rows (1 of 1 extra rows shown, 1 rows in total):
    [31762]
Expected rows (1 of 1 missing rows shown, 1 rows in total):
    [14999]

 If you see how the query in this scenario is constructed using the correlated subquery it should have only reported count(*) as 14999 because the orders table only have 15000 records in total. So, this query reporting so many extra rows shows that the semantics of correlated subquery aren’t somehow working in this case and that should generally be a defect.

I ran the select * from orders instead of select count(*) in the testcase and it seems there are lots of duplicate rows in the query output. This may explain why the row count was so much more than 14999 in this case. 
```
java.lang.AssertionError: For query: 
 SELECT * FROM orders o WHERE EXISTS(SELECT 1 FROM orders i WHERE o.orderkey < i.orderkey AND i.orderkey % 1000 = 0)
not equal
Actual rows (100 of 2062 extra rows shown, 17061 rows in total):
    [6535, 244, F, 207170.87, 1992-06-13, 1-URGENT, Clerk#000000650, 0, s. ironic, even theodolites haggle ]
    [6535, 244, F, 207170.87, 1992-06-13, 1-URGENT, Clerk#000000650, 0, s. ironic, even theodolites haggle ]
    [6560, 1204, F, 318034.04, 1995-01-05, 5-LOW, Clerk#000000328, 0, ven packages. deposits wake furiously enticingly regular theodolites; un]
    [6560, 1204, F, 318034.04, 1995-01-05, 5-LOW, Clerk#000000328, 0, ven packages. deposits wake furiously enticingly regular theodolites; un]
[...]

PR #22998, opened 2024-06-13
[native] Error out on StreamingAggregation

karteekmurthys:
## Description
This change disables StreamingAggregation by default. Refer this issue https://github.com/prestodb/presto/issues/22585. Prestissimo creates a StreamingAgg when there are pregrouped keys in the AggregationNode. This assumption breaks when the source to the Aggregation is a NestedLoopJoin.

## Test Plan

Migrated the Presto tests related correlated subqueries under presto-native.

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* ... :pr:`12345`
* ... :pr:`12345`

Hive Connector Changes
* ... :pr:`12345`
* ... :pr:`12345`
```

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp
+++ b/presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp
@@ -989,7 +989,11 @@ VeloxQueryPlanConverterBase::toVeloxQueryPlan(
   bool streamable = !node->preGroupedVariables.empty() &&
       node->groupingSets.groupingSetCount == 1 &&
       node->groupingSets.globalGroupingSets.empty();
-
+  // TODO karteekmurthys@ Re-enable after
+  // fixing:https://github.com/prestodb/presto/issues/22585
+  if (streamable) {
+    VELOX_UNSUPPORTED("StreamingAggregation is not supported");
+  }
   // groupIdField and globalGroupingSets are required for producing default
   // output rows for global grouping sets when there are no input rows.
   // Global grouping sets can be present without groupIdField in Final


mbasmanova (commented):
I feel the fix belongs to Presto's optimizer. Somewhere it has an assumption that LEFT JOIN preserves the order of probe-size rows. This assumption is not true for non-equi LEFT JOINs in Native Execution. It would be nice to find that place and add a check there for native execution.

aditi-pandit:
@karteekmurthys : Can you add a test ? We could add the query with incorrect results we were debugging.

aditi-pandit:
> I feel the fix belongs to Presto's optimizer. Somewhere it has an assumption that LEFT JOIN preserves the order of probe-size rows. This assumption is not true for non-equi LEFT JOINs in Native Execution. It would be nice to find that place and add a check there for native execution.

@mbasmanova : We are working to track down that code. In the interim we want to add this workaround to avoid the incorrect results.

karteekmurthys:
> @karteekmurthys : Can you add a test ? We could add the query with incorrect results we were debugging.

Yes, I am working on it. the PR is not ready yet.

karteekmurthys:
@aditi-pandit Please review, ported the tests.

mbasmanova:
@karteekmurthys @aditi-pandit Folks, please, reconsider this fix. Presto-to-Velox plan conversion is not the right place for it. Do take the time to debug fully and identify a place in the optimizer that needs to be made conditional on native execution. Thanks.

aditi-pandit:
> @karteekmurthys @aditi-pandit Folks, please, reconsider this fix. Presto-to-Velox plan conversion is not the right place for it. Do take the time to debug fully and identify a place in the optimizer that needs to be made conditional on native execution. Thanks.

@mbasmanova : I agree that it would be great to disable the optimizer logic. But Im curious why you have concerns about this code. This fix saves the engine from itself. We have incorrect results right now which is a very severe problem. Even if we fix the optimizer code, we might still want to retain this code as it catches a big vulnerability. Its possible that in the future someone could still write optimizer code making that assumption, and it would need to be caught in code review by someone who understands native engine as well. There aren't many engineers who understand both pieces.

If the concern is that we will lose track on the optimizer fix, I assure you we are not closing this thread. We are looking at the optimizer code.

mbasmanova:
@aditi-pandit Aditi, indeed, correctness issues are very important. We must ensure Prestissimo returns correct results. 

The problem with this change is that it adds a one-off logic to Presto-to-Velox plan translation that attempts to "fix" an optimizer bug. This logic is very limited and hard to reason about.

To prevent against future breakages one needs to invest in a robust suite of tests. Specifically, we should add a test that fails without the changes.

mbasmanova (changes requested):
See comments above.

aditi-pandit:
@mbasmanova : Was looking at the optimizer code bit more today. It seems like the addLocalExchanges rule is where the problem might have been introduced (see explanation at https://github.com/prestodb/presto/issues/22585#issuecomment-2168942249)

Do you have any suggestions ?

aditi-pandit:
@karteekmurthys : Please can you rebase this PR. We are considering submitting this as Meta encountered this issue in production as well.

presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp
```diff
@@ -1112,6 +1114,11 @@ core::JoinType toJoinType(protocol::JoinType type) {
 
   VELOX_UNSUPPORTED("Unknown join type");
 }
+
+inline bool isNestedLoopJoin(core::JoinType joinType) {
```
aditi-pandit, line 1118: @karteekmurthys : You also have to check for node->criteria.empty() here.

presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp
```diff
@@ -1638,7 +1645,15 @@ core::PlanNodePtr VeloxQueryPlanConverterBase::toVeloxQueryPlan(
   }
   if (auto aggregation =
           std::dynamic_pointer_cast<const protocol::AggregationNode>(node)) {
-    return toVeloxQueryPlan(aggregation, tableWriteInfo, taskId);
+    bool isStreamingAggAllowed = true;
+    if (auto joinNode = std::dynamic_pointer_cast<protocol::JoinNode>(
```
rschlussel, line 1649: what if nested loop join is not a direct source, but is in the plan tree (e.g. there's a project node in between)?

aditi-pandit: @rschlussel : We saw this property set by the optimizer in aggregation over nested loop join via StreamPropertyDerivations https://github.com/prestodb/presto/pull/23315. But you are right that the ideal fix should look for all nodes that can carry forward projections of the grouping keys.

presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp
```diff
@@ -1111,6 +1113,13 @@ core::JoinType toJoinType(protocol::JoinType type) {
 
   VELOX_UNSUPPORTED("Unknown join type");
 }
+
+inline bool isNestedLoopJoin(const std::shared_ptr<protocol::JoinNode> node) {
+  auto joinType = toJoinType(node->type);
+  return node->criteria.empty() || core::isInnerJoin(joinType) ||
```
aditi-pandit, line 1119: @karteekmurthys The condition is just node->criteria.empty(). The rest of the clauses catch all joinTypes. Unless I am missing something.

presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp
rschlussel: My question is what if you have like so


aggregation
   |
filter or identity project (any node that propagates the properties of its input)
  |
nested loop join
 |
something sorted

the nested loop join will mess up the sortedness, but we won't catch it with this code because it's not a direct source of the aggregation node.  so we will still have a correctness bug.

aditi-pandit: @rschlussel : Yes, the ideal fix should check for any nested loop join between aggregation and the TableScan it is originating from. I believe @karteekmurthys  is going to change this code.

karteekmurthys: @rschlussel I am trying to see if I can do a recursive search for a nestedloopjoin node and  if present skip streaming.

presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp
```diff
@@ -1637,7 +1639,18 @@ core::PlanNodePtr VeloxQueryPlanConverterBase::toVeloxQueryPlan(
   }
   if (auto aggregation =
           std::dynamic_pointer_cast<const protocol::AggregationNode>(node)) {
-    return toVeloxQueryPlan(aggregation, tableWriteInfo, taskId);
+    bool isStreamingAggAllowed = true;
```
aditi-pandit, line 1642: @karteekmurthys Nit : Abstract this into a function.

Please also add a comment about why this check is required.

presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp
```diff
@@ -1637,7 +1639,18 @@ core::PlanNodePtr VeloxQueryPlanConverterBase::toVeloxQueryPlan(
...
+    auto parent = aggregation->source;
+    while (parent.get()) {
```
rschlussel, line 1644: You can exit the loop early if isStreamingAggAllowed becomes false

presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp
```diff
@@ -1637,7 +1639,18 @@ core::PlanNodePtr VeloxQueryPlanConverterBase::toVeloxQueryPlan(
...
+      if (auto joinNode =
+              std::dynamic_pointer_cast<protocol::JoinNode>(parent)) {
+        isStreamingAggAllowed = !joinNode->criteria.empty();
```
rschlussel, line 1647: If you aren't exiting the loop early, then you need to && this with isStreamingAggAllowed so you don't overwrite a false value with a true value (e.g. you have a join on top of a join, and only the top one is nestedLoopJoin)

rschlussel: Edit: sorry misread the code, I see you are exiting early, but you should only exit if it's false. Otherwise you should continue.

karteekmurthys: @rschlussel I am still working on this. The `source` needs to be moved to the PlanNode and I cannot edit the presto_protocol.h directly for that. I am still figuring out best way to check the ancestors.

rschlussel:
Can we do a quick fix where we disable streaming aggregation for all cases (even if there's not a nested loop join, so we can quickly mitigate the correctness bug, and then we can iterate to limit the scope to just nested loop joins, or the optimizer change)?

presto-native-execution/presto_cpp/presto_protocol/presto_protocol.h
```diff
@@ -297,6 +297,7 @@ void from_json(const json& j, std::shared_ptr<ConnectorTableHandle>& p);
 namespace facebook::presto::protocol {
 struct PlanNode : public JsonEncodedSubclass {
   PlanNodeId id = {};
+  std::shared_ptr<PlanNode> source = {};
```
aditi-pandit, line 300: @karteekmurthys : Don't think you can change this without java side changes.

presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp
aditi-pandit: @rschlussel : This fix can be simplified since we only allow Project, Filter or Limit, DistinctLimit in between. In any other case streaming should be immediately returned false.

Think Karteek is changing to that.

But if this doesn't wrap up in the next round, we can do with never setting pre-grouping.

presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp
```diff
@@ -1637,7 +1639,9 @@ core::PlanNodePtr VeloxQueryPlanConverterBase::toVeloxQueryPlan(
   }
   if (auto aggregation =
           std::dynamic_pointer_cast<const protocol::AggregationNode>(node)) {
-    return toVeloxQueryPlan(aggregation, tableWriteInfo, taskId);
+    auto parent = aggregation->source;
```
aditi-pandit, line 1642: @karteekmurthys :
Nit : parent is not needed.

You can simply change line 990 to streamable = false (leaving a comment). There isn't a need to pass a variable from this point

karteekmurthys: Done.

presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp
```diff
@@ -986,9 +986,9 @@ VeloxQueryPlanConverterBase::toVeloxQueryPlan(
       VELOX_UNSUPPORTED("Unsupported aggregation step");
   }
 
-  bool streamable = !node->preGroupedVariables.empty() &&
-      node->groupingSets.groupingSetCount == 1 &&
-      node->groupingSets.globalGroupingSets.empty();
+  /// TODO karteekmurthys@ Re-enable after
```
aditi-pandit, line 989: Nit : There isn't a need to surface this in external documentation.

Just using // is fine.

karteekmurthys: Done.

aditi-pandit (approved):
Thanks @karteekmurthys

presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp
```diff
@@ -986,9 +986,9 @@ VeloxQueryPlanConverterBase::toVeloxQueryPlan(
       VELOX_UNSUPPORTED("Unsupported aggregation step");
   }
 
-  bool streamable = !node->preGroupedVariables.empty() &&
-      node->groupingSets.groupingSetCount == 1 &&
-      node->groupingSets.globalGroupingSets.empty();
+  // TODO karteekmurthys@ Re-enable after
+  // fixing:https://github.com/prestodb/presto/issues/22585
+  bool streamable = false;
```
mbasmanova, line 991: Is this safe? If planner planned streaming agg it means it has some assumptions that are wrong. Failing the query might be safer.

aditi-pandit: @mbasmanova : streamable just sets preGroupedVariables for Velox AggregationNode.

https://github.com/prestodb/presto/blob/master/presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp#L1010 

Presto planner doesn't really specify streaming agg. The streaming agg choice is made in Velox planner based on preGrouped variables https://github.com/facebookincubator/velox/blob/main/velox/exec/LocalPlanner.cpp#L499

So I think blanking out preGrouped variables is sufficient for letting the aggregation proceed with HashAggregation.

Let me know if you think otherwise.

mbasmanova: @aditi-pandit Aditi, you are correct that replacing streaming agg with hash agg is safe (just less efficient). However, my concern is that there could be other changes to the plan made by the optimizer based on wrong assumptions that led it to specify pre-grouped fields "incorrectly". In general, it is not safe to "question" the optimizer. However, here we say that we've seen cases where agg with pre-grouped keys was a result of optimizer making incorrect assumptions, hence, we are not following that choice. What I'm saying is that if part of the plan cannot be trusted, I don't think we can trust the rest of the plan. Hence, I'm saying that it would be safer to reject the whole plan and fail.

aditi-pandit: @mbasmanova : The pre-Grouping seemed broken only with nestedLoopJoin, but okay otherwise. 

The current fix to disable the streaming agg always is also aggressive to avoid fine-tuning plan patterns for aggregations over Nested Loop joins.

I feel rejecting the whole plan would be super conservative. But if you want that behavior in Meta production I won't question it. We will likely change the behavior internally at IBM.

rschlussel: My understanding is that a proper fix will be ready in Velox soon, so hopefully no Presto mitigation will be needed at all. But FWIW, I don't think there are other things relying on the pregrouped aggregation variables, so I think that aspect of just converting to a regular aggregation that ignores the pregrouped variables is pretty safe for a short term fix (wouldn't rely on it longer term because of course things can change). 

The riskier part that neither PR addresses is whether other optimizations unrelated to aggregation are also depending on nested loop join maintaining the ordering of the input. Would we plan a merge join after a nested loop join because we think the data is ordered?  Or drop an order by after a nested loop join because we think the data is already ordered?  If we'd like to be conservative/thorough, the safest thing to do is rather than specifically handling aggregation, fail all queries that are using NestedLoopJoin, which is the source of the incorrect optimization.

amitkdutta: @karteekmurthys Just discussed with @aditi-pandit. We agreed to fail the queries in this case, similar to what @rschlussel and @mbasmanova suggested. Lets change it to throw an exception instead of disabling streaming aggregation.

karteekmurthys: We currently use NestedLoopJoin when there is no Join criteria sepcified. If we totally disable it, we don't have alternative way to implement this situation.

```
if (node->criteria.empty()) {
    const bool isNestedJoinType = core::isInnerJoin(joinType) ||
        core::isLeftJoin(joinType) || core::isRightJoin(joinType) ||
        core::isFullJoin(joinType);
    if (isNestedJoinType) {
      return std::make_shared<core::NestedLoopJoinNode>(
          node->id,
          joinType,
          node->filter ? exprConverter_.toVeloxExpr(*node->filter) : nullptr,
          toVeloxQueryPlan(node->left, tableWriteInfo, taskId),
          toVeloxQueryPlan(node->right, tableWriteInfo, taskId),
          toRowType(node->outputVariables, typeParser_));
    }
    VELOX_UNSUPPORTED(
        "JoinNode has empty criteria that cannot be "
        "satisfied by NestedJoinNode or HashJoinNode");
  }
```

@mbasmanova Can you confirm what is the consequence of disabling Nestedloopjoin for Meta's workloads?

presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp
```diff
@@ -989,7 +989,11 @@ VeloxQueryPlanConverterBase::toVeloxQueryPlan(
   bool streamable = !node->preGroupedVariables.empty() &&
       node->groupingSets.groupingSetCount == 1 &&
       node->groupingSets.globalGroupingSets.empty();
-
+  // TODO karteekmurthys@ Re-enable after
+  // fixing:https://github.com/prestodb/presto/issues/22585
+  if (streamable) {
```
rschlussel, line 994: sorry for all the back and forth. instead of failing in plan conversion for aggregation as you are doing here, can we fail in plan conversion for join if it's a nested loop join? That would be a more robust mitigation because it's the root cause of the issue, and there are other optimizations aside from streaming aggregation that might be assuming nested loop join is preserving order when it's not e.g. we might remove an order by because we think the input is already ordered or plan a merge join because we think the input is ordered.

presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp
rschlussel: If we fail for this case it means that a small percentage of queries will start to fail in our native workers and need to run in Java clusters (and even though it is a small percentage, you are right that it's still a much larger impact than just failing for streaming aggregations, and it's not impact we want to take on lightly).  still we think it's the right think to do for the short term because even though most nested loop join queries are probably fine, we can't guarantee the correctness of queries that have nested loop joins in them.

presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp
rschlussel: oh looks like https://github.com/prestodb/presto/pull/23341/files is already doing this.

kgpai: FYI I am turning off NLJ here : https://github.com/prestodb/presto/pull/23341 .

presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp
aditi-pandit: @rschlussel : NLJ is used for any join where there isn't any equality condition between join columns. Sounds like you have have very little traffic for this.

@karteekmurthys : We should do the conservative approach in the open source project. We will have to internally enable this for us at IBM separately. Please follow up with Ethan for that.

pedroerp:
This PR should address the NLJ issue in Velox: https://github.com/facebookincubator/velox/pull/10651

PR #23003, opened 2024-06-13
Implement array_sum function in Java from SQL

abhinavmuk04:
## Description
Implement array_sum function in Java from SQL

## Motivation and Context
array_sum is currently written as a SQL Function, but it is not efficient. Implementing it as a Java function will help improve performance. Code is written with two hidden scalar functions, because Java will not allow for two functions with same scalar function name but different types to work. Additionally, generic implementation would return Integer as Integer as opposed to BigInt which does not match previous behavior.

## Impact
Utilizing SqlArraySumBenchmark class, previous version showed 633.26 cpu ms and current version shows 378.283 cpu ms

## Test Plan
Existing tests pass, showing that changes match previous behavior. Additionally, tests whose behavior was changed have been amended.

## Contributor checklist

- [ ] CI passed.

## Release Notes

== NO RELEASE NOTE ==

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/operator/scalar/ArraySumBigIntFunction.java
+++ b/presto-main/src/main/java/com/facebook/presto/operator/scalar/ArraySumBigIntFunction.java
@@ -1,0 +1,59 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.operator.scalar;
+
+import com.facebook.presto.common.block.Block;
+import com.facebook.presto.common.type.Type;
+import com.facebook.presto.spi.function.Description;
+import com.facebook.presto.spi.function.OperatorDependency;
+import com.facebook.presto.spi.function.ScalarFunction;
+import com.facebook.presto.spi.function.SqlType;
+import com.facebook.presto.spi.function.TypeParameter;
+
+import java.lang.invoke.MethodHandle;
+
+import static com.facebook.presto.common.function.OperatorType.ADD;
+import static com.facebook.presto.util.Failures.internalError;
+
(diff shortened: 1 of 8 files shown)


kaikalur:
Add some explanation for why you are doing it this way

presto-main/src/main/java/com/facebook/presto/operator/scalar/ArraySumBigIntFunction.java
```diff
@@ -0,0 +1,60 @@
...
+import com.facebook.presto.spi.function.TypeParameter;
+
+import java.lang.invoke.MethodHandle;
+
+import static com.facebook.presto.common.function.OperatorType.ADD;
+import static com.facebook.presto.spi.function.SqlFunctionVisibility.HIDDEN;
+import static com.facebook.presto.util.Failures.internalError;
+
+@Description("Returns the sum of all array elements, or 0 if the array is empty. Ignores null elements.")
+@ScalarFunction(value = "arraySumBigInt", visibility = HIDDEN)
```
rschlussel, line 31: this should be called array_sum, and not be hidden

presto-main/src/main/java/com/facebook/presto/operator/scalar/ArraySumDoubleFunction.java
```diff
@@ -0,0 +1,62 @@
...
+
+import java.lang.invoke.MethodHandle;
+
+import static com.facebook.presto.common.function.OperatorType.ADD;
+import static com.facebook.presto.common.type.TypeUtils.readNativeValue;
+import static com.facebook.presto.spi.function.SqlFunctionVisibility.HIDDEN;
+import static com.facebook.presto.util.Failures.internalError;
+
+@Description("Returns the sum of all array elements, or 0 if the array is empty. Ignores null elements.")
+@ScalarFunction(value = "arraySumDouble", visibility = HIDDEN)
```
rschlussel, line 32: call this array_sum and mark it not hidden.

kaikalur: See the original comment - there are subtle type issues and we don't want to spend too much time working on this as java is going away. So I suggested this as a pragmatic way of getting the benefit for users without too much soon-to-be-thrown-away work

rschlussel: I see, I missed that part.  So we can keep the two implementations, but still can directly call this array_sum and not make it hidden so we can get rid of the sql function

rschlussel (commented):
Can you move the tests from TestArraySqlFunctions to a new class TestArraySum since they are no longer a sql function.  changes otherwise look good.

rschlussel (approved):

Issue #22995, opened 2024-06-12
[native] Possible SEGV in HttpClient::createTransaction


czentgr:
## Your Environment

* Presto version used: latest
* Storage (HDFS/S3/GCS..):
* Data source and connector used:
* Deployment (Cloud or On-prem): Cloud, CentOS9 base
* [Pastebin](https://pastebin.com/) link to the complete debug logs:

* compiler / OS: gcc12 / CentOS9 

## Expected Behavior
Tasks get executed.

## Current Behavior
When receiving a task request the presto_server executable may SEGV in HttpClient::createTransaction.
This code was added newly with PR https://github.com/prestodb/presto/pull/22903.

The signal address has varied from nullptr to what looks to be a valid address.

Example 1 with null ptr:
```
I20240610 15:45:20.713101    62 TaskManager.cpp:607] Deleting task 20240610_154517_00010_p52u3.12.0.0.0
*** Aborted at 1718034320 (unix time) try "date -d @1718034320" if you are using GNU date ***
PC: @                0x0 (unknown)
*** SIGSEGV (@0x0) received by PID 8 (TID 0x7f18f37fe640) from PID 0; stack trace: ***
I20240610 15:45:20.727877    68 Task.cpp:1871] Terminating task 20240610_154517_00010_p52u3.11.0.0.0 with state Finished after running for 1009 ms.
I20240610 15:45:20.733484    68 TaskManager.cpp:607] Deleting task 20240610_154517_00010_p52u3.11.0.0.0
    @     0x7f51c55a26f0 (unknown)
    @          0x1ed036d _ZZN8facebook6presto4http10HttpClient17createTransactionEPN8proxygen22HTTPTransactionHandlerEENKUlPNS3_15HTTPSessionBaseEE_clES7_
[...]

PR #23018, opened 2024-06-17
[native] Add check if session is reusable

majetideepak:
## Description
Check if the session is reusable before reusing it.
Resolves https://github.com/prestodb/presto/issues/22995

## Contributor checklist

- [ ] CI passed.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/http/HttpClient.cpp
+++ b/presto-native-execution/presto_cpp/main/http/HttpClient.cpp
@@ -460,6 +460,7 @@ folly::SemiFuture<proxygen::HTTPTransaction*> HttpClient::createTransaction(
     if (!session) {
       return nullptr;
     }
+    VELOX_CHECK(session->isReusable());
     VLOG(3) << "Reuse idle connection from different thread to "
             << address_.describe();
     auto* evb = session->getEventBase();


aditi-pandit:
@majetideepak : From your comment [here](https://github.com/prestodb/presto/issues/22995#issuecomment-2172781637 ) it seems like you didn't see any re-use after this fix. So is this ready for review ?

Don't want to regress anything internally in Meta if they are not encountering it.

@amitkdutta

majetideepak:
@aditi-pandit I think this fix is reasonable given the session API documentation says `isReusable` must be true for a session to be reused.
The current situation is bad since we are easily hitting the sigsegv in the OSS environment.
Since this is a blocking issue, I feel we should merge this and then look into the reuse issue.

Yuhta:
It seems a bug in proxygen so let's fix it there instead.  In the meantime anyone is affected can disable connection pool.

majetideepak:
The config to disable the connection pool is `exchange.http-client.enable-connection-pool`. I am closing this fix for now as it essentially disables the connection pool as well.

majetideepak:
> It seems a bug in proxygen so let's fix it there instead

@Yuhta, @amitkdutta It occurred to me that this PR helps catch such bugs in proxygen. If not this, should we at least add a `VELOX_CHECK`?

Yuhta:
@majetideepak It will crash all the time and we will be forced to disable connection pool even we are not observing crashes before.

majetideepak:
> It will throw all the time and we will be forced to disable connection pool even we are not observing crashes before.

@Yuhta If I understood your comment here https://github.com/prestodb/presto/issues/22995#issuecomment-2176106700 correctly, isn't the expectation that the `session->isReusable()` should always be true? If that is not the case, is there any other precondition/check that we can add to avoid a crash?

majetideepak:
Is `isReusable` buggy in Meta's setup too?

Yuhta:
@majetideepak The value of `isReusable` is just our guess and it might not be aligned with how `ServerIdleSessionController` is supposed to do (our other internal usage of `ServerIdleSessionController` do not check `isReusable` either).  So let's file the bug to expect the we don't get invalid `sock_` from `ServerIdleSessionController`.  Also did you run TSAN to make sure it's not some racing condition that caused this?

majetideepak:
@czentgr ran with TSAN enabled. He did not find anything from that run.

majetideepak:
> isReusable is just our guess and it might not be aligned with how ServerIdleSessionController is supposed to do

Closing again!

PR #23024, opened 2024-06-17
[native] Add isSidecarEnabled config flag to query runner

pdabre12:


tdcmeehan:
Let's add this with a test when those become available (so that we don't add code that's not used).

PR #23042, opened 2024-06-20
Log more detailed messages when HBO stats are missing

abhinavmuk04:
## Description
Log more detailed messages when HBO stats are missing

## Motivation and Context
Oftentimes, when HBO stats are missing users do not know why. There are many reasons why HBO statistics can fail to be fetched, but with this change it will make it more clear to the user why it may be the case.

## Impact
This will improve clarity, and help users when debugging any issues where HBO stats may not appear even when they think it should

## Contributor checklist

- [x] CI passed.

## Release Notes

== NO RELEASE NOTE ==

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/cost/HistoryBasedPlanStatisticsCalculator.java
+++ b/presto-main/src/main/java/com/facebook/presto/cost/HistoryBasedPlanStatisticsCalculator.java
@@ -42,6 +42,7 @@ import static com.facebook.presto.SystemSessionProperties.useHistoryBasedPlanSta
 import static com.facebook.presto.common.RuntimeMetricName.HISTORY_OPTIMIZER_QUERY_REGISTRATION_GET_PLAN_NODE_HASHES;
 import static com.facebook.presto.common.RuntimeMetricName.HISTORY_OPTIMIZER_QUERY_REGISTRATION_GET_STATISTICS;
 import static com.facebook.presto.common.RuntimeUnit.NANO;
+import static com.facebook.presto.common.RuntimeUnit.NONE;
 import static com.facebook.presto.cost.HistoricalPlanStatisticsUtil.getSelectedHistoricalPlanStatisticsEntry;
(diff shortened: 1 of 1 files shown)


presto-main/src/main/java/com/facebook/presto/cost/HistoryBasedPlanStatisticsCalculator.java
```diff
@@ -176,6 +182,7 @@ private Map<PlanCanonicalizationStrategy, PlanNodeWithHash> getPlanNodeHashes(Pl
     private PlanNodeStatsEstimate getStatistics(PlanNode planNode, Session session, Lookup lookup, PlanNodeStatsEstimate delegateStats)
     {
         if (!useHistoryBasedPlanStatisticsEnabled(session)) {
+            LOG.info("History based statistics are disabled for this session");
```
jaystarshot, line 185: This can create too much logging.

jaystarshot (commented):
This change can create too much logging

presto-main/src/main/java/com/facebook/presto/cost/HistoryBasedPlanStatisticsCalculator.java
abhinavmuk04: @jaystarshot After removing this, do you still think too much logging will occur if a user is looking to debug why HBO was not used

jaystarshot: If thats your goal, why not add runtime metrics and publish them through the event listener?

feilong-liu: Let's collect the information and record the reasons in an external table using the event listener so as to make it easy to use.

In addition to why HBO is not available for a whole query, sometimes HBO is available for some of the nodes but not for others, you will also need to record why a specific plan node lacks HBO statistics.

PR #23044, opened 2024-06-20
[native]Add system config for global memory arbitration in velox

xiaoxmeng:
```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/common/Configs.cpp
+++ b/presto-native-execution/presto_cpp/main/common/Configs.cpp
@@ -190,6 +190,7 @@ SystemConfig::SystemConfig() {
           BOOL_PROP(kEnableSerializedPageChecksum, true),
           BOOL_PROP(kUseMmapAllocator, true),
           STR_PROP(kMemoryArbitratorKind, ""),
+          BOOL_PROP(kMemoryArbitratorGlobalArbitrationEnabled, false),
           NUM_PROP(kQueryMemoryGb, 38),
           NUM_PROP(kQueryReservedMemoryGb, 4),
           BOOL_PROP(kEnableVeloxTaskLogging, false),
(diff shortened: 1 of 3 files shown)


presto-native-execution/presto_cpp/main/common/Configs.cpp
```diff
@@ -478,6 +479,10 @@ std::string SystemConfig::memoryArbitratorKind() const {
   return optionalProperty<std::string>(kMemoryArbitratorKind).value_or("");
 }
 
+bool SystemConfig::memoryArbitratorGlobalArbitrationEnabled() const {
+  return optionalProperty<bool>(kMemoryArbitratorGlobalArbitrationEnabled).value_or("");
```
tanjialiang, line 483: .value_or(false)

tanjialiang (dismissed):

presto-native-execution/presto_cpp/main/common/Configs.cpp
xiaoxmeng: All the system configs in Prestissimo doesn't provide the default here. I follow the same pattern.

aditi-pandit:
@xiaoxmeng : Please can you add a release note documenting this property.

To clarify by default a query can reclaim from another query, but this property restricts arbitration to only look at other tasks of the same query ? Or is it something else ?

xiaoxmeng:
> @xiaoxmeng : Please can you add a release note documenting this property.
> 
> To clarify by default a query can reclaim from another query, but this property restricts arbitration to only look at other tasks of the same query ? Or is it something else ?

By default, this option is off at Velox side. This provides a system config to turn it on in Prestissimo

tanjialiang (approved):

presto-native-execution/presto_cpp/main/common/Configs.cpp
tanjialiang: Yeah but you are assigning a string to a bool

tanjialiang (approved):

Issue #23001, opened 2024-06-13
Allow C++-defined system session properties


tdcmeehan:
## Background

With the introduction of the `presto-native-execution` module, Presto now has forked the execution engine into a separate process.  This allows for the execution engine to be written in C++ and to be more efficient than the Java execution engine.  However, this also means that the execution engine code is now entirely separate from the coordinator code, and that session properties which are set on the coordinator are not automatically propagated to the execution engine without a subsequent registration of the session property in the C++ code.  Furthermore, because the code has been forked, there are many session properties which are not supported in the C++ execution engine, and likewise, there are many session properties which are not supported by Java workers.

Ideally, only session properties which are supported by both the coordinator and the execution engine should be settable by the user.  For example, `SHOW SESSION` should only return session properties that are meaningful to the user.  This issues adds a mechanism to allow for session properties to be defined in C++ and to be settable by the user.

## Expected Behavior or Use Case
Hide session properties for evaluation engines which are not being used.

## Presto Component, Service, or Connector
Presto SPI, native execution module

## Possible Implementation
#22829 added a new sidecar process type to Presto.  This sidecar is a separate process which shares the same code as the Presto C++ worker.

A new endpoint will be added to the Presto sidecar which will return the session property mapping for all built-in and externally registered session properties which are implemented in C++.  It will contain a subset of the fields found in `PropertyMetadata`, and it will be serialized as JSON.

To support the introduction of new system session properties, a new SPI will be introduced which will allow for the registration of session properties which are not present in-memory in the Java runtime.  This SPI will be used to register the session properties which are returned by the sidecar process.  This SPI will be similar to the existing `Connector#getSessionProperties` SPI, except it will allow the registration of system session properties which are not connector-specific.

A new module will be developed which has the sole purpose of retrieving C++-defined session properties from the Presto sidecar process and provide them to the Presto coordinator using the new SPI.

In the `presto-native-execution` module, a new class will be developed which will define C++-specific session properties, and will expose them as an endpoint in the sidecar process.  On the worker side, it will be used to validate the session property, and to provide a mapping to a corresponding system property (if it exists), and ultimately to a default in Velox `QueryCtx`.  Each session property will be translated to a value in the `QueryCtx` for Velox, but in the future it may be used for any Presto C++-specific session properties that may exist as well.

[...]

PR #23045, opened 2024-06-20
[native] Adds a native session property provider

pdabre12:
## Description
Adds a native session property manager.

## Motivation and Context
To resolve https://github.com/prestodb/presto/issues/23001

## Impact
With this change, we can now retrieve session properties which are only supported by both the coordinator and the execution engine.

## Test Plan
Tests are included. 

## Contributor checklist

- [ ] CI passed.

## Release Notes

## Release Notes

```
== RELEASE NOTES ==

Coordinator Plugin Changes
* Add native system session property provider. :pr:`23045`
```

cc: @tdcmeehan @abevk2023 @Joe-Abraham


tdcmeehan:
We'll need to spend some time consolidating these commits.  It would be great to think about the following commits:

1) A commit which introduces the necessary SPI.  This includes the definition of the SessionPropertyMetadata.
2) A commit which integrates that SPI with the Presto runtime
3) An implmenentation of this plugin for the sidecar
4) An implementation of this plugin with the C++-specific worker properties, and with a config toggle to enable/disable

presto-main/src/main/java/com/facebook/presto/metadata/SessionPropertyManager.java
```diff
@@ -61,23 +74,84 @@ public final class SessionPropertyManager
...
+            NodeManager nodeManager)
+    {
+        requireNonNull(sessionPropertyProviderName, "sessionPropertyProviderName is null");
+        SystemSessionPropertyProviderFactory factory = systemSessionPropertyProviderFactories.get(sessionPropertyProviderName);
+        checkState(factory != null, "No factory for session property provider factory %s", sessionPropertyProviderName);
+        SystemSessionPropertyProvider sessionPropertyProvider = factory.create(properties,
+                new SessionPropertyContext(typeManager, nodeManager));
+
+        if (systemSessionPropertyProviders.putIfAbsent(sessionPropertyProviderName, sessionPropertyProvider) != null) {
+            throw new IllegalArgumentException(format("System session property provider is already registered for property provider [%s] ", sessionPropertyProviderName));
```
elharo, line 144: string concatenation is simpler and less bug prone

presto-main/src/main/java/com/facebook/presto/metadata/SessionPropertyManager.java
```diff
@@ -61,23 +74,84 @@ public final class SessionPropertyManager
...
+        }
+    }
+
+    public void addSessionPropertyProviderFactory(SystemSessionPropertyProviderFactory factory)
+    {
+        if (systemSessionPropertyProviderFactories.putIfAbsent(factory.getName(), factory) != null) {
+            throw new IllegalArgumentException(format("System Session property provider factory '%s' is already registered", factory.getName()));
```
elharo, line 151: string concatenation is simpler and less bug prone, especially when the placeholder is the final thing in the string

presto-main/src/main/java/com/facebook/presto/metadata/SessionPropertyManager.java
```diff
@@ -78,6 +84,16 @@ public SessionPropertyManager(List<PropertyMetadata<?>> systemSessionProperties)
...
     }
 
+    @Inject
+    public SessionPropertyManager(
+            SystemSessionProperties systemSessionProperties,
+            JavaFeaturesConfig javaFeaturesConfig,
+            NodeSpillConfig nodeSpillConfig)
+    {
+        this(systemSessionProperties.getSessionProperties());
+        this.systemSessionPropertyProviders.put("java-worker", new JavaWorkerSystemSessionPropertyProvider(javaFeaturesConfig, nodeSpillConfig));
```
tdcmeehan: Let's create a Map in Guice that registers all built-in session property providers, and then just inject the map and put all of them.

presto-main/src/main/java/com/facebook/presto/sessionpropertyproviders/JavaWorkerSystemSessionPropertyProvider.java
```diff
@@ -0,0 +1,223 @@
...
+    public static final String ORDER_BY_SPILL_ENABLED = "order_by_spill_enabled";
+    public static final String AGGREGATION_OPERATOR_UNSPILL_MEMORY_LIMIT = "aggregation_operator_unspill_memory_limit";
+    public static final String TOPN_OPERATOR_UNSPILL_MEMORY_LIMIT = "topn_operator_unspill_memory_limit";
+    public static final String TEMP_STORAGE_SPILLER_BUFFER_SIZE = "temp_storage_spiller_buffer_size";
+    public static final String OFFSET_CLAUSE_ENABLED = "offset_clause_enabled";
+    public static final String TABLE_WRITER_MERGE_OPERATOR_ENABLED = "table_writer_merge_operator_enabled";
+
+    private final List<PropertyMetadata<?>> sessionProperties;
+
+    public JavaWorkerSystemSessionPropertyProvider(JavaFeaturesConfig javaFeaturesConfig, NodeSpillConfig nodeSpillConfig)
```
tdcmeehan, line 51: As a follow-on to my previous comment, let's `@Inject` the dependencies here and statically define it in Guice.

presto-main/src/test/java/com/facebook/presto/sql/planner/TestLogicalPlanner.java
```diff
@@ -454,7 +454,7 @@ public void testDistinctOverConstants()
                                         project(ImmutableMap.of("hash", expression("combine_hash(bigint '0', coalesce(\"$operator$hash_code\"(orderstatus_35), 0))")),
                                                 project(
                                                         ImmutableMap.of("orderstatus_35", expression("'F'")),
-                                                tableScan("orders", ImmutableMap.of())))))));
+                                                        tableScan("orders", ImmutableMap.of())))))));
```
tdcmeehan, line 457: Please revert this whitespace only change

tdcmeehan (changes requested):
[Separate out java worker properties and introduce JavaFeaturesConfig](https://github.com/prestodb/presto/pull/23045/commits/42570a3fd84facbb7bad9a0acc93fc390cd28b15) ❓

tdcmeehan (commented):
`Introduce session properties SPI` ✅

presto-main/src/main/java/com/facebook/presto/metadata/SessionPropertyManager.java
```diff
@@ -155,9 +213,18 @@ private void updateWorkerSessionProperties()
 
     private List<PropertyMetadata<?>> getWorkerSessionProperties()
     {
-        return systemSessionPropertyProviders.values().stream()
-                .flatMap(manager -> manager.getSessionProperties().stream())
-                .collect(toImmutableList());
+        // Filter out java-worker related properties if native execution and coordinator sidecar is enabled
```
tdcmeehan, line 216: Coming back to my comments on the previous commits, instead of doing this, can we simply not register the Java worker session properties if `nativeExecution` and `coordinatorSidecar` are enabled. That way we don't need to filter here.

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/NoOpPlugin.java
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+package com.facebook.presto.sidecar;
+
+import com.facebook.presto.spi.Plugin;
+
+/**
+ * Todo: Remove this class when support for CoordinatorPlugin is added in presto-maven-plugin.
+ */
+public class NoOpPlugin
```
tdcmeehan, line 21: Can't we just manually check in the `META-INF` file?

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/sessionpropertyproviders/NativeSystemSessionPropertyProvider.java
```diff
@@ -0,0 +1,232 @@
...
+    private final JsonCodec<List<SessionPropertyMetadata>> nativeSessionPropertiesJsonCodec;
+
+    @Inject
+    public NativeSystemSessionPropertyProvider(
+            JsonCodec<List<SessionPropertyMetadata>> nativeSessionPropertiesJsonCodec,
+            NodeManager nodeManager, TypeManager typeManager)
+    {
+        this.nativeSessionPropertiesJsonCodec = nativeSessionPropertiesJsonCodec;
+        this.nodeManager = nodeManager;
+        this.typeManager = typeManager;
```
tdcmeehan:
```java

    @Inject
    public NativeSystemSessionPropertyProvider(
            JsonCodec<List<SessionPropertyMetadata>> nativeSessionPropertiesJsonCodec,
            NodeManager nodeManager,
            TypeManager typeManager)
    {
        this.nativeSessionPropertiesJsonCodec = requireNonNull(nativeSessionPropertiesJsonCodec, "nativeSessionPropertiesJsonCodec is null");
        this.nodeManager = requireNonNull(nodeManager, "nodeManager is null");
```

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/sessionpropertyproviders/NativeSystemSessionPropertyProvider.java
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    @Override
+    public List<PropertyMetadata<?>> getSessionProperties()
+    {
+        return fetchSessionProperties();
+    }
+
+    private List<PropertyMetadata<?>> fetchSessionProperties()
```
tdcmeehan, line 89: Instead of throwing dummy values, let's just throw, since we add this functionality in the subsequent commit.

tdcmeehan (commented):
[[native] Integrate session property SPI with Presto engine](https://github.com/prestodb/presto/pull/23045/commits/b88081fd8a7a18509ec7ec3e75b78dc0fa93030d) ❓

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/NativeQueryRunnerUtils.java
```diff
@@ -389,7 +396,8 @@ public static void createTableToTestHiddenColumns(QueryRunner queryRunner)
             try {
                 TimeUnit.SECONDS.sleep(2);
             }
-            catch (InterruptedException e) { }
```
tdcmeehan, line 392: Please revert

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/PrestoNativeQueryRunnerUtils.java
```diff
@@ -438,6 +445,12 @@ public static Optional<BiFunction<Integer, URI, Process>> getExternalWorkerLaunc
                                 "system-memory-gb=4%n" +
                                 "http-server.http.port=%d", discoveryUri, port);
 
+                        if (sidecarEnabled.get()) {
```
tdcmeehan, line 448: Can you help me understand why we are doing this?

tdcmeehan (commented):
[[native] Integrate session property plugin with sidecar](https://github.com/prestodb/presto/pull/23045/commits/11a8ab32dd3509dc76d304d083820d518a7e3baa) ❓

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/PrestoNativeQueryRunnerUtils.java
pdabre12: Removed these changes.

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/NoOpPlugin.java
pdabre12: `PrestoServer` does not find the `NativeSidecarPlugin` when I try to manually check in the `META-INF` file. 
The one way it works is packaging the `presto-native-sidecar-plugin ` module as `presto-plugin`. In that case, it requires an implementation of `Plugin.class` in the module.

pdabre12: Adding to the above discussion:

There are two ways to load the plugins:

1. Using the directory path (`../presto-native-sidecar-plugin/target`)
2. Using the `pom.xml` (`../presto-native-sidecar-plugin/pom.xml`)

All plugins implementing the `Plugin.class` are loaded via the `pom.xml` approach. We can do the same for `CoordinatorPlugin`, but there’s a slight complication.

In the `PluginDiscovery` class (see [this link](https://github.com/prestodb/presto/blob/26fd8dbd976e2b4341558c953e54bcfbad0a6198/presto-main/src/main/java/com/facebook/presto/server/PluginDiscovery.java#L50)), there's a check that ensures the artifact has the `presto-plugin` extension. This check fails because the new `presto-native-sidecar-plugin` module is not packaged as a `presto-plugin`, and hence doesn’t have the required `presto-plugin` extension.

If we were to package the `presto-native-sidecar-plugin` as a `presto-plugin`, it would require an implementation of `Plugin.class`, as specified [here](https://github.com/prestodb/presto-maven-plugin/blob/master/src/main/java/com/facebook/presto/maven/ServiceDescriptorGenerator.java#L82).

A potential fix would be to update the [presto-maven-plugin](https://github.com/prestodb/presto-maven-plugin/tree/master) repository to support packaging of the CoordinatorPlugin as a `presto-plugin`. Specifically, we should allow a module to be packaged as a `presto-plugin` if it contains either `Plugin.class` or `CoordinatorPlugin` implementations. Currently, the packaging fails if it doesn’t detect a `Plugin.class` implementation. Hence, the need for a `NoOpPlugin` class here.

presto-main/src/main/java/com/facebook/presto/metadata/SessionPropertyManager.java
```diff
@@ -61,21 +71,88 @@ public final class SessionPropertyManager
...
+    {
+        this(systemSessionProperties.getSessionProperties(), systemSessionPropertyProviders);
+        this.functionAndTypeManager = requireNonNull(functionAndTypeManager, "functionAndTypeManager is null");
+        this.nodeManager = requireNonNull(nodeManager, "nodeManager is null");
+        this.nativeExecution = requireNonNull(featuresConfig, "featuresConfig is null").isNativeExecutionEnabled();
+        this.coordinatorSidecar = requireNonNull(serverConfig, "serverConfig is null").isCoordinatorSidecarEnabled();
+    }
+
+    // required to load a "native" session property provider
+    public void loadSessionPropertyProvider()
```
tdcmeehan, line 127: I think the way this should work is, we should iterate over all values in the map, and load them all.  The Presto runtime cannot make any assumptions on what plugins are loaded, except for the Java session properties.

tdcmeehan (changes requested):

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/NoOpPlugin.java
tdcmeehan: Can you create an issue in the Presto Maven plugin repo?

presto-main/src/main/java/com/facebook/presto/testing/TestingSession.java
```diff
@@ -54,7 +59,8 @@ private TestingSession() {}
 
     public static SessionBuilder testSessionBuilder()
     {
-        return testSessionBuilder(new SessionPropertyManager());
+        return testSessionBuilder(new SessionPropertyManager(new SystemSessionProperties(),
```
tdcmeehan, line 62: Please remove this new constructor that adds the mapping.  Instead, in `SessionPropertyManager`, create a factory method that creates a new `SessionPropertyManager` and just loads up the Java session properties after construction.

tdcmeehan (changes requested):

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/NoOpPlugin.java
pdabre12: Opened an issue: https://github.com/prestodb/presto-maven-plugin/issues/20

rschlussel (commented):
This PR is super well organized. I love how it's broken down.
A couple general questions
1. what if there is a name conflict between a property provided by the plugin and a coordinator property?  Maybe we should add a prefix for worker properties similar to how connector properties get a prefix of the catalog name. 
2. Thinking about what we call this SPI, I'm wondering we should call it WorkerSessionPropertiesProvider instead of SystemSessionPropertiesProvider, as it's specifically about supplying session properties for the worker.  Coordinator properties still need to be separate, as any properties that determine the behavior of the coordinator should not be provided by a plugin (except maybe through e.g. plugin optimizers or something).  From that perspective, the JavaWorkerProperties should be treated exactly the same as the NativeWorkerProperties

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/ImplementOffset.java
```diff
@@ -78,8 +79,10 @@ public Pattern<OffsetNode> getPattern()
     @Override
     public Result apply(OffsetNode parent, Captures captures, Context context)
     {
-        if (!isOffsetClauseEnabled(context.getSession())) {
-            throw new PrestoException(NOT_SUPPORTED, "Offset support is not enabled");
+        if (!isNativeExecutionEnabled(context.getSession())) {
```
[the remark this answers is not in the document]
pdabre12: Yes, thank you for catching this, I will address this.

pdabre12:
@rschlussel Thank you for your comments.

> This PR is super well organized. I love how it's broken down. A couple general questions
> 
> 1. what if there is a name conflict between a property provided by the plugin and a coordinator property?  Maybe we should add a prefix for worker properties similar to how connector properties get a prefix of the catalog name.
> 2. Thinking about what we call this SPI, I'm wondering we should call it WorkerSessionPropertiesProvider instead of SystemSessionPropertiesProvider, as it's specifically about supplying session properties for the worker.  Coordinator properties still need to be separate, as any properties that determine the behavior of the coordinator should not be provided by a plugin (except maybe through e.g. plugin optimizers or something).  From that perspective, the JavaWorkerProperties should be treated exactly the same as the NativeWorkerProperties

Yes, there could be a name conflict between the properties, how does `native-worker.` or `sidecar-plugin.`  sound for the native worker properties and `java-worker.`  for the java worker properties?

Also calling the SPI WorkerSessionPropertiesProvider makes sense. I will incorporate this.

rschlussel:
> Yes, there could be a name conflict between the properties, how does `native-worker.` or `sidecar-plugin.` sound for the native worker properties and `java-worker.` for the java worker properties?

Those prefixes sound good.  Realizing we'll need a migration plan since people are probably relying on old names without a prefix. I don't want to block this change on that, so if needed we can merge without the prefix and add it in a follow on PR.

tdcmeehan:
+1 on the prefix idea.  To mitigate the effects of migration to the prefix based system, we could have the concept of a default prefix, similar to functions.

rschlussel:
we can't really have a default because we need the native-worker (or java-worker) properties to not conflict with coordinator properties, but both are expected to be in use in a typical prod environment.

tdcmeehan:
@rschlussel The way I see it, coordinator properties are prefixed with something like `default.`.  It has special meaning because if a prefix is not specified, then it is assumed to be the default.  A session property provider can opt to give itself the prefix `default` as well, perhaps by just naming its properties file in that way similar to connectors.  If it does, then when we try to register properties, we will detect that there is a conflict and throw an exception, causing the plugin to not load.

pdabre12:
@rschlussel Address your comments % the prefix related changes. 
Depending on the discussion, I will address the prefix related changes in a follow-up PR.
Can you please have another look?

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/sessionpropertyproviders/NativeSystemSessionPropertyProvider.java
```diff
@@ -50,28 +62,41 @@ public class NativeSystemSessionPropertyProvider
...
+            @ForSidecarInfo HttpClient httpClient,
             JsonCodec<List<SessionPropertyMetadata>> nativeSessionPropertiesJsonCodec,
             NodeManager nodeManager,
             TypeManager typeManager)
     {
         this.nativeSessionPropertiesJsonCodec = requireNonNull(nativeSessionPropertiesJsonCodec, "nativeSessionPropertiesJsonCodec is null");
         this.nodeManager = requireNonNull(nodeManager, "nodeManager is null");
         this.typeManager = requireNonNull(typeManager, "typeManager is null");
+        this.httpClient = requireNonNull(httpClient, "typeManager is null");
+        this.memoizedSessionPropertiesSupplier = Suppliers.memoizeWithExpiration(this::fetchSessionProperties, 1, TimeUnit.HOURS);
```
rschlussel, line 80: should we make this time limit configurable?

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/sessionpropertyproviders/NativeSystemSessionPropertyProvider.java
```diff
@@ -107,4 +132,15 @@ else if (type == TINYINT) {
         }
         return propertyMetadata;
     }
+
+    private URI getSidecarLocation()
+    {
+        Node sidecarNode = nodeManager.getSidecarNode();
+        return HttpUriBuilder.uriBuilder()
+                .scheme("http") // The sidecar is presumed to be colocated with the coordinator
```
rschlussel, line 140: what about this assumes the sidecar is colocated? I thought the earlier commit added support for the sidecar to be any worker.

rschlussel (commented):
Thanks for the changes!
A few more questions as I'm thinking about how this will fit with our production environment:
1. How do you set up having workers as sidecars? The commit that adds support for workers as sidecars looks like it just ensures that you can have multiples.  Is there a way to say use the whole worker pool for sidecars?
2. Since you delete the native session properties, does that mean that we can no longer run native workers without a sidecar?  What happens if coordinatorSidecarEnabled is set to false?
3. We have a heterogeneous environment where some of our clusters are cpp clusters and some are Java clusters. The same user query might sometimes run on cpp and sometimes in Java, so the user may specify both java and cpp session properties in their query as they don't know where they will be routed.  We need to ensure that this use case continues to work. Can we make it so that there's an option to not fail queries with invalid session properties if they specify properties from the other cluster type (and add a test for this case)?

presto-main/src/test/java/com/facebook/presto/sql/planner/iterative/rule/test/RuleTester.java
```diff
@@ -83,7 +84,9 @@ public RuleTester(List<Plugin> plugins, Map<String, String> sessionProperties, O
 
     public RuleTester(List<Plugin> plugins, Map<String, String> sessionProperties, Optional<Integer> nodeCountForStats, ConnectorFactory connectorFactory)
     {
-        this(plugins, sessionProperties, new SessionPropertyManager(), nodeCountForStats, connectorFactory);
+        this(plugins, sessionProperties,
```
rschlussel, line 87: formatting nit.  If we don't have everything on one line, we do one argument per line with the first argument on the line after the open parentheses. as follows:
```
myMethod(
      arg1,
      arg2,
      arg3);
```

pdabre12:
> Thanks for the changes! A few more questions as I'm thinking about how this will fit with our production environment:
> 
> 1. How do you set up having workers as sidecars? The commit that adds support for workers as sidecars looks like it just ensures that you can have multiples.  Is there a way to say use the whole worker pool for sidecars?
> 2. Since you delete the native session properties, does that mean that we can no longer run native workers without a sidecar?  What happens if coordinatorSidecarEnabled is set to false?
> 3. We have a heterogeneous environment where some of our clusters are cpp clusters and some are Java clusters. The same user query might sometimes run on cpp and sometimes in Java, so the user may specify both java and cpp session properties in their query as they don't know where they will be routed.  We need to ensure that this use case continues to work. Can we make it so that there's an option to not fail queries with invalid session properties if they specify properties from the other cluster type (and add a test for this case)?

1. Setting up workers as sidecars can be controlled using the config property `native-sidecar=true`. Every worker that needs to be set-up as a sidecar needs this config property.
2. That's actually a good point, it slipped my mind. In that case, deleting the native session properties won't work. I was thinking about adding a new `NativeWorkerSessionPropertyProvider` class and moving all the native worker related properties there. If coordinatorSidecarEnabled it set to false, inject the `NativeWorkerSessionPropertyProvider` class. If coordinatorSidecarEnabled is set to true, do not inject the `NativeWorkerSessionPropertyProvider` class instead we can pull the native session properties from the sidecar. 
3. In a heterogeneous environment where some of the clusters are cpp clusters and some are Java clusters, the only way that I can think of right now to not fail on an invalid property is to inject the `JavaWorkerSessionPropertyProvider` always irrespective of the fact that coordinatorSidecarEnabled / nativeExecution is set to true. In that case, if a `java-worker` only session property was redirected to a cpp cluster, it will still be able to read the property. Eventually, in the future  if the production environment consists only of cpp clusters, we can move away from injecting the `JavaWorkerSessionPropertyProvider`. 
What do you think?

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/sessionpropertyproviders/NativeSystemSessionPropertyProvider.java
pdabre12: Yes, the comment can be removed.

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/sessionpropertyproviders/NativeSystemSessionPropertyProvider.java
pdabre12: Yes, let me add a config property for this.

rschlussel:
This all sounds good. We can make it configurable whether to inject both or not, so people who need it can do it and when environments become 100% cpp we stop, and companies can do it at their own speed (until we fully remove java support from Presto). It would be cool if the java/native properties for the wrong cluster type had some setup that made them still hidden from show session commands.

pdabre12:
Maybe I can use the `nativeExecution` flag to conditionally hide session properties. If it's enabled, hide all `java-worker` only properties, if it's not hide all `native-worker` only properties?

rschlussel:
> Maybe I can use the `nativeExecution` flag to conditionally hide session properties. If it's enabled, hide all `java-worker` only properties, if it's not hide all `native-worker` only properties?

great idea!

pdabre12:
@rschlussel As discussed , addressed the suggestions and opened two issues, https://github.com/prestodb/presto/issues/23827 and https://github.com/prestodb/presto/issues/23828.
Can you please review again? Thanks !

tdcmeehan (commented):
`Add sidecars to NodeManager and add PluginNodeManager` ✅

presto-native-sidecar-plugin/src/test/java/com/facebook/presto/sidecar/TestNativeSidecarPluginSystemPropertyProvider.java
```diff
@@ -0,0 +1,90 @@
...
+    {
+        @Language("SQL") String sql = "SHOW SESSION";
+        MaterializedResult actualResult = computeActual(sql);
+        List<MaterializedRow> actualRows = actualResult.getMaterializedRows();
+        List<MaterializedRow> filteredRows = excludeSystemSessionProperties(actualRows);
+        assertFalse(filteredRows.isEmpty());
+    }
+
+    @Test
+    public void testJavaWorkerProperties()
```
rschlussel, line 73: why does this fail.  I thought the Java properties are registered even when using native worker.  Can you add a test (maybe i missed it?) of reading java properties on a native worker setup and native properties in a java worker set up?

presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
```diff
@@ -347,7 +335,7 @@ public final class SystemSessionProperties
     public static final String NATIVE_ROW_NUMBER_SPILL_ENABLED = "native_row_number_spill_enabled";
     public static final String NATIVE_TOPN_ROW_NUMBER_SPILL_ENABLED = "native_topn_row_number_spill_enabled";
     public static final String NATIVE_SPILLER_NUM_PARTITION_BITS = "native_spiller_num_partition_bits";
-    private static final String NATIVE_EXECUTION_ENABLED = "native_execution_enabled";
+    public static final String NATIVE_EXECUTION_ENABLED = "native_execution_enabled";
```
tdcmeehan, line 331: Why is this public?

presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
```diff
@@ -1774,6 +1696,21 @@ public SystemSessionProperties(
                         "The max partial aggregation memory when data reduction is optimal.",
                         1L << 26,
                         false),
+                stringProperty(
+                        NATIVE_EXECUTION_PROGRAM_ARGUMENTS,
```
tdcmeehan, line 1555: Is this is a new session property?

presto-main/src/main/java/com/facebook/presto/metadata/SessionPropertyManager.java
```diff
@@ -124,13 +202,27 @@ public Optional<PropertyMetadata<?>> getConnectorSessionPropertyMetadata(Connect
...
 
+    private Map<String, PropertyMetadata<?>> getWorkerSessionProperties()
+    {
+        List<PropertyMetadata<?>> workerSessionPropertiesList = workerSessionPropertyProviders.values().stream()
+                .flatMap(manager -> manager.getSessionProperties().stream())
+                .collect(toImmutableList());
+        Map<String, PropertyMetadata<?>> workerSessionProperties = new ConcurrentHashMap<>();
+        workerSessionPropertiesList.forEach(sessionProperty -> {
+            requireNonNull(sessionProperty, "sessionProperty is null");
+            // TODO: Implement fail fast in case of duplicate entries.
```
tdcmeehan, line 223: Can we just validate now?

presto-main/src/main/java/com/facebook/presto/sessionpropertyproviders/NativeWorkerSessionPropertyProvider.java
```diff
@@ -0,0 +1,216 @@
...
+                        3,
+                        !nativeExecution),
+                booleanProperty(
+                        NATIVE_DEBUG_VALIDATE_OUTPUT_FROM_OPERATORS,
+                        "If set to true, then during execution of tasks, the output vectors of " +
+                                "every operator are validated for consistency. This is an expensive check " +
+                                "so should only be used for debugging. It can help debug issues where " +
+                                "malformed vector cause failures or crashes by helping identify which " +
+                                "operator is generating them.",
+                        false,
```
rschlussel, line 139: the native_debug... fields should be hidden always (that was their default before)

presto-main/src/main/java/com/facebook/presto/metadata/SessionPropertyManager.java
```diff
@@ -47,35 +58,100 @@
...
+import static com.google.common.collect.ImmutableList.toImmutableList;
 import static java.lang.String.format;
 import static java.util.Objects.requireNonNull;
+import static java.util.concurrent.TimeUnit.HOURS;
 
 public final class SessionPropertyManager
 {
     private static final JsonCodecFactory JSON_CODEC_FACTORY = new JsonCodecFactory();
-    private final ConcurrentMap<String, PropertyMetadata<?>> systemSessionProperties = new ConcurrentHashMap<>();
+    private final ConcurrentMap<String, PropertyMetadata<?>> defaultOnSessionPropertyProvider = new ConcurrentHashMap<>();
```
tdcmeehan, line 77: This naming is confusing--it is not a session property provider, it is a map of session properties?

tdcmeehan (commented):
`Separate out java worker properties and introduce JavaFeaturesConfig` ❓

presto-main/src/main/java/com/facebook/presto/server/ServerMainModule.java
```diff
@@ -793,6 +801,18 @@ public ListeningExecutorService createResourceManagerExecutor(ResourceManagerCon
         binder.bind(NodeStatusNotificationManager.class).in(Scopes.SINGLETON);
 
         binder.bind(PlanCheckerProviderManager.class).in(Scopes.SINGLETON);
+
+        // Worker session property providers
+        MapBinder<String, WorkerSessionPropertyProvider> mapBinder =
+                newMapBinder(binder, String.class, WorkerSessionPropertyProvider.class);
+        mapBinder.addBinding("java-worker").to(JavaWorkerSessionPropertyProvider.class).in(Scopes.SINGLETON);
```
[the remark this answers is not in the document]
rschlussel: We need these even with sidecar enabled, because in our environment queries can bounce between java and native clusters, so users may specify properties for both.  We want to hide those properties, but we don't want the queries to fail.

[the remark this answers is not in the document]
tdcmeehan: How about we make the default false (disable the properties), but add a config flag that turns it on if needed?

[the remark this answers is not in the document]
pdabre12: Yes, that is the plan. 
Opened an issue for the same: https://github.com/prestodb/presto/issues/23827
Can I add this support in a follow-up PR?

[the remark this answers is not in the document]
tdcmeehan: SGTM!

presto-main/src/main/java/com/facebook/presto/metadata/SessionPropertyManager.java
pdabre12: I was planning on validating the session properties using the prefix, and the support for prefix was supposed to be added in a follow-up PR (tracked here: https://github.com/prestodb/presto/issues/23828), so I can address this then.

rschlussel:
@pdabre12 If it will take more time to address the comments about adding new tests can you pull out [Add sidecars to NodeManager and add PluginNodeManager](https://github.com/prestodb/presto/pull/23045/commits/421612803b9e57c62f4f962af00539f901967258) into a separate PR so that we can merge it quickly and use it in https://github.com/prestodb/presto/pull/23596. That will remove the dependency between these 2 in progress PRs.

pdabre12:
@rschlussel Opened a PR here: https://github.com/prestodb/presto/pull/23863

presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
pdabre12: Removed this property.

presto-native-sidecar-plugin/src/test/java/com/facebook/presto/sidecar/TestNativeSidecarPluginSystemPropertyProvider.java
pdabre12: Added the test cases :
For native properties in a java worker: https://github.com/prestodb/presto/blob/4eb5cf4a6017ca11ee531164114f1f20f35a2422/presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java#L

For java properties in a native worker setup:
https://github.com/prestodb/presto/blob/4eb5cf4a6017ca11ee531164114f1f20f35a2422/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeGeneralQueries.java#L1178

pdabre12:
@tdcmeehan @rschlussel Addressed the comments, please take a look. Also added a CI job for running native-sidecar tests.

tdcmeehan (approved):

rschlussel (approved):

tdcmeehan (approved):

jaystarshot:
@pdabre12  @tdcmeehan  Does this change warrant a release note?

tdcmeehan:
@jaystarshot it's in the above PR, but maybe the tool had trouble parsing it.

jaystarshot:
got it, I don't think **Coordinator Plugin Changes** is a topic yet. Is this change for some sidecar or for general presto c++ coordinator?

tdcmeehan:
Let's put it under general C++ changes

PR #23047, opened 2024-06-21
Treat 0 Estimations as NaN

abhinavmuk04:
## Description
WIP

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* ... :pr:`12345`
* ... :pr:`12345`

Hive Connector Changes
* ... :pr:`12345`
* ... :pr:`12345`
```

```
== NO RELEASE NOTE ==
```


presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/DetermineJoinDistributionType.java
```diff
@@ -134,7 +135,10 @@ private PlanNode getCostBasedJoin(JoinNode joinNode, Context context)
             }
         }
 
-        if (possibleJoinNodes.stream().anyMatch(result -> result.getCost().hasUnknownComponents()) || possibleJoinNodes.isEmpty()) {
+        PlanNodeStatsEstimate buildSideStats = context.getStatsProvider().getStats(joinNode.getRight());
+        boolean isBuildSideStatsUnknown = buildSideStats.getOutputRowCount() == 0 && buildSideStats.confidenceLevel() == LOW;
```
ClarenceThreepwood, line 139: What does rowcount == 0 add here? Wouldn't it make more sense to just check for low confidence level for the build side?

abhinavmuk04: This is specifically to handle the case where multiple filters are being pushed in and it results in an estimate of 0. Trying to handle the misleading case where a 0 estimate does not correlate to it be truly 0

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/DetermineJoinDistributionType.java
```diff
@@ -134,7 +135,10 @@ private PlanNode getCostBasedJoin(JoinNode joinNode, Context context)
...
+
+        if (isBuildSideStatsUnknown || possibleJoinNodes.stream().anyMatch(result -> result.getCost().hasUnknownComponents()) || possibleJoinNodes.isEmpty()) {
```
feilong-liu, line 141: This will not work. It will enter the logic where there are unknown cost for the possible joins, but actually it is not the case here when we have valid estimations for both sides of the join with one side to be low confident 0.

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/DetermineJoinDistributionType.java
```diff
@@ -134,7 +135,10 @@ private PlanNode getCostBasedJoin(JoinNode joinNode, Context context)
...
             // TODO: currently this session parameter is added so as to roll out the plan change gradually, after proved to be a better choice, make it default and get rid of the session parameter here.
             if (isUseBroadcastJoinWhenBuildSizeSmallProbeSizeUnknownEnabled(context.getSession()) && possibleJoinNodes.stream().anyMatch(result -> ((JoinNode) result.getPlanNode()).getDistributionType().get().equals(REPLICATED))) {
                 JoinNode broadcastJoin = (JoinNode) getOnlyElement(possibleJoinNodes.stream().filter(result -> ((JoinNode) result.getPlanNode()).getDistributionType().get().equals(REPLICATED)).map(x -> x.getPlanNode()).collect(toImmutableList()));
```
feilong-liu, line 144: At least this getOnlyElement will fail

feilong-liu (commented):
Also we need to have a session property to guard this feature.

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/DetermineJoinDistributionType.java
feilong-liu: Perhaps make getJoinNodeWithCost to return unknown cost is a better solution

steveburnett:
I think a word ("joins") is missing from the release note entry. Suggested edit:

```
* Add treatment of low confidence, zero estimations as ``UNKNOWN`` during joins, with the ``treat-low-confidence-zero-estimation-as-unknown`` session property :pr:`23047`
```

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/DetermineJoinDistributionType.java
```diff
@@ -114,6 +117,11 @@ public static boolean isBelowMaxBroadcastSize(JoinNode joinNode, Context context
 
         PlanNode buildSide = joinNode.getRight();
         PlanNodeStatsEstimate buildSideStatsEstimate = context.getStatsProvider().getStats(buildSide);
+
+        if (treatLowConfidenceZeroEstimationAsUnknownEnabled(context.getSession()) && buildSideStatsEstimate.confidenceLevel() == LOW && buildSideStatsEstimate.getOutputRowCount() == 0) {
```
feilong-liu, line 121: I see this "buildSideStatsEstimate.confidenceLevel() == LOW && buildSideStatsEstimate.getOutputRowCount() == 0)" is being called three times in this file, maybe consider having it in a separate function.

feilong-liu (approved):

PR #23059, opened 2024-06-24
Add analyzer warning for common NaN pitfalls

rschlussel:
## Description
Add an analyzer warning for division that can produce nans and for comparison operations involving doubles.  The warning will be returned regardless of whether the query is actually affected by nans, as we are unable to know the actual input until runtime. This feature is gated by the configuration property `warn-on-possible-nans` and session property `warn_on_possible_nans`

## Motivation and Context
Division of real and double type is a common accidental source of nans.  Comparison operations are the most significant of the semantic changes for the new nan semantic. Warn for both of these cases to help users adjust to the new nan semantics.

## Impact
Warning for division or comparison operators involving double or real types when warn_on_possible_nans is true.

## Test Plan
new unit tests, also ran a few queries and saw the warnings.

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Add configuration property ``warn-on-possible-nans`` and session property ``warn_on_possible_nans`` to produce a warning on division operations or comparison operations involving double or real types. Division operations are common causes of accidental creation of nans, and the semantics of comparison operations involving nans changed considerably in the most recent Presto release. 
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
+++ b/presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
@@ -347,6 +347,7 @@ public final class SystemSessionProperties
     public static final String DEFAULT_VIEW_SECURITY_MODE = "default_view_security_mode";
     public static final String JOIN_PREFILTER_BUILD_SIDE = "join_prefilter_build_side";
     public static final String OPTIMIZER_USE_HISTOGRAMS = "optimizer_use_histograms";
+    public static final String WARN_ON_COMMON_NAN_PATTERNS = "warn_on_common_nan_patterns";
 
     private final List<PropertyMetadata<?>> sessionProperties;
 
@@ -1937,6 +1938,10 @@ public final class SystemSessionProperties
                 booleanProperty(OPTIMIZER_USE_HISTOGRAMS,
                         "whether or not to use histograms in the CBO",
                         featuresConfig.isUseHistograms(),
+                        false),
+                booleanProperty(WARN_ON_COMMON_NAN_PATTERNS,
+                        "Whether to give a warning for some common issues relating to NaNs",
+                        featuresConfig.getWarnOnCommonNanPatterns(),
                         false));
     }
 
@@ -3228,5 +3233,10 @@ public final class SystemSessionProperties
     public static boolean shouldOptimizerUseHistograms(Session session)
     {
         return session.getSystemProperty(OPTIMIZER_USE_HISTOGRAMS, Boolean.class);
+    }
+
+    public static boolean warnOnCommonNanPatterns(Session session)
+    {
+        return session.getSystemProperty(WARN_ON_COMMON_NAN_PATTERNS, Boolean.class);
     }
 }
--- a/presto-main/src/main/java/com/facebook/presto/sql/analyzer/ExpressionAnalyzer.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/analyzer/ExpressionAnalyzer.java
@@ -157,6 +157,7 @@ import static com.facebook.presto.spi.StandardWarningCode.SEMANTIC_WARNING;
 import static com.facebook.presto.sql.NodeUtils.getSortItemsFromOrderBy;
 import static com.facebook.presto.sql.analyzer.Analyzer.verifyNoAggregateWindowOrGroupingFunctions;
 import static com.facebook.presto.sql.analyzer.Analyzer.verifyNoExternalFunctions;
+import static com.facebook.presto.sql.analyzer.ExpressionTreeUtils.isConstant;
 import static com.facebook.presto.sql.analyzer.ExpressionTreeUtils.isNonNullConstant;
 import static com.facebook.presto.sql.analyzer.ExpressionTreeUtils.tryResolveEnumLiteralType;
 import static com.facebook.presto.sql.analyzer.FunctionArgumentCheckerForAccessControlUtils.getResolvedLambdaArguments;
@@ -601,7 +602,16 @@ public class ExpressionAnalyzer
         protected Type visitComparisonExpression(ComparisonExpression node, StackableAstVisitorContext<Context> context)
         {
             OperatorType operatorType = OperatorType.valueOf(node.getOperator().name());
-            return getOperator(context, node, operatorType, node.getLeft(), node.getRight());
+            Type outputType = getOperator(context, node, operatorType, node.getLeft(), node.getRight());
+            // this needs to be checked after the call to getOperator(), because that's where the argument types get analyzed
+            if (sqlFunctionProperties.shouldWarnOnCommonNanPatterns() &&
+                    (TypeUtils.isApproximateNumericType(getExpressionType(node.getLeft())) || TypeUtils.isApproximateNumericType(getExpressionType(node.getRight())))) {
+                warningCollector.add(new PrestoWarning(
+                        SEMANTIC_WARNING,
+                        "Comparison operations involving DOUBLE or REAL types may include NaNs in the input. " +
+                                "Consider filtering out NaN values from your comparison input using the is_nan() function."));
+            }
+            return outputType;
         }
 
         @Override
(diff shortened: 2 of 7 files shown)


amitkdutta (approved):
Looks good. Thanks @rschlussel

presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
```diff
@@ -1937,6 +1938,10 @@ public SystemSessionProperties(
                 booleanProperty(OPTIMIZER_USE_HISTOGRAMS,
                         "whether or not to use histograms in the CBO",
                         featuresConfig.isUseHistograms(),
+                        false),
+                booleanProperty(WARN_ON_POSSIBLE_NANS,
+                        "Whether to give a warning for some common issues relating to Nans",
```
elharo, line 1943: NaN

presto-main/src/main/java/com/facebook/presto/sql/analyzer/ExpressionAnalyzer.java
```diff
@@ -601,7 +601,16 @@ protected Type visitLogicalBinaryExpression(LogicalBinaryExpression node, Stacka
...
             OperatorType operatorType = OperatorType.valueOf(node.getOperator().name());
-            return getOperator(context, node, operatorType, node.getLeft(), node.getRight());
+            Type outputType = getOperator(context, node, operatorType, node.getLeft(), node.getRight());
+            // this needs to be checked after the call to getOperator(), because that's where the argument types get analyzed
+            if (sqlFunctionProperties.shouldWarnOnPossibleNans() &&
+                    (TypeUtils.isApproximateNumericType(getExpressionType(node.getLeft())) || TypeUtils.isApproximateNumericType(getExpressionType(node.getRight())))) {
+                warningCollector.add(new PrestoWarning(
+                        SEMANTIC_WARNING,
+                        "Semantics of comparisons involving NaNs changed in release 0.288. Comparisons involving DOUBLE/REAL types may produce" +
+                                " different results than previously if there are NaNs being compared."));
```
elharo, line 611: there are NaNs being compared --> NaNs are compared

presto-main/src/main/java/com/facebook/presto/sql/analyzer/ExpressionAnalyzer.java
```diff
@@ -601,7 +601,16 @@ protected Type visitLogicalBinaryExpression(LogicalBinaryExpression node, Stacka
...
         {
             OperatorType operatorType = OperatorType.valueOf(node.getOperator().name());
-            return getOperator(context, node, operatorType, node.getLeft(), node.getRight());
+            Type outputType = getOperator(context, node, operatorType, node.getLeft(), node.getRight());
+            // this needs to be checked after the call to getOperator(), because that's where the argument types get analyzed
+            if (sqlFunctionProperties.shouldWarnOnPossibleNans() &&
+                    (TypeUtils.isApproximateNumericType(getExpressionType(node.getLeft())) || TypeUtils.isApproximateNumericType(getExpressionType(node.getRight())))) {
+                warningCollector.add(new PrestoWarning(
+                        SEMANTIC_WARNING,
+                        "Semantics of comparisons involving NaNs changed in release 0.288. Comparisons involving DOUBLE/REAL types may produce" +
```
elharo, line 610: do we want to include a specific version here? Also, I think this depends on a flag?

elharo (commented):
This warning strikes me as potentially noisy. It effectively warns not about NaNs but about double division. Furthermore there is no SQL change the user can or should make to silence the warning.

rschlussel:
@elharo I was also worried about being noisy, but we don't have a way to warn for actual nans encountered during runtime. I figured we could have this enabled when we first roll out the nan changes to help users identify that their queries might be affected, and then we could turn off the warning.

rschlussel (commented):
Thanks @elharo for the comments. I was trying to balance
1. giving a warning that would help users identify that their queries may be affected by the new NaN semantics (because my observation was that most users didn't realize they were producing or operating on NaNs) and to help oncalls trying to validate the new release.
2. Making the message not too confusing because the warning doesn't indicate that you are affected, only that you are doing an operation that is a common issue
3. Not too noisy.

I opted to have the warning off by default, and we can enable it in our environment during the migration to the new semantics, so it will be noisy, but during the time that we prefer being noisy to avoid wrong results.  And hopefully the messages convey what I was hoping to without being confusing, but definitely would take suggestions if you think they can be improved.

presto-main/src/main/java/com/facebook/presto/sql/analyzer/ExpressionAnalyzer.java
```diff
@@ -732,7 +742,13 @@ protected Type visitArithmeticUnary(ArithmeticUnaryExpression node, StackableAst
         @Override
         protected Type visitArithmeticBinary(ArithmeticBinaryExpression node, StackableAstVisitorContext<Context> context)
         {
-            return getOperator(context, node, OperatorType.valueOf(node.getOperator().name()), node.getLeft(), node.getRight());
+            Type returnType = getOperator(context, node, OperatorType.valueOf(node.getOperator().name()), node.getLeft(), node.getRight());
+            if (sqlFunctionProperties.shouldWarnOnPossibleNans() && node.getOperator() == ArithmeticBinaryExpression.Operator.DIVIDE && TypeUtils.isApproximateNumericType(returnType)) {
```
elharo, line 746: This isn't the only way to get NaN. You can also get it from a sqrt or power function, mulitplying Inf by 0, and probably other ways I'm not thinking of.

elharo (commented):
What bothers me most is that this warns on all floating point division, as if that's the problem, but we really only care about certain uncommon floating point division. Most floating point division is not a problem.

presto-main/src/main/java/com/facebook/presto/sql/analyzer/ExpressionAnalyzer.java
rschlussel: yeah, there are definitely other ways of producing nans. Maybe I'll call this warn_on_common_nan_patterns?

rschlussel:
> What bothers me most is that this warns on all floating point division, as if that's the problem, but we really only care about certain uncommon floating point division. Most floating point division is not a problem.

I can remove the warning for division, but it was the most common source of nans in our verifier results, and I'd like to err on the side of noisier if it's going to help the migration.  i should modify it to only warn if both the numerator and denominator are not constant. what do you think?

feilong-liu (dismissed):

presto-main/src/main/java/com/facebook/presto/sql/analyzer/ExpressionAnalyzer.java
```diff
@@ -601,7 +602,17 @@ protected Type visitLogicalBinaryExpression(LogicalBinaryExpression node, Stacka
...
         {
             OperatorType operatorType = OperatorType.valueOf(node.getOperator().name());
-            return getOperator(context, node, operatorType, node.getLeft(), node.getRight());
+            Type outputType = getOperator(context, node, operatorType, node.getLeft(), node.getRight());
+            // this needs to be checked after the call to getOperator(), because that's where the argument types get analyzed
+            if (sqlFunctionProperties.shouldWarnOnCommonNanPatterns() &&
+                    (TypeUtils.isApproximateNumericType(getExpressionType(node.getLeft())) || TypeUtils.isApproximateNumericType(getExpressionType(node.getRight())))) {
+                warningCollector.add(new PrestoWarning(
+                        SEMANTIC_WARNING,
+                        "Semantics of comparisons involving NaNs are changing. Comparisons involving DOUBLE/REAL types may produce" +
```
tdcmeehan, line 611: Can this be reworded to simply indicate that expressions involving doubles may produce NaNs?  i.e. not in the context of an evolving process, but as a simple statement of fact?

rschlussel: How about something like: `Comparison operations involving DOUBLE or REAL types may include NaNs in the input. Consider filtering out NaN values from your comparison using the is_nan() function.`

rschlussel: I've updated to this wording

tdcmeehan: SGTM!

tdcmeehan (approved):

feilong-liu (approved):

PR #23068, opened 2024-06-24
[native]Add to config largest size class pages

xiaoxmeng:
```
== NO RELEASE NOTE ==
```


aditi-pandit:
@xiaoxmeng : Thanks for this PR.

We should document the config and add some details on its tuning expectations.

tanjialiang (approved):

xiaoxmeng:
> @xiaoxmeng : Thanks for this PR.
> 
> We should document the config and add some details on its tuning expectations. Or add details in Velox config https://facebookincubator.github.io/velox/configs.html
> 
> @tdcmeehan
@aditi-pandit this is not Velox config but an option in velox memory manager. This is config for Prestissimo. We might need to add a config documentation for Prestissimo.

Issue #23063, opened 2024-06-24
Writer scaling fails for Parquet with smaller files


ZacBlanco:
## Your Environment

* Presto version used: 0.287
* Storage: S3
* Data source and connector used: Hive and Iceberg
* Deployment (Cloud or On-prem): Cloud

## Expected Behavior

`INSERT` statements with large amounts of data being written into parquet format should properly scale writers when `scale-writers=true` is set no matter the size of the files or format.

## Current Behavior

We have observed in internal benchmarking that depending on the data source and writer options, that there are some cases where the Parquet writer does not properly scale out to make use of the entire cluster. This occurs specifically with the parquet writer is writing to partitioned tables.

We discovered this issue when comparing hardware metrics while running some `INSERTS` on different formats and datasets. i.e. the following two charts are from writing to the `tpcds.sf100.inventory` table where one table is partitioned and the other is not.

<details><summary>Non-partitioned Insert</summary>
<p>
 <img src="https://github.com/prestodb/presto/assets/3464508/793cf480-2b79-42dc-b92e-bb0bda840d9e"></<img>
</p>
</details>

<details><summary>Partitioned Insert</summary>
<p>
<img src="https://github.com/prestodb/presto/assets/3464508/be1919e9-1d5b-4962-a1a7-6062eb48f911"></img>
</p>
</details>  

The big difference between these two charts shows that in the partitioned insert, only one worker is utilized, indicated by the large amount of network RX compared to every other node in the cluster. This has been traced to being an issue with the way that `ScaledWriterScheduler` works in conjunction with the `ParquetWriter`.

The `ScaledWriterScheduler` increases the number of writers based on the condition that there are full tasks _and_ that the amount of data written per node meets a minimum threshold. What we see is that when using the Parquet writer that the amount of written data to the nodes is almost always 0. When it is greater than 0, the `fullTasks` threshold never seems to be met.

The reason this occurs is that the `ParquetWriter` aggressively caches data in memory before flushing it to disk. Only once data is flushed to disk are the written bytes counted. If you are working on a partitioned table where the file(s) in each partition are less than the writer's buffer then the bytes written isn't reported until the writer is closed. At this point when the writer is closed, a new tasks becomes available. Hence the writes never scale out to additional workers because every time a worker finished writing a file, only then are the bytes written being is reported.

## Possible Solution

I see a few solutions here:

1. Change the Parquet writer to report written bytes that are not flushed to disk yet
[...]

PR #23069, opened 2024-06-24
Use buffered bytes in Parquet written bytes count

ZacBlanco:
Bytes buffered in memory should be considered written for statistical bookkeeping. This fixes issues where writer scaling doesn't work when there are files smaller than the configured parquet row group size.

## Description

Adds buffered bytes in both the legacy and optimized parquet writers' count of written bytes.

## Motivation and Context

See #23063 

## Impact

No API impact

## Test Plan

Existing tests

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-hive/src/main/java/com/facebook/presto/hive/ParquetRecordWriterUtil.java
+++ b/presto-hive/src/main/java/com/facebook/presto/hive/ParquetRecordWriterUtil.java
@@ -23,6 +23,7 @@ import org.apache.hadoop.io.Text;
 import org.apache.hadoop.io.Writable;
 import org.apache.hadoop.mapred.JobConf;
 import org.apache.hadoop.mapred.Reporter;
+import org.apache.parquet.column.ColumnWriteStore;
 import org.apache.parquet.hadoop.ParquetFileWriter;
 import org.apache.parquet.hadoop.ParquetOutputFormat;
 import org.apache.parquet.hadoop.ParquetRecordWriter;
@@ -30,6 +31,7 @@ import org.apache.parquet.hadoop.ParquetRecordWriter;
 import java.io.IOException;
 import java.lang.reflect.Field;
 import java.util.Properties;
+import java.util.concurrent.atomic.AtomicReference;
 
 import static com.facebook.presto.hive.HiveSessionProperties.getParquetWriterBlockSize;
 import static com.facebook.presto.hive.HiveSessionProperties.getParquetWriterPageSize;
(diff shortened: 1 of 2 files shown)


presto-hive/src/main/java/com/facebook/presto/hive/ParquetRecordWriterUtil.java
```diff
@@ -39,12 +41,14 @@ public final class ParquetRecordWriterUtil
...
     private static final Field INTERNAL_WRITER_FIELD;
     private static final Field FILE_WRITER_FIELD;
+    private static final AtomicReference<Field> COLUMN_WRITE_STORE_FIELD;
 
     static {
         try {
             REAL_WRITER_FIELD = ParquetRecordWriterWrapper.class.getDeclaredField("realWriter");
             INTERNAL_WRITER_FIELD = ParquetRecordWriter.class.getDeclaredField("internalWriter");
             FILE_WRITER_FIELD = INTERNAL_WRITER_FIELD.getType().getDeclaredField("parquetFileWriter");
+            COLUMN_WRITE_STORE_FIELD = new AtomicReference<>(null);
```
ZacBlanco, line 51: The `columnStore` field exists in the `InternalParquetRecordWriter` class which is declared as package-private. This means that we don't have the means to even access the class directly in the codebase. However, we can get the `Class<T>` instance from an existing reference. However, that can't be done until we initialize at least one writer.

Use an atomic reference and do a CAS when creating Parquet writers to initialize the `Field` instance if it hasn't been already

presto-hive/src/main/java/com/facebook/presto/hive/ParquetRecordWriterUtil.java
```diff
@@ -85,7 +94,7 @@ public void write(Writable value)
                     throws IOException
             {
                 recordWriter.write(value);
-                length = fileWriter.getPos();
+                length = fileWriter.getPos() + columnWriteStore.getBufferedSize();
```
yingsu00, line 97: I think it's better not to change the meaning of length, instead, the addition should be done in getWrittenBytes()

presto-parquet/src/main/java/com/facebook/presto/parquet/writer/ParquetWriter.java
```diff
@@ -121,7 +121,7 @@ public ParquetWriter(OutputStream outputStream,
 
     public long getWrittenBytes()
     {
-        return outputStream.longSize();
+        return outputStream.longSize() + getBufferedBytes();
```
yingsu00, line 124: The change should happen in ParquetFileWriter, not here. If you add the buffered bytes here, the getWrittenBytes() would return the sum of written bytes and buffered bytes and won't match the function name anymore. Check how OrcFileWriter does it:
```
 @Override
    public long getWrittenBytes()
    {
        return orcWriter.getWrittenBytes() + orcWriter.getBufferedBytes();
    }
```

yingsu00 (dismissed):
The logical changes look good to me, but I'm not sure about the change to get columnStore. @tdcmeehan Will you be able to review that part? Thanks!

presto-hive/src/main/java/com/facebook/presto/hive/ParquetRecordWriterUtil.java
tdcmeehan: I think this is fine, but it definitely deserves a comment that basically says this so people aren't scratching their heads years down the road why the code is written in this way.

tdcmeehan (changes requested):

presto-hive/src/main/java/com/facebook/presto/hive/ParquetRecordWriterUtil.java
ZacBlanco: added

tdcmeehan (approved):

Issue #23074, opened 2024-06-25
Inline cosntant cross joins


kaikalur:
select map(...)[x], x from  T

is a lot more efficient than

select m[x], x from T cross join (select map(...) m)

even when m is a literal map. This inlining can be huge - like 10x. cpu time and actually almost 100x our new bcu metric

CC: @feilong-liu

PR #23081, opened 2024-06-26
Remove cross join if one side of input is single row constant

feilong-liu:
## Description
Fix https://github.com/prestodb/presto/issues/23074

When one side of cross join inputs is single row with constant values, we can remove this cross join and replace it with a project node. This will simplify the plan, and enable more following optimizations, such as filter pushdown etc.

## Motivation and Context
We observed savings enabled by this optimization in our workload.

## Impact
We observed savings enabled by this optimization in our workload.

## Test Plan
Unit tests

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Add an optimization to remove cross join when one side of inputs is single row of constant values. The optimization is controlled by session property `remove_cross_join_with_constant_single_row_input` :pr:`23081`
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/RemoveCrossJoinWithConstantInput.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/RemoveCrossJoinWithConstantInput.java
@@ -1,0 +1,151 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.sql.planner.iterative.rule;
+
+import com.facebook.presto.Session;
+import com.facebook.presto.matching.Captures;
+import com.facebook.presto.matching.Pattern;
+import com.facebook.presto.metadata.FunctionAndTypeManager;
+import com.facebook.presto.spi.plan.FilterNode;
+import com.facebook.presto.spi.plan.JoinType;
+import com.facebook.presto.spi.plan.PlanNode;
+import com.facebook.presto.spi.plan.ProjectNode;
+import com.facebook.presto.spi.plan.ValuesNode;
+import com.facebook.presto.spi.relation.RowExpression;
+import com.facebook.presto.spi.relation.VariableReferenceExpression;
+import com.facebook.presto.sql.planner.iterative.Rule;
+import com.facebook.presto.sql.planner.plan.JoinNode;
+import com.facebook.presto.sql.relational.RowExpressionDeterminismEvaluator;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.function.Function;
+import java.util.stream.IntStream;
+
+import static com.facebook.presto.SystemSessionProperties.isRemoveCrossJoinWithConstantSingleRowInputEnabled;
+import static com.facebook.presto.sql.planner.PlannerUtils.addProjections;
+import static com.facebook.presto.sql.planner.plan.Patterns.join;
+import static com.google.common.base.Preconditions.checkState;
+import static com.google.common.collect.ImmutableMap.toImmutableMap;
+
+/**
+ * When one side of a cross join is one single row of constant, we can remove the cross join and replace it with a project.
+ * <pre>
+ * - Cross Join
+ *      - table scan
+ *          left_field
+ *      - values // only one row
+ *          right_field := 1
+ * </pre>
+ * into
+ * <pre>
+ * - project
+ *      left_field := left_field
+ *      right_field := 1
+ *      - table scan
+ *          left_field
+ * </pre>
+ */
+public class RemoveCrossJoinWithConstantInput
+        implements Rule<JoinNode>
+{
+    private final RowExpressionDeterminismEvaluator rowExpressionDeterminismEvaluator;
+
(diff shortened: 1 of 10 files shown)


kaikalur (changes requested):
Please add a couple of tests with maps and arrays literals where we see the most impact

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/RemoveCrossJoinWithConstantInput.java
```diff
@@ -0,0 +1,125 @@
...
+ *          leftkey
+ * </pre>
+ */
+public class RemoveCrossJoinWithConstantInput
+        implements Rule<JoinNode>
+{
+    @Override
+    public Pattern<JoinNode> getPattern()
+    {
+        return join().matching(x -> x.getType().equals(JoinType.INNER) && x.getCriteria().isEmpty());
```
kaikalur, line 64: What about filte in the join node? WE should push it down as well?

kaikalur (commented):
Add tests with cross join a constant with filter after the join

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/RemoveCrossJoinWithConstantInput.java
feilong-liu: Good point. I remember that the filter field in join node is only populated for outer join node, but not for inner join node. But still this case needs to be handled in this rule.
I updated the code to create a filter node, and this filter node will be pushed down with push filter down rules later.

feilong-liu:
> Please add a couple of tests with maps and arrays literals where we see the most impact

Added these tests

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/RemoveCrossJoinWithConstantInput.java
```diff
@@ -0,0 +1,145 @@
...
+            return Result.empty();
+        }
+        PlanNode resultNode = addProjections(joinInput, context.getIdAllocator(), mapping.get());
+        if (node.getFilter().isPresent()) {
+            resultNode = new FilterNode(node.getSourceLocation(), context.getIdAllocator().getNextId(), resultNode, node.getFilter().get());
+        }
+        return Result.ofPlanNode(resultNode);
+    }
+
+    private boolean outputSingleConstantRow(PlanNode planNode, Context context)
```
kaikalur, line 110: naming - isOutputSingeConstantRow

kaikalur (approved):

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/RemoveCrossJoinWithConstantInput.java
ajaygeorge: @kaikalur can you please mark this comment as resolved if your original comment as addressed.

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/RemoveCrossJoinWithConstantInput.java
```diff
@@ -0,0 +1,145 @@
...
+        return isRemoveCrossJoinWithConstantSingleRowInputEnabled(session);
+    }
+
+    @Override
+    public Result apply(JoinNode node, Captures captures, Context context)
+    {
+        PlanNode singleValueInput;
+        PlanNode joinInput;
+        if (isOutputSingleConstantRow(context.getLookup().resolve(node.getRight()), context)) {
+            singleValueInput = context.getLookup().resolve(node.getRight());
```
ajaygeorge, line 89: is `context.getLookup().resolve(..)` method call cheap.? Asking since we first do the call in if condition and then again in the block.
Maybe a small nit is to move the `context.getLookup().resolve(..)` calls up since you need to call it for both left and right regardless of which if/else block you are in.

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/RemoveCrossJoinWithConstantInput.java
```diff
@@ -0,0 +1,145 @@
...
+                    .collect(toImmutableMap(idx -> valuesNode.getOutputVariables().get(idx), idx -> valuesNode.getRows().get(0).get(idx)));
+            mapping = mapping.entrySet().stream().collect(toImmutableMap(
+                    Map.Entry::getKey, entry -> entry.getValue() instanceof VariableReferenceExpression ? assignments.get(entry.getValue()) : entry.getValue()));
+        }
+        boolean allDeterministic = mapping.entrySet().stream().allMatch(entry -> rowExpressionDeterminismEvaluator.isDeterministic(entry.getValue()));
+        if (allDeterministic) {
+            return Optional.of(mapping);
+        }
+        return Optional.empty();
+    }
```
ajaygeorge, line 144: some nits :
1. some opportunities to avoid code duplication

these lines are repeated twice and can be refactored to a method if it makes sense.
`entry -> entry.getValue() instanceof VariableReferenceExpression ? assignments.get(entry.getValue()) : entry.getValue()`

`mapping.entrySet().stream().collect(toImmutableMap(
                    Map.Entry::getKey, entry -> entry.getValue() instanceof VariableReferenceExpression ? assignments.get(entry.getValue()) : entry.getValue()));`
                    
 2. Add a line break before `checkState` for readability.
 
 3. 
 since we are operating on values we don't have to reach out to entrySet and then get the value. Instead directly call values() .
 
 `boolean allDeterministic = mapping.entrySet().stream().allMatch(entry -> rowExpressionDeterminismEvaluator.isDeterministic(entry.getValue()));`
->
 `boolean allDeterministic = mapping.values().stream().allMatch(rowExpressionDeterminismEvaluator::isDeterministic);`

ajaygeorge (approved):
Looks good.
Added some nits. Feel free to skip if the suggestions doesn't make sense.

ajaygeorge (approved):
Looks good.

PR #23095, opened 2024-06-27
[Iceberg] Return predicate columns in table statistics

ZacBlanco:
## Description

Returns `TableStatistics` instances which include `VariableStatsEstimate`s
corresponding to the selected columns in the table scan and the columns
which are in the table layout predicates.

## Motivation and Context

Queries which filter on a column but don't use it in their output might have
the assignment pruned from `PruneUnreferencedAssignments`. When the
assignment is pruned from the TableScanNode, it is no longer passed in
the set of selected columns in the `TableScanStatsRule`. However, it's
important to include information on columns not in the assignments
(such as those applied only in a predicate) because those additional
columns might be used to compute a more accurate output cardinality.

Specifically we saw this case in TPC-DS Q82 where the estimates for
an Iceberg and Hive table scan which had the exact same statistics
came back with very different estimates:

##### Iceberg

```
- TableScan[PlanNodeId 0][TableHandle {connectorId='iceberg', connectorHandle='item$data@Optional[8250555819003230185]', layout='Optional[item$data@Optional[8250555819003230185]]'}, grouped = false] => [i_item_sk:integer, i_item_id:varchar, i_item_desc:varchar, i_current_price:decimal(7,2)]
            Estimates: {source: CostBasedSourceInfo, rows: 80,872 (10.83MB), cpu: 11,350,978.15, memory: 0.00, network: 0.00}
```

##### Hive

```
- TableScan[PlanNodeId 0][TableHandle {connectorId='hive', connectorHandle='HiveTableHandle{schemaName=tpcds_sf1000_parquet_v2, tableName=item, analyzePartitionValues=Optional.empty}', layout='Optional[tpcds_sf1000_parquet_v2.item{domains={i_current_price=[ [["69.00", "99.00"]] ], i_manufact_id=[ [["105"], ["137"], ["180"], ["513"]] ], i_item_sk=[ [(<min>, <max>)] ]}}]'}, grouped = false] => [i_item_sk:integer, i_item_id:varchar(16), i_item_desc:varchar(200), i_current_price:decimal(7,2)]
            Estimates: {source: CostBasedSourceInfo, rows: 359 (49.14kB), cpu: 50,320.82, memory: 0.00, network: 0.00}
```

The approach in this PR mirrors the approach in the Hive connector 

https://github.com/prestodb/presto/blob/dcb4ed50bc228b268b021e6a43473bcdeaf2fde6/presto-hive/src/main/java/com/facebook/presto/hive/HiveMetadata.java#L866-L878

## Impact

Some additional information is included in the returned TableStatistics. Should
help improve join orders.

## Test Plan

- New unit test to verify behavior of method which calculates the set of columns to include in the statistics calculations
- Integration test to ensure that the table scan node generated by a query with a set of selected and non-selected columns with filter pushdown enabled and disabled on partitioned and non-partitioned tables should result in the same set of columns generated by the Iceberg metadata's `getTableStatistics` when executing the `TableScanStatsRule`

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergHiveMetadata.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergHiveMetadata.java
@@ -87,7 +87,6 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.TimeZone;
-import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import static com.facebook.presto.hive.HiveStatisticsUtil.createPartitionStatistics;
@@ -139,6 +138,7 @@ import static com.google.common.collect.ImmutableSet.toImmutableSet;
 import static java.util.Collections.emptyList;
 import static java.util.Locale.ENGLISH;
 import static java.util.Objects.requireNonNull;
+import static java.util.function.Function.identity;
 import static org.apache.iceberg.TableMetadata.newTableMetadata;
 import static org.apache.iceberg.TableProperties.OBJECT_STORE_PATH;
 import static org.apache.iceberg.TableProperties.WRITE_DATA_LOCATION;
@@ -446,16 +446,12 @@ public class IcebergHiveMetadata
         return ImmutableList.of(schemaNameOrNull);
     }
 
+    @Override
     public TableStatistics getTableStatistics(ConnectorSession session, ConnectorTableHandle tableHandle, Optional<ConnectorTableLayoutHandle> tableLayoutHandle, List<ColumnHandle> columnHandles, Constraint<ColumnHandle> constraint)
     {
         IcebergTableHandle handle = (IcebergTableHandle) tableHandle;
         org.apache.iceberg.Table icebergTable = getIcebergTable(session, handle.getSchemaTableName());
-        TableStatistics icebergStatistics = TableStatisticsMaker.getTableStatistics(session, typeManager,
-                tableLayoutHandle
-                        .map(IcebergTableLayoutHandle.class::cast)
-                        .map(IcebergTableLayoutHandle::getValidPredicate),
-                constraint, handle, icebergTable,
-                columnHandles.stream().map(IcebergColumnHandle.class::cast).collect(Collectors.toList()));
+        TableStatistics icebergStatistics = super.getTableStatistics(session, tableHandle, tableLayoutHandle, columnHandles, constraint);
         EnumSet<ColumnStatisticType> mergeFlags = getHiveStatisticsMergeStrategy(session);
         return tableLayoutHandle.map(IcebergTableLayoutHandle.class::cast).map(layoutHandle -> {
             TupleDomain<VariableReferenceExpression> predicate = layoutHandle.getValidPredicate()
@@ -470,7 +466,8 @@ public class IcebergHiveMetadata
                     .orElse(icebergStatistics);
             TableStatistics.Builder filteredStatsBuilder = TableStatistics.builder()
                     .setRowCount(mergedStatistics.getRowCount());
-            for (ColumnHandle colHandle : columnHandles) {
+            Set<ColumnHandle> fullColumnSet = icebergStatistics.getColumnStatistics().keySet();
+            for (ColumnHandle colHandle : fullColumnSet) {
                 IcebergColumnHandle icebergHandle = (IcebergColumnHandle) colHandle;
                 if (mergedStatistics.getColumnStatistics().containsKey(icebergHandle)) {
                     ColumnStatistics stats = mergedStatistics.getColumnStatistics().get(icebergHandle);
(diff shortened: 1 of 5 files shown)


ZacBlanco:
FYI @hantangwangd

hantangwangd:
Thanks for the change, it's a great improvement.

One problem to be considered, the `predicateColumns` in `IcebergTableLayoutHandle` may be different between `pushdown_filter_enabled` being true and false. As far as I can see, when `pushdown_filter_enabled` is true, the partition columns would not exist in `predicateColumns` after filter pushing down, that is, we may miss the partition columns after `combineSelectedAndPredicateColumns` under current implementation logic.

One feasible solution may be adding a new method similar to `getValidPredicate()` in `IcebergTableLayoutHandle` to handle the gap between the two conditions.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/hive/TestIcebergHiveStatistics.java
```diff
@@ -332,6 +340,50 @@ public void testHiveStatisticsMergeFlags()
...
+        Session session = Session.builder(getSession())
+                .setCatalogSessionProperty("iceberg", PUSHDOWN_FILTER_ENABLED, Boolean.toString(pushdownFilterEnabled))
+                .build();
+        try {
+            TableStatistics est = getScanStatsEstimate(session, "SELECT k from scanFilterStats WHERE j > 2 AND i = 3");
+            assertEquals(est.getColumnStatistics().size(), 3);
+        }
+        finally {
+            getQueryRunner().execute("DROP TABLE scanFilterStats");
+        }
```
hantangwangd, line 368: Seems that the session property `iceberg.pushdown_filter_enabled` not working. I believe it is missed in `AbstractTestingPrestoClient.toClientSession(...)`.

And furthermore, if `iceberg.pushdown_filter_enabled` works, here we would only get an exception, as pushdown_filter_enabled currently do not support java worker. So maybe we should figure out another way to test such scenarios.

ZacBlanco: > I believe it is missed in AbstractTestingPrestoClient.toClientSession(...).

I'm not sure what you mean here. Could you elaborate?

> And furthermore, if iceberg.pushdown_filter_enabled works, here we would only get an exception when it's true

`pushdown_filter_enabled` only throws an error once the splits are generated. In this test, we're not running the query. Just generating the query plan, so we shouldn't see an error thrown here. It should work fine in pure Java

hantangwangd: > I'm not sure what you mean here. Could you elaborate?

We can make a breakpoint in method `IcebergFilterPushdown.optimize(...)` to observe the validity of session property `iceberg.pushdown_filter_enabled`. Seems it was missed after `session.beginTransactionId(...)` and `AbstractTestingPrestoClient.toClientSession(...)` in method `TestIcebergHiveStatistics.getScanStatsEstimate(...)`.

> In this test, we're not running the query. Just generating the query plan, so we shouldn't see an error thrown. It should work fine in pure Java

It seems that `getQueryRunner().createPlan(txnSession, sql, WarningCollector.NOOP)` in `TestIcebergHiveStatistics.getScanStatsEstimate(...)` will firstly execute the sql, then fetch its corresponding plan. So it would throw an error in java worker when `pushdown_filter_enabled` is true.

ZacBlanco: You were correct about the execution. I think it was an issue with the way I generated the logical plan. The session property was not being propagated properly to the connector (it was always false). I dug up some more examples and found a better way to get the logical plan which doesn't involve executing the query. I also confirmed that the pushdown property is set correctly. Thanks for pointing out the issue

hantangwangd (dismissed):
Thanks for the fix. LGTM!

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergHiveMetadata.java
```diff
@@ -467,7 +462,7 @@ public TableStatistics getTableStatistics(ConnectorSession session, ConnectorTab
                     .orElse(icebergStatistics);
             TableStatistics.Builder filteredStatsBuilder = TableStatistics.builder()
                     .setRowCount(mergedStatistics.getRowCount());
-            for (ColumnHandle colHandle : columnHandles) {
+            for (ColumnHandle colHandle : icebergStatistics.getColumnStatistics().keySet()) {
```
imjalpreet, line 465: nit: Since there are multiple usages of `icebergStatistics.getColumnStatistics().keySet()` we can probably assign it to a variable.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestStatisticsUtil.java
```diff
@@ -253,6 +269,61 @@ public void testCalculateTableSize()
...
+        includedColumns = combineSelectedAndPredicateColumns(ImmutableList.of(k, l), Optional.of(builder.build()));
+        assertEquals(new HashSet<>(includedColumns), ImmutableSet.of(i, j, k, l));
+    }
+
+    private static IcebergColumnHandle generateIcebergColumnHandle(String name, boolean partitioned)
+    {
+        return new IcebergColumnHandle(new ColumnIdentity(ThreadLocalRandom.current().nextInt(), name, TypeCategory.PRIMITIVE, ImmutableList.of()),
+                INTEGER,
+                Optional.empty(),
+                partitioned ? BaseHiveColumnHandle.ColumnType.PARTITION_KEY : REGULAR,
```
imjalpreet, line 323: nit: Please add static import for PARTITION_KEY

imjalpreet (commented):
LGTM, minor nits.

agrawalreetika:
Thanks for the fix @ZacBlanco
Looks like when we had filter pushdown enabled for Icberg runs with Prestissimo it was messing up Join order multiple tpc-ds query. I think this change will help in removing plan differences on some tpc-ds queries compared to Hive connector.

agrawalreetika (approved):
Thanks for the fix @ZacBlanco
Looks like when we had filter pushdown enabled for Icberg runs with Prestissimo it was messing up Join order multiple tpc-ds query. I think this change will help in removing plan differences on some tpc-ds queries compared to Hive connector.

hantangwangd (approved):

Issue #23062, opened 2024-06-24
[native] Flaky test TaskManagerTest.buildSpillDirectoryFailure


rschlussel:
example failure: https://app.circleci.com/pipelines/github/prestodb/presto/17128/workflows/75e87d3d-61cf-4f44-bc5d-ed355fdb0884/jobs/67544
```
9: E20240624 16:06:09.054556  6425 Exceptions.h:67] Line: /root/project/presto-native-execution/velox/velox/exec/Task.cpp:1869, Function:terminate, Expression:  Aborted for external error, Source: RUNTIME, ErrorCode: INVALID_STATE
9: E20240624 16:06:09.054900  6425 TaskManager.cpp:282] There are 1 zombie Task that satisfy cleanup conditions but could not be cleaned up, because the Task are referenced by more than 1 owners. RUNNING[0] FINISHED[0] CANCELED[0] ABORTED[1] FAILED[0]  Sample task IDs (shows only 20 IDs): 
9: E20240624 16:06:09.054929  6425 TaskManager.cpp:292] Zombie Task [1/1]: Extra Refs: 1, test.0.0.0.0
9: E20240624 16:06:12.241359  6425 Exceptions.h:67] Line: /root/project/presto-native-execution/velox/velox/exec/tests/utils/QueryAssertions.cpp:1458, Function:waitForAllTasksToBeDeleted, Expression: numDeletedTasks == numCreatedTasks (87 vs. 88) 88 tasks have been created while only 87 have been deleted after waiting for 3000000 us, Source: RUNTIME, ErrorCode: INVALID_STATE
9: unknown file: Failure
9: C++ exception with description "Exception: VeloxRuntimeError
9: Error Source: RUNTIME
9: Error Code: INVALID_STATE
9: Reason: (87 vs. 88) 88 tasks have been created while only 87 have been deleted after waiting for 3000000 us
9: Retriable: False
9: Expression: numDeletedTasks == numCreatedTasks
9: Function: waitForAllTasksToBeDeleted
9: File: /root/project/presto-native-execution/velox/velox/exec/tests/utils/QueryAssertions.cpp
9: Line: 1458
9: Stack trace:
9: # 0  
9: # 1  
9: # 2  
9: # 3  
9: # 4  
9: # 5  
9: # 6  
9: # 7  
9: # 8  
9: # 9  
9: # 10 
9: # 11 
9: # 12 
9: # 13 
9: # 14 
9: # 15 
9: # 16 
9: # 17 
[...]

PR #23098, opened 2024-06-27
[native] Fix TaskManagerTest.buildSpillDirectoryFailure

spershin:
## Description
There is a race in the test itself, which we fix here:
- taskManager_->cleanOldTasks() can be called before the Driver is destroyed.
- Since the Driver holds the Task, the Task is considered a zombie inside cleanOldTasks() and is NOT removed from the map.
- So, destructor of the Task is not called.
- Then we call waitForAllTasksToBeDeleted() which simply compares how many ctors and dtors we had called. They don't match.
- Test fails.

Fixes https://github.com/prestodb/presto/issues/23062

## Test Plan
Stress testing.
`buck2 test @mode/opt github/presto-trunk/presto-native-execution/presto_cpp/main/tests:presto_main_test -- TaskManagerTest.buildSpillDirectoryFailure --stress-runs 100
`

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp
+++ b/presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp
@@ -54,6 +54,26 @@ namespace facebook::presto {
 
 namespace {
 
+// Repeatedly calls for cleanOldTasks() for a while to ensure that we overcome a
+// potential race condition where we call cleanOldTasks() before some Tasks are
+// ready to be cleaned.
+void waitForAllOldTasksToBeCleaned(
+    TaskManager* taskManager,
+    uint64_t maxWaitUs) {
+  taskManager->cleanOldTasks();
+
+  uint64_t waitUs = 0;
+  while (taskManager->getNumTasks() > 0) {
+    constexpr uint64_t kWaitInternalUs = 1'000;
+    std::this_thread::sleep_for(std::chrono::microseconds(kWaitInternalUs));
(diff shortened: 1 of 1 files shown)


tanjialiang:
Thanks for the fix. Does it make sense to just extend the test from velox's OperatorTestBase and reuse the rich utilities there?

spershin:
@tanjialiang 
This test fixes the use of TaskManager's cleanOldTasks(). TaskManager is not in Velox, so we cannot use rich utilities from Velox.

presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp
```diff
@@ -54,6 +54,26 @@ namespace facebook::presto {
 
 namespace {
 
+// Repeatedly calls for cleanOldTasks() for a while to ensure that we overcome a
```
elharo, line 57: for cleanOldTasks() -->= cleanOldTasks()

elharo (approved):

amitkdutta (approved):
Thanks for the fix @spershin

presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp
spershin: Thanks!
Let's fix this some other time.
Now it is important to land this to remove the noise.

PR #23099, opened 2024-06-28
[verifier] Add 'control_query_utf8' and 'test_query_utf8' columns.

spershin:
## Description
Add 'control_query_utf8' and 'test_query_utf8' columns to the verifier query table.

## Motivation and Context
We realized that existing 'control_query' and 'test_query' columns are of latin1 format, which
results is some utf8 characters loss.
Direct conversion of the table columns is failing due to the existing data, so we have come
up with a workaround via the additional temporary columns.
At some point we can remove either the new columns or the old ones and keep the new ones.

The new columns are used if not null, otherwise the old ones are used.

## Test Plan
Plan to make a custom build and test it on a altered table.

```
== NO RELEASE NOTE ==
```


steveburnett (approved):
LGTM! (docs)

PR #23123, opened 2024-07-02
Use property to deactivate UI profile

ZacBlanco:
## Description
In lieu of the activeByDefault, use a property instead to disable building the UI

## Motivation and Context

We have seen build failures occurring when users try to introduce new maven profiles to the POMs.

The UI currently uses the `activeByDefault` directive. It doesn't do what the user might expect. Other profiles which are activeByDefault can override this in the root POM but not submodule POMs, so the the build of the child  module might not be able to find the presto-ui dependency, causing a build failure.

This changes instead uses the `<property>` profile activation which will only deactivate this profile in the case a specific property (skipUI) is present. This can be passed to the maven arguments as simply `-DskipUI`, similar to the surefire plugin's `-DskipTests`

See the following for more discussion:
https://stackoverflow.com/questions/5309379/how-to-keep-maven-profiles-which-are-activebydefault-active-even-if-another-prof

## Impact

CI/scripts which build presto and skip the UI may need to update their command line arguments to use `-DskipUI` instead.

## Test Plan

Local build with various combinations of arguments including and not including `-DskipUI`

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,9 +91,10 @@ resource JAR is included on the classpath of Presto coordinator, it will be able
 resources.
 
 None of the Java code relies on the Presto UI project being compiled, so it is possible to exclude
-this UI when building Presto. It can be excluded by disabling the `ui` maven profile with `-P \!ui`:
+this UI when building Presto. Add the property `-DskipUI` to the maven command to disable building
+the `ui` maven module.
 
-    ./mvnw clean install -P \!ui
+    ./mvnw clean install -DskipUI
 
 You must have [Node.js](https://nodejs.org/en/download/) and [Yarn](https://yarnpkg.com/en/) installed to build the UI. When using  Maven to build
 the project, Node and yarn are installed in the `presto-ui/target` folder. Add the node and yarn
--- a/pom.xml
+++ b/pom.xml
@@ -2489,7 +2489,9 @@
         <profile>
             <id>ui</id>
             <activation>
-                <activeByDefault>true</activeByDefault>
+                <property>
+                    <name>!skipUI</name>
+                </property>
             </activation>
             <modules>
                 <module>presto-ui</module>
--- a/presto-main/pom.xml
+++ b/presto-main/pom.xml
@@ -576,11 +576,13 @@
     <profiles>
         <profile>
             <!-- Requires the UI as a dependency by default. If the UI is   -->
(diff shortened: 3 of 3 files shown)


steveburnett (approved):
LGTM! (docs)
Viewed README.md in GitHub, no concerns. Thanks!

yhwang (approved):
/LGTM

Thanks

AbhijitKulkarni1:
I'm also encountering the same problem with workflow executions regarding JDK21 support. Could you please review and update the Maven execution commands if needed, in the workflow files to ensure that this profile is skipped wherever not needed?

ZacBlanco:
In general the profile should be active. If you're seeing issues in your JDK 21 build you can apply this PR as a patch temporarily and rebase off master once merged. It should fix your issues

ZacBlanco:
@elharo @rschlussel I think we should make this change. I'm not sure how it affects your internal build systems. Could you take a look?

swapsmagic (approved):
Given the change does not disable building UI should not impact the meta builds as nothing is changing as long as the `skipUI` property is not being passed. So should be good to merge.

README.md
```diff
@@ -91,9 +91,10 @@ resource JAR is included on the classpath of Presto coordinator, it will be able
...
 
 None of the Java code relies on the Presto UI project being compiled, so it is possible to exclude
-this UI when building Presto. It can be excluded by disabling the `ui` maven profile with `-P \!ui`:
+this UI when building Presto. Add the property `-DskipUI` to the maven command to disable building
+the `ui` maven module.
 
-    ./mvnw clean install -P \!ui
+    ./mvnw clean install -DskipUI
 
 You must have [Node.js](https://nodejs.org/en/download/) and [Yarn](https://yarnpkg.com/en/) installed to build the UI. When using  Maven to build
```
elharo, line 99: Is this still true?

elharo (approved):

tdcmeehan (approved):

README.md
ZacBlanco: For development, I believe it's still easier to have them installed on your local system. Technically don't need them since we use the maven plugin now, but having to reference the executable from the `./target` is a little cumbersome. @yhwang may want to update this? It is adjacent to this change though.

Issue #22841, opened 2024-05-27
Security Vulnerability in guava


KarthikaPKumar:
## Package
com.google.guava

## Affected versions
25.1-android
26.0-jre
30.1.1-jre
31.0.1-jre

## Patched versions
32.0.0-android
32.0.0-jre

## Summary
Vulnerable versions of guava 25.1-android, 26.0-jre, 30.1.1-jre, 31.0.1-jre is still being used in the presto for several connectors as well.The CVE's detected are https://github.com/advisories/GHSA-5mg8-w23w-74h3, https://github.com/advisories/GHSA-7g45-4rm6-3mm3.

Description -:
[CVE-2020-8908](https://nvd.nist.gov/vuln/detail/cve-2020-8908) -:
A temp directory creation vulnerability exists in all versions of Guava, allowing an attacker with access to the machine to potentially access data in a temporary directory created by the Guava API com.google.common.io.Files.createTempDir(). By default, on unix-like systems, the created directory is world-readable (readable by an attacker with access to the system). The method in question has been marked deprecated in versions 30.0 and later and should not be used.

[CVE-2023-2976](https://nvd.nist.gov/vuln/detail/CVE-2023-2976)
Use of Java's default temporary directory for file creation in FileBackedOutputStream in Google Guava versions 1.0 to 31.1 on Unix systems and Android Ice Cream Sandwich allows other users and apps on the machine with access to the default Java temporary directory to be able to access the files created by the class.

PR #23127, opened 2024-07-03
Upgrade guava to 32.1.0-jre due CVE-2023-2976

denodo-research-labs:
## Description
Presto issue #22841 

## Motivation and Context
Solve CVE of severity HIGH.

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Upgrade guava to 32.1.0-jre due CVE-2023-2976 :pr:`23127`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -74,7 +74,10 @@
         <dep.docker-java.version>3.3.0</dep.docker-java.version>
         <dep.jayway.version>2.6.0</dep.jayway.version>
         <dep.ratis.version>2.2.0</dep.ratis.version>
+        <dep.errorprone.version>2.18.0</dep.errorprone.version>
+        <dep.guava.version>32.1.0-jre</dep.guava.version>
         <dep.jackson.version>2.11.0</dep.jackson.version>
+        <dep.j2objc.version>2.8</dep.j2objc.version>
         <!--
           America/Bahia_Banderas has:
            - offset change since 1970 (offset Jan 1970: -08:00, offset Jan 2018: -06:00)
@@ -2240,12 +2243,19 @@
                 <artifactId>datasketches-java</artifactId>
                 <version>5.0.1</version>
             </dependency>
-            
+
+            <dependency>
+                <groupId>com.google.errorprone</groupId>
+                <artifactId>error_prone_annotations</artifactId>
+                <version>${dep.errorprone.version}</version>
+            </dependency>
+
             <dependency>
                 <groupId>com.google.j2objc</groupId>
                 <artifactId>j2objc-annotations</artifactId>
-                <version>1.3</version>
+                <version>${dep.j2objc.version}</version>
             </dependency>
+
         </dependencies>
     </dependencyManagement>
 
--- a/presto-hive/src/main/java/com/facebook/presto/hive/HiveSplitSource.java
+++ b/presto-hive/src/main/java/com/facebook/presto/hive/HiveSplitSource.java
@@ -26,6 +26,7 @@ import com.facebook.presto.spi.PrestoException;
 import com.facebook.presto.spi.connector.ConnectorPartitionHandle;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
+import com.google.common.util.concurrent.FluentFuture;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.SettableFuture;
@@ -328,7 +329,7 @@ class HiveSplitSource
                     {
                         checkArgument(bucketNumber.isPresent(), "bucketNumber must be present");
                         if (!allSplitLoaded.isDone()) {
-                            return allSplitLoaded.transform(ignored -> ImmutableList.of(), executor);
+                            return FluentFuture.from(allSplitLoaded).transform(ignored -> ImmutableList.of(), executor);
                         }
                         return immediateFuture(function.apply(getSplits(bucketNumber.getAsInt(), maxSize)).getResult());
                     }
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: presto-hive/src/main/java/com/facebook/presto/hive/HiveSplitSource.java [hunk 1]
```diff
@@ -328,7 +329,7 @@ public synchronized ListenableFuture<List<ConnectorSplit>> borrowBatchAsync(Opti
                     {
                         checkArgument(bucketNumber.isPresent(), "bucketNumber must be present");
                         if (!allSplitLoaded.isDone()) {
-                            return allSplitLoaded.transform(ignored -> ImmutableList.of(), executor);
+                            return FluentFuture.from(allSplitLoaded).transform(ignored -> ImmutableList.of(), executor);
```
elharo, line 332: How is this change related?

presto-jdbc/pom.xml
```diff
@@ -80,8 +80,24 @@
             <artifactId>guava</artifactId>
             <exclusions>
                 <exclusion>
-                    <groupId>*</groupId>
-                    <artifactId>*</artifactId>
+                    <artifactId>checker-qual</artifactId>
```
elharo, line 83: Why do these need to be excluded?

Also, groupId conventionally comes before artifactId. This made me look twice.

pom.xml
```diff
@@ -74,7 +74,10 @@
         <dep.docker-java.version>3.3.0</dep.docker-java.version>
         <dep.jayway.version>2.6.0</dep.jayway.version>
         <dep.ratis.version>2.2.0</dep.ratis.version>
+        <dep.errorprone.version>2.18.0</dep.errorprone.version>
+        <dep.guava.version>32.1.0-jre</dep.guava.version>
```
elharo, line 78: Where was this set before? Was this pulled in via airlift? If so, we might want to update airlift instead.

pom.xml
```diff
@@ -74,7 +74,10 @@
...
         <dep.jackson.version>2.11.0</dep.jackson.version>
+        <dep.j2objc.version>2.8</dep.j2objc.version>
```
elharo, line 80: Do we need j2bjc? I see it in a few poms, but not in Java code anywhere. Can we instead remove it where it appears?

presto-hive/src/main/java/com/facebook/presto/hive/HiveSplitSource.java
denodo-research-labs: Fixes a compilation error due to this change introduced in Guava 27.0: [AbstractFuture doesn't expose FluentFuture APIs anymore](https://github.com/google/guava/commit/0f8d3608af401d9090642da2b88f9a8b5999ff1f)

pom.xml
denodo-research-labs: guava version is pulled via airbase, but we see no movement on the forked airbase regarding to version upgrades.

errorprone version is introduced in this PR.

presto-jdbc/pom.xml
denodo-research-labs: Done: groupid before artifactid.

Only failureaccess is needed, all others are excluded.

pom.xml
elharo: I personally don't know why we use airbase, but I assume there's a reason. Given that we do, I think the default approach is to fix that rather introducing the new guava version here. Otherwise it's still possible we'll pull in the wrong guava version somewhere and have dependency convergence issues.

denodo-research-labs: It seems that in the past airbase was used to pull versions, but it does not appear to be the approach being used now:
- the pom also overrides other versions declared in airbase such as slf4j, jackson, joda, etc... 
- airbase v102 is in use since May 2021

pom.xml
denodo-research-labs: Yes we need it. It is often used in the guava code, e.g. in `com.google.common.util.concurrent.AbstractFuture` class, and this class is used via `com.google.common.util.concurrent.SettableFuture` by `com.facebook.presto.hive.HiveSplitSource`

elharo: but shouldn't guava supply that then? Does it need to be in our poms? 

Also should we just use the guava BOM instead of setting these dependencies up manually?

denodo-research-labs: We are pinning the version to avoid conflicts among different versions of the same dependency (Maven Enforcer plugin).

tdcmeehan (approved):

presto-jdbc/pom.xml
```diff
@@ -80,8 +80,24 @@
             <artifactId>guava</artifactId>
             <exclusions>
                 <exclusion>
-                    <groupId>*</groupId>
-                    <artifactId>*</artifactId>
+                    <groupId>com.google.errorprone</groupId>
```
agrawaldevesh, line 83: I believe this is wrong and is causing shading issues in presto-jdbc: org/checkerframework is now creeping into presto-jdbc jar. The groupId should be com.google.errorprone for the checker-qual

Please fix asap.

agrawaldevesh (commented):
Please fix the highlighted issue asap since this is breaking projects that use presto-jdbc

presto-jdbc/pom.xml
agrawaldevesh: I am fixing this in https://github.com/prestodb/presto/pull/23304

hokitorres: Thanks, we assume there is no way to provide a test to avoid this issue, is there?

denodo-research-labs: Thanks, we assume there is no way to provide a test to avoid this issue, is there?

Issue #20847, opened 2023-09-13
[Native] Incorrect results for CAST from varchar(n) to varchar(m)


aditi-pandit:
There is a problem in casting between varchar type columns (with different lengths) in Prestissimo.

In the following SQL the input is a varchar[8] type column cast into varchar(3) and C++ worker returns incorrect results. This works correctly in Presto.

presto:tpch> select cast(varchar_value as varchar(3)) from (values ('9898.122')) t(varchar_value);
_col0
9898.122
Note:-> This is incorrect output. The correct result should be 989

System information

## Your Environment

* Presto version used: master branch
Velox System Info v0.0.2
Commit: fb5afdc532f3248b49302a4bdf0dbe406903882b
CMake Version: 3.26.4
System: Darwin-22.4.0
Arch: x86_64
C++ Compiler: /Library/Developer/CommandLineTools/usr/bin/c++
C++ Compiler Version: 14.0.3.14030022
C Compiler: /Library/Developer/CommandLineTools/usr/bin/cc
C Compiler Version: 14.0.3.14030022
CMake Prefix Path: /Library/Developer/CommandLineTools/SDKs/MacOSX13.1.sdk/usr;/usr/local;/usr;/;/usr/local/Cellar/cmake/3.26.4;/usr/local;/usr/X11R6;/usr/pkg;/opt;/sw;/opt/local

## Expected Behavior
Result of below query should be 989

## Current Behavior
Result of below query is 9898.12

## Steps to Reproduce
presto:tpch> select cast(varchar_value as varchar(3)) from (values ('9898.122')) t(varchar_value);
_col0
9898.122
Note:-> This is incorrect output. The correct result should be 989

aditi-pandit:
The underlying reason is that there isn't a concept of varchar(n) in Velox. In Velox VARCHAR type has a vector of values each of which has its own length. There isn't a notion of a cap on that length.

So in transforming this CAST it becomes VARCHAR -> VARCHAR which is essentially a no-op.

There are multiple solutions here:
i) CAST(VARCHAR(n), VARCHAR(m)) can be evaluated as substr when n > m and is a no-op when n < m. The Prestissimo expression conversion can transform this CAST -> substr
ii) Error for CAST expression instead of silently giving wrong results.
iii) Support varchar(n) notion in Velox and enhance expression compilation to address this case.

aditi-pandit:
[...]

PR #23132, opened 2024-07-03
[native] Convert cast to substr when target is capped varchar

### LINKED ISSUE #7264: Request for supporting SSE-KMS
Currently, Presto supports SSE ([hive.s3.sse.enabled](https://prestodb.io/docs/0.157/connector/hive.html)) w.r.t AES256
It would be nice if it supports SSE-KMS where a customer can specify the [kms-key-id](http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingKMSEncryption.html) for SSE.
Like, hive.s3.sse.kms-key-id

gggrace14:
Convert cast to the call to substr() if the target type is varchar with max length in plan in PrestoToVeloxExpression. Raise exception for try_cast, which will handle later when we align the try_cast behavior.

Resolve https://github.com/facebookincubator/velox/issues/7264

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/types/PrestoToVeloxExpr.cpp
+++ b/presto-native-execution/presto_cpp/main/types/PrestoToVeloxExpr.cpp
@@ -159,6 +159,40 @@ std::vector<TypedExprPtr> VeloxExprConverter::toVeloxExpr(
 }
 
 namespace {
+static const char* kVarchar = "varchar";
+
+/// Convert cast of varchar to substr if target type is varchar with max length.
+/// Throw an exception for cast of varchar to varchar with max length.
+std::optional<TypedExprPtr> convertCastToVarcharWithMaxLength(
+    const std::string& returnType,
+    const std::vector<TypedExprPtr>& args,
+    bool nullOnFailure) {
+  if (nullOnFailure) {
+    VELOX_NYI("TRY_CAST of varchar to {} is not supported.", returnType);
+  }
+
+  // Parse the max length from the return type string in the format of
+  // varchar(max_length). Assume return type string is valid given
+  // TypeParser.yy.
+  char* end;
+  const auto length =
+      strtol(returnType.data() + strlen(kVarchar) + 1, &end, 10);
+  VELOX_DCHECK(errno != ERANGE);
+  VELOX_DCHECK(end == returnType.data() + returnType.size() - 1);
+
+  VELOX_DCHECK_EQ(args.size(), 1);
+  const auto arg = args[0];
+
+  return std::make_shared<CallTypedExpr>(
+      arg->type(),
+      std::vector<TypedExprPtr>{
+          arg,
+          std::make_shared<ConstantTypedExpr>(velox::BIGINT(), 1LL),
+          std::make_shared<ConstantTypedExpr>(velox::BIGINT(), (int64_t)length),
+      },
+      "presto.default.substr");
+}
+
 /// Converts cast and try_cast functions to CastTypedExpr with nullOnFailure
 /// flag set to false and true appropriately.
 /// Removes cast to Re2JRegExp type. Velox doesn't have such type and uses
@@ -166,6 +200,8 @@ namespace {
 /// regular expressions needlessly.
 /// Removes cast to CodePoints type. Velox doesn't have such type and uses
 /// different mechanisms to implement trim functions efficiently.
+/// Convert cast of varchar to substr if the target type is varchar with max
+/// length. Throw an exception for cast of varchar to varchar with max length.
 std::optional<TypedExprPtr> tryConvertCast(
     const protocol::Signature& signature,
     const std::string& returnType,
@@ -217,6 +253,15 @@ std::optional<TypedExprPtr> tryConvertCast(
 
   if (returnType == kCodePoints) {
     return args[0];
+  }
+
+  // When the return type is varchar with max length, truncate if only the
+  // argument type is varchar (or varchar with max length). Non-varchar argument
+  // types are not truncated.
+  if (returnType.find(kVarchar) == 0 &&
+      args[0]->type()->kind() == TypeKind::VARCHAR &&
+      returnType.size() > strlen(kVarchar)) {
+    return convertCastToVarcharWithMaxLength(returnType, args, nullOnFailure);
   }
 
   auto type = typeParser->parse(returnType);
--- a/presto-native-execution/presto_cpp/main/types/tests/RowExpressionTest.cpp
+++ b/presto-native-execution/presto_cpp/main/types/tests/RowExpressionTest.cpp
@@ -49,6 +49,54 @@ class RowExpressionTest : public ::testing::Test {
     ASSERT_EQ(cexpr->value().toJson(cexpr->type()), value);
   }
 
+  std::string makeCastToVarchar(
+      bool isTryCast,
+      const std::string& inputType,
+      const std::string& returnType) {
+    std::string signatureNameField = isTryCast
+        ? R"("name": "presto.default.try_cast")"
+        : R"("name": "presto.default.$operator$cast")";
+    std::string inputTypeField = fmt::format("\"{}\"", inputType);
+    std::string returnTypeField =
+        fmt::format("\"returnType\": \"{}\"", returnType);
+
+    std::string result = R"##(
+      {
+        "@type": "call",
+        "arguments": [
+          {
+            "@type": "variable",
+            "name": "my_col",
+            "type": )##" +
+        inputTypeField + R"##(
+          }
+        ],
+        "displayName": "CAST",
+        "functionHandle": {
+          "@type": "$static",
+          "signature": {
+            "argumentTypes": [
+    )##" +
+        inputTypeField + R"##(
+            ],
+            "kind": "SCALAR",
+    )##" +
+        signatureNameField + R"##(,
+            "longVariableConstraints": [],
+    )##" +
+        returnTypeField + R"##(,
+            "typeVariableConstraints": [],
+            "variableArity": false
+          }
+        },
+    )##" +
+        returnTypeField + R"##(
+      }
+    )##";
+
+    return result;
+  }
+
   std::shared_ptr<memory::MemoryPool> pool_;
   std::unique_ptr<VeloxExprConverter> converter_;
   TypeParser typeParser_;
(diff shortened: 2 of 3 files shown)


presto-native-execution/presto_cpp/main/types/PrestoToVeloxExpr.cpp
```diff
@@ -159,13 +159,48 @@ std::vector<TypedExprPtr> VeloxExprConverter::toVeloxExpr(
 }
 
 namespace {
+static const char* kVarchar = "varchar";
+
+/// Convert cast to a call to substr if target type is varchar with max length.
+/// Convert try_cast to a call to try(substr) if target type is varchar with max
```
mbasmanova, line 165: @gggrace14 try_cast is not equivalent to try(substr). See https://velox-lib.io/blog/#try-and-trycast

mbasmanova (commented):
@gggrace14 Ge, please, add a test.

aditi-pandit:
Resolves https://github.com/prestodb/presto/issues/20847 as well

gggrace14:
> @gggrace14 Ge, please, add a test.

Unit tests are added.

gggrace14:
> @gggrace14 Ge, please, add a test.

Unit tests are added.

aditi-pandit:
@gggrace14 : Wanted to mention that @czentgr has been working to support varchar(n) end to end in Velox. So this code is likely to be removed after that makes it in.

amitkdutta:
> @gggrace14 : Wanted to mention that @czentgr has been working to support varchar(n) end to end in Velox. So this code is likely to be removed after that makes it in.

Yea, makes sense. Once we have `varchar(n)` support, we can remove this code.

amitkdutta (approved):
Looks good to me. Thanks @gggrace14

aditi-pandit:
@gggrace14 : Please add an e2e test for this CAST sql in AbstractTestNativeGeneralQueries.java https://github.com/prestodb/presto/blob/master/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeGeneralQueries.java

presto-native-execution/presto_cpp/main/types/PrestoToVeloxExpr.cpp
```diff
@@ -219,6 +252,12 @@ std::optional<TypedExprPtr> tryConvertCast(
     return args[0];
   }
 
+  if (returnType.find(kVarchar) == 0 &&
+      args[0]->type()->kind() == TypeKind::VARCHAR &&
+      returnType.size() > strlen(kVarchar)) {
```
aditi-pandit, line 257: Maybe abstract a constexpr for strlen(kVarchar) since its repeated.

presto-native-execution/presto_cpp/main/types/PrestoToVeloxExpr.cpp
```diff
@@ -159,13 +159,46 @@ std::vector<TypedExprPtr> VeloxExprConverter::toVeloxExpr(
...
+std::optional<TypedExprPtr> convertCastToVarcharWithMaxLength(
+    const std::string& returnType,
+    const std::vector<TypedExprPtr>& args,
+    bool nullOnFailure) {
+  if (nullOnFailure) {
+    VELOX_NYI("TRY_CAST of varchar to {} is not supported.", returnType);
+  }
+
+  char* end;
+  const auto length =
```
aditi-pandit, line 175: Please add a comment that you are trying to parse the length from a type string that looks like varchar[length] here.

Most of the code in this class uses Velox TypeParser https://github.com/facebookincubator/velox/blob/main/velox/type/parser/TypeParser.yy to parse returnType. Might be good to enhance it for varchar(n) instead of the custom code.

gggrace14:
> @gggrace14 : Wanted to mention that @czentgr has been working to support varchar(n) end to end in Velox. So this code is likely to be removed after that makes it in.

Hi @aditi-pandit yes, that is expected. This could be a temporary solution to align with Presto behavior before we have varchar(n).

gggrace14:
> @gggrace14 : Please add an e2e test for this CAST sql in AbstractTestNativeGeneralQueries.java https://github.com/prestodb/presto/blob/master/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeGeneralQueries.java

e2e test cases are added.

presto-native-execution/presto_cpp/main/types/PrestoToVeloxExpr.cpp
gggrace14: Comment is added.

czentgr: This would require to have a VARCHAR(N) type available as it would need to be created as a result of the parsing.
I have this change in the prototype.

presto-native-execution/presto_cpp/main/types/tests/RowExpressionTest.cpp
```diff
@@ -49,6 +49,54 @@ class RowExpressionTest : public ::testing::Test {
     ASSERT_EQ(cexpr->value().toJson(cexpr->type()), value);
   }
 
+  std::string makeCastToVarchar(
+      bool isTryCase,
```
aditi-pandit, line 53: Nit : Rename variable to isTryCast.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeGeneralQueries.java
```diff
@@ -428,6 +428,16 @@ public void testCast()
         assertQuery("SELECT CAST(true as VARCHAR), CAST(false as VARCHAR)");
         assertQuery("SELECT CAST(0.0 as VARCHAR)");
 
+        // Cast to varchar(n).
+        assertQuery("SELECT CAST('string_longer_than_1_character' AS VARCHAR(1)");
```
aditi-pandit, line 432: Some of the test queries are on constant string arguments. Such CAST expressions get evaluated at the co-ordinator itself in the constant folding optimizer rule. Please add tests that are on column references here.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeGeneralQueries.java
```diff
@@ -428,6 +428,16 @@ public void testCast()
...
+        assertQuery("SELECT CAST(comments as VARCHAR(1)) FROM orders)");
+        assertQuery("SELECT CAST('string_shorter_than_1000_character' AS VARCHAR(1000)");
+        assertQuery("SELECT CAST(comments as VARCHAR(1000)) FROM orders WHERE LENGTH(comment) < 1000");
```
aditi-pandit, line 435: Does this data have comments > varchar(1000) ? So there might not be any truncation.

presto-native-execution/presto_cpp/main/types/PrestoToVeloxExpr.cpp
```diff
@@ -219,6 +255,12 @@ std::optional<TypedExprPtr> tryConvertCast(
     return args[0];
   }
 
+  if (returnType.find(kVarchar) == 0 &&
```
aditi-pandit, line 258: Please add a comment that only varchar type is truncated. Non-varchar types are not truncated.

Also do you handle a case where we are going from varchar(m) to varchar(n) ?

presto-native-execution/presto_cpp/main/types/tests/RowExpressionTest.cpp
```diff
@@ -490,6 +538,80 @@ TEST_F(RowExpressionTest, call) {
   }
 }
 
+TEST_F(RowExpressionTest, castToVarchar) {
+  // CAST(varchar_col AS varchar)
+  {
+    json j = json::parse(makeCastToVarchar(false, "varchar", "varchar"));
+    std::shared_ptr<protocol::CallExpression> p = j;
```
aditi-pandit, line 545: Can we assign p directly without having variable j in between ?

aditi-pandit (commented):
Thanks @gggrace14

presto-native-execution/presto_cpp/main/types/tests/RowExpressionTest.cpp
gggrace14: A good catch. Thanks, Aditi!

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeGeneralQueries.java
gggrace14: Yes, you're right. I'm removing the test cases on constant string, as I have included test cases on column references of orders. They're redundant.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeGeneralQueries.java
gggrace14: Yes, this test case is to test no truncation for varchar values shorter than 1000 if the max length of return type is set to 1000.

presto-native-execution/presto_cpp/main/types/tests/RowExpressionTest.cpp
gggrace14: Yes, it works. Revising.

(Thought about it before, but turned out using the same pattern as the rest of unit tests.)

presto-native-execution/presto_cpp/main/types/PrestoToVeloxExpr.cpp
gggrace14: Comment is added.

Yes, this condition includes the case of casting from varchar(m) to varchar(n), by `args[0]->type()->kind() == TypeKind::VARCHAR`. varchar(m) is also parsing as ScalarType<TypeKind::VARCHAR>, as below if I'm looking at the right place. https://github.com/facebookincubator/velox/blob/037c2befd7bce504c0a09ccc777d578206e057a5/velox/type/fbhive/HiveTypeParser.cpp#L108-L120

aditi-pandit: I see.. Seems like the HiveTypeParser translates VARCHAR(n) to VARCHAR. It will need to be enhanced to handle VARCHAR(n) in the future.

aditi-pandit: I see.. Seems like the HiveTypeParser translates VARCHAR(n) to VARCHAR. It will need to be enhanced to handle VARCHAR(n) in the future.

aditi-pandit: I see.. Seems like the HiveTypeParser translates VARCHAR(n) to VARCHAR. It will need to be enhanced to handle VARCHAR(n) in the future.

aditi-pandit (approved):
Thanks for the iterations @gggrace14

gggrace14:
> Thanks for the iterations @gggrace14

Thank you, Aditi, for the review!

PR #23143, opened 2024-07-08
Fix invalid limit clause

ajaygeorge:
## Description
This commit tries to fix Invalid limit clauses that
currently throws a NumberFormatException which is a GENERIC_INTERNAL_ERROR.
Instead the NumberFormatException is caught and a PrestoException with
INVALID_LIMIT_CLAUSE is thrown which is a USER_ERROR.

## Motivation and Context
Correctly classify errors.
Sample query id 20240708_160446_69458_yvcau

## Test Plan
Unit Tests

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```


swapsmagic (dismissed):
thanks for fixing this.

feilong-liu (dismissed):

presto-main/src/test/java/com/facebook/presto/sql/planner/TestLogicalPlanner.java
```diff
@@ -1364,6 +1368,24 @@ public void testLimitZero()
...
+    public void testInvalidLimit()
+    {
+        try {
+            assertPlan(
+                    "SELECT orderkey FROM orders LIMIT 10000000000000000000000",
+                    output(
+                            values("NOOP")));
+            fail("PrestoException not thrown for invalid limit");
+        }
+        catch (Exception e) {
```
elharo, line 1381: only catch PrestoException here. Then you remove the next two lines

ajaygeorge: That would defeat the purpose of assertions.

say if the code throws a RuntimeException instead of the expected PrestoException.

```
    @Test
    public void testInvalidLimit()
    {
        try {
           throw new RuntimeException();
        }
        catch (PrestoException prestoException) {
            assertEquals(prestoException.getErrorCode(), INVALID_LIMIT_CLAUSE.toErrorCode());
            assertEquals(prestoException.getMessage(), "Invalid limit: 10000000000000000000000");
        }
    }
```

  We still get a test failure but not for **failure** of assertions but for the RuntimeException bubbling up.
    
  The idea here is to assert that the Exception that we get is indeed a PrestoException and not something else.
 for e.g. with this code

```
    @Test
    public void testInvalidLimit()
    {
        try {
           throw new RuntimeException();
        }
        catch (Exception e) {
            assertTrue(e instanceof PrestoException);
            PrestoException prestoException = (PrestoException) e;
            assertEquals(prestoException.getErrorCode(), INVALID_LIMIT_CLAUSE.toErrorCode());
            assertEquals(prestoException.getMessage(), "Invalid limit: 10000000000000000000000");
        }
    }
```
you will get an assertion failure for assertTrue line.

tdcmeehan (dismissed):

presto-main/src/test/java/com/facebook/presto/sql/planner/TestLogicalPlanner.java
elharo: Not quite. Catch PrestoException only. If anything else is thrown, that fails the test. The runtime exception bubboing up is exactly what we want in that case.

elharo (dismissed):

presto-main/src/test/java/com/facebook/presto/sql/planner/TestLogicalPlanner.java
ajaygeorge: I think it is a question of readability.
for the current code you get
```
java.lang.AssertionError:
Expected :true
Actual   :false

	at org.testng.Assert.fail(Assert.java:110)
	at org.testng.Assert.failNotEquals(Assert.java:1413)
	at org.testng.Assert.assertTrue(Assert.java:56)
	at org.testng.Assert.assertTrue(Assert.java:66)
	at com.facebook.presto.sql.planner.TestLogicalPlanner.testInvalidLimit(TestLogicalPlanner.java:1377)
```

with your proposal we get just the stacktrace without any pointers as to why the test failed.

```
java.lang.RuntimeException
	at com.facebook.presto.sql.planner.TestLogicalPlanner.testInvalidLimit(TestLogicalPlanner.java:1373)
```

If I add a message to the assertTrue I get an even more meaningful message back to the user.

```
assertTrue(e instanceof PrestoException, "Expected PrestoException but found " + e);
```

```
java.lang.AssertionError: Expected PrestoException but found java.lang.RuntimeException
Expected :true
Actual   :false

	at org.testng.Assert.fail(Assert.java:110)
	at org.testng.Assert.failNotEquals(Assert.java:1413)
	at org.testng.Assert.assertTrue(Assert.java:56)
	at com.facebook.presto.sql.planner.TestLogicalPlanner.testInvalidLimit(TestLogicalPlanner.java:1377)
```

Hope why keeping the assertion adds more clarity to a user looking at a test failure.

feilong-liu (approved):

PR #23147, opened 2024-07-08
Support query retry for HBO

abhinavmuk04:
## Description
Implement retry query support for HBO in Presto, which retries failed queries and tries to retry if the query plan has changed. If the plan has changed it will re-execute, otherwise after the plan comparison we immediately fail it back

## Motivation and Context
Many times queries fail because of cases where estimates are inaccurate and various out-of-memory, errors can occur. With this change enabled, users will be able to have an automatic retry which will determine if a retry of the query can prevent their query from failing entirely

## Impact
Now when a query fails and a user reruns it, ending up in the query succeeding, with this session property enabled, we can speed up this process to ensure it automatically retries to see if HBO can help the query

## Test Plan
Various created unit tests deeply testing different parts of the new and amended code's functionality. Also E2E tested feature within verified cluster

Example Query:
[Original](https://www.internalfb.com/intern/presto/query/?query_id=20240724_225224_00678_vwf49#plan) (Failure) ------> [Retry](https://www.internalfb.com/intern/presto/query/?query_id=20240724_225325_00679_vwf49#plan) (Successful w/Plan Change) 

## Contributor checklist

- [x] CI passed.

## Release Notes
```
* Add support for query retry when History-Based Optimization can help a failed query,
   with the ``retry-query-with-history-based-optimization`` session property :pr:`23147`
```


tdcmeehan:
I understand this is draft, but what is the design or idea behind this change?

abhinavmuk04:
> I understand this is draft, but what is the design or idea behind this change?

The goal is to implement a simple version of retry query support for HBO in Presto, which retries failed queries and checks if the query plan has changed before re-executing. Currently, just building out basic portions.

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
```diff
@@ -383,26 +388,57 @@ private synchronized Optional<QueryResults> getCachedResult(long token)
...
+
+        if (queryResults.getError() != null && historyBasedOptimizationEnabled && underOneRetryCount && underRetryExecutionTime && retryQueryWithHistoryBasedOptimizationEnabled(session)) {
+            if (lastPlanHashes == null) {
+                lastPlanHashes = currentPlanHashes;
+            }
+        }
+        else if (queryManager.getQueryRetryCount(queryId) == 1 && retryQueryWithHistoryBasedOptimizationEnabled(session)) {
+            if (currentPlanHashes.size() == lastPlanHashes.size()) {
+                for (int planHash = 0; planHash < currentPlanHashes.size(); planHash++) {
+                    if (!currentPlanHashes.get(planHash).getInfo().getHash().equals(lastPlanHashes.get(planHash).getInfo().getHash())) {
```
jaystarshot, line 407: Note Don't think its a deal breaker but the plan hashes can change if the underlying data has changed. (if someone added a new partition)

abhinavmuk04: Thinking of potentially future work that can be built of this. For the first version I wanted to keep it simple, but I agree with you. [Plan Hash vs Plan comparison doc](https://docs.google.com/document/d/1w6yQojt6wLfs509uXNaoAx1tNgWw8inIsNARsawLeQg/edit?usp=sharing)

presto-tests/src/test/java/com/facebook/presto/tests/TestHistoryBasedRetry.java
```diff
@@ -0,0 +1,222 @@
...
+                ListenableFuture<?> future = executor.submit(() -> queryRunner.execute(session, sql));
+                queryFutures.add(future);
+            }
+            waitForQueryToFinish(queryFutures);
+
+            DistributedQueryRunner x = (DistributedQueryRunner) queryRunner;
+            List<BasicQueryInfo> queryInfos = x.getCoordinator().getQueryManager().getQueries();
+
+            for (BasicQueryInfo info : queryInfos) {
+                System.out.println(info.getErrorCode());
```
elharo, line 93: delete this line; never call System.out.println in production or tests; only when debugging

presto-tests/src/test/java/com/facebook/presto/tests/TestHistoryBasedRetry.java
```diff
@@ -0,0 +1,222 @@
...
```
elharo, line 131: delete

presto-tests/src/test/java/com/facebook/presto/tests/TestHistoryBasedRetry.java
```diff
@@ -0,0 +1,222 @@
...
+
+    @Test
+    public void testQueryRetryWithHBONewQueryPlan()
+            throws Exception
+    {
+        int totalQueries = 1;
+        int retryLimit = 3;
+        int retryNum = 0;
+
+        String sql = "select o.orderkey, l.partkey, l.mapcol[o.orderkey] from (select orderkey, partkey, mapcol " +
```
elharo, line 111: nit, but I find SQL a little easier to read when all caps are used for keywords; e.g. SELECT

presto-tests/src/test/java/com/facebook/presto/tests/TestHistoryBasedRetry.java
```diff
@@ -0,0 +1,222 @@
...
+                ListenableFuture<?> future = executor.submit(() -> queryRunner.execute(session, query));
+                queryFutures.add(future);
+            }
+            waitForQueryToFinish(queryFutures);
+
+            DistributedQueryRunner x = (DistributedQueryRunner) queryRunner;
+            List<BasicQueryInfo> queryInfos = x.getCoordinator().getQueryManager().getQueries();
+
+            for (BasicQueryInfo info : queryInfos) {
+                System.out.println(info.getErrorCode());
```
elharo, line 166: delete

presto-tests/src/test/java/com/facebook/presto/tests/TestHistoryBasedRetry.java
```diff
@@ -0,0 +1,222 @@
...
+    public void testQueryRetryWithHBOForceFail()
+            throws Exception
+    {
+        int totalQueries = 1;
+        int retryLimit = 1;
+        int retryNum = 0;
+
+        String query = "SELECT if(COUNT(*)=1,1,fail(1, 'failed')) FROM part p LEFT JOIN lineitem l ON p.partkey = l.partkey WHERE l.comment like '%a%'";
+
+        Session session = Session.builder(createSession())
```
elharo, line 150: Haven't looked to closely, but a lot of these methods seem to have a lot of duplicate code. It might be possible to reuse some of that, but don't bother if it's too much trouble

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
```diff
@@ -383,26 +390,56 @@ private synchronized Optional<QueryResults> getCachedResult(long token)
...
+        if (queryResults.getError() != null && historyBasedOptimizationEnabled && underOneRetryCount && underRetryExecutionTime && retryQueryWithHistoryBasedOptimizationEnabled(session)) {
+            CanonicalPlanWithInfo currentQueryTopLevelPlanHash = queryManager.getFullQueryInfo(queryId).getPlanCanonicalInfo().get(0);
+
+            if (originalBeforeRetryQueryId == null && previousQueryTopLevelPlanHash == null) {
+                originalBeforeRetryQueryId = queryId;
+                previousQueryTopLevelPlanHash = currentQueryTopLevelPlanHash;
+            }
         }
+        else if (queryManager.getQueryRetryCount(queryId) == 1 && retryQueryWithHistoryBasedOptimizationEnabled(session)) {
+            CanonicalPlanWithInfo currentQueryTopLevelPlanHash = queryManager.getFullQueryInfo(queryId).getPlanCanonicalInfo().get(0);
```
feilong-liu, line 407: Directly use the existing hash may not work, at least for changes made in DeterminJoinDistributionType, as the hash is taken before this broadcast

feilong-liu (commented):
Have you done an end to end test for the change? I think for the plan hash comparison, it may not work if we are targeting the

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
```diff
@@ -98,6 +103,8 @@ class Query
 {
     private static final Logger log = Logger.get(Query.class);
     private static final Base64.Encoder BASE64_ENCODER = Base64.getEncoder();
+    private static QueryId originalBeforeRetryQueryId;
```
agrawaldevesh, line 106: Can there only be one retry ? What if we call this originalQueryId, or perhaps we maintain this as a list (of query ids), where there should be one to begin with but the 0'th item contains the original attempt and then subsequent ones get appended. or would that be an overkill ? 

Same for the plan hashes -- but maybe its ok for the hash, since we only care about the "previous"

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
```diff
@@ -383,26 +390,56 @@ private synchronized Optional<QueryResults> getCachedResult(long token)
     private synchronized QueryResults getNextResultWithRetry(long token, UriInfo uriInfo, String scheme, DataSize targetResultSize, boolean binaryResults)
     {
         QueryResults queryResults = getNextResult(token, uriInfo, scheme, targetResultSize, binaryResults);
-        if (queryResults.getError() == null || !queryResults.getError().isRetriable()) {
-            return queryResults;
+
+        boolean historyBasedOptimizationEnabled = useHistoryBasedPlanStatisticsEnabled(session) && trackHistoryBasedPlanStatisticsEnabled(session);
+        boolean underOneRetryCount = queryManager.getQueryRetryCount(queryId) < 1;
```
agrawaldevesh, line 395: how about hasNotRetried ?

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
```diff
@@ -383,26 +390,56 @@ private synchronized Optional<QueryResults> getCachedResult(long token)
...
+        boolean underRetryExecutionTime = queryManager.getQueryInfo(queryId).getQueryStats().getExecutionTime().toMillis() <= getQueryRetryMaxExecutionTime(session).toMillis();
```
agrawaldevesh, line 396: Or do you want to record the start time (or maybe the create time) of the original query and only retry if you are under the budget ?

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
```diff
@@ -98,6 +103,8 @@ class Query
 {
     private static final Logger log = Logger.get(Query.class);
     private static final Base64.Encoder BASE64_ENCODER = Base64.getEncoder();
+    private static QueryId originalBeforeRetryQueryId;
+    private static CanonicalPlanWithInfo previousQueryTopLevelPlanHash;
```
agrawaldevesh, line 107: Should these be Optionals instead of using null ?

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
```diff
@@ -383,26 +390,56 @@ private synchronized Optional<QueryResults> getCachedResult(long token)
...
+
+            if (originalBeforeRetryQueryId == null && previousQueryTopLevelPlanHash == null) {
+                originalBeforeRetryQueryId = queryId;
+                previousQueryTopLevelPlanHash = currentQueryTopLevelPlanHash;
+            }
         }
+        else if (queryManager.getQueryRetryCount(queryId) == 1 && retryQueryWithHistoryBasedOptimizationEnabled(session)) {
+            CanonicalPlanWithInfo currentQueryTopLevelPlanHash = queryManager.getFullQueryInfo(queryId).getPlanCanonicalInfo().get(0);
+            if (previousQueryTopLevelPlanHash.equals(currentQueryTopLevelPlanHash)) {
+                queryManager.failQuery(queryId, new PrestoException(GENERIC_INTERNAL_ERROR, "Since the plan hashes did not change, your retry query will not execute"));
```
agrawaldevesh, line 409: Hmm, perhaps we want to just return the original error (code and message). Or perhaps you can tweak the message to include this information. Also, in this error message do you want to record the old/new query-id's also ?

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
```diff
@@ -383,26 +390,56 @@ private synchronized Optional<QueryResults> getCachedResult(long token)
...
+                return queryResults;
+            }
```
agrawaldevesh, line 411: In the else branch, should we do some logging on why the retry codepath wasn't chosen ? (if my understanding is right)

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
```diff
@@ -383,26 +390,56 @@ private synchronized Optional<QueryResults> getCachedResult(long token)
...
         }
 
-        // no support for transactions
-        if (session.getTransactionId().isPresent() &&
-                !transactionManager.getOptionalTransactionInfo(session.getRequiredTransactionId()).map(TransactionInfo::isAutoCommitContext).orElse(true)) {
-            return queryResults;
+        // no more retries left
+        if (!underOneRetryCount) {
+            originalBeforeRetryQueryId = null;
```
agrawaldevesh, line 441: Do these two originalBeforeRetryQueryId and previousQueryTopLevel ... need to be member variables ?

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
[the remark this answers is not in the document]
abhinavmuk04: @feilong-liu Checking the retry query count will work as it stays static across queries and their retry.

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
```diff
@@ -383,26 +390,56 @@ private synchronized Optional<QueryResults> getCachedResult(long token)
...
     {
         QueryResults queryResults = getNextResult(token, uriInfo, scheme, targetResultSize, binaryResults);
-        if (queryResults.getError() == null || !queryResults.getError().isRetriable()) {
-            return queryResults;
+
+        boolean historyBasedOptimizationEnabled = useHistoryBasedPlanStatisticsEnabled(session) && trackHistoryBasedPlanStatisticsEnabled(session);
+        boolean underOneRetryCount = queryManager.getQueryRetryCount(queryId) < 1;
+        boolean underRetryExecutionTime = queryManager.getQueryInfo(queryId).getQueryStats().getExecutionTime().toMillis() <= getQueryRetryMaxExecutionTime(session).toMillis();
+
+        if (queryResults.getError() != null && historyBasedOptimizationEnabled && underOneRetryCount && underRetryExecutionTime && retryQueryWithHistoryBasedOptimizationEnabled(session)) {
```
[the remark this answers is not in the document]
abhinavmuk04: @feilong-liu This first if statement will be hit only during the original query execution. During the retry the plan change logic comes into play

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
abhinavmuk04: @agrawaldevesh I think that in this case we only want to check for one query as retries aside from that are for transient errors

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
abhinavmuk04: @agrawaldevesh just previousQueryTopLevel is stored in the event of a retry, this way we can quickly compare the plan hashes

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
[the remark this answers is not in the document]
feilong-liu: Can you run an e2e test to verify?
My understanding is this if statement will be hit during the original query execution, and it will move to submit a retry query without checking plan change (and if other conditions met).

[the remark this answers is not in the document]
abhinavmuk04: @feilong-liu In order to see the plan change we need to replan with the statistics, we do not fully retry the query if the plan has not changed

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
abhinavmuk04: @agrawaldevesh Do you think this may cause too much logging? Maybe logging only if retry is enabled?

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
abhinavmuk04: @feilong-liu Will add this as an issue and work on it in a separate PR as discussed

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
```diff
@@ -383,26 +391,38 @@ private synchronized Optional<QueryResults> getCachedResult(long token)
     private synchronized QueryResults getNextResultWithRetry(long token, UriInfo uriInfo, String scheme, DataSize targetResultSize, boolean binaryResults)
     {
         QueryResults queryResults = getNextResult(token, uriInfo, scheme, targetResultSize, binaryResults);
-        if (queryResults.getError() == null || !queryResults.getError().isRetriable()) {
```
feilong-liu, line 386: Keep the error null check and return results if no error

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
```diff
@@ -383,26 +391,38 @@ private synchronized Optional<QueryResults> getCachedResult(long token)
...
+        if (historyBasedOptimizationEnabled && hasNotRetried && retryConditionsMet(queryResults) && retryQueryWithHistoryBasedOptimizationEnabled(session)) {
+            if (!originalBeforeRetryQueryId.isPresent() && !previousQueryTopLevelPlanHash.isPresent()) {
+                originalBeforeRetryQueryId = Optional.of(queryId);
+                previousQueryTopLevelPlanHash = getCurrentTopLevelPlanHash();
+                previousQueryFailureError = Optional.of(queryResults.getError());
+            }
         }
+        else if (queryManager.getQueryRetryCount(queryId) == 1 && retryQueryWithHistoryBasedOptimizationEnabled(session) && retryConditionsMet(queryResults)) {
+            if (planHashesEqual(getCurrentTopLevelPlanHash())) {
+                queryManager.failQuery(queryId, new PrestoException(GENERIC_INTERNAL_ERROR, "Since the plan hashes did not change, your retry query will not execute." +
```
feilong-liu, line 407: Ideally we should keep the retry transparent to the user. Is it possible to have the user only see the failure of the original query without awareness of the retry?

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
```diff
@@ -687,6 +707,53 @@ private static URI findCancelableLeafStage(StageInfo stage)
...
+        if (queryResults.getError() == null) {
+            return false;
+        }
+
+        if (!retryQueryWithHistoryBasedOptimizationEnabled(session) && !queryResults.getError().isRetriable()) {
+            return false;
+        }
+
+        // check if we have exceeded the global limit
+        retryCircuitBreaker.incrementFailure();
```
feilong-liu, line 729: This functions is called multiple times above, and it may increment the failure multiple times in one run.

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
```diff
@@ -383,26 +391,38 @@ private synchronized Optional<QueryResults> getCachedResult(long token)
...
 
-        // check if we have exceeded the global limit
-        retryCircuitBreaker.incrementFailure();
-        if (!retryCircuitBreaker.isRetryAllowed() || hasProducedResult) {
-            return queryResults;
+        boolean historyBasedOptimizationEnabled = useHistoryBasedPlanStatisticsEnabled(session) && trackHistoryBasedPlanStatisticsEnabled(session);
+        boolean hasNotRetried = queryManager.getQueryRetryCount(queryId) < 1;
+
+        if (historyBasedOptimizationEnabled && hasNotRetried && retryConditionsMet(queryResults) && retryQueryWithHistoryBasedOptimizationEnabled(session)) {
+            if (!originalBeforeRetryQueryId.isPresent() && !previousQueryTopLevelPlanHash.isPresent()) {
```
feilong-liu, line 399: If it's the first run, we should not have originalBeforeRetryQueryId and previousQueryTopLevelPlanHash present?

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
```diff
@@ -687,6 +707,53 @@ private static URI findCancelableLeafStage(StageInfo stage)
         return stage.getSelf();
     }
 
+    private boolean planHashesEqual(Optional<CanonicalPlanWithInfo> currentTopLevelPlanHash)
```
feilong-liu, line 710: This is a relatively simple function and only called once, may not need to be a separate function

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
```diff
@@ -383,26 +391,38 @@ private synchronized Optional<QueryResults> getCachedResult(long token)
...
+                return queryResults;
+            }
+        }
 
-        // no support for transactions
-        if (session.getTransactionId().isPresent() &&
-                !transactionManager.getOptionalTransactionInfo(session.getRequiredTransactionId()).map(TransactionInfo::isAutoCommitContext).orElse(true)) {
-            return queryResults;
+        // no more retries left
+        if (!hasNotRetried) {
```
feilong-liu, line 422: Can it be moved to the else if block in L405?

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
```diff
@@ -383,26 +390,56 @@ private synchronized Optional<QueryResults> getCachedResult(long token)
...
-                queryManager.getQueryInfo(queryId).getQueryStats().getExecutionTime().toMillis() > getQueryRetryMaxExecutionTime(session).toMillis()) {
-            return queryResults;
+            // check if we have exceeded the global limit
+            retryCircuitBreaker.incrementFailure();
+            if (!retryCircuitBreaker.isRetryAllowed() || hasProducedResult) {
+                return queryResults;
+            }
+
+            // check if we have exceeded the local limit
+            if (queryManager.getQueryRetryCount(queryId) >= getQueryRetryLimit(session) ||
```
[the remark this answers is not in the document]
feilong-liu: Double checked the code, I think we do not need this for HBO retry as we are always doing at most one retry

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
```diff
@@ -383,26 +390,56 @@ private synchronized Optional<QueryResults> getCachedResult(long token)
...
 
-        // check if we have exceeded the local limit
-        if (queryManager.getQueryRetryCount(queryId) >= getQueryRetryLimit(session) ||
-                queryManager.getQueryInfo(queryId).getQueryStats().getExecutionTime().toMillis() > getQueryRetryMaxExecutionTime(session).toMillis()) {
-            return queryResults;
+            // check if we have exceeded the global limit
+            retryCircuitBreaker.incrementFailure();
+            if (!retryCircuitBreaker.isRetryAllowed() || hasProducedResult) {
+                return queryResults;
+            }
```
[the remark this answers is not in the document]
feilong-liu: Can you check how the retryCircuitBreaker works? Now I am not sure if we want to do this for HBO retry.

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
abhinavmuk04: @feilong-liu When the query is running they will see both queryIds as the retry takes it course. We can tweak the fail message, but the user will always see the retry in this case

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
abhinavmuk04: @feilong-liu I am looking at the code and am not sure if we need this for the retry

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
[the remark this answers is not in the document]
abhinavmuk04: @feilong-liu Yeah, I agree

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
feilong-liu: Let's make it explicit and return early

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
```diff
@@ -383,27 +391,33 @@ private synchronized Optional<QueryResults> getCachedResult(long token)
...
-        if (queryManager.getQueryRetryCount(queryId) >= getQueryRetryLimit(session) ||
-                queryManager.getQueryInfo(queryId).getQueryStats().getExecutionTime().toMillis() > getQueryRetryMaxExecutionTime(session).toMillis()) {
-            return queryResults;
+        if (historyBasedOptimizationEnabled && hasNotRetried && retryConditionsMet(queryResults) && retryQueryWithHistoryBasedOptimizationEnabled(session)) {
+            originalBeforeRetryQueryId = Optional.of(queryId);
+            previousQueryTopLevelPlanHash = getCurrentTopLevelPlanHash();
+            previousQueryFailureError = Optional.of(queryResults.getError());
         }
+        else if (queryManager.getQueryRetryCount(queryId) == 1 && retryQueryWithHistoryBasedOptimizationEnabled(session) && retryConditionsMet(queryResults)) {
+            if (getCurrentTopLevelPlanHash().hashCode() == previousQueryTopLevelPlanHash.hashCode()) {
```
feilong-liu, line 404: Comparing hash code is the same as comparing the whole class. I am thinking maybe we do not need to compare the whole class, comparing the hash may be enough.

presto-tests/src/test/java/com/facebook/presto/tests/TestHistoryBasedRetry.java
```diff
@@ -0,0 +1,214 @@
...
+    {
+        executor.shutdownNow();
+    }
+
+    @Test
+    public void testQueryRetryWithHBONaturalFail()
+            throws Exception
+    {
+        int totalQueries = 1;
+        int retryLimit = 3;
```
feilong-liu, line 70: I don't think this is used, the current code retry at most once

presto-tests/src/test/java/com/facebook/presto/tests/TestHistoryBasedRetry.java
```diff
@@ -0,0 +1,214 @@
...
+
+    private static Session createSession()
+    {
+        return testSessionBuilder()
+                .setCatalog("tpch")
+                .setSchema("tiny")
+                .setSystemProperty(USE_HISTORY_BASED_PLAN_STATISTICS, "true")
+                .setSystemProperty(TRACK_HISTORY_BASED_PLAN_STATISTICS, "true")
+                .setSystemProperty(RETRY_QUERY_WITH_HISTORY_BASED_OPTIMIZATION, "true")
+                .setSystemProperty(USE_PERFECTLY_CONSISTENT_HISTORIES, "true")
```
feilong-liu, line 207: Why set this session property?

presto-tests/src/test/java/com/facebook/presto/tests/TestHistoryBasedRetry.java
```diff
@@ -0,0 +1,214 @@
...
+                .setSystemProperty(JOIN_DISTRIBUTION_TYPE, "PARTITIONED")
+                .setSystemProperty(JOIN_REORDERING_STRATEGY, "NONE")
```
feilong-liu, line 209: Why disabling cost based join planning?

presto-tests/src/test/java/com/facebook/presto/tests/TestHistoryBasedRetry.java
```diff
@@ -0,0 +1,214 @@
...
+                "FROM (SELECT *, map(array[1], array[2]) mapcol FROM lineitem)) l " +
+                "JOIN orders o ON l.partkey = o.custkey WHERE length(comment) > 10";
+
+        Session session = Session.builder(createSession())
+                .setSystemProperty(QUERY_RETRY_LIMIT, String.valueOf(retryLimit))
+                .build();
+
+        try (QueryRunner queryRunner = createQueryRunner(session)) {
+            List<ListenableFuture<?>> queryFutures = new ArrayList<>();
+            for (int i = 0; i < totalQueries; i++) {
```
feilong-liu, line 83: totalQueries is hardcoded to 1, for loop not needed here

presto-tests/src/test/java/com/facebook/presto/tests/TestHistoryBasedRetry.java
```diff
@@ -0,0 +1,214 @@
...
+            waitForQueryToFinish(queryFutures);
+
+            retryNum = getRetryCount(queryRunner);
+        }
+
+        assertEquals(retryNum, 1, "Retry count should be one as the query plan has changed.");
+    }
+
+    @Test
+    public void testQueryRetryWithHBONewQueryPlan()
```
feilong-liu, line 96: What's the difference between this test and "testQueryRetryWithHBONaturalFail"?

feilong-liu (commented):
Can you also shadow production traffic with the feature turned on?

presto-tests/src/test/java/com/facebook/presto/tests/TestHistoryBasedRetry.java
abhinavmuk04: @feilong-liu I added it as part of this to ensure that the singular retry holds true

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
```diff
@@ -383,27 +390,33 @@ private synchronized Optional<QueryResults> getCachedResult(long token)
...
-        // check if we have exceeded the local limit
-        if (queryManager.getQueryRetryCount(queryId) >= getQueryRetryLimit(session) ||
-                queryManager.getQueryInfo(queryId).getQueryStats().getExecutionTime().toMillis() > getQueryRetryMaxExecutionTime(session).toMillis()) {
-            return queryResults;
+        if (historyBasedOptimizationEnabled && hasNotRetried && retryConditionsMet(queryResults) && retryQueryWithHistoryBasedOptimizationEnabled(session)) {
+            originalBeforeRetryQueryId = Optional.of(queryId);
+            previousQueryTopLevelPlanHash = getCurrentTopLevelPlanHash();
+            previousQueryFailureError = Optional.of(queryResults.getError());
         }
+        else if (queryManager.getQueryRetryCount(queryId) == 1 && retryQueryWithHistoryBasedOptimizationEnabled(session) && retryConditionsMet(queryResults)) {
```
feilong-liu, line 402: Should this also only execute when historyBasedOptimizationEnabled is true?

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
```diff
@@ -687,6 +700,50 @@ private static URI findCancelableLeafStage(StageInfo stage)
...
+        }
+
+        if (!retryQueryWithHistoryBasedOptimizationEnabled(session)) {
+            if (!queryResults.getError().isRetriable()) {
+                return false;
+            }
+
+            // check if we have exceeded the global limit
+            retryCircuitBreaker.incrementFailure();
+            if (!retryCircuitBreaker.isRetryAllowed() || hasProducedResult) {
```
feilong-liu, line 716: HBO retry should also stop retry if hasProducedResult is true.

presto-tests/src/test/java/com/facebook/presto/tests/TestHistoryBasedRetry.java
```diff
@@ -0,0 +1,178 @@
...
+    public void shutdown()
+    {
+        executor.shutdownNow();
+    }
+
+    @Test
+    public void testQueryRetryWithHBONaturalFail()
+            throws Exception
+    {
+        int totalQueries = 1;
```
feilong-liu, line 66: Unused variable

presto-tests/src/test/java/com/facebook/presto/tests/TestHistoryBasedRetry.java
```diff
@@ -0,0 +1,178 @@
...
+        int retryLimit = 3;
+        int retryNum = 0;
+
+        String sql = "SELECT o.orderkey, l.partkey, l.mapcol[o.orderkey] FROM (select orderkey, partkey, mapcol " +
+                "FROM (SELECT *, map(array[1], array[2]) mapcol FROM lineitem)) l " +
+                "JOIN orders o ON l.partkey = o.custkey WHERE length(comment) > 10";
+
+        Session session = Session.builder(createSession())
+                .setSystemProperty(QUERY_RETRY_LIMIT, String.valueOf(retryLimit))
```
feilong-liu, line 75: This is not used in HBO retry?

presto-tests/src/test/java/com/facebook/presto/tests/TestHistoryBasedRetry.java
```diff
@@ -0,0 +1,178 @@
...
+        }
+
+        assertEquals(retryNum, 1, "Retry count should be one as the query plan has changed.");
+    }
+
+    @Test
+    public void testQueryRetryWithHBOForceFail()
+            throws Exception
+    {
+        int retryLimit = 1;
```
feilong-liu, line 96: Similar

presto-tests/src/test/java/com/facebook/presto/tests/TestHistoryBasedRetry.java
```diff
@@ -0,0 +1,178 @@
...
+        int retryNum = 0;
+
+        String query = "SELECT if(COUNT(*)=1,1,fail(1, 'failed')) FROM part p LEFT JOIN lineitem l ON p.partkey = l.partkey WHERE l.comment like '%a%'";
+
+        Session session = Session.builder(createSession())
+                .setSystemProperty(QUERY_RETRY_LIMIT, String.valueOf(retryLimit))
```
feilong-liu, line 102: Similar

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
abhinavmuk04: @feilong-liu Yes, since we do not want retries caused by transient errors to enter this logic

presto-tests/src/test/java/com/facebook/presto/tests/TestHistoryBasedRetry.java
abhinavmuk04: @feilong-liu I wanted to set it to a number higher than 1 in one of the tests to ensure the retry count stayed as 1

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
feilong-liu: Can you add "historyBasedOptimizationEnabled" in the check condition then?

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
```diff
@@ -383,27 +390,37 @@ private synchronized Optional<QueryResults> getCachedResult(long token)
...
+        boolean historyBasedOptimizationEnabled = useHistoryBasedPlanStatisticsEnabled(session) && trackHistoryBasedPlanStatisticsEnabled(session);
+        boolean hasNotRetried = queryManager.getQueryRetryCount(queryId) < 1;
+
+        if (historyBasedOptimizationEnabled && hasNotRetried && retryConditionsMet(queryResults) && retryQueryWithHistoryBasedOptimizationEnabled(session)) {
+            originalBeforeRetryQueryId = Optional.of(queryId);
+            previousQueryTopLevelPlanHash = getCurrentTopLevelPlanHash();
+            previousQueryFailureError = Optional.of(queryResults.getError());
         }
+        else if (queryManager.getQueryRetryCount(queryId) == 1 && retryQueryWithHistoryBasedOptimizationEnabled(session) && retryConditionsMet(queryResults) && historyBasedOptimizationEnabled) {
+            if (getCurrentTopLevelPlanHash() == previousQueryTopLevelPlanHash) {
```
feilong-liu, line 407: I think we need to rule out the cases where both are empty.
Also == is for reference check, value check should use equals method

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
```diff
@@ -383,27 +390,39 @@ private synchronized Optional<QueryResults> getCachedResult(long token)
...
+
+        if (historyBasedOptimizationEnabled && hasNotRetried && retryConditionsMet(queryResults) && retryQueryWithHistoryBasedOptimizationEnabled(session)) {
+            originalBeforeRetryQueryId = Optional.of(queryId);
+            previousQueryTopLevelPlanHash = getCurrentTopLevelPlanHash();
+            previousQueryFailureError = Optional.of(queryResults.getError());
         }
+        else if (queryManager.getQueryRetryCount(queryId) == 1 && retryQueryWithHistoryBasedOptimizationEnabled(session) && retryConditionsMet(queryResults) && historyBasedOptimizationEnabled) {
+            Optional<Integer> currentTopLevelPlanHash = getCurrentTopLevelPlanHash();
+
+            if (previousQueryFailureError.isPresent() && currentTopLevelPlanHash.isPresent() && currentTopLevelPlanHash.equals(previousQueryTopLevelPlanHash)) {
```
feilong-liu, line 409: We should also fail the query if both are empty, i.e. HBO somehow not work for both original and retry queries, hence no plan change from HBO expected.

presto-main/src/main/java/com/facebook/presto/server/protocol/Query.java
```diff
@@ -383,27 +390,45 @@ private synchronized Optional<QueryResults> getCachedResult(long token)
...
+                queryManager.failQuery(queryId, new PrestoException(GENERIC_INTERNAL_ERROR, "Since the plan hashes did not change, your retry query will not execute." +
+                        "Your original error was " + previousQueryFailureError.get() + ". Original QueryId: " + originalBeforeRetryQueryId +
+                        ". Retry QueryId: " + queryId));
+            }
+
+            if (!previousQueryTopLevelPlanHash.isPresent() && !currentTopLevelPlanHash.isPresent()) {
+                queryManager.failQuery(queryId, new PrestoException(GENERIC_INTERNAL_ERROR, "Since the plan hashes are both empty, your retry query will not execute." +
+                        "Your original error was " + previousQueryFailureError.get() + ". Original QueryId: " + originalBeforeRetryQueryId +
+                        ". Retry QueryId: " + queryId));
+            }
```
feilong-liu, line 419: Overall looks good to me.
Just I do not think we need to give so much low level detail on plan hash is empty or change etc. in the error. combine these two if blocks, and simply say that the plan has not changed is enough.

abhinavmuk04: @feilong-liu Sounds good thanks!

feilong-liu (approved):

Issue #23156, opened 2024-07-09
[native] TestPrestoNativeIcebergTpcdsQueriesParquetUsingThrift.doDeletesAndQuery is failing


majetideepak:
[ERROR]   TestPrestoNativeIcebergTpcdsQueriesParquetUsingThrift.doDeletesAndQuery:399->verifyDeletes:390->AbstractTestQueryFramework.computeScalar:149->AbstractTestQueryFramework.computeActual:139 ? Runtime size <= capacity_ (18446744073709551499 vs. 1440) Split [Hive: file:/tmp/iceberg_data/HIVE/tpcds/store_sales/data/593eb530-2b2c-4b34-8bf9-caba66ad83d3.parquet 0 - 2071074] Task 20240709_152552_00114_sdniy.1.0.0.0 Operator: TableScan[0] 0
[INFO] 
[ERROR] Tests run: 132, Failures: 1, Errors: 0, Skipped: 0

The test started failing with this PR merge https://github.com/prestodb/presto/pull/23138

zacw7:
looking

zacw7:
The breaking change is https://github.com/facebookincubator/velox/pull/10261
@yingsu00 Could you please help fix this test?

tanjialiang:
> I will fix this.

We are reverting https://github.com/facebookincubator/velox/pull/10261 @zacw7 identified, followed up by an advance PR.

zacw7:
https://github.com/prestodb/presto/pull/23170 is merged and the test no longer fails. Closing the issue. Thanks!

PR #23157, opened 2024-07-09
Revert "[native]Add data cache related server operations"

majetideepak:
This reverts commit b73ab7df31e4d969c44fd953e5cb8e36a18eb55b.

Fix deterministic CI failure.
Resolves: https://github.com/prestodb/presto/issues/23156

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/PrestoServerOperations.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoServerOperations.cpp
@@ -14,8 +14,6 @@
 #include "presto_cpp/main/PrestoServerOperations.h"
 #include <velox/common/base/Exceptions.h>
 #include <velox/common/base/VeloxException.h>
-#include <velox/common/caching/AsyncDataCache.h>
-#include <velox/common/caching/SsdCache.h>
 #include <velox/common/process/TraceContext.h>
 #include "presto_cpp/main/PrestoServer.h"
 #include "presto_cpp/main/ServerOperation.h"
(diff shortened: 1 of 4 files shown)


xiaoxmeng (changes requested):
@majetideepak do you know what's the failure of the test? I think the server operation change is unrelated to the failed test. I am on vocation but you can also ask @zacw7 to help fix this. Thanks!

zacw7 (changes requested):
This change itself doesn't have issue. It just includes a breaking change in Velox: https://github.com/facebookincubator/velox/pull/10261

tdcmeehan:
@zacw7 @xiaoxmeng unfortunately, this commit included *both* an update to Velox and the server operations endpoint.  So to avoid keeping trunk in a broken state, let's revert, then re-add without the Velox bump.

@yingsu00 please note the above, can you please address?  Thanks.

tdcmeehan (approved):

xiaoxmeng:
> @zacw7 @xiaoxmeng unfortunately, this commit included _both_ an update to Velox and the server operations endpoint. So to avoid keeping trunk in a broken state, let's revert, then re-add without the Velox bump.
> 
> @yingsu00 please note the above, can you please address? Thanks.

@tdcmeehan since we know the issue, why not make a forward fix from Velox side?

zacw7:
This change and the advanced Velox version are already being released to prod. Instead of reverting a whole bunch of change together, is it possible to revert the one breaking change in Velox if we really want to fix the broken trunk ASAP to minimize the disruption.

tdcmeehan:
It's fine to revert in Velox first, then advance Velox here.  But I have a question @zacw7: if we were to revert this PR here, are you suggesting this might affect your production deployment?

majetideepak:
I am closing this in favor of advancing Velox.

zacw7:
> It's fine to revert in Velox first, then advance Velox here. But I have a question @zacw7: if we were to revert this PR here, are you suggesting this might affect your production deployment?

It doesn't affect deployment directly. But this is already being rolled out to prod so we would prefer not have it reverted.

Issue #22728, opened 2024-05-13
IN (VALUES ..) should be translated as simple IN LIST


kaikalur:
SELECT * FROM nation where nationkey IN ( values 1,2) 

generates a SemiJoin (like it should!). But this can be a simple IN list! Could be due to some tools not understanding you can do IN for complex types (like rows)

PR #23079, opened 2024-06-25
Add release notes for 0.288

wanglinsong:
# Missing Release Notes
## Ajay Gupte
- [ ] https://github.com/prestodb/presto/pull/22631 BEFORE syntax support for presto engine. (Merged by: Ajay Gupte)

## Amit Dutta
- [ ] https://github.com/prestodb/presto/pull/22968 [native] Remove unused include. (Merged by: Amit Dutta)
- [ ] https://github.com/prestodb/presto/pull/22963 [native] Minor cosmetic fixes. (Merged by: Amit Dutta)
- [ ] https://github.com/prestodb/presto/pull/22767 [native] Advance velox. (Merged by: Amit Dutta)
- [ ] https://github.com/prestodb/presto/pull/22692 [native] Advance velox. (Merged by: Amit Dutta)
- [ ] https://github.com/prestodb/presto/pull/22662 [native] Move static functions from PrestoTask to Util. (Merged by: Amit Dutta)
- [ ] https://github.com/prestodb/presto/pull/22578 [native] Advance velox and remove an unused variable. (Merged by: Amit Dutta)

## Christian Zentgraf
- [ ] d2b2a11f31d8e04ac108572659343242c13accb2 [native] Update dockerfiles to use Centos9
- [ ] 8374c2564cf913d362ed8561a3392ebc2108d373 [native] Advance Velox

## Deepak Majeti
- [ ] https://github.com/prestodb/presto/pull/22815 [native] Make node.id optional (Merged by: Aditi Pandit)
- [ ] https://github.com/prestodb/presto/pull/21929 [native] Add dockerfiles for dependencies and runtime and remove existing dockerfiles (Merged by: Timothy Meehan)

## Emanuel F
- [ ] https://github.com/prestodb/presto/pull/22930 Add table for users to quickly navigate documentation (Merged by: Steve Burnett)

## Jialiang Tan
- [ ] https://github.com/prestodb/presto/pull/22751 [native] Remove spill stats reporting from presto native (Merged by: tanjialiang)
- [ ] https://github.com/prestodb/presto/pull/22720 [native] Migrate away stats reporting for allocator and cache (Merged by: tanjialiang)

## Jimmy Lu
- [ ] https://github.com/prestodb/presto/pull/22922 [native] Transfer idle HTTPSession between IO threads in exchange source (Merged by: Xiaoxuan)

## Karteekmurthys
[...]

PR #23161, opened 2024-07-10
Transform in values to in filter

jackychen718:
## Description
IN (VALUES ...) should be translated into simple IN LIST

## Motivation and Context
The issue comes from #22728 IN (VALUES ..) should be translated as simple IN LIST
If VALUES table only has one column, transformed it into IN LIST

## Impact
Add another optimization rule: 
If filteringSource of SemiJoinNode comes from a ValueNode with only one column, transform it into a ProjectNode with outputVariable of the SemiJoinNode assigned to the IN LIST predicate. The outputVariable will be filtered in a later stage. 

## Test Plan
Test the rule fires on in the scenario and returns correct query plan.

## Contributor checklist

- [ ] CI passed.

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/TransformInValuesToInFilter.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/TransformInValuesToInFilter.java
@@ -1,0 +1,97 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.sql.planner.iterative.rule;
+
+import com.facebook.presto.Session;
+import com.facebook.presto.matching.Captures;
+import com.facebook.presto.matching.Pattern;
+import com.facebook.presto.spi.plan.Assignments;
+import com.facebook.presto.spi.plan.PlanNode;
+import com.facebook.presto.spi.plan.ProjectNode;
+import com.facebook.presto.spi.plan.ValuesNode;
+import com.facebook.presto.spi.relation.RowExpression;
+import com.facebook.presto.sql.planner.iterative.Rule;
+import com.facebook.presto.sql.planner.plan.SemiJoinNode;
+import com.google.common.collect.ImmutableList;
+
+import static com.facebook.presto.SystemSessionProperties.isTransformInValuesToInFilterEnabled;
+import static com.facebook.presto.common.type.BooleanType.BOOLEAN;
+import static com.facebook.presto.spi.relation.SpecialFormExpression.Form.IN;
+import static com.facebook.presto.sql.planner.plan.AssignmentUtils.identityAssignments;
+import static com.facebook.presto.sql.planner.plan.Patterns.filteringSource;
+import static com.facebook.presto.sql.planner.plan.Patterns.project;
+import static com.facebook.presto.sql.planner.plan.Patterns.semiJoin;
+import static com.facebook.presto.sql.planner.plan.Patterns.source;
+import static com.facebook.presto.sql.planner.plan.Patterns.values;
+import static com.facebook.presto.sql.relational.Expressions.specialForm;
+
+/**
+ * This optimizer looks for SemiJoinNode whose filteringSource has only one column, then transform the SemijoinNode into ProjectNode with predicate variable for filtering.
+ * <p/>
+ * Plan before optimizer:
+ * <pre>
+ * SemiJoinNode (semiJoinOutput variable c):
+ *   - source (sourceJoinVariable a)
+ *   - filteringSource (one column variable b)
+ *       - ProjectNode
+ *            - ValuesNode
+ * </pre>
+ * <p/>
+ * Plan after optimizer:
+ * <pre>
+ * ProjectNode:
+ *   - source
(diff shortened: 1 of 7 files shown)


elharo (commented):
Does this resolve https://github.com/prestodb/presto/issues/22728?

jackychen718:
Yes. It resolves #22728.

elharo (dismissed):
If you say

fixes #22728 

in the PR description then Github autocloses the bug when the PR is merged.

jackychen718:
Thanks for your advice.

steveburnett:
If this doesn't need a release note - and I don't believe it does - please add this so this PR isn't in the "Missing Release Notes" section of the release notes process (see #23079 for an example). 

## Release Notes
```
== NO RELEASE NOTE ==
```

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/TransformInValuesToInFilter.java
```diff
@@ -0,0 +1,97 @@
...
+ * ProjectNode:
+ *   - source
+ *   - assignments(identityAssignments of source output,c in b)
+ * </pre>
+ */
+public class TransformInValuesToInFilter
+        implements Rule<SemiJoinNode>
+{
+    private static final Pattern<SemiJoinNode> PATTERN = semiJoin().with(filteringSource()
+            .matching(project().with(source()
```
feilong-liu, line 62: It's possible that there are multiple projects between values and semi join nodes

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/TransformInValuesToInFilter.java
```diff
@@ -0,0 +1,97 @@
...
+    public boolean isEnabled(Session session)
+    {
+        return isTransformInValuesToInFilterEnabled(session);
+    }
+
+    @Override
+    public Result apply(SemiJoinNode semiJoinNode, Captures captures, Context context)
+    {
+        PlanNode source = semiJoinNode.getSource();
+        return context.getLookup().resolveGroup(semiJoinNode.getFilteringSource()).findFirst()
```
feilong-liu, line 81: Having the whole logic within the single return statement may not be a good idea.

presto-main/src/main/java/com/facebook/presto/sql/planner/plan/Patterns.java
```diff
@@ -227,6 +227,11 @@ public static Property<PlanNode, PlanNode> source()
                 Optional.empty());
     }
 
+    public static Property<SemiJoinNode, PlanNode> filteringSource()
```
feilong-liu, line 230: Put it in a statics class "SemiJoin" like other node specific properties below

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/TransformInValuesToInFilter.java
kaikalur: Yeah makes it hard to read

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/TransformInValuesToInFilter.java
```diff
@@ -0,0 +1,97 @@
...
+                .flatMap(projectNode -> context.getLookup().resolveGroup(projectNode.getSources().get(0)).findFirst())
+                .map(ValuesNode.class::cast)
+                .map(ValuesNode::getRows)
+                // check that all values are only a single row expression (no struct/row types)
```
kaikalur, line 85: Hmm why not struct/row? How about array tyoes?

kaikalur (changes requested):
Also add an e2e test

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/TransformInValuesToInFilter.java
jackychen718: yes. I see that, in previous version, it has more than one projects. However, in current version, it has only one project there.Do you have some ideas how can I match multiple levels to catch the ValuesNode?

steveburnett:
Nit of formatting the release note entry to use \` and not ', and add a space between the session property and the PR number.

```
== RELEASE NOTES ==

General Changes
* Add an optimization rule that transform key IN (VALUES ...) form into IN LIST form if VALUES table only has one column. The optimization is controlled by session property `transform_in_values_to_in_filter` :pr:`23161`
```

presto-main/src/main/java/com/facebook/presto/sql/planner/plan/Patterns.java
jackychen718: fixed

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/TransformInValuesToInFilter.java
kaikalur: Not sure about the pattern but values->project* should still be constant? Maybe we have a separate inline constant values rule that'll be generally useful and this rule. So if you have something like:

VALUES('abcd')->Project(subtr(field_0, 1, 2)) 

you can rewrite that as:

VALUES('ab')

Using the RowExpressionInterpreter?

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/TransformInValuesToInFilter.java
jackychen718: In fact the apply function works for struct/row and array. The only problem is how to math the pattern of variable projects between values and semi join nodes.

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/TransformInValuesToInFilter.java
kaikalur: #23196

Issue #23165, opened 2024-07-10
Decouple presto-verifier from Hive and other connectors


elharo:
presto-verifier depends on various connectors, especially Hive but also MySQL; mostly through the use of HiveErrorCode and similar in PrestoExceptionClassifier. This causes problems for connectors not in this repo like Prism.

Break this coupling so that presto-verifier no longer depends on specific connectors.

PR #23168, opened 2024-07-10
Decouple from HiveErrorCode.HIVE_TOO_MANY_OPEN_PARTITIONS

elharo:
Part of #23165

### DIFF
--- a/presto-verifier/pom.xml
+++ b/presto-verifier/pom.xml
@@ -31,17 +31,6 @@
 
         <dependency>
             <groupId>com.facebook.presto</groupId>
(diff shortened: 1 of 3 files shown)


amitkdutta (dismissed):
Looks good. Thanks @elharo

abhiseksaikia (approved):
Thanks for the fix! I guess a long term change could be to provide an option to register additional FailureResolvers which are connector specific and remove the dependency on presto-hive-* from presto-verifier.  

Please fix the checkstyle violation

NikhilCollooru (approved):

PR #23172, opened 2024-07-10
Refactor and Modify IcebergWritableTableHandle

imjalpreet:
## Description

This PR includes the below changes as part of the effort to support Iceberg writes with Presto C++:
- [x] Add compression codec in IcebergWritableTableHandle: We need to send compression codec to Velox as part of IcebergWritableTableHandle.
- [x] Add IcebergInsertTableHandle and IcebergOutputTableHandle: Presto C++ protocol requires separate implementations for ConnectorInsertTableHandle and ConnectorOutputTableHandle

## Motivation and Context

As part of the effort to support Iceberg Writes with Presto C++

## Impact

No user impact

## Contributor checklist

- [x] CI passed.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergOutputTableHandle.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergOutputTableHandle.java
@@ -1,0 +1,51 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.iceberg;
+
+import com.facebook.presto.hive.HiveCompressionCodec;
+import com.facebook.presto.spi.ConnectorOutputTableHandle;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.List;
+import java.util.Map;
+
+public class IcebergOutputTableHandle
+        extends IcebergWritableTableHandle
+        implements ConnectorOutputTableHandle
+{
+    @JsonCreator
+    public IcebergOutputTableHandle(
+            @JsonProperty("schemaName") String schemaName,
+            @JsonProperty("tableName") IcebergTableName tableName,
+            @JsonProperty("schemaAsJson") String schemaAsJson,
+            @JsonProperty("partitionSpecAsJson") String partitionSpecAsJson,
+            @JsonProperty("inputColumns") List<IcebergColumnHandle> inputColumns,
+            @JsonProperty("outputPath") String outputPath,
+            @JsonProperty("fileFormat") FileFormat fileFormat,
+            @JsonProperty("compressionCodec") HiveCompressionCodec compressionCodec,
+            @JsonProperty("storageProperties") Map<String, String> storageProperties)
+    {
+        super(
+                schemaName,
+                tableName,
+                schemaAsJson,
+                partitionSpecAsJson,
+                inputColumns,
+                outputPath,
+                fileFormat,
+                compressionCodec,
+                storageProperties);
+    }
+}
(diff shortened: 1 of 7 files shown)


presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergOutputTableHandle.java
```diff
@@ -0,0 +1,51 @@
...
+import com.facebook.presto.hive.HiveCompressionCodec;
+import com.facebook.presto.spi.ConnectorOutputTableHandle;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.List;
+import java.util.Map;
+
+public class IcebergOutputTableHandle
+        extends IcebergWritableTableHandle
```
ZacBlanco, line 25: Wondering if it's better to subclass or rather just keep a field of type `IcebergWritableTableHandle`?

I think in general, it's better to just make this a field unless you think there is a strong benefit to subclassing. Using a field should also make generating with the native protos easier, as I believe there's some extra metadata you need to add if there are subclasses?

ZacBlanco (commented):
I would like to know why velox can't support the same type for insert and output? It seems like an arbitrary restriction that could be fixed in velox.

Otherwise, I think it would actually make proto generation easier if we don't subclass. I could be wrong though as I haven't looked at the proto generation procedures recently

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergOutputTableHandle.java
imjalpreet: > I would like to know why velox can't support the same type for insert and output? It seems like an arbitrary restriction that could be fixed in velox.

So, the limitation currently is in protocol generation, if we use the same class type for multiple abstract classes, one of them gets overridden by the other during generation. In the current protocol generation, it is not able to use multiple parent classes. I also wanted to avoid this change, and I tried a few approaches to make it work with the same class but there were issues in Prestissimo. It looked like a bigger change to support this in Prestissimo so I moved to this approach. 

I don't currently have all the issues faced so I will try it again and add the issues here for your reference.

imjalpreet: > Wondering if it's better to subclass or rather just keep a field of type IcebergWritableTableHandle?

Sorry, to confirm, are you saying we should add a field of type IcebergWritableTableHandle in IcebergOutputTableHandle(and IcebergInsertTableHandle) rather than subclassing? Please correct me in case you meant something different.

ZacBlanco: > we should add a field of type IcebergWritableTableHandle in IcebergOutputTableHandle(and IcebergInsertTableHandle) rather than subclassing

Your understanding is correct

hantangwangd (commented):
The change makes sense considering of the limitation in protocol generation as you described above. LGTM!

Besides, maybe we should also modify the code in java in a subsequent PR to uniformly use the compression codec maintained in `IcebergWritableTableHandle` after this change.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergOutputTableHandle.java
imjalpreet: @ZacBlanco I also wanted to understand what advantages you see if we include a field rather than making these two a subclass of `IcebergWritableTableHandle`. In the current implementation, protocol generation would be independent of `IcebergWritableTableHandle` since we don't want the hierarchy in C++ but if we add it as a field we would need to include it in dependent classes as well. I don't mind making this change but we would need to do additional refactoring on the java side in the classes `IcebergAbstractMetadata` and `IcebergPageSinkProvider` if I don't subclass.

Let me know your views and based on that we can finalise the approach.

ZacBlanco: It's basically a design decision of [Composition or Inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance).

In my experience designing protocols/data definitions for serialized objects, it's most common to use composition over of inheritance to represent serializable classes. It tends to be cleaner and easier to implement. Most DDLs such as protobuf, openapi, etc tend to not support polymorphic structures.

Not that we'll move off of our esoteric java class parsing-based protocol generation any time soon, but I think composition tends to be a better for wire protocols. If we were to ever move to something like openapi/protobuf in the future one of the first challenges will probably be how to get rid of all polymporphic structures in serializable classes.

Anyways, to prevent additional refactors for this PR, I am fine to keep it as is. Just some food for thought in the future.

ZacBlanco (approved):

ZacBlanco (approved):

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergOutputTableHandle.java
imjalpreet: @ZacBlanco thanks for sharing this. I will read and look into this and see how we can improve our current implementation and we can work on refactoring it if it helps improve the code quality.

PR #23173, opened 2024-07-10
Add wrappers for Iceberg Schema and Iceberg PartitionSpec

imjalpreet:
## Description

As part of an ongoing effort to support iceberg writes in Presto C++, since we don't have iceberg C++ library support to parse the JSON strings for Schema and PartitionSpec, we will send the respective Presto wrapper classes as part of the IcebergInsertTableHandle and InsertOutputTableHandle

Depends on #23172 

## Motivation and Context

To prevent re-implementing JSON parsers for Iceberg Schema and Iceberg PartitionSpec in Presto C++ and eliminate the need to maintain the code with a fast-paced iceberg library, we have created wrapper classes and sending them instead of the JSON strings to workers during writes.

## Impact

No user impact

## Test Plan

CI build pipelines and manually verified writing partitioned and non-partitioned iceberg tables using Java Workers to ensure no impact.

## Contributor checklist

- [x] CI passed.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/PartitionFields.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/PartitionFields.java
@@ -18,7 +18,10 @@ import org.apache.iceberg.PartitionSpec;
 import org.apache.iceberg.Schema;
 import org.apache.iceberg.expressions.Term;
 
+import javax.annotation.Nullable;
+
 import java.util.List;
+import java.util.Optional;
 import java.util.function.Consumer;
 import java.util.regex.MatchResult;
 import java.util.regex.Matcher;
@@ -58,7 +61,15 @@ public final class PartitionFields
 
     public static PartitionSpec parsePartitionFields(Schema schema, List<String> fields)
     {
-        PartitionSpec.Builder builder = PartitionSpec.builderFor(schema);
+        return parsePartitionFields(schema, fields, null);
+    }
+
+    public static PartitionSpec parsePartitionFields(Schema schema, List<String> fields, @Nullable Integer specId)
+    {
+        PartitionSpec.Builder builder = Optional.ofNullable(specId)
+                .map(id -> PartitionSpec.builderFor(schema).withSpecId(id))
+                .orElseGet(() -> PartitionSpec.builderFor(schema));
+
         for (String field : fields) {
             parsePartitionField(builder, field);
         }
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/PrestoIcebergNestedField.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/PrestoIcebergNestedField.java
@@ -1,0 +1,101 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.iceberg;
+
+import com.facebook.presto.common.type.Type;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.Objects;
+import java.util.Optional;
+
+import static java.util.Objects.requireNonNull;
+
+public class PrestoIcebergNestedField
+{
+    private final boolean optional;
+    private final int id;
+    private final String name;
+    private final Type prestoType;
+    private final Optional<String> doc;
+
+    @JsonCreator
+    public PrestoIcebergNestedField(
+            @JsonProperty("optional") boolean optional,
+            @JsonProperty("id") int id,
+            @JsonProperty("name") String name,
+            @JsonProperty("prestoType") Type prestoType,
+            @JsonProperty("doc") Optional<String> doc)
+    {
+        this.optional = optional;
+        this.id = id;
+        this.name = requireNonNull(name, "name is null");
+        this.prestoType = requireNonNull(prestoType, "prestoType is null");
+        this.doc = requireNonNull(doc, "doc is null");
+    }
+
+    @JsonProperty
+    public boolean isOptional()
+    {
+        return optional;
+    }
+
+    @JsonProperty
+    public int getId()
+    {
+        return id;
+    }
+
+    @JsonProperty
+    public String getName()
+    {
+        return name;
+    }
+
+    @JsonProperty
+    public Type getPrestoType()
+    {
+        return prestoType;
+    }
+
+    @JsonProperty
+    public Optional<String> getDoc()
+    {
+        return doc;
+    }
+
+    @Override
+    public int hashCode()
+    {
+        return Objects.hash(optional, id, name, prestoType, doc);
+    }
+
+    @Override
+    public boolean equals(Object obj)
+    {
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null || getClass() != obj.getClass()) {
+            return false;
+        }
+
+        PrestoIcebergNestedField other = (PrestoIcebergNestedField) obj;
+        return Objects.equals(this.optional, other.optional) &&
+                Objects.equals(this.id, other.id) &&
+                Objects.equals(this.name, other.name) &&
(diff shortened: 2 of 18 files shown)


presto-iceberg/src/main/java/com/facebook/presto/iceberg/PrestoIcebergNestedField.java
```diff
@@ -0,0 +1,74 @@
...
+{
+    private final String isOptional;
+    private final int id;
+    private final String name;
+    private final Type prestoType;
+    private final String doc;
+
+    @JsonCreator
+    public PrestoIcebergNestedField(
+            @JsonProperty("isOptional") String isOptional,
```
ZacBlanco, line 32: why is this a string and not boolean?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/PrestoIcebergNestedField.java
```diff
@@ -0,0 +1,74 @@
...
+            @JsonProperty("id") int id,
+            @JsonProperty("name") String name,
+            @JsonProperty("prestoType") Type prestoType,
+            @JsonProperty("doc") String doc)
+    {
+        this.isOptional = requireNonNull(isOptional, "isOptional is null");
+        this.id = id;
+        this.name = requireNonNull(name, "name is null");
+        this.prestoType = requireNonNull(prestoType, "prestoType is null");
+        this.doc = doc;
```
ZacBlanco, line 42: requireNonNull? If this can be null, I would make the field an `Optional`

presto-iceberg/src/main/java/com/facebook/presto/iceberg/PrestoIcebergNestedField.java
imjalpreet: Yes, this was temporary, I was trying out something. For some reason, when taskUpdateRequest is getting serialized on the coordinator and then deserialized on the worker, the true boolean value becomes false after deserialization. 

To see why that is happening:

https://github.com/prestodb/presto/blob/c254d194e5151d1b01ee276001372f0f229191f1/presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTask.java#L859

This will be part of updateRequest.writeInfo. Before this object is serialized it has the correct value of true for `isOptional` but after the serialization, if I am trying to deserialize it using `taskUpdateRequestCodec.fromBytes(taskUpdateRequestJson)` the object returned has a value of false for `isOptional`. 

Now, when I rename the field `isOptional` to just `optional`, it can serialize and deserialize correctly. 

I will try to understand how the serialization/deserialization is affected by the name of a boolean field. In case you have an idea, please let me know. 

In any case, I will rename the field in this PR.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/PrestoIcebergNestedField.java
imjalpreet: I will make it Optional, it can be null based on the iceberg library.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/PrestoIcebergNestedField.java
ZacBlanco: I think renaming to avoid the `is` collision should be fine. I think you are seeing serialization issues because the json property here is defined as `isOptional` but your `@JsonProperty` method for the field is `isOptional`, where it probably should be `getIsOptional` or `isIsOptional` (I don't know which is correct). Another option is to keep the method name and change the `@JsonProperty` annotation to be `@JsonProperty("isOptional")`.

I think it's just best to avoid `is` in the property name anyways to prevent future confusion.

imjalpreet: Yes, since this was a wrapper class for one of the iceberg library class NestedField, I was trying to keep the field names the same, and the field name in the library was isOptional. As you pointed out, I think the issue would be the getter method's name due to which it is not able to pick the correct value. 

I have anyways renamed the field to `optional` now since that looks better.

ZacBlanco (commented):
LGTM for now. I'll take a final pass once the table handle PR merges

imjalpreet:
I have rebased it with the latest master, it's ready for further review.

hantangwangd (commented):
Overall looks good to me. Would you mind adding a test case to show that the core fields are the same as original ones after experiencing the entire conversion path as follows:

```
PartitionSpec(including a Schema) -> IcebergPartitionSpec -> json string -> IcebergPartitionSpec -> PartitionSpec
```

presto-iceberg/src/main/java/com/facebook/presto/iceberg/PartitionFields.java
```diff
@@ -58,7 +59,15 @@ private PartitionFields() {}
 
     public static PartitionSpec parsePartitionFields(Schema schema, List<String> fields)
     {
-        PartitionSpec.Builder builder = PartitionSpec.builderFor(schema);
+        return parsePartitionFields(schema, fields, null);
+    }
+
+    public static PartitionSpec parsePartitionFields(Schema schema, List<String> fields, Integer specId)
```
ZacBlanco, line 65: nit: annotate with `@Nullable`

presto-iceberg/src/main/java/com/facebook/presto/iceberg/TypeConverter.java
```diff
@@ -209,16 +227,55 @@ public static org.apache.iceberg.types.Type fromRow(RowType type, int startId)
...
 
+    public static org.apache.iceberg.types.Type fromRow(
+            RowType type,
+            String columnName,
+            Map<String, Integer> columnNameToIdMapping)
+    {
+        List<Types.NestedField> fields = new ArrayList<>();
+        for (RowType.Field field : type.getFields()) {
+            String name = field.getName().orElseThrow(() ->
+                    new PrestoException(NOT_SUPPORTED, "Row type field does not have a name: " + type.getDisplayName()));
```
ZacBlanco, line 238: Does this mean we can't support anonymous row types?

ZacBlanco (approved):
LGTM one nit, one question

yingsu00 (commented):
@imjalpreet Mostly look good. In addition to what @hantangwangd suggested, could you please add tests to make sure the conversion both ways work as expected and won't throw any exceptions? In the future if the Iceberg lib is updated, these tests will indicate if the conversion is still valid.

yingsu00:
@aditi-pandit Do you also want to review this PR?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/TypeConverter.java
imjalpreet: Yes, although I haven't changed anything as part of this PR, it has been like this.

imjalpreet:
@hantangwangd @yingsu00 Sorry, I was caught up with something else, I have added the tests now.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/TypeConverter.java
aditi-pandit: @imjalpreet : If this error is only for the purpose of converting to Iceberg, then maybe we should change message to "Cannot convert Row type field " + type.getDisplayName() + " to Iceberg"

presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestNestedFieldConverter.java
```diff
@@ -0,0 +1,269 @@
...
+                {1, "decimal"},
+                {1, "varchar"},
+                {1, "varbinary"},
+                {1, "row"},
+                {1, "date"}
+        };
+    }
+
+    @Test(dataProvider = "allTypes")
+    public void testToPrestoNestedField(int id, String name)
```
aditi-pandit, line 71: Also add a negative test case for the anonymous row field type

aditi-pandit (commented):
Change looks good overall, except for a testing related comment.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestNestedFieldConverter.java
hantangwangd: Maybe also add some tests for real nested types, for example:

```
row(a int, b varchar, c array(row(d int)), e map(varchar, row(f varchar)))
```

imjalpreet: @aditi-pandit @hantangwangd I have updated the PR with more tests.

hantangwangd (approved):
Thanks for the fix, lgtm.

aditi-pandit (approved):
Thanks @imjalpreet for iterating over the comments.

PR #22365, opened 2024-03-29
[Iceberg] Add Histogram Statistic Support

ZacBlanco:
## Description

This changes adds support for the histogram statistic type to Iceberg. There are 3 related commits

1. Adds support for collecting a superset of supported statistics in the HMS and Iceberg. This is required to allow histograms to be supported while the HMS is in use.
2. Adds the histogram implementation for Iceberg.
3. Adds documentation for histograms statistics.

## Motivation and Context

Histograms result in better accuracy for output row counts from filter nodes when calculating statistics. This should result in closer estimates in the CBO, resulting in better query plans.

## Impact

1. ANALYZE for Iceberg and Hive will now store statistics in the table's Puffin files.
4. Writes new puffin blob type containing the histogram.

## Test Plan

- New tests to ensure histograms are collected for supported types
- Tests that ensure the histograms result in more accurate plan stat estimates for non-uniform data

## Contributor checklist

- [x] Please make sure your submission complies with our [development](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#development), [formatting](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#formatting), [commit message](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#commit-formatting-and-pull-requests), and [attribution guidelines](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#attribution).
- [x] PR description addresses the issue accurately and concisely.  If the change is non-trivial, a GitHub Issue is referenced.
- [x] Documented new properties (with its default value), SQL syntax, functions, or other functionality.
- [x] If release notes are required, they follow the [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines).
- [x] Adequate tests were added if applicable.
- [x] CI passed.

## Release Notes
Please follow [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines) and fill in the release notes below.

```
== RELEASE NOTES ==

General Changes
* Add support for the histogram statistic type. :pr:`22365`
```

PR #23177, opened 2024-07-10
[Iceberg] Add StatisticsFileCache

ZacBlanco:
## Description

This change adds a connector-wide cache for `StatisticsFile`s. This prevents loading the same statisticsFile multiple times into the connector. It improves query planning times and reduces memory usage by having only one set of statistics available from each file.

The cache is designed to be connector-wide to reduce memory consumption. Since separate queries can reference the same tables, we prevent memory utilization by sharing the cache across queries. The files are immutable and unique to a table and snapshot (hashed and equated on properties of file path, footer size, blob metadata, etc), so caching between queries/users/sessions shouldn't be an issue. Cache entries should never become stale

## Motivation and Context

I noticed after experimenting with histograms in the Iceberg connector that planning was quite slow for two reasons

1. We were always hitting the FS for `StatisticsFile`s
2. Every time we load the blobs from a StatisticFile, even if using the same file as a previous query, we would allocate a new buffer to represent the statistics. These statistics are not evacuated from memory until the query metadata for an execution expires from `query.max-age`. For smaller statistics, this is not really a problem. However,  histograms can be large (KBs to MBs in size). Having the cache can prevent making copies of redundant data, leading to lower memory pressure and smaller chance of choking up the coordinator.

The cache inside of the Iceberg connector does not fully alleviate the issue. If a user has thousands of tables/columns, memory pressure can still occur when there is large amount of historical query metadata in the heap. The only way to prevent the pressure is to eventually store some query metadata (or maybe just statistics?) in a cache which is freed once the coordinator reaches high memory pressure.

## Impact

- Addition to the `ConnectorHistogram` SPI interface to add a `getMemoryUtilization` method. This will be useful in the future when implementing real histograms in #22365 .

## Test Plan

- Added an E2E test to verify cache is hit when table statistics are retrieved.

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Add a configurable-sized cache for Iceberg table puffin files to improve query planning time controlled by the  `iceberg.max-statistics-file-cache-size` configuration property. :pr:`23177`
```


steveburnett (dismissed):
LGTM! (docs) 

Pull branch, local docs build, reviewed compiled page in local build. Looks good, thanks!

ClarenceThreepwood (dismissed):
lgtm overall

hantangwangd (commented):
Overall looks good to me. Bring up one question for discussing, as the Guava cache is a light weight LRU implementation, it may not adapt well to some scenarios. So do you think it's useful to provide a way to configure not to use statistics file cache in some situations, or have a way to extend and configure to use other enhanced cache implementations, for example the ones based on Caffeine or Redis?

presto-iceberg/src/test/java/com/facebook/presto/iceberg/hive/TestIcebergDistributedHive.java
```diff
@@ -59,6 +59,13 @@ public void testPartShowStatsWithFilters()
         // Hive doesn't support returning statistics on partitioned tables
     }
 
+    @Override
+    public void testStatisticsFileCache()
+            throws Exception
+    {
+        // hive doesn't load Iceberg statistics files when metastore is in use
+    }
+
```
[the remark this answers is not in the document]
ZacBlanco: This comment may be slightly wrong. We don't write them but we do load them. I have a commit which changes this behavior in my histogram PR: https://github.com/prestodb/presto/pull/22365/commits/b1751e2efa65efb2edc1fa7d942ef80f78996c88

I intend to re-enable this test in that commit once this PR merges

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergCommonModule.java
```diff
@@ -168,6 +172,22 @@ public void setup(Binder binder)
...
+    @Provides
+    public StatisticsFileCache createStatisticsFileCache(IcebergConfig config, MBeanExporter exporter)
+    {
+        Cache<StatisticsFile, Map<Integer, ColumnStatistics>> delegate = CacheBuilder.newBuilder()
+                .maximumWeight(config.getMaxStatisticsFileCacheSize().toBytes())
+                .<StatisticsFile, Map<Integer, ColumnStatistics>>weigher((key, entry) ->
+                        (int) entry.values().stream().mapToLong(StatisticsUtil::getColumnStatisticsMemoryUtilization)
+                                .sum())
+                .recordStats()
+                .build();
```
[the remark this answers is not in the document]
ZacBlanco: Unfortunately, there isn't much more we can configure with the guava caches unless we allow time or cache entry amount-based evictions. Time-based evictions don't make sense because the statistics files are immutable so the entries will never be stale. If old entries they fill up the cache, then you can just wait until they get evicted by newer ones. I don't see a reason to try to expire early.  Otherwise, having to modify cache size via number of entries makes it much more difficult to properly size the coordinator heap. So, I think this is best approach of the available eviction options.

I am open to ideas which gives users more flexibility. However, I think if we move towards a cache key implementation which is more granular then there I don't see much more value in configuring other parts of the cache.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/TableStatisticsMaker.java
```diff
@@ -497,34 +512,40 @@ private Optional<StatisticsFile> getClosestStatisticsFileForSnapshot(IcebergTabl
     /**
      * Builds a map of field ID to ColumnStatistics for a particular {@link StatisticsFile}.
      */
-    private Map<Integer, ColumnStatistics.Builder> loadStatisticsFile(StatisticsFile file)
+    private Map<Integer, ColumnStatistics> loadStatisticsFile(StatisticsFile file, StatisticsFileCache statisticsFileCache)
```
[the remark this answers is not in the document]
ZacBlanco: I think you have a good point. I will consider creating a separate cache key which includes the table and columns as well. I think the StatisticsFile still needs to be in the cache key because different snapshots can reference different files. Open to more ideas for the cache key as well.

presto-main/src/main/java/com/facebook/presto/cost/DisjointRangeDomainHistogram.java
```diff
@@ -68,6 +69,9 @@
 public class DisjointRangeDomainHistogram
         implements ConnectorHistogram
 {
+    private static final long INSTANCE_SIZE = ClassLayout.parseClass(DisjointRangeDomainHistogram.class).instanceSize();
+    private static final long RANGE_SIZE = ClassLayout.parseClass(Range.class).instanceSize();
```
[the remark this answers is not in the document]
ZacBlanco: Technically not entirely accurate, but there isn't a great way to measure it. The inner fields of `Range` are of type `Cut<C>` which is not publicly accessible, so we cannot call `ClassLayout.parseClass` on it.

Anyways, this will actually be updated in the histogram PR (#22365) after this merges because one of the commits in that PR replaces the `Range` class here instead with Presto's `Range` instead. I would opt to leave this as is for now anyways because of that PR and also the fact that this class usually has very little contribution to memory except for cases when users have large amounts of filters in a query (e.g. 10k values in an `IN` clause)

presto-iceberg/src/test/java/com/facebook/presto/iceberg/hive/TestIcebergDistributedHive.java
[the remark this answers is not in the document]
hantangwangd: Sure, got it.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergCommonModule.java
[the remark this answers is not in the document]
ZacBlanco: I don't think there's strong precedent within the codebase generally to support user-provided cache implementations. In general Presto makes heavy use of guava caches, but little of the exposed to the user except for some of the removal properties (size/number of entries/weight). There are at least 82 references to guava's `Cache.newBuilder` in the codebase from a quick glance.

If we had a better framework for providing a custom cache interface already, I would not be as opposed. However, It is probably a big chunk of work just support it within Iceberg. I would happily support a user-provided cache implementation in a later PR though. Let's create an issue to track it? It could be good issue to pick up for a future contributor.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/TableStatisticsMaker.java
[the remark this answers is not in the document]
ZacBlanco: The latest commit now uses a new cache key which is based off of the file and column ID. I didn't think table name was necessary since each file is specific to a table anyways. The logic for `loadStatisticsFile` has increased in complexity though. I would look more closely at that method during the next review.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergCommonModule.java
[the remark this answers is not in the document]
ZacBlanco: Also re-disabling cache: I would just set the new cache size configuration to 0B and it will effectively disable caching. One entry might get stored at a time, but for all intents and purposes the cache will almost always return null, causing the cache-miss logic to be executed.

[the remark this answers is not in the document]
hantangwangd: > I would happily support a user-provided cache implementation in a later PR though. Let's create an issue to track it? It could be good issue to pick up for a future contributor.

Sure, we can go back here and think about this issue if we encounter problems in production environment.

> I would just set the new cache size configuration to 0B and it will effectively disable caching.

OK, that would be fine.

hantangwangd (commented):
Thanks for the fix, lgtm.

hantangwangd (approved):

presto-iceberg/src/main/java/com/facebook/presto/iceberg/TableStatisticsMaker.java
```diff
@@ -497,8 +516,19 @@ private Optional<StatisticsFile> getClosestStatisticsFileForSnapshot(IcebergTabl
     /**
      * Builds a map of field ID to ColumnStatistics for a particular {@link StatisticsFile}.
      */
-    private Map<Integer, ColumnStatistics.Builder> loadStatisticsFile(StatisticsFile file)
+    private Map<Integer, ColumnStatistics> loadStatisticsFile(StatisticsFile file, StatisticsFileCache statisticsFileCache, List<Integer> columnIds)
     {
+        // first, try to load all stats from the cache. If the map doesn't contain all keys, load the missing
```
aaneja, line 521: Do you have an estimate of how big the puffin file gets for an Iceberg table (with say, 10s of columns) ?  

I ask because, if the file size is small enough (and given the fact that Puffin supports [compression](https://iceberg.apache.org/puffin-spec/#compression-codecs) the number of columns may not be an issue), we may be better off caching the entire puffin file in memory instead of caching at a per-column level (similar to how we cache parquet footers)

With the current implementation, repeated file IO will be performed every time we want to access a new column's stats.
This will be avoided if cache the entire `StatisticsFile`'s byte-stream instead

ZacBlanco: > Do you have an estimate of how big the puffin file gets for an Iceberg table (with say, 10s of columns) ?

It depends on histogram configuration, but using TPC-DS SF1k with the histogram K parameter configured to 32768, each histogram consumes ~1MiB of space. So for a table with up to 100 columns which all support histograms, you'd use 100MB. The smallest configuration takes ~1KiB. If you have a large schema with many tables, it's conceivable that you could easily reach the cache limits if you stored all the columns. I think it makes sense to cache the used columns.

Also, I don't have data on the Theta sketches which we use to store NDV counts, but I believe the sketch can also consumes in the range of ~KiBs but I haven't done any thorough experimentation.

I don't have data on compression of the Puffin files. We don't currently compress the files in our writer. I think adding compression is trivial though. From an information-theoretic perspective, sketches are supposed to be information-dense which would lead me to believe that compression won't affect the size much. I need to do some experimentation though. I will update later with some results.

The difference between puffin and the parquet footers is going to be that the amount of data stored per-column is potentially much larger. I felt re-reading the puffin files for new columns was a reasonable tradeoff in order to not pollute the cache with unused data which is especially important for wide tables.

aaneja: I'm OK with starting with this fine grained cache and measuring usage. 

Specifically can we track a metric with  - 
1. Metric Value :  Size of puffin file fetched
2. Dimension : Number of columns with stats in the file
3. Dimension : File name/ File name MD5
4. Dimension : Was an existing column's stats cached aka have we read this file before (This one may be hard to track given the cache key choice)

This metric, along with cache hit/miss counts should give us a good picture of if we need to tweak what we cache

ZacBlanco: I'm somewhat familiar with our JMX interfaces, but I'm not sure it's capable of tracking these metrics along all the  dimensions. I can record separate independent distributions for the puffin file sizes, columns amount, etc. I don't know if we have the ability to collect these so we can correlate them. Do you know of any existing examples I could look at which does something similar?

aaneja: You can log a `RuntimeMetric` using `ConnectorSession#getRuntimeStats()`, [an example](https://github.com/prestodb/presto/blob/06edc73cca2f83916e8a18d371bba56e6b32f93f/presto-hive/src/main/java/com/facebook/presto/hive/statistics/ParquetQuickStatsBuilder.java#L364-L365).  

Since we don't have a way to track dimensions on the metric, we could model the dimensions in the metric name, e.g
```
session.getRuntimeStats().addMetricValue("<TableOrPartitionIdentifier>/PuffinFileSize", RuntimeUnit.BYTE, 100_000L);
session.getRuntimeStats().addMetricValue("<TableOrPartitionIdentifier>/<FileNameOrMD5>/ColumnCount", RuntimeUnit.NONE, 10L);
session.getRuntimeStats().addMetricValue("<TableOrPartitionIdentifier>/CacheMissButOtherColumnStatsPresent", RuntimeUnit.NONE, 1L);
```

ZacBlanco: Thanks for the pointers. I added metrics for file size and column count, as well as partial cache misses.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/TableStatisticsMaker.java
```diff
@@ -497,8 +527,38 @@ private Optional<StatisticsFile> getClosestStatisticsFileForSnapshot(IcebergTabl
...
+            session.getRuntimeStats().addMetricValue(
+                    String.format(STATISITCS_CACHE_METRIC_PARTIAL_MISS_FORMAT,
+                            tableHandle.getSchemaTableName(),
+                            murmur3_128().hashString(file.path(), UTF_8)),
+                    RuntimeUnit.NANO, 1);
+        }
+
+        String fileSizeMetricName = String.format(STATISITCS_CACHE_METRIC_FILE_SIZE_FORMAT,
+                tableHandle.getSchemaTableName(),
+                murmur3_128().hashString(file.path(), UTF_8));
```
aaneja, line 552: Isn't the file path unique for every stats file ? Also, does it always include the table name ? If so, we can skip hashing the `file.path()` and also skip the table name for this metric

presto-iceberg/src/main/java/com/facebook/presto/iceberg/TableStatisticsMaker.java
```diff
@@ -497,8 +527,38 @@ private Optional<StatisticsFile> getClosestStatisticsFileForSnapshot(IcebergTabl
...
+        Set<Integer> missingStats = columnIds.stream().filter(id -> !cachedStats.containsKey(id)).collect(toImmutableSet());
+        if (missingStats.isEmpty()) {
+            return cachedStats;
+        }
+        if (!cachedStats.isEmpty()) {
+            session.getRuntimeStats().addMetricValue(
+                    String.format(STATISITCS_CACHE_METRIC_PARTIAL_MISS_FORMAT,
+                            tableHandle.getSchemaTableName(),
+                            murmur3_128().hashString(file.path(), UTF_8)),
+                    RuntimeUnit.NANO, 1);
```
aaneja, line 547: typo: Should be `RuntimeUnit.NONE`

presto-iceberg/src/main/java/com/facebook/presto/iceberg/statistics/StatisticsFileCache.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+package com.facebook.presto.iceberg.statistics;
+
+import com.facebook.presto.spi.statistics.ColumnStatistics;
+import com.google.common.cache.Cache;
+import com.google.common.cache.ForwardingCache.SimpleForwardingCache;
+
+public class StatisticsFileCache
+        extends SimpleForwardingCache<StatisticsFileCacheKey, ColumnStatistics>
+{
```
aaneja, line 22: Can you add cache metrics ? We do this e.g in [CachingDirectoryLister](https://github.com/prestodb/presto/blob/e2ec28f6ebca39357bed856441f66aebce9b2667/presto-hive/src/main/java/com/facebook/presto/hive/CachingDirectoryLister.java#L159-L205).

aaneja (commented):
Could you please add DistibutionStat's  for the RuntimeMetric's ? 
- These would be super helpful for users who do not have metric monitoring in place
- They provide a simpler picture of overall distribution without needing a deep dive

steveburnett:
Nit, suggest a minor revision to release note entry to label this as a configuration property. 

```
== RELEASE NOTES ==

General Changes
* Add a configurable-sized cache for Iceberg table puffin files to improve query planning time controlled by the  `iceberg.max-statistics-file-cache-size` configuration property. :pr:`23177` 
```

presto-iceberg/src/main/java/com/facebook/presto/iceberg/statistics/StatisticsFileCache.java
ZacBlanco: We already expose cache metrics? It's done here: https://github.com/prestodb/presto/blob/961f4beb44d853ae51b3152b72d4b4496f25857f/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergCommonModule.java#L183-L184

presto-iceberg/src/main/java/com/facebook/presto/iceberg/TableStatisticsMaker.java
ZacBlanco: The stats files are always unique, but it is not guaranteed that the path always includes the table name. People can configure the storage directory however they like which means it may or may not include the table name. Generally with default settings, the table names exist in the path though

aaneja: Okay lets include the table name in the metric :+1: 
My reasoning behind having the full file path in the metric name was to help users identify which stats files are large in size. Since the path is unique, should we skip hashing it altogether ?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/statistics/StatisticsFileCache.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    public void recordFileSize(long size)
+    {
+        fileSizes.add(size);
+    }
+
+    @Managed
+    @Nested
+    public DistributionStat getColumnCountDistribution()
```
aaneja, line 48: TY for adding these!

aaneja (dismissed):
LGTM % one question

hantangwangd (approved):

presto-iceberg/src/main/java/com/facebook/presto/iceberg/TableStatisticsMaker.java
ZacBlanco: Updated to remove the hash

aaneja (approved):

hantangwangd (approved):

ZacBlanco:
@tdcmeehan could you take a look and give a stamp?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/statistics/StatisticsFileCacheKey.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.Objects;
+
+public class StatisticsFileCacheKey
+{
+    private final StatisticsFile file;
+    private final int columnId;
+
+    public StatisticsFileCacheKey(StatisticsFile file, int columnId)
+    {
+        this.file = file;
```
tdcmeehan, line 27:
```suggestion
        this.file = requireNonNull(file, "file is null");
```

presto-spi/src/main/java/com/facebook/presto/spi/statistics/ConnectorHistogram.java
```diff
@@ -61,4 +61,9 @@ public interface ConnectorHistogram
      * @return the value in the distribution corresponding to the percentile
      */
     Estimate inverseCumulativeProbability(double percentile);
+
+    /**
+     * @return an approximation of the memory utilization of this histogram.
+     */
+    long getMemoryUtilization();
```
tdcmeehan, line 68: Can we call it `getEstimatedSize` for consistency?

presto-spi/src/main/java/com/facebook/presto/spi/statistics/ColumnStatistics.java
```diff
@@ -14,15 +14,21 @@
...
 
 import java.util.Objects;
 import java.util.Optional;
 
+import static com.facebook.presto.spi.statistics.DoubleRange.RANGE_SIZE;
+import static com.facebook.presto.spi.statistics.Estimate.ESTIMATE_SIZE;
 import static java.lang.String.format;
 import static java.util.Objects.requireNonNull;
 
 public final class ColumnStatistics
```
tdcmeehan, line 27: Can we add some tests that show that the memory estimations are reasonably accurate?

tdcmeehan (changes requested):
Just some nits.  Main thing is it would be nice to have a test that breaks if the memory estimates change for some reason.

presto-spi/src/main/java/com/facebook/presto/spi/statistics/ColumnStatistics.java
ZacBlanco: A test in the SPI: `TestColumnStatistics` was added to address this

presto-spi/src/main/java/com/facebook/presto/spi/statistics/ConnectorHistogram.java
ZacBlanco: updated

presto-iceberg/src/main/java/com/facebook/presto/iceberg/statistics/StatisticsFileCacheKey.java
ZacBlanco: updated

tdcmeehan (dismissed):

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergErrorCode.java
```diff
@@ -38,7 +38,8 @@ public enum IcebergErrorCode
     ICEBERG_INVALID_TABLE_TIMESTAMP(12, USER_ERROR),
     ICEBERG_ROLLBACK_ERROR(13, EXTERNAL),
     ICEBERG_INVALID_FORMAT_VERSION(14, USER_ERROR),
-    ICEBERG_UNKNOWN_MANIFEST_TYPE(15, EXTERNAL);
+    ICEBERG_UNKNOWN_MANIFEST_TYPE(15, EXTERNAL),
+    ICEBERG_CACHE_WRITE_ERROR(15, INTERNAL_ERROR);
```
hantangwangd, line 42: A duplicate code here, is this intentional?

hantangwangd (commented):
One little thing, otherwise lgtm.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergErrorCode.java
ZacBlanco: This was probably unintentional after fixing a merge conflict

ZacBlanco: fixed

tdcmeehan (approved):

hantangwangd (approved):

PR #23179, opened 2024-07-11
[Draft] Add support to filter out plans that use Timestamp with Timezone

kgpai:
## Description
This change causes the native worker to throw if  Timestamp with Timezone is encountered in the plan. This is gated on the configuration property 'fail_on_timestamp_with_timezone'.

## Motivation and Context
Due to issue https://github.com/facebookincubator/velox/issues/10338, there can be cases where comparisons with Timestamp with Timezone can be wrong. 

## Impact
Will add further details soon.

## Test Plan
Working on adding tests.

```
== NO RELEASE NOTE ==
```


presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp
```diff
@@ -1924,6 +1927,120 @@ core::PlanFragment VeloxQueryPlanConverterBase::toVeloxQueryPlan(
...
+  // Is it one of the joinNodes with timestamp with tz ?
+  if (auto joinNode =
+          std::dynamic_pointer_cast<const core::HashJoinNode>(planNode)) {
+    if (exprHasTimestampWithTz({joinNode->filter()})) {
+      return true;
+    }
+  }
+
+  if (auto joinNode =
+          std::dynamic_pointer_cast<const core::AbstractJoinNode>(planNode)) {
```
kagamiori, line 1986: Can AbstractJoinNode be used standalone?

presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp
```diff
@@ -1924,6 +1927,120 @@ core::PlanFragment VeloxQueryPlanConverterBase::toVeloxQueryPlan(
...
+  if (auto tableScanNode =
+          std::dynamic_pointer_cast<const core::TableScanNode>(planNode)) {
+    if (auto hiveTableHandle =
+            std::dynamic_pointer_cast<connector::hive::HiveTableHandle>(
+                tableScanNode->tableHandle())) {
+      if (exprHasTimestampWithTz({hiveTableHandle->remainingFilter()})) {
+        return true;
+      }
+    }
+  }
```
kagamiori, line 2016: AggregationNode::aggregates_::call is a CallTypedExprPtr that can contain a LambdaTypedExprPtr as input. Should we check it too?

presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp
kgpai: Yes, it would cover both HashJoin and MergeJoin.

aditi-pandit:
@kgpai @gggrace14 @kagamiori : Am curious why the invalidation happens during Prestissimo plan conversion instead of the Presto sql parser/semantic validation of the plan ? This code is brittle and could need a lot of tweaking. Its better to reject such a query at the co-ordinator itself if we are using native engine.

It would be cleaner to drop the Timestamp with Timezone type and functions from the list here for native engine
https://github.com/prestodb/presto/blob/master/presto-main/src/main/java/com/facebook/presto/metadata/BuiltInTypeAndFunctionNamespaceManager.java#L681

https://github.com/prestodb/presto/blob/master/presto-main/src/main/java/com/facebook/presto/metadata/BuiltInTypeAndFunctionNamespaceManager.java#L681

There could be other approaches as well.

@tdcmeehan : Would be great to hear your thoughts.

amitkdutta:
> @kgpai @gggrace14 @kagamiori : Am curious why the invalidation happens during Prestissimo plan conversion instead of the Presto sql parser/semantic validation of the plan ? This code is brittle and could need a lot of tweaking. Its better to reject such a query at the co-ordinator itself if we are using native engine.
> 
> It would be cleaner to drop the Timestamp with Timezone type and functions from the list here for native engine https://github.com/prestodb/presto/blob/master/presto-main/src/main/java/com/facebook/presto/metadata/BuiltInTypeAndFunctionNamespaceManager.java#L681
> 
> https://github.com/prestodb/presto/blob/master/presto-main/src/main/java/com/facebook/presto/metadata/BuiltInTypeAndFunctionNamespaceManager.java#L681
> 
> There could be other approaches as well.
> 
> @tdcmeehan : Would be great to hear your thoughts.

@aditi-pandit We plan to fix this in velox as quickly as possible. We are putting this interim solution guarded by a query config and intend to remove it as soon as velox support is landed (which we expected to be done in a few weeks).

tdcmeehan:
Understood that this is a temporary measure.  But I don't understand the hesitation around adding this into the coordinator?  It would just be a better check--it would occur during analysis, rather than during execution (which is *after* resources have been allocated, after queueing, tasks have been spun up and splits have begun to be enumerated).  I also think it would just be simpler to put it there.

amitkdutta:
> Understood that this is a temporary measure. But I don't understand the hesitation around adding this into the coordinator? It would just be a better check--it would occur during analysis, rather than during execution (which is _after_ resources have been allocated, after queueing, tasks have been spun up and splits have begun to be enumerated). I also think it would just be simpler to put it there.

@tdcmeehan @aditi-pandit Yea coordinator is definitely a better place, I am trying here 
https://github.com/prestodb/presto/pull/23191
I stumbled upon into two cases (https://github.com/prestodb/presto/pull/23191#issuecomment-2224750063), will be really helpful if you can guide and take a look. This change is very important and also time sensitive.

PR #22764, opened 2024-05-15
Add support for memoization to resource group state endpoint

swapsmagic:
## Description
Adding memoization support to /v1/resourceGroupState endpoint. 

## Motivation and Context
This will help reduce load on the coordinator to
calculate the resource group state for concurrent/frequent requests in a short period of time.

## Impact
Less contention on coordinator from /v1/resourceGroupState endpoint

## Test Plan
unit test

## Contributor checklist

- [x] Please make sure your submission complies with our [development](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#development), [formatting](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#formatting), [commit message](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#commit-formatting-and-pull-requests), and [attribution guidelines](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#attribution).
- [x] PR description addresses the issue accurately and concisely.  If the change is non-trivial, a GitHub Issue is referenced.
- [x] Documented new properties (with its default value), SQL syntax, functions, or other functionality.
- [x] If release notes are required, they follow the [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines).
- [x] Adequate tests were added if applicable.
- [x] CI passed.

## Release Notes
Please follow [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines) and fill in the release notes below.

```
== RELEASE NOTES ==

General Changes
* Add support for memoizing in resource group state info endpoint. This can be enabled by setting `cluster-resource-group-state-info-expiration-duration` to a non-zero duration. :pr:`22764`
```

PR #23192, opened 2024-07-12
[native] Add support for cache periodic full persistence

zacw7:
## Description
Currently, maxWriteRatio(AsyncDataCache::Options), the maximum ratio of the number of in-memory cache entries being written to the SSD cache over the total number of cache entries, is set to the default value of 0.7, meaning that we only persist 70% of cache entries to the SSD and upon server restart 30% of cache will be lost.

This change adds an optional periodic daemon to flush cache data to ssd. This can help to make sure all the in-memory data is flushed to ssd.

```
== RELEASE NOTES ==

General Changes
* Add support for persisting full memory cache to SSD periodically on Presto C++ worker. This can be enabled by setting `cache.velox.full-persistence-interval` to a non-zero value. :pr:`23192`
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/PrestoServer.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoServer.cpp
@@ -119,6 +119,14 @@ bool isCacheTtlEnabled() {
   return false;
 }
 
+bool isCachePeriodicFullPersistenceEnabled() {
+  const auto* systemConfig = SystemConfig::instance();
+  return systemConfig->asyncDataCacheEnabled() &&
+      systemConfig->asyncCacheSsdGb() > 0 &&
+      systemConfig->asyncCacheFullPersistenceInterval() >
+      std::chrono::seconds::zero();
+}
+
 } // namespace
 
 std::string nodeState2String(NodeState nodeState) {
@@ -944,6 +952,49 @@ void PrestoServer::addServerPeriodicTasks() {
         },
         ttlCheckInterval,
         "cache_ttl");
+  }
+
+  if (isCachePeriodicFullPersistenceEnabled()) {
+    PRESTO_STARTUP_LOG(INFO)
+        << "Initializing cache periodic full persistence task...";
+    auto* cache = velox::cache::AsyncDataCache::getInstance();
+    VELOX_CHECK_NOT_NULL(cache);
+    auto* ssdCache = cache->ssdCache();
+    VELOX_CHECK_NOT_NULL(ssdCache);
+    const auto* systemConfig = SystemConfig::instance();
+    const int64_t cacheFullPersistenceIntervalUs =
+        std::chrono::duration_cast<std::chrono::microseconds>(
+            systemConfig->asyncCacheFullPersistenceInterval())
+            .count();
+    const auto asyncCacheSsdCheckpointGb =
+        systemConfig->asyncCacheSsdCheckpointGb();
+    periodicTaskManager_->addTask(
+        [asyncCacheSsdCheckpointGb, cache, ssdCache]() {
+          try {
+            if (!ssdCache->startWrite()) {
+              return;
+            }
+            LOG(INFO) << "Persisting full cache to SSD...";
+            cache->saveToSsd(true);
+            ssdCache->waitForWriteToFinish();
+            LOG(INFO) << "Cache full persistence completed.";
+
+            if (asyncCacheSsdCheckpointGb == 0) {
+              return;
+            }
+
+            if (!ssdCache->startWrite()) {
+              return;
+            }
+
+            ssdCache->checkpoint();
+            ssdCache->waitForWriteToFinish();
+          } catch (const std::exception& e) {
+            LOG(ERROR) << "Failed to persistent cache to SSD: " << e.what();
+          }
+        },
+        cacheFullPersistenceIntervalUs,
+        "cache_full_persistence");
   }
 }
 
--- a/presto-native-execution/presto_cpp/main/common/Configs.cpp
+++ b/presto-native-execution/presto_cpp/main/common/Configs.cpp
@@ -189,6 +189,7 @@ SystemConfig::SystemConfig() {
           NUM_PROP(kAsyncCacheMaxSsdWriteRatio, 0.7),
           NUM_PROP(kAsyncCacheSsdSavableRatio, 0.125),
           NUM_PROP(kAsyncCacheMinSsdSavableBytes, 1 << 24 /*16MB*/),
+          STR_PROP(kAsyncCacheFullPersistenceInterval, "0s"),
           BOOL_PROP(kAsyncCacheSsdDisableFileCow, false),
           BOOL_PROP(kSsdCacheChecksumEnabled, false),
           BOOL_PROP(kSsdCacheReadVerificationEnabled, false),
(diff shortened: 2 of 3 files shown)


presto-native-execution/presto_cpp/main/common/Configs.cpp
```diff
@@ -189,6 +189,7 @@ SystemConfig::SystemConfig() {
           NUM_PROP(kAsyncCacheMaxSsdWriteRatio, 0.7),
           NUM_PROP(kAsyncCacheSsdSavableRatio, 0.125),
           NUM_PROP(kAsyncCacheMinSsdSavableBytes, 1 << 24 /*16MB*/),
+          STR_PROP(kAsyncCacheFullPersistenceInterval, "0s"),
```
aditi-pandit, line 192: @zacw7 : Is it possible to use a negative value here instead ? 0 gives the impression that it always runs (though the code doesn't run it)

aditi-pandit (commented):
Thanks @zacw7.

Please can you add a release note for this functionality. Would greatly help users.

presto-native-execution/presto_cpp/main/common/Configs.cpp
zacw7: To avoid confusion I've updated the comment, thanks for pointing that out!

I would prefer keep 0 as the default value here as it seems more aligned with the convention of similar configs (ex. https://github.com/prestodb/presto/pull/22764). If more folks think it gives the wrong impression I'm open to have it changed to a negative value? WDYT @tanjialiang

zacw7:
> Thanks @zacw7.
> 
> Please can you add a release note for this functionality. Would greatly help users.

release note added. thanks!

steveburnett:
Nit, please add the PR number to the release note entry following the [Release Notes Guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines): 

```
== RELEASE NOTES ==

General Changes
* Add support for persisting full memory cache to SSD periodically on Presto C++ worker. This can be enabled by setting `cache.velox.full-persistence-interval` to a non-zero value. :pr:`23192`
```

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -945,6 +953,49 @@ void PrestoServer::addServerPeriodicTasks() {
         ttlCheckInterval,
         "cache_ttl");
   }
+
+  if (isCachePeriodicFullPersistenceEnabled()) {
+    PRESTO_STARTUP_LOG(INFO)
+        << "Initializing cache periodic full persistence task...";
+    const int64_t cacheFullPersistenceInterval =
+        std::chrono::duration_cast<std::chrono::seconds>(
+            SystemConfig::instance()->asyncCacheFullPersistenceInterval())
```
tanjialiang, line 962: We can take SystemConfig->instance() as a local v

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -945,6 +953,49 @@ void PrestoServer::addServerPeriodicTasks() {
...
+
+          if (asyncCacheSsdCheckpointGb == 0) {
+            return;
+          }
+
+          if (!ssdCache->startWrite()) {
+            return;
+          }
+          ssdCache->checkpoint();
+          ssdCache->waitForWriteToFinish();
```
tanjialiang, line 993: Let's add exception handling for these operations.

tanjialiang (changes requested):

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -945,6 +953,49 @@ void PrestoServer::addServerPeriodicTasks() {
...
+        [asyncCacheSsdCheckpointGb]() {
+          auto* cache = velox::cache::AsyncDataCache::getInstance();
+          if (cache == nullptr) {
+            return;
+          }
+
+          auto* ssdCache = cache->ssdCache();
+          if (ssdCache == nullptr) {
+            return;
+          }
```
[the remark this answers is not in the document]
zacw7: If cache and ssdcache both are non-null, shall we just ignore it instead of failing the server hard? Similarly, we still allow the server operation api to be called even when there is no cache.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -915,12 +923,14 @@ void PrestoServer::updateAnnouncerDetails() {
 }
 
 void PrestoServer::addServerPeriodicTasks() {
+  const auto* systemConfig = SystemConfig::instance();
```
tanjialiang, line 926: I think you need to capture this pointer by copy in the below lambda. Or you can simply put the SystemConfig::instance() call inside lambda. Yeah let's just move it inside lambda. There is not a strong need of having it outside.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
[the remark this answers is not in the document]
tanjialiang: What I meant is something like this
```
if (isCachePeriodicFullPersistenceEnabled()) {
  VELOX_CHECK_NOT_NULL(cache);
  VELOX_CHECK_NOT_NULL(ssdCache);
}
```
Or are there any scenarios that they can actually be NULL when the if condition is true?

tanjialiang (commented):
Thanks @zacw7 for the revision. Some minors

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -945,6 +953,48 @@ void PrestoServer::addServerPeriodicTasks() {
...
+
+            if (!ssdCache->startWrite()) {
+              return;
+            }
+            ssdCache->checkpoint();
+            ssdCache->waitForWriteToFinish();
+            LOG(INFO) << "Cache full persistence completed.";
+          } catch (const VeloxRuntimeError& e) {
+            LOG(ERROR) << "Failed to persistent cache to SSD: " << e.what();
+          }
```
tanjialiang, line 993: Do we expect other exception types? E.g. exceptions thrown by os when writing and etc. Do we want to have a wild catch at the end? Thanks

tanjialiang (dismissed):

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -945,6 +953,48 @@ void PrestoServer::addServerPeriodicTasks() {
...
+          try {
+            if (!ssdCache->startWrite()) {
+              return;
+            }
+            LOG(INFO) << "Persisting full cache to SSD...";
+            cache->saveToSsd(true);
+            ssdCache->waitForWriteToFinish();
+
+            if (asyncCacheSsdCheckpointGb == 0) {
+              return;
```
aditi-pandit, line 982: Maybe we should add some LOG messages before both returns here to correlate in the log.

zacw7: Good catch. I've moved the log for completion ahead. For checkpoint, we can rely on the logging inside the functions. No need for extra logging here.

tanjialiang (approved):

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -945,6 +953,49 @@ void PrestoServer::addServerPeriodicTasks() {
...
+    const auto asyncCacheSsdCheckpointGb =
+        systemConfig->asyncCacheSsdCheckpointGb();
+    periodicTaskManager_->addTask(
+        [asyncCacheSsdCheckpointGb, cache, ssdCache]() {
+          try {
+            if (!ssdCache->startWrite()) {
+              return;
+            }
+            LOG(INFO) << "Persisting full cache to SSD...";
+            cache->saveToSsd(true);
```
xiaoxmeng, line 978:
```
saveToSsd(false)
```

PR #23195, opened 2024-07-12
Send periodic query progress events

pgupta2:
## Description
Add support for sending periodic progress events for Queued/Running queries using EventListener interface. 

## Motivation and Context
This change enables coordinators to push their current state of Queued/Running queries as progress events which can be logged to different sources for analysis purposes. Progress event is sent every 1 minute for each queued/running queries.

## Impact
None

## Test Plan
Added a Unit test to check for queryProgress events. 
Existing tests

## Contributor checklist

- [ ] CI passed.

```
== RELEASE NOTES ==
General Changes
* Add a new configuration property ``event.query-progress-publish-interval`` to specify the time interval at which progress events should be generated. Default is every 1 minute. :pr:`23195`
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/event/QueryMonitor.java
+++ b/presto-main/src/main/java/com/facebook/presto/event/QueryMonitor.java
@@ -46,6 +46,7 @@ import com.facebook.presto.operator.OperatorStats;
 import com.facebook.presto.operator.TableFinishInfo;
 import com.facebook.presto.operator.TaskStats;
 import com.facebook.presto.server.BasicQueryInfo;
+import com.facebook.presto.server.BasicQueryStats;
 import com.facebook.presto.spi.ConnectorId;
 import com.facebook.presto.spi.QueryId;
 import com.facebook.presto.spi.eventlistener.OperatorStatistics;
@@ -57,6 +58,7 @@ import com.facebook.presto.spi.eventlistener.QueryIOMetadata;
 import com.facebook.presto.spi.eventlistener.QueryInputMetadata;
 import com.facebook.presto.spi.eventlistener.QueryMetadata;
 import com.facebook.presto.spi.eventlistener.QueryOutputMetadata;
+import com.facebook.presto.spi.eventlistener.QueryProgressEvent;
 import com.facebook.presto.spi.eventlistener.QueryStatistics;
 import com.facebook.presto.spi.eventlistener.QueryUpdatedEvent;
 import com.facebook.presto.spi.eventlistener.ResourceDistribution;
@@ -171,6 +173,30 @@ public class QueryMonitor
     public void queryUpdatedEvent(QueryInfo queryInfo)
     {
         eventListenerManager.queryUpdated(new QueryUpdatedEvent(createQueryMetadata(queryInfo)));
+    }
+
+    public void publishQueryProgressEvent(long monotonicallyIncreasingEventId, BasicQueryInfo queryInfo)
+    {
+        eventListenerManager.publishQueryProgress(new QueryProgressEvent(
+                monotonicallyIncreasingEventId,
+                new QueryMetadata(
+                        queryInfo.getQueryId().toString(),
+                        queryInfo.getSession().getTransactionId().map(TransactionId::toString),
+                        queryInfo.getQuery(),
+                        queryInfo.getQueryHash(),
+                        queryInfo.getPreparedQuery(),
+                        queryInfo.getState().toString(),
+                        queryInfo.getSelf(),
+                        Optional.empty(),
+                        Optional.empty(),
+                        Optional.empty(),
+                        Optional.empty(),
+                        ImmutableList.of(),
+                        queryInfo.getSession().getTraceToken()),
+                createQueryStatistics(queryInfo),
+                createQueryContext(queryInfo.getSession(), queryInfo.getResourceGroupId()),
+                queryInfo.getQueryType(),
+                ofEpochMilli(queryInfo.getQueryStats().getCreateTime().getMillis())));
     }
 
     public void queryImmediateFailureEvent(BasicQueryInfo queryInfo, ExecutionFailureInfo failure)
@@ -429,6 +455,46 @@ public class QueryMonitor
                 queryStats.getCompletedDrivers(),
                 queryInfo.isFinalQueryInfo(),
                 queryStats.getRuntimeStats());
+    }
+
+    private QueryStatistics createQueryStatistics(BasicQueryInfo basicQueryInfo)
+    {
+        BasicQueryStats queryStats = basicQueryInfo.getQueryStats();
+
+        return new QueryStatistics(
+                ofMillis(queryStats.getTotalCpuTime().toMillis()),
+                ofMillis(0),
+                ofMillis(queryStats.getElapsedTime().toMillis()),
+                ofMillis(queryStats.getWaitingForPrerequisitesTime().toMillis()),
+                ofMillis(queryStats.getQueuedTime().toMillis()),
+                ofMillis(0),
+                ofMillis(0),
+                ofMillis(0),
+                ofMillis(0),
+                ofMillis(0),
+                Optional.of(ofMillis(0)),
+                ofMillis(queryStats.getExecutionTime().toMillis()),
+                queryStats.getPeakRunningTasks(),
+                queryStats.getPeakUserMemoryReservation().toBytes(),
+                queryStats.getPeakTotalMemoryReservation().toBytes(),
+                0,
+                0,
+                0,
+                0,
+                0,
+                queryStats.getRawInputDataSize().toBytes(),
+                queryStats.getRawInputPositions(),
+                0,
+                0,
+                0,
+                0,
+                0,
+                0,
+                queryStats.getCumulativeUserMemory(),
+                queryStats.getCumulativeTotalMemory(),
+                queryStats.getCompletedDrivers(),
+                false,
+                new RuntimeStats());
     }
 
     private QueryContext createQueryContext(SessionRepresentation session, Optional<ResourceGroupId> resourceGroup)
--- a/presto-main/src/main/java/com/facebook/presto/event/QueryMonitorConfig.java
+++ b/presto-main/src/main/java/com/facebook/presto/event/QueryMonitorConfig.java
@@ -14,16 +14,21 @@
 package com.facebook.presto.event;
 
 import com.facebook.airlift.configuration.Config;
+import com.facebook.airlift.configuration.ConfigDescription;
 import io.airlift.units.DataSize;
 import io.airlift.units.DataSize.Unit;
+import io.airlift.units.Duration;
 import io.airlift.units.MaxDataSize;
 import io.airlift.units.MinDataSize;
 
 import javax.validation.constraints.NotNull;
 
+import static java.util.concurrent.TimeUnit.MINUTES;
+
 public class QueryMonitorConfig
 {
     private DataSize maxOutputStageJsonSize = new DataSize(16, Unit.MEGABYTE);
+    private Duration queryProgressPublishInterval = new Duration(0, MINUTES);
 
     @MinDataSize("1kB")
     @MaxDataSize("1GB")
@@ -37,6 +42,19 @@ public class QueryMonitorConfig
     public QueryMonitorConfig setMaxOutputStageJsonSize(DataSize maxOutputStageJsonSize)
     {
         this.maxOutputStageJsonSize = maxOutputStageJsonSize;
+        return this;
+    }
+
+    public Duration getQueryProgressPublishInterval()
+    {
+        return queryProgressPublishInterval;
+    }
+
+    @Config("event.query-progress-publish-interval")
+    @ConfigDescription("How frequently to publish query progress events. 0 duration disables the publication of these events.")
+    public QueryMonitorConfig setQueryProgressPublishInterval(Duration queryProgressPublishInterval)
+    {
+        this.queryProgressPublishInterval = queryProgressPublishInterval;
         return this;
     }
 }
--- a/presto-main/src/main/java/com/facebook/presto/event/QueryProgressMonitor.java
+++ b/presto-main/src/main/java/com/facebook/presto/event/QueryProgressMonitor.java
@@ -1,0 +1,86 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.event;
+
+import com.facebook.presto.dispatcher.DispatchManager;
+import com.facebook.presto.server.BasicQueryInfo;
+import io.airlift.units.Duration;
+
+import javax.annotation.PostConstruct;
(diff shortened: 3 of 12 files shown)


presto-main/src/main/java/com/facebook/presto/dispatcher/DispatchManager.java
```diff
@@ -154,6 +168,9 @@ public DispatchManager(
         this.queryTracker = new QueryTracker<>(queryManagerConfig, dispatchExecutor.getScheduledExecutor(), clusterQueryTrackerService);
 
         this.securityConfig = requireNonNull(securityConfig, "securityConfig is null");
+
+        this.queryMonitor = requireNonNull(queryMonitor, "queryMonitor is null");
+        this.queryHeartbeatExecutor = Executors.newSingleThreadScheduledExecutor(threadsNamed("query-heartbeat-management-%s"));
```
prithvip, line 173: Cleaner to instantiate this in CoordinatorModule and bind it to a annotated ScheduledExecutorService here - see CoordinatorModule.java for how we do this in other cases

presto-main/src/main/java/com/facebook/presto/dispatcher/DispatchManager.java
```diff
@@ -163,6 +180,14 @@ public DispatchManager(
...
     {
         queryTracker.start();
+        queryHeartbeatExecutor.scheduleWithFixedDelay(() -> {
+            try {
+                sendQueryHeartbeats();
+            }
+            catch (Throwable e) {
+                log.error(e, "Error sending Heartbeat Event");
+            }
+        }, 1, 1, TimeUnit.MINUTES);
```
prithvip, line 190: Make this configurable

presto-main/src/main/java/com/facebook/presto/dispatcher/DispatchManager.java
```diff
@@ -163,6 +180,14 @@ public DispatchManager(
     public void start()
     {
         queryTracker.start();
+        queryHeartbeatExecutor.scheduleWithFixedDelay(() -> {
+            try {
+                sendQueryHeartbeats();
+            }
+            catch (Throwable e) {
+                log.error(e, "Error sending Heartbeat Event");
```
prithvip, line 188: Issues in sending the heartbeat should be caught and handled by QueryMonitor, DispatchManager shouldn't worry about that or log anything.

presto-main/src/main/java/com/facebook/presto/dispatcher/DispatchManager.java
```diff
@@ -453,6 +479,16 @@ public void cancelQuery(QueryId queryId)
                 .ifPresent(DispatchQuery::cancel);
     }
 
+    private void sendQueryHeartbeats()
+    {
+        List<BasicQueryInfo> basicQueryInfoList = queryTracker.getAllQueries().stream().map(q -> q.getBasicQueryInfo()).collect(Collectors.toList());
```
prithvip, line 484: DispatchManager::getQueries already does this, let's use that.

prithvip (commented):
It might be better (single responsibility principle) if there was a new singleton class that calls DispatchManager::getQueries periodically and sends the heartbeat, and we can wire in any heartbeat-specific configuration there. "Query heartbeat" is overloaded in DispatchManager context, since DispatchQuery has recordHeartbeat(), which is used for purging abandoned queries, and thats totally different purpose than this "heartbeat". "sendQueryStateInfos" might be better terminology and would match with QueryStateInfoResource::getQueryStateInfos API.

prithvip (dismissed):

pgupta2:
@prithvip @tdcmeehan : Gentle reminder for the review!  Thanks.

presto-main/src/main/java/com/facebook/presto/event/QueryMonitor.java
```diff
@@ -169,6 +171,34 @@ public void queryUpdatedEvent(QueryInfo queryInfo)
...
+                            Optional.empty(),
+                            ImmutableList.of(),
+                            queryInfo.getSession().getTraceToken()),
+                    createQueryStatistics(queryInfo),
+                    createQueryContext(queryInfo.getSession(), queryInfo.getResourceGroupId()),
+                    queryInfo.getQueryType(),
+                    ofEpochMilli(queryInfo.getQueryStats().getCreateTime().getMillis())));
+        }
+        catch (Throwable e) {
+            log.error(e, "Error sending Query Progress Event");
```
prithvip, line 198: I don't think we need try-catch here, because any issues should be handled by the specific EventListener implementation which does not throw checked exception. The try-catch should happen there.

jaystarshot:
Can you add how progress events can potentially be analyzed? 
I feel that there is more value in adding whole cluster level events instead of individual query events

tdcmeehan (changes requested):
Can you please add some unit tests?

pgupta2:
> Can you add how progress events can potentially be analyzed? I feel that there is more value in adding whole cluster level events instead of individual query events

Query progress events can be used to figure out the workload distribution by users or workload type which can be used for doing more realtime quota enforcement. Another use case would be to get insights into the queuing state of the cluster at real time and identify the workloads/users contributing to it. Query progress events allows us to aggregate data on different dimensions as needed. Cluster level events are very coarse and will not expose such detailed information.

pgupta2:
> Can you please add some unit tests?

@tdcmeehan : Added a unit test and updated the PR. Thanks.

presto-main/src/main/java/com/facebook/presto/event/QueryProgressMonitor.java
```diff
@@ -0,0 +1,74 @@
...
+    {
+        this.queryMonitor = requireNonNull(queryMonitor, "queryMonitor is null");
+        this.dispatchManager = requireNonNull(dispatchManager, "dispatchManager is null");
+        this.queryProgressPublishInterval = requireNonNull(queryMonitorConfig, "queryMonitorConfig is null").getQueryProgressPublishInterval();
+    }
+
+    @PostConstruct
+    public void start()
+    {
+        queryProgressMonitorExecutor.scheduleWithFixedDelay(
```
tdcmeehan, line 53: For a heavily contended cluster, I think this would be pretty wasteful, because most queries will publish identical information each time even though they're queued and it won't change until they dequeue.  What do you think about making it based off of query state transition?  I.e. it publishes once per query state transition, and when it transitions into EXECUTING then we start a periodic timer?

presto-spi/src/main/java/com/facebook/presto/spi/eventlistener/QueryProgressEvent.java
```diff
@@ -0,0 +1,69 @@
...
+package com.facebook.presto.spi.eventlistener;
+
+import com.facebook.presto.common.resourceGroups.QueryType;
+
+import java.time.Instant;
+import java.util.Optional;
+
+import static java.util.Objects.requireNonNull;
+
+public class QueryProgressEvent
```
tdcmeehan, line 23: I think what makes this different from similar classes in this package is this is designed to publish multiple events, instead of an event per state transition or at query start/finish.  If you imagine that someone might put this into a queue that has at least once semantics, then it might be possible for messages to arrive out of order or duplicated.  That's not a problem for peer classes in this package because you know the state transition order, or you know that you only need to look at one query completed even if there's a duplicate, but for this class, it's impossible to infer correct ordering.  This may lead to strange queries or incorrect analysis of the events.  I'd recommend you include a monotonically increasing identifier that downstream systems can use to infer ordering and to de-duplicate at the reader side.

tdcmeehan (changes requested):

tdcmeehan:
> > Can you add how progress events can potentially be analyzed? I feel that there is more value in adding whole cluster level events instead of individual query events
> 
> Query progress events can be used to figure out the workload distribution by users or workload type which can be used for doing more realtime quota enforcement. Another use case would be to get insights into the queuing state of the cluster at real time and identify the workloads/users contributing to it. Query progress events allows us to aggregate data on different dimensions as needed. Cluster level events are very coarse and will not expose such detailed information.

Shouldn't higher level quota enforcement work hand-in-hand with local resource management in the resource group?  If so, wouldn't it be better to publish these events at the granularity of resource group IDs?  If not, how do you ensure that local resources are in line with global quota?

Likewise, wouldn't events at the resource group level give you the insights needed to identify the workloads and users contributing to the queueing state?

presto-main/src/main/java/com/facebook/presto/event/QueryProgressMonitor.java
pgupta2: I feel that the decision to process/ignore events for QUEUED queries should be left to the consumer itself. It also depends on the use case for which these events are being utilized. For instance, within Meta itself, these events will be used for two different purposes: 

1) Resource utilization of in-progress queries to perform quota enforcement and throttling. This consumer simply ignore events for QUEUED queries and only logs events for running queries. 

2) Getting the current queuing state of the cluster and power certain alerts/dashboard. This requires getting events for queued queries at periodic interval in order to generate a recent snapshot. Another problem I see is that if a query fails after queueing and before logging query completed event (in case of coordinator crash or cluster restart), then we will not be able to figure out if a query is in QUEUED state because its waiting for resources or whether it died due to to some reason. So, having periodic samples will surface out more accurate snaphot of the cluster state.

pgupta2:
> Shouldn't higher level quota enforcement work hand-in-hand with local resource management in the resource group? If so, wouldn't it be better to publish these events at the granularity of resource group IDs? If not, how do you ensure that local resources are in line with global quota?

My understanding might be wrong, but I feel that higher level quota enforcement can be different from local resource management provided by resource groups. AFAIK, RG provides quota management on concurrency and memory share. So, if a resource group is configured to use 50% memory share and with 100 concurrency, it can run 100 queries in parallel which can consume a total of 50% cluster memory. Global quota management could happen at team, org, product level wherein a team might have quota to use X memory resources per day and once they have consumed these resources, their workload will get blocked/throttled. This enforcement cannot be achieved at RG level IMO. Also, an org might have multiple teams using different resource groups for their workload and we need to aggregate resource usage at org level to perform such quota enforcement. 


> Likewise, wouldn't events at the resource group level give you the insights needed to identify the workloads and users contributing to the queueing state?

Yes, we should be able to collect this data at RG level. However, even at RG level, we will be collecting BasicQueryInfos for all running queries and publishing them at periodic interval. In terms of information published, it will most likely still be same, just grouped at RG level.

presto-spi/src/main/java/com/facebook/presto/spi/eventlistener/QueryProgressEvent.java
pgupta2: @tdcmeehan : Added an increasing event Id field in the QueryProgressEvent class.

pgupta2:
@tdcmeehan : Gentle reminder for the review! Thanks.

pgupta2:
@tdcmeehan : Another gentle reminder for the review!! :)

steveburnett:
Suggest adding the PR number to the release note entry per the [Release Notes Guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines). 

```
== RELEASE NOTES ==
General Changes
* Add a new configuration property ``event.query-progress-publish-interval`` to specify the time interval at which progress events should be generated. Default is every 1 minute. :pr:`23195`
```

pgupta2:
@tdcmeehan : Ping.

prithvip (approved):

presto-main/src/main/java/com/facebook/presto/event/QueryProgressMonitor.java
```diff
@@ -0,0 +1,77 @@
...
+    @PreDestroy
+    public void stop()
+    {
+        queryProgressMonitorExecutor.shutdown();
+    }
+
+    private void publishQueryProgressEvent()
+    {
+        for (BasicQueryInfo basicQueryInfo : dispatchManager.getQueries()) {
+            if (basicQueryInfo.getState() != FINISHED && basicQueryInfo.getState() != FAILED) {
```
tdcmeehan, line 72:
```suggestion
            if (!basicQueryInfo.getState().isDone()) {
```

presto-main/src/main/java/com/facebook/presto/event/QueryMonitorConfig.java
```diff
@@ -14,16 +14,21 @@
...
 import io.airlift.units.MinDataSize;
 
 import javax.validation.constraints.NotNull;
 
+import static java.util.concurrent.TimeUnit.MINUTES;
+
 public class QueryMonitorConfig
 {
     private DataSize maxOutputStageJsonSize = new DataSize(16, Unit.MEGABYTE);
+    private Duration queryProgressPublishInterval = new Duration(1, MINUTES);
```
tdcmeehan, line 31: Can this be defaulted to 0 (don't publish events)?

presto-main/src/main/java/com/facebook/presto/event/QueryProgressMonitor.java
```diff
@@ -0,0 +1,77 @@
...
+    {
+        this.queryMonitor = requireNonNull(queryMonitor, "queryMonitor is null");
+        this.dispatchManager = requireNonNull(dispatchManager, "dispatchManager is null");
+        this.queryProgressPublishInterval = requireNonNull(queryMonitorConfig, "queryMonitorConfig is null").getQueryProgressPublishInterval();
+    }
+
+    @PostConstruct
+    public void start()
+    {
+        queryProgressMonitorExecutor.scheduleWithFixedDelay(
```
tdcmeehan, line 56: Can you check if the duration is 0, and if so, make this a no-op?  And also make the executor loaded lazily?

tdcmeehan (changes requested):

tdcmeehan:
> Yes, we should be able to collect this data at RG level. However, even at RG level, we will be collecting BasicQueryInfos for all running queries and publishing them at periodic interval. In terms of information published, it will most likely still be same, just grouped at RG level.

From your message earlier, I read two use cases: more realtime quota enforcement, and identifying sources of queueing.  For both of these, I'm wondering if you get sufficient information from resource group nodes.  Instead of publishing `BasicQueryInfos`, I'm wondering if you can publish `ResourceGroupInfo`s, perhaps with additional set of `QueryIds` if you want to drill down into particular queries.  Overall, I'm just wondering how practical this change is for most people, since it seems like a ton of information, and you think that this sort of information is spooled somewhere and hence would require potentially costly, online storage to provide that spooling.  And if the main things you need are all for the purposes of aggregating it to the resource group level, why not just publish at the resource group level?

pgupta2:
>> Instead of publishing BasicQueryInfos, I'm wondering if you can publish ResourceGroupInfos, perhaps with additional set of QueryIds if you want to drill down into particular queries. 

Just to confirm my understanding, are you suggesting that we can publish `ResourceGroupInfos` which will contain information about all running queries in the given resource group as a single event instead of publishing information about each query as a separate event? 

`ResourceGroupInfo` only contains information about running queries so we might have to extend it to expose queued queries information as well. It can be done but first I want to ensure that my above understanding is correct or not.

presto-main/src/main/java/com/facebook/presto/event/QueryMonitorConfig.java
```diff
@@ -39,4 +44,17 @@ public QueryMonitorConfig setMaxOutputStageJsonSize(DataSize maxOutputStageJsonS
...
         return this;
     }
+
+    public Duration getQueryProgressPublishInterval()
+    {
+        return queryProgressPublishInterval;
+    }
+
+    @Config("event.query-progress-publish-interval")
+    @ConfigDescription("How frequently to publish query progress events")
```
tdcmeehan, line 54:
```suggestion
    @ConfigDescription("How frequently to publish query progress events.  0ms disables the publication of these events.")
```

steveburnett:
Please add documentation for this new configuration property to https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/admin/properties.rst .

tdcmeehan (approved):

PR #23202, opened 2024-07-13
Add array_contains_array function

jainavi17:
UDF to return whether all elements of an second array are present in the first array

Test plan

Added unit tests.
Build successfully using the following terminal command
- `mvn clean install -Dtest=TestArrayOperators -Dmaven.javadoc.skip=true -DskipUI -T1C -fn -pl presto-main`

```
== RELEASE NOTES ==

General Changes
- Add function :func:`array_contains_array`
```

### DIFF
--- a/presto-docs/src/main/sphinx/functions/array.rst
+++ b/presto-docs/src/main/sphinx/functions/array.rst
@@ -78,7 +78,7 @@ Array Functions
 .. function:: array_has_duplicates(array(T)) -> boolean
 
     Returns a boolean: whether ``array`` has any elements that occur more than once.
-    Throws an exception if any of the elements are rows or arrays that contain nulls. 
+    Throws an exception if any of the elements are rows or arrays that contain nulls.
 
     SELECT array_has_duplicates(ARRAY[1, 2, null, 1, null, 3]) -- true
     SELECT array_has_duplicates(ARRAY[ROW(1, null), ROW(1, null)]) -- "map key cannot be null or contain nulls"
@@ -267,6 +267,10 @@ Array Functions
 .. function:: contains(x, element) -> boolean
 
     Returns true if the array ``x`` contains the ``element``.
+
+.. function:: contains_all(x, y) -> boolean
+
+    Returns true if the array ``x`` contains all the element of array ``y``.
 
 .. function:: element_at(array(E), index) -> E
 
--- a/presto-main/src/main/java/com/facebook/presto/operator/scalar/ArrayContainsAllFunction.java
+++ b/presto-main/src/main/java/com/facebook/presto/operator/scalar/ArrayContainsAllFunction.java
@@ -1,0 +1,52 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.operator.scalar;
+
+import com.facebook.presto.common.block.Block;
+import com.facebook.presto.common.type.StandardTypes;
+import com.facebook.presto.common.type.Type;
+import com.facebook.presto.operator.aggregation.TypedSet;
+import com.facebook.presto.spi.function.Description;
+import com.facebook.presto.spi.function.ScalarFunction;
+import com.facebook.presto.spi.function.SqlNullable;
+import com.facebook.presto.spi.function.SqlType;
+import com.facebook.presto.spi.function.TypeParameter;
+
+@ScalarFunction("contains_all")
+@Description("Returns true if all elements of the second array are present in the first array")
+public final class ArrayContainsAllFunction
+{
+    private ArrayContainsAllFunction() {}
+
+    @TypeParameter("T")
+    @SqlType(StandardTypes.BOOLEAN)
+    @SqlNullable
+    public static Boolean containsAll(
+            @TypeParameter("T") Type elementType,
+            @SqlType("array(T)") Block firstArray,
+            @SqlType("array(T)") Block secondArray)
+    {
+        TypedSet firstSet = new TypedSet(elementType, firstArray.getPositionCount(), "containsAll");
+        for (int i = 0; i < firstArray.getPositionCount(); i++) {
+            firstSet.add(firstArray, i);
+        }
+
+        for (int i = 0; i < secondArray.getPositionCount(); i++) {
+            if (!firstSet.contains(secondArray, i)) {
+                return false;
+            }
+        }
+        return true;
+    }
+}
--- a/presto-main/src/test/java/com/facebook/presto/type/TestArrayOperators.java
+++ b/presto-main/src/test/java/com/facebook/presto/type/TestArrayOperators.java
@@ -1485,6 +1485,36 @@ public class TestArrayOperators
     }
 
     @Test
+    public void testArrayContainsArray()
+    {
+        // Test Overlapping Arrays
+        assertFunction("CONTAINS_ALL(ARRAY [1, 2, 3], ARRAY [2])", BooleanType.BOOLEAN, true);
+        assertFunction("CONTAINS_ALL(ARRAY [1, 2, 3], ARRAY [2, 3])", BooleanType.BOOLEAN, true);
+        assertFunction("CONTAINS_ALL(ARRAY [1, 2, 3], ARRAY [1, 2, 3])", BooleanType.BOOLEAN, true);
+
+        // Test Disjoint Arrays
+        assertFunction("CONTAINS_ALL(ARRAY [1, 2, 3], ARRAY [4])", BooleanType.BOOLEAN, false);
+        assertFunction("CONTAINS_ALL(ARRAY [1, 2, 3], ARRAY [4, 5])", BooleanType.BOOLEAN, false);
+
+        // Test Empty Arrays
+        assertFunction("CONTAINS_ALL(ARRAY [], ARRAY [])", BooleanType.BOOLEAN, true);
+        assertFunction("CONTAINS_ALL(ARRAY [1, 2, 3], ARRAY [])", BooleanType.BOOLEAN, true);
+        assertFunction("CONTAINS_ALL(ARRAY [], ARRAY [1, 2, 3])", BooleanType.BOOLEAN, false);
+
+        // Test Arrays with Different Data Types
+        assertFunction("CONTAINS_ALL(ARRAY [1.0, 2.0, 3.0], ARRAY [2.0])", BooleanType.BOOLEAN, true);
+        assertFunction("CONTAINS_ALL(ARRAY [1.0, 2.0, 3.0], ARRAY [4.0])", BooleanType.BOOLEAN, false);
+        assertFunction("CONTAINS_ALL(ARRAY ['a', 'b', 'c'], ARRAY ['b'])", BooleanType.BOOLEAN, true);
+        assertFunction("CONTAINS_ALL(ARRAY ['a', 'b', 'c'], ARRAY ['d'])", BooleanType.BOOLEAN, false);
+
+        // Test Edge Cases
+        assertFunction("CONTAINS_ALL(ARRAY [NULL, 2, 3], ARRAY [2])", BooleanType.BOOLEAN, true);
+        assertFunction("CONTAINS_ALL(ARRAY [1, NULL, 3], ARRAY [NULL])", BooleanType.BOOLEAN, true);
+        assertFunction("CONTAINS_ALL(ARRAY [1, 2, 3], ARRAY [NULL])", BooleanType.BOOLEAN, false);
+        assertFunction("CONTAINS_ALL(ARRAY [NULL, NULL], ARRAY [NULL])", BooleanType.BOOLEAN, true);
+    }
+
+    @Test
     public void testDistinctFrom()
     {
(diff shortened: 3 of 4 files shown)


presto-docs/src/main/sphinx/functions/array.rst
```diff
@@ -268,6 +268,10 @@ Array Functions
 
     Returns true if the array ``x`` contains the ``element``.
 
+.. function:: contains_all(x, y) -> boolean
+
+    Returns true if the array ``x`` contains all the element of array ``y``.
```
elharo, line 273: elements
or
each element of the array

presto-main/src/test/java/com/facebook/presto/type/TestArrayOperators.java
```diff
@@ -1484,6 +1484,36 @@ public void testComparison()
         assertFunction("ARRAY [ARRAY [1, 2], ARRAY [3, 4, 5]] < ARRAY [ARRAY [1, 2], ARRAY [3, 4, 5]]", BOOLEAN, false);
     }
 
+    @Test
+    public void testArrayContainsArray()
```
elharo, line 1488: This is about 20 separate tests wrapped into one method. Splitting them out into 20 test methods makes them unit tests that can pass or fail and be debugged independently. As is a single failure hides subsequent failures.

elharo (changes requested):
Implementation looks good. I think this needs an issue, and perhaps an RFC. 

Also if it's in the code is it a UDF or just a function?

kaikalur:
I suggest array_contains_all as the name

presto-main/src/main/java/com/facebook/presto/operator/scalar/ArrayContainsAllFunction.java
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        for (int i = 0; i < secondArray.getPositionCount(); i++) {
+            if (!firstSet.contains(secondArray, i)) {
+                return false;
+            }
+        }
+        return true;
+    }
+}
```
kaikalur, line 52: dupes handling? Looks like you want to ignore dupes. So is_array_subset might work better

kaikalur: If you want to keep it array-like semantics, sort both and compare

kaikalur (changes requested):

presto-main/src/main/java/com/facebook/presto/operator/scalar/ArrayContainsAllFunction.java
kaikalur: yeah best might be:

array_sort(a2) = slice(array_sort(a1), 1, cardinality(a2))

keeps it clean array semantics.

presto-docs/src/main/sphinx/functions/array.rst
steveburnett, line 273:
```suggestion
    Returns true if the array ``x`` contains all the elements of array ``y``.
```

steveburnett (changes requested):
Thanks for the doc! Nit only.

presto-main/src/test/java/com/facebook/presto/type/TestArrayOperators.java
jainavi17: Not sure if I followed it. @elharo are you recommending to split all the test into individual method?

presto-main/src/main/java/com/facebook/presto/operator/scalar/ArrayContainsAllFunction.java
jainavi17: > dupes handling? Looks like you want to ignore dupes. So is_array_subset might work better

Yes dupes are ignored. The function will check if all the distinct elements of array2 are present in array1.

steveburnett (dismissed):
LGTM! (docs)

Pull updated branch, new local doc build, docs look good. Thanks!

presto-main/src/test/java/com/facebook/presto/type/TestArrayOperators.java
elharo: No, absolutely not. I feel like I need to write some articles about this because devs keep making the same bad arguments (and by no means just in this project) instead of simply following long-established best practices for unit tests. To be explicit:

1. One test per one test method. Independent tests should pass or fail independently.
2. Use fixtures for common state (though you don't even need that here).
3. Use more test classes. Don't put all tests in one test class.
4. Most importantly, no cargo cult programming. Do what's right even if existing code does it wrong. Consistency with bad code is not a virtue.

jainavi17: @elharo Created separate test cases for individual test type. Let me know if it looks good.

elharo (approved):

presto-docs/src/main/sphinx/functions/array.rst
```diff
@@ -40,6 +40,10 @@ Array Functions
     Returns the average of all non-null elements of the ``array``. If there is no non-null elements, returns
     ``null``.
 
+.. function:: array_contains_all(x, y) -> boolean
+
+    Returns true if the array ``x`` contains all the elements of array ``y``.
```
kaikalur, line 45: Mention about null behavior and also make sure it's consistent with our recent changes to other udfs

presto-docs/src/main/sphinx/functions/array.rst
```diff
@@ -78,7 +82,7 @@ Array Functions
 .. function:: array_has_duplicates(array(T)) -> boolean
 
     Returns a boolean: whether ``array`` has any elements that occur more than once.
-    Throws an exception if any of the elements are rows or arrays that contain nulls. 
+    Throws an exception if any of the elements are rows or arrays that contain nulls.
```
kaikalur, line 85: This is non-standard. In sql we don't throw exceptions. You should define a clean null semantics. @rschlussel - can you take a look?

presto-docs/src/main/sphinx/functions/array.rst
kaikalur: Is it a better version of:

cardinality(array_instrsect(x, y)) > 0

? if so that'll be the best way to do express it. Since you are using a set, maybe use something to indicate it's a set containment semantics by naming it appropriately - like: array_superset(x, y) where x is a superset of y might be best

presto-main/src/main/java/com/facebook/presto/operator/scalar/ArrayContainsAllFunction.java
```diff
@@ -0,0 +1,52 @@
...
+    private ArrayContainsAllFunction() {}
+
+    @TypeParameter("T")
+    @SqlType(StandardTypes.BOOLEAN)
+    @SqlNullable
+    public static Boolean containsAll(
+            @TypeParameter("T") Type elementType,
+            @SqlType("array(T)") Block firstArray,
+            @SqlType("array(T)") Block secondArray)
+    {
```
kaikalur, line 39: Add short-circuit empty checks. - they help

kaikalur (changes requested):

presto-docs/src/main/sphinx/functions/array.rst
rschlussel: I think this is equivalent to`cardinality(array_intersect(x,y)) = cardinality(y)`.  `cardinality(array_instersect(x, y)) > 0` would be arrays_overlap

presto-main/src/main/java/com/facebook/presto/operator/scalar/ArrayContainsAllFunction.java
```diff
@@ -0,0 +1,52 @@
...
+        TypedSet firstSet = new TypedSet(elementType, firstArray.getPositionCount(), "arrayContainsAll");
```
rschlussel, line 40: Can you benchmark this version vs. using OptimizedTypeSet and checking the cardinality of the intersection of the two sets?  I expect that using OptimizedTypedSet would be faster except for the case where a large second array is able to short circuit very early.

presto-main/src/test/java/com/facebook/presto/type/TestArrayContainsAll.java
```diff
@@ -0,0 +1,112 @@
...
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+import static com.facebook.presto.SystemSessionProperties.FIELD_NAMES_IN_JSON_CAST_ENABLED;
+import static com.facebook.presto.common.type.UnknownType.UNKNOWN;
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.fail;
+
+public class TestArrayContainsAll
```
rschlussel, line 32: can you add a test for arrays of complex types (array of arrays and array of rows)

presto-main/src/main/java/com/facebook/presto/operator/scalar/ArrayContainsAllFunction.java
kaikalur: Actually curions - how about CONTAINS? @rschlussel  did you fix that as well to be IS DISTINCT FROM? This function should behave like contains for that part

rschlussel: contains fails for all null values. If we want this to fail for all nulls, we can do that as well, but we shouldn't have failures only for nulls nested in complex types, but otherwise compare nulls as equal (which is what typedset does by default)

elharo:
FWIW, if anything, this function does not exist in Spark SQL, so adding this function increases the divergence between Presto SQL and Spark SQL. I don't know how much weight that carries here, though I do know some people care about this

kaikalur:
> FWIW, if anything, this function does not exist in Spark SQL, so adding this function increases the divergence between Presto SQL and Spark SQL. I don't know how much weight that carries here, though I do know some people care about this

That's not a concern especially because adding udfs to Spark is easier

aditi-pandit:
@jainavi17 : Please can you create an issue to follow up for the Native engine.

presto-docs/src/main/sphinx/functions/array.rst
jainavi17: These seems to be unrelated. Only removing the spaces.

presto-docs/src/main/sphinx/functions/array.rst
jainavi17: ARRAY_INTERSET just returns the distinct element. ARRAY_CONTAINS_ALL will check if array X has all the elements of ARRAY including duplicates.

jainavi17:
> @jainavi17 : Please can you create an issue to add this function for the Native engine.

Sure

presto-docs/src/main/sphinx/functions/array.rst
```diff
@@ -78,7 +82,7 @@ Array Functions
...
 
     SELECT array_has_duplicates(ARRAY[1, 2, null, 1, null, 3]) -- true
```
steveburnett, line 87: Please fix the formatting for the code example - see screenshot of local doc build. 
Screenshot 2024-09-04 at 1 06 55 PM

steveburnett (changes requested):
Just a nit of formatting.

presto-docs/src/main/sphinx/functions/array.rst
jainavi17: Should we do this in a differ PR as this PR is for ARRAY_CONTAINS_ALL?

jainavi17:
Hi @NikhilCollooru  , if you could review the PR please.

presto-docs/src/main/sphinx/functions/array.rst
steveburnett: You're right, ignore this. I'll deal with it in a separate PR.

steveburnett (approved):
LGTM! (docs)

Pull updated branch, new local doc build. Thanks!

presto-docs/src/main/sphinx/functions/array.rst
steveburnett: #23583

jainavi17:
@elharo if you could take a look at the changes please.

presto-main/src/main/java/com/facebook/presto/operator/scalar/ArrayContainsAllFunction.java
rschlussel: @jainavi17 did you have a chance to benchmark vs. a version using OptimizedTypeSet?

presto-main/src/test/java/com/facebook/presto/type/TestArrayContainsAll.java
```diff
@@ -0,0 +1,129 @@
...
+    public void testDifferentDataTypes()
+    {
+        assertFunction("ARRAY_CONTAINS_ALL(ARRAY [1.0, 2.0, 3.0], ARRAY [2.0])", BooleanType.BOOLEAN, true);
+        assertFunction("ARRAY_CONTAINS_ALL(ARRAY [1.0, 2.0, 3.0], ARRAY [4.0])", BooleanType.BOOLEAN, false);
+        assertFunction("ARRAY_CONTAINS_ALL(ARRAY ['a', 'b', 'c'], ARRAY ['b'])", BooleanType.BOOLEAN, true);
+        assertFunction("ARRAY_CONTAINS_ALL(ARRAY ['a', 'b', 'c'], ARRAY ['d'])", BooleanType.BOOLEAN, false);
+    }
+
+    @Test
+    public void testNulls()
```
rschlussel, line 91: add a test for array of arrays/rows containing nulls

rschlussel:
> > @jainavi17 : Please can you create an issue to add this function for the Native engine.
> 
> Native engine as in?

native engine means that we need a corresponding c++ function for this as we are in the process of migrating from java workers to c++ workers.

jainavi17:
Hi @rschlussel I tried to do benchmark test but somehow my IDE is breaking and I'm unable to set up the benchmark test. Is there any way I can do it on command line? Thanks

PR #23219, opened 2024-07-15
Make division by zero query failed errors clearer

### LINKED ISSUE #23215: Query failed: / by zero --> division by zero
I had to look at this error message twice before I figured it out. Spelling out "division" instead of using the forward slash here would be preferable

abhinavmuk04:
## Description
Make division by zero query failed errors more clear and ensuring standardization between different error messages for a singular exception

## Motivation and Context
fixed #23215

## Impact
Make debugging and error information more clear

## Test Plan
Existing and amended tests pass without regression

## Contributor checklist

- [x] CI passed.

## Release Notes
```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/type/BigintOperators.java
+++ b/presto-main/src/main/java/com/facebook/presto/type/BigintOperators.java
@@ -111,7 +111,7 @@ public final class BigintOperators
             return left / right;
         }
         catch (ArithmeticException e) {
-            throw new PrestoException(DIVISION_BY_ZERO, e);
+            throw new PrestoException(DIVISION_BY_ZERO, "Division by zero");
         }
     }
 
@@ -123,7 +123,7 @@ public final class BigintOperators
             return left % right;
         }
         catch (ArithmeticException e) {
-            throw new PrestoException(DIVISION_BY_ZERO, e);
+            throw new PrestoException(DIVISION_BY_ZERO, "Division by zero");
         }
     }
 
--- a/presto-docs/src/main/sphinx/functions/conditional.rst
+++ b/presto-docs/src/main/sphinx/functions/conditional.rst
@@ -159,7 +159,7 @@ Query failure without ``TRY``:
 
 .. code-block:: none
 
-    Query failed: / by zero
+    Query failed: Division by zero
 
 Default values with ``TRY`` and ``COALESCE``:
 
--- a/presto-main/src/main/java/com/facebook/presto/type/DoubleOperators.java
+++ b/presto-main/src/main/java/com/facebook/presto/type/DoubleOperators.java
@@ -88,7 +88,7 @@ public final class DoubleOperators
             return left / right;
         }
         catch (ArithmeticException e) {
-            throw new PrestoException(DIVISION_BY_ZERO, e);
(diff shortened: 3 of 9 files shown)


elharo (commented):
Looks good. Please double check that no other tests depend on the exact wording.

elharo (commented):
I think this is used in AbstractTestQueries too

presto-spi/src/main/java/com/facebook/presto/spi/PrestoException.java
```diff
@@ -45,12 +45,16 @@ public ErrorCode getErrorCode()
...
     {
         String message = super.getMessage();
+
         if (message == null && getCause() != null) {
             message = getCause().getMessage();
         }
         if (message == null) {
             message = errorCode.getName();
         }
+        if (message.equals("/ by zero")) {
```
elharo, line 55: This is an unusual place to patch this. Where is the division by zero and the arithmetic exception? That's where I'd expect the message to be set.

abhinavmuk04: Seems like all the other division by zero errors are handled in my change. The only other place where the code would be changed is arithmeticexception class (which is read only) or in abstract classes which wouldn't work

elharo: Where do we call the Guava method that initially creates this message and throws this exception? That's where I'd look.

abhinavmuk04: @elharo After further research, seems like the best quick fix is to patch it here and then make the PR in guava and wait for them to change it. All the error messages are coming from read only files

elharo (approved):

presto-main/src/main/java/com/facebook/presto/type/BigintOperators.java
```diff
@@ -111,7 +111,7 @@ public static long divide(@SqlType(StandardTypes.BIGINT) long left, @SqlType(Sta
             return left / right;
         }
         catch (ArithmeticException e) {
-            throw new PrestoException(DIVISION_BY_ZERO, e);
+            throw new PrestoException(DIVISION_BY_ZERO, "Division by zero");
```
feilong-liu, line 114: Personally I think stack trace is more helpful, and the error code name should be enough to convey the information of the added message

abhinavmuk04: @feilong-liu Makes sense, only thing is that when tests fail the error message is sometimes the first thing that is looked at and this can just improve overall clarity. What do you think?

feilong-liu: The bigger concern is that the stack trace coming with exception may be lost. Can you run a test query to check what the failure message and stack track will be like with this change?

abhinavmuk04: You are right, it can cause confusion this way

elharo: you can supply both the message and the underlying exception

PR #23245, opened 2024-07-18
Evaluate project node on values node

### LINKED ISSUE #23196: Pushdown projects into value node
When someone projects an expression from values node, we should be able to inline and simplify those into the value node itself. Today we don't do that:

```
presto:di> explain select substr(x, 1,2) from (values 'abcd', 'efgh') AS T(x);
                                                      Query Plan
----------------------------------------------------------------------------------------------------------------------
 - Output[PlanNodeId 8][_col0] => [substr:varchar(4)]
         Estimates: {source: CostBasedSourceInfo, rows: 2 (110B), cpu: 220.00, memory: 0.00, network: 0.00}
         _col0 := substr (1:16)
     - Project[PlanNodeId 4][projectLocality = LOCAL] => [substr:varchar(4)]
             Estimates: {source: CostBasedSourceInfo, rows: 2 (110B), cpu: 220.00, memory: 0.00, network: 0.00}
             substr := substr(field, BIGINT'1', BIGINT'2') (1:38)
         - LocalExchange[PlanNodeId 195][ROUND_ROBIN] () => [field:varchar(4)]
                 Estimates: {source: CostBasedSourceInfo, rows: 2 (110B), cpu: 110.00, memory: 0.00, network: 0.00}
             - Values[PlanNodeId 0] => [field:varchar(4)]
                     Estimates: {source: CostBasedSourceInfo, rows: 2 (110B), cpu: 0.00, memory: 0.00, network: 0.00}
                     (VARCHAR'abcd')
                     (VARCHAR'efgh')

(1 row)
```

So we should pushdown the project nodes into the values clause and simplfiy the expression so that other rules can take advantage of this. Rouhgly we want to make the above to be:

```
select x from (values  substr('abcd',1,2),  substr('efgh',1,2)) AS T(x)
```

CC: @feilong-liu

jackychen718:
## Description
Fix #23196

## Motivation and Context
When someone projects an expression from values node, we should be able to inline and simplify those into the value node itself. #23196

## Impact
Add another optimization rule: InlineProjectionsOnValues

## Test Plan
unit test to verify it works properly.

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Add session property ``inline_projections_on_values`` and configuration property ``optimizer.inline-projections-on-values`` to evaluate project node on values node :pr:`23245`.

```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java
@@ -317,6 +317,7 @@ public class FeaturesConfig
 
     private boolean useNewNanDefinition = true;
     private boolean warnOnPossibleNans;
+    private boolean isInlineProjectionsOnValuesEnabled;
 
     public enum PartitioningPrecisionStrategy
     {
@@ -3198,6 +3199,19 @@ public class FeaturesConfig
     public FeaturesConfig setWarnOnCommonNanPatterns(boolean warnOnPossibleNans)
     {
         this.warnOnPossibleNans = warnOnPossibleNans;
+        return this;
+    }
+
+    public boolean getInlineProjectionsOnValues()
+    {
+        return isInlineProjectionsOnValuesEnabled;
+    }
+
+    @Config("optimizer.inline-projections-on-values")
+    @ConfigDescription("Inline deterministic projections on values input")
+    public FeaturesConfig setInlineProjectionsOnValues(boolean isInlineProjectionsOnValuesEnabled)
+    {
+        this.isInlineProjectionsOnValuesEnabled = isInlineProjectionsOnValuesEnabled;
         return this;
     }
 }
--- a/presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java
+++ b/presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java
@@ -55,6 +55,7 @@ import static com.facebook.presto.SystemSessionProperties.ENABLE_INTERMEDIATE_AG
 import static com.facebook.presto.SystemSessionProperties.FIELD_NAMES_IN_JSON_CAST_ENABLED;
 import static com.facebook.presto.SystemSessionProperties.GENERATE_DOMAIN_FILTERS;
 import static com.facebook.presto.SystemSessionProperties.HASH_PARTITION_COUNT;
+import static com.facebook.presto.SystemSessionProperties.INLINE_PROJECTIONS_ON_VALUES;
 import static com.facebook.presto.SystemSessionProperties.ITERATIVE_OPTIMIZER_TIMEOUT;
 import static com.facebook.presto.SystemSessionProperties.JOIN_PREFILTER_BUILD_SIDE;
 import static com.facebook.presto.SystemSessionProperties.KEY_BASED_SAMPLING_ENABLED;
@@ -7712,5 +7713,37 @@ public abstract class AbstractTestQueries
                 .setSystemProperty(OPTIMIZE_HASH_GENERATION, optimizeHashGeneration)
                 .build();
         assertQuery(session, "SELECT DISTINCT x FROM (VALUES (REAL '0.0'), (REAL '-0.0')) t(x)", "SELECT  CAST(0.0 AS REAL)");
+    }
+
+    @Test
+    public void testEvaluateProjectOnValues()
+    {
+        Session session = Session.builder(getSession())
+                .setSystemProperty(INLINE_PROJECTIONS_ON_VALUES, "true")
+                .build();
+        assertQuery(session,
+                "SELECT MAP(ARRAY[1,2,3],ARRAY['one','two','three'])[x] from (values 1,2) as t(x)",
+                "SELECT * FROM (VALUES 'one','two')");
+        assertQuery(session,
+                "SELECT CASE WHEN x<y THEN x ELSE y END from (values (1,2),(3,4)) as t(x,y)",
+                "SELECT * FROM (VALUES 1,3)");
+        assertQuery(session,
+                "SELECT MAP(ARRAY[ARRAY[1,1],ARRAY[2,2]],ARRAY[1,2])[x] from (values ARRAY[1,1]) as t(x)",
+                "SELECT * FROM (VALUES 1)");
+        assertQuery(session,
+                "SELECT SUBSTR(Y,1,1) FROM (SELECT SUBSTR(X, 1,2) FROM (VALUES 'abcd', 'efgh') AS T(X)) AS T(Y)",
+                "SELECT * FROM (VALUES 'a','e')");
+        assertQuery(session,
+                "SELECT a * 2, a * 4 from (VALUES 2, 4) t(a)",
+                "SELECT * FROM (VALUES (4,8),(8,16))");
+        assertQuery(session,
+                "SELECT a + 1 FROM (VALUES (1, 2), (3, 4)) t(a, b)",
+                "SELECT * FROM (VALUES 2, 4)");
+        assertQuery(session,
+                "WITH t1 as (SELECT a + 1 as a FROM (VALUES 6) as t(a)) SELECT a * 2, a - 1 FROM t1",
+                "SELECT * FROM (VALUES (14, 6))");
+        assertQuery(session,
+                "SELECT a * 2, a - 1 FROM (SELECT x * 2 as a FROM (VALUES 15) t(x))",
+                "SELECT * FROM (VALUES (60, 29))");
     }
 }
--- a/presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
+++ b/presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
@@ -352,6 +352,7 @@ public final class SystemSessionProperties
     public static final String JOIN_PREFILTER_BUILD_SIDE = "join_prefilter_build_side";
     public static final String OPTIMIZER_USE_HISTOGRAMS = "optimizer_use_histograms";
     public static final String WARN_ON_COMMON_NAN_PATTERNS = "warn_on_common_nan_patterns";
+    public static final String INLINE_PROJECTIONS_ON_VALUES = "inline_projections_on_values";
 
     private final List<PropertyMetadata<?>> sessionProperties;
 
@@ -1970,6 +1971,10 @@ public final class SystemSessionProperties
                 booleanProperty(WARN_ON_COMMON_NAN_PATTERNS,
                         "Whether to give a warning for some common issues relating to NaNs",
                         featuresConfig.getWarnOnCommonNanPatterns(),
+                        false),
+                booleanProperty(INLINE_PROJECTIONS_ON_VALUES,
+                        "Whether to evaluate project node on values node",
+                        featuresConfig.getInlineProjectionsOnValues(),
                         false));
     }
 
@@ -3286,5 +3291,10 @@ public final class SystemSessionProperties
     public static boolean warnOnCommonNanPatterns(Session session)
     {
         return session.getSystemProperty(WARN_ON_COMMON_NAN_PATTERNS, Boolean.class);
+    }
+
+    public static boolean isInlineProjectionsOnValues(Session session)
+    {
+        return session.getSystemProperty(INLINE_PROJECTIONS_ON_VALUES, Boolean.class);
     }
 }
--- a/presto-main/src/main/java/com/facebook/presto/sql/planner/PlanOptimizers.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/planner/PlanOptimizers.java
@@ -45,6 +45,7 @@ import com.facebook.presto.sql.planner.iterative.rule.ImplementBernoulliSampleAs
 import com.facebook.presto.sql.planner.iterative.rule.ImplementFilteredAggregations;
 import com.facebook.presto.sql.planner.iterative.rule.ImplementOffset;
 import com.facebook.presto.sql.planner.iterative.rule.InlineProjections;
+import com.facebook.presto.sql.planner.iterative.rule.InlineProjectionsOnValues;
 import com.facebook.presto.sql.planner.iterative.rule.InlineSqlFunctions;
 import com.facebook.presto.sql.planner.iterative.rule.LeftJoinNullFilterToSemiJoin;
 import com.facebook.presto.sql.planner.iterative.rule.LeftJoinWithArrayContainsToEquiJoinCondition;
@@ -477,6 +478,15 @@ public class PlanOptimizers
                                 new TransformUncorrelatedInPredicateSubqueryToSemiJoin(),
                                 new TransformCorrelatedScalarAggregationToJoin(metadata.getFunctionAndTypeManager()),
                                 new TransformCorrelatedLateralJoinToJoin(metadata.getFunctionAndTypeManager()))),
+                new IterativeOptimizer(
+                        metadata,
+                        ruleStats,
+                        statsCalculator,
+                        estimatedExchangesCostCalculator,
+                        ImmutableSet.<Rule<?>>builder()
+                                .add(new InlineProjectionsOnValues(metadata.getFunctionAndTypeManager()))
+                                .addAll(new SimplifyRowExpressions(metadata).rules())
+                                .build()),
                 new IterativeOptimizer(
                         metadata,
                         ruleStats,
--- a/presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/InlineProjectionsOnValues.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/InlineProjectionsOnValues.java
@@ -1,0 +1,135 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.sql.planner.iterative.rule;
+
+import com.facebook.presto.Session;
+import com.facebook.presto.matching.Capture;
+import com.facebook.presto.matching.Captures;
+import com.facebook.presto.matching.Pattern;
+import com.facebook.presto.metadata.FunctionAndTypeManager;
+import com.facebook.presto.spi.plan.ProjectNode;
+import com.facebook.presto.spi.plan.ValuesNode;
+import com.facebook.presto.spi.relation.DeterminismEvaluator;
+import com.facebook.presto.spi.relation.RowExpression;
+import com.facebook.presto.spi.relation.VariableReferenceExpression;
+import com.facebook.presto.sql.planner.iterative.Rule;
+import com.facebook.presto.sql.relational.RowExpressionDeterminismEvaluator;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Streams;
+
+import java.util.AbstractMap.SimpleImmutableEntry;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+import static com.facebook.presto.SystemSessionProperties.isInlineProjectionsOnValues;
+import static com.facebook.presto.matching.Capture.newCapture;
+import static com.facebook.presto.sql.planner.RowExpressionVariableInliner.inlineVariables;
+import static com.facebook.presto.sql.planner.plan.Patterns.project;
+import static com.facebook.presto.sql.planner.plan.Patterns.source;
+import static com.facebook.presto.sql.planner.plan.Patterns.values;
+import static com.google.common.base.Verify.verify;
+import static com.google.common.collect.ImmutableList.toImmutableList;
+import static com.google.common.collect.ImmutableMap.toImmutableMap;
+import static java.util.Objects.requireNonNull;
+
+/**
+ * This optimizer looks for ProjectNode followed by a ValuesNode and get the ProjectNode Evaluated.
+ * When this rule is used on iterative optimizer, the rule could apply iteratively.
+ * <p/>
+ * Plan before optimizer:
+ * <pre>
+ * ProjectNode (outputVariables)
+ *   - ValuesNode
+ * </pre>
+ * <p/>
+ * Plan after optimizer:
+ * <pre>
+ * ValuesNode (outputVariables)
+ * </pre>
+ */
+public class InlineProjectionsOnValues
+        implements Rule<ProjectNode>
+{
+    private static final Capture<ValuesNode> CHILD = newCapture();
+
+    private static final Pattern<ProjectNode> PATTERN = project()
+            .with(source().matching(values().capturedAs(CHILD)));
+
+    private final FunctionAndTypeManager functionAndTypeManager;
+
+    public InlineProjectionsOnValues(FunctionAndTypeManager functionAndTypeManager)
+    {
+        this.functionAndTypeManager = requireNonNull(functionAndTypeManager, "functionManager is null");
+    }
+
+    @Override
+    public Pattern<ProjectNode> getPattern()
+    {
+        return PATTERN;
+    }
+
+    @Override
+    public boolean isEnabled(Session session)
+    {
+        return isInlineProjectionsOnValues(session);
+    }
+
+    @Override
+    public Result apply(ProjectNode projectNode, Captures captures, Context context)
+    {
+        ValuesNode source = captures.get(CHILD);
+        List<List<RowExpression>> rows = source.getRows();
+        List<VariableReferenceExpression> valuesOutputVariables = source.getOutputVariables();
+        List<VariableReferenceExpression> projectOutputVariables = projectNode.getOutputVariables();
+        List<RowExpression> projectRowExpressions = projectNode.getAssignments()
+                .getExpressions()
+                .stream()
+                .collect(toImmutableList());
+
+        // exclude non-deterministic function
+        DeterminismEvaluator determinismEvaluator = new RowExpressionDeterminismEvaluator(functionAndTypeManager);
+        if (!projectRowExpressions.stream().allMatch(determinismEvaluator::isDeterministic)) {
+            return Result.empty();
+        }
+        if (!rows.stream().allMatch(row -> row.stream()
+                .allMatch(determinismEvaluator::isDeterministic))) {
+            return Result.empty();
+        }
+
+        //rewrite ProjectNode assignment expressions
+        ImmutableList.Builder<List<RowExpression>> rowExpressionsListBuilder = ImmutableList.builder();
+        for (List<RowExpression> rowExpressions : rows) {
+            verify(rowExpressions.size() == valuesOutputVariables.size(), "Output variable does not match its value in ValuesNode");
+            Map<VariableReferenceExpression, RowExpression> mapping = Streams.zip(
+                            valuesOutputVariables.stream(),
+                            rowExpressions.stream(),
+                            SimpleImmutableEntry::new)
+                    .collect(toImmutableMap(Map.Entry::getKey, Map.Entry::getValue));
+            List<RowExpression> rowExpressionsInProject = projectRowExpressions.stream()
+                    .map(expression -> inlineVariables(mapping, expression))
+                    .collect(toImmutableList());
+            rowExpressionsListBuilder.add(rowExpressionsInProject);
+        }
+
+        ValuesNode updatedProject = new ValuesNode(
+                source.getSourceLocation(),
+                context.getIdAllocator().getNextId(),
+                projectOutputVariables,
+                rowExpressionsListBuilder.build(),
+                Optional.empty());
+
+        return Result.ofPlanNode(updatedProject);
+    }
+}
--- a/presto-main/src/test/java/com/facebook/presto/sql/analyzer/TestFeaturesConfig.java
+++ b/presto-main/src/test/java/com/facebook/presto/sql/analyzer/TestFeaturesConfig.java
@@ -277,7 +277,8 @@ public class TestFeaturesConfig
                 .setRemoveCrossJoinWithSingleConstantRow(true)
                 .setUseHistograms(false)
                 .setUseNewNanDefinition(true)
-                .setWarnOnCommonNanPatterns(false));
+                .setWarnOnCommonNanPatterns(false)
+                .setInlineProjectionsOnValues(false));
     }
 
     @Test
@@ -499,6 +500,7 @@ public class TestFeaturesConfig
                 .put("optimizer.use-histograms", "true")
                 .put("use-new-nan-definition", "false")
                 .put("warn-on-common-nan-patterns", "true")
+                .put("optimizer.inline-projections-on-values", "true")
                 .build();
 
         FeaturesConfig expected = new FeaturesConfig()
@@ -717,7 +719,8 @@ public class TestFeaturesConfig
                 .setRemoveCrossJoinWithSingleConstantRow(false)
                 .setUseHistograms(true)
                 .setUseNewNanDefinition(false)
-                .setWarnOnCommonNanPatterns(true);
+                .setWarnOnCommonNanPatterns(true)
+                .setInlineProjectionsOnValues(true);
         assertFullMapping(properties, expected);
     }
 
--- a/presto-main/src/test/java/com/facebook/presto/sql/planner/iterative/rule/TestInlineProjectionsOnValues.java
+++ b/presto-main/src/test/java/com/facebook/presto/sql/planner/iterative/rule/TestInlineProjectionsOnValues.java
@@ -1,0 +1,61 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.sql.planner.iterative.rule;
+
+import com.facebook.presto.spi.plan.ValuesNode;
+import com.facebook.presto.sql.planner.iterative.rule.test.BaseRuleTest;
+import org.testng.annotations.Test;
+
+import static com.facebook.presto.SystemSessionProperties.INLINE_PROJECTIONS_ON_VALUES;
+import static com.facebook.presto.common.type.BigintType.BIGINT;
+import static com.facebook.presto.sql.planner.assertions.PlanMatchPattern.node;
+import static com.facebook.presto.sql.planner.iterative.rule.test.PlanBuilder.assignment;
+import static com.facebook.presto.sql.relational.Expressions.call;
+
+public class TestInlineProjectionsOnValues
+        extends BaseRuleTest
+{
+    @Test
+    public void testDoesNotFireOn()
+    {
+        tester().assertThat(new InlineProjectionsOnValues(tester.getMetadata().getFunctionAndTypeManager()))
+                .setSystemProperty(INLINE_PROJECTIONS_ON_VALUES, "true")
+                .on(p -> p.project(p.project(p.values(p.getIdAllocator().getNextId(), p.variable("a")),
+                                assignment(p.variable("c"), p.variable("a"))),
+                        assignment(p.variable("d"), p.variable("c"))))
+                .doesNotFire();
+    }
+
+    @Test
+    public void testDoesNotFireOnWithNonDeterministicFunction()
+    {
+        tester().assertThat(new InlineProjectionsOnValues(tester.getMetadata().getFunctionAndTypeManager()))
+                .setSystemProperty(INLINE_PROJECTIONS_ON_VALUES, "true")
+                .on(p -> p.project(p.values(p.getIdAllocator().getNextId(), p.variable("a")),
+                        assignment(p.variable("b"), call(tester.getMetadata().getFunctionAndTypeManager(), "random", BIGINT, p.variable("a")))))
+                .doesNotFire();
+    }
+
+    @Test
+    public void testFireOnProjectFollowedByValues()
+    {
+        tester().assertThat(new InlineProjectionsOnValues(tester.getMetadata().getFunctionAndTypeManager()))
+                .setSystemProperty(INLINE_PROJECTIONS_ON_VALUES, "true")
+                //  Form the input planNode: ProjectNode -> ValuesNode
+                .on(p -> p.project(p.values(p.getIdAllocator().getNextId(), p.variable("a")),
+                        assignment(p.variable("c"), p.variable("a"))))
+                // Ensure the PlanNode is optimized to a ValuesNode
+                .matches(node(ValuesNode.class));
+    }
+}


kaikalur:
Please add some actual query tests with a bit more complex expressions like map/array constructors etc. which is quite common. Also add multiple columns tests.

jackychen718:
Fixed @kaikalur @ZacBlanco

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
```diff
@@ -0,0 +1,137 @@
...
+        implements Rule<ProjectNode>
+{
+    private static final Pattern<ProjectNode> PATTERN = project().with(source().matching(values()));
+
+    private final FunctionAndTypeManager functionAndTypeManager;
+
+    public EvaluateProjectOnValues(FunctionAndTypeManager functionAndTypeManager)
+    {
+        requireNonNull(functionAndTypeManager, "functionManager is null");
+        this.functionAndTypeManager = functionAndTypeManager;
```
ZacBlanco, line 73: nit: You can combine this into a single line
```suggestion
        this.functionAndTypeManager = requireNonNull(functionAndTypeManager, "functionManager is null");
```

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
```diff
@@ -0,0 +1,137 @@
...
+                RowExpressionInterpreter elementInterpreter = new RowExpressionInterpreter(element,
+                        functionAndTypeManager,
+                        context.getSession().toConnectorSession(),
+                        EVALUATED);
+                valuesMapForResolver.put(valuesOutputVariables.get(index).getName(), elementInterpreter.evaluate());
+            });
+            VariableResolver variableResolver = new Interpreters.LambdaVariableResolver(valuesMapForResolver);
+            //evaluate each row of the ProjectNode
+            List<Object> rowValue = new ArrayList<>();
+            for (RowExpression rowExpression : projectRowExpressions) {
```
ZacBlanco, line 115: I would pick a style (functional or imperative) and stick to it rather than mixing the two. For this loop you can instead do `projectRowExpressions.stream().map(...).collect(toImmutableList())`

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
```diff
@@ -0,0 +1,137 @@
...
+                RowExpressionInterpreter interpreter = new RowExpressionInterpreter(
+                        rowExpression,
+                        functionAndTypeManager,
+                        context.getSession().toConnectorSession(),
+                        OPTIMIZED);
+                rowValue.add(interpreter.optimize(variableResolver));
+            }
+            return rowValue;
+        }).collect(toList());
```
ZacBlanco, line 124:
```suggestion
        }).collect(toImmutableList());
```

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
```diff
@@ -0,0 +1,137 @@
...
+
+        //Form the ValuesNode transformed from ProjectNode
+        List<Type> types = projectNode.getOutputVariables().stream().map(RowExpression::getType).collect(toList());
```
ZacBlanco, line 127:
```suggestion
        List<Type> types = projectNode.getOutputVariables().stream().map(RowExpression::getType).collect(toImmutableList());
```

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
```diff
@@ -0,0 +1,137 @@
...
+        List<List<RowExpression>> valuesRows = rowValues.stream().map(row -> {
+            List<RowExpression> listRowExpression = new ArrayList<>();
+            IntStream.range(0, row.size()).forEach(index -> listRowExpression.add(new ConstantExpression(row.get(index), types.get(index))));
+            return listRowExpression;
+        }).collect(toList());
```
ZacBlanco, line 132:
```suggestion
        }).collect(toImmutableList());
```

presto-main/src/test/java/com/facebook/presto/sql/planner/iterative/rule/TestEvaluateProjectOnValues.java
```diff
@@ -0,0 +1,47 @@
...
+
+public class TestEvaluateProjectOnValues
+        extends BaseRuleTest
+{
+    @Test
+    public void testDoesNotFireOn()
+    {
+        tester().assertThat(new EvaluateProjectOnValues(tester.getMetadata().getFunctionAndTypeManager()))
+                .setSystemProperty(EVALUATE_PROJECT_ON_VALUES, "true")
+                .on(p -> p.project(p.project(p.values(p.getIdAllocator().getNextId(), 2, p.variable("a")),
```
ZacBlanco, line 32: it's difficult to tell what this test case is trying to accomplish. Can you write a comment?

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
```diff
@@ -0,0 +1,137 @@
...
+        List<List<Object>> rowValues = rows.stream().map(row -> {
+            //Prepare to set up variable resolver
+            Map<String, Object> valuesMapForResolver = new HashMap<>();
+            IntStream.range(0, row.size()).forEach(index -> {
+                RowExpression element = row.get(index);
+                RowExpressionInterpreter elementInterpreter = new RowExpressionInterpreter(element,
+                        functionAndTypeManager,
+                        context.getSession().toConnectorSession(),
+                        EVALUATED);
+                valuesMapForResolver.put(valuesOutputVariables.get(index).getName(), elementInterpreter.evaluate());
```
ZacBlanco, line 110: We should probably verify at the beginning of the loop that `valuesOutputVariables` size is equal to each row's inner list size.

ZacBlanco (commented):
Great job! Minor nits on my first pass

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
```diff
@@ -0,0 +1,137 @@
...
+        ValuesNode source = (ValuesNode) optionalSource.get();
+        Collection<RowExpression> projectRowExpressions = projectNode.getAssignments().getExpressions();
+        List<List<RowExpression>> rows = source.getRows();
+        List<VariableReferenceExpression> valuesOutputVariables = source.getOutputVariables();
+
+        //evaluate ProjectNode
+        List<List<Object>> rowValues = rows.stream().map(row -> {
+            //Prepare to set up variable resolver
+            Map<String, Object> valuesMapForResolver = new HashMap<>();
+            IntStream.range(0, row.size()).forEach(index -> {
```
ZacBlanco, line 104: maybe consider using `Streams.zip` here instead? e.g.

```
Streams.zip(row.stream(), valuesOutputVariables.stream(), (column, variable) -> {
    // logic
})
```

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
```diff
@@ -0,0 +1,137 @@
...
+                rowValue.add(interpreter.optimize(variableResolver));
+            }
+            return rowValue;
+        }).collect(toList());
+
+        //Form the ValuesNode transformed from ProjectNode
+        List<Type> types = projectNode.getOutputVariables().stream().map(RowExpression::getType).collect(toList());
+        List<List<RowExpression>> valuesRows = rowValues.stream().map(row -> {
+            List<RowExpression> listRowExpression = new ArrayList<>();
+            IntStream.range(0, row.size()).forEach(index -> listRowExpression.add(new ConstantExpression(row.get(index), types.get(index))));
```
ZacBlanco, line 130: another place to consider `Streams.zip`

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
kaikalur: I agree though I think a judicious mix of things is ok.

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
```diff
@@ -0,0 +1,137 @@
...
+import static com.facebook.presto.sql.planner.plan.Patterns.values;
+import static java.util.Objects.requireNonNull;
+import static java.util.stream.Collectors.toList;
+
+/**
+ * This optimizer looks for ProjectNode followed by a ValuesNode and get the ProjectNode Evaluated.
+ * <p/>
+ * Plan before optimizer:
+ * <pre>
+ * ProjectNode (outputVariables)
```
kaikalur, line 54: How about multiple projects - it could happen. Will this rule be applied again?

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
```diff
@@ -0,0 +1,137 @@
...
+        //evaluate ProjectNode
+        List<List<Object>> rowValues = rows.stream().map(row -> {
+            //Prepare to set up variable resolver
+            Map<String, Object> valuesMapForResolver = new HashMap<>();
+            IntStream.range(0, row.size()).forEach(index -> {
+                RowExpression element = row.get(index);
+                RowExpressionInterpreter elementInterpreter = new RowExpressionInterpreter(element,
+                        functionAndTypeManager,
+                        context.getSession().toConnectorSession(),
+                        EVALUATED);
```
kaikalur, line 109: Does this handle non-deterministic functions? I think we should not apply this rule for non-deterministic like 

select random(x) from (values 1,2) AS T(x)

kaikalur (changes requested):

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
ZacBlanco: What's the reasoning not evaluate a non-deterministic function? Is there a big difference evaluating it on a worker later on vs the coordinator?

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
jackychen718: fixed

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
jackychen718: I refactored a little bit. Hope it is better structured.

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
jackychen718: fixed.

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
jackychen718: fixed.

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
jackychen718: fixed

presto-main/src/test/java/com/facebook/presto/sql/planner/iterative/rule/TestEvaluateProjectOnValues.java
jackychen718: Sure

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
jackychen718: OK.

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
kaikalur: In general there could be surprises so we generally short-circuit lot of optimizations when we see non-deterministic expressions. Also I'm not a fan of RowExpressionInterpreter :( it's implementing the "engine" for constants and I'm sure it works differently from the java and differently differently from native lol

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
kaikalur: Add a test case for two projections for completeness?

kaikalur (approved):

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
jackychen718: sure

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
jackychen718: non-deterministic function excluded.

kaikalur: please add a test like: select random(x) from (values 1,2) AS T(x)

presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java
```diff
@@ -3200,4 +3201,17 @@ public FeaturesConfig setWarnOnCommonNanPatterns(boolean warnOnPossibleNans)
...
         return this;
     }
+
+    public boolean getEvaluateProjectOnValues()
+    {
+        return isEvaluateProjectOnValuesEnabled;
+    }
+
+    @Config("optimizer.evaluate-project-on-values")
+    @ConfigDescription("Evaluate ProjectNode from ValuesNode")
```
ZacBlanco, line 3211:
```suggestion
    @ConfigDescription("evaluate deterministic projections on VALUES inputs")
```

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
jackychen718: Fixed.

presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java
jackychen718: ok.

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
jackychen718: fixed

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
jackychen718: fixed

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
```diff
@@ -0,0 +1,142 @@
...
+        List<List<RowExpression>> rows = source.getRows();
+        List<VariableReferenceExpression> valuesOutputVariables = source.getOutputVariables();
+        Set<Map.Entry<VariableReferenceExpression, RowExpression>> projectAssignmentEntries = projectNode.getAssignments().entrySet();
+        List<VariableReferenceExpression> projectOutputVariables = projectAssignmentEntries.stream().map(Map.Entry::getKey).collect(toImmutableList());
+        List<RowExpression> projectRowExpressions = projectAssignmentEntries.stream().map(Map.Entry::getValue).collect(toImmutableList());
+        // exclude non-deterministic function
+        if (!projectRowExpressions.stream()
+                .filter(expression -> expression instanceof CallExpression)
+                .map(callExpression -> functionAndTypeManager.getFunctionMetadata(((CallExpression) callExpression).getFunctionHandle()))
+                .filter(functionMetadata -> !functionMetadata.isDeterministic()).collect(toImmutableList()).isEmpty()) {
```
ZacBlanco, line 106: can be replaced with `.allMatch(FunctionMetadata::isDeterministic)`

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
```diff
@@ -0,0 +1,142 @@
...
+                            rowExpression,
+                            functionAndTypeManager,
+                            context.getSession().toConnectorSession(),
+                            OPTIMIZED).optimize(variableResolver))
+                    .collect(toImmutableList());
+        }).collect(toImmutableList());
+
+        //Form the ValuesNode transformed from ProjectNode
+        List<List<RowExpression>> valuesRows = rowValues.stream().map(row ->
+                        Streams.zip(row.stream(), projectOutputVariables.stream(), (elementValue, projectOutputVariable) -> (RowExpression) new ConstantExpression(elementValue, projectOutputVariable.getType()))
```
ZacBlanco, line 135: This line is very long. Can you put the arguments on new lines?

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java
```diff
@@ -7712,4 +7713,24 @@ public void testRealDistinctPositiveAndNegativeZero(String optimizeHashGeneratio
...
+        assertQuery(session,
+                "SELECT CASE WHEN x<y THEN x ELSE y END from (values (1,2),(3,4)) as t(x,y)",
+                "SELECT * FROM (VALUES 1,3)");
+        assertQuery(session,
+                "SELECT MAP(ARRAY[ARRAY[1,1],ARRAY[2,2]],ARRAY[1,2])[x] from (values ARRAY[1,1]) as t(x)",
+                "SELECT * FROM (VALUES 1)");
+        assertQuery(session,
+                "SELECT SUBSTR(Y,1,1) FROM (SELECT SUBSTR(X, 1,2) FROM (VALUES 'abcd', 'efgh') AS T(X)) AS T(Y)",
+                "SELECT * FROM (VALUES 'a','e')");
+    }
```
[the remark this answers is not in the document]
ZacBlanco: Also try with subquery expressions

```sql
WITH t1 as (SELECT a + 1 as a FROM (VALUES 6) as t(a)) SELECT a * 2, a - 1 FROM t1;
```

ZacBlanco (commented):
last set of comments

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
```diff
@@ -0,0 +1,142 @@
...
+import static com.facebook.presto.SystemSessionProperties.isEvaluateProjectOnValues;
+import static com.facebook.presto.spi.relation.ExpressionOptimizer.Level.EVALUATED;
+import static com.facebook.presto.spi.relation.ExpressionOptimizer.Level.OPTIMIZED;
+import static com.facebook.presto.sql.planner.plan.Patterns.project;
+import static com.facebook.presto.sql.planner.plan.Patterns.source;
+import static com.facebook.presto.sql.planner.plan.Patterns.values;
+import static com.google.common.base.Verify.verify;
+import static com.google.common.collect.ImmutableMap.toImmutableMap;
+import static java.util.Objects.requireNonNull;
+import static org.glassfish.jersey.internal.util.collection.ImmutableCollectors.toImmutableList;
```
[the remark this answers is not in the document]
jackychen718: Fixed.

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java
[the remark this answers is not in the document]
jackychen718: ok.

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
jackychen718: fixed.

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
jackychen718: sure

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
```diff
@@ -0,0 +1,144 @@
...
+                    }).collect(toImmutableMap(Map.Entry::getKey, Map.Entry::getValue));
+            VariableResolver variableResolver = new Interpreters.LambdaVariableResolver(valuesMapForResolver);
+            //evaluate each row of the ProjectNode
+            return projectRowExpressions.stream().map(rowExpression -> new RowExpressionInterpreter(
+                            rowExpression,
+                            functionAndTypeManager,
+                            context.getSession().toConnectorSession(),
+                            OPTIMIZED).optimize(variableResolver))
+                    .collect(toImmutableList());
+        }).collect(toImmutableList());
```
feilong-liu, line 132: This part seems a bit complex and I have a hard time to understand it. Can you check the `ExpressionRewriter` in `CommonSubExpressionRewriter` class, and see if you can rewrite the expression in project assignments with values in Values node, and it should be later evaluated by `RowExpressionOptimizer`.

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
```diff
@@ -0,0 +1,144 @@
...
+            return Result.empty();
+        }
+        ValuesNode source = (ValuesNode) optionalSource.get();
+        List<List<RowExpression>> rows = source.getRows();
+        List<VariableReferenceExpression> valuesOutputVariables = source.getOutputVariables();
+        Set<Map.Entry<VariableReferenceExpression, RowExpression>> projectAssignmentEntries = projectNode.getAssignments().entrySet();
+        List<VariableReferenceExpression> projectOutputVariables = projectAssignmentEntries.stream().map(Map.Entry::getKey).collect(toImmutableList());
+        List<RowExpression> projectRowExpressions = projectAssignmentEntries.stream().map(Map.Entry::getValue).collect(toImmutableList());
+        // exclude non-deterministic function
+        if (!projectRowExpressions.stream()
```
feilong-liu, line 104: Should also check deterministic for the expressions in Values node

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
```diff
@@ -0,0 +1,144 @@
...
+        Optional<PlanNode> optionalSource = context.getLookup().resolveGroup(projectNode.getSource()).findFirst();
+        if (!optionalSource.isPresent()) {
+            return Result.empty();
+        }
+        ValuesNode source = (ValuesNode) optionalSource.get();
+        List<List<RowExpression>> rows = source.getRows();
+        List<VariableReferenceExpression> valuesOutputVariables = source.getOutputVariables();
+        Set<Map.Entry<VariableReferenceExpression, RowExpression>> projectAssignmentEntries = projectNode.getAssignments().entrySet();
+        List<VariableReferenceExpression> projectOutputVariables = projectAssignmentEntries.stream().map(Map.Entry::getKey).collect(toImmutableList());
+        List<RowExpression> projectRowExpressions = projectAssignmentEntries.stream().map(Map.Entry::getValue).collect(toImmutableList());
```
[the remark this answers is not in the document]
jackychen718: Althought I haven`t found an example,is it possible that projectOutputVariables and projectRowExpressions does not match because projectNode.getAssignments().getExpressions() returns a Collection of RowExpression rather than a List of  RowExpression?

kaikalur:
I'm actually wondering if we should just fireup a localqueryrunner and get the results (after checking there are no non-deterministic expressions in the whole subtree) and create values node from it

jackychen718:
It is a good idea but I have difficulty to use localqueryrunner to evaluate a PlanNode. Mostly,it is use to execute a SQL string. Could you please give me some guidance?

kaikalur:
Looking at:

https://github.com/prestodb/presto/blob/e884a445f27fb86337bdb9e5a9ea8f4e413fb5ba/presto-main/src/main/java/com/facebook/presto/testing/LocalQueryRunner.java#L1143

And also executeInternal that just gets a plan executes it, looks like you can try and just refactor that out intp executePlan method.

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
```diff
@@ -0,0 +1,144 @@
...
+        // exclude non-deterministic function
+        if (!projectRowExpressions.stream()
+                .filter(expression -> expression instanceof CallExpression)
+                .map(callExpression -> functionAndTypeManager.getFunctionMetadata(((CallExpression) callExpression).getFunctionHandle()))
+                .allMatch(FunctionMetadata::isDeterministic)) {
```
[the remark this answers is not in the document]
jackychen718: fixed

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
jackychen718: I used _ExpressionRewriter_ to rewrite the expression in ProjectNode. However, evaluating _ProjectNode_ requires evaluating _ValuesNode_ first to provide the resolver to the _optimize_ function of _RowExpressionOptimizer_. The overall implementation does not simplify too much.

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
jackychen718: fixed

ZacBlanco:
I can see how using a LocalQueryRunner could be beneficial because if you have an entirely deterministic subplan, you can optimize it entirely in the coordinator without having to juggle with the expressions or other plan nodes. However, I think we should keep the scope of this PR small. Using the LocalQueryRunner will be quite a large change since it also has to be moved out of the test-scope within `presto-main`. It will probably bring some other headaches, not to mention the actual latency of using the query runner within the rule and figuring out whether it's acceptable or not for the optimizer.

The `RowExpressionInterpreter`, at least on the Java side, uses the exact same generated bytecode that the the workers use to evaluate the functions, so I don't think there should be any worry about differences between execution results if you're sticking to Java.

Another issue that could occur is the coordinator's expression evaluation result diverging on native and java clusters. I know there have been some discrepancies between results of native and Java function implementations that we've been trying to resolve, but I believe they are minimal, if any still exist. If the function is one of the known ones with diverging behavior, then there could be issues when this rule is applied. However, you would still run into this problem if you were to use the LocalQueryRunner since it is Java-based.

The way I see it, the headache of the `RowExpressionInterpreter` is the complexity of using it within the rule to evaluate all of the `VALUES` and project expressions and having to juggle those around. If we confine the scope for this PR to match only `Project->Values` I think the complexity is acceptable. We could open another issue for evaluating entirely deterministic sub-plans.

kaikalur:
> The RowExpressionInterpreter, at least on the Java side, uses the exact same generated bytecode that the the workers use to evaluate the functions, so I don't think there should be any worry about differences between execution results if you're sticking to Java.

Hmm not exactly! It has things like CASE implemented internally. Not sure if it follows all the subtle semantics, overflow things etc

kaikalur:
> > The RowExpressionInterpreter, at least on the Java side, uses the exact same generated bytecode that the the workers use to evaluate the functions, so I don't think there should be any worry about differences between execution results if you're sticking to Java.
> 
> Hmm not exactly! It has things like CASE implemented internally. Not sure if it follows all the subtle semantics, overflow things etc

 But I agree - we can keep this PR simple. But we should explore the localqueryrunner idea further

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
ZacBlanco: If I'm understanding correctly, I think what we actually want here is the `RowExpressionVariableInliner`? The `ExpressionRewriter` replaces a `RowExpression` with a `VariableReferenceExpression`. I think we want the other way around when constructing expressions to evaluate. If we can re-write the projection input expression to replace the reference to the reference coming from the ValuesNode, then we only need to run the Interpreter once. Is that correct @feilong-liu ?

ZacBlanco:
> > The RowExpressionInterpreter, at least on the Java side, uses the exact same generated bytecode that the the workers use to evaluate the functions, so I don't think there should be any worry about differences between execution results if you're sticking to Java.
> 
> Hmm not exactly! It has things like CASE implemented internally. Not sure if it follows all the subtle semantics, overflow things etc

Thanks for the pointer. I was not aware of that. Though, if we're not confident in it, then we should probably either write some more test cases which do make us confident in the behavior, or remove it entirely to opt for something like LocalQueryRunner which uses the execution codepath.

kaikalur:
> > > The RowExpressionInterpreter, at least on the Java side, uses the exact same generated bytecode that the the workers use to evaluate the functions, so I don't think there should be any worry about differences between execution results if you're sticking to Java.
> > 
> > 
> > Hmm not exactly! It has things like CASE implemented internally. Not sure if it follows all the subtle semantics, overflow things etc
> 
> Thanks for the pointer. I was not aware of that. Though, if we're not confident in it, then we should probably either write some more test cases which do make us confident in the behavior, or remove it entirely to opt for something like LocalQueryRunner which uses the execution codepath.

Yeah we could write some. But IMO there should be one way to execute things and constant folding etc should just use that. LocalQueryRunner will be quite good to have at some point as a cleanup.

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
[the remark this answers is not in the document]
feilong-liu: I think it should be fine. The map in assignment is backed by unmodifiable LinkedHashMap which is order preserving.

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
feilong-liu: @ZacBlanco You are right, we need to use RowExpressionVariableInliner here.

@jackychen718  Can you simplify the logic here with the RowExpressionVariableInliner?

Also is it possible to get rid of the RowExpressionInterpreter in this optimizer? I think the RowExpressionOptimizer which ran later should be able to evaluate it (worth to try e2e to verify)

kaikalur:
> > > The RowExpressionInterpreter, at least on the Java side, uses the exact same generated bytecode that the the workers use to evaluate the functions, so I don't think there should be any worry about differences between execution results if you're sticking to Java.
> > 
> > 
> > Hmm not exactly! It has things like CASE implemented internally. Not sure if it follows all the subtle semantics, overflow things etc
> 
> Thanks for the pointer. I was not aware of that. Though, if we're not confident in it, then we should probably either write some more test cases which do make us confident in the behavior, or remove it entirely to opt for something like LocalQueryRunner which uses the execution codepath.

Just as we are talking about it lol probably in rowexpressioninterpreter:
#23281#issuecomment-2246626612

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
jackychen718: fixed

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
```diff
@@ -0,0 +1,154 @@
...
+        if (!rows.stream().allMatch(row -> row.stream()
+                .allMatch(determinismEvaluator::isDeterministic))) {
+            return Result.empty();
+        }
+
+        ExpressionOptimizer expressionOptimizer = new RowExpressionOptimizer(metadata);
+        //Rewrite ProjectNode Expression
+        Map<RowExpression, VariableReferenceExpression> mapForRowExpressionRewriter = source.getOutputVariables().stream()
+                .map(variableReferenceExpression ->
+                        new AbstractMap.SimpleImmutableEntry<RowExpression, VariableReferenceExpression>((RowExpression) variableReferenceExpression, variableReferenceExpression))
```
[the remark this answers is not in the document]
jackychen718: fixed

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
```diff
@@ -0,0 +1,154 @@
...
+        implements Rule<ProjectNode>
+{
+    private static final Pattern<ProjectNode> PATTERN = project().with(source().matching(values()));
+
+    private final Metadata metadata;
+    private final FunctionAndTypeManager functionAndTypeManager;
+
+    public EvaluateProjectOnValues(Metadata metadata)
+    {
+        this.metadata = requireNonNull(metadata, "MetaData is null");
```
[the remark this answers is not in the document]
jackychen718: fixed

tdcmeehan:
Using the `ExpressionOptimizer` should always be consistent with evaluating on the worker.  As @ZacBlanco mentioned, if we are worried these are inconsistent, then we should write a test case that proves they're consistent.  The rules are extensive, but not that complicated, and are already well tested, so I think the burden of proof is to demonstrate when they're not consistent.  And anyway, for [RFC-0004](https://github.com/prestodb/rfcs/pull/13), we'll be adding an expression fuzzer which should for all practical purposes prove that constant folding is 100% consistent with evaluation on the workers.

I don't like the idea of executing queries on a `LocalQueryRunner` because we are working towards a future where C++ is our default mode of deployment, and this would bake in Java inconsistencies into that query runner.  But the idea is potentially still sound.  I propose that we could simply use the [Presto sidecar](https://github.com/prestodb/rfcs/blob/main/RFC-0003-native-spi.md#coordinator-sidecar).  This would both solve the consistency problem, and also open new optimization opportunities, at no additional cost since we're already bearing the cost of deploying the sidecar.

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
```diff
@@ -0,0 +1,150 @@
...
+        List<List<RowExpression>> rowExpressionsList = rows.stream().map(row -> {
+            verify(row.size() == valuesOutputVariables.size(), "Output variable does not match its value in ValuesNode");
+            Map<VariableReferenceExpression, RowExpression> mapping = Streams.zip(valuesOutputVariables.stream(),
+                    row.stream(),
+                    SimpleImmutableEntry::new)
+                    .collect(toImmutableMap(Map.Entry::getKey, Map.Entry::getValue));
+            return projectRowExpressions.stream()
+                .map(expression -> inlineVariables(mapping, expression))
+                .collect(toImmutableList());
+        }).collect(Collectors.toList());
```
tdcmeehan, line 125:
```suggestion
            Map<VariableReferenceExpression, RowExpression> mapping = Streams.zip(
                    valuesOutputVariables.stream(),
                    row.stream(),
                    SimpleImmutableEntry::new)
                    .collect(toImmutableMap(Map.Entry::getKey, Map.Entry::getValue));
            return projectRowExpressions.stream()
                .map(expression -> inlineVariables(mapping, expression))
                .collect(toImmutableList());
        }).collect(toImmutableList());
```

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
```diff
@@ -0,0 +1,150 @@
...
+        //Form the ValuesNode transformed from ProjectNode
+        List<List<RowExpression>> valuesRows = rowValues.stream().map(row -> Streams.zip(row.stream(),
+                projectOutputVariables.stream(),
+                (elementValue, projectOutputVariable) -> (RowExpression) new ConstantExpression(elementValue, projectOutputVariable.getType()))
+                .collect(toImmutableList()))
+                .collect(toImmutableList());
+        ValuesNode updatedProject = new ValuesNode(source.getSourceLocation(),
+                context.getIdAllocator().getNextId(),
+                projectOutputVariables, valuesRows,
+                Optional.empty());
```
tdcmeehan, line 146:
```suggestion
        ValuesNode updatedProject = new ValuesNode(
                source.getSourceLocation(),
                context.getIdAllocator().getNextId(),
                projectOutputVariables,
                valuesRows,
                Optional.empty());
```

tdcmeehan (changes requested):

kaikalur:
> I don't like the idea of executing queries on a `LocalQueryRunner` because we are working towards a future where C++ is our default mode of deployment, and this would bake in Java inconsistencies into that query runner.

That's already the case between the interpreter and runtime! We should be principled in our eval. Native cannot come in the way of that

tdcmeehan:
We are working on resolving these inconsistencies.  Keeping a forked eval will lock in the inconsistencies forever.

kaikalur:
> We are working on resolving these inconsistencies. Keeping a forked eval will lock in the inconsistencies forever.

Why? Eval is eval is eval so if/when we switch it should be fixed.

kaikalur:
> @kaikalur so this point, eval is a leaky abstraction, and impacts the runtime and the optimizer. As Presto moves from Java to C++, we are abstracting the leaky parts into plugins, and these plugins are implemented differently between the Java eval and the C++ eval. This would be considered to be a new leak in our eval abstraction, and we could hide this behind a plugin if we chose. The underlying implementation could use something like a local query runner for the Java eval, and perhaps the sidecar for C++. The important thing is that we simply don't hardcode the LocalQueryRunner, but so long as we don't do that and we are principled around the design of the plugin, then we could do it two different ways for both eval engines.

Mechanism aside my bigger issue is we should have a single way to eavaluate something - I dont like the word plugin. Its integral to Presto it should be core.

tdcmeehan:
Presto doesn't work without plugins.  Some plugins are enabled by default, and some can be overridden.  Core plugins are enabled by default--for example, the system connector, the built in function namespace manager.  It's no different.

kaikalur:
> Presto doesn't work without plugins. Some plugins are enabled by default, and some can be overridden. Core plugins are enabled by default--for example, the system connector, the built in function namespace manager. It's no different. Plugins abstracting the eval are enabled by default, but can be overridden for early adopters of C++. Over time, C++ will become the default.

I dont consider builtin function namespace manager as core. Eval *should not* be a plugin. Its like the kernel it does basic things and implements SQL semantics which plugins are free to break lol

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
```diff
@@ -0,0 +1,150 @@
...
+            return Result.empty();
+        }
+        if (!rows.stream().allMatch(row -> row.stream()
+                .allMatch(determinismEvaluator::isDeterministic))) {
+            return Result.empty();
+        }
+        ExpressionOptimizer expressionOptimizer = new RowExpressionOptimizer(metadata);
+
+        //rewrite ProjectNode assignment expressions
+        List<List<RowExpression>> rowExpressionsList = rows.stream().map(row -> {
```
tdcmeehan, line 116: I see we're operating over lists of lists for the following three blocks.  Can the outer list be converted to a for loop, and within this loop we operate over a single list?  I.e. can we unnest this?  It would simplify this code.

kaikalur:
> I don't know why you don't consider the built in function namespace manager a plugin, it implements `FunctionNamespaceManager` (which is a plugin) and is instantiated alongside other function namespace managers. It is both core and a plugin. Same for the system connector, for our block encodings, types, and resource groups. Whatever is core is defined by us and what is loaded by default, so I wouldn't place so much emphasis on the word plugin--think rather in terms of outcomes.

The way I see "plugins" - they are there to suppor different behavior from core and something like builtin functions are core.

kaikalur:
> > I don't know why you don't consider the built in function namespace manager a plugin, it implements `FunctionNamespaceManager` (which is a plugin) and is instantiated alongside other function namespace managers. It is both core and a plugin. Same for the system connector, for our block encodings, types, and resource groups. Whatever is core is defined by us and what is loaded by default, so I wouldn't place so much emphasis on the word plugin--think rather in terms of outcomes.
> 
> The way I see "plugins" - they are there to suppor different behavior from core and something like builtin functions are core.

In fact, we have other hacks like "metadata query optimization" which eval even more differently. For the longterm project health, there should be one and only one way to eval the core sql constructs. Yes connector plugins do their own thing - that also bothers me. We need to find a way to make all these use the same eval engine consistently.

tdcmeehan:
> > > I don't know why you don't consider the built in function namespace manager a plugin, it implements `FunctionNamespaceManager` (which is a plugin) and is instantiated alongside other function namespace managers. It is both core and a plugin. Same for the system connector, for our block encodings, types, and resource groups. Whatever is core is defined by us and what is loaded by default, so I wouldn't place so much emphasis on the word plugin--think rather in terms of outcomes.
> > 
> > 
> > The way I see "plugins" - they are there to suppor different behavior from core and something like builtin functions are core.
> 
> In fact, we have other hacks like "metadata query optimization" which eval even more differently. For the longterm project health, there should be one and only one way to eval the core sql constructs. Yes connector plugins do their own thing - that also bothers me. We need to find a way to make all these use the same eval engine consistently.

@kaikalur we have been discussing exactly this at the native worker working group meetings, perhaps let's brainstorm there and see if our proposal matches the outcomes you are expecting.  https://calendar.google.com/calendar/u/0/embed?src=linuxfoundation.org_vrjlva5b0u73ps75fvnv5sasi4@group.calendar.google.com&ctz=America/Los_Angeles

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
jackychen718: fixed

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
```diff
@@ -0,0 +1,165 @@
...
+        List<List<RowExpression>> rowExpressionsListInValuesNode = new ArrayList<>();
+        for (List<Object> rowValues : rowValuesList) {
+            List<RowExpression> rowExpressionsInValuesNode = Streams.zip(
+                    rowValues.stream(),
+                    projectOutputVariables.stream(),
+                    (elementValue, projectOutputVariable) ->
+                        (RowExpression) new ConstantExpression(elementValue, projectOutputVariable.getType())
+                ).collect(toImmutableList());
+            rowExpressionsListInValuesNode.add(rowExpressionsInValuesNode);
+        }
```
tdcmeehan, line 154: Can't these three individual for loops be collapsed into a single for loop?

jackychen718: Sorry for the misunderstanding.

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
jackychen718: fixed. @tdcmeehan

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/EvaluateProjectOnValues.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    @Override
+    public Result apply(ProjectNode projectNode, Captures captures, Context context)
+    {
+        Optional<PlanNode> optionalSource = context.getLookup().resolveGroup(projectNode.getSource()).findFirst();
+        if (!optionalSource.isPresent()) {
+            return Result.empty();
+        }
+        ValuesNode source = (ValuesNode) optionalSource.get();
```
feilong-liu, line 92: Check the example here https://github.com/prestodb/presto/blob/c9d1e8a382cf0ad25e28f94b8f81821b65769e0a/presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/InlineProjections.java#L57
It will help to simplify the code here

tdcmeehan:
Am I correct in reading the current version of the code just inlines the projections, and delegates the evaluation to something else like `SimplifyRowExpressions`?  If so, should we rename the optimizer rule to `InlineProjectionsOnValues` or something similar?

jackychen718:
Changed the rule to be `InlineProjectionsOnValues`

tdcmeehan (approved):
LGTM

tdcmeehan:
@jackychen718 feel free to merge this PR.

jackychen718:
@elharo @ClarenceThreepwood @jaystarshot @presto-oss Please review the pr to merge.

elharo (approved):

Issue #23246, opened 2024-07-18
Session is completely corrupted by the failed statement in a non-autocommit transaction


hantangwangd:
Sometimes we want to manually start a transaction and execute some statements in it, then finally uniformly commit the transaction.

However, if any one of the statements failed when executing, it would cause a completely corruption of current session. That is, we can't execute any SQLs any more even including `rollback` because of the exception as follows:

```
Current transaction is aborted, commands ignored until end of transaction block
```

Currently, we can do noting but quit and reconnect to use a new session.

## Expected Behavior

Enable to successfully execute `rollback` to end the transaction block

## Current Behavior

Execution of `rollback` fails because of the current transaction being aborted

## Possible Solution

Do not check transaction's state when executing `rollback`.

## Steps to Reproduce

1. Open a CLI console, and use hive connector, for example:

```
use hive.default;
```

2. Execute the following statements:

```
set role admin;        // To make sure the problem always shows up

create table test_non_autocommit_table (a bigint, b double, c varchar) with(partitioned_by = array['c']);

start transaction;

insert into test_non_autocommit_table values(1, 1.1, '1001'), (2, 1.2, '1002');

// would fail because inserting into the same partition in the same transaction
insert into test_non_autocommit_table values(3, 1.3, '1002');
```

3. After the fail insertion, now we can not execute any sql because of `Current transaction is aborted ...`.

PR #23247, opened 2024-07-18
Fix rollback after statement fail in non-autocommit transaction

hantangwangd:
## Description

This PR fix the session corruption cause by a failed statement in non-autocommit transaction, enable executing `rollback` to end the aborted transaction block.

## Motivation

Fix: #23246 

## Test Plan

 - Newly added test cases in hive and iceberg to simulate the scenarios described in #23246
 - Manually confirm the fix in CLI console

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-hive/src/main/java/com/facebook/presto/hive/security/SqlStandardAccessControl.java
+++ b/presto-hive/src/main/java/com/facebook/presto/hive/security/SqlStandardAccessControl.java
@@ -656,20 +656,22 @@ public class SqlStandardAccessControl
     @Override
     public void checkCanSetRole(ConnectorTransactionHandle transaction, ConnectorIdentity identity, AccessControlContext context, String role, String catalogName)
     {
-        SemiTransactionalHiveMetastore metastore = getMetastore(transaction);
-        MetastoreContext metastoreContext = new MetastoreContext(
-                identity, context.getQueryId().getId(),
-                context.getClientInfo(),
-                context.getClientTags(),
-                context.getSource(),
-                Optional.empty(),
-                false,
-                HiveColumnConverterProvider.DEFAULT_COLUMN_CONVERTER_PROVIDER,
-                context.getWarningCollector(),
-                context.getRuntimeStats());
-        if (!isRoleApplicable(metastore, identity, new PrestoPrincipal(USER, identity.getUser()), metastoreContext, role)) {
-            denySetRole(role);
-        }
+        Optional<SemiTransactionalHiveMetastore> metastoreOptional = getMetastore(transaction);
+        metastoreOptional.ifPresent(metastore -> {
+            MetastoreContext metastoreContext = new MetastoreContext(
+                    identity, context.getQueryId().getId(),
+                    context.getClientInfo(),
+                    context.getClientTags(),
+                    context.getSource(),
+                    Optional.empty(),
+                    false,
+                    HiveColumnConverterProvider.DEFAULT_COLUMN_CONVERTER_PROVIDER,
+                    context.getWarningCollector(),
+                    context.getRuntimeStats());
+            if (!isRoleApplicable(metastore, identity, new PrestoPrincipal(USER, identity.getUser()), metastoreContext, role)) {
+                denySetRole(role);
+            }
+        });
     }
 
     @Override
@@ -702,8 +704,9 @@ public class SqlStandardAccessControl
 
     private boolean isAdmin(ConnectorTransactionHandle transaction, ConnectorIdentity identity, MetastoreContext metastoreContext)
     {
-        SemiTransactionalHiveMetastore metastore = getMetastore(transaction);
-        return isRoleEnabled(identity, (PrestoPrincipal p) -> metastore.listRoleGrants(metastoreContext, p), ADMIN_ROLE_NAME);
+        return getMetastore(transaction)
+                .map(metastore -> isRoleEnabled(identity, (PrestoPrincipal p) -> metastore.listRoleGrants(metastoreContext, p), ADMIN_ROLE_NAME))
+                .orElse(false);
     }
 
     private boolean isDatabaseOwner(ConnectorTransactionHandle transaction, ConnectorIdentity identity, MetastoreContext metastoreContext, String databaseName)
@@ -717,22 +720,24 @@ public class SqlStandardAccessControl
             return true;
         }
 
-        SemiTransactionalHiveMetastore metastore = getMetastore(transaction);
-        Optional<Database> databaseMetadata = metastore.getDatabase(metastoreContext, databaseName);
-        if (!databaseMetadata.isPresent()) {
+        Optional<SemiTransactionalHiveMetastore> metastoreOptional = getMetastore(transaction);
+        return metastoreOptional.map(metastore -> {
+            Optional<Database> databaseMetadata = metastore.getDatabase(metastoreContext, databaseName);
+            if (!databaseMetadata.isPresent()) {
+                return false;
+            }
+
+            Database database = databaseMetadata.get();
+
+            // a database can be owned by a user or role
+            if (database.getOwnerType() == USER && identity.getUser().equals(database.getOwnerName())) {
+                return true;
+            }
+            if (database.getOwnerType() == ROLE && isRoleEnabled(identity, (PrestoPrincipal p) -> metastore.listRoleGrants(metastoreContext, p), database.getOwnerName())) {
+                return true;
+            }
             return false;
-        }
-
-        Database database = databaseMetadata.get();
-
-        // a database can be owned by a user or role
-        if (database.getOwnerType() == USER && identity.getUser().equals(database.getOwnerName())) {
-            return true;
-        }
-        if (database.getOwnerType() == ROLE && isRoleEnabled(identity, (PrestoPrincipal p) -> metastore.listRoleGrants(metastoreContext, p), database.getOwnerName())) {
-            return true;
-        }
-        return false;
+        }).orElse(false);
     }
 
     private boolean isTableOwner(ConnectorTransactionHandle transaction, ConnectorIdentity identity, MetastoreContext metastoreContext, SchemaTableName tableName)
@@ -760,10 +765,11 @@ public class SqlStandardAccessControl
             return true;
         }
 
-        SemiTransactionalHiveMetastore metastore = getMetastore(transaction);
-        return listEnabledTablePrivileges(metastore, tableName.getSchemaName(), tableName.getTableName(), identity, metastoreContext)
-                .filter(privilegeInfo -> !grantOptionRequired || privilegeInfo.isGrantOption())
-                .anyMatch(privilegeInfo -> privilegeInfo.getHivePrivilege().equals(requiredPrivilege));
+        return getMetastore(transaction)
+                .map(metastore -> listEnabledTablePrivileges(metastore, tableName.getSchemaName(), tableName.getTableName(), identity, metastoreContext)
+                        .filter(privilegeInfo -> !grantOptionRequired || privilegeInfo.isGrantOption())
+                        .anyMatch(privilegeInfo -> privilegeInfo.getHivePrivilege().equals(requiredPrivilege)))
+                .orElse(false);
     }
 
     private boolean hasGrantOptionForPrivilege(ConnectorTransactionHandle transaction, ConnectorIdentity identity, MetastoreContext metastoreContext, Privilege privilege, SchemaTableName tableName)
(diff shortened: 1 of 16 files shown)


presto-hive/src/test/java/com/facebook/presto/hive/TestHiveIntegrationSmokeTest.java
```diff
@@ -6830,6 +6830,37 @@ public void testAlterColumnNotNull()
...
+        Session session = getQueryRunner().getDefaultSession();
+        assertUpdate(session, "CREATE TABLE test_non_autocommit_table (a bigint, b double, c varchar) with(partitioned_by = array['c'])");
+        try {
+            transaction(getQueryRunner().getTransactionManager(), getQueryRunner().getAccessControl())
+                    .execute(session, txnSession3 -> {
+                        assertUpdate(txnSession3, "insert into test_non_autocommit_table values(1, 1.1, '1001'), (2, 1.2, '1002')", 2);
+                        assertQueryFails(txnSession3, "insert into test_non_autocommit_table values(3, 1.3, '1002')",
+                                "Inserting into a partition that were added, altered, or inserted into in the same transaction is not supported");
+
+                        // can not execute any SQLs except `rollback` in current session
```
elharo, line 6845: nit: cannot

presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedSmokeTestBase.java
```diff
@@ -1660,6 +1661,39 @@ public void testMetadataDeleteOnNonIdentityPartitionColumn(String version, Strin
...
+        try {
+            transaction(getQueryRunner().getTransactionManager(), getQueryRunner().getAccessControl())
+                    .execute(Session.builder(session)
+                                    .setCatalogSessionProperty(ICEBERG_CATALOG, "merge_on_read_enabled", "false")
+                                    .build(),
+                            txnSession -> {
+                                // Query should fail because of the conflicts between session property and table property in table mode validation
+                                assertQueryFails(txnSession, "select * from test_non_autocommit_table", "merge-on-read table mode not supported yet");
+
+                                // can not execute any SQLs except `rollback` in current session
```
elharo, line 1678: cannot

presto-spark-base/src/main/java/com/facebook/presto/spark/PrestoSparkQueryExecutionFactory.java
```diff
@@ -149,6 +149,7 @@
 import static com.facebook.presto.spark.util.PrestoSparkUtils.getActionResultWithTimeout;
 import static com.facebook.presto.spi.StandardErrorCode.GENERIC_INTERNAL_ERROR;
 import static com.facebook.presto.spi.StandardErrorCode.INVALID_SESSION_PROPERTY;
+import static com.facebook.presto.sql.QueryUtil.isRollBack;
```
elharo, line 152: This couples to a new package & module. Probably not worth it for the trivial method you're importing.

hantangwangd: Thanks for the comment, I also struggled here for a while. Fixed as your suggestion.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedSmokeTestBase.java
```diff
@@ -1660,6 +1661,33 @@ public void testMetadataDeleteOnNonIdentityPartitionColumn(String version, Strin
...
     }
 
+    @Test
+    public void testNoneAutoCommitTransactionWithFailAndRollback()
+    {
+        Session session = getQueryRunner().getDefaultSession();
+        assertUpdate(session, "CREATE TABLE test_non_autocommit_table (a bigint, b double)");
+        transaction(getQueryRunner().getTransactionManager(), getQueryRunner().getAccessControl())
+                .execute(Session.builder(session)
+                                .setCatalogSessionProperty(ICEBERG_CATALOG, "merge_on_read_enabled", "false")
```
ZacBlanco, line 1671: I think you can use the `fail` function instead to fail a query?

```sql
SELECT fail('forced failure')
```

presto-hive/src/test/java/com/facebook/presto/hive/TestHiveIntegrationSmokeTest.java
```diff
@@ -6830,6 +6830,31 @@ public void testAlterColumnNotNull()
...
 
+    @Test
+    public void testNoneAutoCommitTransactionWithFailAndRollback()
+    {
+        Session session = getQueryRunner().getDefaultSession();
+        assertUpdate(session, "CREATE TABLE test_non_autocommit_table (a bigint, b double, c varchar) with(partitioned_by = array['c'])");
+        transaction(getQueryRunner().getTransactionManager(), getQueryRunner().getAccessControl())
+                .execute(session, txnSession3 -> {
+                    assertUpdate(txnSession3, "insert into test_non_autocommit_table values(1, 1.1, '1001'), (2, 1.2, '1002')", 2);
+                    assertQueryFails(txnSession3, "insert into test_non_autocommit_table values(3, 1.3, '1002')",
```
ZacBlanco, line 6841: Similar to other comment, if you need to force a query failure I would use the `fail` function

presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedSmokeTestBase.java
hantangwangd: Thank you, it's a convenient tool to simulate a query failure, fixed!

tdcmeehan (commented):
It seems like `ignoreTransactionState` is more like `isRollingBack`.  Do you think that's an accurate way of describing it?

hantangwangd:
> It seems like `ignoreTransactionState` is more like `isRollingBack`. Do you think that's an accurate way of describing it?

My thought is, we may want to support other statements in the future for this situation (such as `commit`,  although the commit result should fail, but it should be able to close the entire transaction too). So name this flag `ignoreTransactionState` rather than `isRollback`, do you think it makes sense?

elharo:
Can you resolve the conflicts?

hantangwangd:
> Can you resolve the conflicts?

Sure, the conflicts are resolved!

presto-hive/src/main/java/com/facebook/presto/hive/security/SqlStandardAccessControl.java
```diff
@@ -618,6 +618,9 @@ public void checkCanRevokeRoles(ConnectorTransactionHandle transactionHandle, Co
     public void checkCanSetRole(ConnectorTransactionHandle transaction, ConnectorIdentity identity, AccessControlContext context, String role, String catalogName)
     {
         SemiTransactionalHiveMetastore metastore = getMetastore(transaction);
```
tdcmeehan, line 620: Can you help me understand why we need to check for null as part of this bug fix?

hantangwangd: When a statement in a non-autocommit transaction fails, `InMemoryTransactionManager.abortInternal()` is triggered, which causes the relevant hive connector transaction manager to clean up it's corresponding connector transaction. However, the transaction in `InMemoryTransactionManager` is just marked as failed, but not removed, referring to https://github.com/prestodb/presto/blob/master/presto-main/src/main/java/com/facebook/presto/execution/QueryStateMachine.java#L918.

Therefore, if we do not check the null value here, an NPE exception will be encountered when executing the rollback statement.

justrelax19:
@tdcmeehan and @hantangwangd any update on the above PR since its been open for about 5 months now and if we can merge and close it ?
Because we are also receiving the `aborted transaction block` issue and seems like the above PR will fix it.

ZacBlanco (commented):
I mostly agree with Tim's thought from his comment [here](https://github.com/prestodb/presto/pull/23247#pullrequestreview-2257941028)

But I also understand your reasoning in the response. Can we add a well-written javadoc which explains the meaning and use of this field so that it's clearer to developers the intention of this name? `isRollback` seems clear to me, but `isIgnoreTransactionState` does not

presto-hive/src/main/java/com/facebook/presto/hive/security/SqlStandardAccessControl.java
```diff
@@ -800,6 +803,6 @@ private boolean hasAdminOptionForRoles(ConnectorTransactionHandle transaction, C
     private SemiTransactionalHiveMetastore getMetastore(ConnectorTransactionHandle transaction)
     {
         TransactionalMetadata metadata = hiveTransactionManager.get(transaction);
-        return metadata.getMetastore();
+        return metadata == null ? null : metadata.getMetastore();
```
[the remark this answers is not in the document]
hantangwangd: Good suggestion, I have changed the method to return an `Optional<SemiTransactionalHiveMetastore>`, and check everywhere we call `getMetastore(transaction)`. Please take a look when convenient!

hantangwangd:
> I mostly agree with Tim's thought from his comment [here](https://github.com/prestodb/presto/pull/23247#pullrequestreview-2257941028)
> 
> But I also understand your reasoning in the response. Can we add a well-written javadoc which explains the meaning and use of this field so that it's clearer to developers the intention of this name? `isRollback` seems clear to me, but `isIgnoreTransactionState` does not

Thanks for the feedback, I agree with you that `isIgnoreTransactionState` might bring some confusions to future developers. To make its intention more clearer, and to make it applicable to some other statements (for example, maybe `commit` in the future) as well, do you think it makes sense to call it `enableRollback`? I'm open on this, and do not have a strong inclination, please let me know your thought.

ZacBlanco (changes requested):

presto-main-base/src/main/java/com/facebook/presto/dispatcher/DispatchManager.java
```diff
@@ -462,6 +462,11 @@ public void cancelQuery(QueryId queryId)
                 .ifPresent(DispatchQuery::cancel);
     }
 
+    private boolean isRollBack(String sql)
+    {
+        return sql != null && sql.trim().equalsIgnoreCase("rollback");
+    }
+
```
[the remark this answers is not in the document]
hantangwangd: @ZacBlanco Thanks for the suggestion. I've add the method `isRollback` in `PreparedQuery`, and use it in `DispatchManager`, it seems more elegant than before. Please take a look when convenient.

presto-hive/src/main/java/com/facebook/presto/hive/security/SqlStandardAccessControl.java
```diff
@@ -788,18 +795,22 @@ private boolean hasAdminOptionForRoles(ConnectorTransactionHandle transaction, C
...
     }
 
-    private SemiTransactionalHiveMetastore getMetastore(ConnectorTransactionHandle transaction)
+    private Optional<SemiTransactionalHiveMetastore> getMetastore(ConnectorTransactionHandle transaction)
     {
         TransactionalMetadata metadata = hiveTransactionManager.get(transaction);
-        return metadata.getMetastore();
+        // In some scenarios, for example, when a statement in a non-autocommit transaction fails,
+        // the corresponding transaction metadata could be null.
+        return Optional.ofNullable(metadata == null ? null : metadata.getMetastore());
```
ZacBlanco, line 814: The idiomatic way to write this would be 
```suggestion
        return Optional.ofNullable(metadata).map(HiveMetadata::getMetastore);
```

presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedSmokeTestBase.java
```diff
@@ -1880,6 +1884,38 @@ public void testMetadataDeleteOnNonIdentityPartitionColumn(String version, Strin
         metadataDeleteOnTruncateTransform(version, mode, errorMessage);
     }
 
+    @Test
+    public void testNoneAutoCommitTransactionWithFailAndRollback()
```
ZacBlanco, line 1888: Do you think it's possible to put this inside one of the base test classes instead? I think that it's a case which should be tested outside of the Iceberg and Hive connectors too

ZacBlanco (changes requested):
Two minor things, otherwise LGTM

presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedSmokeTestBase.java
hantangwangd: Sure, good point, I have moved the test to `AbstractTestDistributedQueries`, so that it will be tested by all the subclasses except `Cassandra`. Please take a look when available, thanks a lot!

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestDistributedQueries.java
```diff
@@ -243,6 +246,45 @@ public void testCreateTable()
         assertFalse(getQueryRunner().tableExists(getSession(), "test_create_like"));
     }
 
+    @Test
+    public void testNoneAutoCommitTransactionWithFailAndRollback()
```
ZacBlanco, line 250: nit: typo?
```suggestion
    public void testNonAutoCommitTransactionWithFailAndRollback()
```

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestDistributedQueries.java
```diff
@@ -243,6 +246,45 @@ public void testCreateTable()
...
+                            // execute `rollback` successfully
+                            assertUpdate(txnSession, "rollback");
+                        });
+
+        assertQuery("select count(*) from test_non_autocommit_table", "values(0)");
+        try {
+            assertUpdate("drop table if exists test_non_autocommit_table");
+        }
+        catch (Exception e) {
+            // ignored for connector compatibility
```
ZacBlanco, line 284: What does "for connector compatibility imply" here? Do some connectors behave differently?

ZacBlanco (approved):
Thanks for this fix! LGTM

PR #23255, opened 2024-07-18
[Native] Add alias substring for substr function

aditi-pandit:
## Description
Presto allows usage of substring as an alias of substr.

e2e tests

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/types/PrestoToVeloxExpr.cpp
+++ b/presto-native-execution/presto_cpp/main/types/PrestoToVeloxExpr.cpp
@@ -51,6 +51,9 @@ std::string mapScalarFunction(const std::string& name) {
       {"presto.default.$operator$not_equal", "presto.default.neq"},
       {"presto.default.$operator$subtract", "presto.default.minus"},
       {"presto.default.$operator$subscript", "presto.default.subscript"},
+      {"presto.default.$operator$subscript", "presto.default.subscript"},
+      // Substring.
+      {"substring", "presto.default.substr"},
       // Special form function overrides.
       {"presto.default.in", "in"},
   };
--- a/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeGeneralQueries.java
+++ b/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeGeneralQueries.java
@@ -822,9 +822,12 @@ public abstract class AbstractTestNativeGeneralQueries
     public void testStringFunctions()
     {
         // Substr, length, trim.
-        assertQuery("SELECT substr(comment, 1, 10), length(comment), trim(comment) FROM orders");
+        assertQuery("SELECT substr(comment, 5), length(comment), trim(comment) FROM orders");
         assertQuery("SELECT substr(comment, 1, 10), length(comment), ltrim(comment) FROM orders");
         assertQuery("SELECT substr(comment, 1, 10), length(comment), rtrim(comment) FROM orders");
+
+        assertQuery("SELECT substring(comment, 5), length(comment), trim(comment) FROM orders");
+        assertQuery("SELECT substring(comment, 1, 10), length(comment), ltrim(comment) FROM orders");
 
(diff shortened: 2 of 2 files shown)


presto-native-execution/presto_cpp/main/types/PrestoToVeloxExpr.cpp
```diff
@@ -51,6 +51,9 @@ std::string mapScalarFunction(const std::string& name) {
       {"presto.default.$operator$not_equal", "presto.default.neq"},
       {"presto.default.$operator$subtract", "presto.default.minus"},
       {"presto.default.$operator$subscript", "presto.default.subscript"},
+      {"presto.default.$operator$subscript", "presto.default.subscript"},
+      // Substring.
+      {"substring", "presto.default.substr"},
```
mbasmanova, line 56: This change should go into velox/functions/prestosql/registration/StringFunctionsRegistration.cpp

aditi-pandit: @mbasmanova : There is this extra complexity for the alias. Its registered in TeradataFunctionsPlugin. Its not a built-in.

We will need to take a call on if and how we expose this in Prestissimo. Its automatically available when we use the Presto java builds from the site. I'll check the Prestissimo docker images as well.

The Teradata string functions are mostly aliases (there is one exception). But there are couple TeradataDateFunctions to implement.

So there would be few more steps here:
i) Implement all Teradata functions in Velox. Expose them in a namespace.
ii) Load TeradataFunctionsPlugin in Prestissimo co-ordinator.

We encountered this because someone was trying a DB2 SQL on Prestissimo. But they are open to using substr.  So we are flexible about this issue for now. 

I'm interested in hearing your thoughts about this though.

@tdcmeehan

mbasmanova: @aditi-pandit Aditi, thank you for clarifying. I didn't realize this function / alias is not a built-in, but rather a Terradata specific extension. If that's the case, we need to figure out how to properly define these and not mix them with build-in functions.

aditi-pandit: @mbasmanova : Thanks for clarifying. I had interpret your comment to mean that you are comfortable having an alias substring for substr in Velox.

My understanding is that in the future with the native sidecar, Velox will report all functions (with their namespaces) to the co-ordinator. The co-ordinator will only use those metadata for planning. So there will not be a notion of function plugin at the co-ordinator anymore. There will be only Velox functions.

Given this thinking, our option is to implement all Teradata functions in Velox in their own namespace. Or then deprecate the Teradata substring alias with the built in alias as in the linked PR.

wdyt ?

BTW, there isn't any urgency on our part for this. We can make a call after native sidecar is enabled.

tdcmeehan: To me, the value of adding `substring` as an alias for the similar-sounding `substr` is minimal.  And I would say it's a non-goal to make functions consistent between DB2 and Presto.  Meanwhile, there is ever-growing cognitive overhead for increasing Presto's function footprint--minimal at each introduction, but ever-increasing.  So in my opinion, this would belong to something like an internal fork.  Or, in the future, we can load it as a shared library when we support that capability.  But for now, I'm not in favor of adding it as a built-in alias, just because it doesn't seem like the future cost is worth the small value we get out of it.

aditi-pandit: @tdcmeehan, @mbasmanova : As discussed in the Architecture review meeting yesterday, its actually substring alias in the 2016 SQL standard. However, the syntax is different :
```
<character substring function> ::=
  SUBSTRING <left paren> <character value expression> FROM <start position>
      [ FOR <string length> ] [ USING <char length units> ] <right paren>
<regular expression substring function> ::=
SUBSTRING <left paren> <character value expression> SIMILAR <character value expression>
      ESCAPE <escape character> <right paren>
```

If we want to make substring a built-in we should conform to this syntax.

Lets discuss that in another issue.

aditi-pandit:
Closing as per https://github.com/prestodb/presto/pull/23255#discussion_r1692272893

PR #23257, opened 2024-07-19
Aggressively prune query info after completion

### LINKED ISSUE #23254: Histograms can consume significant amounts of memory in query history
## Expected Behavior

We should try to keep the coordinator memory footprint as lean as possible to prevent delays in query optimization and dispatching. The current behavior for keeping queries around is configurable, but default values can lead to high memory
pressure when using real histograms, even on smaller benchmarks like TPC-DS.

## Current Behavior

After execution of a query finishes, the query is added to an expiration queue: https://github.com/prestodb/presto/blob/05217403bed3c01ea59eb36d683b62598f37ef1e/presto-main/src/main/java/com/facebook/presto/execution/QueryTracker.java#L197-L201

While in the expiration queue, we hold onto the full query information until the queue reaches its max size, configured as `query.max-history`. https://github.com/prestodb/presto/blob/05217403bed3c01ea59eb36d683b62598f37ef1e/presto-main/src/main/java/com/facebook/presto/execution/QueryTracker.java#L307-L322

Only after the queue size is reached can we start pruning info. This is problematic if memory usage balloons just from the queries in the expiration queue.

## Possible Solution

Provide different levels of pruning as arguments to `TrackedQuery#pruneInfo`: https://github.com/prestodb/presto/blob/05217403bed3c01ea59eb36d683b62598f37ef1e/presto-main/src/main/java/com/facebook/presto/execution/QueryTracker.java#L412).

Then, we can call `query.pruneInfo()` at different times with different levels to remove different bits of information.

## Steps to Reproduce

1. Collect large histograms with #22365 
2. Run full TPC-DS query set on Presto
3. Watch memory utilization increase. Force GCs along the way. Heap will not decrease.
4. Take heap dump and look for retained memory under ColumnStatistics. It will be very large.

ZacBlanco:
## Description

This change aggressively prunes query plan metadata before query expiry. It is mostly limited to query plan statistics.

With this change and with histograms enabled in #22365, this cuts the coordinator heap utilization by anywhere between 50% and 75% after running a portion of the TPC-DS benchmark (3-4GiB -> 600-800MiB)

## Motivation and Context

Previously, we only pruned information from old queries after the expiration queue filled. With the addition of histograms query statistics can take two orders of magnitude more memory than previous statistics. Without pruning statistics we will much more quickly hit memory limits of the JVM.

Specifically, these are the areas where we try to clear references to save memory:

- Session: `planNodeStatsMap` and `planCostMapMap`
- QueryStateMachine: clear statistics from `inputs` and `planStatsAndCosts`
- StateMachine: Once final queryInfo reaches a terminal state, we should clear the event listeners to release references to local variables which maintain the SqlQuerySchedule and Stage execution information
- Prune all histograms from `finalQueryInfo`'s `planStatsAndCosts` and from the entire `StageInfo` graph.

Fixes #23254 

## Impact

- Results from the QueryInfo API after execution finishes will always return column statistics with empty histograms

## Test Plan

- Added a new test which verifies that a query completed event is generated before all queryInfo pruning is done

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/dispatcher/FailedDispatchQuery.java
+++ b/presto-main/src/main/java/com/facebook/presto/dispatcher/FailedDispatchQuery.java
@@ -115,7 +115,10 @@ public class FailedDispatchQuery
     public void cancel() {}
 
     @Override
-    public void pruneInfo() {}
+    public void pruneExpiredQueryInfo() {}
+
+    @Override
+    public void pruneFinishedQueryInfo() {}
 
     @Override
     public QueryId getQueryId()
@@ -194,5 +197,5 @@ public class FailedDispatchQuery
 
     @Override
     public void setResourceGroupQueryLimits(ResourceGroupQueryLimits resourceGroupQueryLimits)
-    { }
+    {}
 }
--- a/presto-main/src/main/java/com/facebook/presto/execution/QueryStateMachine.java
+++ b/presto-main/src/main/java/com/facebook/presto/execution/QueryStateMachine.java
@@ -19,7 +19,9 @@ import com.facebook.presto.common.ErrorCode;
 import com.facebook.presto.common.resourceGroups.QueryType;
 import com.facebook.presto.common.transaction.TransactionId;
 import com.facebook.presto.common.type.Type;
+import com.facebook.presto.cost.PlanNodeStatsEstimate;
 import com.facebook.presto.cost.StatsAndCosts;
+import com.facebook.presto.cost.VariableStatsEstimate;
 import com.facebook.presto.execution.QueryExecution.QueryOutputInfo;
 import com.facebook.presto.execution.StateMachine.StateChangeListener;
 import com.facebook.presto.memory.VersionedMemoryPoolId;
@@ -38,13 +40,17 @@ import com.facebook.presto.spi.plan.PlanNodeId;
 import com.facebook.presto.spi.resourceGroups.ResourceGroupId;
 import com.facebook.presto.spi.security.AccessControl;
 import com.facebook.presto.spi.security.SelectedRole;
+import com.facebook.presto.spi.statistics.ColumnStatistics;
+import com.facebook.presto.spi.statistics.TableStatistics;
 import com.facebook.presto.sql.planner.CanonicalPlanWithInfo;
+import com.facebook.presto.sql.planner.PlanFragment;
 import com.facebook.presto.transaction.TransactionInfo;
 import com.facebook.presto.transaction.TransactionManager;
 import com.google.common.base.Ticker;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import com.google.common.collect.Streams;
 import com.google.common.util.concurrent.FutureCallback;
@@ -93,6 +99,8 @@ import static com.facebook.presto.util.Failures.toFailure;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkState;
 import static com.google.common.collect.ImmutableList.toImmutableList;
+import static com.google.common.collect.ImmutableMap.toImmutableMap;
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
 import static com.google.common.util.concurrent.MoreExecutors.directExecutor;
 import static io.airlift.units.DataSize.succinctBytes;
 import static java.lang.String.format;
@@ -1046,7 +1054,24 @@ public class QueryStateMachine
      * Remove large objects from the query info object graph, e.g : plan, stats, stage summaries, failed attempts
      * Used when pruning expired queries from the state machine
      */
-    public void pruneQueryInfo()
+    public void pruneQueryInfoExpired()
+    {
+        Optional<QueryInfo> finalInfo = finalQueryInfo.get();
+        if (!finalInfo.isPresent() || !finalInfo.get().getOutputStage().isPresent()) {
+            return;
+        }
+        QueryInfo queryInfo = finalInfo.get();
+        QueryInfo prunedQueryInfo;
+
+        prunedQueryInfo = pruneExpiredQueryInfo(queryInfo, getMemoryPool());
+        finalQueryInfo.compareAndSet(finalInfo, Optional.of(prunedQueryInfo));
+    }
+
+    /**
+     * Remove the largest objects from the query info object graph, e.g : extraneous stats, costs,
+     * and histograms to reduce memory utilization
+     */
+    public void pruneQueryInfoFinished()
     {
         Optional<QueryInfo> finalInfo = finalQueryInfo.get();
         if (!finalInfo.isPresent() || !finalInfo.get().getOutputStage().isPresent()) {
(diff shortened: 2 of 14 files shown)


aaneja (commented):
Can you check (and maybe add/link me to a test that asserts) that the [queryCompletedEvent](https://github.com/prestodb/presto/blob/f661a1ce171a58697030f09c870c013a252ff55a/presto-main/src/main/java/com/facebook/presto/event/QueryMonitor.java#L251) is fired _before_ the query info pruning is done ?

presto-tests/src/test/java/com/facebook/presto/tests/TestQueryManager.java
```diff
@@ -278,4 +304,174 @@ public void testQueryCountMetrics()
             Thread.sleep(1000);
         }
     }
+
+    @Test
+    public void testQueryCompletedInfoNotPruned()
```
ZacBlanco, line 309: Added the new test here @aaneja

aaneja (dismissed):

presto-main/src/main/java/com/facebook/presto/execution/QueryStateMachine.java
```diff
@@ -1046,28 +1055,167 @@ public QueryInfo updateQueryInfo(Optional<StageInfo> stageInfo)
...
+                planStatsAndCosts.getAndUpdate(stats -> Optional.ofNullable(stats)
+                        .map(QueryStateMachine::pruneHistogramsFromStatsAndCosts)
+                        .orElse(null));
+                prunedQueryInfo = pruneFinishedQueryInfo(queryInfo, inputs.get());
+                break;
+            case EXPIRED:
+                prunedQueryInfo = pruneExpiredQueryInfo(queryInfo, getMemoryPool());
+                break;
+            default:
+                prunedQueryInfo = queryInfo;
```
hantangwangd, line 1088: nit: Is it better to throw an exception here? As it shouldn't happen at all.

hantangwangd (dismissed):
LGTM, just one nit.

presto-tests/src/test/java/com/facebook/presto/tests/TestQueryManager.java
```diff
@@ -278,4 +302,199 @@ public void testQueryCountMetrics()
...
+                        33,
+                        DataSize.valueOf("34GB"),
+                        DataSize.valueOf("35GB"),
+                        DataSize.valueOf("36GB"),
+                        ImmutableList.of(),
+                        ImmutableList.of(),
+                        new RuntimeStats()),
+                Optional.empty(),
+                Optional.empty(),
+                ImmutableMap.of(),
```
elharo, line 466: Tangential and non-blocking, but I'd like to get in the habit of using Collections.emptyMap instead of ImmutableList.of to reduce our dependence on and vulnerability to third party libraries. ImmutableMap.of hasn't been needed for years now.

presto-tests/src/test/java/com/facebook/presto/tests/TestQueryManager.java
```diff
@@ -278,4 +302,199 @@ public void testQueryCountMetrics()
...
+                                "warning message")),
+                ImmutableSet.of(),
+                Optional.empty(),
+                false,
+                Optional.empty(),
+                Optional.empty(),
+                Optional.empty(),
+                Optional.empty(),
+                ImmutableMap.of(),
+                ImmutableSet.of(),
```
elharo, line 489: Collections.emptySet

presto-tests/src/test/java/com/facebook/presto/tests/TestQueryManager.java
```diff
@@ -278,4 +302,199 @@ public void testQueryCountMetrics()
...
+                StatsAndCosts.empty(),
+                ImmutableList.of(),
+                ImmutableList.of(),
```
elharo, line 492: Collections.emptyList

presto-tests/src/test/java/com/facebook/presto/tests/TestQueryManager.java
```diff
@@ -278,4 +302,199 @@ public void testQueryCountMetrics()
...
+                ImmutableMap.of(),
+                ImmutableMap.of(),
+                ImmutableSet.of(),
+                Optional.empty(),
+                false,
+                "33",
+                Optional.empty(),
+                null,
+                EXCEEDED_GLOBAL_MEMORY_LIMIT.toErrorCode(),
+                ImmutableList.of(
```
elharo, line 477: here you can use Arrays.asList
It's just a test so immutability doesn't help

presto-main/src/main/java/com/facebook/presto/execution/QueryStateMachine.java
ZacBlanco: Since this isn't the only place that `PruneLevel` is used, the thinking was that if someone were to add another enum entry in the future, it may not necessarily be desirable for this method to throw an exception because the implementation isn't entirely done in this method. 

For example, `PruneLevel` is also used here: https://github.com/prestodb/presto/blob/735877a34c1541af1a5900536633bdf60d124ace/presto-main/src/main/java/com/facebook/presto/execution/SqlQueryExecution.java#L750-L762

presto-tests/src/test/java/com/facebook/presto/tests/TestQueryManager.java
ZacBlanco: If this is the case, then we should update the contributing guide to state this: https://github.com/prestodb/presto/blob/9a3c695b1403a14e9e4b80f60920f1675115aec2/CONTRIBUTING.md?plain=1#L199

presto-main/src/main/java/com/facebook/presto/execution/QueryStateMachine.java
hantangwangd: IMO, throw an exception by default can make the responsibilities more clearer, even considering the scenario you mentioned above.

For example, if a coder add another enum entry in `PruneLevel` in future, but explicitly do not want to support it in `QueryStateMachine.pruneQueryInfo(level)` for some reasonable reasons, it's this coder's responsibility to make sure not invoking the function `stateMachine.pruneQueryInfo(level)` with the newly added unsupported level, or he/she will get a explicitly exception.

Meanwhile, if a coder add another enum entry in `PruneLevel` in future, but forget to support it in `QueryStateMachine.pruneQueryInfo(level)`, he/she will receive a reminder immediately when invoking `stateMachine.pruneQueryInfo(level)` with the newly added level. It's better than being ignored silently, which may result in not being able to detect this issue for a long while.

What's your opinion?

presto-main/src/main/java/com/facebook/presto/execution/QueryStateMachine.java
```diff
@@ -1046,28 +1055,167 @@ public QueryInfo updateQueryInfo(Optional<StageInfo> stageInfo)
...
     {
         Optional<QueryInfo> finalInfo = finalQueryInfo.get();
         if (!finalInfo.isPresent() || !finalInfo.get().getOutputStage().isPresent()) {
             return;
         }
 
         QueryInfo queryInfo = finalInfo.get();
+        QueryInfo prunedQueryInfo;
+        switch (level) {
+            case FINISHED:
```
jaystarshot, line 1068: PruneLevel has values `FINISHED` and `EXPIRED`. They don't sound like levels

ZacBlanco: I updated the enum name to `PruneEvent`

presto-main/src/main/java/com/facebook/presto/execution/QueryStateMachine.java
ZacBlanco: I can understand your reasoning. It has been updated to throw an exception instead

hantangwangd (approved):
Thanks for the fix.

presto-main/src/main/java/com/facebook/presto/execution/QueryTracker.java
```diff
@@ -409,6 +414,20 @@ public interface TrackedQuery
         void fail(Throwable cause);
 
         // XXX: This should be removed when the client protocol is improved, so that we don't need to hold onto so much query history
-        void pruneInfo();
+        void pruneInfo(PruneEvent event);
+    }
+
+    public enum PruneEvent
```
tdcmeehan, line 420: I would just add two methods, rather than introduce a prune event concept.  Perhaps `pruneExpiredQueryInfo` and `pruneFinishedQueryInfo`?

ZacBlanco: updated

presto-main/src/main/java/com/facebook/presto/dispatcher/FailedDispatchQuery.java
```diff
@@ -115,7 +115,10 @@ public void fail(Throwable throwable) {}
     public void cancel() {}
 
     @Override
-    public void pruneInfo() {}
+    public void pruneQueryExpired() {}
+
+    @Override
+    public void pruneQueryFinished() {}
```
tdcmeehan, line 121: How about this naming? 

```suggestion
    public void pruneExpiredQueryInfo() {}

    @Override
    public void pruneFinishedQueryInfo() {}
```

ZacBlanco: updated

tdcmeehan (approved):

Issue #22996, opened 2024-06-12
Add a UDF for splitting array into slices of given length


kaikalur:
SLICES(array<t>, sz)->array<array<t>>

slices(array[1,2,3,4,5,6], 3)->array[array[1,2,3], array[4,5,6]]

PR #23264, opened 2024-07-19
Add a function for splitting array into slices of given length

abhinavmuk04:
## Description
Add a function for splitting array into slices of given length

## Motivation and Context
(https://github.com/prestodb/presto/issues/22996)

## Impact
UDF for more features within Presto

## Test Plan
Various unit tests, similar to other existing UDF tests

## Contributor checklist

- [x] CI passed.

## Release Note

```
== RELEASE NOTES ==
* Add a new built-in function for array splitting. Given an array and an integer N, split it into pieces of size N  :pr:`23264`
```

### DIFF
--- a/presto-docs/src/main/sphinx/functions/array.rst
+++ b/presto-docs/src/main/sphinx/functions/array.rst
@@ -207,6 +207,16 @@ Array Functions
         SELECT array_sort_desc(ARRAY [null, 100, null, 1, 10, 50]); -- [100, 50, 10, 1, null, null]
         SELECT array_sort_desc(ARRAY [ARRAY ["a", null], null, ARRAY ["a"]); -- [["a", null], ["a"], null]
 
+.. function:: array_split_into_chunks(array(T), int) -> array(array(T))
+
+    Returns an ``array`` of arrays splitting the input ``array`` into chunks of given length.
+    If the ``array`` is not evenly divisible it will split into as many possible chunks and return
+    the left over elements for the last ``array``. Ignores null inputs, but not elements.
+
+        SELECT array_split_into_chunks(ARRAY [1, 2, 3, 4], 3); -- [[1, 2, 3], [4]]
+        SELECT array_split_into_chunks(null, null); -- null
+        SELECT array_split_into_chunks(array[1, 2, 3, cast(null as int)], 2]); -- [[1, 2], [3, null]]
+
 .. function:: array_sum(array(T)) -> bigint/double
 
     Returns the sum of all non-null elements of the ``array``. If there is no non-null elements, returns ``0``.
--- a/presto-main/src/main/java/com/facebook/presto/operator/scalar/sql/ArraySqlFunctions.java
+++ b/presto-main/src/main/java/com/facebook/presto/operator/scalar/sql/ArraySqlFunctions.java
@@ -37,6 +37,24 @@ public class ArraySqlFunctions
                 "s -> if(s[2] = 0, cast(null as double), s[1] / cast(s[2] as double)))";
     }
 
+    @SqlInvokedScalarFunction(value = "array_split_into_chunks", deterministic = true, calledOnNullInput = false)
+    @Description("Returns an array of arrays splitting input array into chunks of given length. " +
+            "If array is not evenly divisible it will split into as many possible chunks and " +
+            "return the left over elements for the last array. Returns null for null inputs, but not elements.")
+    @TypeParameter("T")
+    @SqlParameters({@SqlParameter(name = "input", type = "array(T)"), @SqlParameter(name = "sz", type = "int")})
+    @SqlType("array(array(T))")
+    public static String arraySplitIntoChunks()
+    {
+        return "RETURN IF(sz <= 0, " +
+                "fail('Invalid slice size: ' || cast(sz as varchar) || '. Size must be greater than zero.'), " +
+                "IF(cardinality(input) / sz > 10000, " +
+                "fail('Cannot split array of size: ' || cast(cardinality(input) as varchar) || ' into more than 10000 parts.'), " +
+                "transform(" +
+                "sequence(1, cardinality(input), sz), " +
+                "x -> slice(input, x, sz))))";
+    }
+
     @SqlInvokedScalarFunction(value = "array_frequency", deterministic = true, calledOnNullInput = false)
     @Description("Returns the frequency of all array elements as a map.")
     @TypeParameter("T")
--- a/presto-main/src/test/java/com/facebook/presto/operator/scalar/sql/TestArraySqlFunctions.java
+++ b/presto-main/src/test/java/com/facebook/presto/operator/scalar/sql/TestArraySqlFunctions.java
@@ -22,6 +22,9 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import org.testng.annotations.Test;
 
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
 import static com.facebook.presto.block.BlockAssertions.createMapType;
 import static com.facebook.presto.common.type.BigintType.BIGINT;
 import static com.facebook.presto.common.type.BooleanType.BOOLEAN;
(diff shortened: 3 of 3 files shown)


presto-main/src/main/java/com/facebook/presto/operator/scalar/sql/ArraySqlFunctions.java
```diff
@@ -37,6 +37,18 @@ public static String arrayAverage()
...
+    @SqlInvokedScalarFunction(value = "array_split", deterministic = true, calledOnNullInput = false)
+    @Description("Returns an array of arrays splitting input array into slices of given length. Ignores null elements.")
+    @TypeParameter("T")
+    @SqlParameters({@SqlParameter(name = "input", type = "array(T)"), @SqlParameter(name = "sz", type = "int")})
+    @SqlType("array(array(T))")
+    public static String arraySplit()
+    {
+        return "RETURN transform(" +
+                "sequence(1, cardinality(input), sz), " +
+                "x -> slice(input, x, least(sz, cardinality(input) - x + sz)))";
```
kaikalur, line 49: you don't need the least - slice already does that!

kaikalur (approved):

elharo (commented):
Is there a reason this is a UDF and not just a function?

presto-main/src/main/java/com/facebook/presto/operator/scalar/sql/ArraySqlFunctions.java
```diff
@@ -37,6 +37,18 @@ public static String arrayAverage()
...
+    @SqlInvokedScalarFunction(value = "array_split", deterministic = true, calledOnNullInput = false)
+    @Description("Returns an array of arrays splitting input array into slices of given length. Ignores null elements.")
+    @TypeParameter("T")
+    @SqlParameters({@SqlParameter(name = "input", type = "array(T)"), @SqlParameter(name = "sz", type = "int")})
+    @SqlType("array(array(T))")
+    public static String arraySplit()
+    {
+        return "RETURN transform(" +
+                "sequence(1, cardinality(input), sz), " +
+                "x -> slice(input, x, sz))";
```
kaikalur, line 49: In Presto we have a limit of 10000 elements for SEQUENCE function. So will be good to check and fail:

RETURN IF(cardinality(input)/sz > 10000, fail('Cannot split array of size: ' || cast(cardinality(input) as varchar))', transfrorm...)

abhinavmuk04:
> Is there a reason this is a UDF and not just a function?

@elharo With things such as preventing division by zero and checking for larger element it allows for a bit more customization and can be beneficial as a UDF

kaikalur (dismissed):

elharo (changes requested):
As far as I can tell this is **NOT** a UDF. It is a new built-in function. Is there something I'm missing?

presto-main/src/main/java/com/facebook/presto/operator/scalar/sql/ArraySqlFunctions.java
```diff
@@ -37,6 +37,22 @@ public static String arrayAverage()
                 "s -> if(s[2] = 0, cast(null as double), s[1] / cast(s[2] as double)))";
     }
 
+    @SqlInvokedScalarFunction(value = "array_split", deterministic = true, calledOnNullInput = false)
+    @Description("Returns an array of arrays splitting input array into slices of given length. Ignores null elements.")
```
elharo, line 41: what happens if the array size is not evenly divisible by the slice size? This should be documented.

kaikalur:
we use the term loosely also SQL UDFs is a term we use for all these functions that we add now. No need to split hairs

elharo (commented):
presto-docs/src/main/sphinx/functions/array.rst needs to be updated in this PR to reflect the new function

presto-docs/src/main/sphinx/functions/array.rst
```diff
@@ -207,6 +207,14 @@ Array Functions
         SELECT array_sort_desc(ARRAY [null, 100, null, 1, 10, 50]); -- [100, 50, 10, 1, null, null]
         SELECT array_sort_desc(ARRAY [ARRAY ["a", null], null, ARRAY ["a"]); -- [["a", null], ["a"], null]
 
+.. function:: array_split(array(T), int) -> array(array(T))
+
+    Returns an ``array`` of arrays splitting the input ``array`` into slices of given length.
```
elharo, line 212: I'm a little concerned that it's more natural to split an array into N pieces than pieces of size N. I made that mistake myself when I first read this PR. The function name doesn't make it clear which is being done here, and both could be named array_split and have the same signature. Are we confident users will understand this?

kaikalur: Maybe call it array_split_into_chunks where typically it's the chunk size

presto-docs/src/main/sphinx/functions/array.rst
```diff
@@ -207,6 +207,14 @@ Array Functions
         SELECT array_sort_desc(ARRAY [null, 100, null, 1, 10, 50]); -- [100, 50, 10, 1, null, null]
         SELECT array_sort_desc(ARRAY [ARRAY ["a", null], null, ARRAY ["a"]); -- [["a", null], ["a"], null]
 
+.. function:: array_split_into_chunks(array(T), int) -> array(array(T))
+
+    Returns an ``array`` of arrays splitting the input ``array`` into chunks of given length.
+    If the ``array`` is not evenly divisible it will split into as many possible chunks and return
+    the left over elements for the last ``array``. Ignores null elements.
+
+        SELECT array_split(ARRAY [1, 2, 3, 4], 3); -- [[1, 2, 3], [4]]
```
feilong-liu, line 216: array_split -> array_split_into_chunks

presto-main/src/main/java/com/facebook/presto/operator/scalar/sql/ArraySqlFunctions.java
```diff
@@ -37,6 +37,24 @@ public static String arrayAverage()
                 "s -> if(s[2] = 0, cast(null as double), s[1] / cast(s[2] as double)))";
     }
 
+    @SqlInvokedScalarFunction(value = "array_split_into_chunks", deterministic = true, calledOnNullInput = false)
+    @Description("Returns an array of arrays splitting input array into chunks of given length. " +
+            "If array is not evenly divisible it will split into as many possible chunks and " +
+            "return the left over elements for the last array. Ignores null elements.")
```
feilong-liu, line 43: Why ignoring NULLs?

presto-main/src/test/java/com/facebook/presto/operator/scalar/sql/TestArraySqlFunctions.java
```diff
@@ -52,6 +55,57 @@ public void testArrayAverage()
         assertFunction("array_average(null)", DOUBLE, null);
     }
 
+    @Test
```
feilong-liu, line 58: Also add test with NULLs?

presto-main/src/test/java/com/facebook/presto/operator/scalar/sql/TestArraySqlFunctions.java
```diff
@@ -52,6 +55,57 @@ public void testArrayAverage()
...
+        assertFunction("array_split_into_chunks(array[bigint '2', bigint '3', bigint '4' , bigint '5'], 4)", new ArrayType(new ArrayType(BIGINT)), ImmutableList.of(ImmutableList.of(2L, 3L, 4L, 5L)));
+        assertFunction("array_split_into_chunks(array[bigint '-66', bigint '3', bigint '-66' , bigint '5'], 1)", new ArrayType(new ArrayType(BIGINT)), ImmutableList.of(ImmutableList.of(-66L), ImmutableList.of(3L), ImmutableList.of(-66L), ImmutableList.of(5L)));
+        assertFunction("array_split_into_chunks(array[bigint '-1', bigint '2', bigint '3' , bigint '-11'], 6)", new ArrayType(new ArrayType(BIGINT)), ImmutableList.of(ImmutableList.of(-1L, 2L, 3L, -11L)));
+        assertFunction("array_split_into_chunks(array[bigint '1', bigint '2', bigint '3', bigint '4', bigint '5', bigint '6', bigint '7'], 3)", new ArrayType(new ArrayType(BIGINT)), ImmutableList.of(ImmutableList.of(1L, 2L, 3L), ImmutableList.of(4L, 5L, 6L), ImmutableList.of(7L)));
+        assertInvalidFunction("array_split_into_chunks(array[bigint '-1', bigint '2', bigint '3' , bigint '-11'], 0)", StandardErrorCode.GENERIC_USER_ERROR, "Invalid slice size: 0. Size must be greater than zero.");
```
feilong-liu, line 67: Also test this 10000 limits?

presto-docs/src/main/sphinx/functions/array.rst
```diff
@@ -207,6 +207,14 @@ Array Functions
         SELECT array_sort_desc(ARRAY [null, 100, null, 1, 10, 50]); -- [100, 50, 10, 1, null, null]
         SELECT array_sort_desc(ARRAY [ARRAY ["a", null], null, ARRAY ["a"]); -- [["a", null], ["a"], null]
 
+.. function:: array_split_into_chunks(array(T), int) -> array(array(T))
+
+    Returns an ``array`` of arrays splitting the input ``array`` into chunks of given length.
+    If the ``array`` is not evenly divisible it will split into as many possible chunks and return
+    the left over elements for the last ``array``. Ignores null inputs, but not elements.
```
feilong-liu, line 214: > Ignores null inputs, but not elements.

I guess you refer to return NULL by "Ignores" here? Worth to add a few examples below to make it clear.

kaikalur: Actually this is pretty standard presto behavior - say returns null when any of the inputs are null to be more explicit.

aditi-pandit:
@feilong-liu : Thanks for this function. Could we add a test in https://github.com/prestodb/presto/blob/master/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeArrayFunctionQueries.java as well.

Since this is a SQL function, the native engine should be able to execute it as well.

amitkdutta:
> @feilong-liu : Thanks for this function. Could we add a test in https://github.com/prestodb/presto/blob/master/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeArrayFunctionQueries.java as well.
> 
> Since this is a SQL function, the native engine should be able to execute it without any changes.
> 
> @amitkdutta

@aditi-pandit We have disabled inline-sql-function in all our production systems. There is a historical reason for it, I heard its due to regresssion in planning time, but I am not totally sure (CC: @tdcmeehan). In java, this means planning time the function is passed as-is to worker. In worker, inline-sql-function config is not respected and java worker can expand and execute the functions. In case of C++, the expansion is not there so C++ worker fails to run the function, and in some cases, functions can be implemented in worker (both native and java) to be very efficient. Hence, we are trying to add the functions as a regular function if possible.

kaikalur:
> > @feilong-liu : Thanks for this function. Could we add a test in https://github.com/prestodb/presto/blob/master/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeArrayFunctionQueries.java as well.
> > Since this is a SQL function, the native engine should be able to execute it without any changes.
> > @amitkdutta
> 
> @aditi-pandit We have disabled inline-sql-function in all our production systems. There is a historical reason for it, I heard its due to regresssion in planning time, but I am not totally sure (CC: @tdcmeehan). In java, this means planning time the function is passed as-is to worker. In worker, inline-sql-function config is not respected and java worker can expand and execute the functions. In case of C++, the expansion is not there so C++ worker fails to run the function, and in some cases, functions can be implemented in worker (both native and java) to be very efficient. Hence, we are trying to add the functions as a regular function if possible.

This is not acceptable. SQL functions is a firstclass feature and we should support them.

kaikalur:
> @feilong-liu : Thanks for this function. Could we add a test in https://github.com/prestodb/presto/blob/master/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeArrayFunctionQueries.java as well.
> 
> Since this is a SQL function, the native engine should be able to execute it without any changes.
> 
> @amitkdutta

By definition we should be able to support all presto features in the backend especially sql functions. There are tests added and native should run all tests if they want to be compatible. It's too much burden to test both. It should be part of CI

feilong-liu (approved):

aditi-pandit:
@amitkdutta : We too have disabled inline-sql-function based on your guidance. Though the current setup is very hacky.

We are also on the wrong track to add new functionality as SQL functions. Might be better to make them Java functions, and expect a C++ implementation as well.

Agree with @kaikalur, that SQL functions should be supported at the backend.

Another option is to have a per-function annotation to enable/disable individual SQL functions.

PR #23267, opened 2024-07-21
[Iceberg] Support procedure `remove_orphan_files`

hantangwangd:
## Description

This PR support the procedure `remove_orphan_files` for iceberg. It can be used to remove files which are not referenced in any metadata files of an Iceberg table and can thus be considered "orphaned".

See examples as follow:

 - Remove any files which are not known to the table `db.sample` and older than specified timestamp::

```
CALL iceberg.system.remove_orphan_files('db', 'sample', TIMESTAMP '2023-08-31 00:00:00.000');
```

 - Remove any files which are not known to the table `db.sample` and created 3 days ago (by default)::

```
CALL iceberg.system.remove_orphan_files(schema => 'db', table_name => 'sample');
```

## Motivation and Context

Support removing orphan files that are not referenced in any metadata files for Iceberg

## Test Plan

 - Newly added test cases in `TestRemoveOrphanFilesProcedure`

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==


Iceberg Connector Changes
* Add procedure `remove_orphan_files` to remove orphan files that are not referenced in any metadata files for Iceberg. :pr:`23267`
```


presto-iceberg/src/main/java/com/facebook/presto/iceberg/procedure/RemoveOrphanFiles.java
```diff
@@ -0,0 +1,226 @@
...
+            if (!fileSystem.exists(subFolderPath)) {
+                return;
+            }
+            RemoteIterator<LocatedFileStatus> allFiles = fileSystem.listFiles(subFolderPath, true);
+            while (allFiles.hasNext()) {
+                LocatedFileStatus entry = allFiles.next();
+                if (entry.getModificationTime() <= expiration && !validFiles.contains(entry.getPath().getName())) {
+                    filesToDelete.add(entry.getPath().toString());
+                    if (filesToDelete.size() >= BATCH_DELETE_FILES_COUNT) {
+                        LOGGER.debug("Deleting %s files while removing orphan files for table %s", filesToDelete, schemaTableName);
```
kiersten-stokes, line 173: I wonder if it might be better to move the file listing to the end of the log message for ease of reading since the file names can be quite long. Something like

```suggestion
                        LOGGER.debug("Deleting the following orphan files for table %s: %s", filesToDelete, schemaTableName);
```

presto-iceberg/src/main/java/com/facebook/presto/iceberg/procedure/RemoveOrphanFiles.java
```diff
@@ -0,0 +1,226 @@
...
+                        filesToDelete.clear();
+                    }
+                }
+                else {
+                    LOGGER.debug("%s retained while removing orphan files %s", entry.getPath().toString(), schemaTableName.getTableName());
+                }
+            }
+
+            if (!filesToDelete.isEmpty()) {
+                LOGGER.debug("Deleting %s files while removing orphan files for table %s", filesToDelete, schemaTableName);
```
kiersten-stokes, line 184: And same here :)

kiersten-stokes (commented):
This will be nice to have! Just NITs from me

presto-iceberg/src/main/java/com/facebook/presto/iceberg/procedure/RemoveOrphanFiles.java
hantangwangd: Thanks for the review and suggestion, fixed! Please take a look when convenient.

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -800,6 +800,32 @@ Examples:
 
     CALL iceberg.system.expire_snapshots(schema => 'schema_name', table_name => 'table_name', snapshot_ids => ARRAY[10001, 10002]);
 
+Remove Orphan Files
+^^^^^^^^^^^^^^^^^^^
+
+Used to remove files which are not referenced in any metadata files of an Iceberg table and can thus be considered "orphaned".
```
steveburnett, line 806:
```suggestion
Use to remove files which are not referenced in any metadata files of an Iceberg table.
```

steveburnett (changes requested):
Thanks for the doc! Local doc build with the new table looks good. A minor suggested rephrase for active voice, and shortening for readability.

presto-docs/src/main/sphinx/connector/iceberg.rst
hantangwangd: Thanks for the suggestion, fixed! Please take a look when available.

steveburnett (approved):
LGTM! (docs) 

presto-iceberg/src/main/java/com/facebook/presto/iceberg/procedure/RemoveOrphanFiles.java
```diff
@@ -0,0 +1,226 @@
...
+        scanAndDeleteInvalidFiles(icebergTable, clientSession, schemaTableName, expiration, validDataFileNames.build(), "data");
+        scanAndDeleteInvalidFiles(icebergTable, clientSession, schemaTableName, expiration, validMetadataFileNames.build(), "metadata");
+    }
+
+    private void scanAndDeleteInvalidFiles(Table table, ConnectorSession session, SchemaTableName schemaTableName, long expiration, Set<String> validFiles, String subFolder)
+    {
+        try {
+            List<String> filesToDelete = new ArrayList<>();
+            FileSystem fileSystem = getFileSystem(session, this.hdfsEnvironment, schemaTableName, new Path(table.location()));
+            Path subFolderPath = new Path(table.location(), subFolder);
```
ZacBlanco, line 163: You should instead use the table properties `write.data.path` and `write.metadata.path` as they are not guaranteed to be `data` and `metadata`: https://iceberg.apache.org/docs/1.5.2/configuration/#write-properties

ZacBlanco (changes requested):
A few initial comments. Great to see us getting more parity with Iceberg's Spark procedures

presto-iceberg/src/main/java/com/facebook/presto/iceberg/procedure/RemoveOrphanFiles.java
hantangwangd: Referring to Iceberg's implementation, I extracted two methods `IcebergUtil.metadataLocation(icebergTable)` and `IcebergUtil.dataLocation(icebergTable)`. They took into account the relevant table properties to obtain metadata location and data location of the target Iceberg table. Please take a look when available, thanks!

Besides, we currently haven't supported these additional table properties in presto, so temporarily unable to add test cases to show the scenarios that table properties affecting the default value of metadata location and data location. We can add these test cases in 'TestRemoveOrphanFilesPocedure' once introducing the table properties.

ZacBlanco: > Besides, we currently haven't supported these additional table properties in presto, so temporarily unable to add test cases to show the scenarios that table properties affecting the default value of metadata location and data location.

One way to get around this might be to create a new `Table` instance through the Java API with the properties set, and then call the `register_table` procedure?

kiersten-stokes (approved):

ZacBlanco (commented):
Two final things. Also, I would like to see a test which exercises the default expiry limit. I don't have a good idea on how to do that in a reasonable time frame within the test, so I am happy to leave as is if you think it would be difficult

presto-iceberg/src/main/java/com/facebook/presto/iceberg/procedure/RemoveOrphanFiles.java
```diff
@@ -0,0 +1,232 @@
...
+        }
+        catch (IOException e) {
+            throw new PrestoException(ICEBERG_FILESYSTEM_ERROR, "Failed accessing data for table: " + schemaTableName, e);
+        }
+    }
+
+    private String extractFileName(String path)
+    {
+        return path.substring(path.lastIndexOf('/') + 1);
+    }
```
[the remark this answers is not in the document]
hantangwangd: In some rare circumstances, the authority/scheme of data files may not equal between table metadata and the real storage system. Referring to the comment for `DeleteOrphanFiles.prefixMismatchMode(...)` and test cases `testRemoveOrphanFilesProcedureWithPrefixMode()` and `testRemoveOrphanFilesProcedureWithEqualAuthorities()` in `TestRemoveOrphanFilesProcedure`.

So we use a file name rather than the fully qualified path of the file to choose an `IGNORE` strategy, that is, we do not delete the files that has mismatches in authority/scheme, which is more safer than the `DELETE` strategy.

Besides, this method has exactly the same implementation logic as `org.apache.hadoop.fs.Path.getName()`. The file paths maintained in Iceberg table metadata are all fully qualified paths, which are suitable for constructing a Hadoop Path (see the comment in `ContentFile.path()` and `ManifestFile.path()`). That means, we can omit the verification of these paths (which would be applied when create a new `Path` instance) and just get their names.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/procedure/TestRemoveOrphanFilesProcedure.java
```diff
@@ -0,0 +1,270 @@
...
+            // Generate 2 files in iceberg table metadata location
+            assertTrue(generateFile(session, TEST_SCHEMA, tableName, metadataLocation(table), "file_4.test"));
+            assertTrue(generateFile(session, TEST_SCHEMA, tableName, metadataLocation(table), "file_5.test"));
+
+            int metadataFilesCountMiddle = allMetadataFilesCount(session, table);
+            int dataFilesCountMiddle = allDataFilesCount(session, table);
+
+            // Remove all orphan files older than now
+            waitUntilAfter(System.currentTimeMillis());
+            String formattedDateTime = getTimestampString(System.currentTimeMillis(), zoneId);
```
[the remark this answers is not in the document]
hantangwangd: The provided timestamp should be `System.currentTimeMillis() + 10` if we use the method you mentioned above.

However, I think it introduces some logical complexity, although it seems more concise in code implementation.
For other people who will read these test cases in future, it's not so straightforward to express what we want to do.

The method `waitUntilAfter(currentTimeMillis())` plays a role of some kind of barrier. Meanwhile, it hardly causes any observable waiting, that is, no more than 1ms. I think compare to the cost of waiting, the benefit of making the logic more clearer is greater. What do you think?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/procedure/RemoveOrphanFiles.java
hantangwangd: Thanks, this is feasible. I have added test cases for both `HADOOP` and `HIVE` catalog separately as there are some differences between them. `HADOOP` catalog do not support table property `WRITE_METADATA_LOCATION` on table creation, while the implementation of `HIVE` catalog in presto is based on self implemented `HiveTableOperations` rather than using Iceberg lib's `HiveCatalog`, so the behavior of `createTable()/dropTable()/loadTable()` are different between the two. Please take a look when available.

hantangwangd:
> I would like to see a test which exercises the default expiry limit. I don't have a good idea on how to do that in a reasonable time frame within the test, so I am happy to leave as is if you think it would be difficult

Seems it's indeed a bit difficult to test this scenario unless we change the implementation of `remove_orphan_files` to support configuring the default value of `older_than`. I'm not so sure if it's worthing do this. Maybe we can supplement such test cases when we figure out a better time test frame?

ZacBlanco (approved):

presto-iceberg/src/main/java/com/facebook/presto/iceberg/procedure/RemoveOrphanFiles.java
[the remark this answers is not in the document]
ZacBlanco: Makes sense, thanks for the explanation

tdcmeehan (approved):

PR #23271, opened 2024-07-22
Use assertEquals instead of assertTrue when testing equality

elharo:
## Description
Use assertEquals instead of assertTrue when testing equality

## Motivation and Context
More informative failure messages

## Impact
none

## Test Plan
mvn test

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergSplitManager.java
+++ b/presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergSplitManager.java
@@ -41,6 +41,7 @@ import static com.facebook.presto.iceberg.IcebergQueryRunner.createIcebergQueryR
 import static com.facebook.presto.iceberg.IcebergSessionProperties.PUSHDOWN_FILTER_ENABLED;
 import static com.facebook.presto.spi.connector.NotPartitionedPartitionHandle.NOT_PARTITIONED;
 import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertNotNull;
 import static org.testng.Assert.assertTrue;
 import static org.testng.Assert.fail;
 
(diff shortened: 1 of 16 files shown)


presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergSplitManager.java
```diff
@@ -180,7 +180,7 @@ private void validateSplitsPlannedForSql(SplitManager splitManager,
     {
         Session session = sessionWithFilterPushdown(filterPushdown);
         List<TableScanNode> tableScanNodes = getTableScanFromOptimizedPlanOfSql(sql, session);
-        assertTrue(tableScanNodes != null && tableScanNodes.size() == 1);
+        assertEquals(tableScanNodes.size(), 1);
```
hantangwangd, line 183: This change will cause the original `AssertionError` changing to a `NPE` when `tableScanNodes` happens to be null. Feels like it went from an expected error to an unexpected one. Is that okay?

elharo: That's fine. In both cases the test fails, which is what we want. Now it's more obvious why the test failed: wrong number of nodes or null nodes, so there's more information.

ZacBlanco: Rather than throwing an NPE due to `tableScanNodes` being null I think it would be better to add an `assertNotNull` beforehand as it will be clearer to future authors that this value must be non-null

elharo: done

ZacBlanco (approved):

hantangwangd (approved):

rschlussel (approved):

Goop-arsh (commented):
T

Issue #23186, opened 2024-07-11
NaN in coordinator UI on load


elharo:
When the coordinator UI first loads it shows "NaN" for a number of values until this is resolved with initial values a few seconds in.

My guess is that it starts with default values of zero for a number of variables, divides zero by zero, gets NaN, and loads that straight into the UI without checking. It should probably instead check for NaN and display an empty string or a no data message instead. Or maybe it should check for zero divisor before it even does the division.

<img width="1422" alt="Screenshot 2024-07-11 at 7 50 33 AM" src="https://github.com/prestodb/presto/assets/1005544/4c4f9bc9-e02d-4e98-b91c-1172723cf726">

elharo:
suspect the issue is in the precisionRound function in utils.js which does not handle NaN

hussein-mamane:
hello @elharo I am new to open source contributions but I am a solo programmer for years, I need some time to dig in the code about presto, this seems to be an easy if-else fix, can I have it and a week. Thank you.

PR #23285, opened 2024-07-24
Fixing NaN in coordinator UI on load

anilsomisetty:
## Description
Currently on loading coordinator UI it shows NaN for some values, this change will check for NaN and would return Unknown in it's place.

## Motivation and Context
Fixing issue https://github.com/prestodb/presto/issues/23186

## Impact
After this change NaN would be shown as Unkown initially on load

image

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-ui/src/utils.js
+++ b/presto-ui/src/utils.js
@@ -325,6 +325,9 @@ export function computeRate(count: number, ms: number): number {
 }
 
 export function precisionRound(n: number): string {
+    if (n === undefined) {
+        return "";
+    }
     if (n < 10) {
         return n.toFixed(2);
     }


anilsomisetty:
yeah empty string seems right

image

yhwang:
LGTM with one minor comment: I believe those are statistics. Shouldn't those undefined values be `0`? unless we want to distinguish between undefined and 0.

yhwang (approved):

elharo (approved):

elharo:
Empty string seems correct to me. 0 and undefined/unknown are not the same

PR #23286, opened 2024-07-24
[native] Fix exchange HTTP executors shutdown order

Yuhta:
CPU executor should always be shutdown first, otherwise it would cause crash during shutdown.


amitkdutta (approved):
Thanks @Yuhta

Issue #22835, opened 2024-05-25
ImmutableList.of() --> Collections.emptyList()


elharo:
Now that we're on JDK 8, Collections.emptyList() is preferable to ImmutableList.of().

Should be a straightforward search and replace followed by import cleanup.

jasonmokk:
Hello, I'd like to work on this issue.

ScrapCodes:
May be I can try this!

ScrapCodes:
```
prashant@prashant:~/work/presto$ find . -type f -exec grep -nHI -e "ImmutableList.of" \{\} + | wc -l
13518
```

elharo:
I don't think that will work since you didn't include the parentheses.

ScrapCodes:
@elharo understood that and corrected myself.

tdcmeehan:
@elharo what's the motivation for this change?

elharo:
tldr; JDK > 3rd party library

https://stackoverflow.com/questions/43291202/collections-emptylist-vs-guavas-immutablelist-of

tdcmeehan:
@elharo from the SO post, I see one answer that lists two reasons:

1) Why introduce a third party dependency when you can use whatever is already built-in to the JDK?
2) The JDK factory uses a dedicated implementation for the empty list, vs. Guava uses an immutable list with no elements.

For point 1, I don't understand what the benefit is, unless we have some sort of overarching goal to remove our Guava dependency.  We already have this dependency because we prefer immutability, so I don't understand why we would want to differentiate empty vs. non-empty immutable collections.

For point 2, I simply don't understand what sort of practical benefit this has.  Can you attempt to explain it?

PR #23296, opened 2024-07-25
Immutable still good but Guava not required

elharo:
Don't specify that we prefer Guava over JDK built-in immutability

## Description
Remove "in Guava"

## Motivation and Context
Multiple PRs have pointed to this as a reason to use ImmutableList.of() instead of Collections.emptyList() etc. 

## Impact
none

## Test Plan
CI

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -196,7 +196,7 @@ We recommend you use IntelliJ as your IDE. The code style template for the proje
   time0Bto100KB.add(nanos, NANOSECONDS);
   ```
     
-  * Prefer Immutable collections in Guava when possible. For example, instead of using
+  * Prefer Immutable collections when possible. For example, instead of using
     
   ```java
   expressions.stream()


amitkdutta (approved):

jaystarshot:
Maybe we should explicitly state that using guava is not preferred and shouldn't be used if that is the goal since I saw similar comments across different PRs.  Else we should explicitly state that using both is okay.
I think Guava could be better in a way sometimes since its more explicit(using immutable means its immutable) but java  `Collections.emptyList()` may return an immutable list but its not clear. Also I am not sure if both offer the same degree of immutability.

elharo:
It depends on circumstances whether Guava is required. In Java 8 we don't have a JDK-only non-empty immutable list. In Java 9+ we do, so when we move to Java 11 as the minimum would be a good time to recommend against guava for this. But I certainly do want to allow Collections,.emptyList() for now since it's in the JDK and it is clearer about what it is than ImmutableList.of(). It's an empty list, and the JDK guarantees this is immutable.

elharo:
Hmm, just checked and my memory was incorrect. We do have Collections.unmodifiableList in Java 8 too so we really don't need Guava for this at all.

tdcmeehan (changes requested):
As I recall, the objection we had to this the last time this was brought up in #22835 was twofold:

1) Immutability is good, and there shouldn't be two ways of doing things.  Won't it clutter up the code to have to make the distinction between empty and non-empty lists?  Why not just continue to use Guava since we need that for anything non-empty?  What do we gain by introducing two ways of doing something?
2) Guava collections have a nice feature that copying elements from one to another are no-cost.

hantangwangd:
> We do have Collections.unmodifiableList in Java 8 too so we really don't need Guava for this at all.

Seems `Collections.unmodifiableList` is not really immutable, refer to the example as follows:

```
    public static void main(String[] args) throws Exception
    {
        List<String> lists = new ArrayList<>();
        lists.add("test1");
        lists.add("test2");
        lists.add("test3");
        List<String> unmodifiableList = Collections.unmodifiableList(lists);

        System.out.println(unmodifiableList);
        lists.add("test4");
        System.out.println(unmodifiableList);
    }
```

elharo:
Yes, but this list is immutable:

```
  public static void main(String[] args) throws Exception
    {
        List<String> unmodifiableList = Collections.unmodifiableList(Arrays.asList("test1", "test2", "test3"));
        System.out.println(unmodifiableList);
    }
```

The cases we care about are typically when a list is passed into or returned from a public API. In these cases, if the list is an unmodifiable view, we're good.

steveburnett (approved):
LGTM! (docs)

View .md file in GithHub to verify the text changes are valid as written. Other reviewer's requested changes should be addressed.

PR #23312, opened 2024-07-26
Allow Verifier to validate varchar as floating point

gggrace14:
## Description

Treat varchar column as double and apply floating point column validation, if it is detected that the varchar column is derived from casting floating points. This applies to the simple varchar column, array(varchar) column and map(varchar, varchar). Structured column types with deeper nested varchars are not supported.

Add verifier config --validate-string-as-double.

Resolve https://github.com/prestodb/presto/issues/23348

## Motivation and Context
Today Presto Verifier reports MISMATCH in ColumnValidator for simple columns or structured output columns composed by varchar, for real world test cases. This raises false negative and creates overhead for Verifier users utilizing the Verifier results. Verifier needs to apply floating point error margin for this case.

## Impact
Remove false negative when validating varchar columns derived from casting floating points. When --validate-string-as-double is set to true, Verifier control and test checksum queries will include more columns and become more expensive.

## Test Plan
Unit tests and production tests.

## Release Notes

```
== RELEASE NOTES ==

Verifier Changes
* Add verifier config --validate-string-as-double to control applying floating point validation to the column composed of varchar, if it is detected that the varchar column is derived from casting floating points. :pr:`23312`

### DIFF
--- a/presto-docs/src/main/sphinx/admin/verifier.rst
+++ b/presto-docs/src/main/sphinx/admin/verifier.rst
@@ -158,30 +158,75 @@ queries.
 
 * **Floating Point Columns**
     * For ``DOUBLE`` and ``REAL`` columns, 4 columns are generated for verification:
-       * Sum of the finite values of the column
-       * ``NAN`` count of the column
-       * Positive infinity count of the column
-       * Negative infinity count of the column
+        * Sum of the finite values of the column
+        * ``NAN`` count of the column
+        * Positive infinity count of the column
+        * Negative infinity count of the column
     * Checks if ``NAN`` count, positive and negative infinity count matches.
     * Checks the nullity of control sum and test sum.
     * If either control mean or test mean very close 0, checks if both are close to 0.
     * Checks the relative error between control sum and test sum.
+* **VARCHAR Columns**
+    * For ``VARCHAR`` columns, a simple checksum column is generated for verification using the :func:`!checksum`.
+    * If ``validate-string-as-double`` is set, seven columns below are generated. If ``NULL`` counts are equal before and after casting all values to ``DOUBLE``, apply the floating point validation. Otherwise, check if the simple checksum matches.
+        * Checksum
+        * Count of the ``NULL`` values
+        * Count of the ``NULL`` values after all values are casted to ``DOUBLE``
+      After casting all values to ``DOUBLE``
+        * Sum of the finite values
+        * ``NAN`` count of the values
+        * Positive infinity count of the values
+        * Negative infinity count of the values
 * **Array Columns**
-    * 2 columns are generated for verification:
-       * Sum of the cardinality
-       * Array checksum
-    * For an array column ``arr`` of type ``array(E)``:
-       * If ``E`` is not orderable, array checksum is ``checksum(arr)``.
-       * If ``E`` is orderable, array checksum ``coalesce(checksum(try(array_sort(arr))), checksum(arr))``.
+    * For an array column ``arr`` of type ``array(E)``, three columns are generated for verification:
+        * Sum of the cardinality
+        * Checksum of the cardinality
+        * Array checksum
+            * If ``E`` is not orderable, array checksum is ``checksum(arr)``.
+            * If ``E`` is orderable, array checksum is ``coalesce(checksum(try(array_sort(arr))), checksum(arr))``.
+    * If ``use-error-margin-for-floating-point-arrays`` is set and E is ``DOUBLE`` or ``REAL``, these six columns below are generated instead. Check if the sum of the cardinality matches and if the checksum of the cardinality matches. Apply the floating point validation to the rest of results.
+        * Sum of the cardinality
+        * Checksum of the cardinality
+        * Sum of the finite elements of all array values
+        * Count of the ``NAN`` elements of all array values
+        * Count of the positive infinity elements of all array values
+        * Count of the negative infinity elements of all array values
+    * If ``validate-string-as-double`` is set and E is ``VARCHAR``, these nine columns below are generated instead. Check if the sum and the checksum of the cardinality match. If ``NULL`` counts are equal before and after casting all array elements to ``DOUBLE``, apply the floating point validation. Otherwise, check if the array checksum matches.
+        * Sum of the cardinality
+        * Checksum of the cardinality
+        * Array checksum ``checksum(array_sort(arr))``
+        * Count of the ``NULL`` elements of all array values
+        * Count of the ``NULL`` elements after all array elements are casted to ``DOUBLE``
+      After casting all array elements to ``DOUBLE``
+        * Sum of the finite elements of all array values
+        * Count of the ``NAN`` elements of all array values
+        * Count of the positive infinity elements of all array values
+        * Count of the negative infinity elements of all array values
 * **Map Columns**
-    * 4 columns are generated for verification:
-       * Sum of the cardinality
-       * Checksum of the map
-       * Array checksum of the key set
-       * Array checksum of the value set
+    * For a map column of type ``map(K, V)``, four columns are generated for verification:
+        * Sum of the cardinality
+        * Checksum of the map
+        * Array checksum of the key set
+        * Array checksum of the value set
+    * If ``validate-string-as-double`` is set and K is ``VARCHAR``, six additional columns are generated:
+        * Count of the ``NULL`` elements of all key sets
+        * Count of the ``NULL`` elements of the key sets after all map keys are casted to ``DOUBLE``
+      After casting all map keys to ``DOUBLE``
+        * Sum of the finite elements of all key sets
+        * Count of the ``NAN`` elements of all key sets
+        * Count of the positive infinity elements of all key sets
+        * Count of the negative infinity elements of all key sets
+    * If ``validate-string-as-double`` is set and V is ``VARCHAR``, six additional columns are generated:
+        * Count of the ``NULL`` elements of all value sets
+        * Count of the ``NULL`` elements of the value sets after all map values are casted to ``DOUBLE``
+      After casting all map values to ``DOUBLE``
+        * Sum of the finite elements of all value sets
+        * Count of the ``NAN`` elements of all value sets
+        * Count of the positive infinity elements of all value sets
+        * Count of the negative infinity elements of all value sets
 * **Row Columns**
     * Checksums row fields recursively according to the type of the fields.
-* For all other column types, generates a simple checksum using the :func:`!checksum` function.
+For all other column types, generates a simple checksum using the :func:`!checksum` function.
 
 Determinism
 -----------
--- a/presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnValidator.java
+++ b/presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnValidator.java
@@ -13,6 +13,7 @@
  */
 package com.facebook.presto.verifier.checksum;
 
+import com.facebook.presto.common.type.AbstractVarcharType;
 import com.facebook.presto.common.type.ArrayType;
 import com.facebook.presto.common.type.RowType;
 import com.facebook.presto.common.type.Type;
@@ -49,57 +50,45 @@ public class ArrayColumnValidator
 {
     private final FloatingPointColumnValidator floatingPointValidator;
     private final boolean useErrorMarginForFloatingPointArrays;
+    private final boolean validateStringAsDouble;
 
     @Inject
     public ArrayColumnValidator(VerifierConfig config, FloatingPointColumnValidator floatingPointValidator)
     {
         this.floatingPointValidator = requireNonNull(floatingPointValidator, "floatingPointValidator is null");
         this.useErrorMarginForFloatingPointArrays = config.isUseErrorMarginForFloatingPointArrays();
+        this.validateStringAsDouble = config.isValidateStringAsDouble();
     }
 
     @Override
     public List<SingleColumn> generateChecksumColumns(Column column)
     {
         Type columnType = column.getType();
-        boolean useFloatingPointPath = useFloatingPointPath(column);
+        ImmutableList.Builder<SingleColumn> builder = ImmutableList.builder();
 
-        // For arrays of floating point numbers we have a different processing, akin to FloatingPointColumnValidator.
-        if (useFloatingPointPath) {
-            Type elementType = ((ArrayType) columnType).getElementType();
-            Expression expression = elementType.equals(DOUBLE) ? column.getExpression() : new Cast(column.getExpression(), new ArrayType(DOUBLE).getDisplayName());
-
-            Expression sum = functionCall(
-                    "sum",
-                    functionCall("array_sum", functionCall("filter", expression, generateLambdaExpression("is_finite"))));
-            Expression nanCount = functionCall(
-                    "sum",
-                    functionCall("cardinality", functionCall("filter", expression, generateLambdaExpression("is_nan"))));
-            Expression posInfCount = functionCall(
-                    "sum",
-                    functionCall("cardinality", functionCall("filter", expression, generateInfinityLambdaExpression(ArithmeticUnaryExpression.Sign.PLUS))));
-            Expression negInfCount = functionCall(
-                    "sum",
-                    functionCall("cardinality", functionCall("filter", expression, generateInfinityLambdaExpression(ArithmeticUnaryExpression.Sign.MINUS))));
-            Expression arrayCardinalityChecksum = functionCall("checksum", functionCall("cardinality", expression));
-            Expression arrayCardinalitySum = new CoalesceExpression(
-                    functionCall("sum", functionCall("cardinality", expression)), new LongLiteral("0"));
-            return ImmutableList.of(
-                    new SingleColumn(sum, Optional.of(delimitedIdentifier(FloatingPointColumnValidator.getSumColumnAlias(column)))),
-                    new SingleColumn(nanCount, Optional.of(delimitedIdentifier(FloatingPointColumnValidator.getNanCountColumnAlias(column)))),
-                    new SingleColumn(posInfCount, Optional.of(delimitedIdentifier(FloatingPointColumnValidator.getPositiveInfinityCountColumnAlias(column)))),
-                    new SingleColumn(negInfCount, Optional.of(delimitedIdentifier(FloatingPointColumnValidator.getNegativeInfinityCountColumnAlias(column)))),
-                    new SingleColumn(arrayCardinalityChecksum, Optional.of(delimitedIdentifier(getCardinalityChecksumColumnAlias(column)))),
-                    new SingleColumn(arrayCardinalitySum, Optional.of(delimitedIdentifier(getCardinalitySumColumnAlias(column)))));
-        }
-
+        // coalesce(checksum(try(array_sort(array_column))), checksum(array_column))
         Expression checksum = generateArrayChecksum(column.getExpression(), columnType);
+        // checksum(cardinality(array_column))
         Expression arrayCardinalityChecksum = functionCall("checksum", functionCall("cardinality", column.getExpression()));
+        // coalesce(sum(cardinality(array_column)), 0)
         Expression arrayCardinalitySum = new CoalesceExpression(
                 functionCall("sum", functionCall("cardinality", column.getExpression())), new LongLiteral("0"));
-        return ImmutableList.of(
-                new SingleColumn(checksum, Optional.of(delimitedIdentifier(getChecksumColumnAlias(column)))),
-                new SingleColumn(arrayCardinalityChecksum, Optional.of(delimitedIdentifier(getCardinalityChecksumColumnAlias(column)))),
-                new SingleColumn(arrayCardinalitySum, Optional.of(delimitedIdentifier(getCardinalitySumColumnAlias(column)))));
+
+        // For arrays of floating point numbers we have a different processing, akin to FloatingPointColumnValidator.
+        if (useFloatingPointPath(column)) {
+            builder.addAll(generateFloatingPointArrayChecksumColumns(column));
+        }
+        else if (useStringAsDoublePath(column)) {
+            builder.add(new SingleColumn(checksum, delimitedIdentifier(getChecksumColumnAlias(column))));
+            builder.addAll(generateStringArrayChecksumColumns(column));
+        }
+        else {
+            builder.add(new SingleColumn(checksum, delimitedIdentifier(getChecksumColumnAlias(column))));
+        }
+        builder.add(new SingleColumn(arrayCardinalityChecksum, delimitedIdentifier(getCardinalityChecksumColumnAlias(column))));
+        builder.add(new SingleColumn(arrayCardinalitySum, delimitedIdentifier(getCardinalitySumColumnAlias(column))));
+
+        return builder.build();
     }
 
     @Override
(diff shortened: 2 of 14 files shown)

### CODE REVIEW

**Comment by aditi-pandit** [comment]
@gggrace14 : Thanks for this code.

Did you intend to Advance Velox in this PR ? Please can you separate it to its own PR. We are suggesting that as a best practice.

**Comment by gggrace14** [comment]
> @gggrace14 : Thanks for this code.
> 
> Did you intend to Advance Velox in this PR ? Please can you separate it to its own PR. We are suggesting that as a best practice.

Hi @aditi-pandit , I'm not meant to advance velox by this change. I'll change to leave that part out. I forgot this matter.

**Comment by steveburnett** [comment]
Suggest change to release note entry to follow the [Release Notes Guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines): 

```
== RELEASE NOTES ==

Verifier Changes
* Add configuration property --validate-string-as-double to control applying floating point validation to columns composed of varchar, if it is detected that the varchar column is derived from casting floating points. :pr:`23312`
```

**Review by woshizhouyihao — APPROVED** [review]

**Review by amitkdutta — COMMENTED** [review]
Genearlly looks OK to me. Thanks so much @gggrace14 for working on it. I have one question: why don't we make `validate-string-as-double` default to `true`? Any issue you envision by not making it default?

**Review by mbasmanova — COMMENTED** [review]
@gggrace14 Ge, let's document this enhancement in https://prestodb.io/docs/current/admin/verifier.html

Do we have other recent enhancements documented already? If not, let's prioritize documenting these as well.

CC: @steveburnett

File: presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/MapColumnValidator.java [hunk 1]
```diff
@@ -23,79 +24,190 @@
...
 
+        builder.add(new SingleColumn(checksum, Optional.of(delimitedIdentifier(getChecksumColumnAlias(column)))));
+        builder.add(new SingleColumn(keysChecksum, Optional.of(delimitedIdentifier(getKeysChecksumColumnAlias(column)))));
+
         // We need values checksum in one case only: when key is a floating point type and value is not.
         // In such case, when both column checksum and the key checksum do not match, we cannot tell if the values match or not.
         // Meaning we cannot resolve column mismatch, because value type is not a floating type and a mismatch in the values could indicate a real correctness issue.
         // In order to resolve column mismatch in such a situation, generate an extra checksum for the values.
-        if (isFloatingPointType(keyType) && !isFloatingPointType(valueType)) {
+        if ((isFloatingPointType(keyType) || shouldValidateStringAsDouble(keyType)) && !isFloatingPointType(valueType)) {
```
rschlussel, line 80: should we also skip this if the value type is a string that's really a double?

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/MapColumnValidator.java
```diff
@@ -23,79 +24,190 @@
...
+            Column keysColumn = getAsDoubleArrayColumn(getKeysColumn(column));
+            keysFloatingPointChecksum = Optional.of(FloatingPointColumnValidator.toColumnChecksum(keysColumn, checksumResult, checksumResult.getRowCount()));
+        }
+        Optional<FloatingPointColumnChecksum> valuesFloatingPointChecksum = Optional.empty();
+        if (isDoubleValueAsString) {
+            Column valuesColumn = getAsDoubleArrayColumn(getValuesColumn(column));
+            valuesFloatingPointChecksum = Optional.of(FloatingPointColumnValidator.toColumnChecksum(valuesColumn, checksumResult, checksumResult.getRowCount()));
+        }
+        Object valuesChecksum = null;
+        if ((isFloatingPointType(keyType) || isDoubleKeyAsString) && !isFloatingPointType(valueType)) {
```
rschlussel, line 169: same question about why don't we skip if the value type isDoubleValueAsString

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/MapColumnValidator.java
```diff
@@ -23,79 +24,190 @@
...
         Expression mapCardinalityChecksum = functionCall("checksum", functionCall("cardinality", column.getExpression()));
+        // coalesce(sum(cardinality(map_column)), 0)
         Expression mapCardinalitySum = new CoalesceExpression(
                 functionCall("sum", functionCall("cardinality", column.getExpression())),
                 new LongLiteral("0"));
 
+        builder.add(new SingleColumn(checksum, Optional.of(delimitedIdentifier(getChecksumColumnAlias(column)))));
+        builder.add(new SingleColumn(keysChecksum, Optional.of(delimitedIdentifier(getKeysChecksumColumnAlias(column)))));
+
         // We need values checksum in one case only: when key is a floating point type and value is not.
```
rschlussel, line 76: If we're doing floating point validation on the map now does this still apply? wouldn't we always want to do floating point validation on the values if they are floating point types?

gggrace14:
> @gggrace14 Ge, let's document this enhancement in https://prestodb.io/docs/current/admin/verifier.html
> 
> Do we have other recent enhancements documented already? If not, let's prioritize documenting these as well.
> 
> CC: @steveburnett

@mbasmanova Masha, sure, I will add to verifier.rst and include in this PR.
For recent enhancement with --control/test.reuse-table and --function-substitutes, I've added the documentation to verifier.rst. I will check again if I need to add more to other parts of verifier.rst.

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ColumnValidatorUtil.java
```diff
@@ -0,0 +1,83 @@
...
+            // count_if(column is null).
+            return ImmutableList.of(
+                    new SingleColumn(functionCall("count_if",
+                            new IsNullPredicate(column.getExpression())), delimitedIdentifier(getNullCountColumnAlias(column))),
+                    new SingleColumn(functionCall("count_if",
+                            new IsNullPredicate(asDoubleColumn.getExpression())), delimitedIdentifier(getNullCountColumnAlias(asDoubleColumn))));
+        }
+    }
+
+    // Check if the column is derived from a floating point column by comparing the count of the original null values and the count of null values when casting the column back
```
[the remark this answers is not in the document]
gggrace14: Revised with your suggestion. Yes, that's exactly what I meant.

gggrace14:
> Genearlly looks OK to me. Thanks so much @gggrace14 for working on it. I have one question: why don't we make `validate-string-as-double` default to `true`? Any issue you envision by not making it default?

@amitkdutta, thanks a lot for the review!
Right now with this change, turning on --validate-string-as-double will add a couple aggregate column to the CHECKSUM queries. They're potentially expensive and make the CHECKSUM queries run longer in the control cluster. Already observed slowness of the CHECKSUM queries and we have 3 pairs of CONTROL & TEST CHECKSUM queries. So our approach is to only turn on this config when retesting the cases with COLUMN_MISMATCH previously.

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/ArrayColumnValidator.java
```diff
@@ -166,37 +205,37 @@ private static Expression generateLambdaExpression(String functionName)
...
-                        (long) nanCount,
-                        (long) checksumResult.getChecksum(FloatingPointColumnValidator.getPositiveInfinityCountColumnAlias(column)),
-                        (long) checksumResult.getChecksum(FloatingPointColumnValidator.getNegativeInfinityCountColumnAlias(column)),
-                        cardinalitySum)));
+        Object checksum = checksumResult.getChecksum(getChecksumColumnAlias(column));
+
+        if (useStringAsDoublePath) {
+            Column asDoubleArrayColumn = getAsDoubleArrayColumn(column);
+            return new ArrayColumnChecksum(
+                    checksum,
```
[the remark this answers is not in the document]
gggrace14: Revised and I can set it to null. This field will not be used for the branch of useStringAsDouble. I included it for debugging.

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/MapColumnValidator.java
gggrace14: Nope, for map of floating points, we're not doing floating points validation today. And I think we need to add that. I see you discussed about this in https://github.com/prestodb/presto/pull/22793#issuecomment-2127656021. And I agree. Either me or Sergey could add that later. But it needs a separate PR, given this PR is already very large.

rschlussel: I think I'm a bit confused because we are generating all the columns for floating point values in maps as we would for floating point validation.  Is the difference that we are then comparing that they are exactly equal rather than using an error margin?

presto-docs/src/main/sphinx/admin/verifier.rst
```diff
@@ -166,19 +166,70 @@ queries.
     * Checks the nullity of control sum and test sum.
     * If either control mean or test mean very close 0, checks if both are close to 0.
     * Checks the relative error between control sum and test sum.
+* **VARCHAR Columns**
+    * For ``VARCHAR`` columns, a simple checksum column is generated for verification using :func:`!checksum`.
+    * If ``validate-string-as-double`` is set, 6 additional columns below are generated. If ``NULL`` counts
+      are equal before and after casting all values to ``DOUBLE``, apply the aforementioned floating point validation.
```
steveburnett, line 172:
```suggestion
      are equal before and after casting all values to ``DOUBLE``, apply the floating point validation.
```

presto-docs/src/main/sphinx/admin/verifier.rst
```diff
@@ -166,19 +166,70 @@ queries.
     * Checks the nullity of control sum and test sum.
     * If either control mean or test mean very close 0, checks if both are close to 0.
     * Checks the relative error between control sum and test sum.
+* **VARCHAR Columns**
+    * For ``VARCHAR`` columns, a simple checksum column is generated for verification using :func:`!checksum`.
+    * If ``validate-string-as-double`` is set, 6 additional columns below are generated. If ``NULL`` counts
```
steveburnett, line 171:
```suggestion
    * If ``validate-string-as-double`` is set, these six columns are generated. If ``NULL`` counts
```

presto-docs/src/main/sphinx/admin/verifier.rst
```diff
@@ -166,19 +166,70 @@ queries.
...
-       * If ``E`` is not orderable, array checksum is ``checksum(arr)``.
-       * If ``E`` is orderable, array checksum ``coalesce(checksum(try(array_sort(arr))), checksum(arr))``.
+    * For an array column ``arr`` of type ``array(E)``, 3 columns are generated for verification:
+        * Sum of the cardinality
+        * Checksum of the cardinality
+        * Array checksum
+            * If ``E`` is not orderable, array checksum is ``checksum(arr)``.
+            * If ``E`` is orderable, array checksum is ``coalesce(checksum(try(array_sort(arr))), checksum(arr))``.
+    * If ``use-error-margin-for-floating-point-arrays`` is set and E is ``DOUBLE`` or ``REAL``,
+      6 columns below are generated instead. Check if the sum of the cardinality matches and if the checksum of the cardinality
```
steveburnett, line 189:
```suggestion
      these six columns are generated instead. Check if the sum of the cardinality matches and if the checksum of the cardinality
```

presto-docs/src/main/sphinx/admin/verifier.rst
```diff
@@ -166,19 +166,70 @@ queries.
...
+      matches. And apply the aforementioned floating point validation to the rest of results.
```
steveburnett, line 190:
```suggestion
      matches. Apply the aforementioned floating point validation to the rest of results.
```

presto-docs/src/main/sphinx/admin/verifier.rst
```diff
@@ -166,19 +166,70 @@ queries.
...
     * If either control mean or test mean very close 0, checks if both are close to 0.
     * Checks the relative error between control sum and test sum.
+* **VARCHAR Columns**
+    * For ``VARCHAR`` columns, a simple checksum column is generated for verification using :func:`!checksum`.
+    * If ``validate-string-as-double`` is set, 6 additional columns below are generated. If ``NULL`` counts
+      are equal before and after casting all values to ``DOUBLE``, apply the aforementioned floating point validation.
+      Otherwise, check if the simple checksum matches.
+        * Count of the ``NULL`` values
+        * Count of the ``NULL`` values after all values are casted to ``DOUBLE``
+        * After casting all values to ``DOUBLE``
```
steveburnett, line 176:
```suggestion
        After casting all values to ``DOUBLE``
```
Suggest not formatting this line as an item in the list, to help the reader by showing them six items instead of the formatting suggesting there are seven.

presto-docs/src/main/sphinx/admin/verifier.rst
```diff
@@ -166,19 +166,70 @@ queries.
...
-    * 2 columns are generated for verification:
-       * Sum of the cardinality
-       * Array checksum
-    * For an array column ``arr`` of type ``array(E)``:
-       * If ``E`` is not orderable, array checksum is ``checksum(arr)``.
-       * If ``E`` is orderable, array checksum ``coalesce(checksum(try(array_sort(arr))), checksum(arr))``.
+    * For an array column ``arr`` of type ``array(E)``, 3 columns are generated for verification:
+        * Sum of the cardinality
+        * Checksum of the cardinality
+        * Array checksum
```
steveburnett, line 185:
```suggestion
        Array checksum
```
Suggest not formatting this line as an item in the list, to help the reader. The two sub-items formatting is good, as your description clearly tells the reader the next two are an either/or.

presto-docs/src/main/sphinx/admin/verifier.rst
```diff
@@ -166,19 +166,70 @@ queries.
...
+    * If ``validate-string-as-double`` is set and E is ``VARCHAR``, 9 columns below are generated instead.
+      Check if the sum and the checksum of the cardinality match. If ``NULL`` counts are equal before and after
+      casting all array elements to ``DOUBLE``, apply the aforementioned floating point validation. Otherwise, check
+      if the array checksum matches.
+        * Sum of the cardinality
+        * Checksum of the cardinality
+        * Array checksum ``checksum(array_sort(arr))``
+        * Count of the ``NULL`` elements of all array values
+        * Count of the ``NULL`` elements after all array elements are casted to ``DOUBLE``
+        * After casting all array elements to ``DOUBLE``
```
steveburnett, line 206:
```suggestion
        After casting all array elements to ``DOUBLE``
```
Suggest not formatting this line as an item in the list, to help the reader.

presto-docs/src/main/sphinx/admin/verifier.rst
```diff
@@ -166,19 +166,70 @@ queries.
...
+        * Count of the ``NULL`` elements of the key sets after all map keys are casted to ``DOUBLE``
+        * After casting all map keys to ``DOUBLE``
+            * Sum of the finite elements of all key sets
+            * Count of the ``NAN`` elements of all key sets
+            * Count of the positive infinity elements of all key sets
+            * Count of the negative infinity elements of all key sets
+    * Additionally if ``validate-string-as-double`` is set and V is ``VARCHAR``, 6 additional columns are generated:
+        * Count of the ``NULL`` elements of all value sets
+        * Count of the ``NULL`` elements of the value sets after all map values are casted to ``DOUBLE``
+        * After casting all map values to ``DOUBLE``
```
steveburnett, line 228:
```suggestion
        After casting all map values to ``DOUBLE``
```
Suggest not formatting this line as an item in the list, to help the reader.

presto-docs/src/main/sphinx/admin/verifier.rst
```diff
@@ -166,19 +166,70 @@ queries.
...
+        * Array checksum of the value set
+    * If ``validate-string-as-double`` is set and K is ``VARCHAR``, 6 additional columns are generated:
+        * Count of the ``NULL`` elements of all key sets
+        * Count of the ``NULL`` elements of the key sets after all map keys are casted to ``DOUBLE``
+        * After casting all map keys to ``DOUBLE``
+            * Sum of the finite elements of all key sets
+            * Count of the ``NAN`` elements of all key sets
+            * Count of the positive infinity elements of all key sets
+            * Count of the negative infinity elements of all key sets
+    * Additionally if ``validate-string-as-double`` is set and V is ``VARCHAR``, 6 additional columns are generated:
```
steveburnett, line 225:
```suggestion
    * If ``validate-string-as-double`` is set and V is ``VARCHAR``, 6 additional columns are generated:
```

presto-docs/src/main/sphinx/admin/verifier.rst
```diff
@@ -166,19 +166,70 @@ queries.
...
-       * Array checksum of the value set
+    * For a map column of type ``map(K, V)``, 4 columns are generated for verification:
+        * Sum of the cardinality
+        * Checksum of the map
+        * Array checksum of the key set
+        * Array checksum of the value set
+    * If ``validate-string-as-double`` is set and K is ``VARCHAR``, 6 additional columns are generated:
+        * Count of the ``NULL`` elements of all key sets
+        * Count of the ``NULL`` elements of the key sets after all map keys are casted to ``DOUBLE``
+        * After casting all map keys to ``DOUBLE``
```
steveburnett, line 220:
```suggestion
        After casting all map keys to ``DOUBLE``
```
Suggest not formatting this line as an item in the list, to help the reader.

presto-docs/src/main/sphinx/admin/verifier.rst
```diff
@@ -166,19 +166,70 @@ queries.
     * Checks the nullity of control sum and test sum.
     * If either control mean or test mean very close 0, checks if both are close to 0.
     * Checks the relative error between control sum and test sum.
+* **VARCHAR Columns**
+    * For ``VARCHAR`` columns, a simple checksum column is generated for verification using :func:`!checksum`.
+    * If ``validate-string-as-double`` is set, 6 additional columns below are generated. If ``NULL`` counts
+      are equal before and after casting all values to ``DOUBLE``, apply the aforementioned floating point validation.
+      Otherwise, check if the simple checksum matches.
+        * Count of the ``NULL`` values
```
steveburnett, line 174: Error in local doc build: 

`/Users/steveburnett/Documents/GitHub/presto/presto-docs/src/main/sphinx/admin/verifier.rst:174: ERROR: Unexpected indentation.`

Suggest unindenting this list, and any nested list within it, and do a local doc build until these similar errors do not appear. See [Building the Documentation](https://github.com/prestodb/presto/tree/master/presto-docs#building-the-documentation) for more information. 

Screenshot of local doc build, showing a vertical gray line in the generated file. 
Screenshot 2024-08-05 at 11 27 58 AM

presto-docs/src/main/sphinx/admin/verifier.rst
```diff
@@ -166,19 +166,70 @@ queries.
...
+    * For an array column ``arr`` of type ``array(E)``, 3 columns are generated for verification:
+        * Sum of the cardinality
+        * Checksum of the cardinality
+        * Array checksum
+            * If ``E`` is not orderable, array checksum is ``checksum(arr)``.
+            * If ``E`` is orderable, array checksum is ``coalesce(checksum(try(array_sort(arr))), checksum(arr))``.
+    * If ``use-error-margin-for-floating-point-arrays`` is set and E is ``DOUBLE`` or ``REAL``,
+      6 columns below are generated instead. Check if the sum of the cardinality matches and if the checksum of the cardinality
+      matches. And apply the aforementioned floating point validation to the rest of results.
+        * Sum of the cardinality
```
steveburnett, line 191: Error in local doc build: 

`/Users/steveburnett/Documents/GitHub/presto/presto-docs/src/main/sphinx/admin/verifier.rst:191: ERROR: Unexpected indentation.`

Suggest unindenting this list, and any nested list within it, and do a local doc build until these similar errors do not appear. See [Building the Documentation](https://github.com/prestodb/presto/tree/master/presto-docs#building-the-documentation) for more information.

presto-docs/src/main/sphinx/admin/verifier.rst
```diff
@@ -166,19 +166,70 @@ queries.
...
+        * Checksum of the cardinality
+        * Sum of the finite elements of all array values
+        * Count of the ``NAN`` elements of all array values
+        * Count of the positive infinity elements of all array values
+        * Count of the negative infinity elements of all array values
+    * If ``validate-string-as-double`` is set and E is ``VARCHAR``, 9 columns below are generated instead.
+      Check if the sum and the checksum of the cardinality match. If ``NULL`` counts are equal before and after
+      casting all array elements to ``DOUBLE``, apply the aforementioned floating point validation. Otherwise, check
+      if the array checksum matches.
+        * Sum of the cardinality
```
steveburnett, line 201: Error in local doc build: 

`/Users/steveburnett/Documents/GitHub/presto/presto-docs/src/main/sphinx/admin/verifier.rst:201: ERROR: Unexpected indentation.`

Suggest unindenting this list, and any nested list within it, and do a local doc build until these similar errors do not appear. See [Building the Documentation](https://github.com/prestodb/presto/tree/master/presto-docs#building-the-documentation) for more information.

steveburnett (changes requested):
Thanks for the doc! A few nits of formatting and suggestions of rephrasing, and some errors in the local doc build about indenting lists and nested lists.

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/MapColumnValidator.java
gggrace14: Hi Rebecca, for the branch of map of floating point, we're not generating columns for floating point validation today.   And I keep it as it is with this PR. We do the checksum exact match. So this MapColumnValidator component probably often fails I guess. (Attaching the logic block below) There's another component following this ColumnValidator component in the verifier workflow, called FailureResolver. For map mismatch returned by ColumnValidator, if it sees the map has a floating point key, the FailureResolver auto resolves the mismatches and considers it correct.

I think this MapColumnValidator + FailureResolver for map is too inaccurate. And we should do floating point validation in MapColumnValidator, and get rid of FailureResolve. The recent https://github.com/prestodb/presto/pull/22143 improved ArrayColumnValidator to do floating point validation for array. And we should make the same change for map. That needs a separate PR.

Then in this PR which is for map of varchar, I make the path for map of varchar directly do floating point validation. We're generating floating point columns only for this path.

https://github.com/prestodb/presto/blob/7dae0d6f98ecba6a0b6edd1c073f0b4cbdba460d/presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/MapColumnValidator.java#L73-L82

https://github.com/prestodb/presto/blob/f16032e9e395afc4576eea0337c74cbee3920569/presto-verifier/src/main/java/com/facebook/presto/verifier/resolver/StructuredColumnMismatchResolver.java#L76-L77

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/MapColumnValidator.java
gggrace14: Hi Rebecca, as in another comment, after this MapColumnValidator component, there's another related component, called FailureResolver. MapColumnValidator does checksum exact match today instead of floating point validation. When MapColumnValidator returns mismatch, if the FailureResolver sees the map comprised of floating point, the FailureResolver auto resolves the mismatch and considers it correct.

The reason that we skip here today for floating point is because of this FailureResolver approach. And I think we should get rid of FailureResolver and make MapColumnValidator do floating point validation, as ArrayColumnValidator does wtih https://github.com/prestodb/presto/pull/22143, for floating point.

In this PR for map of varchar, I directly do floating point validation for varchar. So we need to check on both key column and value column, during which we do floating point validation if either of them is a floating point column or a varchar column that's really a double.

presto-docs/src/main/sphinx/admin/verifier.rst
gggrace14: @steveburnett , thanks a lot for pointing to the local build. That's very convenient! I've fixed the build errors.

presto-docs/src/main/sphinx/admin/verifier.rst
gggrace14: Need to keep this as one item in the list. "Array checksum" is one of the three columns generated.

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/MapColumnValidator.java
rschlussel: I see so am I correct that with your change if it's a double string we'll use floating point validation if the config property is set, but if it's just a double we don't do floating point validation?  Can we have floating point types go through the same path?

steveburnett (dismissed):
LGTM! (docs)

Pull updated branch, new local doc build, everything looks good. Thanks!

presto-verifier/src/main/java/com/facebook/presto/verifier/checksum/MapColumnValidator.java
gggrace14: > I see so am I correct that with your change if it's a double string we'll use floating point validation if the config property is set, but if it's just a double we don't do floating point validation?

@rschlussel Rebecca, yes, exactly, that is what this PR has.

Regarding having the true floating point type go through the same path, yes, we could and should do that next. The floating point validation is abstracted into a method in FloatingPointColumnValidator. And we could call that, after refactoring a bit of this MapColumnValidator.validate(). We prefer to make it a separate PR. These changes need quite a bit testing to make sure free of false positive and negative. I would like to separate the change.

rschlussel (approved):
Thanks for the explanations.  The current state after this PR is pretty weird where maps with numeric strings will use floating point validation but maps with actual double types don't do that.  Like you mentioned, we need a follow up to bring floating point keys/values in line with string as double and consolidate the code.  As long as you have plans to do that, it's fine to merge this first.

gggrace14:
> Thanks for the explanations. The current state after this PR is pretty weird where maps with numeric strings will use floating point validation but maps with actual double types don't do that. Like you mentioned, we need a follow up to bring floating point keys/values in line with string as double and consolidate the code. As long as you have plans to do that, it's fine to merge this first.

I see. Agree that the state is weird, and needs a lot of details if to communicate with others. We would add the floating point validation for map as the next step.

PR #23315, opened 2024-07-26
[Do not review][Native] Do not carry over partitioning in StreamPropertyDerivations for nested loop joins

aditi-pandit:
## Description
Nested joins do not maintain order of rows in Native execution. So do not carry the properties forward in StreamPropertyDerivations class.

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* ... :pr:`12345`
* ... :pr:`12345`

Hive Connector Changes
* ... :pr:`12345`
* ... :pr:`12345`
```

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/StreamPropertyDerivations.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/StreamPropertyDerivations.java
@@ -98,20 +98,20 @@ public final class StreamPropertyDerivations
 {
     private StreamPropertyDerivations() {}
 
-    public static StreamProperties derivePropertiesRecursively(PlanNode node, Metadata metadata, Session session, TypeProvider types, SqlParser parser)
+    public static StreamProperties derivePropertiesRecursively(PlanNode node, Metadata metadata, Session session, TypeProvider types, SqlParser parser, boolean nativeExecution)
     {
         List<StreamProperties> inputProperties = node.getSources().stream()
-                .map(source -> derivePropertiesRecursively(source, metadata, session, types, parser))
+                .map(source -> derivePropertiesRecursively(source, metadata, session, types, parser, nativeExecution))
                 .collect(toImmutableList());
-        return StreamPropertyDerivations.deriveProperties(node, inputProperties, metadata, session, types, parser);
+        return StreamPropertyDerivations.deriveProperties(node, inputProperties, metadata, session, types, parser, nativeExecution);
     }
 
-    public static StreamProperties deriveProperties(PlanNode node, StreamProperties inputProperties, Metadata metadata, Session session, TypeProvider types, SqlParser parser)
+    public static StreamProperties deriveProperties(PlanNode node, StreamProperties inputProperties, Metadata metadata, Session session, TypeProvider types, SqlParser parser, boolean nativeExecution)
     {
-        return deriveProperties(node, ImmutableList.of(inputProperties), metadata, session, types, parser);
+        return deriveProperties(node, ImmutableList.of(inputProperties), metadata, session, types, parser, nativeExecution);
     }
 
-    public static StreamProperties deriveProperties(PlanNode node, List<StreamProperties> inputProperties, Metadata metadata, Session session, TypeProvider types, SqlParser parser)
+    public static StreamProperties deriveProperties(PlanNode node, List<StreamProperties> inputProperties, Metadata metadata, Session session, TypeProvider types, SqlParser parser, boolean nativeExecution)
     {
         requireNonNull(node, "node is null");
         requireNonNull(inputProperties, "inputProperties is null");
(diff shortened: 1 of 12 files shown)


aditi-pandit:
@karteekmurthys

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/StreamPropertyDerivations.java
```diff
@@ -185,9 +187,16 @@ public StreamProperties visitJoin(JoinNode node, List<StreamProperties> inputPro
                             .translate(column -> PropertyDerivations.filterOrRewrite(outputs, node.getCriteria(), column))
                             .unordered(unordered);
                 case LEFT:
-                    return leftProperties
-                            .translate(column -> PropertyDerivations.filterIfMissing(outputs, column))
-                            .unordered(unordered);
+                    if (nativeExecution && node.getCriteria().isEmpty()) {
+                        // This maps to a NestedLoopJoin in Native engine. The NestedLoopJoin output is not
```
mbasmanova, line 191: @rschlussel @feilong-liu Rebecca, Feilong, would you help take a first pass for this change? At a high level this seems reasonable, but I haven't been working in the optimizer code for quite some time.

presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/ValidateStreamingAggregations.java
```diff
@@ -43,10 +43,22 @@
 public class ValidateStreamingAggregations
         implements Checker
 {
+    private final boolean nativeExecution;
+
+    public ValidateStreamingAggregations()
```
rschlussel, line 48: why do we need this zero argument constructor?

presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/ValidateStreamingJoins.java
```diff
@@ -45,10 +45,22 @@
 public class ValidateStreamingJoins
         implements Checker
 {
+    private final boolean nativeExecution;
+
+    public ValidateStreamingJoins()
```
rschlussel, line 50: why do we need the zero argument constructor?

presto-spark-base/src/main/java/com/facebook/presto/spark/planner/optimizers/PickJoinSides.java
```diff
@@ -104,7 +104,8 @@ public Result apply(JoinNode joinNode, Captures captures, Context context)
         // if we don't have exact costs for the join, but based on source tables we think the left side
         // is very small or much smaller than the right, then flip the join.
         if (rightSize > leftSize || (isSizeBasedJoinDistributionTypeEnabled(context.getSession()) && (Double.isNaN(leftSize) || Double.isNaN(rightSize)) && isLeftSideSmall(joinNode, context))) {
-            rewrittenNode = createRuntimeSwappedJoinNode(joinNode, metadata, sqlParser, context.getLookup(), context.getSession(), context.getVariableAllocator(), context.getIdAllocator());
+            // This is never used for Prestissimo.
```
rschlussel, line 107: i don't like that this assumption isn't enforced by the code.  Could we instead pass the native execution config to this optimizer like we do with other optimizers (may need to inject FeaturesConfig into adaptivePlanOptimizers, but that's fine).

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/StreamPropertyDerivations.java
rschlussel: does only left join with no criteria produce nested loop join?  What happens for inner join?  Also, would it be enough to just change unordered to false for these cases, but otherwise keep the other things the same (does the stream distribution or partitioning change)?

aditi-pandit: @rschlussel : That's a good point. I was trying to contain this fix, but this change would be applicable for inner too. This change has come by in correlated queries that seem to use LEFT join predominantly.

But just changing unordered to false didn't work. I had to change the partitioning as well for the change to take effect. The addLocalExchanges seems to leverage distribution as well. More at https://github.com/prestodb/presto/issues/22585#issuecomment-2168942249

aditi-pandit:
@rschlussel : Thanks for your review. 

I hear more recently that Velox team is planning to fix NestedLoopJoin. This PR will not be required if the Velox change happens.

This PR has wide changes in the optimizer. So I will resume work on it only if the Velox change will not happen.

pedroerp:
This PR should address the NLJ issue in Velox, so this probably won't be needed: 
 https://github.com/facebookincubator/velox/pull/10651

aditi-pandit:
Closing as the underlying NestedLoopJoin in Velox is fixed to maintain probe row ordering now. https://github.com/facebookincubator/velox/pull/10651

Issue #23252, opened 2024-07-18
Equality semantics of TIMESTAMP WITH TIME ZONE type can cause inconsistent behavior


bikramSingh91:
Equality semantics of TIMESTAMP WITH TIME ZONE type can cause inconsistent behavior

**Context:**
TIMESTAMP WITH TIME ZONE (further abbreviated as TS_w_TZ) logical type is backed by BIGINT physical type. The timestamp values are stored in memory as 64-bit integers using an encoding that doesn't allow for direct comparisons of these integers.
It physically packs two integers in a single 64-bit word, using 12 bits for timezone ID, and 52 bits for a millisecond-precision timestamp in UTC.

Therefore, two TS_w_TZ values are compared using only their UTC time, ignoring the timezone information. This means that 4 PM EST would compare as equal to 1 PM PST. 
https://github.com/prestodb/presto/blob/master/presto-main/src/main/java/com/facebook/presto/type/TimestampWithTimeZoneOperators.java#L79

While this may seem reasonable at first glance, it leads to inconsistencies in certain scenarios. For example:

Consider the following:
```
with a as (select from_iso8601_timestamp(x) as ts from unnest(array['2024-01-01T04:01:02-04:00', '2024-01-01T01:01:02-07:00']) as t(x))
select a.ts, hour(a.ts) from a;

 ts                             | _col1 
--------------------------------+-------
 2024-01-01 04:01:02.000 -04:00 |     4 
 2024-01-01 01:01:02.000 -07:00 |     1 
```

Now, if we join this table with itself on the "ts" columns as the join key, we should ideally expect 4 results, which are all possible combinations: (a, a), (a, b), (b, a), (b, b) as follows:
```
  ts                             | _col1 | ts_(1)                         | _col3 
--------------------------------+-------+--------------------------------+-------
 2024-01-01 04:01:02.000 -04:00 |     4 | 2024-01-01 04:01:02.000 -04:00 |     4 
 2024-01-01 04:01:02.000 -04:00 |     4 | 2024-01-01 01:01:02.000 -07:00 |     1 
 2024-01-01 01:01:02.000 -07:00 |     1 | 2024-01-01 04:01:02.000 -04:00 |     4 
 2024-01-01 01:01:02.000 -07:00 |     1 | 2024-01-01 01:01:02.000 -07:00 |     1 
```

However, the actual result we get is  (a, a), (a, a), (b, b), (b, b) as follows:

```
[...]

PR #23319, opened 2024-07-29
Fix assignments for equijoin on TimestampWithTimeZone

tdcmeehan:
TimestampWithTimeZone can be represented in multiple equivalant ways. This breaks an optimization introduced by f6c2252 because while the values may be semantically equivalent, the contents may still differ.  To support this, and future types that may end up needing this support (such as JSON), a new flag is introduced to the type to indicate whether or not the values of the type may support more than one representation.

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Fix equijoin over timestamp with timezone types (:pr:`23319`)

SPI Changes
* Add ``equalValuesAreIdentical`` to ``Type``.  Override this method to return ``false`` when the values of the type may have more than one canonical representation.
```

### DIFF
--- a/presto-common/src/main/java/com/facebook/presto/common/type/TimestampWithTimeZoneType.java
+++ b/presto-common/src/main/java/com/facebook/presto/common/type/TimestampWithTimeZoneType.java
@@ -29,6 +29,19 @@ public final class TimestampWithTimeZoneType
         super(parseTypeSignature(StandardTypes.TIMESTAMP_WITH_TIME_ZONE));
     }
 
+    /**
+     * Timestamp with time zone represents a single point in time.  Multiple timestamps with timezones may
+     * each refer to the same point in time.  For example, 9:00am in New York is the same point in time as
+     * 2:00pm in London.  While those two timestamps may be encoded differently, they each refer to the same
+     * point in time.  Therefore, it's possible encode multiple timestamps which each represent the same
+     * point in time, and hence it's not safe to use equality as a proxy for identity.
+     */
+    @Override
+    public boolean equalValuesAreIdentical()
+    {
+        return false;
+    }
+
     @Override
     public Object getObjectValue(SqlFunctionProperties properties, Block block, int position)
     {
--- a/presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/UnaliasSymbolReferences.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/UnaliasSymbolReferences.java
@@ -578,7 +578,7 @@ public class UnaliasSymbolReferences
 
             if (node.getType().equals(INNER)) {
                 canonicalCriteria.stream()
-                        .filter(clause -> clause.getLeft().getType().equals(clause.getRight().getType()))
+                        .filter(clause -> clause.getLeft().getType().equals(clause.getRight().getType()) && clause.getLeft().getType().equalValuesAreIdentical())
                         .filter(clause -> node.getOutputVariables().contains(clause.getLeft()))
                         .forEach(clause -> map(clause.getRight(), clause.getLeft()));
             }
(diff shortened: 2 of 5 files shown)


presto-common/src/main/java/com/facebook/presto/common/type/TimestampWithTimeZoneType.java
```diff
@@ -29,6 +29,12 @@ private TimestampWithTimeZoneType()
         super(parseTypeSignature(StandardTypes.TIMESTAMP_WITH_TIME_ZONE));
     }
 
+    @Override
+    public boolean valuesAreCanonical()
```
rschlussel, line 33: I think something like `equalValuesAreIdentical()` would be clearer. I wouldn't know what "valuesAreCanonical" is supposed to mean without looking it up.

rschlussel (commented):
change looks reasonable (pending tests, etc).

presto-common/src/main/java/com/facebook/presto/common/type/TimestampWithTimeZoneType.java
tdcmeehan: Sounds reasonable to me.

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/UnaliasSymbolReferences.java
```diff
@@ -578,7 +578,7 @@ public PlanNode visitJoin(JoinNode node, RewriteContext<Void> context)
 
             if (node.getType().equals(INNER)) {
                 canonicalCriteria.stream()
-                        .filter(clause -> clause.getLeft().getType().equals(clause.getRight().getType()))
```
jaystarshot, line 581: Shouldn't we not skip canonicalization of the joinCriteria itself? [here](https://github.com/prestodb/presto/blob/9651f49cafa297f68460b04ad41b5d02695ac664/presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/UnaliasSymbolReferences.java#L813)

tdcmeehan: This statement maps an equijoin condition from the build side to the probe side.  So if `a = b`, and `a` and `b` are both types where `equalValuesAreIdentical` is `true`, then there's no difference between them, and the optimizer can just say `b` is an alias for `a`.  Downstream, we will canonicalize all references to the output variables from this join.  So when we encounter downstream references to `b`, when we canonicalize it, we'll map it to `a`.  So in effect, `b` becomes `a` downstream.  This change prevents that from happening--`b` will still be `b` if its a tstz.

jaystarshot: Ahh i see i thought the canonicalization has already happened in [here](https://github.com/prestodb/presto/blob/9651f49cafa297f68460b04ad41b5d02695ac664/presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/UnaliasSymbolReferences.java#L813) whereas that was just replacement.

feilong-liu (approved):

jaystarshot (approved):

rschlussel (approved):

aaneja (approved):
I think we do a similar de-aliasing in other places, like [EqualityInference#generateEqualitiesPartitionedBy](https://github.com/prestodb/presto/blob/01293f696c9aed537ce68672e22575a631fdd1fa/presto-main/src/main/java/com/facebook/presto/sql/planner/EqualityInference.java#L190) that may need to be fixed too.  
I don't have a repro case for this though. For now, I think this is good fix to merge

tdcmeehan:
Let's reopen #23252 if similar cases are discovered.

PR #23321, opened 2024-07-29
[native] Fix HTTP client connection pool shutdown order

Yuhta:
Connection pool needs to be destroyed after CPU threads are joined but before IO threads are joined, otherwise we got crashes using it in CPU threads.


amitkdutta (dismissed):
Looks good. Thanks @Yuhta 
It will be great if you shadow real traffic for some time and do a few restarts just to verify nothing else regresses.

amitkdutta (approved):
Looks good. Thanks @Yuhta for iterating on it.

aditi-pandit (approved):
Thanks @Yuhta

Issue #22031, opened 2024-02-27
TestHudiIntegration test failures in pipeline


czentgr:
https://github.com/prestodb/presto/actions/runs/8070233507/job/22047081223?pr=22024

## Your Environment
n/a

## Current Behavior
 Error:  Failures: 
Error:    TestHudiIntegration.testDemoQuery1:94->AbstractTestQueryFramework.assertQuery:159 Execution of 'actual' query failed: SELECT symbol, max(ts) FROM stock_ticks_mor_rt GROUP BY symbol HAVING symbol = 'GOOG'
Error:    TestHudiIntegration.testDemoQuery2:117->AbstractTestQueryFramework.assertQuery:159 Execution of 'actual' query failed: SELECT "_hoodie_record_key", symbol, ts, volume, open, close  FROM stock_ticks_mor_rt WHERE symbol = 'GOOG'
Error:    TestHudiIntegration.testQueryWithPartitionColumn:137->AbstractTestQueryFramework.assertQuery:159 Execution of 'actual' query failed: SELECT symbol, ts, dt FROM stock_ticks_mor_rt WHERE symbol = 'GOOG' AND dt = '2018-08-31'

Main cause appears to be

```
Caused by: java.lang.UnsupportedOperationException: Not implemented by the RawLocalFileSystem FileSystem implementation
	at org.apache.hadoop.fs.FileSystem.getScheme(FileSystem.java:219)
	at org.apache.hadoop.fs.HadoopExtendedFileSystem.getScheme(HadoopExtendedFileSystem.java:71)
	at org.apache.hudi.common.fs.FSUtils.isGCSFileSystem(FSUtils.java:649)
	at org.apache.hudi.common.table.log.HoodieLogFileReader.getFSDataInputStream(HoodieLogFileReader.java:501)
	at org.apache.hudi.common.table.log.HoodieLogFileReader.<init>(HoodieLogFileReader.java:120)
	at org.apache.hudi.common.table.log.HoodieLogFormatReader.<init>(HoodieLogFormatReader.java:70)
	at org.apache.hudi.common.table.log.AbstractHoodieLogRecordReader.scanInternalV1(AbstractHoodieLogRecordReader.java:245)
	... 32 more
```

## Steps to Reproduce
n/a

imjalpreet:
Also observed this here: https://github.com/prestodb/presto/actions/runs/8530290591/job/23367795134?pr=21622

elharo:
[...]

PR #23323, opened 2024-07-29
array_min/array_max throw if first element is an inner Array with a null

kevinwilfong:
## Description
array_min and array_max will always throw an exception if the first element is an inner array with a null.

## Motivation and Context
array_min and array_max always compare the first element with itself.  Besides being unnecessary this produces inconsistent behavior where if the first element is an array (or there is an array recursively in a complex type) that contains a null the function will throw  "ARRAY comparison not supported for arrays with null elements".

This is because the element is of course equal to itself, but everything in it needs to be compared to verify this.

For other elements, the comparison will exit early as soon as it can be confirmed that it is less than/greater than the current candidate, so it may not throw even if it contains an array with a null element.

## Impact
This should slightly improve performance by avoiding an unnecessary comparison, but will also cause the function to throw only if it needs to compare a null element of an inner array in the first element with the corresponding value in a later element.

## Test Plan
Added unit tests to verify a null in the first element does not necessarily cause the function to throw, but it does throw if it is necessary to compare that null with another value.

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

== RELEASE NOTES ==

General Changes
* Fix array_min/array_max to not necessarily throw if the first element recursively contains an inner array with a null element. :pr:``
```


presto-main/src/test/java/com/facebook/presto/type/TestArrayOperators.java
```diff
@@ -690,6 +690,10 @@ public void testArrayMin()
         assertFunction("ARRAY_MIN(ARRAY ['puppies', 'kittens'])", createVarcharType(7), "kittens");
         assertFunction("ARRAY_MIN(ARRAY [TRUE, FALSE])", BOOLEAN, false);
         assertFunction("ARRAY_MIN(ARRAY [NULL, FALSE])", BOOLEAN, null);
+        assertFunction("ARRAY_MIN(ARRAY [ARRAY[1, NULL], ARRAY[2, NULL]])", new ArrayType(INTEGER), Lists.newArrayList(1, null));
```
elharo, line 693: Can we have new test methods instead please? Makes it easier to debug failures.

elharo (commented):
Are these functions documented somewhere? If so, does that need to be updated?

kevinwilfong:
@elharo Thanks for the review!

The functions are documented here

https://github.com/prestodb/presto/blob/772213c4f93bc77b22ffa1c5ae9f781697f86933/presto-docs/src/main/sphinx/functions/array.rst?plain=1#L123-L129

But it does not say anything about this behavior that would need to be updated.

kevinwilfong:
The failing check appears to be some TestHudiIntegration tests that are broken in trunk https://github.com/prestodb/presto/issues/22031

presto-main/src/test/java/com/facebook/presto/type/TestArrayOperators.java
```diff
@@ -696,6 +696,15 @@ public void testArrayMin()
         assertDecimalFunction("ARRAY_MIN(ARRAY [2.22222222222222222, 2.3])", decimal("2.22222222222222222"));
     }
 
+    @Test
+    public void testArrayMinWithNulls()
+    {
+        assertFunction("ARRAY_MIN(ARRAY [ARRAY[1, NULL], ARRAY[2, NULL]])", new ArrayType(INTEGER), Lists.newArrayList(1, null));
```
elharo, line 702: These should be four separate test methods so they can pass or fail independently.

presto-main/src/test/java/com/facebook/presto/type/TestArrayOperators.java
```diff
@@ -726,6 +735,15 @@ public void testArrayMax()
         assertDecimalFunction("ARRAY_MAX(ARRAY [2.22222222222222222, 2.3])", decimal("2.30000000000000000"));
     }
 
+    @Test
+    public void testArrayMaxWithNulls()
```
elharo, line 739: These should be four separate test methods so they can pass or fail independently.

rschlussel (approved):

elharo (dismissed):

xiaoxmeng (approved):

rschlussel (approved):

Issue #21304, opened 2023-11-03
IGNORE NULLs in Window aggregates doesn't seem to work


aditi-pandit:
Window aggregate SQL support IGNORE NULLs clauses. It seems like the results with and without this clause are the same wrt NULLs in the input data.

Examples
Function array_agg(c) over (partition by a order by b rows between 1 preceding and current row)  with and without IGNORE NULLs retuns the same results.
```
presto:tiny> select a, array_agg(c) over (partition by a order by b rows between 1 preceding and current row) from (values (1, 1, 3), (1, 2, null), (1, 4, 2)) as t(a, b, c);
 a |   _col1   
---+-----------
 1 | [3]       
 1 | [3, null] 
 1 | [null, 2] 
(3 rows)
```

```
presto:tiny> select a, array_agg(c) ignore nulls over (partition by a order by b rows between 1 preceding and current row) from (values (1, 1, 3), (1, 2, null), (1, 4, 2)) as t(a, b, c);
 a |   _col1   
---+-----------
 1 | [3]       
 1 | [3, null] 
 1 | [null, 2] 
(3 rows)

```
## Expected Behavior
array_agg as a regular aggregate doesn't have any explicit IGNORE NULLs option so there isn't a basis for comparison. But the results I would've expected is 

```
presto:tiny> select a, array_agg(c) ignore nulls over (partition by a order by b rows between 1 preceding and current row) from (values (1, 1, 3), (1, 2, null), (1, 4, 2)) as t(a, b, c);
 a |   _col1   
---+-----------
 1 | [3]       
 1 | [3] 
 1 | [2] 
(3 rows)
```

Most other systems I checked like DuckDB error on IGNORE NULLs in aggregates. If we don't support any special behavior, then Presto should error as well.

Akanksha-kedia:
Postgresql:

<img width="583" alt="Screenshot 2023-11-17 at 5 40 14 PM" src="https://github.com/prestodb/presto/assets/89628774/fe6db3e6-544c-4ba5-8556-d81d492305ed">

Mysql:
<img width="567" alt="Screenshot 2023-11-17 at 5 41 47 PM" src="https://github.com/prestodb/presto/assets/89628774/2edc90d1-ae3f-485f-b00e-77eee5fe2a6e">

aditi-pandit:
@mbasmanova : Any thoughts ?

mbasmanova:
[...]

PR #23325, opened 2024-07-29
Add warning for IGNORE NULL with Window agggregates

auden-woolfson:
## Description
Added a warning when an IGNORE NULL clause is used on any non lag, lead, first or last value function. Presto will not throw an error in other cases, but it also will not ignore null values. See issue #21304 

== RELEASE NOTES ==

General Changes
* Add a warning when an IGNORE NULL clause is used on any non lag, lead, first, last, or nth value function.  :pr:`23325`

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/sql/analyzer/ExpressionAnalyzer.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/analyzer/ExpressionAnalyzer.java
@@ -47,6 +47,7 @@ import com.facebook.presto.spi.security.AccessControl;
 import com.facebook.presto.spi.security.DenyAllAccessControl;
 import com.facebook.presto.sql.parser.SqlParser;
 import com.facebook.presto.sql.planner.TypeProvider;
+import com.facebook.presto.sql.relational.FunctionResolution;
 import com.facebook.presto.sql.tree.ArithmeticBinaryExpression;
 import com.facebook.presto.sql.tree.ArithmeticUnaryExpression;
 import com.facebook.presto.sql.tree.ArrayConstructor;
@@ -209,6 +210,7 @@ public class ExpressionAnalyzer
     private static final int MAX_NUMBER_GROUPING_ARGUMENTS_INTEGER = 31;
 
     private final FunctionAndTypeResolver functionAndTypeResolver;
+    private final FunctionResolution functionResolution;
     private final Function<Node, StatementAnalyzer> statementAnalyzerFactory;
     private final TypeProvider symbolTypes;
     private final boolean isDescribe;
@@ -261,6 +263,7 @@ public class ExpressionAnalyzer
             Map<NodeRef<Expression>, Type> outerScopeSymbolTypes)
     {
         this.functionAndTypeResolver = requireNonNull(functionAndTypeResolver, "functionAndTypeResolver is null");
+        this.functionResolution = new FunctionResolution(functionAndTypeResolver);
         this.statementAnalyzerFactory = requireNonNull(statementAnalyzerFactory, "statementAnalyzerFactory is null");
         this.transactionId = requireNonNull(transactionId, "transactionId is null");
         this.sessionFunctions = requireNonNull(sessionFunctions, "sessionFunctions is null");
@@ -1113,6 +1116,13 @@ public class ExpressionAnalyzer
                     if (!sortKeyType.isOrderable()) {
                         throw new SemanticException(TYPE_MISMATCH, node, "ORDER BY can only be applied to orderable types (actual: %s)", sortKeyType.getDisplayName());
                     }
+                }
+            }
+
+            if (node.isIgnoreNulls() && node.getWindow().isPresent()) {
+                if (!functionResolution.isWindowValueFunction(function)) {
+                    String warningMessage = createWarningMessage(node, "IGNORE NULLS is not used for aggregate and ranking window functions. This will cause queries to fail in future versions.");
+                    warningCollector.add(new PrestoWarning(SEMANTIC_WARNING, warningMessage));
                 }
             }
 
--- a/presto-main/src/main/java/com/facebook/presto/sql/relational/FunctionResolution.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/relational/FunctionResolution.java
@@ -56,6 +56,12 @@ public final class FunctionResolution
         implements StandardFunctionResolution
 {
     private final FunctionAndTypeResolver functionAndTypeResolver;
+    private final List<QualifiedObjectName> windowValueFunctions = ImmutableList.of(
+            QualifiedObjectName.valueOf(DEFAULT_NAMESPACE, "lead"),
+            QualifiedObjectName.valueOf(DEFAULT_NAMESPACE, "lag"),
+            QualifiedObjectName.valueOf(DEFAULT_NAMESPACE, "first_value"),
+            QualifiedObjectName.valueOf(DEFAULT_NAMESPACE, "last_value"),
+            QualifiedObjectName.valueOf(DEFAULT_NAMESPACE, "nth_value"));
 
     public FunctionResolution(FunctionAndTypeResolver functionAndTypeResolver)
     {
@@ -359,5 +365,10 @@ public final class FunctionResolution
     public boolean isElementAtFunction(FunctionHandle functionHandle)
     {
         return functionAndTypeResolver.getFunctionMetadata(functionHandle).getName().equals(QualifiedObjectName.valueOf(DEFAULT_NAMESPACE, "element_at"));
+    }
+
+    public boolean isWindowValueFunction(FunctionHandle functionHandle)
+    {
+        return windowValueFunctions.contains(functionAndTypeResolver.getFunctionMetadata(functionHandle).getName());
     }
 }
--- a/presto-main/src/test/java/com/facebook/presto/sql/analyzer/TestAnalyzer.java
+++ b/presto-main/src/test/java/com/facebook/presto/sql/analyzer/TestAnalyzer.java
@@ -29,11 +29,13 @@ import com.facebook.presto.spi.WarningCollector;
 import com.facebook.presto.spiller.NodeSpillConfig;
 import com.facebook.presto.sql.planner.CompilerConfig;
 import com.facebook.presto.tracing.TracingConfig;
+import com.google.common.collect.ImmutableList;
 import org.testng.annotations.Test;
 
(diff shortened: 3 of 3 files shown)


auden-woolfson:
Easy CLA should be resolved once I make/push more changes. Had to set local git config

presto-main/src/main/java/com/facebook/presto/sql/analyzer/ExpressionAnalyzer.java
```diff
@@ -1155,6 +1155,13 @@ else if (frame.getType() == GROUPS) {
                 }
             }
 
+            if (node.isIgnoreNulls()) {
+                ImmutableList<String> allowedIgnoreNullFunctionNames = ImmutableList.of("lead", "lag", "first_value", "last_value");
```
tdcmeehan, line 1159: Let's not hardcode this.  Instead, let's create a new method in `FunctionResolution` that identifies a window value function.  I think it will be OK to hardcode it, but inside there, make it a constant.  Also, I think you're missing one, Nth value.  Double check this list by looking at all subclasses of `ValueWindowFunction`.

presto-main/src/main/java/com/facebook/presto/sql/analyzer/ExpressionAnalyzer.java
```diff
@@ -1155,6 +1155,13 @@ else if (frame.getType() == GROUPS) {
...
+                if (!allowedIgnoreNullFunctionNames.contains(node.getName().toString())) {
+                    warningCollector.add(new PrestoWarning(SEMANTIC_WARNING, "IGNORE NULLS does not work for aggregate and ranking window functions"));
```
tdcmeehan, line 1161: Also mention that it will start failing in the future.

tdcmeehan (changes requested):
Great work!  Some suggestions.

presto-main/src/main/java/com/facebook/presto/sql/relational/FunctionResolution.java
```diff
@@ -56,6 +56,13 @@ public final class FunctionResolution
         implements StandardFunctionResolution
 {
     private final FunctionAndTypeResolver functionAndTypeResolver;
+    private final ImmutableList<QualifiedObjectName> windowValueFunctions = ImmutableList.of(
```
tdcmeehan, line 59:
```suggestion
    private final List<QualifiedObjectName> windowValueFunctions = ImmutableList.of(
```

presto-main/src/main/java/com/facebook/presto/sql/analyzer/ExpressionAnalyzer.java
```diff
@@ -1155,6 +1156,14 @@ else if (frame.getType() == GROUPS) {
                 }
             }
 
+            if (node.isIgnoreNulls()) {
+                FunctionResolution functionResolution = new FunctionResolution(functionAndTypeResolver);
```
tdcmeehan, line 1160: Let's make the resolution a member variable of the class

presto-main/src/main/java/com/facebook/presto/sql/analyzer/ExpressionAnalyzer.java
auden-woolfson: Thanks for the feedback. Made all the updates you suggested. I added the list of window value functions to FunctionResolution. Is this the right place for it to live?

tdcmeehan: Yes that's good.  Next let's look for a place to put some tests.  We should be able to construct some simple queries that show a warning is emitted for functions which are not window value functions, and no warning is emitted for window value functions.

presto-main/src/test/java/com/facebook/presto/sql/analyzer/TestAnalyzer.java
```diff
@@ -150,6 +152,28 @@ void testNoORWarning()
...
+                "LAST_VALUE(c)",
+                "LEAD(c, 1)",
+                "LAG(c, 1)");
+
+        for (String function : valueFunctions) {
+            assertNoWarning(analyzeWithWarnings("SELECT a, " + function + " IGNORE NULLS OVER\n" +
+                    "(ORDER BY b)\n FROM (VALUES (1, 1, 3), (1, 2, null), (1, 4, 2)) AS t(a, b, c)"));
+        }
+
+        assertHasWarning(
```
aditi-pandit, line 170: Please add more tests with various aggregate functions and those in this list https://prestodb.io/docs/current/functions/window.html#ranking-functions

presto-main/src/main/java/com/facebook/presto/sql/analyzer/ExpressionAnalyzer.java
```diff
@@ -1155,6 +1158,13 @@ else if (frame.getType() == GROUPS) {
                 }
             }
 
+            if (node.isIgnoreNulls()) {
```
aditi-pandit, line 1161: Most of the window function semantic checks happen here https://github.com/prestodb/presto/blob/master/presto-main/src/main/java/com/facebook/presto/sql/analyzer/ExpressionAnalyzer.java#L981. Any particular reason we didn't add this check after the frame validation in that block ?

@tdcmeehan

aditi-pandit:
@auden-woolfson : Thanks for working on this issue. Please can you merge all your commits into a single one.

presto-main/src/main/java/com/facebook/presto/sql/analyzer/ExpressionAnalyzer.java
auden-woolfson: There wasn't any reason before that I had placed it on that exact line, but it needs to be after the function variable has been initialized. I moved it back to line 1122.

presto-main/src/main/java/com/facebook/presto/sql/analyzer/ExpressionAnalyzer.java
```diff
@@ -1116,6 +1119,13 @@ else if (frame.getType() == GROUPS) {
...
+                if (!functionResolution.isWindowValueFunction(function)) {
+                    String warningMessage = createWarningMessage(node, "IGNORE NULLS does not work for aggregate and ranking window functions, this will cause queries to fail in future versions");
```
tdcmeehan, line 1124:
```suggestion
                    String warningMessage = createWarningMessage(node, "IGNORE NULLS is not used for aggregate and ranking window functions.  This will cause queries to fail in future versions.");
```

presto-main/src/test/java/com/facebook/presto/sql/analyzer/TestAnalyzer.java
```diff
@@ -150,6 +152,36 @@ void testNoORWarning()
         assertNoWarning(analyzeWithWarnings("SELECT * FROM t1 JOIN t2 ON t1.a = t2.a \n" + "AND (t1.b = t2.b OR t1.b > t2.b)"));
     }
 
+    @Test
+    void testIgnoreNullWarning()
```
tdcmeehan, line 156:
```suggestion
    public void testIgnoreNullWarning()
```

tdcmeehan (commented):
LGTM % nits

tdcmeehan:
Code looks good.  Let's reword the commit, something like

`Warn when a non-window value function is specified to ignore nulls`

jaystarshot (commented):
Is this code path executed only once? Can there be multiple warnings for a single query

auden-woolfson:
> Is this code path executed only once? Can there be multiple warnings for a single query

I just tested this with the following query....

`SELECT 
    a,
    array_agg(c) 
    IGNORE NULLS 
    OVER (
        PARTITION BY a 
        ORDER BY b 
        ROWS BETWEEN 1 PRECEDING AND CURRENT ROW
    ) AS agg_1,
    array_agg(c) 
    IGNORE NULLS 
    OVER (
        PARTITION BY a 
        ORDER BY b 
        ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW
    ) AS agg_2
FROM 
    (VALUES (1, 1, 3), (1, 2, NULL), (1, 4, 2)) AS t(a, b, c);
`

I did get two errors, it will error out for each individual "invalid" ignore null.

presto-main/src/main/java/com/facebook/presto/sql/analyzer/ExpressionAnalyzer.java
```diff
@@ -1116,6 +1119,13 @@ else if (frame.getType() == GROUPS) {
                 }
             }
 
+            if (node.isIgnoreNulls()) {
```
aditi-pandit, line 1122: @auden-woolfson : I generally preferred this check to be in the block for window since the validation is specific for it. But if we want to keep this check here, then we should also add node.getWindow().isPresent() to this condition.

auden-woolfson: @aditi-pandit,

For the first option of bringing this logic into the `node.getWindow().isPresent()` conditional block, this would require restructuring the entire function. Since we need the handle, all of the logic surrounding `argumentTypesBuilder` (and a few other lines of code) would need to be put at the top of the function.

I tried the second option of adding the window is present condition to the ignore nulls condition and this is causing tests to fail. `node.getWindow()` is not present when `node.isIgnoreNulls()`. I think the window is only present when the functions to the right of the `OVER` clause in the SQL query are being visited.

The tests are passing, I believe the current implementation is sound.

aditi-pandit: @auden-woolfson : I am fine with the current changes.

aditi-pandit (approved):
Thanks @auden-woolfson

tdcmeehan (approved):
Please add an issue to follow up and convert this warning into a failure after the next release has been cut.

tdcmeehan:
@auden-woolfson please add a release note.  See other PRs and our contributing guide for examples.

steveburnett (approved):
Reviewed text of warning, looks good to me!

steveburnett:
For information on writing and formatting a release note entry, see the [Release Notes Guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines). As a starting point, maybe something like this following draft: 

```
== RELEASE NOTES ==

General Changes
* Add a warning when an IGNORE NULL clause is used on any non lag, lead, first or last value function.  :pr:`23325`
```

PR #23341, opened 2024-07-31
[WIP] [Native] Disable Streaming aggregations in Prestissimo.

kgpai:
## Description
This PR will cause query convertor to throw when converting a plan with streaming aggregations.

## Motivation and Context
Currently there is a bug with Streaming aggregations when the source is a Nested loop join. 

## Test Plan
Adding test cases.


aditi-pandit:
@kgpai @amitkdutta : We have encountered this issue already (https://github.com/prestodb/presto/issues/22585). This drastic fix is not needed.

You can correct the grouping behavior to avoid streaming aggregation using https://github.com/prestodb/presto/pull/22998
The more elaborate fix is to avoid generating such plans in the optimizer https://github.com/prestodb/presto/pull/23315

amitkdutta:
> @kgpai @amitkdutta : We have encountered this issue already (#22585). This drastic fix is not needed.
> 
> You can correct the grouping behavior to avoid streaming aggregation using #22998 The more elaborate fix is to avoid generating such plans in the optimizer #23315
> 
> If there is urgency then #22998 is better for you.

CC: @pedroerp @mbasmanova

presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp
```diff
@@ -960,6 +960,13 @@ VeloxQueryPlanConverterBase::toVeloxQueryPlan(
   std::vector<std::string> aggregateNames;
   std::vector<core::AggregationNode::Aggregate> aggregates;
 
+  // Do not allow streaming aggregations
+  auto streamable = !node->preGroupedVariables.empty()
+      && node->groupingSets.groupingSetCount == 1
+      && node->groupingSets.globalGroupingSets.empty();
+  VELOX_CHECK(!streamable,
```
rschlussel, line 967: Is it possible to just use a regular aggregation instead of failing?

aditi-pandit: @rschlussel : Yes. You are correct. 
There is a nuance though, HashAggregation also has preGrouping related optimizations which streamable should disable.

That is the fix in https://github.com/prestodb/presto/pull/22998.

aditi-pandit:
> > @kgpai @amitkdutta : We have encountered this issue already (#22585). This drastic fix is not needed.
> > You can correct the grouping behavior to avoid streaming aggregation using #22998 The more elaborate fix is to avoid generating such plans in the optimizer #23315
> > If there is urgency then #22998 is better for you.
> 
> CC: @pedroerp @mbasmanova

@pedroerp @mbasmanova : For context we had tried fixing the NestedLoopJoin in Velox to maintain the grouping property on LHS, but there was lot of overhead to maintain RHS to give such behavior. ref https://github.com/facebookincubator/velox/pull/10048

kgpai:
@aditi-pandit Based on inputs from @rschlussel its probably safer to turn off NLJ till its fixed.

aditi-pandit:
> @aditi-pandit Based on inputs from @rschlussel its probably safer to turn off NLJ till its fixed.

@kgpai @rschlussel : Agree this is a more conservative fix. Please feel free to go ahead with it if you prefer.

@karteekmurthys

rschlussel (approved):

aditi-pandit (dismissed):
Thanks @kgpai

kgpai:
I am seeing a couple of failures related to docker compose: https://github.com/prestodb/presto/actions/runs/10204170880/job/28232117475?pr=23341
``` 
++ docker-compose version
presto-hive-hadoop2/bin/common.sh: line 79: docker-compose: command not found
```

kgpai:
@rschlussel @amitkdutta @aditi-pandit Please have a look again, I have gated this under a configuration.

karteekmurthys:
@kgpai Would you please also disable streaming agg when nestedloopjoin is enabled?

aditi-pandit:
> @kgpai Would you please also disable streaming agg when nestedloopjoin is enabled?
> 
> cc: @aditi-pandit

@karteekmurthys : Let Krishna proceed with this change to keep it minimal. You can add the correction in a subsequent PR since you are more familiar with those code-paths.

@kgpai , @amitkdutta

kgpai:
Unfortunately there are a lot more tests than I anticipated that depend on NLJ.

pedroerp:
This PR should address the NLJ issue in Velox, so hopefully this won't be needed:  https://github.com/facebookincubator/velox/pull/10651

Issue #23000, opened 2024-06-13
Add fail-fast function validation support for Presto C++


tdcmeehan:
## Background

Unfenced functions are functions which execute within the same process as the underlying evaluation engine.  Typically, this term is used in the context of user-defined functions, and it is meant to contrast with fenced functions, which execute in a separate process.  Unfenced functions are more efficient than fenced functions, but they are also more dangerous, as they can crash the process in which they are running.  Presto has supported unfenced functions for a very long time, and they are a convenient way to efficiently add new functions in Presto.

Currently, the Presto analyzer only validates functions that are registered through the Presto functions SPI, and also built-in functions which are implemented inJava.  C++ functions are not validated at query planning time, which can lead to runtime errors if the function is not implemented correctly or if the function only exists in C++.

## Expected Behavior or Use Case

To support RFC-0003, and to enable an equivalent SPI that allows for the registration of built-in functions written in C++, we need enhance Presto's SPI to allow for the registration of functions which are not present in-memory in the Java runtime.  This issue is to create a new SPI which will allow out of process, yet built-in, functions to be registered to the Presto analyzer, which will allow these functions to be planned in the same way as existing Java functions, and also to quickly validate incorrect usage of these functions.

## Presto Component, Service, or Connector
Presto SPI, Presto Sidecar, Native execution module

## Possible Implementation
#22829 added a new sidecar process type to Presto.  This sidecar is a separate process which shares the same code as the Presto C++ worker.

A new endpoint will be added to the Presto sidecar which will return the function mapping for all built-in and externally registered functions which are implemented in C++.  This will allow the Presto analyzer to validate the function signatures of these functions at query planning time.

(Currently, there is no way to add new functions to the C++ engine without forking Prestissimo.  However, a separate issue will be created which will enable registering such functions as an externally loaded shared library.  Once this feature is enabled, then we can consider this to be a new SPI which allows for the registration of functions which are not present in-memory in the Java runtime or built-in to the Presto C++ engine.)

An enhancement to the SPI will be added in the Java codebase which will allow for the registration of functions which are not present in-memory in the Java runtime.  This SPI will be used to register the functions which are returned by the sidecar process.  This SPI is the same as the existing `FunctionNamespaceManager` SPI, with some important additions:

* Currently, there is no support in this SPI for registering functions which take in parametric types.  This support will be added.
* Currently, there is no support in this SPI for registering functions which take in a variable number of arguments.  This support will be added.
[...]

PR #23358, opened 2024-08-01
Add a native function namespace manager

pdabre12:
## Description

Adds a native function namespace manager

## Motivation and Context
To help resolve : https://github.com/prestodb/presto/issues/23000

## Test Plan
Unit and end-to-end tests. More comprehensive end-to-end tests will be written in the future.

## Contributor checklist

- [ ] CI passed.

```
== RELEASE NOTES ==

General C++ changes
* Add a native function namespace manager :pr:`23358`
```

Depends on : https://github.com/prestodb/presto/pull/23671


steveburnett (dismissed):
LGTM! (docs)

Pull branch, new local doc build, the doc looks good. Thanks!

steveburnett (dismissed):
LGTM! (docs)

Pull updated branch, new local doc build, doc looks good. Thanks!

presto-common/src/main/java/com/facebook/presto/common/type/TypeSignature.java
```diff
@@ -146,6 +146,25 @@ public List<TypeSignature> getTypeParametersAsTypeSignatures()
         return result;
     }
 
+    public List<TypeSignature> getTypeOrNamedTypeParametersAsTypeSignatures()
+    {
+        List<TypeSignature> result = new ArrayList<>();
+        for (TypeSignatureParameter parameter : parameters) {
+            if ((parameter.getKind() != ParameterKind.TYPE) &&
+                    (parameter.getKind() != ParameterKind.NAMED_TYPE)) {
```
tdcmeehan, line 154:
```suggestion
            if (!(parameter.getKind() == ParameterKind.TYPE &&
                    (parameter.getKind() == ParameterKind.NAMED_TYPE)) {
```

presto-function-namespace-managers-common/src/main/java/com/facebook/presto/functionNamespace/AbstractSqlInvokedFunctionNamespaceManager.java
```diff
@@ -245,7 +246,7 @@ public CompletableFuture<SqlFunctionResult> executeFunction(String source, Funct
                 typeManager.getType(functionMetadata.getReturnType()));
     }
 
-    private static PrestoException convertToPrestoException(UncheckedExecutionException exception, String failureMessage)
+    protected static PrestoException convertToPrestoException(UncheckedExecutionException exception, String failureMessage)
```
tdcmeehan, line 249: Just reimplement where needed, don't change the visibility.

presto-function-namespace-managers-common/src/main/java/com/facebook/presto/functionNamespace/AbstractSqlInvokedFunctionNamespaceManager.java
```diff
@@ -362,10 +363,13 @@ protected AggregationFunctionImplementation sqlInvokedFunctionToAggregationImple
                         "Need aggregationMetadata to get aggregation function implementation");
 
                 AggregationFunctionMetadata aggregationMetadata = function.getAggregationMetadata().get();
+                List<Type> parameters = function.getSignature().getArgumentTypes().stream().map(
```
tdcmeehan, line 366: Remind me why we need this please?

presto-function-namespace-managers-common/pom.xml
```diff
@@ -0,0 +1,104 @@
...
+        <artifactId>presto-root</artifactId>
+        <groupId>com.facebook.presto</groupId>
+        <version>0.290-SNAPSHOT</version>
+    </parent>
+
+    <properties>
+        <air.main.basedir>${project.parent.basedir}</air.main.basedir>
+    </properties>
+
+    <artifactId>presto-function-namespace-managers-common</artifactId>
```
tdcmeehan, line 15: Why do we need the Thrift implementation of function namespace managers in common?  Can this module be slimmed down?

presto-grpc-api/pom.xml
```diff
@@ -80,7 +80,8 @@
 
         <dependency>
             <groupId>com.facebook.presto</groupId>
-            <artifactId>presto-function-namespace-managers</artifactId>
+            <artifactId>presto-function-namespace-managers-common</artifactId>
+            <version>0.290-SNAPSHOT</version>
```
tdcmeehan, line 84:
```suggestion
            <version>${project.version}</version>
```

presto-jdbc/pom.xml
```diff
@@ -38,6 +38,13 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>com.facebook.presto</groupId>
+            <artifactId>presto-function-namespace-managers-common</artifactId>
+            <version>0.290-SNAPSHOT</version>
```
tdcmeehan, line 44:
```suggestion
            <version>${project.version}</version>
```

presto-main/pom.xml
```diff
@@ -469,6 +469,13 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>com.facebook.presto</groupId>
+            <artifactId>presto-function-namespace-managers-common</artifactId>
+            <version>0.290-SNAPSHOT</version>
```
tdcmeehan, line 475:
```suggestion
            <version>${project.version}</version>
```

presto-main/src/main/java/com/facebook/presto/metadata/FunctionAndTypeManager.java
```diff
@@ -256,14 +276,14 @@ public CacheStatsMBean getFunctionResolutionCacheStats()
     public void loadFunctionNamespaceManager(
             String functionNamespaceManagerName,
             String catalogName,
-            Map<String, String> properties)
+            Map<String, String> properties,
+            Optional<NodeManager> nodeManager)
```
tdcmeehan, line 280: Why is it optional?

tdcmeehan (changes requested):
Just some initial comments

tdcmeehan (commented):
`Refactored qualifyObjectName from static to instance method` ✅

tdcmeehan (commented):
`[native] Fix bug when handling multiple params and no params aggregat…` ❓ 

Can you help me understand when this bug manifests?  Is there any way to add a unit test for it?

tdcmeehan (changes requested):
`[native] Add additional function metadata to function signatures endp…` ❌  

Looks good, but there is a `workspace.xml` file that needs to be removed.

tdcmeehan (commented):
`Add a new module presto-function-namespace-managers-common` ❌ 

Previous feedback remains, let's keep `common` as a very slim module.

presto-function-namespace-managers-common/src/main/java/com/facebook/presto/functionNamespace/AbstractSqlInvokedFunctionNamespaceManager.java
pdabre12: Its needed because we changed the `SqlInvokedAggregationFunctionImplementation` constructor here: https://github.com/prestodb/presto/blob/88e1152a75bf1d4edd555c197ab0c27e756d574e/presto-spi/src/main/java/com/facebook/presto/spi/function/SqlInvokedAggregationFunctionImplementation.java

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/functionNamespace/NativeFunctionDefinitionProvider.java
```diff
@@ -0,0 +1,50 @@
...
+    {
+        this.nativeFunctionSignatureMapJsonCodec = nativeFunctionSignatureMapJsonCodec;
+    }
+
+    @Override
+    public UdfFunctionSignatureMap getUdfDefinition(NodeManager nodeManager)
+            throws IllegalStateException
+    {
+        try {
+            throw new UnsupportedOperationException();
```
tdcmeehan: Can we just throw UOE?

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/functionNamespace/NativeFunctionHandle.java
```diff
@@ -0,0 +1,109 @@
...
+        return Objects.hash(signature);
+    }
+
+    @Override
+    public String toString()
+    {
+        return signature.toString();
+    }
+
+    private static void checkArgument(boolean condition, String message, Object... args)
```
tdcmeehan, line 93: We can use the Guava utility.

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/functionNamespace/NativeFunctionNamespaceManager.java
```diff
@@ -0,0 +1,484 @@
...
+        if (nativeFunctionSignatureMap == null || nativeFunctionSignatureMap.isEmpty()) {
+            return latestFunctionsTemp;
+        }
+        populateNameSpaceManager(nativeFunctionSignatureMap);
+        latestFunctionsTemp.putAll(latestFunctions);
+        latestFunctions.clear();
+        return latestFunctionsTemp;
+    }
+
+    private void populateNameSpaceManager(UdfFunctionSignatureMap udfFunctionSignatureMap)
```
tdcmeehan, line 138: `populateNamespaceManager`

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/functionNamespace/NativeFunctionNamespaceManager.java
```diff
@@ -0,0 +1,484 @@
...
+
+    private SqlFunctionSupplier doGetSpecializedFunctionKey(Signature signature)
+    {
+        return functionMetadataManager.getSpecializedFunctionKey(signature);
+    }
+
+    private Map<SqlFunctionId, SqlInvokedFunction> bootstrapNamespace()
+    {
+        UdfFunctionSignatureMap nativeFunctionSignatureMap = functionDefinitionProvider.getUdfDefinition(nodeManager);
+        Map<SqlFunctionId, SqlInvokedFunction> latestFunctionsTemp = new ConcurrentHashMap<>();
```
tdcmeehan, line 128: Please just use an immutable map, there is no need for concurrency here.

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/functionNamespace/NativeFunctionNamespaceManager.java
```diff
@@ -0,0 +1,484 @@
...
+        if (nativeFunctionSignatureMap == null || nativeFunctionSignatureMap.isEmpty()) {
+            return latestFunctionsTemp;
```
tdcmeehan, line 130: `return ImmutableMap.of();`

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/functionNamespace/NativeFunctionNamespaceManager.java
```diff
@@ -0,0 +1,484 @@
...
+                .build(CacheLoader.from(this::doGetSpecializedFunctionKey));
+        this.types = functionMetadataManager.getTypes();
+    }
+
+    private SqlFunctionSupplier doGetSpecializedFunctionKey(Signature signature)
+    {
+        return functionMetadataManager.getSpecializedFunctionKey(signature);
+    }
+
+    private Map<SqlFunctionId, SqlInvokedFunction> bootstrapNamespace()
```
tdcmeehan, line 125: Here is how I would suggest the concurrency control to go.

Wherever we mutate `latestFunctions`, synchronize that.  If it's lightweight and in-memory, just synchronize the method.  This applies to `createFunction` and bootstrapNamespace`.

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/functionNamespace/NativeFunctionNamespaceManager.java
```diff
@@ -0,0 +1,484 @@
...
+            TypeSignature parameterTypeSignature;
+            // If it's a named type signatures only in the case of row signature types.
+            if (isNamedTypeSignature) {
+                parameterTypeSignature = parameter.getNamedTypeSignature().getTypeSignature();
+            }
+            else {
+                parameterTypeSignature = parameter.getTypeSignature();
+            }
+
+            if (parameterTypeSignature.getParameters().isEmpty()) {
```
tdcmeehan, line 400: Let's very clearly call out the hack, where it is needed and why.  Let's prefix the comment with `// HACK:` so it's obvious.

tdcmeehan (commented):
`[native] Add e2e native function validation tests with sidecar` ✅

tdcmeehan (commented):
`Add OpenAPI documentation for /v1/functions` ✅

presto-common/src/main/java/com/facebook/presto/common/type/TypeSignature.java
rschlussel: @tdcmeehan I thought the original is easier to read. but your suggestion should have || instead of &&.

presto-common/src/main/java/com/facebook/presto/common/type/TypeSignature.java
```diff
@@ -146,6 +146,25 @@ public List<TypeSignature> getTypeParametersAsTypeSignatures()
         return result;
     }
 
+    public List<TypeSignature> getTypeOrNamedTypeParametersAsTypeSignatures()
```
rschlussel, line 149: Can you instead modify getTypeParametersAsTypeSignatures. I don't think there's any reason NamedTypeParameters are excluded there.

presto-common/src/main/java/com/facebook/presto/common/type/TypeSignatureUtils.java
```diff
@@ -0,0 +1,158 @@
...
+            //              Type signature base = T
+            //              parameters = empty
+            //            ArgumentType :
+            //              Type signature base = double
+            //              parameters = empty
+            //          return typeSignatureMap = {"T": "double"}
+
+            // If parameter params are of type long e.g decimal(15, 2) or
+            // of type varchar e.g decimal(i4, i5), we don't need to recursively call the function on its params
+            if (parameter.getParameters().isEmpty() || areParametersLongOrVariable(parameter.getParameters())) {
```
rschlussel, line 75: what about parameters that are none of these kinds (long_enum, varchar_enum, distinct_type)?

also, since you've been looking at these different parameter kinds, would be great to add some javadoc to the enum to explain what the different parameter kinds mean.

presto-common/src/main/java/com/facebook/presto/common/type/TypeSignatureUtils.java
```diff
@@ -0,0 +1,158 @@
...
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+import static java.util.Collections.unmodifiableList;
+
+public final class TypeSignatureUtils
+{
+    private TypeSignatureUtils() {}
+
+    public static TypeSignature resolveIntermediateType(TypeSignature actualTypeSignature, List<TypeSignature> actualTypeSignatureParameters, List<TypeSignature> expectedTypeSignatureParameters)
```
rschlussel, line 29: why do we need special logic to handle aggregation function intermediate types?

presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp
```diff
@@ -845,15 +845,17 @@ void VeloxQueryPlanConverterBase::toAggregations(
           auto pos = functionId.find(";", start + 1);
           if (pos == std::string::npos) {
             auto argumentType = functionId.substr(start + 1);
-            aggregate.rawInputTypes.push_back(
-                stringToType(argumentType, typeParser_));
+            if (!argumentType.empty()) {
```
rschlussel, line 848: Can you explain the bug? Can you share an example of how this manifests. Also, add a test.  If this is a bug fix, also consider pulling this out into a separate PR to merge earlier.

presto-common/src/main/java/com/facebook/presto/common/type/TypeSignature.java
tdcmeehan: Edited.  I read this as "if the parameter kind is not a TYPE OR a NAMED TYPE", which to me is just a more natural way of describing the outcome than "if the parameter is not a TYPE AND not a NAMED TYPE".

But I just realized we could just do this, which seems better:

```java
    public List<TypeSignature> getTypeOrNamedTypeParametersAsTypeSignatures()
    {
        {
            List<TypeSignature> result = new ArrayList<>();
            for (TypeSignatureParameter parameter : parameters) {
                switch (parameter.getKind()) {
                    case TYPE:
                        result.add(parameter.getTypeSignature());
                        break;
                    case NAMED_TYPE:
                        result.add(parameter.getNamedTypeSignature().getTypeSignature());
                        break;
                    default:
                        throw new IllegalStateException(
                                format("Expected all parameters to be of kind TYPE or NAMED_TYPE but [%s] kind was found for parameter: [%s]",
                                        parameter.getKind(), parameter));
                }
            }
            return result;
        }
    }

```

presto-common/src/main/java/com/facebook/presto/common/type/TypeSignatureUtils.java
pdabre12: We only need special logic to handle generic aggregate functions. 
To resolve the intermediate type for generic aggregate functions, we rely on utility functions from TypeSignatureUtils. Specifically, we use the parameter types passed down in the function signatures to determine and insert the appropriate intermediateType.

- Why do we need this?
This is required because aggregate function metadata for generic aggregate functions contain generic intermediate args i.e `K,V` and we need a way to resolve these args.

Example:
`max` function:
For the max function, here is the relevant metadata:

```
{
  "aggregateMetadata": {
    "intermediateType": "t",
    "isOrderSensitive": false
  },
  "docString": "native.default.max",
  "functionKind": "AGGREGATE",
  "outputType": "t",
  "paramTypes": [
    "t"
  ],
  "routineCharacteristics": {
    "determinism": "DETERMINISTIC",
    "language": "CPP",
    "nullCallClause": "CALLED_ON_NULL_INPUT"
  },
  "schema": "default",
  "typeVariableConstraints": [
    {
      "comparableRequired": true,
      "name": "t",
      "nonDecimalNumericRequired": false,
      "orderableRequired": true,
      "variadicBound": ""
    }
  ],
  "variableArity": false
}
```

In this case:
- The functionHandle passed is an instance of NativeFunctionHandle.
- Let's suppose the signature that gets passed down is `native.default.max(bigint):bigint`, we use the passed down param type `bigint` to resolve the intermediate type `t` as `bigint`.

`max_by` function: 

```
{
  "aggregateMetadata": {
    "intermediateType": "row(bigint,array(boolean),array(v))",
    "isOrderSensitive": true
  },
  "docString": "native.default.max_by",
  "functionKind": "AGGREGATE",
  "outputType": "array(v)",
  "paramTypes": [
    "v",
    "boolean",
    "bigint"
  ],
  "routineCharacteristics": {
    "determinism": "DETERMINISTIC",
    "language": "CPP",
    "nullCallClause": "CALLED_ON_NULL_INPUT"
  },
  "schema": "default",
  "typeVariableConstraints": [
    {
      "comparableRequired": false,
      "name": "v",
      "nonDecimalNumericRequired": false,
      "orderableRequired": false,
      "variadicBound": ""
    }
  ],
  "variableArity": false
}
```

In this case:

- The functionHandle passed is an instance of NativeFunctionHandle.
- Let's suppose the signature that gets passed down is `native.default.max_by(integer, boolean, bigint):array(integer)`, we use the passed down param types `integer ` to resolve the intermediate type `"intermediateType": "row(bigint,array(boolean),array(v))` as `"intermediateType": "row(bigint,array(boolean),array(integer))`

presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp
pdabre12: Opened a PR: https://github.com/prestodb/presto/pull/24059

presto-main/src/main/java/com/facebook/presto/metadata/FunctionAndTypeManager.java
pdabre12: Because `nodeManager` is only required while loading the `NativeFunctionNamespaceManager` and not the other function namespace managers, hence I thought making it optional is the right way.

tdcmeehan: Can we supply it always and just have other namespace managers not use it?

pdabre12: Also in some of the places where `loadFunctionNamespaceManager` is called I do not have access to an instance of the `NodeManager`.

presto-common/src/main/java/com/facebook/presto/common/type/TypeSignatureUtils.java
```diff
@@ -0,0 +1,158 @@
...
+    {
+        Map<TypeSignature, TypeSignature> typeSignatureMap = getTypeSignatureMap(actualTypeSignatureParameters, expectedTypeSignatureParameters);
+        return resolveTypeSignatures(actualTypeSignature, typeSignatureMap).getTypeSignature();
+    }
+
+    public static Map<TypeSignature, TypeSignature> getTypeSignatureMap(List<TypeSignature> parameters, List<TypeSignature> argumentTypes)
```
tdcmeehan, line 35: I believe this is only used in the tests and in the above method.  Can we make it private?  I think there's sufficient unit test coverage through `resolveIntermediateType`.

tdcmeehan (commented):
`Add utility functions to resolve intermediate type in generic aggrega…`

tdcmeehan (commented):
`Add a new module presto-function-namespace-managers-common` ✅ 

Much better, thanks!

presto-main/src/main/java/com/facebook/presto/server/NativeFunctionNamespaceManagerProvider.java
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+package com.facebook.presto.server;
+
+import com.facebook.presto.spi.NodeManager;
+import com.google.inject.Inject;
+
+import static java.util.Objects.requireNonNull;
+
+public class NativeFunctionNamespaceManagerProvider
```
tdcmeehan, line 21: Why do we need this class?  Seems unused?

presto-main/src/main/java/com/facebook/presto/server/PluginManagerConfig.java
```diff
@@ -114,4 +115,17 @@ public PluginManagerConfig setDisabledConnectors(String disabledConnectors)
...
     }
+
+    @NotNull
+    public String getDefaultNamespacePrefix()
+    {
+        return defaultNamespacePrefix;
+    }
+
+    @Config("presto.default-namespace")
+    public PluginManagerConfig setDefaultNamespacePrefix(String defaultNamespacePrefix)
```
tdcmeehan, line 126: Rather than match a default namespace with a namespace that is configured in the function namespace manager, just make it a boolean flag property on the function namespace manager's config that it is the default namespace.  If none have that specified, use `presto.default`, if more than one have that specified, then throw.  

The place to add this property would be in `StaticFunctionNamespaceStore`.

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/functionNamespace/NativeFunctionNamespaceManager.java
```diff
@@ -0,0 +1,501 @@
...
+            throw new PrestoException(GENERIC_USER_ERROR, format("Function '%s' is missing from cache", functionId.getId()));
+        }
+
+        aggregationImplementationByHandle.put(
+                sqlFunctionHandle,
+                sqlInvokedFunctionToAggregationImplementation(memoizedFunctionsSupplier.get().get(functionId), typeManager));
+        return aggregationImplementationByHandle.get(sqlFunctionHandle);
+    }
+
+    private static SqlInvokedFunction copyFunction(SqlInvokedFunction function)
```
tdcmeehan, line 206: `SqlInvokedFunction` is immutable, why do we need to do this copy?

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/functionNamespace/NativeFunctionNamespaceManager.java
tdcmeehan: It seems like this is several methods long, so let's add an end comment that says when the hack is over.

presto-spi/src/main/java/com/facebook/presto/spi/function/SqlInvokedAggregationFunctionImplementation.java
```diff
@@ -59,4 +73,28 @@ public boolean isOrderSensitive()
...
+    @Override
+    public List<Type> getParameterTypes()
+    {
+        return parameterTypes;
+    }
+
+    @Override
+    public AggregationMetadata getAggregationMetadata()
+    {
+        return null;
```
tdcmeehan, line 86: I believe it should be safe or even preferable to throw here, since they should be called.

tdcmeehan (changes requested):
Looking pretty good, some more comments.

rschlussel (commented):
I'm still reviewing "[native] Introduce native function namespace SPI"

presto-common/src/main/java/com/facebook/presto/common/type/TypeSignatureUtils.java
```diff
@@ -0,0 +1,158 @@
...
+    private static NamedTypeSignature resolveTypeSignatures(TypeSignature typeSignature, Map<TypeSignature, TypeSignature> typeSignatureMap)
+    {
+        if (typeSignatureMap.containsKey(typeSignature)) {
+            TypeSignature resolvedTypeSignature = typeSignatureMap.get(typeSignature);
+            return new NamedTypeSignature(Optional.empty(), resolvedTypeSignature);
+        }
+        List<NamedTypeSignature> namedTypeSignatures = new ArrayList<>();
+        List<TypeSignature> typeSignatures = new ArrayList<>();
+        for (TypeSignature typeParameterSignature : typeSignature.getTypeOrNamedTypeParametersAsTypeSignatures()) {
+            // if base is "row" signature needs to be of type NamedTypeSignature because of the check in HiveUtil.java #L1237
```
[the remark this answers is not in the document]
pdabre12: It is not dependent on something in the hive connector, it's just that a `row` typeSignature expects all its params to be of `NamedTypeSignature`, I can change the wording if needed.

presto-docs/src/main/sphinx/presto_cpp/properties.rst
```diff
@@ -95,6 +95,14 @@ alphabetical order.
   This property provides function signatures for built-in aggregation 
   functions which are compatible with Velox.
 
+``presto.default-namespace``
```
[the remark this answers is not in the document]
pdabre12: So the implementation for this has changed a little bit but the idea remains the same.
The idea is that we have functions residing in two different namespaces (`presto.default` and `native.default`), we allow the switching the namespace that is currently acting as default. 
Adding a config flag gives users the ability to specify what the default namespace is maybe `native.default` or `native-functions.default`, it's totally upto the user to decide that.

That being said, in the latest changes (will push soon),  we have moved this flag to the `function-namespace/<catalog-name>.properties`. Users can mention in the `function-namespace/<catalog-name>.properties` whether they want the function namespace as the currentDefaultNamepace.
For example:
If the user sets `default.namespace=true` in `function-namespace/native.properties`, the currentDefaultNamepace will be `native.default`, we just pick the `catalog name (function-namespace-manager- file name)`, the schema name is constant as `default`.

Let me know if that sounds okay.

presto-function-namespace-managers/src/main/java/com/facebook/presto/functionNamespace/json/JsonBasedUdfFunctionMetadata.java
```diff
@@ -30,9 +31,6 @@
 import static com.google.common.collect.ImmutableList.toImmutableList;
 import static java.util.Objects.requireNonNull;
 
-/**
- * The function metadata provided by the Json file to the {@link JsonFileBasedFunctionNamespaceManager}.
```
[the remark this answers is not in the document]
pdabre12: There was a new module introduced, we moved the file there.

presto-native-execution/presto_cpp/main/types/FunctionMetadata.cpp
```diff
@@ -199,8 +224,22 @@ json buildAggregateMetadata(
       getWindowFunctionSignatures(name).has_value(),
       "Aggregate function {} not registered as a window function",
       name);
+
```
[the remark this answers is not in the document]
pdabre12: I am not sure about the sideeffect for the worker. 
It doesn't seem to have a problem in processing aggregate/ window functions. 
Locally, I extended all the tests cases from `AbstractTestNativeWindowQueries` and all the test cases are passing.

CC : @aditi-pandit

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/AddExchanges.java
```diff
@@ -771,8 +773,34 @@ public PlanWithProperties visitTableFinish(TableFinishNode node, PreferredProper
             PlanNode child = planChild(node, PreferredProperties.any()).getNode();
 
             ExchangeNode gather;
+
+            // If coordinatorSidecar is enabled, we do not have the necessary metadata to
+            // build an accumulator factory to perform the aggregations, hence we introduce an exchange and
+            // perform the final aggregation in the native execution engine.
+            if (coordinatorSidecarEnabled && (child instanceof TableWriterNode || child instanceof UnionNode)) {
```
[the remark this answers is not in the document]
pdabre12: Okay, do you have any suggestions on how I should handle this?

presto-spi/src/main/java/com/facebook/presto/spi/function/SqlInvokedFunction.java
```diff
@@ -47,7 +47,7 @@ public class SqlInvokedFunction
     private final String description;
     private final RoutineCharacteristics routineCharacteristics;
     private final String body;
-
+    private final boolean variableArity;
```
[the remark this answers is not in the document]
pdabre12: I borrowed the naming from : https://github.com/prestodb/presto/blob/1327a3e6c2025659760b8f94142dec0edc663485/presto-spi/src/main/java/com/facebook/presto/spi/function/Signature.java#L44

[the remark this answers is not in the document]
rschlussel: oh i think i maybe misunderstood what this means. is this a boolean to say the number of arguments to the function is variable? when i read the name i though it was about the number of arguments for a particular variable.

presto-native-execution/presto_cpp/main/types/FunctionMetadata.cpp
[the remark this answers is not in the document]
aditi-pandit: I'm unsure of the context here... but all aggregate functions can be used as window functions in SQL queries. With this change seems like you are not building function metadata for aggregate functions as window functions from native side-car. So only pure window functions like rank, nth_value etc will appear as window functions from the side-car. Velox expects registration of each aggregate function as a window function in its registry. Does that not hold on the Presto Java side ?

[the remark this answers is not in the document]
rschlussel: so for presto, i think any aggregate functions can be used as a window function, and we don't need to register it separately. Why doesn't this change break things for prestissimo window functions if we aren't registering them anymore.

presto-spi/src/main/java/com/facebook/presto/spi/function/SqlInvokedFunction.java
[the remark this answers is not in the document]
pdabre12: Yes, its a boolean flag to indicate the number of arguments to the function is variable.

presto-docs/src/main/sphinx/presto_cpp/properties.rst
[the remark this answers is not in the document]
pdabre12: My bad, I thought the question was for the `presto.default-namespace` config on the coordinator side.
For more context on the worker side config: https://github.com/prestodb/presto/issues/23385

presto-native-execution/presto_cpp/main/common/Configs.cpp
```diff
@@ -748,6 +749,10 @@ bool SystemConfig::enableRuntimeMetricsCollection() const {
   return optionalProperty<bool>(kEnableRuntimeMetricsCollection).value();
 }
 
+std::string SystemConfig::prestoDefaultNamespacePrefix() const {
+  return optionalProperty(kPrestoDefaultNamespacePrefix).value().append(".");
```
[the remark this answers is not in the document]
pdabre12: It is required here to register the functions: https://github.com/prestodb/presto/blob/046c0212254f4a9793d533893a245155c7466458/presto-native-execution/presto_cpp/main/PrestoServer.cpp#L1247

presto-native-execution/presto_cpp/main/types/PrestoToVeloxExpr.h
```diff
@@ -14,10 +14,19 @@
 #pragma once
 
 #include <stdexcept>
+#include "presto_cpp/main/common/Configs.h"
 #include "presto_cpp/main/types/TypeParser.h"
 #include "presto_cpp/presto_protocol/core/presto_protocol_core.h"
 #include "velox/core/Expressions.h"
 
+namespace {
+inline std::string formatDefaultNamespacePrefix(
```
aditi-pandit, line 23: Its bit odd this function has name formatDefaultNamespacePrefix and then takes the default namespace prefix as a parameter. Maybe you should call this addDefaultNamespace and pick up the default namespace in the function without passing it as a parameter everywhere.

presto-main/src/main/java/com/facebook/presto/metadata/FunctionAndTypeManager.java
pdabre12: Resolving this conversation, changed `loadFunctionNamespaceManager` to take in NodeManager as a parameter.

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/NativeSidecarPlugin.java
```diff
@@ -26,4 +28,10 @@ public Iterable<WorkerSessionPropertyProviderFactory> getWorkerSessionPropertyPr
     {
         return ImmutableList.of(new NativeSystemSessionPropertyProviderFactory());
     }
+
+    @Override
+    public Iterable<FunctionNamespaceManagerFactory> getFunctionNamespaceManagerFactories()
```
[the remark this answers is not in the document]
pdabre12: I wasn't able to think of a good way to achieve this, is it okay if I create an issue and handle this in a follow-up PR?
@tdcmeehan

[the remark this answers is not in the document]
tdcmeehan: Another complication is that certain aspects of functionality will be dependent on each other.  But to answer the question, I think we can use per-feature flags.  For example, to disable this aspect of the plugin, simply do not enable it to be the default namespace.

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/functionNamespace/NativeFunctionNamespaceManager.java
```diff
@@ -0,0 +1,486 @@
...
+    {
+        SqlFunctionSupplier functionKey;
+        try {
+            functionKey = specializedFunctionKeyCache.getUnchecked(signature);
+        }
+        catch (UncheckedExecutionException e) {
+            throw convertToPrestoException(e, format("Error getting FunctionMetadata for signature: %s", signature));
+        }
+        return functionKey.getFunction();
+    }
```
tdcmeehan, line 453:
```suggestion
    private SqlFunction getSqlFunctionFromSignature(Signature signature)
    {
        try {
            return specializedFunctionKeyCache.getUnchecked(signature).getFunction();
        }
        catch (UncheckedExecutionException e) {
            throw convertToPrestoException(e, format("Error getting FunctionMetadata for signature: %s", signature));
        }
    }
```

presto-spi/src/main/java/com/facebook/presto/spi/function/FunctionMetadataManager.java
```diff
@@ -13,7 +13,17 @@
...
+
 public interface FunctionMetadataManager
 {
     FunctionMetadata getFunctionMetadata(FunctionHandle functionHandle);
+
+    /**
+     * Extends the ability to retrieve specializedFunctionKeys using existing code from BuiltInTypeAndFunctionNamespaceManager.
+     */
+    SqlFunctionSupplier getSpecializedFunctionKey(Signature signature);
+    List<Type> getTypes();
```
tdcmeehan, line 28: Hmm why do we need this?

presto-spi/src/main/java/com/facebook/presto/spi/function/FunctionNamespaceManagerContext.java
```diff
@@ -14,20 +14,38 @@
...
+    private final NodeManager nodeManager;
+    private final Optional<FunctionMetadataManager> functionMetadataManager;
 
-    public FunctionNamespaceManagerContext(TypeManager typeManager)
+    public FunctionNamespaceManagerContext(TypeManager typeManager, NodeManager nodeManager,
+            Optional<FunctionMetadataManager> functionMetadataManager)
     {
         this.typeManager = requireNonNull(typeManager, "typeManager is null");
+        this.nodeManager = requireNonNull(nodeManager, "nodeManager is null");
+        this.functionMetadataManager = requireNonNull(functionMetadataManager, "functionMetadataManager is null");
```
tdcmeehan, line 34: The only creator of this seems to be in `FunctionAndTypeManager`, and it's present there.  Can we make it always present?

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/functionNamespace/NativeFunctionNamespaceManager.java
```diff
@@ -0,0 +1,486 @@
...
+                .collect(onlyElement());
+    }
+
+    @Override
+    public synchronized void createFunction(SqlInvokedFunction function, boolean replace)
+    {
+        checkFunctionLanguageSupported(function);
+        SqlFunctionId functionId = function.getFunctionId();
+        if (!replace && functions.containsKey(function.getFunctionId())) {
+            throw new PrestoException(GENERIC_USER_ERROR, format("Function '%s' already exists", functionId.getId()));
```
tdcmeehan, line 282: I would create a new error code type in StandardErrorCodes.  `DUPLICATE_FUNCTION_ERROR`.

tdcmeehan (commented):
A few nits

presto-spi/src/main/java/com/facebook/presto/spi/function/FunctionMetadataManager.java
pdabre12: Not needed anymore, removed.

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/NativeSidecarPlugin.java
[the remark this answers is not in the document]
pdabre12: That would work for the `FunctionNamespaceManager` SPI, but disabling the `SessionPropertyManager` SPI will not be possible. 
The per-feature flags idea would work, provided the functionalities aren't dependent on each other.
For eg: having a `disable-native-session-property-provider` flag,  if disabled do not load the SessionPropertyManager SPI.

presto-native-execution/presto_cpp/main/types/PrestoToVeloxExpr.h
pdabre12: Yes, calling this `addDefaultNamespace` makes sense. 
I am confused about the picking up of the default namespace from the functions part, the functions passed down to the function won't have any namespace attached.

For eg: 
`presto.default.$operator.$add` should map to` {namespace-prefix}.add` , in this case `native.default.add`
https://github.com/prestodb/presto/blob/40161c0a0b68ee0842ad25a66d849d6425e2e883/presto-native-execution/presto_cpp/main/types/PrestoToVeloxExpr.cpp#L41 

The call to the formatDefaultNamespacePrefix() function will take in the default namespace prefix and the function name, add the default namespace prefix to the function name and return the string.

presto-native-execution/presto_cpp/main/types/PrestoToVeloxExpr.cpp
```diff
@@ -207,6 +231,8 @@ std::optional<TypedExprPtr> tryConvertCast(
     const std::string& returnType,
     const std::vector<TypedExprPtr>& args,
     const TypeParser* typeParser) {
+  static const std::string prestoDefaultNamespacePrefix =
+      SystemConfig::instance()->prestoDefaultNamespacePrefix();
   static const char* kCast = "presto.default.$operator$cast";
```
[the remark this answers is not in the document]
pdabre12: The casts will be passed down by the coordinator with the `presto.default` namespace prefix always, and we do not need to map it to a velox function. 

We only need to call the `formatDefaultNamespacePrefix` function if we need to map a passed down function by the Java coordinator to a velox function.

presto-main/src/main/java/com/facebook/presto/metadata/FunctionAndTypeManager.java
```diff
@@ -363,9 +386,19 @@ public List<SqlFunction> listFunctions(Session session, Optional<String> likePat
         ImmutableList.Builder<SqlFunction> functions = new ImmutableList.Builder<>();
         if (!isListBuiltInFunctionsOnly(session)) {
             functions.addAll(SessionFunctionUtils.listFunctions(session.getSessionFunctions()));
-            functions.addAll(functionNamespaceManagers.values().stream()
-                    .flatMap(manager -> manager.listFunctions(likePattern, escape).stream())
-                    .collect(toImmutableList()));
+
+            // If coordinator sidecar is enabled, filter on the current default namespace
+            if (coordinatorSidecarEnabled) {
```
tdcmeehan, line 391: In case the coordinator sidecar isn't enabled, wouldn't this be the same thing?  Can we remove this branch?

presto-main/src/main/java/com/facebook/presto/metadata/StaticFunctionNamespaceStore.java
```diff
@@ -78,7 +84,14 @@ private void loadFunctionNamespaceManager(String catalogName, Map<String, String
         String functionNamespaceManagerName = properties.remove(FUNCTION_NAMESPACE_MANAGER_NAME);
         checkState(!isNullOrEmpty(functionNamespaceManagerName), "%s property must be present", FUNCTION_NAMESPACE_MANAGER_NAME);
 
-        functionAndTypeManager.loadFunctionNamespaceManager(functionNamespaceManagerName, catalogName, properties);
+        if (Boolean.parseBoolean(properties.get(DEFAULT_NAMESPACE))) {
```
tdcmeehan, line 87:
```suggestion
        if (parseBoolean(properties.get(DEFAULT_NAMESPACE))) {
```

tdcmeehan: I think you'll want to remove this from the properties before sending it to the plugin, in the same way we remove the name.  See line 84.

presto-main/src/main/java/com/facebook/presto/sql/planner/PlanOptimizers.java
```diff
@@ -893,13 +893,13 @@ public PlanOptimizers(
         // Optimizers above this do not need to care about aggregations with the type other than SINGLE
         // This optimizer must be run after all exchange-related optimizers
         builder.add(new IterativeOptimizer(
-                metadata,
```
tdcmeehan, line 896: Revert whitespace only change

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/AddExchanges.java
```diff
@@ -854,6 +854,7 @@ public PlanWithProperties visitTableFinish(TableFinishNode node, PreferredProper
             PlanNode child = planChild(node, PreferredProperties.any()).getNode();
 
             ExchangeNode gather;
+
```
tdcmeehan, line 857: Revert whitespace only change

presto-main/src/main/java/com/facebook/presto/sql/relational/SqlToRowExpressionTranslator.java
```diff
@@ -908,6 +910,9 @@ protected RowExpression visitLikePredicate(LikePredicate node, Context context)
 
             if (node.getEscape().isPresent()) {
                 RowExpression escape = process(node.getEscape().get(), context);
+                if (coordinatorSidecarEnabled) {
```
tdcmeehan, line 913: Instead of switching on `coordinatorSidecarEnabled`, can we introduce a method `functionResolution.supportsLikePattern`, which just checks if the lake pattern function is present.  If so, use it, otherwise, use a varchar version.

presto-main/src/test/java/com/facebook/presto/sql/TestExpressionInterpreter.java
```diff
@@ -174,15 +175,31 @@ public class TestExpressionInterpreter
...
+    private static final TestingPrestoServer server;
+
+    static {
+        try {
+            server = new TestingPrestoServer();
+        }
+        catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
```
tdcmeehan, line 187: This is a bit heavyweight, can we refactor to avoid using a `TestingPrestoServer`?

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/functionNamespace/NativeFunctionNamespaceManager.java
```diff
@@ -0,0 +1,473 @@
...
+        }
+        return functionHandle;
+    }
+
+    /*
+        Todo: Improve the handling of parameter type differentiation in native execution.
+        // HACK: Currently, we lack support for correctly identifying the parameterKind, specifically between TYPE and VARIABLE,
+        in native execution. The following utility functions help bridge this gap by parsing the type signature and verifying whether its base
+        and parameters are of a supported type. The valid types list are non - parametric types that Presto supports.
+     */
```
tdcmeehan, line 338:
```suggestion
        // TODO: Improve the handling of parameter type differentiation in native execution.
        // HACK: Currently, we lack support for correctly identifying the parameterKind, specifically between TYPE and VARIABLE,
        // in native execution. The following utility functions help bridge this gap by parsing the type signature and verifying whether its base
        // and parameters are of a supported type. The valid types list are non - parametric types that Presto supports.
```

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/AddExchanges.java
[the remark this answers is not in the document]
pdabre12: Removed this.

presto-common/src/main/java/com/facebook/presto/common/type/TypeSignatureUtils.java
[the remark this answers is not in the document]
pdabre12: It is a requirement enforced by the type system itself

pdabre12:
@rschlussel Can you please have another look?  Thanks.

presto-spi/src/main/java/com/facebook/presto/spi/function/SqlInvokedAggregationFunctionImplementation.java
```diff
@@ -59,4 +73,28 @@ public boolean isOrderSensitive()
...
+    @Override
+    public Class<? extends Accumulator> getAccumulatorClass()
+    {
+        return null;
+    }
+
+    @Override
+    public Class<? extends GroupedAccumulator> getGroupedAccumulatorClass()
+    {
+        return null;
```
tdcmeehan, line 98: Can we have these throw since they should never be called?

tdcmeehan (changes requested):
Only one minor thing.

@rschlussel would you like to have a look as well?

presto-main/src/main/java/com/facebook/presto/metadata/FunctionAndTypeManager.java
```diff
@@ -589,26 +607,33 @@ public boolean nullIfSpecialFormEnabled()
...
      * @throws PrestoException if function could not be found
      */
-    public FunctionHandle lookupFunction(String name, List<TypeSignatureProvider> parameterTypes)
+    public FunctionHandle lookupFunction(QualifiedObjectName functionName, List<TypeSignatureProvider> parameterTypes)
     {
-        QualifiedObjectName functionName = getFunctionAndTypeResolver().qualifyObjectName(QualifiedName.of(name));
+        Optional<FunctionNamespaceManager<?>> functionNamespaceManager = getServingFunctionNamespaceManager(functionName.getCatalogSchemaName());
+        checkArgument(functionNamespaceManager.isPresent(), "Cannot find function namespace for function '%s'", functionName);
         if (parameterTypes.stream().noneMatch(TypeSignatureProvider::hasDependency)) {
             return lookupCachedFunction(functionName, parameterTypes);
```
rschlussel, line 627: there's too many assumptions going on here now. this will throw an exception later if the fucntionNamepsaceManager here is not the DEFAULT or currentDefault.

presto-main/src/main/java/com/facebook/presto/metadata/FunctionAndTypeManager.java
```diff
@@ -129,6 +131,7 @@ public class FunctionAndTypeManager
     private final LoadingCache<FunctionResolutionCacheKey, FunctionHandle> functionCache;
     private final CacheStatsMBean cacheStatsMBean;
     private final boolean nativeExecution;
+    private CatalogSchemaName currentDefaultNamespace = DEFAULT_NAMESPACE;
```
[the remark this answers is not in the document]
rschlussel: Last comment about default_namespace, but have you checked all the things that are using DEFAULT_NAMESPACE right now? Is it still correct for them to do that?

presto-main/src/main/java/com/facebook/presto/metadata/FunctionAndTypeManager.java
```diff
@@ -726,6 +757,17 @@ private Optional<FunctionNamespaceManager<? extends SqlFunction>> getServingFunc
         return Optional.ofNullable(functionNamespaceManagers.get(typeSignatureBase.getTypeName().getCatalogName()));
     }
 
+    @Override
+    @SuppressWarnings("unchecked")
+    public SpecializedFunctionKey getSpecializedFunctionKey(Signature signature)
```
rschlussel, line 762: calling out to BuiltInTypeAndFunctionNamepsaceManager to use that functionality for other namespace managers seems to violate some of the boundaries of the abstraction. Maybe we should extract that specialized function cache logic out of BuiltInTypeAndFunctionNamespaceManager if it needs to be common to other function namespace managers.

pdabre12:
I will get back to this soon.

presto-spi/src/main/java/com/facebook/presto/spi/function/FunctionMetadataManager.java
```diff
@@ -16,4 +16,9 @@
 public interface FunctionMetadataManager
 {
     FunctionMetadata getFunctionMetadata(FunctionHandle functionHandle);
+
+    /**
```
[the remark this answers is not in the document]
pdabre12: Removed this description, moved this function to `FunctionAndTypeManager`.

presto-main/src/main/java/com/facebook/presto/metadata/FunctionAndTypeManager.java
pdabre12: Added a check.

presto-main/src/main/java/com/facebook/presto/metadata/FunctionAndTypeManager.java
pdabre12: Extracted it to `FunctionAndTypeManager`.

presto-main/src/main/java/com/facebook/presto/metadata/FunctionAndTypeManager.java
```diff
@@ -698,7 +728,8 @@ private FunctionHandle resolveFunctionInternal(Optional<TransactionId> transacti
 
     private FunctionHandle resolveBuiltInFunction(QualifiedObjectName functionName, List<TypeSignatureProvider> parameterTypes)
     {
-        checkArgument(functionName.getCatalogSchemaName().equals(DEFAULT_NAMESPACE), "Expect built-in functions");
+        checkArgument(functionName.getCatalogSchemaName().equals(currentDefaultNamespace) ||
```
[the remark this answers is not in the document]
pdabre12: There might be some functions an eg: `key_sampling_percent` function https://github.com/prestodb/presto/blob/a391cec0e6d9f52c3a2a1ef1eba60044305ebbf5/presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/KeyBasedSampler.java#L151-L155, which would always be under the `JAVA_BUILTIN_NAMESPACE`.

presto-main/src/main/java/com/facebook/presto/metadata/FunctionAndTypeManager.java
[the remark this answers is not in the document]
pdabre12: Added a config in `FunctionsConfig`, the default value is going to be `JAVA_BUILTIN_NAMESPACE` so we always have a default namespace configured.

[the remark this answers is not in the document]
pdabre12: Yes, I went over usages of `DEFAULT_NAMESPACE` (now `JAVA_BUILTIN_NAMESPACE`), they are used in the right places.

pdabre12:
@rschlussel 
Thank you for the review.
Addressed your comments. Can you please have another look?

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/CombineApproxPercentileFunctions.java
```diff
@@ -111,8 +112,14 @@ private static int getPercentilePosition(FunctionHandle functionHandle)
         // approx_percentile(x, percentage, accuracy) -> arguments.size() == 3 && arguments.get(1) is Double
         // approx_percentile(x, w, percentage) -> arguments.size() == 3 && arguments.get(1) is BigInt
         // approx_percentile(x, w, percentage, accuracy) -> arguments.size() == 4
-        checkState(functionHandle instanceof BuiltInFunctionHandle);
-        List<TypeSignature> argumentTypes = ((BuiltInFunctionHandle) functionHandle).getSignature().getArgumentTypes();
+        checkState(functionHandle instanceof BuiltInFunctionHandle || functionHandle instanceof SqlFunctionHandle);
+        List<TypeSignature> argumentTypes;
+        if (functionHandle instanceof BuiltInFunctionHandle) {
```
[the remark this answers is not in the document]
pdabre12: Done, thanks.

presto-main/src/main/java/com/facebook/presto/metadata/FunctionAndTypeManager.java
```diff
@@ -361,15 +379,18 @@ public void registerBuiltInFunctions(List<? extends SqlFunction> functions)
     public List<SqlFunction> listFunctions(Session session, Optional<String> likePattern, Optional<String> escape)
     {
         ImmutableList.Builder<SqlFunction> functions = new ImmutableList.Builder<>();
-        if (!isListBuiltInFunctionsOnly(session)) {
+        if (isListBuiltInFunctionsOnly(session)) {
+            functions.addAll(functionNamespaceManagers.entrySet().stream()
```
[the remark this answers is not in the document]
rschlussel: why stream the map and filter instead of getting the functionNamespacemanager directly?
```
functions.addAll(functionNamespaceManagers.get(
     defaultNamespace.getCatalogName().listFunctions(likePattern,escape).stream()
    .collect(toImmutableList())
```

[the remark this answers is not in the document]
pdabre12: We are configuring function namespaces using `.properties` files located under `etc/function-namespace`. 
The name of the `.properties` file without the `.properties` extension corresponds to the `catalogName`, which is passed to the `FunctionNamespaceManager` factories. These factories then create the respective function namespaces.
This `catalogName` is subsequently registered as a `ServingCatalog` within its corresponding `FunctionNamespaceManager`.

In `FunctionAndTypeManager`, we maintain a map  of the `FunctionNamespaceManager` instances, where:
	•	The key is the catalogName.
	•	The value is the corresponding `FunctionNamespaceManager`.
For example:` {"native": NativeFunctionNamespaceManager}`

To retrieve builtiin functions from the `defaultNamespace`, 
We use `defaultNamespace.getCatalogName()` to get the `catalogName`. If the `catalogName` exists in the map, we fetch the corresponding `FunctionNamespaceManager` (e.g., `NativeFunctionNamespaceManager`) and call `getFunctions()` on it.

The only requirement for users to ensure is that the name of the file (`{catalogName}.properties`) file used to load the functionNamespaces must match the catalog name of the `presto.default-namespace` configuration property for the defaultNamespace, but its not a new requirement, even with the current implementation that is enforced. 
Let me know if that works for you.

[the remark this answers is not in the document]
pdabre12: You are right, thanks for catching that. Changed it.

rschlussel (approved):

presto-spi/src/main/java/com/facebook/presto/spi/ConnectorSystemConfig.java
```diff
@@ -13,7 +13,10 @@
  */
 package com.facebook.presto.spi;
 
+import com.facebook.presto.common.CatalogSchemaName;
+
 public interface ConnectorSystemConfig
 {
     boolean isNativeExecution();
+    CatalogSchemaName getDefaultNamespace();
```
tdcmeehan, line 21: Instead of leaving the responsibility of resolving the default namespace to individual connectors, can we please use `FunctionResolution` to check if a function is of a particular type?  I think this is used in the Iceberg connector--instead of a static mapping of min/max/approx_distinct by name, let's check if the function is any of these by using methods on `StandardFunctionResolution`.

tdcmeehan (changes requested):

presto-spi/src/main/java/com/facebook/presto/spi/ConnectorSystemConfig.java
pdabre12: Done, thanks.

tdcmeehan (approved):

presto-native-execution/presto_cpp/main/common/tests/ConfigTest.cpp
```diff
@@ -12,17 +12,59 @@
...
+    const FunctionMap& functionMap,
+    const std::string& prestoDefaultNamespacePrefix) {
+  static const std::unordered_set<std::string> kBlockList = {
+      "row_constructor", "in", "is_null"};
+
+  std::vector<std::string> prestoDefaultNamespacePrefixParts;
+  folly::split(
+      '.',
+      prestoDefaultNamespacePrefix,
+      prestoDefaultNamespacePrefixParts,
```
aditi-pandit, line 46: Add CHECK that prestoDefaultNamespacePrefixParts has only 2 parts.

presto-native-execution/presto_cpp/main/common/tests/ConfigTest.cpp
```diff
@@ -12,17 +12,59 @@
...
+  for (const auto& [functionName, _] : functionMap) {
+    // Skip internal functions. They don't have any prefix.
+    if ((kBlockList.count(functionName) != 0) ||
+        (functionName.find("$internal$") != std::string::npos)) {
+      continue;
+    }
+
+    std::vector<std::string> parts;
+    folly::split('.', functionName, parts, true);
+    if ((parts.size() != 3) ||
```
aditi-pandit, line 58: Should this be a check instead ? Are there functions that don't follow since (since you added all the built-ins with prefixes) ?

presto-native-execution/presto_cpp/main/types/PrestoToVeloxExpr.h
```diff
@@ -14,10 +14,19 @@
 #pragma once
 
 #include <stdexcept>
+#include "presto_cpp/main/common/Configs.h"
 #include "presto_cpp/main/types/TypeParser.h"
 #include "presto_cpp/presto_protocol/core/presto_protocol_core.h"
 #include "velox/core/Expressions.h"
 
+namespace {
```
aditi-pandit, line 22: Consider moving this to presto_cpp/main/common/Utils.h

aditi-pandit (commented):
@pdabre12 : Have 2 minor comments.

presto-native-execution/presto_cpp/main/common/tests/ConfigTest.cpp
pdabre12: All the functions should have a prefix. Added the check.

aditi-pandit (approved):
Thanks @pdabre12

tdcmeehan (approved):

PR #23378, opened 2024-08-05
[native] Add TaskInfo summarize

guhanjie:
## Description
[native] TaskInfo carries excessive detail infos and metrics to coordinator, leading unnecessary CPU consumption.  
Only `EXPLAIN ANALYZE` queries don't summarize the task infos, 
and [native] TaskInfo lacks summarize implementations, which leading significant serialization and deserialization overhead on coordinator <-> workers.

## Motivation and Context
The same workload(50+K runtime tasks) on [presto-native] vs. [presto-java] consumed more 20+ CPU cores. 
Each http/rpc of TaskInfo on [presto-native] increased extra 10K-30K(even more) bytes to ser-deser at now.  
This PR will save much cpu on coordinator/worker with TaskInfo summarize(API: createOrUpdateTask, getTaskInfo).

## Impact
TaskInfo summarize will strip TaskStats at runtime, and get detail metrics at Task final state.


tdcmeehan:
@guhanjie this is a great catch.  Only `EXPLAIN ANALYZE` queries don't summarize the task infos: https://github.com/prestodb/presto/blob/245c93f60bb514f8e66a233bd95a1e85ceb0b88c/presto-main/src/main/java/com/facebook/presto/execution/SqlQueryExecution.java#L586

I expect this should also free up cycles in the coordinator, perhaps substantially.

guhanjie:
> @guhanjie this is a great catch. Only `EXPLAIN ANALYZE` queries don't summarize the task infos:
> 
> https://github.com/prestodb/presto/blob/245c93f60bb514f8e66a233bd95a1e85ceb0b88c/presto-main/src/main/java/com/facebook/presto/execution/SqlQueryExecution.java#L586
> 
> I expect this should also free up cycles in the coordinator, perhaps substantially.

guhanjie:
> @guhanjie this is a great catch. Only `EXPLAIN ANALYZE` queries don't summarize the task infos:
> 
> https://github.com/prestodb/presto/blob/245c93f60bb514f8e66a233bd95a1e85ceb0b88c/presto-main/src/main/java/com/facebook/presto/execution/SqlQueryExecution.java#L586
> 
> I expect this should also free up cycles in the coordinator, perhaps substantially.

Yeap, actually this PR is to optimize coordinator cycles primarily(found 2x cpu usage vs. [presto-java]), but also it would free up on workers(as serialization of multi-level nested objects, TaskInfo/TaskStats/PipelineStats/OperatorStats/DriverStats/RuntimeStats ...)

presto-native-execution/presto_cpp/main/TaskManager.cpp
```diff
@@ -763,7 +771,7 @@ void TaskManager::cancelAbandonedTasks() {
 
 folly::Future<std::unique_ptr<protocol::TaskInfo>> TaskManager::getTaskInfo(
     const TaskId& taskId,
-    bool summarize,
+    const bool summarize,
```
aditi-pandit, line 774: const not required with bool

presto-native-execution/presto_cpp/main/TaskManager.cpp
```diff
@@ -442,6 +445,7 @@ std::unique_ptr<protocol::TaskInfo> TaskManager::createOrUpdateBatchTask(
     const protocol::TaskId& taskId,
     const protocol::BatchTaskUpdateRequest& batchUpdateRequest,
     const velox::core::PlanFragment& planFragment,
+    const bool summarize,
```
aditi-pandit, line 448: const is not used with bool

presto-native-execution/presto_cpp/main/TaskManager.cpp
```diff
@@ -427,13 +428,15 @@ std::unique_ptr<protocol::TaskInfo> TaskManager::createOrUpdateTask(
     const protocol::TaskId& taskId,
     const protocol::TaskUpdateRequest& updateRequest,
     const velox::core::PlanFragment& planFragment,
+    const bool summarize,
```
aditi-pandit, line 431: const is not used with bool

presto-native-execution/presto_cpp/main/TaskManager.cpp
```diff
@@ -360,6 +360,7 @@ void TaskManager::acknowledgeResults(
 std::unique_ptr<TaskInfo> TaskManager::createOrUpdateErrorTask(
     const TaskId& taskId,
     const std::exception_ptr& exception,
+    const bool summarize,
```
aditi-pandit, line 363: const is not used with bool

presto-native-execution/presto_cpp/main/PrestoTask.h
```diff
@@ -161,9 +163,234 @@ struct PrestoTask {
...
+        taskStats.outputPositions,
+        taskStats.physicalWrittenDataSizeInBytes,
+        taskStats.fullGcCount,
+        taskStats.fullGcTimeInMillis,
+        {}, // stripe PipelineStats
+        {} // stripe runtimeStats
+    };
+  }
+
+  /// Summarize PipelineStats to avoid too heavy ser & de-ser on communication
```
aditi-pandit, line 292: Comments should be complete sentences ending in period.

aditi-pandit:
@guhanjie : Thanks for your code. This is a great catch and would be a nice improvement.

I have some nits so far. Would be great to add a unit test for your code in TaskManagerTest.cpp though. Please can you look into it.

guhanjie:
> @guhanjie : Thanks for your code. This is a great catch and would be a nice improvement.
> 
> I have some nits so far. Would be great to add a unit test for your code in TaskManagerTest.cpp though. Please can you look into it.

Thanks for comments, all done.   
I touched TaskManagerTest.cpp with injecting parameter [summarize = true], to make sure the summarize path correct, and all tests passed.  
Does it mean the new path already covered?

guhanjie:
@aditi-pandit Any question or comments still? I would close it after merged

aditi-pandit:
> @aditi-pandit Thanks for comments, all done. I touched TaskManagerTest.cpp with injecting parameter [summarize = true], to make sure the summarize path correct, and all tests passed. Does it mean the new path already covered?

@guhanjie : Would be great to retrieve the TaskInfo as well in some test and check that its structure matches the summarized one you expect.

guhanjie:
@aditi-pandit done

spershin (commented):
@guhanjie 
As I understand this PR does the following:
1. If the task is not finalized, it strips all pipeline, operator and runtime stats.
2. If the task is finalized it strips all operator stats.

Is this correct?

I'm seeing that we copy the whole structure in all cases before sending it to the coordinator.
Wondering if it would be the best to implement this logic when generating the Presto structure from the Velox structures. I think we might even have some rudimentary logic stripping some metrics in there, but likely disabled under a config parameter.
Such we can save some CPU on copying a bunch of data from Velox to Presto structures and then partially copying Presto structures again.

guhanjie:
@joshk The logic now respected the original implementation on presto coordinator<->worker(java).  
And the parameter[summarize] is optional at runtime, there are 2 cases:
1. explain analyze which not summarize, as described by @tdcmeehan above, maybe we need all detailed metrics
2. normal query which summarize, we don't need too much detailed infos, just save cpu cycles on coordinator
It would be great if some rudimentary logic stripping on velox, but it would be better with parameter rather than static config flag as [explain analyze] maybe not summarize.

guhanjie:
@joshk I have checked it again, and I think no more needs to be done on this PR.   
Let me clear it:
Now taskinfo summarize does the following:
1. If the task is running, we don't need detail stats of pipeline, operator and runtime, so strip them when get.
2. If the task is finalized, we still need final stats of pipeline, operator and runtime(for UI or profiling), so summarize them.

All these stats are passing through Velox -> Worker -> Coordinator.
Worker should hold all detail stats in memory rather than striped, and output as needed(summarize or not).

The target of this PR is to reduce the Serialization and Deserialization of TaskInfo(from Worker to Coordinator, with shallow copy), rather than strip TaskInfo Object holding on Worker or generating on Velox.

The config parameter [kSkipRuntimeStatsInRunningTaskInfo] enables to skip build taskRuntimeStats on running, but still keeping some important metrics, such as "drivers.total", "drivers.running", ... "stuck_op."

aditi-pandit:
@guhanjie : Sergey is proposing to optimize your approach to TaskInfo summarization by building just the limited structures when going from Velox -> Presto, rather than generating the entire Presto TaskInfo structure and then limiting what is transferred to the co-ordinator. 

With his suggestion you avoid a copy from Velox -> Prestissimo structures, in addition of the serialization/deserialization between the worker and co-ordinator. 

The worker holds the detail stats in Velox structures but doesn't transform them to the Presto application when they are not required.

Hope this gives more background.

guhanjie:
@spershin , @aditi-pandit : I see what you are saying, and I think it's ok, I would refine it at time.  
But what I am concerning now is,   
If we use summarize, means we don't need pipeline stats at runtime, and if that, wondering whether we should do copy Velox structures pipeline stats(aka `statsCopy`), would also skip op->stats copy to save cpu in Velox?
Note: Even though, op->stats must summarize at final, so we can get the op summarize stats at final.  

[https://github.com/facebookincubator/velox/blob/d16b195bf4b3b27295ab99a4afe27c4ccacd5eb3/velox/exec/Task.cpp#L2126]

`
TaskStats Task::taskStats() const {
...
    for (auto& op : driver->operators()) {
      auto statsCopy = op->stats(false);
      aggregateOperatorRuntimeStats(statsCopy.runtimeStats);
      addRunningTimeOperatorMetrics(statsCopy);
      taskStats.pipelineStats[statsCopy.pipelineId]
          .operatorStats[statsCopy.operatorId]
          .add(statsCopy);
    }
...
}
`

spershin:
@guhanjie 

Hmm, not sure I understand your concern.

This is how I see this "summarize" feature: it basically strips metrics from pipelines and operators and leaves only the task level metrics. Sort of "return only aggregated per task metrics" rather than detailed per operator (or driver if you wish).

This differs from what native worker does - we always send metrics in operators and pipelines.

In the past we suspected that it can have a performance hit on the coordinator, as native sends many more metrics than java and we even added some config flag to control that - avoid sending the detailed metrics until the query is finished, failed or otherwise 'done with'.

Later we turned off the config property, because it was useful to see the operator stats when the query was still working. For example, when a query got stuck.

My suggestion here is:
1. Maybe we should try reusing this config property in case we might need to turn it on or off.
2. Attempt to strip the detailed metrics before we convert velox -> presto and not after to safe some CPU cycles.

These suggestions might need considerable changes to your PR, though. Maybe we can do them in stages.

aditi-pandit:
@spershin : I like the idea of using a config to enable striping metrics from pipelines and operators since it might be useful to see the metrics while debugging.

guhanjie:
@spershin Thanks for your explanation about history, my concern is just a question, and I think it's ok as it is now on Velox.    
Velox can hold detail metrics for operators and pipelines, we can get it from Task::taskStats() as we need Whenever.    
And I have finished  strip the detailed metrics **when** convert velox -> presto, rathor than **after** as my previous implementation, I think it's better and enough to save CPU cycles.

@aditi-pandit please review it, and feel free to point out any issues you find.

guhanjie:
> @guhanjie
> 
> Hmm, not sure I understand your concern.
> 
> This is how I see this "summarize" feature: it basically strips metrics from pipelines and operators and leaves only the task level metrics. Sort of "return only aggregated per task metrics" rather than detailed per operator (or driver if you wish).
> 
> This differs from what native worker does - we always send metrics in operators and pipelines.
> 
> In the past we suspected that it can have a performance hit on the coordinator, as native sends many more metrics than java and we even added some config flag to control that - avoid sending the detailed metrics until the query is finished, failed or otherwise 'done with'.
> 
> Later we turned off the config property, because it was useful to see the operator stats when the query was still working. For example, when a query got stuck.
> 
> My suggestion here is:
> 
> 1. Maybe we should try reusing this config property in case we might need to turn it on or off.
> 2. Attempt to strip the detailed metrics before we convert velox -> presto and not after to safe some CPU cycles.
> 
> These suggestions might need considerable changes to your PR, though. Maybe we can do them in stages.

@spershin @aditi-pandit Hi Aditi and Sergey, too long time last on this PR, I think we should resolve it promptly. 
I agree with Sergey's suggestion, let's review all the updates:
1. The core of this PR is: Velox worker missed to summarize TaskInfo, which carries detailed op and pipeline info to Coordinator.
2. At first, I tried to strip TaskInfo simply with shallow copy, and add some tests.
3. @spershin pointed there's a config flag on velox, and suggested to strip metrics on converting velox -> presto

I agree with the suggestions. Let's fix the issue promptly over prolonged discussion.  

> Suggestion 1:
I checked the config flag, I'am not sure if it is the following one?
image
image
If it is, it has been already setup on, and it only controls runtimeStats, but not op and pipelineStats.  
If it is not, where's the flag, maybe we can use it?

> Suggestion 2:
Please check the latest commits, which already optimized, strip metrics on converting velox -> presto.

presto-native-execution/presto_cpp/main/TaskResource.cpp
```diff
@@ -351,6 +359,7 @@ proxygen::RequestHandler* TaskResource::createOrUpdateTask(
             taskId,
             updateRequest,
             planFragment,
+            summarize,
```
arhimondr, line 362: `deleteTask` also needs to support the flag: https://github.com/prestodb/presto/blob/5d63246c7dbd93cf0f6f444780bddbc05545bdda/presto-main/src/main/java/com/facebook/presto/server/TaskResource.java#L271

presto-native-execution/presto_cpp/main/PrestoTask.cpp
```diff
@@ -595,197 +598,242 @@ void PrestoTask::updateExecutionInfoLocked(
...
-    }
-
-    for (const auto& driverStat : veloxPipeline.driverStats) {
-      for (const auto& [name, value] : driverStat.runtimeStats) {
-        addRuntimeMetric(taskRuntimeStats, name, value);
+  // strip pipelines when running to avoid too heavy ser & de-ser on
+  // communication.
+  if (stripPipelines) {
+    prestoTaskStats.pipelines = {};
+    for (int i = 0; i < veloxTaskStats.pipelineStats.size(); ++i) {
```
arhimondr, line 605: Maybe extract this loop to the top (to avoid duplication)

presto-native-execution/presto_cpp/main/PrestoTask.cpp
```diff
@@ -452,7 +452,9 @@ protocol::TaskInfo PrestoTask::updateInfoLocked() {
   updateMemoryInfoLocked(veloxTaskStats, currentTimeMs, taskRuntimeStats);
 
   // Update execution related info.
-  updateExecutionInfoLocked(veloxTaskStats, prestoTaskStatus, taskRuntimeStats);
+  bool stripPipelines = summarize && !isFinalState(prestoTaskStatus.state);
```
arhimondr, line 455: nit: `stripPipelineStats`?

presto-native-execution/presto_cpp/main/PrestoTask.cpp
```diff
@@ -580,7 +582,8 @@ void PrestoTask::updateMemoryInfoLocked(
 void PrestoTask::updateExecutionInfoLocked(
     const velox::exec::TaskStats& veloxTaskStats,
     const protocol::TaskStatus& prestoTaskStatus,
-    std::unordered_map<std::string, velox::RuntimeMetric>& taskRuntimeStats) {
+    std::unordered_map<std::string, velox::RuntimeMetric>& taskRuntimeStats,
+    bool stripPipelines) {
```
arhimondr, line 586: ditto

presto-native-execution/presto_cpp/main/PrestoTask.h
```diff
@@ -82,6 +82,8 @@ struct ResultRequest {
         maxSize(_maxSize) {}
 };
 
+bool isFinalState(protocol::TaskState state);
```
arhimondr, line 85: Why moving this?

presto-native-execution/presto_cpp/main/PrestoTask.h
```diff
@@ -161,9 +163,9 @@ struct PrestoTask {
     return updateStatusLocked();
   }
 
-  protocol::TaskInfo updateInfo() {
+  protocol::TaskInfo updateInfo(bool summarize = false) {
```
arhimondr, line 166: Maybe drop the optional parameter and make sure summarize is always passed. All methods invoking this should take into account `sumarize` flag.

presto-native-execution/presto_cpp/main/PrestoTask.h
```diff
@@ -172,7 +174,7 @@ struct PrestoTask {
 
   /// Invoked to update presto task status from the updated velox task stats.
   protocol::TaskStatus updateStatusLocked();
-  protocol::TaskInfo updateInfoLocked();
+  protocol::TaskInfo updateInfoLocked(bool summarize = false);
```
arhimondr, line 177: Ditto

presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp
```diff
@@ -1567,6 +1571,70 @@ TEST_F(TaskManagerTest, buildSpillDirectoryFailure) {
...
+
+  auto infoRequestState = http::CallbackRequestHandlerState::create();
+
+  const auto finalTaskInfo =
+      taskManager_
+          ->getTaskInfo(
+              taskId, true, std::nullopt, std::nullopt, infoRequestState)
+          .get();
+
+  ASSERT_TRUE(finalTaskInfo->stats.pipelines.empty());
```
arhimondr, line 1635: Should final task info include pipeline stats?

presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp
```diff
@@ -1567,6 +1571,70 @@ TEST_F(TaskManagerTest, buildSpillDirectoryFailure) {
...
+                          .planFragment();
+
+  long splitSequenceId{0};
+
+  protocol::TaskId taskId = "scan.0.0.1.0";
+
+  protocol::TaskUpdateRequest updateRequest;
+  updateRequest.sources.push_back(
+      makeSource("0", filePaths, true, splitSequenceId));
+  auto taskInfo = createOrUpdateTask(taskId, updateRequest, planFragment);
```
arhimondr, line 1597: Maybe also test `getInfo` and `delete`

presto-native-execution/presto_cpp/main/PrestoTask.cpp
```diff
@@ -595,197 +598,242 @@ void PrestoTask::updateExecutionInfoLocked(
...
+        prestoTaskStats.outputPositions += prestoPipeline.outputPositions;
+        prestoTaskStats.outputDataSizeInBytes +=
+            prestoPipeline.outputDataSizeInBytes;
       }
 
-      if (!veloxOp.dynamicFilterStats.empty()) {
-        prestoOp.dynamicFilterStats = toPrestoDynamicFilterStats(veloxOp);
+      for (const auto& driverStat : veloxPipeline.driverStats) {
+        for (const auto& [name, value] : driverStat.runtimeStats) {
+          addRuntimeMetric(taskRuntimeStats, name, value);
```
arhimondr, line 695: Do we want to keep runtime metrics when `summarize`? It seems like this is what Java does: https://github.com/prestodb/presto/blob/5d63246c7dbd93cf0f6f444780bddbc05545bdda/presto-main/src/main/java/com/facebook/presto/operator/TaskStats.java#L599

Collection of runtime stats can be turned off with the flag @spershin suggested.

arhimondr (commented):
Thank you for working on this @guhanjie . 

We have observed significantly higher CPU utilization on coordinator for native workers due to ignored "summarize" flag.

steveburnett:
Does this PR need a release note, or `NO RELEASE NOTE`?

presto-native-execution/presto_cpp/main/TaskResource.cpp
spershin: Also
`TaskResource::getTaskInfo(`
and 
`public List<TaskInfo> getAllTaskInfo(`
Although the latter seems like not supported by native worker and might be only for coordinator?

spershin:
> Does this PR need a release note, or `NO RELEASE NOTE`?

@steveburnett 
Probably `NO RELEASE NOTE`, since it is a bugfix?

tdcmeehan:
[We add bug fixes to our release notes](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines#order-of-changes), but in this case, it seems like the highest impact of the PR is to improve coordinator performance under heavy load, so I'd choose something along the lines of 

`Improve coordinator task management performance.`

And put it in the `Native Execution` section.

steveburnett:
> [We add bug fixes to our release notes](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines#order-of-changes), but in this case, it seems like the highest impact of the PR is to improve coordinator performance under heavy load, so I'd choose something along the lines of
> 
> `Improve coordinator task management performance.`
> 
> And put it in the `Native Execution` section.

So, draft release note entry similar to this:

```
== RELEASE NOTES ==

* Improve coordinator task management performance. :pr:`23378`

```

arhimondr:
@guhanjie Hasn't responded. This problem is a high priority problem for Meta. Opening an alternative PR based of this PR here: https://github.com/prestodb/presto/pull/24369

guhanjie:
I'm glad to see this bug has been fixed. Thank you, everyone! @tdcmeehan @aditi-pandit @spershin @steveburnett @arhimondr

PR #23384, opened 2024-08-05
[native] Add presto.default-namespace Prestissimo config property

pdabre12:
## Description

Adds the `presto.default-namespace` config property. 

Resolves: #23385 

## Release Notes

```
== RELEASE NOTES ==
Prestissimo Changes
* Add ``presto.default-namespace`` Prestissimo configuration property :pr:`23384`
```

### DIFF
--- a/presto-docs/src/main/sphinx/presto_cpp/properties.rst
+++ b/presto-docs/src/main/sphinx/presto_cpp/properties.rst
@@ -95,6 +95,14 @@ The configuration properties of Presto C++ workers are described here, in alphab
 
   In-memory cache.
 
+``presto.default-namespace``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``string``
+* **Default value:** ``presto.default``
+
+  Specifies the namespace prefix for native C++ functions.
+
 ``query.max-memory-per-node``
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
--- a/presto-native-execution/presto_cpp/main/common/Configs.h
+++ b/presto-native-execution/presto_cpp/main/common/Configs.h
@@ -585,6 +585,10 @@ class SystemConfig : public ConfigBase {
   static constexpr std::string_view kDriverMaxPagePartitioningBufferSize{
       "driver.max-page-partitioning-buffer-size"};
 
+  // Specifies the default Presto namespace prefix.
+  static constexpr std::string_view kPrestoDefaultNamespacePrefix{
+      "presto.default-namespace"};
+
   SystemConfig();
 
   virtual ~SystemConfig() = default;
@@ -802,6 +806,7 @@ class SystemConfig : public ConfigBase {
   bool enableRuntimeMetricsCollection() const;
 
   bool prestoNativeSidecar() const;
+  std::string prestoDefaultNamespacePrefix() const;
 };
 
 /// Provides access to node properties defined in node.properties file.
--- a/presto-native-execution/presto_cpp/main/types/PrestoToVeloxExpr.cpp
+++ b/presto-native-execution/presto_cpp/main/types/PrestoToVeloxExpr.cpp
@@ -33,24 +33,41 @@ std::string toJsonString(const T& value) {
 }
 
 std::string mapScalarFunction(const std::string& name) {
+  static const std::string prestoDefaultNamespacePrefix =
+      SystemConfig::instance()->prestoDefaultNamespacePrefix();
   static const std::unordered_map<std::string, std::string> kFunctionNames = {
       // Operator overrides: com.facebook.presto.common.function.OperatorType
-      {"presto.default.$operator$add", "presto.default.plus"},
-      {"presto.default.$operator$between", "presto.default.between"},
-      {"presto.default.$operator$divide", "presto.default.divide"},
-      {"presto.default.$operator$equal", "presto.default.eq"},
-      {"presto.default.$operator$greater_than", "presto.default.gt"},
-      {"presto.default.$operator$greater_than_or_equal", "presto.default.gte"},
+      {"presto.default.$operator$add",
+       formatDefaultNamespacePrefix(prestoDefaultNamespacePrefix, "plus")},
+      {"presto.default.$operator$between",
+       formatDefaultNamespacePrefix(prestoDefaultNamespacePrefix, "between")},
+      {"presto.default.$operator$divide",
+       formatDefaultNamespacePrefix(prestoDefaultNamespacePrefix, "divide")},
+      {"presto.default.$operator$equal",
+       formatDefaultNamespacePrefix(prestoDefaultNamespacePrefix, "eq")},
+      {"presto.default.$operator$greater_than",
+       formatDefaultNamespacePrefix(prestoDefaultNamespacePrefix, "gt")},
+      {"presto.default.$operator$greater_than_or_equal",
+       formatDefaultNamespacePrefix(prestoDefaultNamespacePrefix, "gte")},
       {"presto.default.$operator$is_distinct_from",
-       "presto.default.distinct_from"},
-      {"presto.default.$operator$less_than", "presto.default.lt"},
-      {"presto.default.$operator$less_than_or_equal", "presto.default.lte"},
-      {"presto.default.$operator$modulus", "presto.default.mod"},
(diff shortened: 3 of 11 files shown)


steveburnett:
Suggestions: 

* Add documentation for this config property to https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/presto_cpp/properties.rst 

* Add a release note entry following the Release Notes Guidelines, similar to the following draft: 

```
== RELEASE NOTES ==

General Changes
* Add ``presto.default-namespace`` configuration property for Presto C++ :pr:`23384`
```

presto-docs/src/main/sphinx/presto_cpp/properties.rst
```diff
@@ -95,6 +95,14 @@ The configuration properties of Presto C++ workers are described here, in alphab
 
   In-memory cache.
 
+``presto.default-namespace``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``string``
+* **Default value:** ``presto.default``
+
+  Specifies the namespace prefix for presto functions.
```
steveburnett, line 104:
```suggestion
  Specifies the namespace prefix for Presto functions.
```
Just a nit for capitalization.

steveburnett (changes requested):
Thanks for the doc!

pdabre12:
@aditi-pandit @czentgr Please take a look.

steveburnett (dismissed):
LGTM! (docs)

Pull branch, new local build, looks good. Thanks!

pdabre12:
@steveburnett Thank you for you review !

presto-native-execution/presto_cpp/main/types/PrestoToVeloxExpr.cpp
```diff
@@ -33,24 +34,40 @@ std::string toJsonString(const T& value) {
...
+      SystemConfig::instance()->prestoDefaultNamespacePrefix();
   static const std::unordered_map<std::string, std::string> kFunctionNames = {
       // Operator overrides: com.facebook.presto.common.function.OperatorType
-      {"presto.default.$operator$add", "presto.default.plus"},
-      {"presto.default.$operator$between", "presto.default.between"},
-      {"presto.default.$operator$divide", "presto.default.divide"},
-      {"presto.default.$operator$equal", "presto.default.eq"},
-      {"presto.default.$operator$greater_than", "presto.default.gt"},
-      {"presto.default.$operator$greater_than_or_equal", "presto.default.gte"},
+      {"presto.default.$operator$add",
```
aditi-pandit, line 41: @pdabre12 : High level question about this... Should we be using prestoDefaultNamespacePrefix for the key function name in Velox as well ? Since we are registering functions with this prefix, then it should be used here as well.

aditi-pandit (commented):
@pdabre12 : Thanks for this PR. Have one high level question.

presto-native-execution/presto_cpp/main/types/PrestoToVeloxExpr.cpp
pdabre12: @aditi-pandit You are right. Thank you for catching this. 
Updated my PR.

presto-native-execution/presto_cpp/main/common/Configs.h
```diff
@@ -584,6 +584,10 @@ class SystemConfig : public ConfigBase {
   static constexpr std::string_view kDriverMaxPagePartitioningBufferSize{
       "driver.max-page-partitioning-buffer-size"};
 
+  // Specifies the default Presto namespace prefix.
+  static constexpr std::string_view kPrestoDefaultNamespacePrefix{
+      "presto.default-namespace"};
```
aditi-pandit, line 589: @pdabre12 : Should this be just "presto.default" instead of "presto.default-namespace"

presto-native-execution/presto_cpp/main/types/PrestoToVeloxExpr.cpp
```diff
@@ -33,26 +34,45 @@ std::string toJsonString(const T& value) {
...
-       "presto.default.distinct_from"},
-      {"presto.default.$operator$less_than", "presto.default.lt"},
-      {"presto.default.$operator$less_than_or_equal", "presto.default.lte"},
-      {"presto.default.$operator$modulus", "presto.default.mod"},
-      {"presto.default.$operator$multiply", "presto.default.multiply"},
-      {"presto.default.$operator$negation", "presto.default.negate"},
-      {"presto.default.$operator$not_equal", "presto.default.neq"},
-      {"presto.default.$operator$subtract", "presto.default.minus"},
-      {"presto.default.$operator$subscript", "presto.default.subscript"},
+      {fmt::format("{}$operator$add", prestoDefaultNamespacePrefix),
```
aditi-pandit, line 41: Write an inline function to add the defaultNamespacePrefix to both the operator name and Velox function name and use it for all entries below.

presto-native-execution/presto_cpp/main/types/PrestoToVeloxExpr.cpp
```diff
@@ -66,11 +86,15 @@ std::string mapScalarFunction(const std::string& name) {
...
 
 std::string mapAggregateOrWindowFunction(const std::string& name) {
+  static const std::string prestoDefaultNamespacePrefix =
+      SystemConfig::instance()->prestoDefaultNamespacePrefix();
   static const std::unordered_map<std::string, std::string> kFunctionNames = {
-      {"presto.default.$internal$max_data_size_for_stats",
-       "presto.default.max_data_size_for_stats"},
-      {"presto.default.$internal$sum_data_size_for_stats",
-       "presto.default.sum_data_size_for_stats"},
+      {fmt::format(
```
aditi-pandit, line 92: Write an inline function for adding the defaultNamespacePrefix to the function name and use it everywhere.

aditi-pandit (commented):
@pdabre12 : Would be great to add a unit test for this logic as well. 

One possible test could be to retrieve all functions from Velox and check they have the default namespace.

presto-native-execution/presto_cpp/main/common/Configs.h
czentgr: This is the config name. There are no configs that start with `presto.
How about a more specific name like `default-object-namespace` or something like that?

pdabre12: Thanks @czentgr .
Yes it is the config name.  There is a config that start with `presto.` `(presto.version)`.  I thought it might make sense to put this config under `presto.` 
`default-object-namespace` seems fine , but what about `default-presto-object-namespace` or `default-presto-namespace` ?

czentgr: Thanks. I overlooked the one that actually has it in the name :).
But the reason it is there is because it is a "shared" config that also exists in Presto itself. The namespace doesn't really exist there as otherwise we could just use the same name for it. And so using the prefix `presto` seems to me it could make things confusing.
Is there any though that this could change dynamically (i.e. via session property)?

pdabre12: So, if you take a look at issue https://github.com/prestodb/presto/issues/23385, this is going to be a shared config which is going to be introduced in Presto in a follow-up PR. That's why I think it makes sense to have the prefix `presto` before. 
Let me know what your thoughts are.

pdabre12:
@aditi-pandit Incorporated your suggestions, please have another look.

presto-native-execution/presto_cpp/main/types/PrestoToVeloxExpr.cpp
pdabre12: @aditi-pandit Reverted the changes related to the operators and magic literal functions. These functions aren't reported by Prestissimo. Since, these don't get reported by Prestissimo they won't be present in the namespace specified by the `presto.default-namespace` property for eg. `native.default` namespace.  As the operator types and magic literal functions are always passed down by the Java coordinator, it can be a fair assumption that they will have the `presto.default` namespace prefix.

presto-native-execution/presto_cpp/main/common/Configs.h
czentgr: Ok, then I'm good with the name as it affects both sides, like `presto.version`.

pdabre12:
Closing as completed via https://github.com/prestodb/presto/pull/23358.

Issue #23361, opened 2024-08-01
Timestamps are being recorded as normalized to UTC by Iceberg Connector


tdcmeehan:
[According to the Iceberg spec](https://iceberg.apache.org/spec/#parquet), the `timestamp` type in Iceberg should map to the `TIMESTAMP_MICROS` logical type in Parquet.  `TIMESTAMP_MICROS` is parameterized and takes in two parameters: the unit (here, it is microseconds), and whether or not it is adjusted to UTC.  According to the Iceberg spec, the UTC adjustment must be disabled.  This is because [according to the Parquet spec](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#local-semantics-timestamps-not-normalized-to-utc), timestamps without UTC adjustment represent a local time (i.e., a time that is independent of user's timezone).

In Presto, we are always writing `timestamp` values with UTC adjustment enabled:

```
tdcmeehan@Timothys-MBP ~ % parquet meta /var/folders/q5/h4rx7vn15jxdw3ysr50qsw5w0000gn/T/PrestoTest2735905965489955697/iceberg_data/HIVE/tpch/test_timestamp/data/e9f9fe92-09b6-41dd-8721-448d89e9aa27.parquet 

File path:  /var/folders/q5/h4rx7vn15jxdw3ysr50qsw5w0000gn/T/PrestoTest2735905965489955697/iceberg_data/HIVE/tpch/test_timestamp/data/e9f9fe92-09b6-41dd-8721-448d89e9aa27.parquet
Created by: null
Properties: (none)
Schema:
message table {
  optional int64 x (TIMESTAMP(MICROS,true)) = 1;
}


Row group 0:  count: 1  56.00 B records  start: 4  total(compressed): 56 B total(uncompressed):56 B 
--------------------------------------------------------------------------------
   type      encodings count     avg size   nulls   min / max
x  INT64     G RRD     1         56.00 B    0       "2017-05-01T10:12:34.00000..." / "2017-05-01T10:12:34.00000..."
```

The `true` value in the column `x` `TIMESTAMP(MICROS,true)` means it is normalized to UTC.

This is true regardless if the flag `deprecated.legacy-timestamp` is set to true or false.  This is incorrect, because the Iceberg type represents a timestamp without a time zone, and hence it should not be normalized to any particular time zone.

hantangwangd:
[...]

PR #23388, opened 2024-08-06
Persist proper logical type parameters in parquet files

auden-woolfson:
## Description
When reading/writing parquet files, logical type parameters are not being persisted properly (see issue #23361). This was noticed using the timestamp type with iceberg, where `isAdjustedUTC` was always set to true. The parquet reading and writing functionality was updated, and a test was added to the iceberg connector in order to fix this issue.

Similar issues were noticed in the hive connector in #23454, these required some changes to the parquet schema converter.
```
== RELEASE NOTES ==

Presto Parquet Changes
* Fix bug so that proper logical type parameters are now read and written to Parquet files :pr:`23388`

Presto Iceberg Changes
* Add test for logical type storage in parquet files :pr:`23388`
```


presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergFileWriterFactory.java
```diff
@@ -134,6 +138,24 @@ private IcebergFileWriter createParquetWriter(
         List<String> fileColumnNames = icebergSchema.columns().stream()
                 .map(Types.NestedField::name)
                 .collect(toImmutableList());
+
+        List<Types.NestedField> icebergFields = new ArrayList<>(icebergSchema.columns());
+        ListIterator<Types.NestedField> icebergFieldsIterator = icebergFields.listIterator();
+
+        while (icebergFieldsIterator.hasNext()) {
```
ZacBlanco, line 145: why not just do `icebergSchema.columns().map(...)` and collect into a new list?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergFileWriterFactory.java
```diff
@@ -134,6 +138,24 @@ private IcebergFileWriter createParquetWriter(
...
+            Types.NestedField field = icebergFieldsIterator.next();
+            if (field.type() instanceof Types.TimestampType) {
```
ZacBlanco, line 147: Do we need to consider the case for nested types? e.g. Array, Map, and Struct?

ZacBlanco (changes requested):
Please also add a unit test which verifies the behavior for basic timestamp and types with nested timestamps

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergFileWriterFactory.java
auden-woolfson: Possibly, I'm going to build out the test first then see what the results are. If we do need to consider this we will probably need some kind of dfs style function

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergFileWriterFactory.java
```diff
@@ -131,9 +132,40 @@ private IcebergFileWriter createParquetWriter(
...
+                                Types.fromPrimitiveString(Types.TimestampType.withoutZone().toString()));
+                    }
+
+                    if (field.type().isNestedType()) {
+                        return field;
+                    }
+
+                    return field;})
+                .collect(Collectors.toList());
+        icebergSchema = new Schema(updatedFields);
```
hantangwangd, line 167: As far as I can see, the `icebergSchema` has no need to be translated here. Seems that the information loss of `adjustToUTC` is omitted when flushing the parquet message type to the parquet file's footer, and vice versa reading parquet message type from the parquet file's footer metadata.

Referring to `MessageTypeConverter.toParquetSchema()` and `MetadataReader.readTypeSchema()`.

auden-woolfson: I see your point. I think this issue ultimately boils down to the Apache `OriginalType` class missing any parameter information, and the conversion methods basically just defaulting to `isAdjustedToUTC`.  The `PrimitiveType` class we use in `messageTypeConverter` has a logical type annotation, but there is no way to directly build a `LogicalType` from this. 

Within the Apache library, there exists an interface `LogicalTypeAnnotationVisitor`, and it's inheritor `LogicalTypeConverterVisitor` does exactly what we need here, however it is private to `ParquetMetadataConverter`.

This method can only be used through `toParquetMetadata`, which I'm not sure why we aren't currently making use of. Instead of setting the schema of our file metadata using a custom method that currently isn't working for logical types, we could just convert the entire `FileMetaData` object using the method that Apache already provides us.

The other solution would be to copy the implementation of 'LogicalTypeConverterVisitor' into our message type converter so we can get types from annotations.

TLDR: Apache doesn't expose a method for us to keep going with our custom converter implementation, we can either switch to using their meta data converter of copy some of their protected code.

hantangwangd: Yes, you are right. Our own implementation omitted the logical type annotation. Besides the methods you mentioned above which may introduce a lot of change, another shortcut way maybe that we do special handle for the primitive type with logical type annotation of `TimestampLogicalTypeAnnotation` in `TypeVisitor.visit(primitiveType)`, set logical type for it's result `SchemaElement`. And do the opposite thing in `MetadataReader.readTypeSchema(...)`. This is not a universal method, but it can be proven to be feasible.

auden-woolfson: I tried coding it up this way on a separate branch...

```
            public void visit(PrimitiveType primitiveType)
            {
                SchemaElement element = new SchemaElement(primitiveType.getName());
                element.setRepetition_type(toParquetRepetition(primitiveType.getRepetition()));
                element.setType(getType(primitiveType.getPrimitiveTypeName()));

                if (primitiveType.getOriginalType() != null) {
                    element.setConverted_type(getConvertedType(primitiveType.getOriginalType()));
                    LogicalTypeAnnotation annotation = primitiveType.getLogicalTypeAnnotation();

                    if (annotation instanceof LogicalTypeAnnotation.TimestampLogicalTypeAnnotation) {
                        TimeUnit unit = TimeUnit.MICROS(new MicroSeconds());
                        switch(((LogicalTypeAnnotation.TimestampLogicalTypeAnnotation) annotation).getUnit()) {
                            case NANOS:
                                unit = TimeUnit.NANOS(new NanoSeconds());
                                break;
                            case MILLIS:
                                unit = TimeUnit.MILLIS(new MilliSeconds());
                                break;
                        }

                        element.setLogicalType(LogicalType.TIMESTAMP(
                                new TimestampType(((LogicalTypeAnnotation.TimestampLogicalTypeAnnotation) annotation).isAdjustedToUTC(), unit)));
                    }
                }
```

This could obviously be separated into another method if need be. 

I think this will work but wouldn't it be better to use the method Apache already provides? What if in the future we run into a similar problem with another type? We are maintaining a custom type converter that doesn't have any custom functionality apart from the provided one as far as I can tell.

hantangwangd: > wouldn't it be better to use the method Apache already provides?

Of course, that would be better.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergFileWriter.java
```diff
@@ -0,0 +1,177 @@
...
+                    binder.bind(RowExpressionService.class).toInstance(context.getRowExpressionService());
+                    binder.bind(FilterStatsCalculatorService.class).toInstance(context.getFilterStatsCalculatorService());
+                });
+
+        Injector injector = app
+                .doNotInitializeLogging()
+                .setRequiredConfigurationProperties(ImmutableMap.of())
+                .initialize();
+
+        this.icebergFileWriterFactory = injector.getInstance(IcebergFileWriterFactory.class);
```
auden-woolfson, line 143: I've been trying to figure out the best way to instantiate this class in the tests here but it has been a very convoluted process. Right now I have this injector implementation but it is still giving me some explicit bindings error for my hive metastore. Does anybody else know if there is a better way to go about this?

hantangwangd: Maybe we can build `IcebergFileWriterFactory` directly. I think the following code could work, you can have a try:

```
        Session session = getQueryRunner().getDefaultSession();
        ConnectorSession connectorSession = session.toConnectorSession(new ConnectorId("iceberg"));
        TypeManager typeManager = new TypeManager()
        {
            @Override
            public Type getType(TypeSignature signature)
            {
                for (Type type : getTypes()) {
                    if (signature.getBase().equals(type.getTypeSignature().getBase())) {
                        return type;
                    }
                }
                return null;
            }

            @Override
            public Type getParameterizedType(String baseTypeName, List<TypeSignatureParameter> typeParameters)
            {
                return getType(new TypeSignature(baseTypeName, typeParameters));
            }

            @Override
            public boolean canCoerce(Type actualType, Type expectedType)
            {
                throw new UnsupportedOperationException();
            }

            private List<Type> getTypes()
            {
                return ImmutableList.of(BOOLEAN, INTEGER, BIGINT, DOUBLE, VARCHAR, VARBINARY, TIMESTAMP, DATE, HYPER_LOG_LOG);
            }
        };
        HdfsContext hdfsContext = new HdfsContext(connectorSession);
        HdfsEnvironment hdfsEnvironment = IcebergDistributedTestBase.getHdfsEnvironment();
        IcebergFileWriterFactory icebergFileWriterFactory = new IcebergFileWriterFactory(hdfsEnvironment, typeManager,
                new FileFormatDataSourceStats(), new NodeVersion("test"), new OrcFileWriterConfig(), HiveDwrfEncryptionProvider.NO_ENCRYPTION);

        Path path = new Path(createTempDir().getAbsolutePath());
        Schema schema = toIcebergSchema(ImmutableList.of(new ColumnMetadata("a", VARCHAR), new ColumnMetadata("b", INTEGER), new ColumnMetadata("c", TIMESTAMP)));
        IcebergFileWriter icebergFileWriter = icebergFileWriterFactory.createFileWriter(path, schema, new JobConf(), connectorSession,
                hdfsContext, FileFormat.PARQUET, MetricsConfig.getDefault());

        List<Page> input = rowPagesBuilder(VARCHAR, BIGINT, TIMESTAMP)
                .addSequencePage(100, 0, 0, 123)
                .addSequencePage(100, 100, 100, 223)
                .addSequencePage(100, 200, 200, 323)
                .build();
        for (Page page : input) {
            icebergFileWriter.appendRows(page);
        }
        icebergFileWriter.commit();
```

auden-woolfson: Thanks! I was able to add this into the test. It is currently failing but I'm pretty sure this is due to the reader not being fixed yet. Will work on that this afternoon

presto-parquet/src/main/java/com/facebook/presto/parquet/cache/MetadataReader.java
```diff
@@ -366,17 +373,46 @@ private static void readTypeSchema(Types.GroupBuilder<?> builder, Iterator<Schem
...
+                }
+                if (element.getLogicalType().isSetINTEGER()) {
+                    IntType integer = type.getINTEGER();
+                    typeBuilder.as(LogicalTypeAnnotation.intType(integer.bitWidth, integer.isSigned));
+                }
+                if (element.getLogicalType().isSetTIMESTAMP()) {
+                    TimestampType timestamp = type.getTIMESTAMP();
+                    typeBuilder.as(LogicalTypeAnnotation.timestampType(timestamp.isAdjustedToUTC, convertTimeUnit(timestamp.unit)));
+                }
+            }
```
auden-woolfson, line 397: I only had to include these four because they are the ones that include logical type parameters that need to persist through the conversion process. This is an adaptation of the `getLogicalTypeAnnotation` method from the parquet metadata converter. The actual method does not work since it uses a switch statement on the wrong type, or a type that can't be used for some reason...

steveburnett:
Suggest revising the release note entry for consistency with the [Release Notes Guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines). Maybe something like this: 

```
== RELEASE NOTES ==

Iceberg Connector Changes
* Fix bug so that proper logical type parameters are now read and written to Parquet files :pr:`23388`
```

presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergFileWriter.java
```diff
@@ -0,0 +1,171 @@
...
+    private HdfsContext hdfsContext;
+    private ConnectorSession connectorSession;
+
+    @BeforeClass
+    public void setup()
+    {
+        Session session = getQueryRunner().getDefaultSession();
+        ConnectorId connectorId = new ConnectorId("iceberg");
+        this.connectorSession = session.toConnectorSession(connectorId);
+        TypeManager typeManager = new TypeManager()
```
tdcmeehan, line 84: Please move this into a separate class, like `TestingTypeManager`.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergFileWriter.java
```diff
@@ -0,0 +1,171 @@
...
+            ParquetMetadata parquetMetadata = MetadataReader.readFooter(
+                    dataSource,
+                    parquetFile.length(),
+                    Optional.empty(),
+                    false).getParquetMetadata();
+            MessageType writtenSchema = parquetMetadata.getFileMetaData().getSchema();
+            MessageType originalSchema = convert(icebergSchema, "table");
+            assertEquals(originalSchema, writtenSchema);
+        }
+        catch (IOException e) {
```
tdcmeehan, line 167: Just have the test `throws Exception`

presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergFileWriter.java
```diff
@@ -0,0 +1,171 @@
...
+import static com.facebook.presto.iceberg.FileFormat.PARQUET;
+import static com.facebook.presto.iceberg.IcebergAbstractMetadata.toIcebergSchema;
+import static com.facebook.presto.iceberg.IcebergQueryRunner.createIcebergQueryRunner;
+import static com.google.common.io.Files.createTempDir;
+import static org.apache.iceberg.parquet.ParquetSchemaUtil.convert;
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.fail;
+
+public class TestIcebergFileWriter
+        extends AbstractTestQueryFramework
```
tdcmeehan, line 72: Does this need a `QueryRunner`?  I can't tell where we actually run a full query.

presto-parquet/src/main/java/com/facebook/presto/parquet/cache/MetadataReader.java
tdcmeehan: Can you explain this a bit more?  I don't understand why we need this corresponding reader change.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergFileWriter.java
auden-woolfson: We aren't running a full query anywhere in the test, but the query runner is important for getting a valid connector session. There are a lot of variables that go into creating a connector session that actually works here. Maybe we don't need to extend the abstract class, and we can just instantiate the query runner in the before class method.

tdcmeehan: Could we use `TestingSession` to help?  Here's an example from the codebase:

```java
    public static final Session TEST_SESSION = testSessionBuilder()
            .setCatalog("tpch")
            .setSchema(TINY_SCHEMA_NAME)
            .build();
```

auden-woolfson: I tried that and it causes the test to fail due to missing session properties... I used "iceberg" as catalog. There is a method to add properties, I can look into how to manually build a good test session for the use case but so far adding the same properties as I do in the constructor for the query runner hasn't worked

tdcmeehan: This should work, let me know what failed and we can look into it

presto-parquet/src/main/java/com/facebook/presto/parquet/cache/MetadataReader.java
auden-woolfson: This is because we were losing the logical type information upon reading the parquet schema. The test will not pass without this logic in place. It's fine if we don't need logical type info for storage in certain databases/formats but within the parquet module it is important to at least read it so that choice can be made within other connectors

presto-parquet/src/main/java/com/facebook/presto/parquet/writer/ParquetWriter.java
```diff
@@ -259,11 +262,21 @@ static Slice getFooter(List<RowGroup> rowGroups, MessageType messageType)
...
         fileMetaData.setRow_groups(ImmutableList.copyOf(rowGroups));
 
+        // use parquet metadata converter to get schema with proper logical type info
+        org.apache.parquet.hadoop.metadata.FileMetaData parquetMetaDataInput = new org.apache.parquet.hadoop.metadata.FileMetaData(
+                messageType,
+                ImmutableMap.of(),
+                fileMetaData.getCreated_by());
+
+        ParquetMetadataConverter metadataConverter = new ParquetMetadataConverter();
+        FileMetaData parquetMetaData = metadataConverter.toParquetMetadata(1, new ParquetMetadata(parquetMetaDataInput, ImmutableList.of()));
```
tdcmeehan, line 276: I believe this method is stateless.  If so, can you make `metadataConverter` a constant?

presto-parquet/src/main/java/com/facebook/presto/parquet/writer/ParquetWriter.java
```diff
@@ -259,11 +262,21 @@ static Slice getFooter(List<RowGroup> rowGroups, MessageType messageType)
...
+
+        fileMetaData.setSchema(parquetMetaData.getSchema());
```
tdcmeehan, line 278: Can you extract a helper method to encapsulate the conversion logic to get the Parquet schema?

presto-parquet/src/main/java/com/facebook/presto/parquet/cache/MetadataReader.java
tdcmeehan: Do you feel that there are sufficient tests in place to ensure that the type conversion logic works with the Parquet built in converter?

auden-woolfson: Which converter are you referring to? I'm not sure what tests Apache has in place for their libraries...

tdcmeehan: We've deleted the `MessageTypeConverter`, and I'm just wondering if the unit test coverage is sufficient to prove that using the built in converter from Parquet gives us 1:1 output otherwise.

tdcmeehan: Let's work on some of those tests, the scope of the change makes me nervous that the existing unit test coverage isn't sufficient.

presto-parquet/src/test/java/com/facebook/presto/parquet/writer/TestParquetWriter.java
```diff
@@ -77,6 +86,66 @@ public void testRowGroupFlushInterleavedColumnWriterFallbacks()
         }
     }
 
+    @Test
+    public void testMetadataCreation()
```
tdcmeehan, line 90: @auden-woolfson can you help me understand how these tests run before and after your main code changes?

presto-parquet/src/test/java/com/facebook/presto/parquet/writer/TestParquetWriter.java
```diff
@@ -77,6 +86,66 @@ public void testRowGroupFlushInterleavedColumnWriterFallbacks()
...
+
+            org.apache.parquet.schema.Type booleanType = parquetTypes.get(3);
+            assertEquals(booleanType.asPrimitiveType().getPrimitiveTypeName().name(), "BOOLEAN");
+
+            org.apache.parquet.schema.Type timestampType = parquetTypes.get(4);
+            LogicalTypeAnnotation timeAnnotation = timestampType.getLogicalTypeAnnotation();
+            assertEquals(((LogicalTypeAnnotation.TimestampLogicalTypeAnnotation) timeAnnotation).getUnit(), LogicalTypeAnnotation.TimeUnit.MILLIS);
+            assertFalse(((LogicalTypeAnnotation.TimestampLogicalTypeAnnotation) timeAnnotation).isAdjustedToUTC());
+        }
+        catch (Exception e) {
```
tdcmeehan, line 144: Just let it throw (add `throws Exception` to method signature)

presto-parquet/src/test/java/com/facebook/presto/parquet/writer/TestParquetWriter.java
```diff
@@ -77,6 +86,66 @@ public void testRowGroupFlushInterleavedColumnWriterFallbacks()
...
+    public void testMetadataCreation()
+    {
+        temporaryDirectory = createTempDir();
+        parquetFile = new File(temporaryDirectory, randomUUID().toString() + ".parquet");
+        List<Type> types = ImmutableList.of(
+                BIGINT,
+                INTEGER,
+                VARCHAR,
+                BOOLEAN,
+                TIMESTAMP);
```
tdcmeehan, line 99: Is this comprehensive?  For example, what about doubles?

presto-hive/src/test/java/com/facebook/presto/hive/TestHiveFileFormats.java
```diff
@@ -498,6 +507,55 @@ public void testParquetPageSourceSchemaEvolution(int rowCount)
...
+
+        List<PropertyMetadata<?>> allSessionProperties = getAllSessionProperties(
+                new HiveClientConfig(),
+                new ParquetFileWriterConfig().setParquetOptimizedWriterEnabled(true),
+                createOrcHiveCommonClientConfig(true, 100.0));
+
+        TestingConnectorSession session = new TestingConnectorSession(allSessionProperties);
+        File file = createTempFile("logicaltest", ".parquet");
+
+        file.delete();
```
tdcmeehan, line 523: Why delete it twice?

presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergFileWriter.java
```diff
@@ -0,0 +1,181 @@
...
+import static org.apache.iceberg.parquet.ParquetSchemaUtil.convert;
+import static org.testng.Assert.assertEquals;
+
+public class TestIcebergFileWriter
+{
+    private IcebergFileWriterFactory icebergFileWriterFactory;
+    private HdfsContext hdfsContext;
+    private ConnectorSession connectorSession;
+
+    private static class TestingTypeManager
```
tdcmeehan, line 77: Move this to the bottom of this file please

presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergFileWriter.java
```diff
@@ -0,0 +1,181 @@
...
+                .setCatalog(ICEBERG_CATALOG)
+                .setSchema("tpch")
+                .build();
+
+        this.connectorSession = session.toConnectorSession(connectorId);
+        TypeManager typeManager = new TestingTypeManager();
+        this.hdfsContext = new HdfsContext(connectorSession);
+        HdfsEnvironment hdfsEnvironment = IcebergDistributedTestBase.getHdfsEnvironment();
+        this.icebergFileWriterFactory = new IcebergFileWriterFactory(hdfsEnvironment, typeManager,
+                new FileFormatDataSourceStats(), new NodeVersion("test"), new OrcFileWriterConfig(), HiveDwrfEncryptionProvider.NO_ENCRYPTION);
```
tdcmeehan, line 149: Why do we have Orc config settings here when we're suing Parquet?

tdcmeehan (changes requested):

presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergFileWriter.java
auden-woolfson: The iceberg file writer factory requires orc settings to be instantiated

presto-parquet/src/test/java/com/facebook/presto/parquet/writer/TestParquetWriter.java
```diff
@@ -77,6 +89,75 @@ public void testRowGroupFlushInterleavedColumnWriterFallbacks()
...
     }
 
+    @Test
+    public void testMetadataCreation()
+            throws Exception
+    {
+        temporaryDirectory = createTempDir();
+        parquetFile = new File(temporaryDirectory, randomUUID().toString() + ".parquet");
+        List<Type> types = ImmutableList.of(
+                BIGINT,
```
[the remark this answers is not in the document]
auden-woolfson: Some of these types, like UUID, are unsupported for parquet writers still. Supporting them may be out of scope for this issue... maybe we can include all the supported types and create a new issue for the parquet writer to support the remaining ones

presto-parquet/src/main/java/com/facebook/presto/parquet/cache/MetadataReader.java
ZacBlanco: I don't think we should special-case for only certain types when reading logical types. If a logical type is set, we should call a method which does the logical type conversion, and then sets the logical type in the descriptor. If the logical type isn't known during conversion, then throw an exception.

See here for an example: https://github.com/trinodb/trino/blob/0d5576a67a855bf95123acaf3ea836d1963861ef/lib/trino-parquet/src/main/java/io/trino/parquet/reader/MetadataReader.java#L221-L227

auden-woolfson: I don't think this is necessary, the logical type annotations are already of the proper type.  I just added the date type which isn't included in the logic here, to `testWriteParquetFileWithLogicalTypes` and it is still passing. The only purpose of this logic is to make sure that the parameters are correct, so it only needs to check the logical types that take parameters. Adding conditions for each logical type would be overkill, and confusing since most of the conditions wouldn't actually result in any additional functionality. 

That being said, I can extract it into a method named something like `getLogicalTypeAnnotationWithParameters` to make the intentions a little more clear here.

ZacBlanco: For completeness' sake, I do think that we should exhaustively convert all logical types so we don't have to touch this code afterwards. I agree it doesn't actually make much difference except for parameterized types or other logical types which are stored in the same physical format (e.g. long decimal and UUID). I would prefer to do the exhaustive conversion.

On another note, is this the error you encountered with the other approach?

```
java.lang.NoSuchMethodError: org.apache.parquet.format.LogicalType.getSetField()Lcom/facebook/presto/hive/$internal/parquet/org/apache/thrift/TFieldIdEnum;
	at org.apache.parquet.format.converter.ParquetMetadataConverter.getLogicalTypeAnnotation(ParquetMetadataConverter.java:1084)
	at org.apache.parquet.format.converter.ParquetMetadataConverter.buildChildren(ParquetMetadataConverter.java:1715)
	at org.apache.parquet.format.converter.ParquetMetadataConverter.fromParquetSchema(ParquetMetadataConverter.java:1670)
	at org.apache.parquet.format.converter.ParquetMetadataConverter.fromParquetMetadata(ParquetMetadataConverter.java:1526)
	at org.apache.parquet.format.converter.ParquetMetadataConverter.readParquetMetadata(ParquetMetadataConverter.java:1490)
	at org.apache.parquet.hadoop.ParquetFileReader.readFooter(ParquetFileReader.java:591)
	at org.apache.parquet.hadoop.ParquetFileReader.<init>(ParquetFileReader.java:799)
	at org.apache.parquet.hadoop.ParquetFileReader.open(ParquetFileReader.java:654)
	at org.apache.iceberg.parquet.ParquetUtil.fileMetrics(ParquetUtil.java:80)
	at org.apache.iceberg.parquet.ParquetUtil.fileMetrics(ParquetUtil.java:75)
	at com.facebook.presto.iceberg.IcebergParquetFileWriter.lambda$getMetrics$0(IcebergParquetFileWriter.java:77)
	at com.facebook.presto.hive.authentication.NoHdfsAuthentication.doAs(NoHdfsAuthentication.java:23)
```

ZacBlanco (changes requested):
My main recommendation is to parameterize the unit test for each type so it's clear for which type(s), if any fail

presto-parquet/src/main/java/com/facebook/presto/parquet/cache/MetadataReader.java
auden-woolfson: Yes I believe that was the error I got, the `getSetField` method does not exist. 

I can update it to an exhaustive list.

hantangwangd:
I noticed there are CI test failures on iceberg for long, and have investigated the problem for a while.

In the test environment, if we set the logicalType for a primary type through `schemaElement.setLogicalType(logicalType)`, then an exception as follows will be thrown when `ParquetMetadataConverter.getLogicalTypeAnnotation(logicalType)` is called to get the corresponding annotation of the logical type:

```
java.lang.NoSuchMethodError: org.apache.parquet.format.LogicalType.getSetField()Lcom/facebook/presto/hive/$internal/parquet/org/apache/thrift/TFieldIdEnum;
```

But I do not meet this problem in the formal environment. By tracking the classes loaded at their respective runtimes, I found that for `LogicalType`, the class file loaded for it in the formal environment is from `hive-apache-3.0.0-10.jar`:

```
[Loaded org.apache.parquet.format.LogicalType$1 from file:/Users/wangd/.m2/repository/com/facebook/presto/hive/hive-apache/3.0.0-10/hive-apache-3.0.0-10.jar]
```

While in the test environment, the class loaded is from `parquet-format-structures-11.3.1.jar`:

```
[Loaded org.apache.parquet.format.LogicalType$1 from file:/Users/wangd/.m2/repository/org/apache/parquet/parquet-format-structures/1.13.1/parquet-format-structures-1.13.1.jar]
```

I guess this may be the most direct reason, because `ParquetMetadataConverter` is defined only in `hive-apache-3.0.0-10.jar`, so it should uses `LogicalType` defined in `hive-apache-3.0.0-10.jar` which may differs with the one in `parquet-format-structures-11.3.1.jar` considering their method signatures.

This may involve some deeper issues related to Maven dependencies, and I am unsure how to handle it. Perhaps this is also the reason why trino has implemented its own version of `ParquetMetadataConverter`.

presto-parquet/src/test/java/com/facebook/presto/parquet/writer/TestParquetWriter.java
```diff
@@ -109,27 +114,60 @@ public void testRowGroupFlushInterleavedColumnWriterFallbacks()
...
+                        Optional.of("FIXED_LEN_BYTE_ARRAY"),
+                        Optional.of("INT32"),
+                        Optional.of("INT64"),
+                        Optional.empty(),
+                        Optional.empty())
+        }};
+    }
+
+    @Test(dataProvider = "testMetadataCreation")
+    public void testMetadataCreation(List<Type> types, List<Optional<Class<?>>> annotationTypes, List<Optional<String>> primitiveNames)
```
ZacBlanco, line 166: Rather than accepting a list of inputs, the signature should accept a single value for each type
```suggestion
    public void testMetadataCreation(Type type, Optional<Class<?>> annotationType, Optional<String> primitiveName)
```

presto-parquet/src/test/java/com/facebook/presto/parquet/writer/TestParquetWriter.java
```diff
@@ -14,36 +14,69 @@
...
+import static org.testng.Assert.assertThrows;
+import static org.testng.Assert.assertTrue;
 import static org.testng.Assert.fail;
 
 public class TestParquetWriter
 {
     private File temporaryDirectory;
     private File parquetFile;
 
+    private static final DecimalType SHORT_DECIMAL = DecimalType.createDecimalType(18);
```
ZacBlanco, line 71: I believe we use different types for different-sized decimals. Can we have cases for

1. P <= 9 (Int32 storage)
2. 10 <= P <= 18 (Int64 storage)

ZacBlanco (commented):
PR looks good to me now. One more request for a test case. Once other test failures everything else LGTM

presto-parquet/src/test/java/com/facebook/presto/parquet/writer/TestParquetWriter.java
auden-woolfson: Done, thanks for the feedback!

presto-common/src/main/java/com/facebook/presto/common/type/LongDecimalType.java
```diff
@@ -29,7 +29,7 @@
 import static com.facebook.presto.common.type.UnscaledDecimal128Arithmetic.compare;
 import static io.airlift.slice.SizeOf.SIZE_OF_LONG;
 
-final class LongDecimalType
+public final class LongDecimalType
```
tdcmeehan, line 32: Why is this needed?

tdcmeehan (commented):
LGTM % nit about visibility of `LongDecimalType`

presto-common/src/main/java/com/facebook/presto/common/type/LongDecimalType.java
auden-woolfson: It's not anymore, I initially added it to check which decimal type we had in test metadata creation. Good catch

tdcmeehan:
Please squash commits and we'll be good to merge.

ZacBlanco (approved):
LGTM, but you'll need to squash the commits again

hantangwangd (dismissed):
Great, thanks for fixing this maven dependency issue @auden-woolfson. The whole change looks good to me.

auden-woolfson:
Great, thanks for all the help + reviews @ZacBlanco @tdcmeehan @hantangwangd, rebased and squashed

tdcmeehan (approved):

ZacBlanco (approved):

hantangwangd (approved):

tdcmeehan (approved):

PR #23390, opened 2024-08-06
[native] Add new shared arbitrator configs

tanjialiang:
Adding new arbitrator configs to presto native, in preparation of utilizing the new config mechanism for memory arbitrator, and deprecating the old ones.
```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/PrestoServer.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoServer.cpp
@@ -15,7 +15,6 @@
 #include <boost/asio/io_service.hpp>
 #include <boost/asio/ip/host_name.hpp>
 #include <boost/asio/ip/tcp.hpp>
-#include <boost/lexical_cast.hpp>
 #include <glog/logging.h>
 #include "CoordinatorDiscoverer.h"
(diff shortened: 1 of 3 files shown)


xiaoxmeng (approved):
@tanjialiang LGTM. Thanks!

steveburnett:
Suggest documenting these new native configs in https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/presto_cpp/properties.rst.

tanjialiang:
> Suggest documenting these new native configs in https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/presto_cpp/properties.rst.

Good point, I'll add the missing documentations there.

tanjialiang:
> Suggest documenting these new native configs in https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/presto_cpp/properties.rst.

These configs are currently not in use. So adding the documentations in this PR would not make sense. I will add them in the followup PR in which I remove the old configs and make the switch to the new ones.

xiaoxmeng (approved):

PR #19236, opened 2023-03-23
Remove Raptor codebase.

amitkdutta:
Raptor was a meta internal system used for fast analytics. The system is deprecated and shut down with many enhancement made in Presto (e.g. caching).  This PR removes raptor directory.

```
== RELEASE NOTES ==

General Changes
* Remove Raptor module

```

PR #23393, opened 2024-08-07
[native] Disable cache retention with NO_PREFERENCE node selection

zacw7:
```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/QueryContextManager.cpp
+++ b/presto-native-execution/presto_cpp/main/QueryContextManager.cpp
@@ -17,6 +17,7 @@
 #include "presto_cpp/main/common/Configs.h"
 #include "presto_cpp/main/common/Counters.h"
 #include "velox/common/base/StatsReporter.h"
+#include "velox/connectors/hive/HiveConfig.h"
 #include "velox/core/QueryConfig.h"
 #include "velox/type/tz/TimeZoneMap.h"
 
@@ -142,6 +143,25 @@ void updateVeloxConfigs(
         core::QueryConfig::kDriverCpuTimeSliceLimitMs, "1000");
   }
 }
+
+void updateVeloxConnectorConfigs(
+    std::unordered_map<
+        std::string,
+        std::unordered_map<std::string, std::string>>& connectorConfigStrings) {
+  const auto& systemConfig = SystemConfig::instance();
+
+  for (auto& entry : connectorConfigStrings) {
+    auto connectorConfig = entry.second;
+
+    // Do not retain cache if `node_selection_strategy` is explicitly set to
+    // `NO_PREFERENCE`.
+    auto it = connectorConfig.find("node_selection_strategy");
+    if (it == connectorConfig.end() || it->second == "NO_PREFERENCE") {
+      connectorConfig.emplace(
+          connector::hive::HiveConfig::kCacheNoRetentionSession, "true");
+    }
+  }
+}
 } // namespace
 
 QueryContextManager::QueryContextManager(
@@ -172,6 +192,7 @@ std::shared_ptr<core::QueryCtx> QueryContextManager::findOrCreateQueryCtx(
   }
 
   updateVeloxConfigs(configStrings);
+  updateVeloxConnectorConfigs(connectorConfigStrings);
 
   std::unordered_map<std::string, std::shared_ptr<Config>> connectorConfigs;
   for (auto& entry : connectorConfigStrings) {


presto-native-execution/presto_cpp/main/QueryContextManager.cpp
```diff
@@ -141,6 +145,15 @@ void updateVeloxConfigs(
     configStrings.emplace(
         core::QueryConfig::kDriverCpuTimeSliceLimitMs, "1000");
   }
+
+  for (auto& entry : connectorConfigStrings) {
+    auto connectorConfig = entry.second;
+    it = connectorConfig.find("node_selection_strategy");
+    if (it == connectorConfig.end() || it->second != "SOFT_AFFINITY") {
+      connectorConfig.emplace(
```
xiaoxmeng, line 153: The default value for Presto java is true? Thanks!

presto-native-execution/presto_cpp/main/QueryContextManager.cpp
```diff
@@ -141,6 +145,15 @@ void updateVeloxConfigs(
     configStrings.emplace(
         core::QueryConfig::kDriverCpuTimeSliceLimitMs, "1000");
   }
+
+  for (auto& entry : connectorConfigStrings) {
```
xiaoxmeng, line 149: Let's have a separate method updateConnectorConfigs?

presto-native-execution/presto_cpp/main/QueryContextManager.cpp
zacw7: In Java, there are three node selection strategy:  _HARD_AFFINITY_, _SOFT_AFFINITY_ and _NO_PREFERENCE_. _NO_PREFERENCE_ is the default value.

Code pointers shared by @kewang1024 : Splits are set as `cacheable` only when the strategy is _HARD_AFFINITY_ or _SOFT_AFFINITY_.
https://github.com/prestodb/presto/blob/6e4896ddc4ed078cce1f81186e6986ecca94a141/presto-main/src/main/java/com/facebook/presto/execution/scheduler/nodeSelection/SimpleNodeSelector.java#L156-L167

And _HARD_AFFINITY_ is not used anymore, so we expect cache is enabled only when soft affinity is set and the default is without cache.

steveburnett: > And _HARD_AFFINITY_ is not used anymore, so we expect cache is enabled only when soft affinity is set and the default is without cache.

If _HARD_AFFINITY_ is not used anymore, should it be removed from the Presto doc in https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/develop/connectors.rst#connectorsplit ?

kewang1024: HARD_AFFINITY is designed for Raptor, which is not used anymore by Meta.
But open source could still use it, so we can not remove it.

tdcmeehan: Raptor was deleted last year by #19236

steveburnett: Thanks @kewang1024 and @tdcmeehan. If Raptor was removed from OSS Presto over a year ago, is there any other use of _HARD_AFFINITY_ in current Presto that it should stay in the current doc?

kewang1024:
image
It might not be ideal, but it would be least destructive to the current workflow

presto-native-execution/presto_cpp/main/QueryContextManager.cpp
kewang1024: I found other `HARD_AFFINITY` usage for example `LocalFileSplit`

steveburnett: > I found other `HARD_AFFINITY` usage for example `LocalFileSplit`

Thanks @kewang1024! In that case I think it should stay in the current doc.

zacw7:
> One issue I can think of is if the config "node-selection-strategy" is set to "SOFT_AFFINITY" and no session property. Then on prestissimo side, it would be translated to no cache retention?

You are right, I'm surprised that Velox doesn't consider Presto system config properly. This is tricky.

presto-native-execution/presto_cpp/main/QueryContextManager.cpp
xiaoxmeng: @zacw7 then change to it->second == "NO_PREFERENCE"? Also make sure it doesn't break existing meta native use case with SSD cache enabled? thanks!

xiaoxmeng (commented):
@zacw7 LGTM % comment. Thanks!

xiaoxmeng (approved):

PR #23405, opened 2024-08-08
Refactor to better separate queueing and execution phase 

prithvip:
## Description
This change refactors some classes to separate queueing and execution phase in the code organization. It is the first set of changes. 

## Motivation and Context
This is the first set of refactoring to support RFC5 https://github.com/prestodb/rfcs/pull/23. However, this refactor stands on its own for improving code organization, and is valuable regardless of the RFC. 

## Impact
No impact 

## Test Plan
mvn test + verifier run, this is a straightforward refactor 

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/dispatcher/FailedDispatchQueryFactory.java
+++ b/presto-main/src/main/java/com/facebook/presto/dispatcher/FailedDispatchQueryFactory.java
@@ -14,50 +14,25 @@
 package com.facebook.presto.dispatcher;
 
 import com.facebook.presto.Session;
-import com.facebook.presto.event.QueryMonitor;
-import com.facebook.presto.execution.ExecutionFailureInfo;
-import com.facebook.presto.execution.LocationFactory;
-import com.facebook.presto.server.BasicQueryInfo;
 import com.facebook.presto.spi.resourceGroups.ResourceGroupId;
 
-import javax.inject.Inject;
-
 import java.util.Optional;
-import java.util.concurrent.ExecutorService;
 
-import static com.facebook.presto.util.Failures.toFailure;
-import static java.util.Objects.requireNonNull;
-
-public class FailedDispatchQueryFactory
+/**
+ * Factory interface to create FailedDispatchQuery
+ *
+ * This interface is required for https://github.com/prestodb/presto/issues/23455
+ */
+public interface FailedDispatchQueryFactory
 {
-    private final QueryMonitor queryMonitor;
-    private final LocationFactory locationFactory;
-    private final ExecutorService executor;
-
-    @Inject
-    public FailedDispatchQueryFactory(QueryMonitor queryMonitor, LocationFactory locationFactory, DispatchExecutor dispatchExecutor)
-    {
-        this.queryMonitor = requireNonNull(queryMonitor, "queryMonitor is null");
-        this.locationFactory = requireNonNull(locationFactory, "locationFactory is null");
-        this.executor = requireNonNull(dispatchExecutor, "dispatchExecutor is null").getExecutor();
-    }
-
-    public FailedDispatchQuery createFailedDispatchQuery(Session session, String query, Optional<ResourceGroupId> resourceGroup, Throwable throwable)
-    {
-        ExecutionFailureInfo failure = toFailure(throwable);
-        FailedDispatchQuery failedDispatchQuery = new FailedDispatchQuery(
-                session,
-                query,
-                locationFactory.createQueryLocation(session.getQueryId()),
-                resourceGroup,
-                failure,
-                executor);
-
-        BasicQueryInfo queryInfo = failedDispatchQuery.getBasicQueryInfo();
-
-        queryMonitor.queryCreatedEvent(queryInfo);
-        queryMonitor.queryImmediateFailureEvent(queryInfo, failure);
-
-        return failedDispatchQuery;
-    }
+    /**
+     * If DispatchManager fails to create a DispatchQuery, a FailedDispatchQuery is created instead
+     *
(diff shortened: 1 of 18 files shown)


presto-main/src/main/java/com/facebook/presto/dispatcher/FailedDispatchQueryFactory.java
```diff
@@ -14,50 +14,11 @@
...
-import javax.inject.Inject;
-
 import java.util.Optional;
-import java.util.concurrent.ExecutorService;
-
-import static com.facebook.presto.util.Failures.toFailure;
-import static java.util.Objects.requireNonNull;
 
-public class FailedDispatchQueryFactory
+public interface FailedDispatchQueryFactory
```
rschlussel, line 21: can you explain why you extract an interface here?

presto-main/src/main/java/com/facebook/presto/server/protocol/ExecutingQueryResponseProvider.java
```diff
@@ -0,0 +1,41 @@
...
+import com.google.common.util.concurrent.ListenableFuture;
+import io.airlift.units.DataSize;
+import io.airlift.units.Duration;
+
+import javax.ws.rs.core.Response;
+import javax.ws.rs.core.UriInfo;
+
+import java.util.Optional;
+
+public interface ExecutingQueryResponseProvider
```
rschlussel, line 27: can you add Javadoc for this interface?

presto-main/src/main/java/com/facebook/presto/dispatcher/FailedDispatchQueryFactory.java
rschlussel: also add javadoc please (even though it's not new)

rschlussel (commented):
small comments. generally looks good.

presto-main/src/main/java/com/facebook/presto/dispatcher/FailedDispatchQueryFactory.java
prithvip: From high-level point of view, DispatchQuery is created from the DispatchQueryFactory, of which there can be two implementations, LocalDispatchQueryFactory which creates a LocalDispatchQuery, and a possible RemoteDispatchQueryFactory, which could create a RemoteDispatchQuery. Analogously, there could be two implementations to create a FailedDispatchQuery, and that is what this change does. 

From a code organization point of view, FailedLocalDispatchQueryFactory takes a dependency on QueryMonitor, which brings in many heavy weight classes, such as FunctionAndTypeManager, HistoryBasedPlanStatisticsTracker, and more, which in turn, bring in a laundry list of more dependencies. An alternative approach to extracting this interface, could be to split QueryMonitor into a DispatchQueryMonitor, which would have queryCreated and queryImmediateFailure methods, and a ExecutingQueryMonitor, which would have the rest of the methods (like queryCompletedEvent). This way, dependencies such as FunctionAndTypeManager, would not need to be wired into DispatchQueryMonitor. But this would be a much larger refactor, and I felt it cleaner to simply extract the interface from FailedDispatchQueryFactory itself, and allow for multiple implementations, especially since the underlying implementation is not complex, and there would not be much benefit from code re-use.

prithvip:
@rschlussel Added javadocs for ExecutingStatementResponse and FailedDispatchQueryFactory

rschlussel (dismissed):

tdcmeehan (dismissed):
My comment below seems to apply to the whole PR.  Is the idea that these interfaces will be extended into some other, unknown codebase?  If so, please let's settle on a design in an RFC first, as I think this would need to be considered carefully (we don't want to create another brittle dependency between open source and internal code).

presto-main/src/main/java/com/facebook/presto/dispatcher/FailedDispatchQueryFactory.java
rschlussel: Ultimately, the idea is to have other implementations, but the goal for this PR is not really about the specific interfaces, but about removing dependencies in queueing/dispatching code on classes not needed for queueing/dispatching, so it's fine if the interfaces need to change as the RFC is finalized. The end goal that this is working towards is that it should be possible to move queueing stuff out of Presto main to a separate package. There's certainly overlap between determining the right interfaces and figuring out how best to separate out dependencies, but since disentangling the dependencies is itself a substantial undertaking, and most of the work would be needed even if interfaces change, we want to get started with that process now.

prithvip:
> My comment below seems to apply to the whole PR. Is the idea that these interfaces will be extended into some other, unknown codebase? If so, please let's settle on a design in an RFC first, as I think this would need to be considered carefully (we don't want to create another brittle dependency between open source and internal code).

@tdcmeehan @rschlussel Yes, I agree we don't want to create a brittle dependency between internal code and open source code, since someone could change open source code and inadvertently break the internal code, as we have seen happen many times in the past. This refactor is the first step of separating the presto-main module into a presto-dispatcher module and a presto-main module. Right now the classes in presto-main are quite entangled in each other, and this is a step in the disentangling. I do believe that this disentangling stands on its own for better code organization, regardless of the RFC. As an example, there is no reason why code related to the queueing phase needs to have access to the analyzer, since that is only relevant for planning phase. Hence the separation of QueryPreparer and QueryAnalyzer. The other commits in this PR are along those same lines.

prithvip:
This PR is first PR for https://github.com/prestodb/presto/issues/23455

rschlussel (approved):

PR #23407, opened 2024-08-08
Accept query id and slug in QueuedStatement

prithvip:
## Description
This change adds support for QueuedStatementResource to accept a pre-minted query id and slug. 

## Motivation and Context
This is required for RFC5 https://github.com/prestodb/rfcs/pull/23 

## Impact
There is no user-facing or performance impact. 

## Test Plan
Added test in TestServer

## Contributor checklist

- [ ] CI passed.

## Release Notes
```
== RELEASE NOTES ==

General Changes
* Add support in QueuedStatement protocol to accept pre-minted query id and slug :pr:`23407`
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/server/protocol/QueuedStatementResource.java
+++ b/presto-main/src/main/java/com/facebook/presto/server/protocol/QueuedStatementResource.java
@@ -48,6 +48,7 @@ import javax.ws.rs.DefaultValue;
 import javax.ws.rs.GET;
 import javax.ws.rs.HeaderParam;
 import javax.ws.rs.POST;
+import javax.ws.rs.PUT;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
@@ -99,6 +100,7 @@ import static java.util.concurrent.TimeUnit.SECONDS;
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
 import static javax.ws.rs.core.MediaType.TEXT_PLAIN_TYPE;
 import static javax.ws.rs.core.Response.Status.BAD_REQUEST;
+import static javax.ws.rs.core.Response.Status.CONFLICT;
 import static javax.ws.rs.core.Response.Status.NOT_FOUND;
 
 @Path("/")
@@ -218,6 +220,58 @@ public class QueuedStatementResource
                 Optional.of(sessionPropertyManager));
         Query query = new Query(statement, sessionContext, dispatchManager, executingQueryResponseProvider, 0);
         queries.put(query.getQueryId(), query);
+
+        return withCompressionConfiguration(Response.ok(query.getInitialQueryResults(uriInfo, xForwardedProto, xPrestoPrefixUrl, binaryResults)), compressionEnabled).build();
+    }
+
+    /**
+     * HTTP endpoint for submitting queries to the Presto Coordinator.
+     * Presto performs lazy execution. The submission of a query returns
+     * a placeholder for the result set, but the query gets
+     * scheduled/dispatched only when the client polls for results.
+     * This endpoint accepts a pre-minted queryId and slug, instead of
+     * generating it.
+     *
+     * @param statement The statement or sql query string submitted
+     * @param queryId Pre-minted query ID to associate with this query
+     * @param slug Pre-minted slug to protect this query
+     * @param xForwardedProto Forwarded protocol (http or https)
+     * @param servletRequest The http request
+     * @param uriInfo {@link javax.ws.rs.core.UriInfo}
+     * @return {@link javax.ws.rs.core.Response} HTTP response code
+     */
+    @PUT
+    @Path("/v1/statement/{queryId}")
+    @Produces(APPLICATION_JSON)
+    public Response putStatement(
+            String statement,
+            @PathParam("queryId") QueryId queryId,
+            @QueryParam("slug") String slug,
+            @DefaultValue("false") @QueryParam("binaryResults") boolean binaryResults,
+            @HeaderParam(X_FORWARDED_PROTO) String xForwardedProto,
+            @HeaderParam(PRESTO_PREFIX_URL) String xPrestoPrefixUrl,
+            @Context HttpServletRequest servletRequest,
+            @Context UriInfo uriInfo)
+    {
+        if (isNullOrEmpty(statement)) {
+            throw badRequest(BAD_REQUEST, "SQL statement is empty");
+        }
+
+        abortIfPrefixUrlInvalid(xPrestoPrefixUrl);
+
+        // TODO: For future cases we may want to start tracing from client. Then continuation of tracing
+        //       will be needed instead of creating a new trace here.
+        SessionContext sessionContext = new HttpRequestSessionContext(
+                servletRequest,
+                sqlParserOptions,
+                tracerProviderManager.getTracerProvider(),
+                Optional.of(sessionPropertyManager));
+        Query attemptedQuery = new Query(statement, sessionContext, dispatchManager, executingQueryResponseProvider, 0, queryId, slug);
+        Query query = queries.computeIfAbsent(queryId, unused -> attemptedQuery);
+
+        if (attemptedQuery != query && !attemptedQuery.getSlug().equals(query.getSlug()) || query.getLastToken() != 0) {
+            throw badRequest(CONFLICT, "Query already exists");
+        }
 
         return withCompressionConfiguration(Response.ok(query.getInitialQueryResults(uriInfo, xForwardedProto, xPrestoPrefixUrl, binaryResults)), compressionEnabled).build();
     }
@@ -386,7 +440,7 @@ public class QueuedStatementResource
         private final DispatchManager dispatchManager;
         private final ExecutingQueryResponseProvider executingQueryResponseProvider;
         private final QueryId queryId;
-        private final String slug = "x" + randomUUID().toString().toLowerCase(ENGLISH).replace("-", "");
+        private final String slug;
         private final AtomicLong lastToken = new AtomicLong();
         private final int retryCount;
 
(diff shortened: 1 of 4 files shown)


rschlussel (approved):
Add a release note about the protocol change.

presto-main/src/main/java/com/facebook/presto/server/protocol/QueuedStatementResource.java
```diff
@@ -227,6 +227,53 @@ public Response postStatement(
...
+     * @return {@link javax.ws.rs.core.Response} HTTP response code
+     */
+    @POST
+    @Path("/v1/statement/queued/{queryId}")
+    @Produces(APPLICATION_JSON)
+    public Response postStatement(
+            String statement,
+            @PathParam("queryId") QueryId queryId,
+            @QueryParam("slug") String slug,
+            @DefaultValue("false") @QueryParam("binaryResults") boolean binaryResults,
```
tdcmeehan, line 252: This was added for fuzzer testing, and no client presently uses this.  I'd remove it.

presto-main/src/main/java/com/facebook/presto/server/protocol/QueuedStatementResource.java
```diff
@@ -227,6 +227,53 @@ public Response postStatement(
...
+
+        // TODO: For future cases we may want to start tracing from client. Then continuation of tracing
+        //       will be needed instead of creating a new trace here.
+        SessionContext sessionContext = new HttpRequestSessionContext(
+                servletRequest,
+                sqlParserOptions,
+                tracerProviderManager.getTracerProvider(),
+                Optional.of(sessionPropertyManager));
+        Query query = new Query(statement, sessionContext, dispatchManager, queryResultsProvider, 0, queryId, slug);
+        queries.put(query.getQueryId(), query);
```
tdcmeehan, line 272: The best thing to do, to preserve idempotency while also ensuring that retries work, would be to check if the query already exists.  If it does, check if the slug matches.  If it does, then ensure that the token should be `0`, otherwise, fail.

If the query doesn't already exist, then proceed as here.

tdcmeehan (changes requested):
I think we can add some documentation here in `statement.rst`.

presto-main/src/main/java/com/facebook/presto/server/protocol/QueuedStatementResource.java
```diff
@@ -227,6 +227,53 @@ public Response postStatement(
...
     }
 
+    /**
+     * HTTP endpoint for submitting queries to the Presto Coordinator.
+     * Presto performs lazy execution. The submission of a query returns
+     * a placeholder for the result set, but the query gets
+     * scheduled/dispatched only when the client polls for results.
+     * This endpoint accepts a pre-minted queryId and slug, instead of
+     * generating it.
+     * @param statement The statement or sql query string submitted
```
elharo, line 237: nit: blank line before first param

presto-main/src/main/java/com/facebook/presto/server/protocol/QueuedStatementResource.java
tdcmeehan: I just remembered that `POST` is not an idempotent HTTP verb.  Idempotency is a nice attribute, and I'd recommend using `PUT` instead. (The other endpoint doesn't care about idempotency, because you can call `POST` multiple times for the same query, but the only one that matters is the first one whose `nextUri` is followed.  Not so here--creating multiple queries with the same identical query ID is impossible.)

I still recommend the improvement above, because I think the API should guard against adversarial usage, even if it's trivial or unlikely.  For example, `POST`ing an already executing query here would at best cause an error, at worst could be a security issue.

presto-main/src/main/java/com/facebook/presto/server/protocol/QueuedStatementResource.java
```diff
@@ -227,6 +227,53 @@ public Response postStatement(
...
+     * @param queryId Pre-minted query ID to associate with this query
+     * @param slug Pre-minted slug to protect this query
+     * @param xForwardedProto Forwarded protocol (http or https)
+     * @param servletRequest The http request
+     * @param uriInfo {@link javax.ws.rs.core.UriInfo}
+     * @return {@link javax.ws.rs.core.Response} HTTP response code
+     */
+    @POST
+    @Path("/v1/statement/queued/{queryId}")
```
tdcmeehan, line 246:
```suggestion
    @Path("/v1/statement/{queryId}")
```

Can we make this consistent with the `POST`?  I don't think the `queued` signifies any additional information.

ZacBlanco (changes requested):
It would be great to have this added to the OpenAPI spec we're working on getting into the docs: 

https://github.com/prestodb/presto/blob/master/presto-openapi/src/main/resources/queued_statement.yaml

prithvip:
> It would be great to have this added to the OpenAPI spec we're working on getting into the docs:
> 
> https://github.com/prestodb/presto/blob/master/presto-openapi/src/main/resources/queued_statement.yaml

@ZacBlanco Is there a README for how to add and test this?

ZacBlanco:
We don't currently have a readme but it wouldn't hurt to have one. I'll try to submit a PR for that soon with instructions for adding new paths and components.

For the time being, I would look at the existing endpoints in the spec and try mimicking them for your new endpoint. You can refer the official spec for the full set of options: https://spec.openapis.org/oas/latest.html

Once you've added a new path to the YAML you can build the openapi module with .`/mvnw -pl presto-openapi` and check the generated `index.html` in the `target/` directory has your new endpoint that would be enough. We don't have tests but the build will error if your new additions to the YAML don't align with the OpenAPI spec.

prithvip:
@tdcmeehan Addressed all comments: 
1. Changed verb from POST to PUT 
2. Added idempotency checks and tests for same 
3. Added documentation to statement.rst 

@ZacBlanco 
Added docs to OpenAPI

presto-main/src/main/java/com/facebook/presto/server/protocol/QueuedStatementResource.java
prithvip: I'm not sure that no client is using it, and I kept it here since the POST supports it. If we want to remove support for this, we should probably raise an issue first and remove it as part of a separate PR.

steveburnett (dismissed):
LGTM! (docs)

Pull branch, local docs build, looks good. Thanks!

presto-main/src/main/java/com/facebook/presto/server/protocol/QueuedStatementResource.java
prithvip: I asked around and it looks like we are actually using this flag for our client, so we should keep it.

tdcmeehan: Can you share more details on the client that's using it?  Is it an internal client?

presto-main/src/main/java/com/facebook/presto/server/protocol/QueuedStatementResource.java
```diff
@@ -222,6 +224,61 @@ public Response postStatement(
...
+        }
+
+        if (queries.containsKey(queryId)) {
+            Query query = queries.get(queryId);
+            if (!query.getSlug().equals(slug) || query.getLastToken() != 0) {
+                throw badRequest(CONFLICT, "Query already exists");
+            }
+        }
+
+        abortIfPrefixUrlInvalid(xPrestoPrefixUrl);
```
tdcmeehan, line 267:
```suggestion
        abortIfPrefixUrlInvalid(xPrestoPrefixUrl);
        
        if (queries.containsKey(queryId)) {
            Query query = queries.get(queryId);
            if (!query.getSlug().equals(slug) || query.getLastToken() != 0) {
                throw badRequest(CONFLICT, "Query already exists");
            }
            Query query = queries.get(queryId);
            return withCompressionConfiguration(Response.ok(query.getInitialQueryResults(uriInfo, xForwardedProto, xPrestoPrefixUrl, binaryResults)), compressionEnabled).build();
        }
```

I'd recommend returning the same query ID instead of proceeding and creating a new one, to guard against a race condition where we've retried the PUT, but meanwhile a GET has been made on the old copy of the query.

tdcmeehan (changes requested):
Looks good % one more comment

presto-main/src/main/java/com/facebook/presto/server/protocol/QueuedStatementResource.java
prithvip: @tdcmeehan Done, changed to return the same query ID.

presto-main/src/main/java/com/facebook/presto/server/protocol/QueuedStatementResource.java
```diff
@@ -222,6 +224,62 @@ public Response postStatement(
...
+
+        // TODO: For future cases we may want to start tracing from client. Then continuation of tracing
+        //       will be needed instead of creating a new trace here.
+        SessionContext sessionContext = new HttpRequestSessionContext(
+                servletRequest,
+                sqlParserOptions,
+                tracerProviderManager.getTracerProvider(),
+                Optional.of(sessionPropertyManager));
+        Query query = new Query(statement, sessionContext, dispatchManager, executingQueryResponseProvider, 0, queryId, slug);
+        queries.put(query.getQueryId(), query);
```
tdcmeehan, line 278: One more thing just to make it 100% foolproof:

```suggestion
        Query query = queries.computeIfAbsent(queryId, key -> new Query(statement, sessionContext, dispatchManager, executingQueryResponseProvider, 0, queryId, slug));
```

tdcmeehan (changes requested):

presto-main/src/main/java/com/facebook/presto/server/protocol/QueuedStatementResource.java
tdcmeehan: We could combine them into one check, and remove the check you have above.

```suggestion
        Query attemptedQuery = new Query(statement, sessionContext, dispatchManager, executingQueryResponseProvider, 0, queryId, slug);
        Query query = queries.putIfAbsent(query.getQueryId(), unused -> query);
        if (attemptedQuery != query && !attemptedQuery.getSlug().equals(query.getSlug()) || query.getLastToken() != 0) {
            throw badRequest(CONFLICT, "Query already exists");
        }
```

tdcmeehan (approved):

presto-main/src/main/java/com/facebook/presto/server/protocol/QueuedStatementResource.java
prithvip: @amitkdutta do you have any details on this client?

presto-openapi/src/main/resources/queued_statement.yaml
```diff
@@ -46,6 +46,58 @@ paths:
                 $ref: './schemas.yaml/#/components/schemas/QueryResults'
         '400':
           description: Bad request
+    put:
```
ZacBlanco, line 49: This should be put under a separate `paths` component which includes the `queryId` in the path. See the example here: https://swagger.io/docs/specification/describing-parameters/

e.g. for this case

```yaml
paths:
  /v1/statement/{queryId}:
    put:
      parameters:
        - name: queryId
          in: path
          .... 
```


While reviewing, I also noticed some other existing paths in this part of the spec are slightly wrong as well, so I will fix them in a separate PR.

ZacBlanco (changes requested):
Thanks for adding this! Sorry for the slow review. Was OOO last week.

Spec looks good, just need to shift the path component it is under.

presto-openapi/src/main/resources/queued_statement.yaml
prithvip: @ZacBlanco done

ZacBlanco (approved):

prithvip:
@tdcmeehan Could I get another stamp?

tdcmeehan (approved):

steveburnett (commented):
LGTM! (docs)

Issue #22068, opened 2024-03-02
OrcBatchPageSourceFactory catches raw java.lang.Exception


elharo:
This is recommended against by Effective Java. See if this can be cleaned up to catch more specific exceptions. This class might also miss closing some resources along some paths that could be managed with more careful exception handling and try-with-resources.

misterjpapa:
@elharo, please note there over 20 other non-test classes in the presto-hive module in which a raw Exception is caught similar to this issue. See the attached PDF for a list. In most cases the reason is to wrap the Exception with either a RuntimeException or a PrestoException (which extends the RuntimeException) and then rethrow.
[presto-hive catching raw java.lang.Exception - Sheet1.pdf](https://github.com/user-attachments/files/16571590/presto-hive.catching.raw.java.lang.Exception.-.Sheet1.pdf)
It would be relatively straightforward to refactor these calls to use best practices, but adding a unit test for each change may be a larger effort.

PR #23419, opened 2024-08-11
Refactor try-catch clause for createOrcPageSource

misterjpapa:
Refactor the OrcBatchPageSourceFactory.createOrcPageSource method as it catches raw java.lang.Exception which is not a recommended practice. A better practice involves catching the specific Exception types that can be thrown in the try-catch block. Kept theOrcDataSource object outside of a try-with-resources block as it is the responsibility of the ConnectorPageSource to close both the reader and datasource.

Added an unit test to verify that any RuntimeException thrown in try block is still mapped to a PrestoException and thrown as expected.

## Description
Refactor the try-catch block to catch specific IOException and RuntimeException class rather than unspecific Exception class.

## Motivation and Context
Refactor the OrcBatchPageSourceFactory.createOrcPageSource to follow best practices for Closeable resource creation and exception handling.
https://github.com/prestodb/presto/issues/22068

## Impact
None

## Test Plan
Added a new unit test class TestOrcBatchPageSourceFactory with a new unit test to that any RuntimeException thrown in try block is still mapped to a PrestoException and thrown as expected.

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-hive/src/main/java/com/facebook/presto/hive/orc/OrcBatchPageSourceFactory.java
+++ b/presto-hive/src/main/java/com/facebook/presto/hive/orc/OrcBatchPageSourceFactory.java
@@ -258,13 +258,14 @@ public class OrcBatchPageSourceFactory
                     partitionID,
                     rowGroupID);
         }
-        catch (Exception e) {
+        catch (IOException | RuntimeException exception) {
             try {
+                //in case the OrcReader did not close the OrcDataSource
                 orcDataSource.close();
             }
             catch (IOException ignored) {
             }
-            throw mapToPrestoException(e, path, fileSplit);
+            throw mapToPrestoException(exception, path, fileSplit);
         }
     }
 }
--- a/presto-hive/src/test/java/com/facebook/presto/hive/orc/TestOrcBatchPageSourceFactory.java
+++ b/presto-hive/src/test/java/com/facebook/presto/hive/orc/TestOrcBatchPageSourceFactory.java
@@ -1,0 +1,104 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.hive.orc;
+
+import com.facebook.presto.common.predicate.TupleDomain;
+import com.facebook.presto.hive.DwrfEncryptionMetadata;
+import com.facebook.presto.hive.EncryptionInformation;
+import com.facebook.presto.hive.FileFormatDataSourceStats;
+import com.facebook.presto.hive.HiveColumnHandle;
+import com.facebook.presto.hive.HiveDwrfEncryptionProvider;
+import com.facebook.presto.hive.HiveFileSplit;
+import com.facebook.presto.hive.TestOrcBatchPageSourceMemoryTracking;
+import com.facebook.presto.orc.OrcReaderOptions;
+import com.facebook.presto.orc.StorageStripeMetadataSource;
+import com.facebook.presto.orc.StripeMetadataSourceFactory;
+import com.facebook.presto.orc.cache.StorageOrcFileTailSource;
+import com.facebook.presto.spi.ConnectorPageSource;
+import com.facebook.presto.spi.PrestoException;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hive.ql.io.orc.OrcOutputFormat;
+import org.apache.hadoop.hive.ql.io.orc.OrcSerde;
+import org.apache.hadoop.mapred.FileSplit;
+import org.apache.hadoop.mapred.JobConf;
+import org.joda.time.DateTimeZone;
+import org.testng.annotations.Test;
+
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.time.Instant;
+import java.util.List;
+import java.util.Optional;
+
+import static com.facebook.presto.hive.HiveFileContext.DEFAULT_HIVE_FILE_CONTEXT;
+import static com.facebook.presto.hive.HiveTestUtils.FUNCTION_AND_TYPE_MANAGER;
+import static com.facebook.presto.hive.HiveTestUtils.HDFS_ENVIRONMENT;
+import static com.facebook.presto.hive.HiveTestUtils.SESSION;
+import static com.facebook.presto.orc.OrcEncoding.ORC;
+import static org.apache.hadoop.hive.serde2.objectinspector.primitive.PrimitiveObjectInspectorFactory.javaStringObjectInspector;
+import static org.testng.Assert.assertTrue;
+import static org.testng.Assert.fail;
+
+public class TestOrcBatchPageSourceFactory
+{
+    @Test
+    public void testCreateOrcPageSourceException() throws Exception
+    {
+        List<TestOrcBatchPageSourceMemoryTracking.TestColumn> testColumns = ImmutableList.<TestOrcBatchPageSourceMemoryTracking.TestColumn>builder()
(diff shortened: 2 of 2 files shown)


presto-hive/src/test/java/com/facebook/presto/hive/orc/TestOrcBatchPageSourceFactory.java
```diff
@@ -0,0 +1,103 @@
...
+                ImmutableMap.of(),
+                0);
+        OrcReaderOptions orcReaderOptions = null; //need to be set to null to trigger PrestoException
+        Optional<EncryptionInformation> encryptionInformation =
+                Optional.of(EncryptionInformation.fromEncryptionMetadata(DwrfEncryptionMetadata.forPerField(
+                ImmutableMap.of("field1", "test1".getBytes()),
+                ImmutableMap.of(),
+                "test_algo",
+                "test_provider")));
+        try (ConnectorPageSource connectorPageSource = OrcBatchPageSourceFactory.createOrcPageSource(ORC, HDFS_ENVIRONMENT,
```
sdruzkin, line 88: The problem with this unit test is that it does not read from the page source it created.

The reason why the page source factory is implemented as it's is that the life-cycle of the OrcDataSource is attached to the life-cycle of the created OrcPageSource. If you use try-catch-with-resource it will close the OrcDataSource right after creating the OrcPageSource and you won't be able to read anything from it.

sdruzkin (changes requested):

presto-hive/src/test/java/com/facebook/presto/hive/orc/TestOrcBatchPageSourceFactory.java
misterjpapa: Why does this matter? The idea of the unit test was to make sure a RuntimeException was thrown and it was still caught and wrapped by a PrestoException as expected. The ConnectorPageSource object fails to get instantiated due to the orcReaderOptions.

presto-hive/src/main/java/com/facebook/presto/hive/orc/OrcBatchPageSourceFactory.java
```diff
@@ -201,11 +201,10 @@ public static ConnectorPageSource createOrcPageSource(
         checkArgument(domainCompactionThreshold >= 1, "domainCompactionThreshold must be at least 1");
         checkRowIDPartitionComponent(columns, rowIDPartitionComponent);
 
-        OrcDataSource orcDataSource = getOrcDataSource(session, fileSplit, hdfsEnvironment, configuration, hiveFileContext, stats);
         Path path = new Path(fileSplit.getPath());
 
         OrcAggregatedMemoryContext systemMemoryUsage = new HiveOrcAggregatedMemoryContext();
-        try {
+        try (OrcDataSource orcDataSource = getOrcDataSource(session, fileSplit, hdfsEnvironment, configuration, hiveFileContext, stats)) {
```
misterjpapa, line 207: @sdruzkin - looking at your earlier comment "The reason why the page source factory is implemented as it's is that the life-cycle of the OrcDataSource is attached to the life-cycle of the created OrcPageSource. If you use try-catch-with-resource it will close the OrcDataSource right after creating the OrcPageSource and you won't be able to read anything from it" - I should **NOT** wrap the OrcDataSource orcDataSource in a try-catch-with resource here then.

presto-hive/src/test/java/com/facebook/presto/hive/orc/TestOrcBatchPageSourceFactory.java
```diff
@@ -0,0 +1,103 @@
...
+import static org.apache.hadoop.hive.serde2.objectinspector.primitive.PrimitiveObjectInspectorFactory.javaStringObjectInspector;
+import static org.testng.Assert.assertTrue;
+import static org.testng.Assert.fail;
+
+public class TestOrcBatchPageSourceFactory
+{
+    @Test
+    public void testCreateOrcPageSourceException() throws Exception
+    {
+        Random random = new Random();
```
elharo, line 61: random in unit tests is a big no-no as it makes tests non-reproducible.

presto-hive/src/test/java/com/facebook/presto/hive/orc/TestOrcBatchPageSourceFactory.java
```diff
@@ -0,0 +1,103 @@
...
+        List<TestOrcBatchPageSourceMemoryTracking.TestColumn> testColumns = ImmutableList.<TestOrcBatchPageSourceMemoryTracking.TestColumn>builder()
+                .add(new TestOrcBatchPageSourceMemoryTracking.TestColumn("p_empty_string", javaStringObjectInspector, () -> "", true))
+                .add(new TestOrcBatchPageSourceMemoryTracking.TestColumn("p_string", javaStringObjectInspector, () -> Long.toHexString(random.nextLong()), false))
```
elharo, line 64: here you can use an arbitrary constant value like 893247L instead of a random one

presto-hive/src/test/java/com/facebook/presto/hive/orc/TestOrcBatchPageSourceFactory.java
```diff
@@ -0,0 +1,103 @@
...
+                .build();
+        File tempFile = File.createTempFile("presto_test_orc_batch_page_source_factory", "orc");
+        tempFile.delete();
```
elharo, line 67: You create the file and immediately delete it? This is strange

presto-hive/src/test/java/com/facebook/presto/hive/orc/TestOrcBatchPageSourceFactory.java
```diff
@@ -0,0 +1,103 @@
...
+                TupleDomain.all(), DateTimeZone.getDefault(), FUNCTION_AND_TYPE_MANAGER, true,
+                new FileFormatDataSourceStats(), 100, new StorageOrcFileTailSource(),
+                StripeMetadataSourceFactory.of(new StorageStripeMetadataSource()), DEFAULT_HIVE_FILE_CONTEXT, orcReaderOptions,
+                encryptionInformation, HiveDwrfEncryptionProvider.NO_ENCRYPTION.toDwrfEncryptionProvider(), SESSION, Optional.empty())) {
+            fail("createOrcPageSource call should have resulted in an PrestoException.");
+        }
+        catch (PrestoException pe) {
+            assertTrue(pe.getMessage().contains("orcReaderOptions is null"), "Incorrect PrestoException was thrown.");
+        }
+        catch (IOException e) {
```
elharo, line 99: simpler to let the IOexception be thrown which would also fail the test

misterjpapa: My thinking was that since IOException was the other exception that was explicitly caught and wrapped in a PrestoException I wanted a fail statement if an IOException was thrown eg: "escaped". Now after writing this comment I guess I should have been catching all exception types, so you are right should just let it be thrown for simplicity sake.

presto-hive/src/test/java/com/facebook/presto/hive/orc/TestOrcBatchPageSourceFactory.java
misterjpapa: I used the same setup for the test https://github.com/prestodb/presto/blob/be299743bf1b2feaaf0fbfaa3d09ce052908a22b/presto-hive/src/test/java/com/facebook/presto/hive/TestOrcBatchPageSourceMemoryTracking.java#L168  - looking at the code the idea is to get the absolute path of the temporary file as it is later used by a writer to write content into that  path after which a file is created.

presto-hive/src/main/java/com/facebook/presto/hive/orc/OrcBatchPageSourceFactory.java
misterjpapa: @sdruzkin- Thank you Sergii for your review and explanation of the close contract between the OrcDataSource, OrcReader, and OrcBatchPageSource and why a try-catch-with resource cannot be used here. This makes the refactor requested in this issue more trivial now.

presto-hive/src/test/java/com/facebook/presto/hive/orc/TestOrcBatchPageSourceFactory.java
elharo: That feels rather roundabout. I suspect there's a simple and more reliable way to do this without creating and deleting the file. Just use the path and System.getProperty("java.io.tmpdir") for instance.

sdruzkin:
Folks, I'm not sure I see much value in the current revision of this PR. I suggest to abandon it.

PR #23423, opened 2024-08-12
[native] Add hook for Velox plan validation.

amitkdutta:
## Description
This PR adds a hook to further check presto to velox converted plan.

## Motivation and Context
After Presto plan is converted to Velox plan, it is often required to disable certain features based on the generated plan. The reasoning is even though the plan is converted properly, due to arbitary limitation in Velox (either correctness or performacne), we may want to fail the query execution. Adding a hook (similar to spilling directory), allows us to customize query execution and make it configurable based on current capabilities of Velox. 

## Impact
None

## Test Plan
Ran worker and cooridnator, printed logs inside dummy validator and observed it was printed.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/PrestoServer.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoServer.cpp
@@ -476,8 +476,12 @@ void PrestoServer::run() {
         << "Spilling root directory: " << baseSpillDirectory;
   }
 
+  initPrestoToVeloxPlanValidator();
   taskResource_ = std::make_unique<TaskResource>(
-      *taskManager_, pool_.get(), httpSrvCpuExecutor_.get());
+      pool_.get(),
+      httpSrvCpuExecutor_.get(),
+      getPlanValidator(),
+      *taskManager_);
   taskResource_->registerUris(*httpServer_);
   if (systemConfig->enableSerializedPageChecksum()) {
     enableChecksum();
@@ -1247,6 +1251,15 @@ void PrestoServer::enableWorkerStatsReporting() {
   // This flag must be set to register the counters.
   facebook::velox::BaseStatsReporter::registered = true;
   registerStatsCounters();
+}
+
+void PrestoServer::initPrestoToVeloxPlanValidator() {
+  VELOX_CHECK_NULL(planValidator_);
+  planValidator_ = std::make_shared<PrestoToVeloxPlanValidator>();
+}
+
+PrestoToVeloxPlanValidator* PrestoServer::getPlanValidator() {
+  return planValidator_.get();
 }
 
 void PrestoServer::populateMemAndCPUInfo() {
--- a/presto-native-execution/presto_cpp/main/PrestoServer.h
+++ b/presto-native-execution/presto_cpp/main/PrestoServer.h
@@ -25,6 +25,7 @@
 #include "presto_cpp/main/PeriodicHeartbeatManager.h"
 #include "presto_cpp/main/PrestoExchangeSource.h"
 #include "presto_cpp/main/PrestoServerOperations.h"
+#include "presto_cpp/main/types/PrestoToVeloxPlanValidator.h"
 #include "velox/common/caching/AsyncDataCache.h"
 #include "velox/common/memory/MemoryAllocator.h"
 #if __has_include("filesystem")
@@ -172,6 +173,11 @@ class PrestoServer {
   /// Invoked to enable stats reporting and register counters.
   virtual void enableWorkerStatsReporting();
 
+  /// Invoked to initialize Presto to Velox plan validator.
+  virtual void initPrestoToVeloxPlanValidator();
+
+  PrestoToVeloxPlanValidator* getPlanValidator();
+
   /// Invoked to get the list of filters passed to the http server.
   std::vector<std::unique_ptr<proxygen::RequestHandlerFactory>>
   getHttpServerFilters();
@@ -231,6 +237,8 @@ class PrestoServer {
 
   // Executor for spilling.
   std::shared_ptr<folly::CPUThreadPoolExecutor> spillerExecutor_;
+
+  std::shared_ptr<PrestoToVeloxPlanValidator> planValidator_;
 
   std::unique_ptr<http::HttpClientConnectionPool> exchangeSourceConnectionPool_;
 
--- a/presto-native-execution/presto_cpp/main/TaskResource.h
+++ b/presto-native-execution/presto_cpp/main/TaskResource.h
@@ -15,6 +15,7 @@
 
 #include "presto_cpp/main/TaskManager.h"
 #include "presto_cpp/main/http/HttpServer.h"
+#include "presto_cpp/main/types/PrestoToVeloxPlanValidator.h"
 #include "velox/common/memory/Memory.h"
 
 namespace facebook::presto {
@@ -22,11 +23,13 @@ namespace facebook::presto {
 class TaskResource {
  public:
   explicit TaskResource(
-      TaskManager& taskManager,
       velox::memory::MemoryPool* pool,
-      folly::Executor* httpSrvCpuExecutor)
+      folly::Executor* httpSrvCpuExecutor,
+      PrestoToVeloxPlanValidator* planValidator,
+      TaskManager& taskManager)
       : httpSrvCpuExecutor_(httpSrvCpuExecutor),
         pool_{pool},
+        planValidator_(planValidator),
         taskManager_(taskManager) {}
 
   void registerUris(http::HttpServer& server);
@@ -97,6 +100,7 @@ class TaskResource {
 
   folly::Executor* const httpSrvCpuExecutor_;
   velox::memory::MemoryPool* const pool_;
+  PrestoToVeloxPlanValidator* const planValidator_;
 
   TaskManager& taskManager_;
 };
--- a/presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp
+++ b/presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp
@@ -244,8 +244,14 @@ class TaskManagerTest : public testing::Test {
 
     taskManager_ = std::make_unique<TaskManager>(
         driverExecutor_.get(), httpSrvCpuExecutor_.get(), nullptr);
+
+    auto validator =
+        std::make_shared<facebook::presto::PrestoToVeloxPlanValidator>();
     taskResource_ = std::make_unique<TaskResource>(
-        *taskManager_.get(), leafPool_.get(), httpSrvCpuExecutor_.get());
+        leafPool_.get(),
+        httpSrvCpuExecutor_.get(),
+        validator.get(),
+        *taskManager_.get());
 
     auto httpServer = std::make_unique<http::HttpServer>(
         httpSrvIOExecutor_,
--- a/presto-native-execution/presto_cpp/main/types/PrestoToVeloxPlanValidator.cpp
+++ b/presto-native-execution/presto_cpp/main/types/PrestoToVeloxPlanValidator.cpp
@@ -1,0 +1,22 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "presto_cpp/main/types/PrestoToVeloxPlanValidator.h"
+
+namespace facebook::presto {
+bool PrestoToVeloxPlanValidator::validatePlanFragment(
+    const velox::core::PlanFragment& fragment) {
+  return true;
+}
+} // namespace facebook::presto
--- a/presto-native-execution/presto_cpp/main/types/PrestoToVeloxPlanValidator.h
+++ b/presto-native-execution/presto_cpp/main/types/PrestoToVeloxPlanValidator.h
@@ -1,0 +1,23 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+#include "velox/core/PlanFragment.h"
+
+namespace facebook::presto {
+class PrestoToVeloxPlanValidator {
+ public:
+  virtual bool validatePlanFragment(const velox::core::PlanFragment& fragment);
+  virtual ~PrestoToVeloxPlanValidator() = default;
+};
+} // namespace facebook::presto
--- a/presto-native-execution/presto_cpp/main/TaskResource.cpp
+++ b/presto-native-execution/presto_cpp/main/TaskResource.cpp
@@ -344,6 +344,7 @@ proxygen::RequestHandler* TaskResource::createOrUpdateTask(
           VeloxInteractiveQueryPlanConverter converter(queryCtx.get(), pool_);
           planFragment = converter.toVeloxQueryPlan(
               prestoPlan, updateRequest.tableWriteInfo, taskId);
+          planValidator_->validatePlanFragment(planFragment);
         }
 
         return taskManager_.createOrUpdateTask(
--- a/presto-native-execution/presto_cpp/main/types/CMakeLists.txt
+++ b/presto-native-execution/presto_cpp/main/types/CMakeLists.txt
@@ -15,8 +15,10 @@ add_library(presto_type_converter OBJECT TypeParser.cpp)
 target_link_libraries(presto_type_converter velox_type_parser)
 
 add_library(
-  presto_types OBJECT PrestoToVeloxQueryPlan.cpp PrestoToVeloxExpr.cpp
-                      PrestoToVeloxSplit.cpp PrestoToVeloxConnector.cpp)
+  presto_types OBJECT
+  PrestoToVeloxQueryPlan.cpp PrestoToVeloxExpr.cpp
+  PrestoToVeloxPlanValidator.cpp PrestoToVeloxSplit.cpp
+  PrestoToVeloxConnector.cpp)
 add_dependencies(presto_types presto_operators presto_type_converter velox_type
                  velox_type_fbhive velox_dwio_dwrf_proto)
 


aditi-pandit:
@amitkdutta : Can you give some examples of such validation ? Its quite hard to back-track from a failure at this point to the Presto PlanNode that got translated. Its simpler to fail during translation imo.

@tdcmeehan and @BryanCutler are also working of fail fast plan validation for Presto/Prestissimo. Lets decide on something together.

presto-native-execution/presto_cpp/main/PrestoServer.h
```diff
@@ -172,6 +173,11 @@ class PrestoServer {
   /// Invoked to enable stats reporting and register counters.
   virtual void enableWorkerStatsReporting();
 
+  /// Invoked to get a planValidator instance. The validator will be used to
+  /// validate a velox planfragment in TaskResoruce.
```
xiaoxmeng, line 177:
```
nit: s/in/by/
```

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -1249,6 +1252,13 @@ void PrestoServer::enableWorkerStatsReporting() {
   registerStatsCounters();
 }
 
+std::shared_ptr<facebook::presto::PrestoToVeloxPlanValidator>
+PrestoServer::getPlanValidator() {
+  auto defaultPlanValidator =
```
xiaoxmeng, line 1257:
```
static auto defaultPlanValidator =
```

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -1249,6 +1252,13 @@ void PrestoServer::enableWorkerStatsReporting() {
   registerStatsCounters();
 }
 
+std::shared_ptr<facebook::presto::PrestoToVeloxPlanValidator>
```
xiaoxmeng, line 1255: We can just return a raw pointer and let presto server to hold the reference.

presto-native-execution/presto_cpp/main/TaskResource.h
```diff
@@ -99,6 +103,7 @@ class TaskResource {
   velox::memory::MemoryPool* const pool_;
 
   TaskManager& taskManager_;
+  std::shared_ptr<facebook::presto::PrestoToVeloxPlanValidator> planValidator_;
```
xiaoxmeng, line 106:
```
facebook::presto::PrestoToVeloxPlanValidator* const planValidator_;
```
And move const member first.

presto-native-execution/presto_cpp/main/types/PrestoToVeloxPlanValidator.h
```diff
@@ -0,0 +1,23 @@
...
+ * limitations under the License.
+ */
+#pragma once
+#include "velox/core/PlanFragment.h"
+
+namespace facebook::presto {
+class PrestoToVeloxPlanValidator {
+ public:
+  virtual bool validatePlanFragment(const velox::core::PlanFragment& fragment);
+  virtual ~PrestoToVeloxPlanValidator() {}
```
xiaoxmeng, line 21:
```
virtual ~PrestoToVeloxPlanValidator() = default;
```

xiaoxmeng (commented):
@amitkdutta LGTM % nits. Thanks!

amitkdutta:
> @amitkdutta : Can you give some examples of such validation ? Its quite hard to back-track from a failure at this point to the Presto PlanNode that got translated. Its simpler to fail during translation imo.
> 
> @tdcmeehan and @BryanCutler are also working of fail fast plan validation for Presto/Prestissimo. Lets decide on something together.

@aditi-pandit Thanks for the review. Indeed its hard to walk back to the plan fragment. The purpose of this hook is to allow such walk (even ineffieicnet) if necessary. A good example of it is [Nested loop Join disablement](https://github.com/prestodb/presto/pull/23341) or [disabling timestamp with timezone](https://github.com/prestodb/presto/pull/23179). Key points are:
- Sometimes we observe correctness issues with certain features (e.g. NLJ) while the system is already in production. At that time we want to disable (or enable) something based on plan fragment on the fly with custom configurations. This kind of usage may not be generally used (or can be too much Meta specific), hence we want to handle the hook based on runtime requirements instead of introducing random plan conversion errors.
- There are cases where `time stamp with time zone` are only used for casting and not for general equality/ordering etc. In such use cases, we want to have a validator which block queries if it finds `time stamp with time zone` for ordering purpose, but allow for other cases.

These use cases might sound custom (and indeed they are). However they serve a purpose where a feature is not fully ready, but part of it can be used (or not used) based on generated plan fragment. Having a hook allows to do such manipulation, hence this PR. 

The validation call back here always returns true here. If we find somehting that can be genrally used as validation, we can add here.

This is similar to what has been done with spilling directory fetching hook, which allows us to override and provide a custom implementation that dynamically (and periodically) get spill path.

We will definitely work with @tdcmeehan for fail fast validation, and I assume that will be more generic. If needed, we can remove this default validator also once the general one is more valid. But I assume the general one will do such validation during plan conversion code. Having a hook after plan conversion gives some flexibility (and deployment speed) for partial features.

tdcmeehan:
> There are cases where time stamp with time zone are only used for casting and not for general equality/ordering etc. In such use cases, we want to have a validator which block queries if it finds time stamp with time zone for ordering purpose, but allow for other cases.

Can't this be done as a plan checker?

I think this PR has two problems.  The first is that it's too late--resources have already been allocated by the time this check occurs.  The second is, as @aditi-pandit pointed out, it makes it hard to work backwards to the plan.

If this can wait, we're working on a plan checker now.  It uses a plugin system, so Meta can write whatever code they want and place it behind a plugin.  We can work on getting this out in a reasonable time frame, or we can work together to get it out even quicker.  If it can't, then I propose we can let this PR pass, but once we introduce the plan checker we'll deprecate this and eventually delete.  I consider this two phased solution, while not ideal, to be better than introducing Meta-specific requirements directly into open source code.  WDYT?

amitkdutta:
> > There are cases where time stamp with time zone are only used for casting and not for general equality/ordering etc. In such use cases, we want to have a validator which block queries if it finds time stamp with time zone for ordering purpose, but allow for other cases.
> 
> Can't this be done as a plan checker?
> 
> I think this PR has two problems. The first is that it's too late--resources have already been allocated by the time this check occurs. The second is, as @aditi-pandit pointed out, it makes it hard to work backwards to the plan.
> 
> If this can wait, we're working on a plan checker now. It uses a plugin system, so Meta can write whatever code they want and place it behind a plugin. We can work on getting this out in a reasonable time frame, or we can work together to get it out even quicker. If it can't, then I propose we can let this PR pass, but once we introduce the plan checker we'll deprecate this and eventually delete. I consider this two phased solution, while not ideal, to be better than introducing Meta-specific requirements directly into open source code. WDYT?

@tdcmeehan This change is not necessarily Meta specific though. [We have built hooks in PrestoServer to allow customers customize different aspects of the server](https://github.com/prestodb/presto/blob/master/presto-native-execution/presto_cpp/main/PrestoServer.h#L119-L173). This one just adds a hook after plan conversion in the worker. We can also add a config in worker to completely disable the hook, and Meta (or others) can utilize it if any tricky scenario arises. I understand its not optimal, and I also see at some point it will be deleted when the Velox eval becomes mature.

presto-native-execution/presto_cpp/main/PrestoServer.h
```diff
@@ -172,6 +173,11 @@ class PrestoServer {
   /// Invoked to enable stats reporting and register counters.
   virtual void enableWorkerStatsReporting();
 
+  /// Invoked to get a planValidator instance. The validator will be used to
+  /// validate a velox planfragment by TaskResoruce.
+  virtual std::shared_ptr<facebook::presto::PrestoToVeloxPlanValidator>
```
xiaoxmeng, line 178:
```
virtual PrestoToVeloxPlanValidator*  getPlanValidator()
```
Drop facebook::presto:: as we already in this namespace?

presto-native-execution/presto_cpp/main/PrestoServer.h
```diff
@@ -232,6 +238,8 @@ class PrestoServer {
   // Executor for spilling.
   std::shared_ptr<folly::CPUThreadPoolExecutor> spillerExecutor_;
 
+  std::shared_ptr<facebook::presto::PrestoToVeloxPlanValidator> planValidator_;
```
xiaoxmeng, line 241: We probably need two APIs one is to init planValidator_ and the other is to get plan validator? The latter returns a raw pointer.

amitkdutta:
> @amitkdutta that's all understood, but we also need to avoid doing things two different ways. We're working on a solution that doesn't have the deficiencies I listed, so I expect that when we have a plan checker, this can be deleted, not just disabled.

@tdcmeehan Of course, we should not keep duplicated items. We can remove it once we merge PlanChecker. Will be great if you can share some details about it also, we can go through it and also discuss in the bi-weekly meeting we have.

presto-native-execution/presto_cpp/main/TaskResource.h
```diff
@@ -99,6 +102,7 @@ class TaskResource {
   velox::memory::MemoryPool* const pool_;
 
   TaskManager& taskManager_;
+  facebook::presto::PrestoToVeloxPlanValidator* planValidator_;
```
xiaoxmeng, line 105:
```
PrestoToVeloxPlanValidator* const planValidator_;
```
Put const members first. Drop facebook::presto::

pedroerp:
@tdcmeehan do you have a timeline for the plugin-based solution to be ready? We all agree doing this at query planning stage on the coordinator would be ideal, but depending on how long this would take to be available, it may make sense to have a quick way to toggle on/off traffic in the interim, even if suboptimal.

xiaoxmeng:
> @amitkdutta that's all understood, but we also need to avoid doing things two different ways. We're working on a solution that doesn't have the deficiencies I listed, so I expect that when we have a plan checker, this can be deleted, not just disabled.

@tdcmeehan  does the plan checker allow to customize or override some of the check logic which is specific to a customer like Meta without changing OSS code? Thanks!

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -1249,6 +1253,14 @@ void PrestoServer::enableWorkerStatsReporting() {
   registerStatsCounters();
 }
 
+void PrestoServer::initPrestoToVeloxPlanValidator() {
+  planValidator_ = std::make_shared<PrestoToVeloxPlanValidator>();
```
xiaoxmeng, line 1257:
```
VELOX_CHECK_NULL(planValidator_);
planValidator_ = ...
```

xiaoxmeng (dismissed):
@amitkdutta thanks for the update!

xiaoxmeng (approved):

presto-native-execution/presto_cpp/main/types/PrestoToVeloxPlanValidator.cpp
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "presto_cpp/main/types/PrestoToVeloxPlanValidator.h"
+
+namespace facebook::presto {
```
pedroerp, line 17: nit: not sure about Presto, but in Velox there is always a blank line after namespace definition

presto-native-execution/presto_cpp/main/types/PrestoToVeloxPlanValidator.h
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+#include "velox/core/PlanFragment.h"
+
+namespace facebook::presto {
+class PrestoToVeloxPlanValidator {
```
pedroerp, line 18: usually useful to add some documentation here on what this class is intended for, and how it should be used.

presto-native-execution/presto_cpp/main/PrestoServer.h
```diff
@@ -172,6 +173,11 @@ class PrestoServer {
   /// Invoked to enable stats reporting and register counters.
   virtual void enableWorkerStatsReporting();
 
+  /// Invoked to initialize Presto to Velox plan validator.
+  virtual void initPrestoToVeloxPlanValidator();
```
pedroerp, line 177: better to turn this into a factory function so users don't need to learn that it actually has side-effects (it sets the class member, can't be called twice, etc). 
```
virtual std::shared_ptr<VeloxPlanValidator> makePlanValidator();
```

and then callers can use this with more flexibility:

```
validator_ = makePlanValidator();
or
TaskResource(makePlanValidator(), ...);
```

pedroerp: Could also drop the "PrestoTo" from the name since this, in effect, only validates a Velox query fragment.

presto-native-execution/presto_cpp/main/TaskResource.h
```diff
@@ -15,18 +15,21 @@
...
 namespace facebook::presto {
 
 class TaskResource {
  public:
   explicit TaskResource(
-      TaskManager& taskManager,
       velox::memory::MemoryPool* pool,
-      folly::Executor* httpSrvCpuExecutor)
+      folly::Executor* httpSrvCpuExecutor,
+      PrestoToVeloxPlanValidator* planValidator,
```
pedroerp, line 28: usually a good idea to pass the shared_ptr so that lifecycle and ownership are more explicitly defined.

presto-native-execution/presto_cpp/main/TaskResource.h
```diff
@@ -97,6 +100,7 @@ class TaskResource {
 
   folly::Executor* const httpSrvCpuExecutor_;
   velox::memory::MemoryPool* const pool_;
+  PrestoToVeloxPlanValidator* const planValidator_;
```
pedroerp, line 103:
```
VeloxPlanValidatorPtr planValidator_;
```

(using the aliases we usually create for shared_ptr in Velox)

presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp
```diff
@@ -244,8 +244,14 @@ class TaskManagerTest : public testing::Test {
 
     taskManager_ = std::make_unique<TaskManager>(
         driverExecutor_.get(), httpSrvCpuExecutor_.get(), nullptr);
+
+    auto validator =
```
pedroerp, line 248: to really validate it works, you could create a mock validator that throws, and expect the the task to throw that exception.

presto-native-execution/presto_cpp/main/types/PrestoToVeloxPlanValidator.h
```diff
@@ -0,0 +1,23 @@
...
+ public:
+  virtual bool validatePlanFragment(const velox::core::PlanFragment& fragment);
```
pedroerp, line 20: perhaps either leave as pure virtual, or leave the base implementation here in the header so you don't need to define a .cpp that doesn't really have anything?

presto-native-execution/presto_cpp/main/types/PrestoToVeloxPlanValidator.cpp
```diff
@@ -0,0 +1,22 @@
...
+bool PrestoToVeloxPlanValidator::validatePlanFragment(
+    const velox::core::PlanFragment& fragment) {
+  return true;
```
aditi-pandit, line 20: If there are multiple validations like 1. Don't allow TIMESTAMPTZ 2. Don't allow NLJ 3. Allow only particular combination of 1 and 2, then there would be a config for each validation and all checks in the same function. Did you consider allowing multiple validators instead ?

Also, given that I think this class would need to be aware of SystemConfig.

tdcmeehan:
@xiaoxmeng we will add a plan checker SPI, and a default implementation which delegates to the Presto sidecar, to validate that the generated plan may successfully be executed with Prestissimo and Velox (or, that at least the plan can get translated successfully).  This SPI can additionally take additional implementations, which may have Meta-specific logic, the same as our other SPI interfaces (connectors, event listeners, authenticators, etc).

@pedroerp @amitkdutta I agree that this can go out temporarily.  Long term, I don't think these sorts of checks shouldn't go into the worker, and I just wanted to give everyone the heads up that when this solution is ready, this interface might get deprecated, so whatever is added here may need to get migrated.  This design was briefly mentioned in [RFC-0003](https://github.com/prestodb/rfcs/blob/main/RFC-0003-native-spi.md#plannode-support), but we're working on a more detailed RFC specifically for this, and are targeting Q3 to get the code out.

pedroerp:
> @pedroerp @amitkdutta I agree that this can go out temporarily. Long term, I don't think these sorts of checks shouldn't go into the worker, and I just wanted to give everyone the heads up that when this solution is ready, this interface might get deprecated, so whatever is added here may need to get migrated.

+1. This validation is meant to be temporary anyhow, and we will relax the restrictions as bugs are fixed and more features are added to the native stack. We just need to have a killswitch we can use to disable traffic as we roll more of these things out to production.

presto-native-execution/presto_cpp/main/PrestoServer.h
```diff
@@ -232,6 +238,8 @@ class PrestoServer {
   // Executor for spilling.
   std::shared_ptr<folly::CPUThreadPoolExecutor> spillerExecutor_;
 
+  std::shared_ptr<PrestoToVeloxPlanValidator> planValidator_;
```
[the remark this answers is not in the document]
pedroerp: I guess you can only do that if you create a new validator per query. Should be pretty much the same, but this isn't how the code is structured today.

presto-native-execution/presto_cpp/main/TaskResource.h
pedroerp: (or unique_ptr, depending on how you organize things).

amitkdutta:
@tdcmeehan Totally agree with you. We can remove this callback once we have the Planchecker.
@aditi-pandit @pedroerp Let me address your comments in a follow up PR. I will add an example that fails queries with timestamp with timezone.

aditi-pandit:
> > @amitkdutta : Would be good to add the NLJ check as a plan validation to demonstrate the API in a real use-case.
> 
> NLJ is fixed in Velox already, but potentially we could use this to disable timestamp with timezone comparisons which are still broken

Duh... My bad :(  I picked up NLJ from a previous remark. But its fixed in Velox already.

Do we really want to disable timestamp with timezone here now instead of the PlanChecker in the co-ordinator ? Are we planning to  refine that logic in any way ? Even if we are, the PlanChecker is still a better place for that logic. That validation has to look at all PlanNodes, Function Signatures.. have deeper understanding of what individual planNodes do. Thats a better fit in the co-ordinator. And since we already have that code, I feel we shouldn't refactor it for the sake of having an example.

The NLJ situation was a better candidate for Prestissimo PlanValidator. It would be okay to add it as a test and not in the main code path as a compromise.

@pedroerp, @amitkdutta : wdyt ?

pedroerp:
The problem with timestamp with timezone blocker (as I understand) is that it is too coarse; it blocks any usage of the custom type, where the problem is really only when you need rely on its comparisons. If this is something that could be done in the coordinator, even better. Although even in that case, this API provides us more flexibility to quickly iterate on any other query shapes that may need to be blocked.

tdcmeehan:
I think we should create an issue to improve this checker, there's nothing preventing us from refining it.  And I don't think there are features or capabilities in C++ that are superior to the tools in Java that we would use to check for this.

I'm curious why this approach is considered quicker?  Is it because it allows us to write the check in C++?

pedroerp:
I suppose not only in C++, but it allow us to do the check based on a Velox Plan (after the conversion from Presto). Presto Java, in theory, doesn't know the details of that conversion.

tdcmeehan:
@pedroerp do you have an example of what information the Velox plan would contain that wouldn't be available in the Presto plan?  My thinking was it can't contain any essential additional information that isn't known upfront, because we always start with a Presto plan.

pedroerp:
Good question. I suppose we would have most of the same information in both places, minus details of the plan translation/mapping from java to cpp, which may not be available in Java. 

Overall, the idea here is less about re-implementing the blocks we already have (which are trivial and could be implemented somewhere else), and more about having an easy-to-use and convenient kill switch device that let's us disable a very specific slice of traffic. These things are found in production as we roll out additional traffic, so it's very hard to predict what exactly will be needed, but having access to the exact Velox plan being executed gives you the most flexibility.

tdcmeehan:
@pedroerp thanks for the clarification.  But I'm still not sure what exactly makes this check easy to use and convenient?  Is the ease of use due to it being developed as a hook, which means platforms can develop custom implementations internally?  i.e. the idea is you can write a very quick code change to your internal repository, without having to write a patch in open source where there is a desire for understanding and consensus?  Or is the ease of use due to it being a Velox plan, instead of a Presto plan?

PR #23200, opened 2024-07-12
Add plan validator for timestamp with timezone check for prestissimo

feilong-liu:
## Description
Detail in discussion of https://github.com/prestodb/presto/pull/23191

Add a plan checker for Prestissimo, so that queries which have timestamp with timezone type will fail.

## Motivation and Context
Fix https://github.com/facebookincubator/velox/issues/10338

## Impact
Only affect Prestissimo

## Test Plan
existing unit tests and
end to end test with queries in the issue:
```
presto:sf1> select cast(x as timestamp with time zone), count(1) from unnest(array['2024-04-10 07:00 America/Los_Angeles', '2024-04-10 10:00 America/New_York']) as t(x) group by 1;
Query 20240712_205745_00033_5bggp failed: Timestamp with Timezone type is not supported in Prestissimo
java.lang.IllegalStateException: Timestamp with Timezone type is not supported in Prestissimo
	at com.google.common.base.Preconditions.checkState(Preconditions.java:507)
	at com.facebook.presto.sql.planner.sanity.CheckNoTimestampWithNoTimezoneType$Visitor.visitPlan(CheckNoTimestampWithNoTimezoneType.java:63)
	at com.facebook.presto.sql.planner.sanity.CheckNoTimestampWithNoTimezoneType$Visitor.visitPlan(CheckNoTimestampWithNoTimezoneType.java:50)
	at com.facebook.presto.spi.plan.PlanVisitor.visitOutput(PlanVisitor.java:25)
	at com.facebook.presto.spi.plan.OutputNode.accept(OutputNode.java:98)
	at com.facebook.presto.sql.planner.sanity.CheckNoTimestampWithNoTimezoneType.validate(CheckNoTimestampWithNoTimezoneType.java:47)
	at com.facebook.presto.sql.planner.sanity.PlanChecker.lambda$validateFinalPlan$0(PlanChecker.java:81)
	at com.google.common.collect.ImmutableList.forEach(ImmutableList.java:407)
	at com.facebook.presto.sql.planner.sanity.PlanChecker.validateFinalPlan(PlanChecker.java:81)
	at com.facebook.presto.sql.Optimizer.validateAndOptimizePlan(Optimizer.java:131)
	at com.facebook.presto.execution.SqlQueryExecution.lambda$createLogicalPlanAndOptimize$3(SqlQueryExecution.java:557)
	at com.facebook.presto.common.RuntimeStats.profileNanos(RuntimeStats.java:136)
[...]

PR #23374, opened 2024-08-03
Move timezone with timestamp plan checker to intermediate plan stage

feilong-liu:
## Description
This plan check is added in https://github.com/prestodb/presto/pull/23200
However, it does not check the row expressions in table scan node (for example filters pushed down into table scan). Since we will need to check the type of table scan and add logic for all connectors if we want to solve it in the plan checker, I instead choose to move the plan check to the intermediate plan stage, which happens before filter pushdown.

## Motivation and Context
Cover queries which not covered

## Impact
Cover queries which not covered

## Test Plan
Existing unit tests and newly added test

## Contributor checklist

- [ ] Please make sure your submission complies with our [development](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#development), [formatting](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#formatting), [commit message](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#commit-formatting-and-pull-requests), and [attribution guidelines](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#attribution).
- [ ] PR description addresses the issue accurately and concisely.  If the change is non-trivial, a GitHub Issue is referenced.
- [ ] Documented new properties (with its default value), SQL syntax, functions, or other functionality.
- [ ] If release notes are required, they follow the [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines).
- [ ] Adequate tests were added if applicable.
- [ ] CI passed.

## Release Notes
Please follow [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines) and fill in the release notes below.

```
== RELEASE NOTES ==

General Changes
* Improve the plan check for timestamp with timezone type to cover filter pushdown :pr:`12345`
```

PR #23436, opened 2024-08-13
[native] Disable IPAddress type on presto native

mohsaka:
## Description

We have merged in support for IPADDRESS type in velox, however the order by is not supported due to
https://github.com/facebookincubator/velox/issues/10338

As a result, we will block IPAddress type on presto native similar to how timezone with timestamp is disabled via
https://github.com/prestodb/presto/pull/23200

## Motivation and Context

We do not want correctness failures since IPAddress type has been added.

More context:
https://github.com/facebookincubator/velox/pull/10596#discussion_r1710424382

## Test Plan

Followed the tests from the Timestamp with timezone disabling.

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* ... :pr:`12345`
* ... :pr:`12345`

Hive Connector Changes
* ... :pr:`12345`
* ... :pr:`12345`
```

```
== NO RELEASE NOTE ==
```


presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/CheckNoIPAddressType.java
```diff
@@ -0,0 +1,191 @@
...
+import com.facebook.presto.sql.planner.SimplePlanVisitor;
+import com.facebook.presto.sql.planner.TypeProvider;
+import com.facebook.presto.sql.planner.plan.WindowNode;
+
+import java.util.List;
+
+import static com.facebook.presto.type.IpAddressType.IPADDRESS;
+import static com.google.common.base.Preconditions.checkState;
+
+public class CheckNoIPAddressType
```
feilong-liu, line 46: Instead of adding one more plan visitor, I wonder if adding to existing plan visitor is a better idea.

aditi-pandit: @mohsaka  : +1

Please add the changes along with TimestampWithTimezone changes https://github.com/prestodb/presto/pull/23200 and https://github.com/prestodb/presto/pull/23374

mohsaka: Will do

mohsaka: @feilong-liu @aditi-pandit 

So I made a change to the error message string and hasType to support IPAddress type.

However I have a few questions.

1. Should we support/not support Timestamp With Timezone and IPAddress independently? Say for example Timestamp with timezone order by is fixed before IPAddress, then we can disable the types independently? If so I believe that we would have to remove the static final from the error message and in turn remove the static from the nested classes. Since I think the best way to do this would be to set the error message in `hasType` and exit on first occurrence of either type. 

We can also make it more complex by having `hasType` do a complete recursion or at least until all types are hit to make a more complex error message with either or both types included in it. 

If not. 
2. Currently I'm only using the `disableTimeStampWithTimeZoneForNative` config. As an or/and with both configs don't make sense. I was planning on renaming it to something like disablePhysicalMappedTypes. However it seems like this config is used, and I do not want to break deployment configs.
https://github.com/prestodb/presto/pull/23200#discussion_r1676659424

aditi-pandit: @mohsaka : We should enable/disable each independently.

I'm not entirely familiar with Java but is it possible to abstract a common base class to do PlanChecking that forwards hasType() to its child class. We could have 2 child classes for either type.

feilong-liu: @mohsaka you can pass the flag of disabling timestamp and iptype to the checker, and in the plan checker do the check accordingly based on the flag

mohsaka: @feilong-liu Agreed. I think I'm doing something similar to that. Let me push the changes once complete and please take a look. Thanks!

mohsaka: @feilong-liu @aditi-pandit Pushed the changes. Please take a look. 

Decided not to set the member variable string as I did not know if there are multiple callers on different threads. Could cause an issue where a string from one execution's string could appear as an error on another if run at the same time due to static.

Also currently made is so that we do not recurse through the entire plan as it was before. We could still do this and make a more informative string. But I think its unnecessary. 
Ex if there is a Timestamp with timezone type and IPAddress type then it would show both in one error message.

presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/CheckUnsupportedPrestissimoTypes.java
```diff
@@ -33,20 +33,28 @@
...
 import static com.facebook.presto.common.type.TimestampWithTimeZoneType.TIMESTAMP_WITH_TIME_ZONE;
+import static com.facebook.presto.type.IpAddressType.IPADDRESS;
 import static com.google.common.base.Preconditions.checkState;
 
-public class CheckNoTimestampWithTimezoneType
+public class CheckUnsupportedPrestissimoTypes
         implements PlanChecker.Checker
 {
-    private static final String errorMessage = "Timestamp with Timezone type is not supported in Prestissimo";
+    private static FeaturesConfig config;
```
feilong-liu, line 52: Why have it as a static member?

presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/CheckUnsupportedPrestissimoTypes.java
```diff
@@ -33,20 +33,28 @@
...
+
+    public CheckUnsupportedPrestissimoTypes(FeaturesConfig config)
+    {
+        this.config = config;
```
feilong-liu, line 56: this.config = requireNonNull(...)

presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/CheckUnsupportedPrestissimoTypes.java
```diff
@@ -121,71 +130,81 @@ public Void visitValues(ValuesNode node, Void context)
...
         }
     }
 
-    private static boolean hasTimestampWithTimezoneType(Type type)
+    private static String hasType(Type type)
     {
-        if (type.equals(TIMESTAMP_WITH_TIME_ZONE)) {
-            return true;
+        if (type.equals(TIMESTAMP_WITH_TIME_ZONE) && config.isDisableTimeStampWithTimeZoneForNative()) {
+            return "Timestamp with Timezone type is not supported in Prestissimo";
```
feilong-liu, line 193: Consider assign this to a static string variable

presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/CheckUnsupportedPrestissimoTypes.java
```diff
@@ -121,71 +130,81 @@ public Void visitValues(ValuesNode node, Void context)
...
+        }
+        if (type.equals(IPADDRESS) && config.isDisableIPAddressForNative()) {
+            return "IPAddress type is not supported in Prestissimo";
```
feilong-liu, line 196: Seem as above

presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/CheckUnsupportedPrestissimoTypes.java
```diff
@@ -121,71 +130,81 @@ public Void visitValues(ValuesNode node, Void context)
...
-            return hasTimestampWithTimezoneType(((MapType) type).getKeyType()) || hasTimestampWithTimezoneType(((MapType) type).getValueType());
+            String key = hasType(((MapType) type).getKeyType());
+            return key != null ? key : hasType(((MapType) type).getValueType());
         }
         else if (type instanceof RowType) {
-            return ((RowType) type).getTypeParameters().stream().anyMatch(CheckNoTimestampWithTimezoneType::hasTimestampWithTimezoneType);
+            return ((RowType) type).getTypeParameters().stream().map(CheckUnsupportedPrestissimoTypes::hasType).filter(Objects::nonNull).findFirst().orElse(null);
         }
-        return false;
+        return null;
```
feilong-liu, line 208: Consider use Optional instead of return null

presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/CheckUnsupportedPrestissimoTypes.java
```diff
@@ -121,71 +130,81 @@ public Void visitValues(ValuesNode node, Void context)
...
         {
-            checkState(!hasTimestampWithTimezoneType(expression.getType()), errorMessage);
+            String errorMessage = hasType(expression.getType());
+            checkState(errorMessage == null, errorMessage);
             return super.visitIntermediateFormExpression(expression, context);
         }
     }
 
-    private static boolean hasTimestampWithTimezoneType(Type type)
+    private static String hasType(Type type)
```
feilong-liu, line 190: now this function returns error message, consider rename this function

presto-main/src/test/java/com/facebook/presto/sql/planner/sanity/TestCheckNoIPAddressType.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.function.Function;
+
+import static com.facebook.presto.SessionTestUtils.TEST_SESSION;
+import static com.facebook.presto.common.type.BooleanType.BOOLEAN;
+import static com.facebook.presto.common.type.VarcharType.VARCHAR;
+import static com.facebook.presto.sql.planner.iterative.rule.test.PlanBuilder.assignment;
+import static com.facebook.presto.testing.TestingSession.testSessionBuilder;
+import static com.facebook.presto.type.IpAddressType.IPADDRESS;
+
+public class TestCheckNoIPAddressType
```
feilong-liu, line 40: Combine this test with the other existing test, and rename the test

presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/CheckUnsupportedPrestissimoTypes.java
mohsaka: It was due to the classes and functions being in a static context. Removed static from all of them.

presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/CheckUnsupportedPrestissimoTypes.java
mohsaka: Done

presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/CheckUnsupportedPrestissimoTypes.java
mohsaka: Done

presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/CheckUnsupportedPrestissimoTypes.java
mohsaka: Done

presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/CheckUnsupportedPrestissimoTypes.java
mohsaka: Done

presto-main/src/test/java/com/facebook/presto/sql/planner/sanity/TestCheckNoIPAddressType.java
mohsaka: Done

mohsaka:
@feilong-liu Really appreciate the thorough review! Pretty sure I got everything fixed up. Please let me know if there were any misunderstandings or other comments. Thanks!

presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/CheckUnsupportedPrestissimoTypes.java
mohsaka: Done

presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/CheckUnsupportedPrestissimoTypes.java
```diff
@@ -33,41 +33,54 @@
...
+import static com.facebook.presto.type.IpAddressType.IPADDRESS;
 import static com.google.common.base.Preconditions.checkState;
+import static java.util.Objects.requireNonNull;
 
-public class CheckNoTimestampWithTimezoneType
+public class CheckUnsupportedPrestissimoTypes
         implements PlanChecker.Checker
 {
-    private static final String errorMessage = "Timestamp with Timezone type is not supported in Prestissimo";
+    static final String timestampWithTimeszoneErrorMessage = "Timestamp with Timezone type is not supported in Prestissimo";
```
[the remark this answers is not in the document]
mohsaka: Done.

feilong-liu (approved):

aditi-pandit (approved):
Thanks @mohsaka

PR #23438, opened 2024-08-13
[native] Fix Meta internal build caused by folly::StringPiece

tanjialiang:
folly::StringPiece is not compatible with fmt::format(). Change it back to std::string_view
```
== NO RELEASE NOTE ==
```


amitkdutta (approved):
Thanks @tanjialiang

Issue #23375, opened 2024-08-03
Function to aggregate IP prefixes.


matt-calder:
## Summary
I'm proposing to contribute a function `ip_prefix_agg` which accepts an `ARRAY` of `IPPREFIX`s, collapses them into their minimal representation, and returns an `ARRAY` of new `IPPREFIX`s.

For example:
`IP_PREFIX_AGG(ARRAY[IPPREFIX '192.168.1.0/24', IPPREFIX '192.168.0.0/24', IPPREFIX '192.168.2.0/24', IPPREFIX '192.168.9.0/24'])` ==> `ARRAY[IPPREFIX '192.168.0.0/23', IPPREFIX '192.168.2.0/24', IPPREFIX '192.168.9.0/24']`

## Context
Many sizable organizations that have a presence on the Internet need to perform regular and ad-hoc network analytics for auditing, security (e.g., intrusion, bot, & DDoS detection), traffic engineering, performance monitoring, and capacity planning. While Presto currently supports the `IPADDRESS` and `IPPREFIX` types, only a few commonly functions are implemented. Aggregation of address space is critical when operating at Internet scale datasets. Not having aggregation within our primary SQL processing means we have to write tables out to disk, invoke functions from Python's `ipaddress.collapse_addresses` function, then reimport the results as a new table which is very inefficient and time consuming for 10s of millions of rows.

### Background
An IP prefix is a notation for representing a *network*, or group of IP addresses. For example, a */24*, such as 192.0.1.0/24, is 256 IP addresses in the range of 192.0.1.0 to 192.168.1.255. The /24 specifies a *prefix length* of 24-bits. To be a valid IP prefix, the prefix length part must fall on a valid IP address boundary. For example, 192.168.5.7/24 is not valid. This is enforced by the existing `IPPREFIX` type.

As a trivial example, 192.0.2.0/25, 192.0.2.128/25 can be aggregated to 192.0.2.0/24. However, we cannot just merge arbitrary neighboring IP blocks with the same prefix lengths because the starting prefix may not fall on a valid network boundary. For example, 192.168.1.0/24 and 192.168.2.0/24 cannot be aggregated to 192.168.1.0/23.

### Example
[...]

PR #23445, opened 2024-08-14
Function to aggregate IP prefixes

matt-calder:
## Description
A function `ip_prefix_collapse` which accepts an ARRAY of `IPPREFIX`s, collapses them into their minimal representation, and returns an `ARRAY` of new `IPPREFIX`s. Detailed background of problem in [https://github.com/prestodb/presto/issues/23375](https://github.com/prestodb/presto/issues/23375). 

## Motivation and Context
Addresses [https://github.com/prestodb/presto/issues/23375](https://github.com/prestodb/presto/issues/23375)

## Impact
This introduces a new user-facing SQL function called `ip_prefix_collapse`.

## Test Plan
unit tests

`mvn clean install -Dtest=TestIpPrefixFunctions -Dmaven.javadoc.skip=true -DskipUI -T1C -fn -pl presto-main`

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
*  Add function :func:`ip_prefix_collapse` ... :pr:`23445`
```

### DIFF
--- a/presto-docs/src/main/sphinx/functions/ip.rst
+++ b/presto-docs/src/main/sphinx/functions/ip.rst
@@ -49,3 +49,14 @@ IP Functions
         SELECT is_subnet_of(IPPREFIX '192.168.3.131/26', IPPREFIX '192.168.3.144/30'); -- true
         SELECT is_subnet_of(IPPREFIX '64:ff9b::17/64', IPPREFIX '64:ffff::17/64'); -- false
         SELECT is_subnet_of(IPPREFIX '192.168.3.131/26', IPPREFIX '192.168.3.131/26'); -- true
+
+.. function:: ip_prefix_collapse(array(ip_prefix)) -> array(ip_prefix)
+
+    Returns the minimal CIDR representation of the input ``IPPREFIX`` array.
+    Every ``IPPREFIX`` in the input array must be the same IP version (that is, only IPv4 or only IPv6)
+    or the query will fail and raise an error. ::
+
+        SELECT IP_PREFIX_COLLAPSE(ARRAY[IPPREFIX '192.168.0.0/24', IPPREFIX '192.168.1.0/24']); -- [{192.168.0.0/23}]
+        SELECT IP_PREFIX_COLLAPSE(ARRAY[IPPREFIX '2620:10d:c090::/48', IPPREFIX '2620:10d:c091::/48']); -- [{2620:10d:c090::/47}]
+        SELECT IP_PREFIX_COLLAPSE(ARRAY[IPPREFIX '192.168.1.0/24', IPPREFIX '192.168.0.0/24', IPPREFIX '192.168.2.0/24', IPPREFIX '192.168.9.0/24']); -- [{192.168.0.0/23}, {192.168.2.0/24}, {192.168.9.0/24}]
+
--- a/presto-main/src/main/java/com/facebook/presto/operator/scalar/IpPrefixFunctions.java
+++ b/presto-main/src/main/java/com/facebook/presto/operator/scalar/IpPrefixFunctions.java
@@ -24,9 +24,14 @@ import com.facebook.presto.spi.function.SqlType;
 import com.google.common.net.InetAddresses;
 import io.airlift.slice.Slice;
 
+import java.math.BigInteger;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 
+import static com.facebook.presto.operator.scalar.ArraySortFunction.sort;
 import static com.facebook.presto.spi.StandardErrorCode.GENERIC_INTERNAL_ERROR;
 import static com.facebook.presto.spi.StandardErrorCode.INVALID_FUNCTION_ARGUMENT;
 import static com.facebook.presto.type.IpAddressOperators.between;
@@ -44,6 +49,8 @@ import static java.lang.System.arraycopy;
 
 public final class IpPrefixFunctions
 {
+    private static final BigInteger TWO = BigInteger.valueOf(2);
+
     private IpPrefixFunctions() {}
 
     @Description("IP prefix for a given IP address and subnet size")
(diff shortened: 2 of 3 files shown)


presto-docs/src/main/sphinx/functions/ip.rst
```diff
@@ -49,3 +49,13 @@ IP Functions
         SELECT is_subnet_of(IPPREFIX '192.168.3.131/26', IPPREFIX '192.168.3.144/30'); -- true
         SELECT is_subnet_of(IPPREFIX '64:ff9b::17/64', IPPREFIX '64:ffff::17/64'); -- false
         SELECT is_subnet_of(IPPREFIX '192.168.3.131/26', IPPREFIX '192.168.3.131/26'); -- true
+
+.. function:: ip_prefix_agg(array(ip_prefix)) -> array(ip_prefix)
+
+    Returns the minimal CIDR representation of the input ``IPPREFIX`` array.
+    All ``IPPREFIX``s in the input array must have the same IP version (i.e., only IPv4 or only IPv6). ::
```
steveburnett, line 56:
```suggestion
    Every ``IPPREFIX`` in the input array must be the same IP version (that is, only IPv4 or only IPv6). ::
```
* Adding an `s` after the ``IPPREFIX`` here causes the following error in a local doc build: 

  `/Users/steveburnett/Documents/GitHub/presto/presto-docs/src/main/sphinx/functions/ip.rst:55: WARNING: Inline literal start-string without end-string.` 
  Suggest rephrase to avoid the warning. 

* Do not use Latin abbreviations. See the [GitLab documentation style guide recommended word list entry for i.e.](https://docs.gitlab.com/ee/development/documentation/styleguide/word_list.html#ie).

steveburnett (changes requested):
Thanks for the doc! Just two nits, with suggested changes.

steveburnett (dismissed):
LGTM! (docs) 

Pull updated branch, new local doc build, docs look good. Thanks!

elharo (commented):
This seems quite specific. Are there functions like this in any other SQL implementation? Can this be an extension function instead of a core function?

presto-docs/src/main/sphinx/functions/ip.rst
```diff
@@ -49,3 +49,13 @@ IP Functions
         SELECT is_subnet_of(IPPREFIX '192.168.3.131/26', IPPREFIX '192.168.3.144/30'); -- true
         SELECT is_subnet_of(IPPREFIX '64:ff9b::17/64', IPPREFIX '64:ffff::17/64'); -- false
         SELECT is_subnet_of(IPPREFIX '192.168.3.131/26', IPPREFIX '192.168.3.131/26'); -- true
+
+.. function:: ip_prefix_agg(array(ip_prefix)) -> array(ip_prefix)
+
+    Returns the minimal CIDR representation of the input ``IPPREFIX`` array.
+    Every ``IPPREFIX`` in the input array must be the same IP version (that is, only IPv4 or only IPv6). ::
```
elharo, line 56: what if that isn't true? What happens?

presto-main/src/main/java/com/facebook/presto/operator/scalar/IpPrefixFunctions.java
```diff
@@ -44,6 +49,8 @@
 
 public final class IpPrefixFunctions
 {
+    static final BigInteger TWO = BigInteger.valueOf(2);
```
elharo, line 52: private

presto-main/src/main/java/com/facebook/presto/operator/scalar/IpPrefixFunctions.java
```diff
@@ -137,6 +144,227 @@ public static boolean isPrefixSubnetOf(@SqlType(StandardTypes.IPPREFIX) Slice fi
...
+                IPPREFIX.writeSlice(blockBuilder, ipPrefix);
+            }
+        }
+
+        return blockBuilder.build();
+    }
+
+    private static List<Slice> generateMinIpPrefixes(BigInteger firstIpAddress, BigInteger lastIpAddress, int ipVersionMaxBits)
+    {
+        if (firstIpAddress.compareTo(lastIpAddress) > 0) {
```
elharo, line 202: Isn't this guaranteed by sorting above?

matt-calder:
Thank you for the feedback.

> This seems quite specific. 

This function is commonly found in libraries for manipulating networking data and serves a real need big data networking analytics. The most popular one in the analytics work is probably Python's `ipaddress.collapse_addresses` function. However, there are a number of implementations:
* [ip range](https://manpages.org/iprange)
* [aggregate6](https://pypi.org/project/aggregate6/)
* [netaddr.cidr_merge](https://netaddr.readthedocs.io/en/latest/api.html?highlight=merge#netaddr.cidr_merge)
* [https://pypi.org/project/aggregate-prefixes/](https://pypi.org/project/aggregate-prefixes/)
* [caprica](https://github.com/google/capirca/blob/master/capirca/lib/summarizer.py)

> Are there functions like this in any other SQL implementation? 

Yes, PostgreSQL's [inet_merge](https://pgpedia.info/i/inet_merge.html) is similar.

> Can this be an extension function instead of a core function?

* I think since there [are existing `IPPREFIX` functions](https://prestodb.io/docs/current/functions/ip.html) (e.g., `is_subnet_of`, `ip_subnet_range`), it makes sense to keep them together.
* The `IPADDRESS` and `IPPREFIX` types and supporting functions are not in presto-common, only presto-main, which prevents writing extension functions for them (classes in presto-main are on the classpath as test-only).

presto-docs/src/main/sphinx/functions/ip.rst
matt-calder: Updated doc.

presto-main/src/main/java/com/facebook/presto/operator/scalar/IpPrefixFunctions.java
```diff
@@ -137,6 +144,227 @@ public static boolean isPrefixSubnetOf(@SqlType(StandardTypes.IPPREFIX) Slice fi
         return between(ipSubnetMin(second), ipSubnetMin(first), ipSubnetMax(first)) && between(ipSubnetMax(second), ipSubnetMin(first), ipSubnetMax(first));
     }
 
+    @Description("Aggregates an array of IP prefixes into the minimum set of IP prefixes.")
```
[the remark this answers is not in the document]
matt-calder: I tried to rephrase the description that the set is determined by the CIDR representation of IP ranges.

presto-main/src/main/java/com/facebook/presto/operator/scalar/IpPrefixFunctions.java
matt-calder: Right. I've removed this condition.

elharo:
Yes, but Presto is not a library for manipulating networking data. It's a SQL database (or execution engine, or analyzer, or whatever). There are many things commonly found in non-core libraries for Java/Python/Rust/etc. that aren't right for a SQL library. We would not, for example, want to include equivalents of everything in the standard Java class libraries, much less everything in Apache Commons, Guava, Codehaus, and many others. 

The more I think about this the more I'm convinced this should be a UDF and not something we ship in core. If you make it a UDF, you can use it tomorrow for whatever you need without waiting for any approval here.

matt-calder:
> Yes, but Presto is not a library for manipulating networking data. It's a SQL database (or execution engine, or analyzer, or whatever). There are many things commonly found in non-core libraries for Java/Python/Rust/etc. that aren't right for a SQL library. We would not, for example, want to include equivalents of everything in the standard Java class libraries, much less everything in Apache Commons, Guava, Codehaus, and many others.

It isn't manipulating, it is analyzing. And Presto and all other major SQL engines support functions for analyzing networking data. 

> 
> The more I think about this the more I'm convinced this should be a UDF and not something we ship in core. If you make it a UDF, you can use it tomorrow for whatever you need without waiting for any approval here.

It cannot be made a UDF for the reason I mentioned previously -- The IPADDRESS and IPPREFIX types and supporting functions are not in presto-common, only presto-main, which prevents writing extension functions for them (classes in presto-main are on the classpath as test-only).

kaikalur:
@NikhilCollooru  this is a useful function. Let's review it and merge it ASAP

kaikalur:
> > Yes, but Presto is not a library for manipulating networking data. It's a SQL database (or execution engine, or analyzer, or whatever). There are many things commonly found in non-core libraries for Java/Python/Rust/etc. that aren't right for a SQL library. We would not, for example, want to include equivalents of everything in the standard Java class libraries, much less everything in Apache Commons, Guava, Codehaus, and many others.
> 
> It isn't manipulating, it is analyzing. And Presto and all other major SQL engines support functions for analyzing networking data.
> 
> > The more I think about this the more I'm convinced this should be a UDF and not something we ship in core. If you make it a UDF, you can use it tomorrow for whatever you need without waiting for any approval here.
> 
> It cannot be made a UDF for the reason I mentioned previously -- The IPADDRESS and IPPREFIX types and supporting functions are not in presto-common, only presto-main, which prevents writing extension functions for them (classes in presto-main are on the classpath as test-only).

Given we have IPAddress as a built-in type - this is quite reasonable ot have it as part of built-in.

kaikalur:
Also this subtle distinction between udf vs builtin is irrelevant for this PR. If/when there is a clear definition we can refactor it. Let's unblock this here for now.

elharo:
Is it feasible to move the IPADDRESS and IPPREFIX types to presto-common? I.e. is it a simple refactor or does it have other implications and effects?

presto-docs/src/main/sphinx/functions/ip.rst
```diff
@@ -49,3 +49,14 @@ IP Functions
         SELECT is_subnet_of(IPPREFIX '192.168.3.131/26', IPPREFIX '192.168.3.144/30'); -- true
         SELECT is_subnet_of(IPPREFIX '64:ff9b::17/64', IPPREFIX '64:ffff::17/64'); -- false
         SELECT is_subnet_of(IPPREFIX '192.168.3.131/26', IPPREFIX '192.168.3.131/26'); -- true
+
+.. function:: ip_prefix_agg(array(ip_prefix)) -> array(ip_prefix)
+
+    Returns the minimal CIDR representation of the input ``IPPREFIX`` array.
+    Every ``IPPREFIX`` in the input array must be the same IP version (that is, only IPv4 or only IPv6)
+    or the query will fail and raise an error::
```
elharo, line 57: Is there a reason for the double colon here or is that a typo?

presto-main/src/main/java/com/facebook/presto/operator/scalar/IpPrefixFunctions.java
```diff
@@ -137,6 +144,222 @@ public static boolean isPrefixSubnetOf(@SqlType(StandardTypes.IPPREFIX) Slice fi
         return between(ipSubnetMin(second), ipSubnetMin(first), ipSubnetMax(first)) && between(ipSubnetMax(second), ipSubnetMin(first), ipSubnetMax(first));
     }
 
+    @Description("Combines the input set of IP prefixes into the smallest contiguous CIDR ranges possible.")
```
elharo, line 147: Is this smallest or fewest? That is, is it returning a single smallest range that covers everything (smallest range)  or a list of ranges that covers everything (fewest ranges)?

presto-main/src/main/java/com/facebook/presto/operator/scalar/IpPrefixFunctions.java
```diff
@@ -137,6 +144,222 @@ public static boolean isPrefixSubnetOf(@SqlType(StandardTypes.IPPREFIX) Slice fi
...
+    {
+        List<Slice> ipPrefixSlices = new ArrayList<>();
+
+        // i.e., while firstIpAddress <= lastIpAddress
+        while (firstIpAddress.compareTo(lastIpAddress) <= 0) {
+            long rangeBits = findRangeBits(firstIpAddress, lastIpAddress); // find the number of bits for the next prefix in the range
+            int prefixLength = (int) (ipVersionMaxBits - rangeBits);
+
+            InetAddress asInetAddress;
+            try {
```
elharo, line 210: use a longer try block to avoid splitting declaration and assignment. All code that depends on asInetAddress being set can go into the try block.

presto-docs/src/main/sphinx/functions/ip.rst
matt-calder: fixed.

presto-main/src/main/java/com/facebook/presto/operator/scalar/IpPrefixFunctions.java
matt-calder: Agreed -- "fewest ranges" is better.

matt-calder:
> Is it feasible to move the IPADDRESS and IPPREFIX types to presto-common? I.e. is it a simple refactor or does it have other implications and effects?

I briefly looked into it but it was outside my depth for how all the presto projects fit together.

tdcmeehan:
> Given we have IPAddress as a built-in type - this is quite reasonable ot have it as part of built-in.

Agreed.  We already support the IPADDRESS type, and this seems like a reasonable thing to do with IP addresses.  Also, for reference, Postgres supports this type built in as well.

The naming might be confusing.  AFAIA `_agg` functions alway represent aggregation functions, but technically this is a scalar function.  I wonder if something like `ip_prefix_collapse` or `ip_prefix_union` is more appropriate?

presto-main/src/main/java/com/facebook/presto/operator/scalar/IpPrefixFunctions.java
```diff
@@ -137,6 +144,220 @@ public static boolean isPrefixSubnetOf(@SqlType(StandardTypes.IPPREFIX) Slice fi
         return between(ipSubnetMin(second), ipSubnetMin(first), ipSubnetMax(first)) && between(ipSubnetMax(second), ipSubnetMin(first), ipSubnetMax(first));
     }
 
+    @Description("Combines the input set of IP prefixes into the fewest contiguous CIDR ranges possible.")
+    @ScalarFunction("ip_prefix_agg")
```
feilong-liu, line 148: Can you rename to get rid of "agg" to avoid confusion?
Overall lgtm. Thanks for the contribution.

feilong-liu (approved):

feilong-liu (approved):

PR #23449, opened 2024-08-15
[native] Validate configs used by hive connector

pramodsatya:
## Description
Validates properties used by the hive connector, throws a warning if a config will not be used. 
The hive connector configs are typically defined in `etc/hive.properties`.

## Motivation and Context
Ensure invalid configs are detected.

## Test Plan
Verified the warning message is present in presto_server logs.

```
== NO RELEASE NOTE ==
```


presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -75,6 +75,58 @@ constexpr char const* kTaskUriFormat =
     "{}://{}:{}"; // protocol, address and port
 constexpr char const* kConnectorName = "connector.name";
 
+// Set of valid configs for the hive connector.
+const std::unordered_set<std::string> kHiveConfigs = {
```
czentgr, line 79: This file keeps growing ...
I almost think that the Connectors should be refactored a bit - i.e. move `class HivePrestoToVeloxConnector final : public PrestoToVeloxConnector` to their own file.
And there this list and validation can be added. In this case everything is together - the config for the hive connector. 
Or something that logically combines them as applicable together. Just adding them here is a bit too specific.

@majetideepak What do you think?

pramodsatya: Thanks for the suggestion, updated accordingly.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -1094,6 +1096,18 @@ std::vector<std::string> PrestoServer::registerConnectors(
       // make sure connector type is supported
       getPrestoToVeloxConnector(connectorName);
 
+      // Validate hive configs.
+      if (connectorName == kHive || connectorName == kHiveHadoop2) {
```
aditi-pandit, line 1100: Why is the validation done in this manner ? It could be part of the HivePrestoToVeloxConnector constructor itself.

presto-native-execution/presto_cpp/main/types/PrestoToVeloxConnector.h
```diff
@@ -34,6 +35,26 @@ void unregisterPrestoToVeloxConnector(const std::string& connectorName);
 const PrestoToVeloxConnector& getPrestoToVeloxConnector(
     const std::string& connectorName);
 
+velox::dwio::common::FileFormat toVeloxFileFormat(
```
aditi-pandit, line 38: Might be worth moving all the Iceberg related processing that transforms to Hive structures in the HivePrestoToVeloxConnector classes as well.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -36,7 +36,7 @@
 #include "presto_cpp/main/operators/PartitionAndSerialize.h"
 #include "presto_cpp/main/operators/ShuffleRead.h"
 #include "presto_cpp/main/operators/UnsafeRowExchangeSource.h"
-#include "presto_cpp/main/types/PrestoToVeloxConnector.h"
+#include "presto_cpp/main/types/HivePrestoToVeloxConnector.h"
```
aditi-pandit, line 39: We get by here because HivePrestoToVeloxConnector includes PrestoToVeloxConnector. But maybe we should still retain the older include for other direct connectors as well.

aditi-pandit (commented):
Thanks @pramodsatya  for this refactoring. Its a good idea.

aditi-pandit:
@pramodsatya : Separate the refactoring and HiveConfig validation into 2 separate commits (not PRs). Would be logically cleaner.

pramodsatya (commented):
Thanks for the feedback @aditi-pandit @czentgr. I have addressed the comments, could you please take another look?

presto-native-execution/presto_cpp/main/types/HivePrestoToVeloxConnector.h
```diff
@@ -0,0 +1,111 @@
...
+    const std::shared_ptr<protocol::RowExpression>& remainingPredicate,
+    bool isPushdownFilterEnabled,
+    const std::string& tableName,
+    const protocol::List<protocol::Column>& dataColumns,
+    const protocol::TableHandle& tableHandle,
+    const protocol::Map<protocol::String, protocol::String>& tableParameters,
+    const VeloxExprConverter& exprConverter,
+    const TypeParser& typeParser);
+
+void validateHiveConfigs(
```
aditi-pandit, line 50: Why is this an external method ? It can be made local to the cpp file in an anonymous namespace. Same for all above methods as well. They are only used by HivePrestoToVeloxConnector. Only the public interface of HivePrestoToVeloxConnector should be made public imo.

aditi-pandit (commented):
@pramodsatya : Had one high level question. Will do a more line by line review after you address this.

presto-native-execution/presto_cpp/main/types/HivePrestoToVeloxConnector.h
pramodsatya: Thanks for the suggestion, moved `validateHiveConfigs` to anonymous namespace and removed it from header. The rest of the functions, `toHiveTableHandle`, `toRequiredSubfields`, `toHiveColumnType` and `toVeloxFileFormat` are also used by the respective functions in `IcebergPrestoToVeloxConnector`, hence they are exposed in the header file.

pramodsatya (commented):
Thanks for the feedback @aditi-pandit, addressed the comment. Could you please take another look?

presto-native-execution/presto_cpp/main/types/HivePrestoToVeloxConnector.h
aditi-pandit: @pramodsatya : It might be better to move IcebergPrestoToVeloxConnector to HivePrestoToVeloxConnector.h(cpp) files as well. wdyt ?

pramodsatya: @aditi-pandit, thanks for the suggestion, but I feel this would result in a file similar to the current PrestoToVeloxConnector.h(.cpp) files. The intent with this refactor was to breakdown the large file `PrestoToVeloxConnector.cpp`. How do you suggest we proceed with the refactor?

czentgr: Maybe they should be refactored to their own file. PrestoToVeloxConnector being the base and then the individual connectors should be in their own file. So Iceberg would go separately.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -250,10 +253,37 @@ void PrestoServer::run() {
...
 
-  registerPrestoToVeloxConnector(
-      std::make_unique<HivePrestoToVeloxConnector>("hive"));
-  registerPrestoToVeloxConnector(
-      std::make_unique<HivePrestoToVeloxConnector>("hive-hadoop2"));
+  // Update connectorToCatalogInfo_ so the connector configs can be validated
+  // when the connectors are registered.
+  getCatalogConfigs();
+
+  // Update hiveConfigs map for validation when the connector is registered.
```
czentgr, line 260: I think what is confusing to me is that it says "update". But in fact, all that happens here is taking all the various catalog configurations from a per catalog lookup into a single vector to be passed into the connector constructor later on. Nothing is updated (changed), they are only collected into a single vector for easier processing.
Perhaps the comment and naming should reflect that.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -1150,22 +1164,48 @@ std::vector<std::string> PrestoServer::registerConnectors(
...
+        std::make_shared<folly::NamedThreadFactory>("Connector"));
+    PRESTO_STARTUP_LOG(INFO)
+        << "Connector IO executor has " << connectorIoExecutor_->numThreads()
+        << " threads.";
+  }
+
+  std::vector<std::string> catalogNames;
+  for (const auto& entry : connectorToCatalogInfo_) {
+    auto connectorName = entry.first;
+    // make sure connector type is supported
```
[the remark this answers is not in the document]
pramodsatya: Yes it was copied from the original code, should this comment be removed?

presto-native-execution/presto_cpp/main/PrestoServer.cpp
pramodsatya: Thanks for the suggestion, updated the comment and variable names accordingly.

pramodsatya (commented):
Thanks for the feedback @czentgr @aditi-pandit, addressed the comments. Could you please take another look?

presto-native-execution/presto_cpp/main/types/HivePrestoToVeloxConnector.cpp
```diff
@@ -0,0 +1,1326 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "presto_cpp/main/types/HivePrestoToVeloxConnector.h"
+#include <velox/type/fbhive/HiveTypeParser.h>
```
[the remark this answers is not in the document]
pramodsatya: Thanks for pointing it out, this was copied from the existing file where `<>` was used for this header and it is fixed now.

Issue #8713, opened 2017-08-10
Presto use hive data source, time will change?


annisun:
Presto use hive data source, time will change?
In hive,the time is '2017-08-10 00:00:00.000',
but in presto, the time changes like this : '2017-08-10 08:00:00.000'.

Where can I configure it?

fiedukow:
The way TIMESTAMP from Hive is displayed in presto currently can be determined by two factors.
 1) CLI timezone - the value is adjusted by Presto to timezone demanded by user. Timezone can be chosen by passing `-Duser.timezone=<YourTZ>[1]` to jvm while starting cli executable. By default it's of course determined by client operating system settings.
Example CLI start: `java -jar -Duser.timezone=Asia/Kathmandu presto-cli-0.183-SNAPSHOT-executable.jar`
 2) Presto Server timezone - *it's recommended to align TZ in which Presto Server JVM works with timezone used by Hive*. You can do that by applying same flag as to CLI but to the server (`-Duser.timezone` may be added to jvm.config). If for some reason you cannot do that, you can preserve some functionality by setting `hive.time-zone` connector property to timezone of Hive. In that setup you will not be able to do inserts into hive though. Please note that if you will not follow one of those steps Presto configuration is considered invalid, and for that reason the behaviour is undefined and untested - you are basically lying to Presto about what timezone does hive operate in.

To summarise:
Make sure that presto server is configured properly. If you are sure that the configuration is correct, you can adjust timezone in which TIMESTAMP is displayed(!) by setting CLI timezone.

Note that current TIMESTAMP behaviour may change in the future (#7480).

[1] List of possible timezones for JVM: https://garygregory.wordpress.com/2013/06/18/what-are-the-java-timezone-ids/

annisun:
thanks  !
I will try it

fiedukow:
Closing for now. Feel free to reopen if required.

PR #23452, opened 2024-08-15
Use the JVM timezone in HiveQueryRunner

zuyu:
## Description
`HiveMetaData` only checks the JVM time zone:
https://github.com/prestodb/presto/blob/4f74f52f36709944ba9eea18b0559ec7dd691d36/presto-hive/src/main/java/com/facebook/presto/hive/HiveMetadata.java#L3375

The `TIME_ZONE` constant in `HiveQueryRunner` does not work when the time zone was set differently. In other words, the `TIME_ZONE` constant is useless, thus removing it.
https://github.com/prestodb/presto/blob/4f74f52f36709944ba9eea18b0559ec7dd691d36/presto-hive/src/test/java/com/facebook/presto/hive/HiveQueryRunner.java#L86

On the other hand, per #8713, feeding `-Duser.timezone=America/Bahia_Banderas` as a JVM argument should work, but still sees the assertion failure below. Do NOT know why.

## Motivation and Context
Before this PR, running `HiveExternalWorkerQueryRunner` would fail on a machine w/ a different time zone, even w/ adding `-Duser.timezone=America/Bahia_Banderas` as a JVM argument in IntelliJ IDEA.
```sh
2024-08-16T08:15:40.224+0800	INFO	main	stderr	Exception in thread "main" 
2024-08-16T08:15:40.224+0800	INFO	main	stderr	java.lang.AssertionError: Timezone not configured correctly. Add -Duser.timezone=America/Bahia_Banderas to your JVM arguments expected [America/Bahia_Banderas] but found [Asia/Seoul]
2024-08-16T08:15:40.225+0800	INFO	main	stderr		at org.testng.Assert.fail(Assert.java:110)
2024-08-16T08:15:40.225+0800	INFO	main	stderr		at org.testng.Assert.failNotEquals(Assert.java:1413)
2024-08-16T08:15:40.225+0800	INFO	main	stderr		at org.testng.Assert.assertEqualsImpl(Assert.java:149)
2024-08-16T08:15:40.225+0800	INFO	main	stderr		at org.testng.Assert.assertEquals(Assert.java:131)
2024-08-16T08:15:40.225+0800	INFO	main	stderr		at com.facebook.presto.hive.HiveQueryRunner.createQueryRunner(HiveQueryRunner.java:189)
2024-08-16T08:15:40.225+0800	INFO	main	stderr		at com.facebook.presto.hive.HiveQueryRunner.createQueryRunner(HiveQueryRunner.java:161)
2024-08-16T08:15:40.225+0800	INFO	main	stderr		at com.facebook.presto.hive.HiveQueryRunner.createQueryRunner(HiveQueryRunner.java:145)
2024-08-16T08:15:40.225+0800	INFO	main	stderr		at com.facebook.presto.hive.HiveQueryRunner.createQueryRunner(HiveQueryRunner.java:130)
2024-08-16T08:15:40.225+0800	INFO	main	stderr		at com.facebook.presto.nativeworker.PrestoNativeQueryRunnerUtils.createJavaQueryRunner(PrestoNativeQueryRunnerUtils.java:138)
2024-08-16T08:15:40.225+0800	INFO	main	stderr		at com.facebook.presto.nativeworker.PrestoNativeQueryRunnerUtils.createJavaQueryRunner(PrestoNativeQueryRunnerUtils.java:121)
2024-08-16T08:15:40.225+0800	INFO	main	stderr		at com.facebook.presto.nativeworker.PrestoNativeQueryRunnerUtils.createJavaQueryRunner(PrestoNativeQueryRunnerUtils.java:115)
2024-08-16T08:15:40.225+0800	INFO	main	stderr		at com.facebook.presto.nativeworker.PrestoNativeQueryRunnerUtils.createJavaQueryRunner(PrestoNativeQueryRunnerUtils.java:103)
2024-08-16T08:15:40.225+0800	INFO	main	stderr		at com.facebook.presto.nativeworker.HiveExternalWorkerQueryRunner.main(HiveExternalWorkerQueryRunner.java:32)
```

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-hive/src/test/java/com/facebook/presto/hive/HiveQueryRunner.java
+++ b/presto-hive/src/test/java/com/facebook/presto/hive/HiveQueryRunner.java
@@ -65,7 +65,7 @@ import static com.facebook.presto.testing.TestingSession.testSessionBuilder;
 import static com.facebook.presto.tests.QueryAssertions.copyTables;
 import static com.facebook.presto.tpch.TpchMetadata.TINY_SCHEMA_NAME;
 import static java.util.Locale.ENGLISH;
-import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertNotNull;
 
 public final class HiveQueryRunner
 {
@@ -83,7 +83,6 @@ public final class HiveQueryRunner
     public static final String TPCDS_BUCKETED_SCHEMA = "tpcds_bucketed";
     public static final MetastoreContext METASTORE_CONTEXT = new MetastoreContext("test_user", "test_queryId", Optional.empty(), Collections.emptySet(), Optional.empty(), Optional.empty(), false, HiveColumnConverterProvider.DEFAULT_COLUMN_CONVERTER_PROVIDER, WarningCollector.NOOP, new RuntimeStats());
     private static final String TEMPORARY_TABLE_SCHEMA = "__temporary_tables__";
-    private static final DateTimeZone TIME_ZONE = DateTimeZone.forID("America/Bahia_Banderas");
 
     public static DistributedQueryRunner createQueryRunner(TpchTable<?>... tables)
             throws Exception
@@ -186,7 +185,9 @@ public final class HiveQueryRunner
             boolean addJmxPlugin)
             throws Exception
     {
-        assertEquals(DateTimeZone.getDefault(), TIME_ZONE, "Timezone not configured correctly. Add -Duser.timezone=America/Bahia_Banderas to your JVM arguments");
+        DateTimeZone timeZone = DateTimeZone.getDefault();
+        assertNotNull(timeZone, "Timezone not configured correctly. Add -Duser.timezone=America/Bahia_Banderas to your JVM arguments");
+
         setupLogging();
 
         Map<String, String> systemProperties = ImmutableMap.<String, String>builder()
@@ -228,7 +229,7 @@ public final class HiveQueryRunner
 
             Map<String, String> hiveProperties = ImmutableMap.<String, String>builder()
                     .putAll(extraHiveProperties)
-                    .put("hive.time-zone", TIME_ZONE.getID())
+                    .put("hive.time-zone", timeZone.getID())
                     .put("hive.security", security)
                     .put("hive.max-partitions-per-scan", "1000")
                     .put("hive.assume-canonical-partition-keys", "true")


tdcmeehan:
I think the reason we hardcode the timezone in the `HiveQueryRunner` is because `air.test.timezone` is set to `America/Bahia_Banderas`, and there's a specific reason for that: 

https://github.com/prestodb/presto/blob/52c5a38357b9eac05cc6500488a40907f8bfa840/pom.xml#L84-L91

The problem I see with this PR is if you're trying to reproduce a problem in CI, then really we should only allow `America/Bahia_Banderas`, and because many tests implicitly use `HiveQueryRunner`, I'm just worried that it might cause confusion and long debugging time for timezone related tests (why they are behaving differently in local IntelliJ runs vs. CI).  And for that we should be opinionated.

I don't understand the problem you mentioned here:

> On the other hand, per https://github.com/prestodb/presto/issues/8713, feeding -Duser.timezone=America/Bahia_Banderas as a JVM argument should work, but still sees the assertion failure below. Do NOT know why.

I believe this is a problem specific to you.  To check this, I noted that when I run my tests, they are getting set with `America/Bahia_Banderas`, and when I manually add a different one like `America/New_York`, they fail.  I'm wondering if you are adding this as a VM argument to the HiveQueryRunner?

image

zuyu:
Thank @tdcmeehan for taking a look.

The problem for me was that, running `HiveExternalWorkerQueryRunner` Application always fails due to the time zone assertion
https://github.com/prestodb/presto/blob/be299743bf1b2feaaf0fbfaa3d09ce052908a22b/presto-hive/src/test/java/com/facebook/presto/hive/HiveQueryRunner.java#L189

```
java.lang.AssertionError: Timezone not configured correctly. Add -Duser.timezone=America/Bahia_Banderas to your JVM arguments expected [America/Bahia_Banderas] but found [America/Chicago]
```

My local time zone was set differently as `America/Bahia_Banderas`:
Time-zone-setup

And `HiveExternalWorkerQueryRunner` Application JVM setup does not help: removing the time zone argument, or set it to a non-existing one like `Asia/Dummy` has no effects at all.
https://github.com/prestodb/presto/blob/be299743bf1b2feaaf0fbfaa3d09ce052908a22b/presto-native-execution/README.md#L133

With the above info, could you reproduce the assertion failure on your end? Thank you!

Note that I have not tried any tests yet (and I was surprised that the CI were all green w/ this PR).

tdcmeehan:
@zuyu yes, I am able to prove that the `HiveExternalWorkerQueryRunner` fails to start up if I supply any other timezone besides `America/Bahia_Banderas`.  You mentioned you were running this in IntelliJ, and I wanted to confirm, did you add `-Duser.timezone=America/Bahia_Banderas` _as a VM argument_, and not for example a program argument?  A VM argument is a JVM specific argument, and I don't see how it's possible for this to have not taken effect.  I suspect you're setting it incorrectly--please share a screenshot of your IntelliJ run configuration.  Thanks!

zuyu:
Oh, closing this PR, as I did misconfigure the `VM options` as `program arguments` in IntelliJ. Thank you for pointing it out, @tdcmeehan!

Issue #23450, opened 2024-08-15
Incorrect and inconsistent behavior of Like with constant folding


kagamiori:
I see inconsistent behaviors between the two queries below.

Query1: select if(c2 like '' escape '', c0, c1)  from (values (true, false, 'T')) t(c0, c1, c2);
This query throws an exception "Escape string must be a single character".

Query2: select if('T' like '' escape '', c0, c1) from (values (true, false)) t(c0, c1);
This query returns `false` with no error.

Since Query1 and Query2 are semantically equivalent, I expect them to have the same behavior.

The different behavior is because Query1 takes this branch where the `isLikePattern()` method throws the exception.
https://github.com/prestodb/presto/blob/4f74f52f36709944ba9eea18b0559ec7dd691d36/presto-main/src/main/java/com/facebook/presto/sql/planner/RowExpressionInterpreter.java#L936-L937

On the other hand, Query2 takes this branch that doesn't call isLikePattern.
https://github.com/prestodb/presto/blob/4f74f52f36709944ba9eea18b0559ec7dd691d36/presto-main/src/main/java/com/facebook/presto/sql/planner/RowExpressionInterpreter.java#L912

I think Query1 has the correct behavior while Query2 needs to be fixed.

## Your Environment

* Presto version used: Presto-0.289
* Storage (HDFS/S3/GCS..):
* Data source and connector used:
* Deployment (Cloud or On-prem):
* [Pastebin](https://pastebin.com/) link to the complete debug logs:

kagamiori:
cc @kaikalur @rschlussel @tdcmeehan

tdcmeehan:
I think all we need to do is add this validation:

https://github.com/prestodb/presto/blob/04138bc26a4fc801630fe936463d89d071d00c0a/presto-main/src/main/java/com/facebook/presto/type/LikeFunctions.java#L124

To right after this line as well:

https://github.com/prestodb/presto/blob/04138bc26a4fc801630fe936463d89d071d00c0a/presto-main/src/main/java/com/facebook/presto/type/LikeFunctions.java#L151

PR #23456, opened 2024-08-15
Unescape like pattern, and check for one character escape string when in row expression interpreter

auden-woolfson:
## Description
fix bug described in issue #23450 

```
== RELEASE NOTES ==

General Changes
* Fix bug to unescape like pattern and validate escape string with no unresolved value :pr:`23456`
```

### DIFF
--- a/presto-tests/src/main/java/com/facebook/presto/tests/TestLikeQueries.java
+++ b/presto-tests/src/main/java/com/facebook/presto/tests/TestLikeQueries.java
@@ -1,0 +1,43 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.tests;
+
+import com.facebook.presto.Session;
+import com.facebook.presto.testing.QueryRunner;
+import org.testng.annotations.Test;
+
+import static com.facebook.presto.testing.TestingSession.testSessionBuilder;
+
+public class TestLikeQueries
+        extends AbstractTestQueryFramework
+{
+    @Override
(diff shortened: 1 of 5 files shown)


steveburnett:
Nit suggestion to revise release note entry for consistency with the [Order of Changes](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines#order-of-changes) in the Release Notes Guidelines. 

```
== RELEASE NOTES ==

General Changes
* Fix bug to unescape like pattern and validate escape string with no unresolved value :pr:`23456`
```

tdcmeehan:
Thanks for the quick work. Can you please add a test?

tdcmeehan:
This looks good, but I think we should add one more test that shows this works in a query.

presto-tests/src/main/java/com/facebook/presto/tests/TestLikeQueries.java
```diff
@@ -0,0 +1,43 @@
...
+    {
+        Session session = testSessionBuilder()
+                .setCatalog("test")
+                .setSchema("default")
+                .build();
+        return new DistributedQueryRunner(session, 1);
+    }
+
+    @Test
+    public void testLikeQueriesWithEscape()
```
elharo, line 36: This should be 4 separate tests that can pass or fail independently.

presto-tests/src/main/java/com/facebook/presto/tests/TestLikeQueries.java
```diff
@@ -0,0 +1,43 @@
...
+package com.facebook.presto.tests;
+
+import com.facebook.presto.Session;
+import com.facebook.presto.testing.QueryRunner;
+import org.testng.annotations.Test;
+
+import static com.facebook.presto.testing.TestingSession.testSessionBuilder;
+
+public class TestLikeQueries
+        extends AbstractTestQueries
```
tdcmeehan, line 23: This will create a new suite of tests derived from the parent class.  Perhaps instead inherit from `AbstractTestQueryFramework`, which won't do this.

tdcmeehan (changes requested):

tdcmeehan:
Looks good.  Please squash commits and check our contributing guide for commit message guidelines.

tdcmeehan:
@auden-woolfson there is a test failure

auden-woolfson:
> @auden-woolfson there is a test failure

There are 4 test cases that assert like functions with empty escapes in `TestConditions`. Should I just remove these? Or assert that they fail?

tdcmeehan:
LGTM.  Please rebase to fix the conflict.

tdcmeehan (approved):

PR #23459, opened 2024-08-16
[native] Fill protocol::TaskStats::completedDrivers with finished splits

tanjialiang:
Presto Java reports number of drivers to number of splits in Presto
UI because split and driver are 1 to 1 mapping relationship. This is not
true in Prestissimo where 1 driver handles many splits. In order to
quickly unblock developers from viewing the correct number of completed
 splits in Prestissimo's coordinator UI, we put numFinishedSplits in
protocol::TaskStats::completedDrivers. We should really extend the API
(protocol::TaskStats and Presto coordinator UI) to have splits information
as a proper fix.
```
== NO RELEASE NOTE ==
```


Yuhta (approved):

xiaoxmeng (approved):
@tanjialiang thanks for the fix!

PR #23480, opened 2024-08-20
[native] Refactor TaskManagerTest to not create common::WriterOptions instance

kewang1024:
We want to make [common::WriterOptions::processConfigs](https://github.com/facebookincubator/velox/pull/10762/files#diff-5a2dd3766d9a74bbef58d62d96f0abfb111e8e507ce9bcecd35f69d2c8669ed7R623) function pure virtual 
to make sure every data format will implement its own.
As a result, we should not create dwio::common::WriterOptions instance anymore

### DIFF
--- a/presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp
+++ b/presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp
@@ -29,6 +29,7 @@
 #include "velox/dwio/common/FileSink.h"
 #include "velox/dwio/common/WriterFactory.h"
 #include "velox/dwio/common/tests/utils/BatchMaker.h"
+#include "velox/dwio/dwrf/writer/Writer.h"
 #include "velox/exec/Exchange.h"
 #include "velox/exec/Values.h"
 #include "velox/exec/tests/utils/PlanBuilder.h"
@@ -270,6 +271,8 @@ class TaskManagerTest : public testing::Test {
         "http://{}:{}",
         serverAddress.getAddressStr(),
         serverAddress.getPort()));
+    writerFactory_ =
+        dwio::common::getWriterFactory(dwio::common::FileFormat::DWRF);
   }
 
   void TearDown() override {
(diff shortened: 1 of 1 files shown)


xiaoxmeng (dismissed):

presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp
```diff
@@ -29,6 +29,7 @@
 #include "velox/dwio/common/FileSink.h"
 #include "velox/dwio/common/WriterFactory.h"
 #include "velox/dwio/common/tests/utils/BatchMaker.h"
+#include "velox/dwio/dwrf/writer/Writer.h"
```
yingsu00, line 32: @kewang1024 Can you please explain why making this change? The TaskManagerTest is file-format orthogonal and should NOT depend on DWRF. We've been trying to remove DWRF dependencies from HiveConnectorTest as well, e.g. https://github.com/facebookincubator/velox/pull/10150

yingsu00 (changes requested):

majetideepak (dismissed):
Please use `[native]` prefix for the commit.

presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp
xiaoxmeng: @yingsu00 good point. We might need this change as we disable creation of dwio common writer options as it is an abstract class now. If we want to cover parquet file format here, then we can make this a parameterized test with different file formats. But I guess it is not required for TaskManagerTest.

xiaoxmeng: I think this will break Meta internal test as Velox side change on Writer options is landed.

yingsu00 (commented):
I'm removing the "request change" to unblock Meta internal build break for now. We will revisit this issue soon. Since "dwio::common" used to work, I think there may be a better solution than making presto_cpp/main tests depend on DWRF.

yingsu00 (approved):
Approving it for now to unblock Meta workloads

presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp
```diff
@@ -299,7 +300,8 @@ class TaskManagerTest : public testing::Test {
   void writeToFile(
       const std::string& filePath,
       const std::vector<RowVectorPtr>& vectors) {
-    auto options = std::make_shared<dwio::common::WriterOptions>();
+    std::shared_ptr<dwio::common::WriterOptions> options =
```
yingsu00, line 303: @kewang1024 @xiaoxmeng Does the following not work?

const std::shared_ptr<dwio::common::WriterFactory> writerFactory_;

In constructor, initialize the writerFactory_ as:
writerFactory_(dwio::common::getWriterFactory(dwio::common::FileFormat::DWRF))

Then in writeToFile:

auto options = writerFactory_->createWriterOptions();

This way we don't need to include "velox/dwio/dwrf/writer/Writer.h" but just the dwio::common ones.

xiaoxmeng: Yeah, I think it works as in the production code. Thanks!

kewang1024: Thanks @yingsu00 for your good suggestion!

presto-native-execution/presto_cpp/main/tests/TaskManagerTest.cpp
majetideepak: Do we need this header still?

xiaoxmeng (approved):
@kewang1024 thanks!

PR #23482, opened 2024-08-21
[native] Fix Velox connector configs update

zacw7:
The variable needs to be a reference instead of a copy. This is necessary to ensure that changes made to the config are reflected in the query context.

```
== NO RELEASE NOTE ==
```


kewang1024 (approved):
Thanks @zacw7

xiaoxmeng (approved):
@zacw7 thanks for the fix!

PR #23495, opened 2024-08-22
Distinguish scheduler runtime stats for scan stage

NikhilCollooru:
## Description
Distinguish scheduler runtime stats for scan stage scheduling

## Motivation and Context
We want to identify the slowness if any in the scan stage scheduling (split scheduling). So we want to have a different name, so that its easy to identify.

## Impact
None

## Test Plan
run hive query runner and confirm
Screenshot 2024-08-22 at 11 25 17 AM

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-common/src/main/java/com/facebook/presto/common/RuntimeMetricName.java
+++ b/presto-common/src/main/java/com/facebook/presto/common/RuntimeMetricName.java
@@ -41,9 +41,15 @@ public class RuntimeMetricName
     public static final String GET_PARTITIONS_BY_NAMES_TIME_NANOS = "getPartitionsByNamesTimeNanos";
     public static final String GET_TABLE_TIME_NANOS = "getTableTimeNanos";
     public static final String GET_SPLITS_TIME_NANOS = "getSplitsTimeNanos";
+    // CPU time taken to schedule a given stage
     public static final String SCHEDULER_CPU_TIME_NANOS = "schedulerCpuTimeNanos";
+    // Wall time taken to schedule a given stage
     public static final String SCHEDULER_WALL_TIME_NANOS = "schedulerWallTimeNanos";
+    // Blocked time of the scheduler during scheduling a given stage
     public static final String SCHEDULER_BLOCKED_TIME_NANOS = "schedulerBlockedTimeNanos";
+    public static final String SCAN_STAGE_SCHEDULER_CPU_TIME_NANOS = "scanStageSchedulerCpuTimeNanos";
+    public static final String SCAN_STAGE_SCHEDULER_WALL_TIME_NANOS = "scanStageSchedulerWallTimeNanos";
+    public static final String SCAN_STAGE_SCHEDULER_BLOCKED_TIME_NANOS = "scanStageSchedulerBlockedTimeNanos";
     public static final String LOGICAL_PLANNER_TIME_NANOS = "logicalPlannerTimeNanos";
     public static final String OPTIMIZER_TIME_NANOS = "optimizerTimeNanos";
     public static final String GET_CANONICAL_INFO_TIME_NANOS = "getCanonicalInfoTimeNanos";
--- a/presto-main/src/main/java/com/facebook/presto/execution/StageExecutionStateMachine.java
+++ b/presto-main/src/main/java/com/facebook/presto/execution/StageExecutionStateMachine.java
@@ -39,6 +39,9 @@ import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Supplier;
 
 import static com.facebook.presto.common.RuntimeMetricName.GET_SPLITS_TIME_NANOS;
+import static com.facebook.presto.common.RuntimeMetricName.SCAN_STAGE_SCHEDULER_BLOCKED_TIME_NANOS;
+import static com.facebook.presto.common.RuntimeMetricName.SCAN_STAGE_SCHEDULER_CPU_TIME_NANOS;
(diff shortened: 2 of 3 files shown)


elharo (dismissed):

swapsmagic (approved):
LGTM

presto-main/src/main/java/com/facebook/presto/execution/StageExecutionStateMachine.java
```diff
@@ -369,16 +372,16 @@ public void recordGetSplitTime(long startNanos)
         runtimeStats.addMetricValue(GET_SPLITS_TIME_NANOS, NANO, elapsedNanos);
     }
 
-    public void recordSchedulerRunningTime(long cpuTimeNanos, long wallTimeNanos)
+    public void recordSchedulerRunningTime(long cpuTimeNanos, long wallTimeNanos, boolean isScanStage)
     {
-        runtimeStats.addMetricValue(SCHEDULER_CPU_TIME_NANOS, NANO, max(cpuTimeNanos, 0));
-        runtimeStats.addMetricValue(SCHEDULER_WALL_TIME_NANOS, NANO, max(wallTimeNanos, 0));
+        runtimeStats.addMetricValue(isScanStage ? SCAN_STAGE_SCHEDULER_CPU_TIME_NANOS : SCHEDULER_CPU_TIME_NANOS, NANO, max(cpuTimeNanos, 0));
```
rschlussel, line 377: SCHEDULER_CPU_TIME_NANOS name is confusing now for what it represents. I would expect it to include the scheduling stage time as well.  If anyone is relying on this value they also may not realize that it now doesn't include the full scan stage. should we change the name or should we include the scan stage time here as well?

presto-common/src/main/java/com/facebook/presto/common/RuntimeMetricName.java
```diff
@@ -44,6 +44,9 @@ private RuntimeMetricName()
     public static final String SCHEDULER_CPU_TIME_NANOS = "schedulerCpuTimeNanos";
     public static final String SCHEDULER_WALL_TIME_NANOS = "schedulerWallTimeNanos";
     public static final String SCHEDULER_BLOCKED_TIME_NANOS = "schedulerBlockedTimeNanos";
+    public static final String SCAN_STAGE_SCHEDULER_CPU_TIME_NANOS = "scanStageSchedulerCpuTimeNanos";
```
rschlussel, line 47: can you add some comments to the runtime metric about what all these values mean

presto-main/src/main/java/com/facebook/presto/execution/StageExecutionStateMachine.java
```diff
@@ -369,16 +372,16 @@ public void recordGetSplitTime(long startNanos)
         runtimeStats.addMetricValue(GET_SPLITS_TIME_NANOS, NANO, elapsedNanos);
     }
 
-    public void recordSchedulerRunningTime(long cpuTimeNanos, long wallTimeNanos)
+    public void recordSchedulerRunningTime(long cpuTimeNanos, long wallTimeNanos, boolean isScanStage)
```
rschlussel, line 375: nit: instead of passing a flag should we add a new method for recording the new metric and call that directly in SqlStageExecution? I'm not sure if we're saving much by passing this off until later.

rschlussel (commented):
is it possible to have a test for this?

presto-main/src/main/java/com/facebook/presto/execution/StageExecutionStateMachine.java
NikhilCollooru: This metric is introduced recently. So i dont think anyone is using,relying on it yet. 

So this schedulerCpuTime is a metric for every stage. So i want just distinguish the metric for scan stages, so that its easy to track. I did not completely understand your question.

presto-main/src/main/java/com/facebook/presto/execution/StageExecutionStateMachine.java
NikhilCollooru: We dont have access to the runtime stats here, hence we are doing it in the StageExecutionStateMachine

rschlussel: I meant have two methods in StageExecutionStateMachine - recordSchedulerRunningTime() and recordLeafStageSchedulerRunningTime(). And then SqlStageExecution would call the right one depending on if it's a leaf stage or not.  That way StageExecutionStateMachine doesn't need to know anything about the different stages and doesn't have any branching. It just updates the metric that it's asked to update.

presto-main/src/main/java/com/facebook/presto/execution/StageExecutionStateMachine.java
rschlussel: previously scheduler_cpu_time_nanos was the scheduler cpu time for all stages.  With this change, it is the scheduler cpu time for all stages except the scan stage. I think it would be better to either give the metric a name that makes it clear that it's the scheduler time for non-scan stages only, or have it also include the time for scan stages.  I don't have a preference about which way to go here.

NikhilCollooru: I see your point. I made changes so that for leaf scan stage we have both  `schedulerCpuTimeNanos`, `scanStageSchedulerCpuTimeNanos`

rschlussel (approved):

Issue #23501, opened 2024-08-23
Oracle not using BaseJdbcConfigs for user and password


lukmanulhakkeem:
Oracle is not using user and password configs from BaseJdbcConfigs. Currently, we are connecting to Oracle by appending the username and password in the connection URL. 

## Expected Behavior

User and password config from the BaseJdbc config can be used like below. But it is not working now.

```
connector.name=oracle
connection-url=jdbc:oracle:thin:@(DESCRIPTION=(ADDRESS=(PROTOCOL=<tcp>)(HOST=<host>)(PORT=<port>))(CONNECT_DATA=(SID=<sid>)))
connection-user=<user>
connection-password=<password>
```

## Current Behavior

We need to add the user and password in the connection URL even though we have the configs in the BaseJdbc.

`jdbc:oracle:thin:<username>/<password>@//<host>:<port>/<SID/SERVICE_NAME>`

## Possible Solution

It is possible to do a minimal code change to use these configs.

PR #23502, opened 2024-08-23
Oracle user pwd

lukmanulhakkeem:
## Description

Currently we are passing username and password in the connection URL in oracle. But it is better to use the existing basejdbc configs for username and password. This PR is to make changes to use username and password configs for oracle.

## Motivation and Context

https://github.com/prestodb/presto/issues/23501

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* ... :pr:`12345`
* ... :pr:`12345`

Hive Connector Changes
* ... :pr:`12345`
* ... :pr:`12345`
```

```
== NO RELEASE NOTE ==
```


tdcmeehan:
@lukmanulhakkeem should we add some validation that if username and password is present in both URL and config, they are validated to be identical?

lukmanulhakkeem:
Hi @tdcmeehan - I will test this scenario and try to make the changes.

lukmanulhakkeem:
@tdcmeehan - The `connection-user` and `connection-password` configs are overriding the user and password in the connection URL. So if username and password are present in both configs and connection URL, the driver ignores the one in the connection URL and proceeds with the configs. Is any validation needed in that case?

lukmanulhakkeem:
The branch rebase added more commits. I am trying to sync with master

tdcmeehan:
> @tdcmeehan - The `connection-user` and `connection-password` configs are overriding the user and password in the connection URL. So if username and password are present in both configs and connection URL, the driver ignores the one in the connection URL and proceeds with the configs. Is any validation needed in that case?

I think that if a user has specified both, then it would be confusing, and the connector should fail loudly so the user can fix this ambiguity.

lukmanulhakkeem:
@tdcmeehan, I tried the class `java.net.URI` to parse the jdbc url. But Oracle connection url is not compatible with this class. Below is a sample Oracle connection URL

`jdbc:oracle:thin:MYUSER/MYPASSWORD@//example.com:1521/DATABASE`

But the above class expecting a `//` after `thin:`. Also it is expecting a `:` instead of `/` as a separator bw username and password.

steveburnett:
I removed myself as a reviewer and did **not** select prestodb/team-velox to remove. When I attempted to re-add prestodb/team-velox, GitHub blocked that with a message about 15 reviewers being a limit.

tdcmeehan:
@lukmanulhakkeem I see the problem.  Please go ahead with the regular expression approach, but try to make the regular expression as simple as feasible, and please add extensive unit tests.

lukmanulhakkeem:
@tdcmeehan Can i let the presto server fail in this case, or should it run even after the oracle catalog fails?

jaystarshot:
@lukmanulhakkeem  Yes you can add a test for the validator when the validator fails, you can fail the server. Though this will be a breaking change for some people so better to call it out explicitly in the release notes. 
Do other connectors have similar validation when credentials are supplied in both jdbc and config?

lukmanulhakkeem:
@jaystarshot I tested with SQL Server and the behaviour is [same](https://github.com/prestodb/presto/pull/23502#issuecomment-2333047893) there. If we specify credentials in both configs and URL, the one in the configs has the precedence. So is this validation required in oracle?

lukmanulhakkeem:
> @jaystarshot I tested with SQL Server and the behaviour is [same](https://github.com/prestodb/presto/pull/23502#issuecomment-2333047893) there. If we specify credentials in both configs and URL, the one in the configs has the precedence. So is this validation required in oracle?

@jaystarshot could you check this?

jaystarshot:
Sg, can you please add tests to fail if both are present?

lukmanulhakkeem:
But i am asking is this validation really needed? Because the other connectors doesn't have this right?

jaystarshot:
As time mentioned, i think its better to add.

presto-oracle/src/main/java/com/facebook/presto/plugin/oracle/OracleClientModule.java
```diff
@@ -48,7 +53,19 @@ public void configure(Binder binder)
     public static ConnectionFactory connectionFactory(BaseJdbcConfig config, OracleConfig oracleConfig)
             throws SQLException
     {
+        requireNonNull(config, "BaseJdbc config is null");
+
+        boolean iSCredentialsInUrl = JDBCUrlValidator.checkCredentialsInJdbcUrl(config.getConnectionUrl());
+        //Throws an exception when username and password present in both connection url and configs
+        if (iSCredentialsInUrl && (config.getConnectionUser() != null || config.getConnectionPassword() != null)) {
+            System.out.println("both configs and url has the credentials");
```
czentgr, line 61: Given this block throws an exception, do we need the println here?

presto-oracle/src/main/java/com/facebook/presto/plugin/oracle/OracleClientModule.java
```diff
@@ -48,7 +53,19 @@ public void configure(Binder binder)
     public static ConnectionFactory connectionFactory(BaseJdbcConfig config, OracleConfig oracleConfig)
             throws SQLException
     {
+        requireNonNull(config, "BaseJdbc config is null");
+
+        boolean iSCredentialsInUrl = JDBCUrlValidator.checkCredentialsInJdbcUrl(config.getConnectionUrl());
+        //Throws an exception when username and password present in both connection url and configs
+        if (iSCredentialsInUrl && (config.getConnectionUser() != null || config.getConnectionPassword() != null)) {
```
czentgr, line 60: Is || the correct check here? Later the options are only set if both are not null (&&) on line 65.
So this check should also check that both options are present in the config for user and password? Otherwise, this will be true if only connectionUser is not null, irrelevant of the password value (if it exists).

presto-oracle/src/main/java/com/facebook/presto/plugin/oracle/OracleClientModule.java
```diff
@@ -58,4 +75,23 @@ public static ConnectionFactory connectionFactory(BaseJdbcConfig config, OracleC
...
+        //To check the presence of username and password in the connection url
+        public static boolean checkCredentialsInJdbcUrl(String jdbcUrl)
+        {
+            String regex = "jdbc:oracle:thin:([^/]+?)/([^@]+?)@";
+
+            Pattern pattern = Pattern.compile(regex);
+            Matcher matcher = pattern.matcher(jdbcUrl);
+
+            if (matcher.find()) {
+                System.out.println("user and password present in the connection url");
```
czentgr, line 89: Given that later this will throw an exception do we need this printf at this point? or does this need to be behind a debug log if it might be useful for debugging?

czentgr (commented):
And as @jaystarshot mentioned. Can we have some unit tests for the regex?

steveburnett:
@lukmanulhakkeem, when you have time, PTAL at @czentgr's comments.

presto-oracle/src/main/java/com/facebook/presto/plugin/oracle/OracleClientModule.java
```diff
@@ -48,7 +53,19 @@ public void configure(Binder binder)
     public static ConnectionFactory connectionFactory(BaseJdbcConfig config, OracleConfig oracleConfig)
             throws SQLException
     {
+        requireNonNull(config, "BaseJdbc config is null");
+
+        boolean iSCredentialsInUrl = JDBCUrlValidator.checkCredentialsInJdbcUrl(config.getConnectionUrl());
```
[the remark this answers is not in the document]
lukmanulhakkeem: Done

presto-oracle/src/main/java/com/facebook/presto/plugin/oracle/OracleClientModule.java
lukmanulhakkeem: Its not needed. I removed it

presto-oracle/src/main/java/com/facebook/presto/plugin/oracle/OracleClientModule.java
lukmanulhakkeem: I removed the validation in the line 65 you mentioned. I referred MySQL connector and there was no such check. Also we are getting an intuitive logon error when either one of the user or password is missing in the configuration.

presto-oracle/src/main/java/com/facebook/presto/plugin/oracle/OracleClientModule.java
lukmanulhakkeem: Added it under the debug log

lukmanulhakkeem:
@czentgr I tried to address your comments and added a test for the validation. Please review once you have the time

presto-oracle/src/main/java/com/facebook/presto/plugin/oracle/OracleClientModule.java
```diff
@@ -58,4 +73,23 @@ public static ConnectionFactory connectionFactory(BaseJdbcConfig config, OracleC
                 Optional.empty(),
                 connectionProperties);
     }
+    public static class JDBCUrlValidator
+    {
+        //To check the presence of username and password in the connection url
```
czentgr, line 78: Nit: leave a space after // but do we need this comment? maybe a slightly better function name could do?
Such as `findCredentialsInJdbcUrl` and then it is more obvious in connection with the boolean return value?

presto-oracle/src/main/java/com/facebook/presto/plugin/oracle/OracleClientModule.java
```diff
@@ -48,7 +56,14 @@ public void configure(Binder binder)
     public static ConnectionFactory connectionFactory(BaseJdbcConfig config, OracleConfig oracleConfig)
             throws SQLException
     {
-        Properties connectionProperties = new Properties();
+        requireNonNull(config, "BaseJdbc config is null");
+
+        boolean isCredentialsInUrl = JDBCUrlValidator.checkCredentialsInJdbcUrl(config.getConnectionUrl());
+        // Throws an exception when username and password present in both connection url and configs
```
czentgr, line 62: Yes, that's what the code does.
But the comment should explain the reasoning such as:
If username and password are present in both config and url this is an error because no precedence rules are established. By design only one method is valid.

presto-oracle/src/test/java/com/facebook/presto/plugin/oracle/OracleClientModuleTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+import java.sql.SQLException;
+
+import static com.facebook.presto.testing.assertions.Assert.assertEquals;
+import static com.facebook.presto.testing.assertions.Assert.fail;
+
+public class OracleClientModuleTest
+{
+    @Test
+    public void testConnectionFactory_credentialsInUrlAndConfig() throws SQLException
```
czentgr, line 28: There is a limited list of combinations. You could probably write a single test that tests all combinations with the expected result.
For example, what happens if the URL only has the user and the password is set through the config and such combinations.

presto-oracle/src/main/java/com/facebook/presto/plugin/oracle/OracleClientModule.java
lukmanulhakkeem: @czentgr I will remove the comment and rename the function

presto-oracle/src/main/java/com/facebook/presto/plugin/oracle/OracleClientModule.java
lukmanulhakkeem: Is it okay if i am not leaving a comment there?

steveburnett:
Does this need a release note, documentation, or both?

lukmanulhakkeem:
@steveburnett Currently we mentioned about `connection-user` and `password` as like below. 

image

May be we can mention about the validation like below:

"The username and password should be specified either using the `connection-user` and `connection-password` properties or directly in the connection URL. Providing the credentials in both places will result in an error"

cc: @czentgr

lukmanulhakkeem:
@czentgr I tried to address your comments. Please take a pass again and let me know if any changes needed

steveburnett:
> Providing the credentials in both places will result in an error

Yes, that seems useful to mention for the reader.

presto-oracle/src/main/java/com/facebook/presto/plugin/oracle/OracleClientModule.java
```diff
@@ -58,4 +72,22 @@ public static ConnectionFactory connectionFactory(BaseJdbcConfig config, OracleC
                 Optional.empty(),
                 connectionProperties);
     }
+    public static class JDBCUrlValidator
```
czentgr, line 76: Please add a new line to the prior function.

presto-oracle/src/main/java/com/facebook/presto/plugin/oracle/OracleClientModule.java
```diff
@@ -48,7 +56,13 @@ public void configure(Binder binder)
     public static ConnectionFactory connectionFactory(BaseJdbcConfig config, OracleConfig oracleConfig)
             throws SQLException
     {
-        Properties connectionProperties = new Properties();
+        requireNonNull(config, "BaseJdbc config is null");
+
+        boolean isCredentialsInUrl = JDBCUrlValidator.findCredentialsInJdbcUrl(config.getConnectionUrl());
+        if (isCredentialsInUrl && (config.getConnectionUser() != null || config.getConnectionPassword() != null)) {
+            throw new PrestoException(NOT_SUPPORTED, "User credentials in both configs and url");
```
czentgr: Is this explanation sufficient. "both configs"?
Remember that someone will be reading this and needs to understand.
So perhaps better to say "The user credentials were provided using the user and password configuration as well as the URL."

czentgr (commented):
To finish this off:
1. Please rebase.
2. Please make a change to the documentation as discussed in the comments.
3. Please add release notes to the PR description that can be reviewed as well.
4. Tim had requested extensive regex tests for the Oracle URL parsing. The current tests cover the base cases. What about URLs with only the user? We don't allow the password to come from the configuration and the user from the URL, right? Perhaps such combos need to be tested as well.

presto-oracle/src/main/java/com/facebook/presto/plugin/oracle/OracleClientModule.java
lukmanulhakkeem: Thanks for noting that, I will change it

lukmanulhakkeem:
@czentgr @steveburnett 

1. Branch rebased with latest master
2. Documentation changes are added
3. Release note added
4. I tried my best to cover all the test cases. Please let me know if any scenario is missing

presto-docs/src/main/sphinx/connector/oracle.rst
```diff
@@ -41,6 +41,12 @@ for more information.
 The ``connection-user`` and ``connection-password`` are typically required and
 determine the user credentials for the connection, often a service user.
 
+.. note::
+
+   The user and password can be set either in the connection URL or in the 'connection-user' and 'connection-password' properties. Providing the credentials in both places will result in an error.
```
steveburnett:
```suggestion
   The user and password can be set either in the connection URL or in the `connection-user` and `connection-password` properties. Providing the credentials in both places will result in an error.
```

steveburnett (changes requested):
Just a formatting nit. Thanks for the doc!

presto-docs/src/main/sphinx/connector/oracle.rst
```diff
@@ -43,7 +43,7 @@ determine the user credentials for the connection, often a service user.
 
 .. note::
 
-   The user and password can be set either in the connection URL or in the 'connection-user' and 'connection-password' properties. Providing the credentials in both places will result in an error.
+   The user and password can be set either in the connection URL or in the `connection-user` and `connection-password` properties. Providing the credentials in both places will result in an error.
```
steveburnett:
```suggestion
   The user and password can be set either in the connection URL or in the ``connection-user`` and ``connection-password`` properties. Providing the credentials in both places will result in an error.
```
And I gave the wrong formatting. Apologies. Tested and verified this in a local doc build.

steveburnett (changes requested):
I gave the wrong formatting advice, sorry! This is tested and verified in a local doc build.

lukmanulhakkeem:
I will squash the commits once all the review comments are resolved

steveburnett (approved):
LGTM! (docs)

Pull updated branch, new local doc build, looks good. Thanks for the doc!

lukmanulhakkeem:
@czentgr please review this once you have some time. thanks

czentgr (commented):
Please rebase again. Then I'll approve.

lukmanulhakkeem:
@czentgr My apologies for the delay. I have rebased with master and squashed the commits

presto-oracle/src/test/java/com/facebook/presto/plugin/oracle/OracleClientModuleTest.java
```diff
@@ -0,0 +1,109 @@
...
+
+        try {
+            OracleClientModule.connectionFactory(config, oracleConfig);
+            fail("Expected SQLException to be thrown");
+        }
+        catch (PrestoException e) {
+            assertEquals("Credentials are specified in both connection-url and user/password properties. Use only one", e.getMessage());
+        }
+
+        // Credentials in URl but not in the config
```
czentgr: Nit: typo: url instead of URl (lower case is used everywhere else.

lukmanulhakkeem: I have reframed all the comments

lukmanulhakkeem:
@czentgr, Please review this

lukmanulhakkeem:
As per the current regex, '@' is not allowed. I couldn't find any documentation that restricts '@' in the username or password. To check this, I need to create an Oracle database that has a password containing '@'.

Issue #23514, opened 2024-08-24
Function to split an IP prefix into subnets


matt-calder:
This is a proposal for a function `ip_prefix_subnets` that, given an input `ip_prefix` and a subnet `prefix_length`, splits the IP prefix into subnets of the input prefix length. The function will return an array of new prefixes of the new input prefix length.

This function is conceptually a compliment to the `ip_prefix_collapse` function which combines prefixes together into larger IP blocks. 

## Presto Component, Service, or Connector
A new user-facing function.

## Context

While there are a variety of use cases for this function, one of the most useful in an analytical context is helping to perform efficient joins across IP prefix datasets where the primary join key, the IP prefix, data is not all a fixed length. The canonical example of this is joining against IP metadata such as geolocation, ASN, or traffic volume where IP prefixes need to be normalized to a standard length for efficient joins.

elharo:
What are the legal values for ip_prefix_length? What's the behavior if a value outside the legasl values is supplied?

PR #23515, opened 2024-08-24
Function to split an IP prefix into subnets

matt-calder:
## Description
This change introduces a function `ip_prefix_subnets` that, given an input `ip_prefix` and a subnet `prefix_length`, splits the IP prefix into subnets of the input prefix length. The functions return an array of new prefixes of the new input prefix length.

This is a complimentary sister function to the `ip_prefix_collapse` function.

## Motivation and Context

This change addresses [issue 23514](https://github.com/prestodb/presto/issues/23514).

## Impact

This change introduces a new user-facing function `ip_prefix_subnets`.

## Test Plan

Unit tests

`mvn clean install -Dtest=TestIpPrefixFunctions -Dmaven.javadoc.skip=true -DskipUI -T1C -fn -pl presto-main`

## Contributor checklist

- [ ] CI passed.

```
== RELEASE NOTES ==

General Changes
* Add function :func:`ip_prefix_subnets` that splits the input prefix into subnets the size of the new prefix length. ... :pr:`23515`
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/operator/scalar/IpPrefixFunctions.java
+++ b/presto-main/src/main/java/com/facebook/presto/operator/scalar/IpPrefixFunctions.java
@@ -51,6 +51,8 @@ public final class IpPrefixFunctions
 {
     private static final BigInteger TWO = BigInteger.valueOf(2);
 
+    private static final Block emptyBlock = IPPREFIX.createBlockBuilder(null, 0).build();
+
     private IpPrefixFunctions() {}
 
     @Description("IP prefix for a given IP address and subnet size")
@@ -192,6 +194,66 @@ public final class IpPrefixFunctions
             for (Slice ipPrefix : ipPrefixSlices) {
                 IPPREFIX.writeSlice(blockBuilder, ipPrefix);
             }
+        }
+
+        return blockBuilder.build();
+    }
+
+    @Description("Split the input prefix into subnets the size of the new prefix length.")
+    @ScalarFunction("ip_prefix_subnets")
+    @SqlType("array(IPPREFIX)")
+    public static Block ipPrefixSubnets(@SqlType(StandardTypes.IPPREFIX) Slice prefix, @SqlType(StandardTypes.BIGINT) long newPrefixLength)
+    {
+        boolean inputIsIpV4 = isIpv4(prefix);
+
+        if (newPrefixLength < 0 || (inputIsIpV4 && newPrefixLength > 32) || (!inputIsIpV4 && newPrefixLength > 128)) {
+            throw new PrestoException(INVALID_FUNCTION_ARGUMENT, "Invalid prefix length for IPv" + (inputIsIpV4 ? "4" : "6") + ": " + newPrefixLength);
+        }
+
+        int inputPrefixLength = getPrefixLength(prefix);
+        // An IP prefix is a 'network', or group of contiguous IP addresses. The common format for describing IP prefixes is
+        // uses 2 parts separated by a '/': (1) the IP address part and the (2) prefix length part (also called subnet size or CIDR).
+        // For example, in 9.255.255.0/24, 9.255.255.0 is the IP address part and 24 is the prefix length.
+        // The prefix length describes how many IP addresses the prefix contains in terms of the leading number of bits required. A higher number of bits
+        // means smaller number of IP addresses. Subnets inherently mean smaller groups of IP addresses.
+        // We can only disaggregate a prefix if the prefix length is the same length or longer (more-specific) than the length of the input prefix.
+        // E.g., if the input prefix is 9.255.255.0/24, the prefix length can be /24, /25, /26, etc... but not 23 or larger value than 24.
+
+        int newPrefixCount = 0;  // if inputPrefixLength > newPrefixLength, there are no new prefixes and we will return an empty array.
+        if (inputPrefixLength <= newPrefixLength) {
+            // Next, count how many new prefixes we will generate. In general, every difference in prefix length doubles the number new prefixes.
+            // For example if we start with 9.255.255.0/24, and want to split into /25s, we would have 2 new prefixes. If we wanted to split into /26s,
+            // we would have 4 new prefixes, and /27 would have 8 prefixes etc....
+            newPrefixCount = 1 << (newPrefixLength - inputPrefixLength);  // 2^N
+        }
+
+        if (newPrefixCount == 0) {
+            return emptyBlock;
+        }
+
+        BlockBuilder blockBuilder = IPPREFIX.createBlockBuilder(null, newPrefixCount);
+
+        if (newPrefixCount == 1) {
+            IPPREFIX.writeSlice(blockBuilder, prefix); // just return the original prefix in an array
+            return blockBuilder.build(); // returns empty or single entry
+        }
+
+        int ipVersionMaxBits = inputIsIpV4 ? 32 : 128;
+        BigInteger newPrefixIpCount = TWO.pow(ipVersionMaxBits - (int) newPrefixLength);
+
+        Slice startingIpAddressAsSlice = ipSubnetMin(prefix);
+        BigInteger currentIpAddress = toBigInteger(startingIpAddressAsSlice);
+
+        try {
+            for (int i = 0; i < newPrefixCount; i++) {
+                InetAddress asInetAddress = bigIntegerToIpAddress(currentIpAddress);
+                Slice ipPrefixAsSlice = castFromVarcharToIpPrefix(utf8Slice(InetAddresses.toAddrString(asInetAddress) + "/" + newPrefixLength));
+                IPPREFIX.writeSlice(blockBuilder, ipPrefixAsSlice);
+                currentIpAddress = currentIpAddress.add(newPrefixIpCount);   // increment to start of next new prefix
+            }
+        }
+        catch (UnknownHostException ex) {
+            throw new PrestoException(GENERIC_INTERNAL_ERROR, "Unable to convert " + currentIpAddress + " to IP prefix", ex);
         }
 
         return blockBuilder.build();
@@ -356,6 +418,11 @@ public final class IpPrefixFunctions
         }
 
         return ipPartBytes[10] == (byte) 0xff && ipPartBytes[11] == (byte) 0xff;
+    }
+
+    private static int getPrefixLength(Slice ipPrefix)
+    {
+        return ipPrefix.getByte(IPPREFIX.getFixedSize() - 1) & 0xFF;
     }
 
     private static InetAddress toInetAddress(Slice ipAddress)
--- a/presto-docs/src/main/sphinx/functions/ip.rst
+++ b/presto-docs/src/main/sphinx/functions/ip.rst
@@ -60,3 +60,12 @@ IP Functions
         SELECT IP_PREFIX_COLLAPSE(ARRAY[IPPREFIX '2620:10d:c090::/48', IPPREFIX '2620:10d:c091::/48']); -- [{2620:10d:c090::/47}]
         SELECT IP_PREFIX_COLLAPSE(ARRAY[IPPREFIX '192.168.1.0/24', IPPREFIX '192.168.0.0/24', IPPREFIX '192.168.2.0/24', IPPREFIX '192.168.9.0/24']); -- [{192.168.0.0/23}, {192.168.2.0/24}, {192.168.9.0/24}]
 
+.. function:: ip_prefix_subnets(ip_prefix, prefix_length) -> array(ip_prefix)
+
+    Returns the subnets of ``ip_prefix`` of size ``prefix_length``. ``prefix_length`` must be valid ([0, 32] for IPv4
+    and [0, 128] for IPv6) or the query will fail and raise an error. An empty array is returned if ``prefix_length``
+    is shorter (that is, less specific) than ``ip_prefix``. ::
+
+        SELECT IP_PREFIX_SUBNETS(IPPREFIX '192.168.1.0/24', 25); -- [{192.168.1.0/25}, {192.168.1.128/25}]
+        SELECT IP_PREFIX_SUBNETS(IPPREFIX '2a03:2880:c000::/34', 36); -- [{2a03:2880:c000::/36}, {2a03:2880:d000::/36}, {2a03:2880:e000::/36}, {2a03:2880:f000::/36}]
+
--- a/presto-main/src/test/java/com/facebook/presto/operator/scalar/TestIpPrefixFunctions.java
+++ b/presto-main/src/test/java/com/facebook/presto/operator/scalar/TestIpPrefixFunctions.java
@@ -216,4 +216,56 @@ public class TestIpPrefixFunctions
         assertInvalidFunction("IP_PREFIX_COLLAPSE(ARRAY[IPPREFIX '192.168.0.0/22', IPPREFIX '2409:4043:251a:d200::/56'])",
                 "All IPPREFIX elements must be the same IP version.");
     }
+
+    @Test
+    public void testIpPrefixSubnets()
+    {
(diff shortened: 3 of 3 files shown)


presto-main/src/main/java/com/facebook/presto/operator/scalar/IpPrefixFunctions.java
```diff
@@ -197,6 +197,54 @@ public static Block collapseIpPrefixes(@SqlType("array(IPPREFIX)") Block unsorte
...
+    public static Block ipPrefixSubnets(@SqlType(StandardTypes.IPPREFIX) Slice prefix, @SqlType(StandardTypes.BIGINT) long newPrefixLength)
+    {
+        boolean inputIsIpV4 = isIpv4(prefix);
+
+        if (newPrefixLength < 0 || (inputIsIpV4 && newPrefixLength > 32) || (!inputIsIpV4 && newPrefixLength > 128)) {
+            throw new PrestoException(INVALID_FUNCTION_ARGUMENT, "Invalid prefix length for IPv" + (inputIsIpV4 ? "4" : "6") + ": " + newPrefixLength);
+        }
+
+        int inputPrefixLength = getPrefixLength(prefix);
+        // We can only deaggregate a prefix if the prefix length is longer (more-specific) than the prefix length of the input prefix.
```
elharo, line 212: disaggregate is more ocmmon

presto-main/src/main/java/com/facebook/presto/operator/scalar/IpPrefixFunctions.java
```diff
@@ -197,6 +197,54 @@ public static Block collapseIpPrefixes(@SqlType("array(IPPREFIX)") Block unsorte
...
+        // E.g., if the input prefix is /24, the subnetSize must be like /25, /26 etc...
```
elharo, line 213: super nit: etc....

presto-main/src/main/java/com/facebook/presto/operator/scalar/IpPrefixFunctions.java
```diff
@@ -197,6 +197,54 @@ public static Block collapseIpPrefixes(@SqlType("array(IPPREFIX)") Block unsorte
...
+        if (inputPrefixLength > newPrefixLength) {
+            throw new PrestoException(INVALID_FUNCTION_ARGUMENT, "newPrefixLength argument '" + newPrefixLength + "' must be longer (more-specific) than the input prefix's length '" + inputPrefixLength + "'");
```
elharo, line 215: would it be possible to just return the original subnet instead?

steveburnett (dismissed):
LGTM! (docs) 

presto-main/src/main/java/com/facebook/presto/operator/scalar/IpPrefixFunctions.java
matt-calder: This is a good question. After sampling the behavior in a few other libraries it seems there is no clear answer.
* Throws exception
    * [Python3](https://docs.python.org/3/library/ipaddress.html) and [C#](https://github.com/lduchosal/ipnetwork/blob/master/src/System.Net.IPNetwork/IPNetwork2.cs) subnet functions throw exceptions.
* [netaddr](https://netaddr.readthedocs.io/en/latest/tutorial_01.html#summarizing-list-of-addresses-and-subnets) returns an empty list.
* [ipaddress-go library](https://seancfoley.github.io/IPAddress/) returns the supernet.

I'm in favor of the current behavior of generating a query error because it provides the nice property that all output prefixes will be the same prefix length. It forces the user to determine how their use case should handle the less specific cases which I believe is the best policy for data quality.

presto-main/src/main/java/com/facebook/presto/operator/scalar/IpPrefixFunctions.java
```diff
@@ -197,6 +197,54 @@ public static Block collapseIpPrefixes(@SqlType("array(IPPREFIX)") Block unsorte
         return blockBuilder.build();
     }
 
+    @Description("Split the input prefix into subnets the size of the new prefix length.")
+    @ScalarFunction("ip_prefix_subnets")
+    @SqlType("array(IPPREFIX)")
+    public static Block ipPrefixSubnets(@SqlType(StandardTypes.IPPREFIX) Slice prefix, @SqlType(StandardTypes.BIGINT) long newPrefixLength)
```
yingsu00, line 203: static import StandardTypes.IPPREFIX and StandardTypes.BIGINT. 
I know the existing code didnt' use static import either, will you be able to add a second commit to make the change to all functions?

presto-main/src/main/java/com/facebook/presto/operator/scalar/IpPrefixFunctions.java
```diff
@@ -197,6 +197,54 @@ public static Block collapseIpPrefixes(@SqlType("array(IPPREFIX)") Block unsorte
...
+    {
+        boolean inputIsIpV4 = isIpv4(prefix);
+
+        if (newPrefixLength < 0 || (inputIsIpV4 && newPrefixLength > 32) || (!inputIsIpV4 && newPrefixLength > 128)) {
+            throw new PrestoException(INVALID_FUNCTION_ARGUMENT, "Invalid prefix length for IPv" + (inputIsIpV4 ? "4" : "6") + ": " + newPrefixLength);
+        }
+
+        int inputPrefixLength = getPrefixLength(prefix);
+        // We can only disaggregate a prefix if the prefix length is longer (more-specific) than the prefix length of the input prefix.
+        // E.g., if the input prefix is /24, the subnetSize must be like /25, /26 etc....
```
yingsu00, line 213: Use the full prefix string here. e.g. 9.255.255.0/24, or use the word "prefix length" and explain that the prefix is composed of IP/(prefix lengtgh).

yingsu00, line 213: /26 etc....
-->
/26, etc.

presto-main/src/main/java/com/facebook/presto/operator/scalar/IpPrefixFunctions.java
```diff
@@ -197,6 +197,54 @@ public static Block collapseIpPrefixes(@SqlType("array(IPPREFIX)") Block unsorte
...
+        if (inputPrefixLength > newPrefixLength) {
+            throw new PrestoException(INVALID_FUNCTION_ARGUMENT, "newPrefixLength argument '" + newPrefixLength + "' must be the same or longer (more-specific) than the input prefix's length '" + inputPrefixLength + "'");
+        }
+
+        // Count how many new prefixes we will have.  /24 => /26s; 26 - 24 = 2; 2^2 = 4 new prefixes.
```
yingsu00, line 218: Could u please use plain English instead of just some equations? We want to make the code and comment as straightforward as possible.

presto-main/src/main/java/com/facebook/presto/operator/scalar/IpPrefixFunctions.java
matt-calder: Yes. I can send out a new PR that updates the rest of the functions.

matt-calder: Actually, I think the static import of `StandardTypes.IPPREFIX` is in conflict with the static import of  `com.facebook.presto.type.IpPrefixType.IPPREFIX`.

presto-main/src/main/java/com/facebook/presto/operator/scalar/IpPrefixFunctions.java
matt-calder: Took a another pass through these comments based on your feedback.

presto-main/src/main/java/com/facebook/presto/operator/scalar/IpPrefixFunctions.java
matt-calder: @yingsu00 Not a Java expert so not sure the best path forward given the naming conflict.

presto-main/src/main/java/com/facebook/presto/operator/scalar/IpPrefixFunctions.java
```diff
@@ -197,6 +197,65 @@ public static Block collapseIpPrefixes(@SqlType("array(IPPREFIX)") Block unsorte
...
+    public static Block ipPrefixSubnets(@SqlType(StandardTypes.IPPREFIX) Slice prefix, @SqlType(StandardTypes.BIGINT) long newPrefixLength)
+    {
+        boolean inputIsIpV4 = isIpv4(prefix);
+
+        if (newPrefixLength < 0 || (inputIsIpV4 && newPrefixLength > 32) || (!inputIsIpV4 && newPrefixLength > 128)) {
+            throw new PrestoException(INVALID_FUNCTION_ARGUMENT, "Invalid prefix length for IPv" + (inputIsIpV4 ? "4" : "6") + ": " + newPrefixLength);
+        }
+
+        int inputPrefixLength = getPrefixLength(prefix);
+        /*
```
yingsu00, line 212: Use // instead of /*

presto-main/src/main/java/com/facebook/presto/operator/scalar/IpPrefixFunctions.java
yingsu00: @elharo @tdcmeehan What's your thoughts on this?

elharo: I don't have a strong opinion about this. My weak opinion is that compilable SQL functions should not throw errors or otherwise fail at runtime based on the arguments. They should always return a predictable value. Python, C#, and Java are very different things than SQL so I don't find their behavior here a useful precedent.

What happens if input prefix is /24 and subnetsize is also /24?

matt-calder: @elharo If the input prefix is /24 and subnetsize is also /24 then we just return the input prefix.

presto-main/src/main/java/com/facebook/presto/operator/scalar/IpPrefixFunctions.java
```diff
@@ -197,6 +197,61 @@ public static Block collapseIpPrefixes(@SqlType("array(IPPREFIX)") Block unsorte
...
+            throw new PrestoException(INVALID_FUNCTION_ARGUMENT, "Invalid prefix length for IPv" + (inputIsIpV4 ? "4" : "6") + ": " + newPrefixLength);
+        }
+
+        int inputPrefixLength = getPrefixLength(prefix);
+        // An IP prefix is a 'network', or group of contiguous IP addresses. The common format for describing IP prefixes is
+        // uses 2 parts separated by a '/': (1) the IP address part and the (2) prefix length part (also called subnet size or CIDR).
+        // For example, in 9.255.255.0/24, 9.255.255.0 is the IP address part and 24 is the prefix length.
+        // The prefix length describes how many IP addresses the prefix contains in terms of the leading number of bits required. A higher number of bits
+        // means smaller number of IP addresses. Subnets inherently mean smaller groups of IP addresses.
+        // We can only disaggregate a prefix if the prefix length is longer (more-specific) than the prefix length of the input prefix.
```
elharo, line 217: same length or longer
prefix length of the --> length of the

presto-main/src/main/java/com/facebook/presto/operator/scalar/IpPrefixFunctions.java
```diff
@@ -197,6 +197,61 @@ public static Block collapseIpPrefixes(@SqlType("array(IPPREFIX)") Block unsorte
...
+        // E.g., if the input prefix is 9.255.255.0/24, the prefix length can be like /25, /26, etc... but not 23 or larger value than 24.
```
elharo, line 218: can be like /25 --> can be /24, /25

presto-main/src/main/java/com/facebook/presto/operator/scalar/IpPrefixFunctions.java
```diff
@@ -197,6 +197,61 @@ public static Block collapseIpPrefixes(@SqlType("array(IPPREFIX)") Block unsorte
...
+        if (inputPrefixLength > newPrefixLength) {
+            throw new PrestoException(INVALID_FUNCTION_ARGUMENT, "newPrefixLength argument '" + newPrefixLength + "' must be the same or longer (more-specific) than the input prefix's length '" + inputPrefixLength + "'");
```
elharo, line 220: delete "argument"

presto-main/src/main/java/com/facebook/presto/operator/scalar/IpPrefixFunctions.java
matt-calder: The discussion here has stalled a bit. Throwing an error in this case isn't my hill to die on -- I think returning NULL or an empty array could also work. @kaikalur or @feilong-liu, thoughts on this?

steveburnett (dismissed):
(Re-review after close and reopening.)

LGTM! (docs)

presto-main/src/main/java/com/facebook/presto/operator/scalar/IpPrefixFunctions.java
matt-calder: Thanks to @sharadjaiswal for this suggestion:

> While, I gave this example (for `approx_percentile`), in this case, if someone picks an input percentile value of say 1.35, its really unclear what the `approx_percentile` function should do? Should they treat it as 1.00??  - but that’s inferring the intent of the callee. However, returning ip_prefix_subnets(’[12.13.14.15/24](http://12.13.14.15/24)’, 16) as [] seems more correct - we are not inferring intent, we are just saying “there exist no such subnets”, and it saves us from debugging error messages...

steveburnett (approved):
(rerereview by request)

LGTM! (docs)

presto-main/src/main/java/com/facebook/presto/operator/scalar/IpPrefixFunctions.java
```diff
@@ -197,6 +197,63 @@ public static Block collapseIpPrefixes(@SqlType("array(IPPREFIX)") Block unsorte
...
+        if (inputPrefixLength <= newPrefixLength) {
+            // Next, count how many new prefixes we will generate. In general, every difference in prefix length doubles the number new prefixes.
+            // For example if we start with 9.255.255.0/24, and want to split into /25s, we would have 2 new prefixes. If we wanted to split into /26s,
+            // we would have 4 new prefixes, and /27 would have 8 prefixes etc....
+            newPrefixCount = 1 << (newPrefixLength - inputPrefixLength);  // 2^N
+        }
+
+        BlockBuilder blockBuilder = IPPREFIX.createBlockBuilder(null, newPrefixCount);
+        if (newPrefixCount <= 1) {
+            if (newPrefixCount == 1) {
```
kaikalur, line 230: Stylistically, I would make it:

if (newPrefixCount == 1) ... 
else if (newPrefixCount < 1) ...

presto-main/src/main/java/com/facebook/presto/operator/scalar/IpPrefixFunctions.java
```diff
@@ -197,6 +197,63 @@ public static Block collapseIpPrefixes(@SqlType("array(IPPREFIX)") Block unsorte
...
+                IPPREFIX.writeSlice(blockBuilder, prefix); // just return the original prefix in an array
+            }
+            return blockBuilder.build(); // returns empty or single entry
```
kaikalur, line 233: Maybe have a static constant so we don't keep creating empty blocks

steveburnett (dismissed):
LGTM! (docs)

matt-calder:
I was unable to squash and merge this correctly so this branch is FUBAR. Abandoning for new PR.

Issue #18617, opened 2022-11-03
[native] Add support for graceful shutdown 


majetideepak:
Presto orchestration tools benefit from this feature
https://prestodb.io/docs/current/release/release-0.128.html#graceful-shutdown

Example usage:
```
curl -v -X PUT -d '"SHUTTING_DOWN"' -H "Content-type: application/json" \
    http://worker:8080/v1/info/state
```

Issue #23299, opened 2024-07-25
[Native] Support Graceful Shutdown in Prestissimo


ethanyzhang:
In Prestissimo, we do not yet support the PUT method in `/v1/info/state` for transitioning the server to the SHUTING_DOWN state for graceful shutdown

https://prestodb.io/docs/current/release/release-0.128.html#graceful-shutdown

I do see the API exists: https://github.com/prestodb/presto/blob/72e771bb5cf1aa62ea2d708e3501ed42b13ab59d/presto-native-execution/presto_cpp/main/PrestoServer.cpp#L882-L890 but not hooked up with the REST API endpoint. `/v1/info/state` now only supports GET: https://github.com/prestodb/presto/blob/72e771bb5cf1aa62ea2d708e3501ed42b13ab59d/presto-native-execution/presto_cpp/main/PrestoServer.cpp#L349-L357

kewang1024:
Yes, unlike Java, the signal is received by here: https://github.com/prestodb/presto/blob/39d419e8a13b4ff6ec372e91e8cb82bbc36be926/presto-native-execution/presto_cpp/main/SignalHandler.cpp#L29

ethanyzhang:
Shakyan Kushwaha from my team is working on this

amitkdutta:
@ethanyzhang This is intentially not implemented in native worker, because native worker can handle Unix Process signals (e.g. SIGTERM) and trigger shutdown automtaically. This is a common pattern for any UNIX process to handle signal instead of explicit shutdown call (e.g. hitting CTRL + C to any application sends SIGTERM to the application, which is then handled by the application if signal handler is written. In java, such signal handling is likely not possible - hence the API was added to ask worker to shutdown.

ethanyzhang:
@amitkdutta I believe the motivation here is for bringing the cluster auto-scaling we support in our Java offering to Prestissimo. We need to implement the same Java semantics like documented in https://prestodb.io/docs/current/release/release-0.128.html#graceful-shutdown when we need to scale back the cluster so we can call the API to notify a worker to stop like we can do in Java.

ethanyzhang:
@aditi-pandit @majetideepak @tdcmeehan  can you help follow up on this?

majetideepak:
@amitkdutta It is not straightforward to send a unix process signal if the control plane/orchestration is on a remote node.
Ahana's control plane uses REST API to manage the scaling. We want to use the same implementation to manage both Java and C++ clusters.
[...]

PR #23517, opened 2024-08-25
[native] Implement Graceful Shutdown in Native worker

anandamideShakyan:
## Description
Added PUT method in /v1/info/state for transitioning the server to the SHUTTING_DOWN state for graceful shutdown

Resolves https://github.com/prestodb/presto/issues/18617

## Motivation and Context
The Prestissimo project currently lacks support for transitioning the server to the SHUTTING_DOWN state via the PUT method in the /v1/info/state endpoint. Implementing this feature allows for a graceful shutdown of the server, ensuring that ongoing processes are completed without abrupt termination.

## Impact
This implementation introduces a new feature that enables the server to handle a graceful shutdown when the SHUTTING_DOWN state is triggered via a PUT request. By supporting this functionality, the server can now properly manage ongoing requests before shutting down, reducing the risk of data loss or inconsistency.

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* ... :pr:`12345`
* ... :pr:`12345`

Hive Connector Changes
* ... :pr:`12345`
* ... :pr:`12345`
```

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/PrestoServer.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoServer.cpp
@@ -26,6 +26,7 @@
 #include "presto_cpp/main/common/ConfigReader.h"
 #include "presto_cpp/main/common/Counters.h"
 #include "presto_cpp/main/common/Utils.h"
+#include "presto_cpp/main/http/HttpConstants.h"
 #include "presto_cpp/main/http/filters/AccessLogFilter.h"
 #include "presto_cpp/main/http/filters/HttpEndpointLatencyFilter.h"
 #include "presto_cpp/main/http/filters/InternalAuthenticationFilter.h"
@@ -372,6 +373,14 @@ void PrestoServer::run() {
           proxygen::ResponseHandler* downstream) {
         json infoStateJson = convertNodeState(server->nodeState());
         http::sendOkResponse(downstream, infoStateJson);
+      });
+  httpServer_->registerPut(
+      "/v1/info/state",
+      [server = this](
+          proxygen::HTTPMessage* /*message*/,
+          const std::vector<std::unique_ptr<folly::IOBuf>>& body,
+          proxygen::ResponseHandler* downstream) {
+        server->handleGracefulShutdown(body, downstream);
       });
   httpServer_->registerGet(
       "/v1/status",
@@ -911,7 +920,6 @@ void PrestoServer::stop() {
   PRESTO_SHUTDOWN_LOG(INFO)
       << "Waiting for " << shutdownOnsetSec
       << " second(s) before proceeding with the shutdown...";
-
   // Give coordinator some time to receive our new node state and stop sending
   // any tasks.
   std::this_thread::sleep_for(std::chrono::seconds(shutdownOnsetSec));
@@ -1433,6 +1441,27 @@ void PrestoServer::reportServerInfo(proxygen::ResponseHandler* downstream) {
 
 void PrestoServer::reportNodeStatus(proxygen::ResponseHandler* downstream) {
   http::sendOkResponse(downstream, json(fetchNodeStatus()));
+}
+
+void PrestoServer::handleGracefulShutdown(
+    const std::vector<std::unique_ptr<folly::IOBuf>>& body,
+    proxygen::ResponseHandler* downstream) {
+  std::string bodyContent =
+      folly::trimWhitespace(body[0]->moveToFbString()).toString();
+  if (body.size() == 1 && bodyContent == http::kShuttingDown) {
+    LOG(INFO) << "Shutdown requested";
+    if (nodeState() == NodeState::kActive) {
+      // Run stop() in a separate thread to avoid blocking the main HTTP handler
+      // and ensure a timely 200 OK response to the client.
+      std::thread([this]() { this->stop(); }).detach();
+    } else {
+      LOG(INFO) << "Node is inactive or shutdown is already requested";
+    }
+    http::sendOkResponse(downstream);
+  } else {
+    LOG(ERROR) << "Bad Request. Received body content: " << bodyContent;
+    http::sendErrorResponse(downstream, "Bad Request", http::kHttpBadRequest);
+  }
 }
 
 void PrestoServer::registerSidecarEndpoints() {
--- a/presto-native-execution/presto_cpp/main/PrestoServer.h
+++ b/presto-native-execution/presto_cpp/main/PrestoServer.h
@@ -211,6 +211,10 @@ class PrestoServer {
 
   void reportNodeStatus(proxygen::ResponseHandler* downstream);
 
+  void handleGracefulShutdown(
+      const std::vector<std::unique_ptr<folly::IOBuf>>& body,
+      proxygen::ResponseHandler* downstream);
+
   protocol::NodeStatus fetchNodeStatus();
 
   void populateMemAndCPUInfo();
--- a/presto-native-execution/presto_cpp/main/http/HttpConstants.h
+++ b/presto-native-execution/presto_cpp/main/http/HttpConstants.h
@@ -17,11 +17,13 @@ namespace facebook::presto::http {
 const uint16_t kHttpOk = 200;
 const uint16_t kHttpAccepted = 202;
 const uint16_t kHttpNoContent = 204;
+const uint16_t kHttpBadRequest = 400;
 const uint16_t kHttpUnauthorized = 401;
 const uint16_t kHttpNotFound = 404;
 const uint16_t kHttpInternalServerError = 500;
 
 const char kMimeTypeApplicationJson[] = "application/json";
 const char kMimeTypeApplicationThrift[] = "application/x-thrift+binary";
+const char kShuttingDown[] = "\"SHUTTING_DOWN\"";
 static const char kPrestoInternalBearer[] = "X-Presto-Internal-Bearer";
 } // namespace facebook::presto::http
--- a/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/TestPrestoNativeGracefulShutdown.java
+++ b/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/TestPrestoNativeGracefulShutdown.java
@@ -1,0 +1,45 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.nativeworker;
+
+import com.facebook.presto.spi.NodeState;
+import com.facebook.presto.testing.QueryRunner;
+import com.facebook.presto.tests.DistributedQueryRunner;
+import org.testng.annotations.Test;
+
+import static com.facebook.presto.spi.NodeState.SHUTTING_DOWN;
+import static org.testng.Assert.assertEquals;
+
+public class TestPrestoNativeGracefulShutdown
+{
+    @Test
+    public void testGracefulShutdown() throws Exception
+    {
+        QueryRunner queryRunner = PrestoNativeQueryRunnerUtils.createNativeQueryRunner(true);
+        DistributedQueryRunner distributedQueryRunner = (DistributedQueryRunner) queryRunner;
+
+        int responseCode = distributedQueryRunner.sendWorkerRequest(0, "INVALID_BODY");
+        assertEquals(responseCode, 400, "Expected a 400 Bad Request response for invalid body");
+
+        responseCode = distributedQueryRunner.sendWorkerRequest(0, "");
+        assertEquals(responseCode, 400, "Expected a 400 Bad Request response for empty body");
+
+        responseCode = distributedQueryRunner.sendWorkerRequest(0, "\"SHUTTING_DOWN\"");
+        assertEquals(responseCode, 200, "Expected a 200 OK response for valid shutdown request");
+        NodeState state = distributedQueryRunner.getWorkerInfoState(0);
+        assertEquals(state.getValue(), SHUTTING_DOWN.getValue());
+
+        queryRunner.close();
+    }
+}
--- a/presto-tests/src/main/java/com/facebook/presto/tests/DistributedQueryRunner.java
+++ b/presto-tests/src/main/java/com/facebook/presto/tests/DistributedQueryRunner.java
@@ -61,8 +61,12 @@ import org.jdbi.v3.core.Handle;
 import org.jdbi.v3.core.Jdbi;
 
 import java.io.IOException;
+import java.io.OutputStream;
 import java.io.UncheckedIOException;
+import java.net.HttpURLConnection;
 import java.net.URI;
+import java.net.URL;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -413,6 +417,39 @@ public class DistributedQueryRunner
         return state;
     }
 
+    public NodeState getWorkerInfoState(int worker)
+    {
+        URI uri = URI.create(getWorker(worker).getBaseUrl().toString() + "/v1/info/state");
+        Request request = prepareGet()
+                .setHeader(PRESTO_USER, DEFAULT_USER)
+                .setUri(uri)
+                .build();
+
+        NodeState state = client.execute(request, createJsonResponseHandler(jsonCodec(NodeState.class)));
+        return state;
+    }
+
+    public int sendWorkerRequest(int worker, String body)
+    {
+        try {
+            URL url = new URL(getWorker(worker).getBaseUrl().toString() + "/v1/info/state");
+            HttpURLConnection connection = (HttpURLConnection) url.openConnection();
+            connection.setRequestMethod("PUT");
+            connection.setDoOutput(true);
+            connection.setRequestProperty("Content-Type", "application/json");
+
+            try (OutputStream os = connection.getOutputStream()) {
+                byte[] input = body.getBytes(StandardCharsets.UTF_8);
+                os.write(input, 0, input.length);
+            }
+            return connection.getResponseCode();
+        }
+        catch (Exception e) {
+            e.printStackTrace();
+            return 500;
+        }
+    }
+
     private static TestingPrestoServer createTestingPrestoServer(
             URI discoveryUri,
             boolean resourceManager,
@@ -599,6 +636,12 @@ public class DistributedQueryRunner
     {
         checkState(coordinator < coordinators.size(), format("Expected coordinator index %d < %d", coordinator, coordinatorCount));
         return coordinators.get(coordinator);
+    }
+
+    private TestingPrestoServer getWorker(int worker)
+    {
+        checkState(worker < servers.size(), format("Expected worker index %d < %d", worker, servers.size()));
+        return servers.get(worker);
     }
 
     public List<TestingPrestoServer> getCoordinators()


aditi-pandit (commented):
Thanks for this code. @anandamideShakyan : Please can you add a unit test (maybe in ServerOperationsTest or a new unit test file) for this functionality.

anandamideShakyan:
@aditi-pandit Sure, I will add the unit test.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -363,6 +363,44 @@ void PrestoServer::run() {
         json infoStateJson = convertNodeState(server->nodeState());
         http::sendOkResponse(downstream, infoStateJson);
       });
+  httpServer_->registerPut(
+    "/v1/info/state",
+    [server = this](
+        proxygen::HTTPMessage* /*message*/,
+        const std::vector<std::unique_ptr<folly::IOBuf>>& body ,
+        proxygen::ResponseHandler* downstream) {
+         std::string requestBody;
```
czentgr, line 372: Please  move this logic into its own function.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -363,6 +363,44 @@ void PrestoServer::run() {
...
+                } else {
+                    // Print message and indicate that shutdown is already requested
+                    LOG(INFO) << "Shutdown already requested";
+                }
+                http::sendOkResponse(downstream);
+            } else {
+                // Throw a 400 Bad Request error if the body has some other value
+                http::sendErrorResponse(
+                    downstream,
+                    "Bad Request: Body must contain 'SHUTTING_DOWN'.",400);
```
czentgr, line 396: You can define a new constant for bad request in HttpConstants.h and use that.

amitkdutta (changes requested):
@anandamideShakyan Prestoserver handles graceful shutdown mechanism with SIGTERM, which is a standard way to shut down any process. Here is how the signal handler is registered (https://github.com/prestodb/presto/blob/master/presto-native-execution/presto_cpp/main/PrestoServer.cpp#L152)

Also in the [stop method](https://github.com/prestodb/presto/blob/master/presto-native-execution/presto_cpp/main/SignalHandler.cpp#L31)], server goes in [graceful shotdown mode](https://github.com/prestodb/presto/blob/master/presto-native-execution/presto_cpp/main/PrestoServer.cpp#L890)

As a result, no explicit endpoint is required to ask the server to shut down. Just sending SIGTERM like any other process works as expected. In Java, perhaps these signal handling is not possible, hence this end point concept came into development. Native worker even have additional signal handling mechnism to catch debugging information in case of SEGFault (https://github.com/prestodb/presto/blob/master/presto-native-execution/presto_cpp/main/PrestoMain.cpp#L25C27-L25C47)

czentgr (commented):
Thank you for adding this.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
anandamideShakyan: Ok. I will create getRequestBody function for this.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
anandamideShakyan: done

anandamideShakyan:
> @anandamideShakyan Prestoserver handles graceful shutdown mechanism with SIGTERM, which is a standard way to shut down any process. Here is how the signal handler is registered (https://github.com/prestodb/presto/blob/master/presto-native-execution/presto_cpp/main/PrestoServer.cpp#L152)
> 
> Also in the [stop method](https://github.com/prestodb/presto/blob/master/presto-native-execution/presto_cpp/main/SignalHandler.cpp#L31)], server goes in [graceful shotdown mode](https://github.com/prestodb/presto/blob/master/presto-native-execution/presto_cpp/main/PrestoServer.cpp#L890)
> 
> As a result, no explicit endpoint is required to ask the server to shut down. Just sending SIGTERM like any other process works as expected. In Java, perhaps these signal handling is not possible, hence this end point concept came into development. Native worker even have additional signal handling mechnism to catch debugging information in case of SEGFault (https://github.com/prestodb/presto/blob/master/presto-native-execution/presto_cpp/main/PrestoMain.cpp#L25C27-L25C47)
> 
> CC: @spershin

@amitkdutta 

Please refer to this issue : https://github.com/prestodb/presto/issues/23299

amitkdutta:
> > @anandamideShakyan Prestoserver handles graceful shutdown mechanism with SIGTERM, which is a standard way to shut down any process. Here is how the signal handler is registered (https://github.com/prestodb/presto/blob/master/presto-native-execution/presto_cpp/main/PrestoServer.cpp#L152)
> > Also in the [stop method](https://github.com/prestodb/presto/blob/master/presto-native-execution/presto_cpp/main/SignalHandler.cpp#L31)], server goes in [graceful shotdown mode](https://github.com/prestodb/presto/blob/master/presto-native-execution/presto_cpp/main/PrestoServer.cpp#L890)
> > As a result, no explicit endpoint is required to ask the server to shut down. Just sending SIGTERM like any other process works as expected. In Java, perhaps these signal handling is not possible, hence this end point concept came into development. Native worker even have additional signal handling mechnism to catch debugging information in case of SEGFault (https://github.com/prestodb/presto/blob/master/presto-native-execution/presto_cpp/main/PrestoMain.cpp#L25C27-L25C47)
> > CC: @spershin
> 
> @amitkdutta
> 
> Please refer to this issue : #23299

@anandamideShakyan Commented in the issue as well

https://github.com/prestodb/presto/issues/23299#issuecomment-2332210970

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -363,6 +373,40 @@ void PrestoServer::run() {
...
+            if (jsonBody == "SHUTTING_DOWN") {
+                // Print message and initiate shutdown
+                LOG(INFO) << "Shutdown requested";
+                if (server->nodeState() == NodeState::kActive) {
+                    // Call stop() to handle the shutdown process
+                    std::thread([server]() {
+                            server->stop();
+                        }).detach();
+                } else {
+                    // Print message and indicate that shutdown is already requested
```
majetideepak, line 394: State could be `kInActive` as well. So we need to say `Node is inactive or shutdown already requested`

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -363,6 +373,40 @@ void PrestoServer::run() {
...
+                    LOG(INFO) << "Shutdown already requested";
+                }
+                http::sendOkResponse(downstream);
+            } else {
+                // Throw a 400 Bad Request error if the body has some other value
+                http::sendErrorResponse(
+                    downstream,
+                    "Bad Request: Body must contain 'SHUTTING_DOWN'.",http::kHttpBadRequest);
```
majetideepak, line 402: Can you check the Java behavior for this case?

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -363,6 +373,40 @@ void PrestoServer::run() {
...
       });
+  httpServer_->registerPut(
+    "/v1/info/state",
+    [server = this](
+        proxygen::HTTPMessage* /*message*/,
+        const std::vector<std::unique_ptr<folly::IOBuf>>& body ,
+        proxygen::ResponseHandler* downstream) {
+         std::string requestBody=getRequestBody(body);
+         try {
+            folly::dynamic jsonBody = folly::parseJson(requestBody);
```
majetideepak, line 384: Let's use `folly::json_patch::try_parse(folly::parseJson(requestBody))`

majetideepak: We won't need the try-catch with the `folly::json_patch` usage.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -147,6 +148,15 @@ std::string nodeState2String(NodeState nodeState) {
   return fmt::format("<unknown>:{}>", static_cast<int>(nodeState));
 }
 
+std::string getRequestBody(const std::vector<std::unique_ptr<folly::IOBuf>>& body) {
```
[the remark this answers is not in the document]
anandamideShakyan: @czentgr asked me to move the code logic inside a function.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
anandamideShakyan: Ok.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
anandamideShakyan: Done.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
anandamideShakyan: I get the following response for presto java:
Screenshot 2024-10-04 at 11 00 59 AM

It seems that we directly get the error stack. Should I remove "Bad Request:..." message ?

majetideepak: No, this implementation is better.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
majetideepak: Did you push this change?

anandamideShakyan: Actually, I am having issues in using try_parse(), it expects an array as an input something like:  [
            { "op": "replace", "path": "/name", "value": "Jane Doe" },
            { "op": "add", "path": "/location", "value": "New York" }
        ] but our requestBody just contains a string SHUTTING_DOWN so when I am passing it in the function it is throwing parse error. You can see in the attached image I am parsing a string named patchStr and requestBody, patchStr is parsed without am issue whereas requestBody throws parse error (i.e. jsonPatchResult2.hasValue() is 0).
Screenshot 2024-10-04 at 7 19 47 PM
May I move back to using try-catch ?

presto-native-execution/presto_cpp/main/PrestoServer.cpp
majetideepak: I clarified with @czentgr.
We should put the whole block in a function say `handleGracefulShutdown(body)` similar to other `registerGet` implementations. I agree that this will make it more readable.
```
      "/v1/info/state",
      [server = this](
          proxygen::HTTPMessage* /*message*/,
          const std::vector<std::unique_ptr<folly::IOBuf>>& body,
          proxygen::ResponseHandler* /*downstream*/) {
        server->handleGracefulShutdown(body);
      });
```

presto-native-execution/presto_cpp/main/PrestoServer.cpp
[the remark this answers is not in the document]
majetideepak: I clarified with @czentgr.
We should put the whole block in a function say `handleGracefulShutdown(body)` similar to other `registerGet.. {server->reportNodeStatus(downstream)}` implementations. I agree that this will make it more readable.
```
      "/v1/info/state",
      [server = this](
          proxygen::HTTPMessage* /*message*/,
          const std::vector<std::unique_ptr<folly::IOBuf>>& body,
          proxygen::ResponseHandler* /*downstream*/) {
        server->handleGracefulShutdown(body);
      });
```

presto-native-execution/presto_cpp/main/PrestoServer.cpp
majetideepak: I am looking at the folly API on how to use this. There is a test here https://github.com/facebook/folly/blob/e08c228878aaff37aa6abc348154bd5a3e1e82fb/folly/json/test/json_patch_test.cpp#L86

majetideepak: > but our requestBody just contains a string SHUTTING_DOWN so when I am passing it in the function it is throwing parse error

The request is not a json string right? Why parse as json then? Just compare the body directly?

presto-native-execution/presto_cpp/main/PrestoServer.cpp
majetideepak: See discussion above. The body is not JSON string.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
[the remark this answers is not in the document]
anandamideShakyan: Ok, I will make the changes.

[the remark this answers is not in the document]
anandamideShakyan: @majetideepak I have pushed new changes.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -1404,6 +1414,34 @@ void PrestoServer::reportNodeStatus(proxygen::ResponseHandler* downstream) {
   http::sendOkResponse(downstream, json(fetchNodeStatus()));
 }
 
+void PrestoServer::handleGracefulShutdown(const std::vector<std::unique_ptr<folly::IOBuf>>& body , proxygen::ResponseHandler* downstream){
+    std::string requestBody;
+    for (const auto& buf : body) {
+        // Convert IOBuf to fbstring and then to std::string
+        requestBody += buf->moveToFbString().toStdString();
+    }
+    requestBody.erase(requestBody.length()-2);
```
majetideepak, line 1423: Please add a comment describing this erase.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -1404,6 +1414,34 @@ void PrestoServer::reportNodeStatus(proxygen::ResponseHandler* downstream) {
   http::sendOkResponse(downstream, json(fetchNodeStatus()));
 }
 
+void PrestoServer::handleGracefulShutdown(const std::vector<std::unique_ptr<folly::IOBuf>>& body , proxygen::ResponseHandler* downstream){
+    std::string requestBody;
+    for (const auto& buf : body) {
```
majetideepak, line 1419: Isn't it enough to read the first element of the body vector and check that the size of the vector == 1?

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -1404,6 +1414,34 @@ void PrestoServer::reportNodeStatus(proxygen::ResponseHandler* downstream) {
...
+        // Convert IOBuf to fbstring and then to std::string
+        requestBody += buf->moveToFbString().toStdString();
```
majetideepak, line 1421: Can we then use `auto requestBody = body[0]->toString();`?

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -1404,6 +1414,34 @@ void PrestoServer::reportNodeStatus(proxygen::ResponseHandler* downstream) {
...
+     } else {
+         // Print message and indicate that node is inactive or shutdown is already requested
+         LOG(INFO) << "Node is inactive or shutdown is already requested";
+     }
+     http::sendOkResponse(downstream);
+    } else {
+     // Throw a 400 Bad Request error if the body has some other value
+     http::sendErrorResponse(
+         downstream,
+         "Bad Request: Body must contain 'SHUTTING_DOWN'.",http::kHttpBadRequest);
```
majetideepak, line 1441: I am not sure if we should say `Body must contain 'SHUTTING_DOWN'` due to security reasons.
Just `Bad Request` should do?
missing space after comma

presto-native-execution/presto_cpp/main/PrestoServer.h
```diff
@@ -207,6 +207,8 @@ class PrestoServer {
 
   void reportNodeStatus(proxygen::ResponseHandler* downstream);
 
+  void handleGracefulShutdown(const std::vector<std::unique_ptr<folly::IOBuf>>& body , proxygen::ResponseHandler* downstream);
```
majetideepak, line 210: extra space before comma

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -61,6 +62,7 @@
 #include "velox/functions/prestosql/window/WindowFunctionsRegistration.h"
 #include "velox/serializers/PrestoSerializer.h"
 
+
```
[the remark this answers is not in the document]
anandamideShakyan: Done.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -365,6 +367,14 @@ void PrestoServer::run() {
         json infoStateJson = convertNodeState(server->nodeState());
         http::sendOkResponse(downstream, infoStateJson);
       });
+  httpServer_->registerPut(
+     "/v1/info/state",
+     [server = this](
+         proxygen::HTTPMessage* /*message*/,
+         const std::vector<std::unique_ptr<folly::IOBuf>>& body ,
+         proxygen::ResponseHandler* downstream) {
+         server->handleGracefulShutdown(body,downstream);
```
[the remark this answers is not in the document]
anandamideShakyan: Done.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
anandamideShakyan: Done.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
anandamideShakyan: There is no toString() method in folly so I am using moveToFbString() method.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
anandamideShakyan: Done.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
anandamideShakyan: Ok, removing the extra text and adding the missing space.

presto-native-execution/presto_cpp/main/PrestoServer.h
anandamideShakyan: Done.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -1404,6 +1413,31 @@ void PrestoServer::reportNodeStatus(proxygen::ResponseHandler* downstream) {
   http::sendOkResponse(downstream, json(fetchNodeStatus()));
 }
 
+void PrestoServer::handleGracefulShutdown(const std::vector<std::unique_ptr<folly::IOBuf>>& body, proxygen::ResponseHandler* downstream){
+    auto requestBody = body[0]->moveToFbString();
+    // Remove the trailing newline characters from the requestBody
+    requestBody.erase(requestBody.length()-2);
+    if (body.size()==1 && requestBody == "\"SHUTTING_DOWN\"") {
```
majetideepak, line 1420: `body.size() == 1` check should be before `body[0]->...` to handle empty request.
Include the newline in the string check and remove `requestBody.erase(requestBody.length()-2);`
```
if (body.size() == 1 && body[0]->toString() ==  "\"SHUTTING_DOWN\n\"") {

} else {

}
```

anandamideShakyan: Ok I have made the changes. Actually  "\n" is not needed it was a mistake on my side while sending the requests through postman I added a newline in the request body. Should I add code to trim the leading and trailing spaces or I should strictly expect "\"SHUTTING_DOWN\"" in the request body?

majetideepak: > Should I add code to trim the leading and trailing spaces or I should strictly expect ""SHUTTING_DOWN"" in the request body?

Can you check what Java does?

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -1404,6 +1413,28 @@ void PrestoServer::reportNodeStatus(proxygen::ResponseHandler* downstream) {
   http::sendOkResponse(downstream, json(fetchNodeStatus()));
 }
 
+void PrestoServer::handleGracefulShutdown(const std::vector<std::unique_ptr<folly::IOBuf>>& body, proxygen::ResponseHandler* downstream){
```
czentgr, line 1416: The format check is failing. Please fix this (see the Makefile having a format-fix target).

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -1404,6 +1413,28 @@ void PrestoServer::reportNodeStatus(proxygen::ResponseHandler* downstream) {
...
+    if (body.size()==1 && body[0]->moveToFbString() == "\"SHUTTING_DOWN\"") {
+         // Print message and initiate shutdown
```
czentgr, line 1418: We probably don't need all these comments as they explains what happens which is apparent from the code. We use the Velox comment guidelines: https://github.com/facebookincubator/velox/blob/main/CONTRIBUTING.md#coding-best-practices (see "Code Comments" section).

majetideepak: I agree.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
anandamideShakyan: Java allows strings with leading and trailing spaces, that is why I was parsing to json, it was a clean way of removing these spaces. Should I go back to using parseJson() ?

majetideepak: Let's use `folly::trimWhitespace` https://github.com/facebook/folly/blob/main/folly/String.h#L671 if you need to trim spaces.

majetideepak:
Please add a unit test as Aditi requested above.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
anandamideShakyan: Done.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
anandamideShakyan: Done.

anandamideShakyan:
@majetideepak I have added the unit test GracefulShutdownTest.cpp.

majetideepak (commented):
@anandamideShakyan can you check if can use PrestoServerOperations.h?
See the test ServerOperationTest.cpp.
I see that Aditi referenced this as well.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -906,22 +917,30 @@ void PrestoServer::stop() {
       << "Waiting for " << shutdownOnsetSec
       << " second(s) before proceeding with the shutdown...";
 
-  // Give coordinator some time to receive our new node state and stop sending
-  // any tasks.
-  std::this_thread::sleep_for(std::chrono::seconds(shutdownOnsetSec));
+  std::thread shutdownThread([this, shutdownOnsetSec]() {
```
majetideepak, line 920: what is the motivation for a separate shutdown thread?

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -157,7 +158,9 @@ PrestoServer::PrestoServer(const std::string& configDirectoryPath)
 
 PrestoServer::~PrestoServer() {}
 
-void PrestoServer::run() {
+void PrestoServer::run(
+    std::function<void(proxygen::HTTPServer*)> onSuccess,
```
majetideepak, line 162: How are these arguments used?

majetideepak (commented):
> after some test cases we get error The memory manager has already been set

This should not happen. Can you share a reproducer? 
Let's extend `class PrestoServerOperations` if you need to launch presto server in a separate thread.

anandamideShakyan:
> This should not happen. Can you share a reproducer? Let's extend `class PrestoServerOperations` if you need to launch presto server in a separate thread.
@majetideepak 
I extended the ServerOperationTest.cpp to include the graceful shutdown test suite,I get this : 
`
[----------] 7 tests from ServerOperationTest
I20241027 17:00:11.820835 20054 Configs.cpp:120] [PRESTO_STARTUP] Registered properties from '../../../etc/config.properties':
  runtime-metrics-collection-enabled=true
  register-test-functions=true
  http-server.http.port=7777
  shutdown-onset-sec=1
  presto.version=testversion
  discovery.uri=http://127.0.0.1:49595
I20241027 17:00:11.821100 20054 Configs.cpp:120] [PRESTO_STARTUP] Registered properties from '../../../etc/node.properties':
  node.internal-address=127.0.0.1
  node.location=testing-location
  node.environment=testing
I20241027 17:00:11.821332 20054 Configs.cpp:120] [PRESTO_STARTUP] Registered properties from '../../../etc/velox.properties':
  mutable-config=true
I20241027 17:00:11.822000 20054 Configs.cpp:900] [PRESTO_STARTUP] Updated in '../../../etc/velox.properties' from SystemProperties:
  max_partial_aggregation_memory=16777216
  max_page_partitioning_buffer_size=33554432
  max_output_buffer_size=33554432
  presto.array_agg.ignore_nulls=false
I20241027 17:00:11.822346 20054 PrestoServer.cpp:799] [PRESTO_STARTUP] Starting with node memory 40GB
E20241027 17:00:11.822404 20054 Exceptions.h:66] Line: /Users/shakyan/opensource/presto/presto-native-execution/velox/velox/common/memory/Memory.cpp:177, Function:initialize, Expression: instance == nullptr The memory manager has already been set: Memory Manager[capacity UNLIMITED alignment 64B usedBytes 0B number of pools 2
List of root pools:
	__sys_root__
Memory Allocator[MALLOC capacity UNLIMITED allocated bytes 0 allocated pages 0 mapped pages 0]
ARBIRTATOR[NOOP CAPACITY[UNLIMITED]]], Source: RUNTIME, ErrorCode: INVALID_STATE
libc++abi: terminating due to uncaught exception of type facebook::velox::VeloxRuntimeError: Exception: VeloxRuntimeError
Error Source: RUNTIME
Error Code: INVALID_STATE
Reason: The memory manager has already been set: Memory Manager[capacity UNLIMITED alignment 64B usedBytes 0B number of pools 2
List of root pools:
	__sys_root__
Memory Allocator[MALLOC capacity UNLIMITED allocated bytes 0 allocated pages 0 mapped pages 0]
ARBIRTATOR[NOOP CAPACITY[UNLIMITED]]]
Retriable: False
Expression: instance == nullptr
Function: initialize
/Users/shakyan/opensource/presto/presto-native-execution/velox/velox/common/memory/Memory.cpp
Line: 177
Process finished with exit code 134 (interrupted by signal 6:SIGABRT)
`
Steps to reproduce:
Add the following to ServerOperationTest.cpp
```
#include "presto_cpp/main/http/tests/HttpTestBase.h"
#include "presto_cpp/main/tests/PrestoServerWrapper.h"
.......
namespace facebook::presto {
static std::unique_ptr<facebook::presto::PrestoServer> getPrestoServer() {
    auto server = std::make_unique<PrestoServer>("../../../etc");
    return server;
}
........
class ServerOperationTest : public testing::Test {
protected:
    static facebook::presto::test::PrestoServerWrapper* wrapper;
    static folly::SocketAddress* socketAddress;
    static void SetUpTestSuite() {
        folly::SingletonVault::singleton()->registrationComplete();
#ifndef PRESTO_STATS_REPORTER_TYPE
        // Initialize singleton for the reporter.
        folly::Singleton<facebook::velox::BaseStatsReporter> reporter(
            []() { return new facebook::velox::DummyStatsReporter(); });
#endif
        static auto prestoServer = getPrestoServer();
        wrapper = new facebook::presto::test::PrestoServerWrapper(
            std::move(prestoServer));
        socketAddress = new folly::SocketAddress(wrapper->start().get());
    }

    static void TearDownTestSuite() {
        wrapper->stop();
    }
.........
};
facebook::presto::test::PrestoServerWrapper* ServerOperationTest::wrapper =
    nullptr;
folly::SocketAddress* ServerOperationTest::socketAddress = nullptr;
..........
TEST_F(ServerOperationTest, TestGetState) {
    auto memoryPool = memory::MemoryManager::getInstance()->addLeafPool("");
    HttpClientFactory clientFactory;
    auto client = clientFactory.newClient(
        *socketAddress,
        std::chrono::milliseconds(1'000),
        std::chrono::milliseconds(0),
        false,
        memoryPool);
    {
        auto response = sendGet(client.get(), "/v1/info/state").get();
        ASSERT_EQ(response->headers()->getStatusCode(), http::kHttpOk);
    }
}
TEST_F(ServerOperationTest, TestSendPutShuttingDown) {
    auto memoryPool = memory::MemoryManager::getInstance()->addLeafPool("");
    HttpClientFactory clientFactory;
    auto client = clientFactory.newClient(
        *socketAddress,
        std::chrono::milliseconds(1'000),
        std::chrono::milliseconds(0),
        false,
        memoryPool);
    {
        std::string emptyBody = "";
        auto response = sendPut(client.get(), "/v1/info/state", 0, emptyBody).get();
        ASSERT_EQ(
            response->headers()->getStatusCode(),
            http::kHttpBadRequest); // Assuming empty request is bad
        ASSERT_EQ(bodyAsString(*response, memoryPool.get()), "Bad Request");
    }
    {
        std::string invalidBody = "\"SHUTTING_DWN\"";
        auto response =
            sendPut(client.get(), "/v1/info/state", 0, invalidBody).get();
        ASSERT_EQ(response->headers()->getStatusCode(), http::kHttpBadRequest);
        ASSERT_EQ(bodyAsString(*response, memoryPool.get()), "Bad Request");
    }
    {
        std::string body = "\"SHUTTING_DOWN\"";
        auto response = sendPut(client.get(), "/v1/info/state", 0, body).get();
        ASSERT_EQ(response->headers()->getStatusCode(), http::kHttpOk);
    }
}
} // namespace facebook::presto
```
While working on the issue I discussed the errors that I was getting with @nmahadevuni, he suggested me to make gracefulshutdown as a separate file and a separate executable. This solved the issue. Aditi also said that I can use a new file.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
anandamideShakyan: We use it inside PrestoServerWrapper.cpp to get the socket address of the server that we need while running the test suite(see SetUpTestSuite()).

presto-native-execution/presto_cpp/main/PrestoServer.cpp
anandamideShakyan: We were not able to read the server status if we didn't schedule the shutdown in a separate thread because before we could read the status, server was shutting down.

anandamideShakyan:
@majetideepak , I can  add the test suite as a mock test. But it still won't test the whole client->RESTEndpoint->response.
I can shift testcase to ServerOperationsTest.cpp and use serverOperationSetState with actual server created. But we would still have that testcase as a separate executable. Also it doesn't test the REST endpoint, it will call the shutdown method directly. Do you want me to go ahead with this approach?

presto-native-execution/presto_cpp/main/PrestoServer.cpp
majetideepak: It is not ideal to make API changes just for testing. Let's figure out another way to test.

majetideepak:
> test the whole client->RESTEndpoint->response

@anandamideShakyan We already have the setup for such e2e in the Java tests. Can we add a test there?

presto-native-execution/presto_cpp/main/PrestoServer.cpp
anandamideShakyan: We need it to compare it against the status codes in the test cases.

majetideepak: Again not ideal to make changes for testing.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -1421,6 +1429,23 @@ void PrestoServer::reportNodeStatus(proxygen::ResponseHandler* downstream) {
...
 }
 
+void PrestoServer::handleGracefulShutdown(
+    const std::vector<std::unique_ptr<folly::IOBuf>>& body,
+    proxygen::ResponseHandler* downstream) {
+  if (body.size() == 1 &&
+      folly::trimWhitespace(body[0]->moveToFbString()) == "\"SHUTTING_DOWN\"") {
+    LOG(INFO) << "Shutdown requested";
+    if (nodeState() == NodeState::kActive) {
+      std::thread([this]() { this->stop(); }).detach();
```
majetideepak, line 1439: Why do we invoke stop in a separate thread? Please add a comment.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/TestPrestoNativeGracefulShutdown.java
```diff
@@ -0,0 +1,80 @@
...
+            return 500;
+        }
+    }
+
+    private QueryRunner createQueryRunner() throws Exception
+    {
+        return PrestoNativeQueryRunnerUtils.createNativeQueryRunner(true);
+    }
+
+    private String getServerUri(QueryRunner queryRunner)
```
majetideepak, line 74: nit: do we need a function if it is used only once?

majetideepak (commented):
@anandamideShakyan This looks great! I have 2 minor comments.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/TestPrestoNativeGracefulShutdown.java
anandamideShakyan: done

presto-native-execution/presto_cpp/main/PrestoServer.cpp
anandamideShakyan: Added comment.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/TestPrestoNativeGracefulShutdown.java
```diff
@@ -0,0 +1,70 @@
...
+
+        int responseCode = sendShutdownRequest(baseUri, "\"INVALID_SHUTDOWN\"");
+        assertEquals(responseCode, 400, "Expected a 400 Bad Request response for invalid body");
+
+        responseCode = sendShutdownRequest(baseUri, "");
+        assertEquals(responseCode, 400, "Expected a 400 Bad Request response for empty body");
+
+        responseCode = sendShutdownRequest(baseUri, "\"SHUTTING_DOWN\"");
+        assertEquals(responseCode, 200, "Expected a 200 OK response for valid shutdown request");
+
```
majetideepak, line 46: We should add some tests here for the shutdown. Can we check the state here?

majetideepak (commented):
@anandamideShakyan One more comment. 
Can you clarify in the Java implementation if the process also exits(ends) after this request?

anandamideShakyan:
> @anandamideShakyan One more comment. Can you clarify in the Java implementation if the process also exits(ends) after this request?

When we send PUT request to /v1/info/state with "SHUTTING_DOWN" as request body the server initiates shutdown after  waiting for certain time, the api remains active during that duration but once the server shuts down the api also shuts down.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/TestPrestoNativeGracefulShutdown.java
anandamideShakyan: @majetideepak  We can get the state of coordinator by sending get request to v1/info/state after the put request and compare with shutting down, there is a method for it inside distributedQueryRunner named getCoordinatorInfoState but it is private so we cannot call it.

majetideepak:
@anandamideShakyan I am wondering if the worker process exits after the shutdown process. If it does, then we might only need to call `kill(getpid(), SIGTERM);`. The signal handler will call the stop() before that.
https://github.com/prestodb/presto/blob/master/presto-native-execution/presto_cpp/main/SignalHandler.cpp

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/TestPrestoNativeGracefulShutdown.java
```diff
@@ -0,0 +1,99 @@
...
+        assertEquals(state, "\"SHUTTING_DOWN\"");
+
+        queryRunner.close();
+    }
+
+    public static int sendShutdownRequest(String shutdownUrl, String body)
+    {
+        try {
+            URL url = new URL(shutdownUrl);
+            HttpURLConnection connection = (HttpURLConnection) url.openConnection();
```
majetideepak, line 83: DistributedQueryRunner has a HttpClient. Can we not use that?
That API seems simpler. The same is true above to check the status of the node.

anandamideShakyan: getCoordinatorInfoState() is private, so is client. Do you want me to extend DistributedQueryRunner class with setter getters?

anandamideShakyan: Thanks for the insight. The code looks much cleaner. I have one issue though I am not able to extract statusCode from JettyClient because JettyResponse is private and I need to pass a custom responseHandler to get the status code so I am using the same java.net.HttpURLConnection class that I was using in sendShutdownRequest function.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/TestPrestoNativeGracefulShutdown.java
```diff
@@ -0,0 +1,73 @@
...
+        QueryRunner queryRunner = PrestoNativeQueryRunnerUtils.createNativeQueryRunner(true);
+        DistributedQueryRunner distributedQueryRunner = (DistributedQueryRunner) queryRunner;
+
+        int responseCode = distributedQueryRunner.sendWorkerShutdown(0,"INVALID_BODY");
+        assertEquals(responseCode, 400, "Expected a 400 Bad Request response for invalid body");
+
+        responseCode = distributedQueryRunner.sendWorkerShutdown(0,"");
+        assertEquals(responseCode, 400, "Expected a 400 Bad Request response for empty body");
+
+        responseCode =distributedQueryRunner.sendWorkerShutdown(0,"\"SHUTTING_DOWN\"");
```
aditi-pandit, line 45: Can we add a bit more elaborate test ? Like waiting for some more time post this request to check that the worker has infact shutdown.

What are the e2e use-cases of this functionality ? Might be worth having an e2e test for those.

presto-tests/src/main/java/com/facebook/presto/tests/DistributedQueryRunner.java
```diff
@@ -413,6 +420,39 @@ private NodeState getCoordinatorInfoState(int coordinator)
...
+        Request request = prepareGet()
+                .setHeader(PRESTO_USER, DEFAULT_USER)
+                .setUri(uri)
+                .build();
+
+        NodeState state = client.execute(request, createJsonResponseHandler(jsonCodec(NodeState.class)));
+        return state;
+    }
+
+    public int sendWorkerShutdown(int worker,String body)
```
majetideepak, line 435: nit: missing space after `,`.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/TestPrestoNativeGracefulShutdown.java
majetideepak, line 45: nit: space after `=`

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/TestPrestoNativeGracefulShutdown.java
```diff
@@ -0,0 +1,73 @@
...
+        assertEquals(responseCode, 200, "Expected a 200 OK response for valid shutdown request");
+        NodeState state = distributedQueryRunner.getWorkerInfoState(0);
+        assertEquals(state.getValue(), 3);
+
+        queryRunner.close();
+    }
+
+    public static int sendShutdownRequest(String shutdownUrl, String body)
```
majetideepak, line 53: Clean this code up? Not used anymore.

majetideepak (commented):
@anandamideShakyan This looks great. Few minor comments. Thanks for all the refactoring!

majetideepak:
maven checks are also failing.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/TestPrestoNativeGracefulShutdown.java
majetideepak: > What are the e2e use-cases of this functionality ? Might be worth having an e2e test for those.

The graceful shutdown feature in Presto is used by other apps on top of Presto such as auto-scaling. I think ensuring the node goes down is sufficient for the Prestissimo E2E tests.
@anandamideShakyan can you check what tests exist on the Java side?

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/TestPrestoNativeGracefulShutdown.java
anandamideShakyan: Done.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/TestPrestoNativeGracefulShutdown.java
anandamideShakyan: Done.

presto-tests/src/main/java/com/facebook/presto/tests/DistributedQueryRunner.java
anandamideShakyan: Done.

presto-tests/src/main/java/com/facebook/presto/tests/DistributedQueryRunner.java
```diff
@@ -413,6 +417,39 @@ private NodeState getCoordinatorInfoState(int coordinator)
...
+        Request request = prepareGet()
+                .setHeader(PRESTO_USER, DEFAULT_USER)
+                .setUri(uri)
+                .build();
+
+        NodeState state = client.execute(request, createJsonResponseHandler(jsonCodec(NodeState.class)));
+        return state;
+    }
+
+    public int sendWorkerShutdown(int worker, String body)
```
majetideepak, line 432: nit: It is odd to allow a body for the Shutdown.
May be this should be renamed as a generic `sendWorkerRequest`

presto-native-execution/presto_cpp/main/http/HttpConstants.h
```diff
@@ -17,11 +17,13 @@ namespace facebook::presto::http {
...
 const uint16_t kHttpAccepted = 202;
 const uint16_t kHttpNoContent = 204;
+const uint16_t kHttpBadRequest = 400;
 const uint16_t kHttpUnauthorized = 401;
 const uint16_t kHttpNotFound = 404;
 const uint16_t kHttpInternalServerError = 500;
 
 const char kMimeTypeApplicationJson[] = "application/json";
 const char kMimeTypeApplicationThrift[] = "application/x-thrift+binary";
+const char SHUTTING_DOWN[] = "\"SHUTTING_DOWN\"";
```
majetideepak, line 27: Constant should be `kShuttingDown`

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/TestPrestoNativeGracefulShutdown.java
```diff
@@ -0,0 +1,52 @@
...
+        QueryRunner queryRunner = PrestoNativeQueryRunnerUtils.createNativeQueryRunner(true);
+        DistributedQueryRunner distributedQueryRunner = (DistributedQueryRunner) queryRunner;
+
+        int responseCode = distributedQueryRunner.sendWorkerShutdown(0,"INVALID_BODY");
+        assertEquals(responseCode, 400, "Expected a 400 Bad Request response for invalid body");
+
+        responseCode = distributedQueryRunner.sendWorkerShutdown(0,"");
+        assertEquals(responseCode, 400, "Expected a 400 Bad Request response for empty body");
+
+        responseCode = distributedQueryRunner.sendWorkerShutdown(0,"\"SHUTTING_DOWN\"");
```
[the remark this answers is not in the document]
anandamideShakyan: The constant is defined in .h file, I am not sure how I can use that in java. Do I need to add the constant somewhere else for java?

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -1435,6 +1443,30 @@ void PrestoServer::reportNodeStatus(proxygen::ResponseHandler* downstream) {
...
+    } else {
+      LOG(INFO) << "Node is inactive or shutdown is already requested";
+    }
+    http::sendOkResponse(downstream);
+  } else {
+    std::string bodyContent;
+    for (const auto& buf : body) {
+        bodyContent+=std::string(folly::trimWhitespace(buf->clone()->moveToFbString()));
+    }
+    LOG(ERROR) << "Bad Request. Received body content: "<<bodyContent;
```
majetideepak, line 1465: nit: space after `<<`

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/TestPrestoNativeGracefulShutdown.java
```diff
@@ -0,0 +1,45 @@
...
+        int responseCode = distributedQueryRunner.sendWorkerRequest(0,"INVALID_BODY");
+        assertEquals(responseCode, 400, "Expected a 400 Bad Request response for invalid body");
+
+        responseCode = distributedQueryRunner.sendWorkerRequest(0,"");
+        assertEquals(responseCode, 400, "Expected a 400 Bad Request response for empty body");
+
+        responseCode = distributedQueryRunner.sendWorkerRequest(0,"\"SHUTTING_DOWN\"");
+        assertEquals(responseCode, 200, "Expected a 200 OK response for valid shutdown request");
+        NodeState state = distributedQueryRunner.getWorkerInfoState(0);
+        assertEquals(state.getValue(), 3);
```
majetideepak, line 41: nit:: can we compare against the enum value SHUTTING_DOWN?

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/TestPrestoNativeGracefulShutdown.java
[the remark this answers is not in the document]
majetideepak: I missed this is in Java. Nevermind.

majetideepak:
@anandamideShakyan can you fix the maven failure?
@amitkdutta can you take a look at this PR? Thanks.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/TestPrestoNativeGracefulShutdown.java
anandamideShakyan: We are sending "\"SHUTTING_DOWN\"" in the body, whereas the value of SHUTTING_DOWN.toString() is SHUTTING_DOWN so I need to additionally put "" in sendWorkerRequest or create a new enum?

presto-native-execution/presto_cpp/main/PrestoServer.cpp
anandamideShakyan: Actually this is not working right now I am getting empty line, working on a fix.

anandamideShakyan: Done.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/TestPrestoNativeGracefulShutdown.java
majetideepak: I meant using `assertEquals(state.getValue(), NodeState.SHUTTING_DOWN)`

majetideepak (approved):

amitkdutta:
Looks good to me. Just a noob question before merging: this adds a new endpoint to call existing shutdown from an API, but nothing changes from existing automatic signal handling based shutdown and that will work as is. @anandamideShakyan @majetideepak will be great if you can confirm.

majetideepak:
@amitkdutta Yes, no changes to existing signal handling. You also need to remove the requested changes for us to be able to merge :).

tdcmeehan (approved):

PR #23519, opened 2024-08-26
[native] Fix conversion of partitioning const keys.

spershin:
## Description
Constant keys for output partitioning were being converted to nulls.
This caused incorrect query results, as rows were sent to wrong workers.

## Test Plan
Added unit test to reflect the situation.

```
== NO RELEASE NOTE ==
```


aditi-pandit (approved):
Nice catch ! Thanks @spershin

rschlussel (commented):
can you add an end-to-end correctness test too (query that would get wrong results without this change)?

Also, clean up/replace the json plan to remove all internal references, custom fields, table/column names, etc.

aditi-pandit (changes requested):
+1 for @rschlussel's comments about e2e test.

Agree with your changes in the PrestoToVeloxQueryPlan logic. The original code didn't handle the ConstantExpr correctly.

I too am interested to see the e2e query in this case.

spershin:
@rschlussel , @aditi-pandit 
I don't know how to compose a query that would reproduce the issue in the E2E test.
Based on our minimal query that's reproducible in the prod environment
```
SELECT 
  session_guid, 
  set_union(source) source, 
  set_union(source_pipeline) source_pipeline 
FROM 
  (
    SELECT -- Fragment 2
      session_guid session_guid, 
      ARRAY['insight_to_action'] source, 
      ARRAY[version] source_pipeline 
    FROM 
      (
        SELECT 
          * 
        FROM 
          edm_rule_based_session_labels_hourly 
        WHERE 
          ((((ds = '2024-08-10') AND (ts IN ('2024-08-10+07:00:99'))) AND (universe = 'FB')) AND (rule_mode = 'prod'))
      ) 
    UNION ALL 
    SELECT -- Fragment 3
      session_id session_guid, 
      ARRAY['cluster_label_v2'] source, 
      "set_agg"(model) source_pipeline 
    FROM 
      (
        SELECT 
          * 
        FROM 
          fb_cluster_labeling_session_label_table 
        WHERE 
          (((ds = '2024-08-10') AND (cluster_name IN ('http_header_order'))) AND (model = 'rule_based'))
      ) 
    GROUP BY 
      1, 2
  ) 
WHERE 
  session_guid in ('100003902118689:16:1723259469', '888888888888888:16:1723259469', '999999999999999:16:1723259469') 
GROUP BY 
  session_guid;
```

I tried to create a query for tpch, but it does not have constant partitioning keys for some reason and thus does not reproduce the issue:
```
SELECT 
  name, 
  set_union(source) source, 
  set_union(source_pipeline) source_pipeline 
FROM 
  (
    SELECT -- Fragment 2
      name name, -- varchar
      ARRAY['insight_to_action'] source, -- constant
      ARRAY[comment] source_pipeline -- varchar
    FROM 
      (
        SELECT 
          * 
        FROM 
          nation 
      ) 
    UNION ALL 
    SELECT -- Fragment 3
      name name, -- varchar
      ARRAY['cluster_label_v2'] source, -- constant
      set_agg(comment) source_pipeline -- varchar
    FROM 
      (
        SELECT 
          * 
        FROM 
          nation_partitioned_ds 
      ) 
    GROUP BY 
      1, 2
  ) 
WHERE 
  name not in ('100003902118689:16:1723259469', '888888888888888:16:1723259469', '999999999999999:16:1723259469') 
GROUP BY 
  name;
```

Maybe there are some session properties involved, however I'm not sure if any of these would affect it:
```
aggregation_spill_enabled	true
enable_dynamic_filtering	false
inline_sql_functions	false
join_spill_enabled	false
order_by_spill_enabled	true
partitioning_provider_catalog	system
prism_batch.bucket_function_type_for_exchange	HIVE_COMPATIBLE
prism_batch.dynamic_split_sizes_enabled	true
prism_batch.pushdown_filter_enabled	true
spill_enabled	true
```

There could be a questions of scan partitioning, filters on select, table size, etc. that play role.
But I'm unable to come up with a E2E query top reproduce the issue.

spershin:
Updated unit test to check partitionFunctionSpec and cleaned up the test plan.

amitkdutta:
@aditi-pandit @rschlussel Looks like we haven't found a good e2e case. Since it's a bit non trivial to find a case, willing to merge it asap since it resolves a critical correctness bug and already has a good unit test.

aditi-pandit (approved):
@spershin, @amitkdutta  : I'm generally in agreement, but would want to make a revised version of the e2e query. Will give it a shot.

spershin:
> @spershin : Can you give more information about the cardinality of Fragment 2, Fragment 3 and their inner sub-selects ?

@aditi-pandit 
See the cardinalities in the comments to all selects.

```
SELECT -- 366'006 rows if without session_guid filter 
  session_guid, 
  set_union(source) source, 
  set_union(source_pipeline) source_pipeline 
FROM 
  (
    SELECT -- Fragment 2 (313,430)
      session_guid session_guid, 
      ARRAY['insight_to_action'] source, 
      ARRAY[version] source_pipeline 
    FROM 
      (
        SELECT -- 313,430 out of 73,706,428 in that ds
          * 
        FROM 
          edm_rule_based_session_labels_hourly 
        WHERE 
          ((((ds = '2024-08-10') AND (ts IN ('2024-08-10+07:00:99'))) AND (universe = 'FB')) AND (rule_mode = 'prod'))
      ) 
    UNION ALL 
    SELECT -- Fragment 3 (110'019)
      session_id session_guid, 
      ARRAY['cluster_label_v2'] source, 
      "set_agg"(model) source_pipeline 
    FROM 
      (
        SELECT -- 225,540 out of 225'540 in that ds
          * 
        FROM 
          fb_cluster_labeling_session_label_table 
        WHERE 
          (((ds = '2024-08-10') AND (cluster_name IN ('http_header_order'))) AND (model = 'rule_based'))
      ) 
    GROUP BY 
      1, 2
  ) 
WHERE 
  session_guid in ('100003902118689:16:1723259469', '888888888888888:16:1723259469', '999999999999999:16:1723259469') 
GROUP BY 
  session_guid;
```

Note that some tiny variations (w/o obvious connection to the partitioning) here and there make it run correctly.
For example, if I remove the 'session_guid' filter and wrap the query into the count(*) both Java and C++ return the same number - in that case likely something get optimized out as we don't use the fields.
Or if I remove some aggregates - it stops reproducing.

PR #23527, opened 2024-08-27
[native] Add session property for partial aggregation and spill

kewang1024:
```
== RELEASE NOTES ==

General Changes
* Add a session property `native_max_partial_aggregation_memory` which specifies presto native max partial aggregation memory when data reduction is not optimal :pr:`23527`
* Add a session property `native_max_extended_partial_aggregation_memory` which specifies presto native max partial aggregation memory when data reduction is optimal :pr:`23527`
* Add a session property `native_max_spill_bytes ` which specifies presto native max allowed spill bytes :pr:`23527`

```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
+++ b/presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
@@ -350,6 +350,10 @@ public final class SystemSessionProperties
     private static final String NATIVE_EXECUTION_PROGRAM_ARGUMENTS = "native_execution_program_arguments";
     public static final String NATIVE_EXECUTION_PROCESS_REUSE_ENABLED = "native_execution_process_reuse_enabled";
     public static final String NATIVE_DEBUG_VALIDATE_OUTPUT_FROM_OPERATORS = "native_debug_validate_output_from_operators";
+
+    public static final String NATIVE_MAX_PARTIAL_AGGREGATION_MEMORY = "native_max_partial_aggregation_memory";
+    public static final String NATIVE_MAX_EXTENDED_PARTIAL_AGGREGATION_MEMORY = "native_max_extended_partial_aggregation_memory";
+    public static final String NATIVE_MAX_SPILL_BYTES = "native_max_spill_bytes";
     public static final String DEFAULT_VIEW_SECURITY_MODE = "default_view_security_mode";
     public static final String JOIN_PREFILTER_BUILD_SIDE = "join_prefilter_build_side";
     public static final String OPTIMIZER_USE_HISTOGRAMS = "optimizer_use_histograms";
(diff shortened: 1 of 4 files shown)


xiaoxmeng (dismissed):
@kewang1024 LGTM. Thanks!

xiaoxmeng (commented):
@kewang1024 do we still have the sanity check for the native session conversion in Prestissimo? Thanks!

presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
```diff
@@ -350,6 +350,9 @@ public final class SystemSessionProperties
     private static final String NATIVE_EXECUTION_PROGRAM_ARGUMENTS = "native_execution_program_arguments";
     public static final String NATIVE_EXECUTION_PROCESS_REUSE_ENABLED = "native_execution_process_reuse_enabled";
     public static final String NATIVE_DEBUG_VALIDATE_OUTPUT_FROM_OPERATORS = "native_debug_validate_output_from_operators";
+
+    public static final String NATIVE_MAX_PARTIAL_AGGREGATION_MEMORY = "native_max_partial_aggregation_memory";
```
xiaoxmeng, line 354: Do you want to expose max_extended_partial_aggregation_memory as well? Velox supports to dynamically adjust aggregation memory limit in a range defined by the above two configs? Thanks!

kewang1024:
> Should this be documented on https://prestodb.io/docs/current/presto_cpp/properties.html ?

Actually this is session property which is different than the configs, so should not be added here.
But we can start a new wiki to document the session properties, do you think you can help with this effort? @steveburnett

presto-native-execution/presto_cpp/main/SessionProperties.h
```diff
@@ -86,6 +98,8 @@ class SessionProperties {
   /// The maximum allowed spill file size.
   static constexpr const char* kMaxSpillFileSize = "native_max_spill_file_size";
 
+  static constexpr const char* kMaxSpillBytes = "native_max_spill_bytes";
```
aditi-pandit, line 101: It might be good to specify all the operators (Hashjoin, HashAgg, OrderBy, RowNumber, Window etc) to which this value applies

aditi-pandit (commented):
Thanks @kewang1024 for this code.

Also as per @steveburnett's comment we should document atleast native_max_spill_bytes at https://prestodb.io/docs/current/presto_cpp/properties.html

yingsu00 (commented):
@kewang1024 This PR needs the release note. Will you be able to add one? Thanks.

presto-native-execution/presto_cpp/main/SessionProperties.h
kewang1024: I don't have much context on spilling so might not have the full list, @xiaoxmeng do you know the full list? 

I can add in a separate PR. We need this change to make an internal release cut.

steveburnett:
> > Should this be documented on https://prestodb.io/docs/current/presto_cpp/properties.html ?
> 
> Actually this is session property which is different than the configs, so should not be added here. But we can start a new wiki to document the session properties, do you think you can help with setting up new page? @steveburnett

Like the Presto [Properties Reference](https://prestodb.io/docs/current/admin/properties.html) page, the [Presto C++ Properties Reference](https://prestodb.io/docs/current/presto_cpp/properties.html) page begins with the statement "The following pages are not a complete list of all configuration and session properties ". 

I know the sentence before that reads "This section describes Presto C++ configuration properties." which is exclusionary, but that could be changed to "This section describes Presto C++ configuration and session properties." and we don't have a separate session properties page in the Presto doc. (Maybe we should.) 

I don't think we need a separate page for Presto C++ session properties, so I think all three of these session properties could be added to [Presto C++ Properties Reference](https://prestodb.io/docs/current/presto_cpp/properties.html). 

We could highlight the distinction for session properties in the description. For example, in your current draft adding ``max_spill_bytes`` to Presto C++ Properties Reference, the description could read "_This session property_ (my new words emphasized here for clarity) specifies the max spill bytes limit set for each query." 

If you or others feel strongly we need a separate page for Presto C++ Session Properties, I am open to the idea. Help me understand why a separate page is important.

NikhilCollooru (approved):

spershin (approved):

presto-native-execution/presto_cpp/main/SessionProperties.h
xiaoxmeng: @kewang1024  @aditi-pandit: this applies at task level and is not specific to a particular operator.

Issue #23529, opened 2024-08-27
Iceberg `timestamptz` should map to Presto `TIMESTAMP WITH TIME ZONE` type


tdcmeehan:
## Your Environment

* Presto version used: Any
* Storage (HDFS/S3/GCS..): NA
* Data source and connector used: Iceberg
* Deployment (Cloud or On-prem): NA
* [Pastebin](https://pastebin.com/) link to the complete debug logs:

## Expected Behavior
Currently, the Presto Iceberg connector does not differentiate between `timestamp` and `timestamptz`, mapping both to `TIMESTAMP`.  This is incorrect, especially when `deprecated.legacy-timestamp` is set to `false`, because `timestamptz` represents a point in time value, whereas Presto's `TIMESTAMP` represents a timestamp that is not a point in time.

## Current Behavior
All Iceberg `timestamp` and `timestamptz`s are being mapped to `TIMESTAMP`

## Possible Solution
Improve the type mapping to use `TIMESTAMP WITH TIME ZONE`

## Steps to Reproduce

1.
2.
3.
4.

auden-woolfson:
I think updating the timestamp case in `TypeConverter` `toPrestoType` to...

```
  case TIMESTAMP:
      Types.TimestampType timestampType = (Types.TimestampType) type.asPrimitiveType();
      if (timestampType.shouldAdjustToUTC()) {
          return TimestampWithTimeZoneType.TIMESTAMP_WITH_TIME_ZONE;
      }
      return TimestampType.TIMESTAMP;
```
should fix this

PR #22926, opened 2024-06-05
Disable timestamp with time zone in create, alter and insert statements

denodo-research-labs:
Timestamp with time zone is not allowed in DDL statements because inserts fail, but tables with timestamp with time zone can be registered and queried.

## Description
Presto issue #22794 

## Motivation and Context
When querying an Iceberg table in AWS Glue the following error is received: 'Iceberg column type timestamptz is not supported'. This worked before Presto 0.285. It is related to #21096 

## Impact
Create table, alter table and insert statements having "timestamp with time zone" will return error for Iceberg tables:
'Iceberg column type timestamptz is not supported'.
But existent Iceberg tables with timestamp with time zone can be registered and queried.

## Test Plan
Add a new test in IcebergDistributedSmokeTestBase

## Contributor checklist

- [x] Please make sure your submission complies with our [development](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#development), [formatting](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#formatting), [commit message](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#commit-formatting-and-pull-requests), and [attribution guidelines](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#attribution).
- [x] PR description addresses the issue accurately and concisely.  If the change is non-trivial, a GitHub Issue is referenced.
- [x] Documented new properties (with its default value), SQL syntax, functions, or other functionality.
- [x] If release notes are required, they follow the [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines).
- [x] Adequate tests were added if applicable.
- [x] CI passed.

## Release Notes
Please follow [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines) and fill in the release notes below.

```
== RELEASE NOTES ==

Iceberg Connector Changes
* Disable timestamp with time zone in create, alter and insert statements :pr:`22926`

PR #23534, opened 2024-08-27
Fix timestamp with timezone mapping in iceberg type converter

auden-woolfson:
## Description
Fixes bug described in #23529

```
== RELEASE NOTES ==

Iceberg Connector Changes
* Add logic to iceberg type converter for timestamp with timezone :pr:`23534`
```

### DIFF
--- a/presto-docs/src/main/sphinx/connector/iceberg.rst
+++ b/presto-docs/src/main/sphinx/connector/iceberg.rst
@@ -1775,6 +1775,10 @@ Map of Iceberg types to the relevant PrestoDB types:
     - ``TIME``
   * - ``TIMESTAMP``
     - ``TIMESTAMP``
+  * - ``TIMESTAMP``
+    - ``TIMESTAMP_WITH_TIMEZONE``
+  * - ``STRING``
+    - ``VARCHAR``
   * - ``UUID``
     - ``UUID``
   * - ``LIST``
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergHiveMetadata.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergHiveMetadata.java
@@ -124,7 +124,6 @@ import static com.facebook.presto.iceberg.IcebergUtil.isIcebergTable;
 import static com.facebook.presto.iceberg.IcebergUtil.populateTableProperties;
 import static com.facebook.presto.iceberg.IcebergUtil.toHiveColumns;
 import static com.facebook.presto.iceberg.IcebergUtil.tryGetProperties;
-import static com.facebook.presto.iceberg.IcebergUtil.verifyTypeSupported;
 import static com.facebook.presto.iceberg.PartitionFields.parsePartitionFields;
 import static com.facebook.presto.iceberg.PartitionSpecConverter.toPrestoPartitionSpec;
 import static com.facebook.presto.iceberg.SchemaConverter.toPrestoSchema;
@@ -306,8 +305,6 @@ public class IcebergHiveMetadata
         String tableName = schemaTableName.getTableName();
 
         Schema schema = toIcebergSchema(tableMetadata.getColumns());
-
-        verifyTypeSupported(schema);
 
         PartitionSpec partitionSpec = parsePartitionFields(schema, getPartitioning(tableMetadata.getProperties()));
 
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergNativeMetadata.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergNativeMetadata.java
@@ -67,7 +67,6 @@ import static com.facebook.presto.iceberg.IcebergUtil.getColumns;
 import static com.facebook.presto.iceberg.IcebergUtil.getNativeIcebergTable;
 import static com.facebook.presto.iceberg.IcebergUtil.getNativeIcebergView;
 import static com.facebook.presto.iceberg.IcebergUtil.populateTableProperties;
-import static com.facebook.presto.iceberg.IcebergUtil.verifyTypeSupported;
 import static com.facebook.presto.iceberg.PartitionFields.parsePartitionFields;
 import static com.facebook.presto.iceberg.PartitionSpecConverter.toPrestoPartitionSpec;
 import static com.facebook.presto.iceberg.SchemaConverter.toPrestoSchema;
@@ -307,8 +306,6 @@ public class IcebergNativeMetadata
         String tableName = schemaTableName.getTableName();
 
         Schema schema = toIcebergSchema(tableMetadata.getColumns());
-
-        verifyTypeSupported(schema);
 
         PartitionSpec partitionSpec = parsePartitionFields(schema, getPartitioning(tableMetadata.getProperties()));
         FileFormat fileFormat = getFileFormat(tableMetadata.getProperties());
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/TypeConverter.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/TypeConverter.java
@@ -61,6 +61,7 @@ import static com.facebook.presto.common.type.IntegerType.INTEGER;
 import static com.facebook.presto.common.type.RealType.REAL;
 import static com.facebook.presto.common.type.SmallintType.SMALLINT;
 import static com.facebook.presto.common.type.TimestampType.TIMESTAMP;
+import static com.facebook.presto.common.type.TimestampWithTimeZoneType.TIMESTAMP_WITH_TIME_ZONE;
 import static com.facebook.presto.common.type.TinyintType.TINYINT;
 import static com.facebook.presto.common.type.VarbinaryType.VARBINARY;
 import static com.facebook.presto.hive.HiveType.HIVE_BINARY;
@@ -118,6 +119,10 @@ public final class TypeConverter
             case TIME:
                 return TimeType.TIME;
             case TIMESTAMP:
+                Types.TimestampType timestampType = (Types.TimestampType) type.asPrimitiveType();
+                if (timestampType.shouldAdjustToUTC()) {
+                    return TIMESTAMP_WITH_TIME_ZONE;
+                }
                 return TimestampType.TIMESTAMP;
             case STRING:
                 return VarcharType.createUnboundedVarcharType();
--- a/presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedSmokeTestBase.java
+++ b/presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedSmokeTestBase.java
@@ -89,11 +89,11 @@ public abstract class IcebergDistributedSmokeTestBase
     @Test
     public void testTimestampWithTimeZone()
     {
-        assertQueryFails("CREATE TABLE test_timestamp_with_timezone (x timestamp with time zone)", "Iceberg column type timestamptz is not supported");
-        assertQueryFails("CREATE TABLE test_timestamp_with_timezone (x) AS SELECT TIMESTAMP '1969-12-01 00:00:00.000000 UTC'", "Iceberg column type timestamptz is not supported");
-        assertUpdate("CREATE TABLE test_timestamp_with_timezone (x timestamp)");
-        assertQueryFails("ALTER TABLE test_timestamp_with_timezone ADD COLUMN y timestamp with time zone", "Iceberg column type timestamptz is not supported");
+        assertQuerySucceeds("CREATE TABLE test_timestamp_with_timezone (x) AS SELECT TIMESTAMP '1969-12-01 00:00:00.000000 UTC'");
+        assertQuerySucceeds("ALTER TABLE test_timestamp_with_timezone ADD COLUMN y timestamp with time zone");
         dropTable(getSession(), "test_timestamp_with_timezone");
+
+        assertQueryFails("CREATE TABLE test_timestamp_with_timezone (x) WITH ( format = 'ORC') AS SELECT TIMESTAMP '1969-12-01 00:00:00.000000 UTC'", "Unsupported Type: timestamp with time zone");
     }
 
     @Test
--- a/presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergTypes.java
+++ b/presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergTypes.java
@@ -1,0 +1,121 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.iceberg;
+
+import com.facebook.presto.Session;
+import com.facebook.presto.common.type.TimestampType;
+import com.facebook.presto.common.type.TimestampWithTimeZoneType;
+import com.facebook.presto.common.type.Type;
+import com.facebook.presto.testing.MaterializedResult;
+import com.facebook.presto.testing.MaterializedRow;
+import com.facebook.presto.testing.QueryRunner;
+import com.facebook.presto.tests.AbstractTestQueryFramework;
+import com.google.common.collect.ImmutableMap;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+import java.util.List;
+
+import static com.facebook.presto.hive.HiveCommonSessionProperties.PARQUET_BATCH_READ_OPTIMIZATION_ENABLED;
+import static com.facebook.presto.iceberg.IcebergQueryRunner.createIcebergQueryRunner;
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertTrue;
+
+public class TestIcebergTypes
+        extends AbstractTestQueryFramework
+{
+    protected QueryRunner createQueryRunner() throws Exception
+    {
+        return createIcebergQueryRunner(ImmutableMap.of(), ImmutableMap.of());
+    }
+
+    @DataProvider(name = "testTimestampWithTimezone")
+    public Object[][] createTestTimestampWithTimezoneData()
+    {
+        return new Object[][] {
+                {Session.builder(getSession())
+                        .setCatalogSessionProperty("iceberg", PARQUET_BATCH_READ_OPTIMIZATION_ENABLED, "true")
+                        .build()},
+                {Session.builder(getSession())
+                        .setCatalogSessionProperty("iceberg", PARQUET_BATCH_READ_OPTIMIZATION_ENABLED, "false")
+                        .build()}
+        };
+    }
+
+    @Test(dataProvider = "testTimestampWithTimezone")
+    public void testTimestampWithTimezone(Session session)
+    {
+        QueryRunner runner = getQueryRunner();
+        String timestamptz = "TIMESTAMP '1984-12-08 00:10:00 America/Los_Angeles'";
+        String timestamp = "TIMESTAMP '1984-12-08 00:10:00'";
+
+        dropTableIfExists(runner, session.getCatalog().get(), session.getSchema().get(), "test_timestamptz");
+        assertQuerySucceeds(session, "CREATE TABLE test_timestamptz(a TIMESTAMP WITH TIME ZONE, b TIMESTAMP, c TIMESTAMP WITH TIME ZONE)");
+
+        String row = "(" + timestamptz + ", " + timestamp + ", " + timestamptz + ")";
+        for (int i = 0; i < 10; i++) {
+            assertUpdate(session, "INSERT INTO test_timestamptz values " + row, 1);
+        }
+
+        MaterializedResult initialRows = runner.execute(session, "SELECT * FROM test_timestamptz");
+
+        List<Type> types = initialRows.getTypes();
+        assertTrue(types.get(0) instanceof TimestampWithTimeZoneType);
+        assertTrue(types.get(1) instanceof TimestampType);
+
+        List<MaterializedRow> rows = initialRows.getMaterializedRows();
+        for (int i = 0; i < 10; i++) {
+            assertEquals("[1984-12-08T08:10Z[UTC], 1984-12-08T00:10, 1984-12-08T08:10Z[UTC]]", rows.get(i).toString());
+        }
+
+        dropTableIfExists(runner, session.getCatalog().get(), session.getSchema().get(), "test_timestamptz_partition");
+        assertQuerySucceeds(session, "CREATE TABLE test_timestamptz_partition(a TIMESTAMP WITH TIME ZONE, b TIMESTAMP, c TIMESTAMP WITH TIME ZONE) " +
+                "WITH (PARTITIONING = ARRAY['b'])");
+        assertUpdate(session, "INSERT INTO test_timestamptz_partition (a, b, c) SELECT a, b, c FROM test_timestamptz", 10);
+
+        MaterializedResult partitionRows = runner.execute(session, "SELECT * FROM test_timestamptz");
+
+        List<Type> partitionTypes = partitionRows.getTypes();
+        assertTrue(partitionTypes.get(0) instanceof TimestampWithTimeZoneType);
+        assertTrue(partitionTypes.get(1) instanceof TimestampType);
+
+        rows = partitionRows.getMaterializedRows();
+        for (int i = 0; i < 10; i++) {
+            assertEquals("[1984-12-08T08:10Z[UTC], 1984-12-08T00:10, 1984-12-08T08:10Z[UTC]]", rows.get(i).toString());
+        }
+
+        String earlyTimestamptz = "TIMESTAMP '1980-12-08 00:10:00 America/Los_Angeles'";
+        dropTableIfExists(runner, session.getCatalog().get(), session.getSchema().get(), "test_timestamptz_filter");
+        assertQuerySucceeds(session, "CREATE TABLE test_timestamptz_filter(a TIMESTAMP WITH TIME ZONE)");
+
+        for (int i = 0; i < 5; i++) {
+            assertUpdate(session, "INSERT INTO test_timestamptz_filter VALUES (" + earlyTimestamptz + ")", 1);
+        }
+        for (int i = 0; i < 5; i++) {
+            assertUpdate(session, "INSERT INTO test_timestamptz_filter VALUES (" + timestamptz + ")", 1);
+        }
+
+        MaterializedResult lateRows = runner.execute(session, "SELECT a FROM test_timestamptz_filter WHERE a > " + earlyTimestamptz);
+        assertEquals(lateRows.getMaterializedRows().size(), 5);
+
+        MaterializedResult lateRowsFromEquals = runner.execute(session, "SELECT a FROM test_timestamptz_filter WHERE a = " + timestamptz);
+        com.facebook.presto.testing.assertions.Assert.assertEquals(lateRows, lateRowsFromEquals);
+
+        MaterializedResult earlyRows = runner.execute(session, "SELECT a FROM test_timestamptz_filter WHERE a < " + timestamptz);
+        assertEquals(earlyRows.getMaterializedRows().size(), 5);
+
+        MaterializedResult earlyRowsFromEquals = runner.execute(session, "SELECT a FROM test_timestamptz_filter WHERE a = " + earlyTimestamptz);
+        com.facebook.presto.testing.assertions.Assert.assertEquals(earlyRows, earlyRowsFromEquals);
+    }
+}
--- a/presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/decoders/ValuesDecoder.java
+++ b/presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/decoders/ValuesDecoder.java
@@ -64,6 +64,11 @@ public interface ValuesDecoder
                 throws IOException;
     }
 
+    interface PackFunction
+    {
+        long pack(long millis);
+    }
+
     interface TimestampValuesDecoder
             extends ValuesDecoder
     {
--- a/presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/decoders/delta/Int64TimeAndTimestampMicrosDeltaBinaryPackedValuesDecoder.java
+++ b/presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/decoders/delta/Int64TimeAndTimestampMicrosDeltaBinaryPackedValuesDecoder.java
@@ -20,6 +20,8 @@ import org.openjdk.jol.info.ClassLayout;
 
 import java.io.IOException;
 
+import static com.facebook.presto.common.type.DateTimeEncoding.packDateTimeWithZone;
+import static com.facebook.presto.common.type.TimeZoneKey.UTC_KEY;
 import static java.util.concurrent.TimeUnit.MICROSECONDS;
 
 /**
@@ -34,11 +36,14 @@ public class Int64TimeAndTimestampMicrosDeltaBinaryPackedValuesDecoder
 
     private final DeltaBinaryPackingValuesReader innerReader;
 
-    public Int64TimeAndTimestampMicrosDeltaBinaryPackedValuesDecoder(int valueCount, ByteBufferInputStream bufferInputStream)
+    private final PackFunction packFunction;
+
+    public Int64TimeAndTimestampMicrosDeltaBinaryPackedValuesDecoder(int valueCount, ByteBufferInputStream bufferInputStream, boolean withTimezone)
             throws IOException
     {
         innerReader = new DeltaBinaryPackingValuesReader();
         innerReader.initFromPage(valueCount, bufferInputStream);
+        this.packFunction = withTimezone ? millis -> packDateTimeWithZone(millis, UTC_KEY) : millis -> millis;
     }
 
     @Override
@@ -46,7 +51,8 @@ public class Int64TimeAndTimestampMicrosDeltaBinaryPackedValuesDecoder
     {
         int endOffset = offset + length;
         for (int i = offset; i < endOffset; i++) {
-            values[i] = MICROSECONDS.toMillis(innerReader.readLong());
+            long curValue = MICROSECONDS.toMillis(innerReader.readLong());
+            values[i] = packFunction.pack(curValue);
         }
     }
 
--- a/presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/decoders/plain/Int64TimeAndTimestampMicrosPlainValuesDecoder.java
+++ b/presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/decoders/plain/Int64TimeAndTimestampMicrosPlainValuesDecoder.java
@@ -17,6 +17,8 @@ import com.facebook.presto.parquet.batchreader.BytesUtils;
 import com.facebook.presto.parquet.batchreader.decoders.ValuesDecoder.Int64TimeAndTimestampMicrosValuesDecoder;
 import org.openjdk.jol.info.ClassLayout;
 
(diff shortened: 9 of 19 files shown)


agrawalreetika (commented):
Can we add some tests with column types `Timestamp with timezone`?

tdcmeehan:
+1.  Let's add some end to end tests.  Additionally, we may want to remove the validation added here, since I believe we should support this properly now: #22926

hantangwangd:
Also add test cases involving `Timestamp with timezone` in filter conditions and partition columns, I'm a little concerned about the behavior in these scenarios.

auden-woolfson:
> Also add test cases involving `Timestamp with timezone` in filter conditions and partition columns, I'm a little concerned about the behavior in these scenarios.

Just to clarify, do you want the timestamp with timezone to be a part of the table that is being partitioned or the type of the partition column? Currently it is not supported as one of the types for partition columns.

hantangwangd:
> Just to clarify, do you want the timestamp with timezone to be a part of the table that is being partitioned or the type of the partition column? Currently it is not supported as one of the types for partition columns.

Yes, that's right. But I think it's better for us to first figure out how to handle it in these cases when we start to support it.

A very important question is, what format of long type data do we plan to actually store in data files for `Timestamp with timezone`? Presto has a special encoding for data with type of timestamp with timezone, which mix the time zone information with UTC values in millis. Meanwhile, Iceberg spec store the `timestamp tz` data as a UTC values in micros and do not retain the source time zone.

If we store the data following Iceberg spec, then we will lose the information of time zone; and if we store the data following Presto's format, then we may meet problems involving cross-engine compatibility.

cc: @tdcmeehan @agrawalreetika @ZacBlanco

tdcmeehan:
@hantangwangd I don't see this as a choice, we must store the data according to the Iceberg spec, which means we'll lose the embedded time zone information.  This is fine--semantically, it's the same thing, and the only thing that *might* be confusing is the user, when retrieving stored Iceberg timestamps, will see that the timezones have been adjusted to UTC.  But the point in time values will remain the same, and this is merely a limitation of the Iceberg table format.

hantangwangd:
@tdcmeehan Completely agree with your viewpoint.

That means we need to perform transformation logics for data with type of `timestamp with timezone` when writing/reading, parsing filter conditions, and handling partition values, besides doing the type conversion. It's not to say completing all these works all at once, but it can be divided into a series of PRs to complete.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergHiveMetadata.java
```diff
@@ -273,8 +272,6 @@ public ConnectorOutputTableHandle beginCreateTable(ConnectorSession session, Con
 
         Schema schema = toIcebergSchema(tableMetadata.getColumns());
 
-        verifyTypeSupported(schema);
```
ZacBlanco, line 276: What's the reason to remove this?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergNativeMetadata.java
```diff
@@ -166,8 +165,6 @@ public ConnectorOutputTableHandle beginCreateTable(ConnectorSession session, Con
 
         Schema schema = toIcebergSchema(tableMetadata.getColumns());
 
-        verifyTypeSupported(schema);
```
ZacBlanco, line 169: Same question

presto-iceberg/src/main/java/com/facebook/presto/iceberg/TypeConverter.java
```diff
@@ -117,6 +117,10 @@ public static Type toPrestoType(org.apache.iceberg.types.Type type, TypeManager
             case TIME:
                 return TimeType.TIME;
             case TIMESTAMP:
+                Types.TimestampType timestampType = (Types.TimestampType) type.asPrimitiveType();
+                if (timestampType.shouldAdjustToUTC()) {
+                    return TimestampWithTimeZoneType.TIMESTAMP_WITH_TIME_ZONE;
```
ZacBlanco, line 122: nit: static import:

presto-iceberg/src/main/java/com/facebook/presto/iceberg/TypeConverter.java
```diff
@@ -117,6 +117,10 @@ public static Type toPrestoType(org.apache.iceberg.types.Type type, TypeManager
             case TIME:
                 return TimeType.TIME;
             case TIMESTAMP:
+                Types.TimestampType timestampType = (Types.TimestampType) type.asPrimitiveType();
```
ZacBlanco, line 120: nit: static import for `TimestampType`

presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergTypes.java
```diff
@@ -0,0 +1,66 @@
...
+
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertTrue;
+
+public class TestIcebergTypes
+        extends AbstractTestQueryFramework
+{
+    @Override
+    protected QueryRunner createQueryRunner() throws Exception {
+        return IcebergQueryRunner.createIcebergQueryRunner(ImmutableMap.of(), ImmutableMap.of());
```
ZacBlanco, line 22: nit: static import on `IcebergQueryRunner`

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergNativeMetadata.java
auden-woolfson: I deleted this method, it only served to filter out the timestamp type.

ZacBlanco: Hmmm. We also currently don't support the UUID type. I guess it fails in a different way then

auden-woolfson: I think UUID is failing in the type converter default case

agrawalreetika (changes requested):
Overall looks good to me.

1. Please add a document entry in https://prestodb.io/docs/current/connector/iceberg.html#type-mapping
2. Squash all the commits into 1 "Fix timestamp with timezone mapping in iceberg type converter"

presto-iceberg/src/main/java/com/facebook/presto/iceberg/TypeConverter.java
agrawalreetika, line 122: Add static import for this

agrawalreetika (changes requested):

presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergTypes.java
```diff
@@ -0,0 +1,123 @@
...
+{
+    private QueryRunner batchReaderEnabledQueryRunner;
+
+    @Override
+    protected QueryRunner createQueryRunner() throws Exception
+    {
+        this.batchReaderEnabledQueryRunner = createIcebergQueryRunner(
+                ImmutableMap.of(),
+                ImmutableMap.of(),
+                ImmutableMap.of(PARQUET_BATCH_READ_OPTIMIZATION_ENABLED, "true"),
```
hantangwangd, line 47: I think there is no need to refactor the `IcebergQueryRunner.createIcebergQueryRunner(...)`, just set `hive.parquet-batch-read-optimization-enabled` to `true` in `extraProperties` would be ok.

presto-parquet/src/main/java/com/facebook/presto/parquet/reader/LongTimestampMicrosColumnReader.java
```diff
@@ -36,12 +33,7 @@ protected void readValue(BlockBuilder blockBuilder, Type type)
         if (definitionLevel == columnDescriptor.getMaxDefinitionLevel()) {
             long utcMillis = MICROSECONDS.toMillis(valuesReader.readLong());
             // TODO: specialize the class at creation time
-            if (type instanceof TimestampWithTimeZoneType) {
-                type.writeLong(blockBuilder, packDateTimeWithZone(utcMillis, UTC_KEY));
-            }
-            else {
-                type.writeLong(blockBuilder, utcMillis);
-            }
+            type.writeLong(blockBuilder, utcMillis);
```
hantangwangd, line 36: The real long type data for timestamp with tz stored in parquet file is of presto's format? This may lead to many problems. As guided by @tdcmeehan, we should store the data in Iceberg format.

hantangwangd (changes requested):

presto-parquet/src/main/java/com/facebook/presto/parquet/reader/LongTimestampMicrosColumnReader.java
ZacBlanco: We may need to modify the writer as well for this, but I need to verify. I noticed that we were improperly packing the date in the previous logic which led to incorrect values being read.

For example the following set of queries produced bad results:

```
create table t(t TIMESTAMP WITH TIME ZONE);

INSERT INTO t VALUES TIMESTAMP '1980-12-08 0:10:0 America/Los_Angeles';

presto:tpch> SELECT * FROM t;
               t
-------------------------------
 +46764-05-25 18:40:01.825 UTC
```

hantangwangd: > We may need to modify the writer as well for this.

Agree, the bad results in your example seems to be the problem caused by writing data with the incorrect format. So we need to customize a dedicated writer logic for timestamp with tz, which transform the long values encoded in presto's format to long values encoded in iceberg's format.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergTypes.java
auden-woolfson: Maybe there is another way, but adding this to extra properties actually causes the tests to break since it is an unused property there. This sets it as a configuration property, but it needs to be a session property. The session is actually passed to the distributed query runner builder in it's constructor, so we need some way to add properties to that session before building the runner. This is why I decided to make changes to `IcebergQueryRunner`. Please let me know if this clears things up, if you have another approach in mind here I would certainly be open to it!

presto-parquet/src/main/java/com/facebook/presto/parquet/reader/LongTimestampMicrosColumnReader.java
auden-woolfson: I made some changes that I thought would fix this, added the `unpackMillisUtc` method to a new `TimestampWithTimezoneValueWriter` that I added to parquet, but we are losing all timezone information in the unpacking bit shift. This theoretically shouldn't be a problem since we  are supposed to be storing `millisUtc` in the first section of the bits, and the last 12 are reserved for whatever timezone we want to display to the user. However this is clearly not the case, since the filter operations are failing, the program expects the millis UTC to be transformed based on the timezone info, to get an _actual_ millis UTC. 

This is some pretty misleading variable naming and I think it leaves us with two options...

1. leave the system the same and build methods on top of it that apply the timezone part of the ts with tz to the millis part to get a correct instant (and change the variable names)
2. change the way ts with tz is read so at the time of unpacking we already have the millis UTC correct.

Sorry if this is redundant, I couldn't find documentation about how iceberg stores ts w tz under the hood.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergTypes.java
hantangwangd: Ooh, sorry for the overlooking that `hive.parquet-batch-read-optimization-enabled` belongs to the config in `presto-hive-common`. Since `presto-main` does not dependent on `presto-hive-common`, `TestingPrestoServer` does not inject `HiveCommonModule` either. So you are right currently we cannot set `hive.parquet-batch-read-optimization-enabled` in `extraProperties` here.

presto-parquet/src/main/java/com/facebook/presto/parquet/reader/LongTimestampMicrosColumnReader.java
hantangwangd: Referring to https://iceberg.apache.org/spec/#file-system-operations, Iceberg defines timestamp tz as follows:

```
Timestamp, microsecond precision, with timezone

Timestamp values with time zone represent a point in time: values are stored as UTC and do not retain a source time zone (2017-11-16 17:10:34 PST is stored/retrieved as 2017-11-17 01:10:34 UTC and these values are considered identical).
```

So it just contains UTC values in micros and do not retain the source time zone.

By the way, did the filtering operation fail because we forgot to do the same convert for timestamp with tz in `ExpressionConverter.getIcebergLiteralValue(...)`?

auden-woolfson: Not sure about the filtering operation. I had to add `TimestampWithTimezoneType` to the cases in expression converter but so far I have just treated it like the other date time types that are included. I tried adding unpacking to the expression converter for ts w tz and now we have the regular tests passing, but failing with the batch reader enabled. Looking into this now

presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/decoders/ValuesDecoder.java
```diff
@@ -60,6 +60,9 @@ interface Int64TimeAndTimestampMicrosValuesDecoder
         void readNext(long[] values, int offset, int length)
                 throws IOException;
 
+        void readNextWithTimezone(long[] values, int offset, int length)
+                throws IOException;
+
```
auden-woolfson, line 65: This is necessary because the decoder is accesses the individual raw values from parquet, not the batch reader. The batch reader has the column descriptor (metadata) which should tell it whether or not to pack the value with timezone, like is done in the regular column reader. The actual packing should be done within the decoder, so there needs to be a way for decoders to dynamically switch between with and without timezone mode.

The approach I took here is just to copy the readNext method from each implementation and add `packDateTimewithZone`. A stateful approach (instance variable bool withTimestamp or something) would probably be more efficient for future development and code execution but for now I am seeing if this implementation works.

presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/decoders/rle/Int64TimeAndTimestampMicrosRLEDictionaryValuesDecoder.java
```diff
@@ -69,7 +91,13 @@ public void readNext(long[] values, int offset, int length)
                     final LongDictionary localDictionary = dictionary;
                     for (int srcIndex = currentBuffer.length - currentCount; destinationIndex < endIndex; srcIndex++) {
                         long dictionaryValue = localDictionary.decodeToLong(localBuffer[srcIndex]);
-                        values[destinationIndex++] = MICROSECONDS.toMillis(dictionaryValue);
+                        long millisValue = MICROSECONDS.toMillis(dictionaryValue);
+                        if (isWithTimezone()) {
```
ZacBlanco, line 95: Rather than having a branch statement inside the hot loop for a lot of these readers, I think we should set the "reading function" in the constructor of the reader as the reading behavior shouldn't change. Do you see a performance impact when this conditional is introduced?

presto-parquet/src/main/java/com/facebook/presto/parquet/writer/valuewriter/TimestampWithTimezoneValueWriter.java
```diff
@@ -0,0 +1,51 @@
...
+        this.writeMicroseconds = parquetType.isPrimitive() && parquetType.getOriginalType() == OriginalType.TIMESTAMP_MICROS;
+    }
+
+    @Override
+    public void write(Block block)
+    {
+        for (int i = 0; i < block.getPositionCount(); i++) {
+            if (!block.isNull(i)) {
+                long value = unpackMillisUtc(type.getLong(block, i));
+                long scaledValue = writeMicroseconds ? MILLISECONDS.toMicros(value) : value;
```
ZacBlanco, line 45: same comment with avoiding branching inside hot loop

ZacBlanco (changes requested):

presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergTypes.java
```diff
@@ -0,0 +1,123 @@
...
+                Optional.empty(),
+                false);
+        return createIcebergQueryRunner(ImmutableMap.of(), ImmutableMap.of());
+    }
+
+    @DataProvider(name = "testTimestampWithTimezone")
+    public Object[][] createTestTimestampWithTimezoneData()
+    {
+        return new Object[][] {
+                {getQueryRunner()},
```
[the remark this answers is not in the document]
auden-woolfson: Can you elaborate on this please? I'm not sure what you are referring to as the separate test. I can have the data provider pass in one true and one false value and add a condition inside the test function itself, is that what you are asking for here? If so what purpose would that serve? Thanks

steveburnett:
>     1. Please add a document entry in https://prestodb.io/docs/current/connector/iceberg.html#type-mapping

Please let me know when this or other doc is added to the PR, and I'll be happy to review the documentation.

steveburnett (dismissed):
LGTM! (docs)

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -1757,6 +1757,10 @@ Map of Iceberg types to the relevant PrestoDB types:
     - ``TIME``
   * - ``TIMESTAMP``
     - ``TIMESTAMP``
+  * - ``TIMESTAMP``
+    - ``TIMESTAMP_WITH_TIMEZONE``
+  * - ``STRING``
+    - ``VARCHAR``
```
ZacBlanco, line 1763: I think we also need the mapping for the table below (prestodb -> iceberg)?

presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/decoders/delta/Int64TimeAndTimestampMicrosDeltaBinaryPackedValuesDecoder.java
```diff
@@ -34,19 +36,43 @@ public class Int64TimeAndTimestampMicrosDeltaBinaryPackedValuesDecoder
...
     }
 
     @Override
     public void readNext(long[] values, int offset, int length)
     {
         int endOffset = offset + length;
         for (int i = offset; i < endOffset; i++) {
-            values[i] = MICROSECONDS.toMillis(innerReader.readLong());
+            long curValue = MICROSECONDS.toMillis(innerReader.readLong());
+            packFunction.pack(curValue);
```
ZacBlanco, line 75: don't we need to use the return value of the `pack` function?

presto-docs/src/main/sphinx/connector/iceberg.rst
auden-woolfson: Actually I'm not sure about this. The below table contains...

```
  * - ``TIMESTAMP``
    - ``TIMESTAMP WITHOUT ZONE``
  * - ``TIMESTAMP WITH TIMEZONE``
    - ``TIMESTAMP WITH ZONE``
```

I suppose with and without zone just represent `isAdjustedUTC` in this case? Either way it looks like this part is covered. Maybe we should standardize the way we discuss these in the documentation though

auden-woolfson:
Looking into test failures

auden-woolfson:
All presto spark base UTs are passing locally, rebasing on master and trying again

presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergTypes.java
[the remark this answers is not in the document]
ZacBlanco: Creating an entirely new query runner is expensive. We should be able to test with the batch read optimization by just setting a session property.

In this data provider you should can create two sessions using the default session. One with batch read enabled and disabled and then run the test. It will take far less time to run. You will need to update all of the query statements to pass the session variable though.


```java
public Object[][] batchReadSessions()
    {
        return new Object[][] {
                {Session.builder(getSession())
                        .setCatalogSessionProperty("iceberg", PARQUET_BATCH_READ_OPTIMIZATION_ENABLED, "true")
                        .build()},
                {Session.builder(getSession())
                        .setCatalogSessionProperty("iceberg", PARQUET_BATCH_READ_OPTIMIZATION_ENABLED, "false")
                        .build()}
        };
    }
```

This would also allow you to remove the additional logic inside of `createQueryRunner`

presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergTypes.java
```diff
@@ -0,0 +1,124 @@
...
+        runner.execute("INSERT INTO test_timestamptz_partition (a, b, c) SELECT a, b, c FROM test_timestamptz");
+
+        MaterializedResult partitionRows = runner.execute("SELECT * FROM test_timestamptz");
+        List<Type> partitionTypes = partitionRows.getTypes();
+
+        assertTrue(partitionTypes.get(0) instanceof TimestampWithTimeZoneType);
+        assertTrue(partitionTypes.get(1) instanceof TimestampType);
+
+        String earlyTimestamptz = "TIMESTAMP '1980-12-08 00:10:00 America/Los_Angeles'";
+        runner.execute("CREATE TABLE test_timestamptz_filter(a TIMESTAMP WITH TIME ZONE)");
```
ZacBlanco, line 97: assertQuerySucceeds

presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergTypes.java
```diff
@@ -0,0 +1,124 @@
...
+
+        for (int i = 0; i < 5; i++) {
+            runner.execute("INSERT INTO test_timestamptz_filter VALUES (" + earlyTimestamptz + ")");
+        }
+        for (int i = 0; i < 5; i++) {
+            runner.execute("INSERT INTO test_timestamptz_filter VALUES (" + timestamptz + ")");
```
ZacBlanco, line 103: assertUpdate

auden-woolfson:
@ZacBlanco thanks, just pushed the changes

ZacBlanco (changes requested):

presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergTypes.java
```diff
@@ -0,0 +1,139 @@
...
+                ImmutableMap.of(),
+                ImmutableMap.of(PARQUET_BATCH_READ_OPTIMIZATION_ENABLED, "true"),
+                new IcebergConfig().getFileFormat(),
+                true,
+                false,
+                OptionalInt.empty(),
+                Optional.empty(),
+                Optional.empty(),
+                false,
+                Optional.empty());
```
ZacBlanco, line 57: This is also unused

presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/decoders/delta/Int64TimeAndTimestampMicrosDeltaBinaryPackedValuesDecoder.java
```diff
@@ -34,19 +36,31 @@ public class Int64TimeAndTimestampMicrosDeltaBinaryPackedValuesDecoder
...
     {
         innerReader = new DeltaBinaryPackingValuesReader();
         innerReader.initFromPage(valueCount, bufferInputStream);
+        this.withTimezone = withTimezone;
+        if (withTimezone) {
+            this.packFunction = millis -> packDateTimeWithZone(millis, UTC_KEY);
+        }
+        else {
+            this.packFunction = millis -> millis;
+        }
```
agrawalreetika, line 54:
```suggestion
this.packFunction = withTimezone
                ? millis -> packDateTimeWithZone(millis, UTC_KEY)
                : millis -> millis;
```

presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergTypes.java
```diff
@@ -0,0 +1,121 @@
...
+    @DataProvider(name = "testTimestampWithTimezone")
+    public Object[][] createTestTimestampWithTimezoneData()
+    {
+        return new Object[][] {
+                {Session.builder(getSession())
+                        .setCatalogSessionProperty("iceberg", PARQUET_BATCH_READ_OPTIMIZATION_ENABLED, "true")
+                        .build()},
+                {Session.builder(getSession())
+                        .setCatalogSessionProperty("iceberg", PARQUET_BATCH_READ_OPTIMIZATION_ENABLED, "false")
+                        .build()}
```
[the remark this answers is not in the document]
auden-woolfson: This is just to test with batch reader enabled and disabled. We are not changing the data used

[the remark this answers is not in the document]
agrawalreetika: Then I think name of the DataProvider and Object method according to the properties would be better?

presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/decoders/plain/Int64TimeAndTimestampMicrosPlainValuesDecoder.java
```diff
@@ -31,11 +33,27 @@ public class Int64TimeAndTimestampMicrosPlainValuesDecoder
...
         this.byteBuffer = byteBuffer;
         this.bufferOffset = bufferOffset;
         this.bufferEnd = bufferOffset + length;
+        this.withTimezone = withTimezone;
+        if (withTimezone) {
+            packFunction = millis -> packDateTimeWithZone(millis, UTC_KEY);
+        }
+        else {
+            packFunction = millis -> millis;
+        }
```
agrawalreetika, line 56:
```suggestion
this.packFunction = withTimezone ? millis -> packDateTimeWithZone(millis, UTC_KEY) : millis -> millis;
```

presto-docs/src/main/sphinx/connector/iceberg.rst
hantangwangd: nit: Seems `STRING -> VARCHAR` is duplicate with above.

presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/Int64TimeAndTimestampMicrosNestedBatchReader.java
```diff
@@ -16,9 +16,11 @@
 import com.facebook.presto.common.block.Block;
 import com.facebook.presto.common.block.LongArrayBlock;
 import com.facebook.presto.common.block.RunLengthEncodedBlock;
+import com.facebook.presto.common.type.TimestampWithTimeZoneType;
 import com.facebook.presto.parquet.RichColumnDescriptor;
 import com.facebook.presto.parquet.batchreader.decoders.ValuesDecoder.Int64TimeAndTimestampMicrosValuesDecoder;
 import com.facebook.presto.parquet.reader.ColumnChunk;
+import org.apache.parquet.schema.LogicalTypeAnnotation;
```
hantangwangd, line 23: Why add these two import?

presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/decoders/ValuesDecoder.java
```diff
@@ -62,6 +62,8 @@ void readNext(long[] values, int offset, int length)
 
         void skip(int length)
                 throws IOException;
+
+        boolean getWithTimezone();
```
hantangwangd, line 66: Is this method useful?

presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/decoders/delta/Int64TimeAndTimestampMicrosDeltaBinaryPackedValuesDecoder.java
```diff
@@ -34,19 +36,26 @@ public class Int64TimeAndTimestampMicrosDeltaBinaryPackedValuesDecoder
 
     private final DeltaBinaryPackingValuesReader innerReader;
 
-    public Int64TimeAndTimestampMicrosDeltaBinaryPackedValuesDecoder(int valueCount, ByteBufferInputStream bufferInputStream)
+    private final PackFunction packFunction;
+
+    private final boolean withTimezone;
```
hantangwangd, line 41: Do we really need this field? As I understand, `packFunction` has already contains this information.

presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/decoders/delta/Int64TimeAndTimestampMicrosDeltaBinaryPackedValuesDecoder.java
```diff
@@ -65,4 +74,15 @@ public long getRetainedSizeInBytes()
...
+    @Override
+    public boolean getWithTimezone()
+    {
+        return withTimezone;
+    }
+
+    private interface PackFunction
+    {
+        long pack(long millis);
+    }
```
hantangwangd, line 87: Should we extract interface `PackFunction` into some place like `ValuesDecoder` and reuse it? Since currently it has been defined and used in several places.

presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/decoders/plain/Int64TimeAndTimestampMicrosPlainValuesDecoder.java
```diff
@@ -31,11 +33,27 @@ public class Int64TimeAndTimestampMicrosPlainValuesDecoder
 
     private int bufferOffset;
 
+    private final boolean withTimezone;
```
hantangwangd, line 36: The same as above.

presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/decoders/plain/Int64TimeAndTimestampMicrosPlainValuesDecoder.java
```diff
@@ -69,4 +87,15 @@ public long getRetainedSizeInBytes()
...
+    @Override
+    public boolean getWithTimezone()
+    {
+        return withTimezone;
+    }
+
+    private interface PackFunction
+    {
+        long pack(long millis);
+    }
```
hantangwangd, line 100: The same as above.

presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/decoders/rle/Int64TimeAndTimestampMicrosRLEDictionaryValuesDecoder.java
```diff
@@ -34,10 +36,21 @@ public class Int64TimeAndTimestampMicrosRLEDictionaryValuesDecoder
 
     private final LongDictionary dictionary;
 
-    public Int64TimeAndTimestampMicrosRLEDictionaryValuesDecoder(int bitWidth, InputStream inputStream, LongDictionary dictionary)
+    private final boolean withTimezone;
```
hantangwangd, line 39: The same as above.

presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/decoders/rle/Int64TimeAndTimestampMicrosRLEDictionaryValuesDecoder.java
```diff
@@ -110,4 +124,15 @@ public long getRetainedSizeInBytes()
...
+    @Override
+    public boolean getWithTimezone()
+    {
+        return withTimezone;
+    }
+
+    private interface PackFunction
+    {
+        long pack(long millis);
+    }
```
hantangwangd, line 137: The same as above.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedSmokeTestBase.java
```diff
@@ -89,10 +89,8 @@ public void testTimestamp()
     @Test
     public void testTimestampWithTimeZone()
     {
-        assertQueryFails("CREATE TABLE test_timestamp_with_timezone (x timestamp with time zone)", "Iceberg column type timestamptz is not supported");
-        assertQueryFails("CREATE TABLE test_timestamp_with_timezone (x) AS SELECT TIMESTAMP '1969-12-01 00:00:00.000000 UTC'", "Iceberg column type timestamptz is not supported");
-        assertUpdate("CREATE TABLE test_timestamp_with_timezone (x timestamp)");
-        assertQueryFails("ALTER TABLE test_timestamp_with_timezone ADD COLUMN y timestamp with time zone", "Iceberg column type timestamptz is not supported");
+        assertQuerySucceeds("CREATE TABLE test_timestamp_with_timezone (x) AS SELECT TIMESTAMP '1969-12-01 00:00:00.000000 UTC'");
+        assertQuerySucceeds("ALTER TABLE test_timestamp_with_timezone ADD COLUMN y timestamp with time zone");
```
hantangwangd, line 93: Should we add a test case to indicate that timestamp with time zone on ORC is not currently supported?

hantangwangd (changes requested):
Thanks for the change, overall looks good to me. Some little nits.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergQueryRunner.java
```diff
@@ -146,7 +147,7 @@ public static DistributedQueryRunner createIcebergQueryRunner(
             Optional<Path> dataDirectory)
             throws Exception
     {
-        return createIcebergQueryRunner(extraProperties, extraConnectorProperties, format, createTpchTables, addJmxPlugin, nodeCount, externalWorkerLauncher, dataDirectory, false, Optional.empty());
+        return createIcebergQueryRunner(extraProperties, extraConnectorProperties, format, createTpchTables, addJmxPlugin, nodeCount, externalWorkerLauncher, dataDirectory, false);
```
ZacBlanco, line 150: Can we revert this change? I don't see how it's used

presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergQueryRunner.java
```diff
@@ -161,12 +162,13 @@ public static DistributedQueryRunner createIcebergQueryRunner(
             boolean addStorageFormatToPath)
             throws Exception
     {
-        return createIcebergQueryRunner(extraProperties, extraConnectorProperties, format, createTpchTables, addJmxPlugin, nodeCount, externalWorkerLauncher, dataDirectory, addStorageFormatToPath, Optional.empty());
+        return createIcebergQueryRunner(extraProperties, extraConnectorProperties, ImmutableMap.of(), format, createTpchTables, addJmxPlugin, nodeCount, externalWorkerLauncher, dataDirectory, addStorageFormatToPath, Optional.empty());
```
ZacBlanco, line 165: Same here. Is there any reason to change this?

ZacBlanco (commented):
Just some minor issues, otherwise lgtm

presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergQueryRunner.java
auden-woolfson: This is needed because we have to be able to pass extraSessionProperties. Since the other signatures are still used in the code it is logical to just have them pass an empty map to the method that actually contains the logic.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergQueryRunner.java
auden-woolfson: This would require adding another instance method which would only be used within the class.

ZacBlanco (changes requested):

ZacBlanco (approved):

agrawalreetika (approved):

hantangwangd (approved):

tdcmeehan (approved):

tdcmeehan:
Thank you for the reviews @ZacBlanco @agrawalreetika @hantangwangd, and thanks @auden-woolfson for the fix.

PR #23541, opened 2024-08-28
Rephrase session property description

kewang1024:
```
== NO RELEASE NOTE ==
```


presto-docs/src/main/sphinx/presto_cpp/properties.rst
```diff
@@ -148,9 +148,9 @@ The configuration properties of Presto C++ workers are described here, in alphab
 * **Type:** ``integer``
 * **Default value:** ``100UL << 30``
 
-  Specifies the max spill bytes limit set for each query. This is used to cap the
-  storage used for spilling. If it is zero, then there is no limit and spilling
-  might exhaust the storage or takes too long to run.
+  Specifies the max spill bytes limit set for each query on a Prestissimo worker.
+  This is used to cap the storage used for spilling. If it is zero, then there is
+  no limit and spilling might exhaust the storage or takes too long to run.
```
elharo, line 153: takes --> take

presto-docs/src/main/sphinx/presto_cpp/properties.rst
```diff
@@ -148,9 +148,9 @@ The configuration properties of Presto C++ workers are described here, in alphab
 * **Type:** ``integer``
 * **Default value:** ``100UL << 30``
 
-  Specifies the max spill bytes limit set for each query. This is used to cap the
-  storage used for spilling. If it is zero, then there is no limit and spilling
-  might exhaust the storage or takes too long to run.
+  Specifies the max spill bytes limit set for each query on a Prestissimo worker.
```
steveburnett, line 151:
```suggestion
  Specifies the max spill bytes limit set for each query on a Presto C++ worker.
```

presto-docs/src/main/sphinx/presto_cpp/properties.rst
```diff
@@ -148,9 +148,9 @@ The configuration properties of Presto C++ workers are described here, in alphab
...
+  This is used to cap the storage used for spilling. If it is zero, then there is
```
steveburnett, line 152:
```suggestion
  Use this session property to cap the storage used for spilling. If this is set to zero, then there is
```

presto-docs/src/main/sphinx/presto_cpp/properties.rst
steveburnett, line 153:
```suggestion
  no limit and spilling might exhaust the storage or take too long to run.
```

steveburnett (changes requested):
Thanks for the doc! A few nits of phrasing suggestions.

presto-docs/src/main/sphinx/presto_cpp/properties.rst
kewang1024: I changed it to presto native to be consistent

steveburnett: I understand that, and I'd suggested Presto C++ for consistency with the existing documentation: the [Presto C++](https://prestodb.io/docs/current/presto-cpp.html) page in the Presto doc has "Presto C++, sometimes referred to by the development name Prestissimo,...". 

We don't mention "Presto native" in the docs as a synonym for Prestissimo and I want to avoid introducing another label into the docs for the same concept.  If you object to Presto C++, I'd prefer it if you could change it back to Prestissimo instead of Presto native. Does that make sense?

xiaoxmeng (dismissed):
@kewang1024 thanks for the fix.

presto-docs/src/main/sphinx/presto_cpp/properties.rst
xiaoxmeng: yeah, Prestissimo worker seems better and be consistent with other places. Thanks!

kewang1024: Changed to "prestissimo worker"

tdcmeehan: Folks, speaking generally, this is user facing documentation.  For user facing documentation, please use **Presto C++**, and for development, please use **Prestissimo**.

This is the terminology we are already using for user facing documentation on the docs website.  If you would like to use **Prestissimo** for user facing documentation, please open an issue and we can discuss, but let's not start calling this two different things.

Moreover, I have a **practical suggestion for this PR**.  Just remove any reference to "Prestissimo worker" or "Presto C++ worker" and just call it a **worker**, as it is implicit that we are talking about C++ workers.

tdcmeehan (changes requested):

presto-docs/src/main/sphinx/presto_cpp/properties.rst
steveburnett, line 151:
```suggestion
  Specifies the max spill bytes limit set for each query on a worker.
```
New suggestion following @tdcmeehan's suggestion in [his comment](https://github.com/prestodb/presto/pull/23541/files#r1738639670).

steveburnett (changes requested):

PR #23545, opened 2024-08-29
Phase 1. Implementation for RFC-0005: Scalar function stats propagation.

ScrapCodes:
## Description
1. Support for annotating functions with both constant stats and propagating source stats.
2. Added tests for the same.
3. Added Scalar stats calculation based on annotation and tests for the same.

Not added SQLInvokedScalarFunctions.
Not annotated builtin functions, as that is covered in next implementation phase. Not added C++ changes as this phase only covers Java side of changes.

## Motivation and Context
https://github.com/prestodb/rfcs/blob/main/RFC-0005-functions-stats.md

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* ... :pr:`12345`
* ... :pr:`12345`

Hive Connector Changes
* ... :pr:`12345`
* ... :pr:`12345`
```

```
== NO RELEASE NOTE ==
```


steveburnett:
* Suggest including documentation for the new configuration property and session property, following the [Review and Commit Guidelines](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#documentation).

* Suggest revision to the release note entry following the [Release Notes Guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines): 

```
== RELEASE NOTES ==

General Changes
* Add configuration property ``optimizer.scalar-function-stats-propagation-enabled`` and session property ``scalar_function_stats_propagation_enabled`` to enable stats propagation by annotation, supporting `RFC-5 <https://github.com/prestodb/rfcs/blob/main/RFC-0005-functions-stats.md>`_. :pr: `23545`
```

ScrapCodes:
Thank you for the comment @steveburnett , do you have a suggestion on documenting this feature in full detail e.g. developer docs.

steveburnett:
> Thank you for the comment @steveburnett , do you have a suggestion on documenting this feature in full detail e.g. developer docs.

Yes, thanks! If these properties are relevant to Presto, please add entries for them to https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/admin/properties.rst. 

If these are Presto C++ (Prestissimo), please add the entries to https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/presto_cpp/features.rst#session-properties and https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/presto_cpp/properties.rst as appropriate.

presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsCalculator.java
```diff
@@ -246,16 +627,41 @@ private VariableStatsEstimate computeNegationStatistics(CallExpression call, Voi
...
+        }
+
+        private VariableStatsEstimate computeComparisonOperatorStatistics(CallExpression call, Void context)
+        {
+            requireNonNull(call, "call is null");
+            VariableStatsEstimate left = getSourceStats(call, context, 0);
+            VariableStatsEstimate right = getSourceStats(call, context, 1);
+            VariableStatsEstimate.Builder result =
+                    VariableStatsEstimate.builder()
+                            .setAverageRowSize(1.0)
```
ZacBlanco, line 649: I would use the width of the boolean type here rather than hard code a constant

presto-main/src/main/java/com/facebook/presto/operator/scalar/ScalarHeader.java
```diff
@@ -23,13 +27,32 @@ public class ScalarHeader
...
+        this.visibility = visibility;
+        this.deterministic = deterministic;
+        this.calledOnNullInput = calledOnNullInput;
+        this.scalarStatsHeader = scalarStatsHeader;
+    }
+
+    @Override
+    public String toString()
+    {
+        return String.format("ScalarHeader: description: %s visibility:%s deterministic: %b calledOnNullInput %b scalarStatsHeader %s",
```
ZacBlanco, line 54: Consider using guava's `ToStringHelper` here

presto-main/src/main/java/com/facebook/presto/operator/scalar/annotations/ScalarFromAnnotationsParser.java
```diff
@@ -99,11 +106,31 @@ private static List<ScalarHeaderAndMethods> findScalarsInFunctionSetClass(Class<
         return builder.build();
     }
 
+    private static Optional<ScalarStatsHeader> getScalarStatsHeader(Method annotated)
+    {
+        Optional<ScalarStatsHeader> scalarStatsHeader;
+        ScalarFunctionConstantStats statsCalculator = annotated.getAnnotation(ScalarFunctionConstantStats.class);
+        List<java.lang.reflect.Parameter> params = Arrays.stream(annotated.getParameters()).filter(x -> x.getAnnotation(SqlType.class) != null).collect(Collectors.toList());
```
ZacBlanco, line 113: Try to use full variable names instead of `x`

presto-main/src/main/java/com/facebook/presto/operator/scalar/annotations/ScalarFromAnnotationsParser.java
```diff
@@ -99,11 +106,31 @@ private static List<ScalarHeaderAndMethods> findScalarsInFunctionSetClass(Class<
...
+        // Map of (function argument position index) -> (ScalarPropagateSourceStats annotation)
+        Map<Integer, ScalarPropagateSourceStats> argumentStatsResolver = new HashMap<>();
+
+        IntStream.range(0, params.size())
+                .filter(x -> params.get(x).getAnnotation(ScalarPropagateSourceStats.class) != null)
```
ZacBlanco, line 118: same comment on variable name. There are a few other cases in this class. Please fix them as well

presto-spi/src/main/java/com/facebook/presto/spi/function/Signature.java
```diff
@@ -168,6 +169,15 @@ public String toString()
                 "(" + String.join(",", argumentTypes.stream().map(TypeSignature::toString).collect(toList())) + "):" + returnType;
     }
 
+    /*
+     * Canonical (normalized i.e. erased type size bounds) form of signature instance.
+     */
+    public Signature canonicalization()
+    {
+        return new Signature(this.name, this.kind, new TypeSignature(this.returnType.getBase(), Collections.emptyList()),
+                argumentTypes.stream().map(x -> new TypeSignature(x.getBase(), Collections.emptyList())).collect(toList()));
```
ZacBlanco, line 178: nit: argument name for `x ->`

presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsCalculator.java
```diff
@@ -136,13 +153,53 @@ public VariableStatsEstimate visitCall(CallExpression call, Void context)
...
+            if (isStatsPropagationEnabled) {
+                if (functionMetadata.getOperatorType().map(OperatorType::isHashOperator).orElse(false)) {
+                    return computeHashCodeOperatorStatistics(call, context);
+                }
+
+                if (functionMetadata.getOperatorType().map(OperatorType::isComparisonOperator).orElse(false)) {
+                    return computeComparisonOperatorStatistics(call, context);
+                }
+
+                if (call.getDisplayName().equals("concat")) {
```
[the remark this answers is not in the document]
ScrapCodes: concat function is a generated (byte code) function for all arities. It is possible to add/generate annotation as well, but this seemed easier/cleaner to implement. What do you think?

presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsCalculator.java
```diff
@@ -136,13 +153,53 @@ public VariableStatsEstimate visitCall(CallExpression call, Void context)
...
             }
+
+            return computeStatsViaAnnotations(call, context, functionMetadata);
+        }
+
+        private VariableStatsEstimate computeStatsViaAnnotations(CallExpression call, Void context, FunctionMetadata functionMetadata)
+        {
+            // casting session to FullConnectorSession is not ideal.
+            boolean isStatsPropagationEnabled =
+                    SystemSessionProperties.shouldEnableScalarFunctionStatsPropagation(((FullConnectorSession) session).getSession());
```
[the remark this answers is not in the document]
ScrapCodes: This turned out to be larger undertaking, may be it can be a separate PR?

presto-docs/src/main/sphinx/admin/properties.rst
```diff
@@ -854,6 +854,18 @@ This can also be specified on a per-query basis using the ``confidence_based_bro
...
 on a per-query basis using the ``treat-low-confidence-zero-estimation-as-unknown`` session property.
 
+``optimizer.scalar-function-stats-propagation-enabled``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+
+* **Type:** ``boolean``
+* **Default value:** ``false``
+
+Enable scalar functions stats propagation via annotations. Annotations `@ScalarFunctionConstantStats` and
```
steveburnett, line 864:
```suggestion
Enable scalar functions stats propagation via annotations. Annotations ``@ScalarFunctionConstantStats`` and
```

presto-docs/src/main/sphinx/admin/properties.rst
```diff
@@ -854,6 +854,18 @@ This can also be specified on a per-query basis using the ``confidence_based_bro
...
+`@ScalarPropagateSourceStats` will define the behavior of the scalar function stats characteristics. This flag or
```
steveburnett, line 865:
```suggestion
``@ScalarPropagateSourceStats`` define the behavior of the scalar function stats characteristics. 
```

presto-docs/src/main/sphinx/admin/properties.rst
```diff
@@ -854,6 +854,18 @@ This can also be specified on a per-query basis using the ``confidence_based_bro
...
+the session property `scalar_function_stats_propagation_enabled` when set to true will enable the stats
```
steveburnett, line 866:
```suggestion
When set to true, this property enables the stats
```

presto-docs/src/main/sphinx/admin/properties.rst
```diff
@@ -854,6 +854,18 @@ This can also be specified on a per-query basis using the ``confidence_based_bro
...
+propagation via annotations.
```
steveburnett, line 867:
```suggestion
propagation through annotations. This can also be specified on a per-query basis using the ``scalar_function_stats_propagation_enabled`` session property.
```
Avoid using Latin terms and expressions such as via. For discussion and alternate suggestions see the Gitlab documentation style guide [recommended word list entry for via](https://docs.gitlab.com/ee/development/documentation/styleguide/word_list.html#via). 

Also revised the mention of the session property for consistency with the rest of the page.

steveburnett (changes requested):
Thank you for adding the documentation! A few nits of formatting, and a couple of revisions for consistency with the existing documentation.

presto-docs/src/main/sphinx/admin/properties.rst
```diff
@@ -854,6 +854,18 @@ This can also be specified on a per-query basis using the ``confidence_based_bro
...
 on a per-query basis using the ``treat-low-confidence-zero-estimation-as-unknown`` session property.
 
+``optimizer.scalar-function-stats-propagation-enabled``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+
+* **Type:** ``boolean``
+* **Default value:** ``false``
+
+Enable scalar functions stats propagation via annotations. Annotations ``@ScalarFunctionConstantStats`` and
```
steveburnett, line 864:
```suggestion
Enable scalar functions stats propagation using annotations. Annotations ``@ScalarFunctionConstantStats`` and
```

presto-docs/src/main/sphinx/admin/properties.rst
```diff
@@ -854,6 +854,18 @@ This can also be specified on a per-query basis using the ``confidence_based_bro
...
+``@ScalarPropagateSourceStats`` define the behavior of the scalar function stats characteristics. 
+When set to true, this property enables the stats
```
steveburnett, line 866:
```suggestion
When set to ``true``, this property enables the stats
```

steveburnett (changes requested):
Thanks for the quick response! Two final nits that I should have caught in my first pass, and I think this should be everything.

steveburnett (approved):
LGTM! (docs)

Pull updated branch, reviewed new local doc build, looks good. Thanks!

presto-docs/src/main/sphinx/admin/properties.rst
```diff
@@ -854,6 +854,18 @@ This can also be specified on a per-query basis using the ``confidence_based_bro
 Enable treating ``LOW`` confidence, zero estimations as ``UNKNOWN`` during joins. This can also be specified
 on a per-query basis using the ``treat-low-confidence-zero-estimation-as-unknown`` session property.
 
+``optimizer.scalar-function-stats-propagation-enabled``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
steveburnett, line 858:
```suggestion
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
The formatting should be the same number of characters as the characters in the line being formatted.

steveburnett (changes requested):
One nit of formatting.

steveburnett (dismissed):
LGTM! (docs)

presto-main/src/main/java/com/facebook/presto/operator/scalar/annotations/ScalarFromAnnotationsParser.java
```diff
@@ -99,11 +106,37 @@ private static List<ScalarHeaderAndMethods> findScalarsInFunctionSetClass(Class<
...
+        }
+        return scalarStatsHeader;
+    }
+
     private static SqlScalarFunction parseParametricScalar(ScalarHeaderAndMethods scalar, Optional<Constructor<?>> constructor)
     {
         ScalarImplementationHeader header = scalar.getHeader();
 
         Map<SpecializedSignature, ParametricScalarImplementation.Builder> signatures = new HashMap<>();
+        Map<Signature, ScalarStatsHeader> signatureVsStatsHeader = new HashMap<>();
```
aaneja, line 139: We prefer the use of immutable collections, can we switch this to use a `ImmutableMap.builder()`

aaneja: Let's rename this to `signatureToStatsHeader`

presto-main/src/main/java/com/facebook/presto/operator/scalar/annotations/ScalarFromAnnotationsParser.java
```diff
@@ -131,7 +166,10 @@ private static SqlScalarFunction parseParametricScalar(ScalarHeaderAndMethods sc
         header.getOperatorType().ifPresent(operatorType ->
                 validateOperator(operatorType, scalarSignature.getReturnType(), scalarSignature.getArgumentTypes()));
 
-        return new ParametricScalar(scalarSignature, header.getHeader(), implementations);
+        ScalarHeader scalarHeader = header.getHeader();
+        ScalarHeader headerWithStats =
+                new ScalarHeader(scalarHeader.getDescription(), scalarHeader.getVisibility(), scalarHeader.isDeterministic(), scalarHeader.isCalledOnNullInput(), signatureVsStatsHeader);
```
aaneja, line 171: nit : This line is still more than 180 columns

presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsCalculator.java
```diff
@@ -136,13 +148,42 @@ public VariableStatsEstimate visitCall(CallExpression call, Void context)
...
+            if (isStatsPropagationEnabled) {
+                if (functionMetadata.hasStatsHeader() && call.getFunctionHandle() instanceof BuiltInFunctionHandle) {
+                    Signature signature = ((BuiltInFunctionHandle) call.getFunctionHandle()).getSignature().canonicalization();
+                    Optional<ScalarStatsHeader> statsHeader = functionMetadata.getStatsHeader(signature);
+                    if (statsHeader.isPresent()) {
+                        return computeCallStatistics(call, context, statsHeader.get());
+                    }
+                }
+                else {
+                    log.debug("Stats not found for func: " + functionMetadata.getName() + " " + call);
```
aaneja, line 174: Use printf style log statements - `log.debug("Scalar stats annotation missing for function : %s %s", functionMetadata.getName(), call)`
Ideally, we avoid debug logs in Presto

aaneja (commented):
Still reviewing. Posting partial comments

presto-main/src/main/java/com/facebook/presto/operator/scalar/annotations/ScalarFromAnnotationsParser.java
```diff
@@ -99,11 +107,38 @@ private static List<ScalarHeaderAndMethods> findScalarsInFunctionSetClass(Class<
...
+    {
+        Optional<ScalarStatsHeader> scalarStatsHeader;
+        ScalarFunctionConstantStats constantStatsAnnotation =
+                annotated.getAnnotation(ScalarFunctionConstantStats.class);
+        List<java.lang.reflect.Parameter> params =
+                Arrays.stream(annotated.getParameters())
+                        .filter(param -> param.getAnnotation(SqlType.class) != null)
+                        .collect(Collectors.toList());
+        // Map of (function argument position index) -> (ScalarPropagateSourceStats annotation)
+        ImmutableMap.Builder<Integer, ScalarPropagateSourceStats> argumentStatsResolverBuilder = new ImmutableMap.Builder<>();
```
aaneja, line 120: This can be simplified to just a `List<ScalarPropagateSourceStats>`, the list index is the argument position index

presto-spi/src/main/java/com/facebook/presto/spi/function/StatsPropagationBehavior.java
```diff
@@ -0,0 +1,53 @@
...
+    MAX_TYPE_WIDTH_VARCHAR,
+    /** Stats are unknown and thus no action is performed. */
+    UNKNOWN;
+
+    /*
+     * Stats are single argument when their value is calculated by operating on stats from source stats of the argument which
+     * is annotated.
+     */
+    public static final Set<StatsPropagationBehavior> SINGLE_ARGUMENT_STATS =
+            Collections.unmodifiableSet(new HashSet<>(Arrays.asList(UNKNOWN, ROW_COUNT, NON_NULL_ROW_COUNT, USE_TYPE_WIDTH_VARCHAR, USE_SOURCE_STATS)));
```
aaneja, line 47: Lets use `ImmutableSet.of()` to make this more readable

presto-spi/src/main/java/com/facebook/presto/spi/function/StatsPropagationBehavior.java
```diff
@@ -0,0 +1,53 @@
...
+    /** Take max of type width of arguments with varchar type. */
+    MAX_TYPE_WIDTH_VARCHAR,
+    /** Stats are unknown and thus no action is performed. */
+    UNKNOWN;
+
+    /*
+     * Stats are single argument when their value is calculated by operating on stats from source stats of the argument which
+     * is annotated.
+     */
+    public static final Set<StatsPropagationBehavior> SINGLE_ARGUMENT_STATS =
```
aaneja, line 46: IMO it would be easier to instead store `MULTI_ARGUMENT_STATS` which are stats behaviors which use mulitple function arguments to arrive at a new stats value

presto-spi/src/main/java/com/facebook/presto/spi/function/StatsPropagationBehavior.java
```diff
@@ -0,0 +1,53 @@
...
+            Collections.unmodifiableSet(new HashSet<>(Arrays.asList(UNKNOWN, ROW_COUNT, NON_NULL_ROW_COUNT, USE_TYPE_WIDTH_VARCHAR, USE_SOURCE_STATS)));
+
+    public boolean isSingleArgumentStats()
```
aaneja, line 49: nit: this should have a singular form, i.e `isASingleArgumentStat`

presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsAnnotationProcessor.java
```diff
@@ -0,0 +1,255 @@
...
+        if (!operation.isSingleArgumentStats()) {
+            boolean sourceStatUnknown = false;
+            for (int i = 0; i < sourceStats.size() && !sourceStatUnknown; i++) {
+                switch (operation) {
+                    case MAX_TYPE_WIDTH_VARCHAR:
+                        statValue = maxExcludingNaNs(statValue, typeWidthVarchar(callExpression, i));
+                        break;
+                    case USE_MIN_ARGUMENT:
+                        sourceStatUnknown = !isFinite(sourceStats.get(i));
+                        statValue = minExcludingNaNs(statValue, sourceStats.get(i));
```
aaneja, line 151: Instead of tracking `sourceStatUnknown` we could just return `NaN` from here if `sourceStats.get(i)` is a `NaN`

aaneja (commented):
- Still need to review tests
- Can you please squash your commits

presto-spi/src/main/java/com/facebook/presto/spi/function/ScalarFunctionStatsUtils.java
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.spi.function;
+
+public final class ScalarFunctionStatsUtils
```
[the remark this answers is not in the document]
ScrapCodes: Fields in annotation - ScalarFunctionConstantStats, can take double values only.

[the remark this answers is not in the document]
ScrapCodes: I can try for Enum.getValue() passed to annotations.

presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsAnnotationProcessor.java
```diff
@@ -0,0 +1,255 @@
...
+                }
+                if (sum > 0) {
+                    return sum;
+                }
+            }
+        }
+        return NaN;
+    }
+
+    private double firstNonNaN(double... values)
```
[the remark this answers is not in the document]
ScrapCodes: > * This can be moved to MoreMath class too. That class already has a `firstNonNaN` method. Will that suffice ?

No, because that one does not allow all NaNs in arguments.

>     * This function will return `NaN` if all values are `NaN`. Is this expected ?

Yes
> 
>     * You can return early when the first `NaN` is found instead of iterating through the remaining values

Yes

presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsAnnotationProcessor.java
ScrapCodes: I considered that, 

* This way it is explicit, the reader of the code will know our intention.
* If I use min instead of minExcludingNaNs, then I need to add condition if statValue is NaN and sourceStats are defined then based on whether it is USE_MIN_ARGUMENT, USE_MAX... assign the initial value of statValue.
For the sake of conciseness and clarity, I leaned this way.  However, It may be otherwise, I am happy to change it - WDYT?

presto-spi/src/main/java/com/facebook/presto/spi/function/StatsPropagationBehavior.java
ScrapCodes: We would need guava dependency in presto-spi.

presto-main/src/main/java/com/facebook/presto/operator/scalar/annotations/ScalarFromAnnotationsParser.java
ScrapCodes: Challenge here is, what do I add at those argument position where we do not have an annotation. 
```java
if (params.get(paramIndex).getAnnotation(ScalarPropagateSourceStats.class) == null) {
                argumentStatsResolverBuilder.add((ScalarPropagateSourceStats) null);
            }
```
This code produces null pointer exception with Immutable list. 

* One option is to use Optional.valueOf, but that will make it a List<Optional<ScalarPropagateSourceStats>>..

presto-main/src/main/java/com/facebook/presto/operator/scalar/annotations/ScalarFromAnnotationsParser.java
```diff
@@ -99,11 +106,37 @@ private static List<ScalarHeaderAndMethods> findScalarsInFunctionSetClass(Class<
         return builder.build();
     }
 
+    private static Optional<ScalarStatsHeader> getScalarStatsHeader(Method annotated)
```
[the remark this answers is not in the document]
ScrapCodes: Don't remember trying this approach.
* If passed in annotatedElement is a class, but methods of the class may have annotation - that needs to be handled. 
* Currently, we support parametric scalar functions, I have not tried with [SQL scalar functions](https://github.com/prestodb/presto/tree/master/presto-main/src/main/java/com/facebook/presto/operator/scalar/sql)

presto-main/src/main/java/com/facebook/presto/operator/scalar/annotations/ScalarFromAnnotationsParser.java
aaneja: Ok, makes sense then

presto-spi/src/main/java/com/facebook/presto/spi/function/StatsPropagationBehavior.java
aaneja: Ah I see we [have](https://github.com/prestodb/presto/blob/master/presto-spi/pom.xml#L88-L92) it as test only

ScrapCodes:
@aaneja still working on the feedback

presto-main/src/main/java/com/facebook/presto/operator/scalar/annotations/ScalarFromAnnotationsParser.java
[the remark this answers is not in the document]
ScrapCodes: Do you want the relevant code to compute signature of the function be moved to `fromAnnotatedElement` ?

ScrapCodes:
@aaneja Please take a look again!

presto-main/src/test/java/com/facebook/presto/cost/TestScalarStatsCalculator.java
```diff
@@ -73,6 +89,334 @@ public void setUp()
...
+                        maxValue = StatsPropagationBehavior.SUM_ARGUMENTS) @SqlType(StandardTypes.BIGINT) long x,
+                @SqlType(StandardTypes.BIGINT) long y)
+        {
+            return x + y;
+        }
+
+        @ScalarFunction(value = "custom_is_null", calledOnNullInput = true)
+        @LiteralParameters("x")
+        @SqlType(StandardTypes.BOOLEAN)
+        @ScalarFunctionConstantStats(distinctValuesCount = 2.0, nullFraction = 0.0)
```
aaneja, line 114: Can we set `distinctValuesCount` to some value other than `2.0` which is also used in the `custom_is_null` BIGINT variant. GIves us confidence that we are picking the value from the actual annotated header

presto-main/src/test/java/com/facebook/presto/cost/TestScalarStatsCalculator.java
```diff
@@ -73,6 +89,334 @@ public void setUp()
...
+                @ScalarPropagateSourceStats(
+                        propagateAllStats = false,
+                        maxValue = StatsPropagationBehavior.USE_SOURCE_STATS
+                ) @SqlNullable @SqlType(StandardTypes.BIGINT) Long max)
+        {
+            return (long) ((Math.random() * (max - min)) + min);
+        }
+    }
+
+    @Test
```
aaneja, line 176: Can you add negative tests -
1. NDV value set in annotation to higher than output row count
2. Average row size is set to greater than the actual width of the type
3. max is greater than min
4. null fraction is negative or more than `1`

For some of the cases (negative values, out of bound values) would the stats propagation fail to register and would we get the existing do-not-propagate  stats behavior ?

aaneja (changes requested):

presto-main/src/test/java/com/facebook/presto/cost/TestScalarStatsCalculator.java
ScrapCodes: Yes, only exception is average row size which is upper bounded to return type width of the function (if known).

presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsAnnotationProcessor.java
```diff
@@ -0,0 +1,249 @@
...
+                distinctValuesCount = outputRowCount;
+            }
+        }
+        return distinctValuesCount;
+    }
+
+    private double processDistinctValuesCount(double distinctValuesCount)
+    {
+        double distinctValuesCountFinal = firstFiniteValue(getConstantDistinctValuesCount(), distinctValuesCount);
+        if (isFinite(outputRowCount)) {
```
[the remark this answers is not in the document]
ScrapCodes: One option is to add 
SUM_ARGUMENTS_UPPER_BOUND_ROW_COUNT to deal with above case where sum of source stats of more than one argument exceeds row count.
And for all other cases, when NDV is set greater than row count, just reset it to NaN (i.e. unknown).

presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsAnnotationProcessor.java
```diff
@@ -0,0 +1,247 @@
...
+        }
+        return VariableStatsEstimate.builder()
+                .setLowValue(minValue)
+                .setHighValue(maxValue)
+                .setNullsFraction(nullFraction)
+                .setAverageRowSize(minExcludingNaNs(firstFiniteValue(scalarStatsHeader.getAvgRowSize(), averageRowSize), getReturnTypeWidth(callExpression, UNKNOWN)))
+                .setDistinctValuesCount(processDistinctValuesCount(distinctValuesCount)).build();
+    }
+
+    private double getConstantDistinctValuesCount()
```
elharo, line 107: Why is this a double instead of an integer? A number of distinct values feels fundamentally integral to me. If that's not what this is, then the method might need to be renamed.

presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsAnnotationProcessor.java
```diff
@@ -0,0 +1,247 @@
...
+                Long longLiteral = typeSignature.getParameters().get(0).getLongLiteral();
+                if (longLiteral > 0 && longLiteral != VarcharType.UNBOUNDED_LENGTH) {
+                    return longLiteral;
+                }
+            }
+        }
+        return NaN;
+    }
+
+    private double getReturnTypeWidth(CallExpression call, StatsPropagationBehavior operation)
```
elharo, line 195: Again, it's surprising to see a double here. At first read, this really sounds like an integer to me. Why is this a double?

presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsAnnotationProcessor.java
```diff
@@ -0,0 +1,247 @@
...
+            return ((FixedWidthType) call.getType()).getFixedSize();
+        }
+        if (call.getType() instanceof VarcharType) {
+            VarcharType returnType = (VarcharType) call.getType();
+            if (!returnType.isUnbounded()) {
+                return returnType.getLengthSafe();
+            }
+            if (operation == SUM_ARGUMENTS) {
+                // since return type is a varchar and length is unknown,
+                // getting it by doing a SUM each argument's varchar type bounds.
```
elharo, line 207: unclear; please rephrase comment

presto-main/src/main/java/com/facebook/presto/util/MoreMath.java
```diff
@@ -79,6 +81,50 @@ public static double max(double... values)
...
+            if (isFinite(v)) {
+                if (isNaN(sum)) {
+                    sum = v;
+                }
+                else {
+                    sum += v;
+                }
+            }
+        }
+        return sum;
```
elharo, line 97: This can still return NaN if no value is finite

presto-main/src/main/java/com/facebook/presto/util/MoreMath.java
```diff
@@ -79,6 +81,50 @@ public static double max(double... values)
                 .getAsDouble();
     }
 
+    public static double sumExcludingNaNs(double... values)
```
elharo, line 84: needs api doc

presto-main/src/main/java/com/facebook/presto/util/MoreMath.java
```diff
@@ -79,6 +81,50 @@ public static double max(double... values)
...
+                }
+                else {
+                    sum += v;
+                }
+            }
+        }
+        return sum;
+    }
+
+    public static double minExcludingNaNs(double... values)
```
elharo, line 100: needs api doc

presto-main/src/main/java/com/facebook/presto/util/MoreMath.java
```diff
@@ -79,6 +81,50 @@ public static double max(double... values)
...
+        double min = NaN;
+        for (double v : values) {
+            if (isFinite(v)) {
+                if (isNaN(min)) {
+                    min = v;
+                }
+                min = Math.min(min, v);
+            }
+        }
+        return min;
```
elharo, line 111: This can be NaN

presto-main/src/main/java/com/facebook/presto/util/MoreMath.java
```diff
@@ -79,6 +81,50 @@ public static double max(double... values)
...
+        double max = NaN;
+        for (double v : values) {
+            if (isFinite(v)) {
+                if (isNaN(max)) {
+                    max = v;
+                }
+                max = Math.max(max, v);
+            }
+        }
+        return max;
```
elharo, line 125: This can be NaN

elharo (changes requested):

presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsAnnotationProcessor.java
ScrapCodes: I need it because, `VariableStatsEstimate` needs all the values as double. This way we avoid type conversions. https://github.com/prestodb/presto/blob/master/presto-main/src/main/java/com/facebook/presto/cost/VariableStatsEstimate.java

presto-main/src/test/java/com/facebook/presto/cost/TestScalarStatsCalculator.java
ScrapCodes: So, I am thinking of changing this slightly, 
1. If average row size is provided via `ScalarFunctionConstantStats` annotation, then we allow even if the size is greater than functions return type width.
2. If average row size is provided via one of the `StatsPropagationBehavior` values, then we upper bound it to functions return type width - if available.
3. If both 1 and 2 is unknown, then we default it to functions return type width if available.

This way the function developer has greater control.

ScrapCodes: For distinct values count, upper bounding it to row count is as good as unknown. Therefore, the approach here is, when distinctValuesCount is greater than row count and is provided via annotation we set it to unknown. 
A function developer has full control here, for example developer can choose to upper bound or not by selecting the appropriate `StatsPropagationBehavior` value.

ScrapCodes:
Hi @elharo , thanks for taking a look. Relevant docs and comments are added/improved. 
Regarding you comment on why we are using double as the return type for average row size and distinct values count, because these values are actually estimates and may be fractional. Infact for this reason `VariableStatsEstimate` class has all these fields values as double type. We could have used integer at some places for example `getReturnTypeWidth`, but that adds complexity of repeated type casting/conversions. Double type also allows us to set the value to NaN, indicating it as an unknown.

presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsAnnotationProcessor.java
```diff
@@ -0,0 +1,249 @@
...
+            maxValue = firstFiniteValue(maxValue, processSingleArgumentStatistic(
+                    sourceStats.stream().map(VariableStatsEstimate::getHighValue).collect(toImmutableList()),
+                    paramIndexVsStatsMap.getKey(),
+                    applyPropagateAllStats(propagateAllStats, scalarPropagateSourceStats.maxValue())));
+            minValue = firstFiniteValue(minValue, processSingleArgumentStatistic(
+                    sourceStats.stream().map(VariableStatsEstimate::getLowValue).collect(toImmutableList()),
+                    paramIndexVsStatsMap.getKey(),
+                    applyPropagateAllStats(propagateAllStats, scalarPropagateSourceStats.minValue())));
+        }
+        if (!isFinite(maxValue) || !isFinite(minValue)) {
```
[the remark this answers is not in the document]
ScrapCodes: Corrected it by checking only for isNaN()

presto-spi/src/main/java/com/facebook/presto/spi/function/ScalarPropagateSourceStats.java
```diff
@@ -0,0 +1,34 @@
...
+public @interface ScalarPropagateSourceStats
+{
+    boolean propagateAllStats() default true;
+
+    StatsPropagationBehavior minValue() default StatsPropagationBehavior.UNKNOWN;
+    StatsPropagationBehavior maxValue() default StatsPropagationBehavior.UNKNOWN;
+    StatsPropagationBehavior distinctValuesCount() default StatsPropagationBehavior.UNKNOWN;
+    StatsPropagationBehavior avgRowSize() default StatsPropagationBehavior.UNKNOWN;
+    StatsPropagationBehavior nullFraction() default StatsPropagationBehavior.UNKNOWN;
+    StatsPropagationBehavior histogram() default StatsPropagationBehavior.UNKNOWN;
```
ZacBlanco, line 33: I think if there is no use of the `histogram()` it should be left out

presto-spi/src/main/java/com/facebook/presto/spi/function/ScalarFunctionConstantStats.java
```diff
@@ -0,0 +1,63 @@
...
+@Retention(RUNTIME)
+@Target(METHOD)
+public @interface ScalarFunctionConstantStats
+{
+    // Min max value is Infinity if unknown.
+    double minValue() default Double.NEGATIVE_INFINITY;
+    double maxValue() default Double.POSITIVE_INFINITY;
+
+    // Histogram
+    HistogramTypes histogram() default HistogramTypes.UNKNOWN;
```
ZacBlanco, line 44: same comment on histogram

ZacBlanco (changes requested):

presto-main/src/main/java/com/facebook/presto/util/MoreMath.java
```diff
@@ -79,6 +81,40 @@ public static double max(double... values)
                 .getAsDouble();
     }
 
+    /**
+     * Find minimum value from finite arguments, and return NaN otherwise.
```
elharo, line 85: Returns the minimum value of the arguments. Returns NaN if there are no arguments or all arguments are NaN.

presto-main/src/main/java/com/facebook/presto/util/MoreMath.java
```diff
@@ -79,6 +81,40 @@ public static double max(double... values)
...
+                min = Math.min(min, v);
+            }
+        }
+        return min;
+    }
+
+    /**
+     * Find maximum value from finite arguments, and return NaN otherwise.
+     */
+    public static double maxExcludingNaNs(double... values)
```
elharo, line 104: Returns the maximum value of the arguments. Returns NaN if there are no arguments or all arguments are NaN.

presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsCalculator.java
```diff
@@ -101,6 +113,36 @@ public VariableStatsEstimate calculate(RowExpression scalarExpression, PlanNodeS
         return scalarExpression.accept(new RowExpressionStatsVisitor(inputStatistics, session), null);
     }
 
+    private static VariableStatsEstimate estimateCoalesce(PlanNodeStatsEstimate input, VariableStatsEstimate left, VariableStatsEstimate right)
```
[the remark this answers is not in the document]
ScrapCodes: Sorry about this, somehow this creep in as a result of rearranging code using the codestyle guide of Intellij.

presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsAnnotationProcessor.java
```diff
@@ -0,0 +1,236 @@
...
+import static java.lang.Double.isNaN;
+
+public final class ScalarStatsAnnotationProcessor
+{
+    private ScalarStatsAnnotationProcessor()
+    {
+    }
+
+    public static VariableStatsEstimate process(
+            double outputRowCount,
```
elharo, line 50: probably an int

presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsAnnotationProcessor.java
```diff
@@ -0,0 +1,236 @@
...
+            CallExpression callExpression,
+            List<VariableStatsEstimate> sourceStats,
+            ScalarStatsHeader scalarStatsHeader)
+    {
+        double nullFraction = scalarStatsHeader.getNullFraction();
+        double distinctValuesCount = NaN;
```
elharo, line 56: int

presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsAnnotationProcessor.java
```diff
@@ -0,0 +1,236 @@
...
+        double averageRowSize = NaN;
+        double maxValue = scalarStatsHeader.getMax();
+        double minValue = scalarStatsHeader.getMin();
+        for (Map.Entry<Integer, ScalarPropagateSourceStats> paramIndexVsStatsMap : scalarStatsHeader.getArgumentStats().entrySet()) {
```
elharo, line 60: Vs is unclear. Perhaps paramIndexToStatsMap

presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsAnnotationProcessor.java
```diff
@@ -0,0 +1,236 @@
...
+        }
+        return VariableStatsEstimate.builder()
+                .setLowValue(minValue)
+                .setHighValue(maxValue)
+                .setNullsFraction(nullFraction)
+                .setAverageRowSize(firstFiniteValue(scalarStatsHeader.getAvgRowSize(), averageRowSize, returnNaNIfTypeWidthUnknown(getReturnTypeWidth(callExpression, UNKNOWN))))
+                .setDistinctValuesCount(processDistinctValuesCount(outputRowCount, nullFraction, scalarStatsHeader.getDistinctValuesCount(), distinctValuesCount)).build();
+    }
+
+    private static double processDistinctValuesCount(double outputRowCount, double nullFraction, double distinctValuesCountFromConstant, double distinctValuesCount)
```
elharo, line 97: The counts should ints

presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsAnnotationProcessor.java
```diff
@@ -0,0 +1,236 @@
...
+    {
+        if (isFinite(distinctValuesCountFromConstant)) {
+            if (distinctValuesCountFromConstant == NON_NULL_ROW_COUNT.getValue()) {
+                distinctValuesCountFromConstant = outputRowCount * (1 - firstFiniteValue(nullFraction, 0.0));
+            }
+            else if (distinctValuesCount == ROW_COUNT.getValue()) {
```
elharo, line 103: Given that your doing direct equality comparisons with no tolerance, these MUST be ints.

presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsAnnotationProcessor.java
```diff
@@ -0,0 +1,236 @@
...
+        }
+        double distinctValuesCountFinal = firstFiniteValue(distinctValuesCountFromConstant, distinctValuesCount);
+        if (distinctValuesCountFinal > outputRowCount) {
+            distinctValuesCountFinal = NaN;
+        }
+        return distinctValuesCountFinal;
+    }
+
+    private static double processSingleArgumentStatistic(
+            double outputRowCount,
```
elharo, line 115: int

presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsAnnotationProcessor.java
```diff
@@ -0,0 +1,236 @@
...
+            StatsPropagationBehavior operation)
+    {
+        // sourceStatsArgumentIndex is index of the argument on which
+        // ScalarPropagateSourceStats annotation was applied.
+        double statValue = NaN;
+        if (operation.isMultiArgumentStat()) {
+            for (int i = 0; i < sourceStats.size(); i++) {
+                if (i == 0 && operation.isSourceStatsDependentStats() && isFinite(sourceStats.get(i))) {
+                    statValue = sourceStats.get(i);
+                    continue;
```
elharo, line 129: consider using an else block instead of a continue

presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsAnnotationProcessor.java
ScrapCodes: These and others are double values across the code base, I am not sure if we can change that.

presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsAnnotationProcessor.java
ScrapCodes: Thanks for catching this, switched to Double.compare, hopefully that should be okay.

presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsAnnotationProcessor.java
elharo: I'm pretty sure we can. It might be inconvenient to do in this PR, though I'm not sure about that. New methods absolutely can use ints/longs for things that are integers. Just maybe you need to pass in a value from existing code that is already *incorrectly* a double,. If so, and if we need to move forward with this before fixing that, then please file a detailed issue. And also consider using the right type in the new methods and casting the value before passing it into the new methods so the source of the bad type is clear. 

However, very likely many if not all of these new arguments and variables can simply be made ints now.

presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsAnnotationProcessor.java
```diff
@@ -0,0 +1,240 @@
...
+                int sum = 0;
+                for (RowExpression r : callExpression.getArguments()) {
+                    if (r instanceof CallExpression) { // argument is another function call
+                        sum += getReturnTypeWidth((CallExpression) r, UNKNOWN);
+                    }
+                    else {
+                        sum += typeWidthVarchar(r.getType());
+                    }
+                }
+                if (sum > 0) {
```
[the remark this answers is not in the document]
ScrapCodes: This annotation (ScalarPropagateSourceStats) can be applied to the arguments of the scalar function only, so no arguments situation will not happen.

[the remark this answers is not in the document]
elharo: Then this can't be zero or negative and you can simply return the sum.

elharo (changes requested):

presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsAnnotationProcessor.java
aaneja: I think the reason all the members in `VariableStatsEstimate` (and related types like `PlanNodeStatsEstimate`) use doubles for integral properties is because this is the largest width primitive type we can use in Java (BigInteger could be used, but is not a primitive type)

It's fairly easy to build a query that will hit the limits of `long`, e.g - 
```
presto:sf100> explain select 1 from lineitem l1, lineitem l2, lineitem l3 , lineitem l4, lineitem l4;
                                                                                                                         Query Plan                                                                                                                         
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 - Output[PlanNodeId 13][_col0] => [expr:integer]                                                                                                                                                                                                           
         Estimates: {source: CostBasedSourceInfo, rows: 9,223,372,036,854,775,807 (8192PB), cpu: 388,922,817,995,845,400,000,000,000,000,000,000,000,000,000.00, memory: 0.00, network: 388,922,817,995,845,400,000,000,000,000,000,000,000,000,000.00}     
         _col0 := expr (1:16)  
```

In the above cross-join example using TPCH SF100 lineitem, the output row count would be `600037902^4
= 1.29632750431103743243e35` (this is incorrectly truncated to long max in the [TextRenderer](https://github.com/prestodb/presto/blob/03a2bd8bd31dcc51e0f61b389e32d63dde0f807a/presto-main/src/main/java/com/facebook/presto/sql/planner/planPrinter/TextRenderer.java#L249))

elharo: But row count is not a return type width. In the example you cite, my initial reaction would be to use BigInteger, and perhaps consider double if performance really requires it. However, returnTypeWidth fits inside a single byte. There's no reason I can see to use a double here. 

Unless maybe returnTypeWidth is not at all what I think it is? But if so, this method needs to be renamed to be clearer.

aaneja: Gotcha. Agreed about this return type.  

My point was broader about using doubles in the optimizer core stats classes. While VariableStatsEstimate#averageRowSize is unlikely to hit a long max, fields like PlanNodeStatsEstimate#totalSize which represents the estimate of the size of the data flow, can hit limits pretty easily (e.g the `8192PB` in the above query is also incorrect and actual size is much larger)

ScrapCodes: Perhaps his comment is an answer to all your comments regarding double -> int/long across the code base: https://github.com/prestodb/presto/pull/23545#discussion_r1786094895

presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsAnnotationProcessor.java
```diff
@@ -0,0 +1,243 @@
...
+                        case USE_MIN_ARGUMENT:
+                            statValue = min(statValue, sourceStats.get(i));
+                            break;
+                        case USE_MAX_ARGUMENT:
+                            statValue = max(statValue, sourceStats.get(i));
+                            break;
+                        case SUM_ARGUMENTS:
+                            statValue = statValue + sourceStats.get(i);
+                            break;
+                        case SUM_ARGUMENTS_UPPER_BOUNDED_TO_ROW_COUNT:
```
ZacBlanco, line 146: Sorry if I've missed discussion on this in the past but what are the cases where we want the behavior of `SUM_ARGUMENTS` as opposed to `SUM_ARGUMENTS_UPPER_BOUNDED_TO_ROW_COUNT`

presto-main/src/main/java/com/facebook/presto/operator/scalar/annotations/ScalarFromAnnotationsParser.java
```diff
@@ -99,11 +107,38 @@ private static List<ScalarHeaderAndMethods> findScalarsInFunctionSetClass(Class<
...
+    {
+        Optional<ScalarStatsHeader> scalarStatsHeader;
+        ScalarFunctionConstantStats constantStatsAnnotation =
+                annotated.getAnnotation(ScalarFunctionConstantStats.class);
+        List<java.lang.reflect.Parameter> params =
+                Arrays.stream(annotated.getParameters())
```
ZacBlanco, line 116: nit: can we just import `Parameter` instead of using `java.lang.reflect.Parameter`?

ZacBlanco (commented):
Just a few minor things, otherwise lgtm

presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsAnnotationProcessor.java
ScrapCodes: Useful when a function developer want to use SUM of arguments stats for computing distinctValuesCount, which can sometimes exceed the row count and may not be the intention of the developer.

The discussion was, previously we were upper bounding by default which is not so helpful. In this way, for computing distinct values count through SUM_ARGUMENTS, user has better control, whether or not he wants to upper bound to row count. May be you or @aaneja has a better perspective on this.
Link to the discussion: https://github.com/prestodb/presto/pull/23545#discussion_r1778295120

presto-docs/src/main/sphinx/admin/properties.rst
```diff
@@ -854,6 +854,18 @@ This can also be specified on a per-query basis using the ``confidence_based_bro
...
 
+``optimizer.scalar-function-stats-propagation-enabled``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+
+* **Type:** ``boolean``
+* **Default value:** ``false``
+
+Enable scalar functions stats propagation using annotations. Annotations ``@ScalarFunctionConstantStats`` and
+``@ScalarPropagateSourceStats`` define the behavior of the scalar function stats characteristics.
```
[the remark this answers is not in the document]
ScrapCodes: Currently, this is only supported for builtin functions. In a subsequent PR, both will be added the support as well as supporting documentation.

presto-docs/src/main/sphinx/admin/properties.rst
```diff
@@ -884,6 +884,17 @@ This can also be specified on a per-query basis using the ``confidence_based_bro
...
+``optimizer.scalar-function-stats-propagation-enabled``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+
+* **Type:** ``boolean``
+* **Default value:** ``false``
+
+Enable scalar functions stats propagation using annotations. Annotations define the behavior of the scalar
+function's stats characteristics. When set to ``true``, this property enables the stats propagation through annotations.
+This can also be specified on a per-query basis using the ``scalar_function_stats_propagation_enabled`` session property.
```
ZacBlanco, line 896: I think this is better than before. However, users still shouldn't know/care that a function is annotated or even that annotations exist. It is an implementation detail. How about:
```suggestion
When set to ``true``, this property enables statistics to be propagated from function arguments to higher level expressions. When set to ``false`` any function calls result in unknown statistics which can negatively impact query planning.
This can also be specified on a per-query basis using the ``scalar_function_stats_propagation_enabled`` session property.
```

presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsAnnotationProcessor.java
ScrapCodes: The https://github.com/prestodb/presto/pull/23545#discussion_r1793207848 specifies the reason for using doubles here and other places.

ScrapCodes:
Hi @elharo is it good to go?

presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsAnnotationProcessor.java
```diff
@@ -0,0 +1,243 @@
...
+                return sum;
+            }
+        }
+        return -VarcharType.MAX_LENGTH;
+    }
+
+    // Return first 'finite' value from values, else return values[0]
+    private static double firstFiniteValue(double... values)
+    {
+        checkArgument(values.length > 1);
```
elharo, line 226: > 0?

That is, a single element list is OK? or just check values is not empty

presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsAnnotationProcessor.java
```diff
@@ -0,0 +1,243 @@
...
+    {
+        if (argumentType instanceof VarcharType) {
+            if (!((VarcharType) argumentType).isUnbounded()) {
+                return ((VarcharType) argumentType).getLengthSafe();
+            }
+        }
+        return -VarcharType.MAX_LENGTH;
+    }
+
+    private static double returnNaNIfTypeWidthUnknown(int typeWidthValue)
```
elharo, line 182: I'd be tempted to just use a negative value here. Is it realy necessary to use NaN?

ZacBlanco (dismissed):

ScrapCodes:
> I'm still not sold on using doubles for row counts and the like. Doubles are not just a bigger long and have round off and representation problems integers don't. If these things are really going to exceed the size of a long (which I'm not sure they do) then a BigInteger would be preferred.

BigInteger does not exist in java natively. May be this is a decision that presto project has made much before this PR came in. Will it be a good idea, if you can start a mailing group thread with broader audience?

elharo:
BigInteger isn't a primitive type, but java.math.BigInteger is available if you need it. Again, I'm not convinced you do. long should work here.

presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsAnnotationProcessor.java
ScrapCodes: Yes, NaN indicates unknown.

presto-main/src/main/java/com/facebook/presto/cost/ScalarStatsAnnotationProcessor.java
ScrapCodes: We expect at least two arguments in this private function.

ZacBlanco (approved):

aaneja (dismissed):

presto-spi/src/main/java/com/facebook/presto/spi/function/Signature.java
```diff
@@ -168,6 +168,17 @@ public String toString()
                 "(" + String.join(",", argumentTypes.stream().map(TypeSignature::toString).collect(toList())) + "):" + returnType;
     }
 
+    /*
```
aaneja, line 171: I'm not sure if this is really a 'canonical' form. At any rate, since this only finds used in the ScalarFormAnnotationsParser (and other consumers) can we instead move it there as a static method ?

presto-spi/src/main/java/com/facebook/presto/spi/function/ScalarStatsHeader.java
```diff
@@ -0,0 +1,97 @@
...
+    {
+        this.min = min;
+        this.max = max;
+        this.argumentStatsResolver = argumentStatsResolver;
+        this.distinctValuesCount = distinctValuesCount;
+        this.nullFraction = nullFraction;
+        this.avgRowSize = avgRowSize;
+    }
+
+    public ScalarStatsHeader(ScalarFunctionConstantStats methodConstantStats, Map<Integer, ScalarPropagateSourceStats> argumentStatsResolver)
```
aaneja, line 42: Can you reorder your parameters ? The other constructor has `argumentStatsResolver` as the first parameter

aaneja: Can you add a JavaDoc for both constructors ? It's not clear what `method` is in `methodConstantStats`

presto-spi/src/main/java/com/facebook/presto/spi/function/ScalarStatsHeader.java
```diff
@@ -0,0 +1,97 @@
...
+    public ScalarStatsHeader(Map<Integer, ScalarPropagateSourceStats> argumentStatsResolver)
+    {
+        this(argumentStatsResolver, Double.NEGATIVE_INFINITY, Double.POSITIVE_INFINITY, Double.NaN, Double.NaN, Double.NaN);
+    }
+
+    @Override
+    public String toString()
+    {
+        return String.format("distinctValuesCount: %g , nullFraction: %g, avgRowSize: %g, min: %g, max: %g",
+                distinctValuesCount, nullFraction, avgRowSize, min, max);
```
aaneja, line 61: Why are we not using the toString on `argumentStatsResolver` as well ?

presto-spi/src/main/java/com/facebook/presto/spi/function/ScalarFunctionConstantStats.java
```diff
@@ -0,0 +1,60 @@
...
+package com.facebook.presto.spi.function;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * By default, a function is just a “black box” that the database system knows very little about the behavior of.
```
aaneja, line 24: Is this JavaDoc in the right place ? Should it instead be - 
```
This annotation is used for specifying constant values for stats propagation for scalar functions. It takes precedence over any per-argument `ScalarPropagateSourceStats`
```

presto-spi/src/main/java/com/facebook/presto/spi/function/ScalarFunctionConstantStats.java
```diff
@@ -0,0 +1,60 @@
...
+@Target(METHOD)
+public @interface ScalarFunctionConstantStats
+{
+    // Min max value is Infinity if unknown.
+    double minValue() default Double.NEGATIVE_INFINITY;
+    double maxValue() default Double.POSITIVE_INFINITY;
+
+    /**
+     * A constant value for Distinct values count regardless of `input column`'s source stats.
+     * e.g. a perfectly random generator may result in distinctValuesCount of `ScalarFunctionStatsUtils.ROW_COUNT`.
```
aaneja, line 45: The example is unclear. Can you reword it be more precise - 
```
For example the `random(double)` scalar function can set this `ScalarFunctionStatsUtils.ROW_COUNT`
```

Also, IMO, this and other examples can live on the interface's javadoc itself

presto-main/src/main/java/com/facebook/presto/metadata/BuiltInTypeAndFunctionNamespaceManager.java
```diff
@@ -1181,6 +1182,19 @@ else if (function instanceof SqlInvokedFunction) {
...
+            return new FunctionMetadata(
+                    signature.getName(),
+                    signature.getArgumentTypes(),
+                    signature.getReturnType(),
+                    signature.getKind(),
+                    JAVA,
+                    function.isDeterministic(),
+                    function.isCalledOnNullInput(),
+                    sqlFunction.getComplexTypeFunctionDescriptor(),
+                    sqlFunction.getScalarHeader().getSignatureToScalarStatsHeadersMap());
```
aaneja, line 1196: Instead of passing the whole Map as an argument to FunctionMetadata, can we not just lookup the `ScalarStatsHeader` to see if it exists for the `signature` and pass in an `Optional<ScalarStatsHeader` ?

aaneja (changes requested):

presto-spi/src/main/java/com/facebook/presto/spi/function/Signature.java
ScrapCodes: For example a scalar string function signature such as `varchar(y) fun1(varchar(x) , varchar(z))` since the values x, y, z depend on invocation of this scalar function (e.g. `varchar(20) fun1(varchar(10) , varchar(10))` ). Now to convert it into a form that can be equated, it is be transformed to `varchar fun1(varchar , varchar)`. This is the reason for calling it a canonical form (since the original function signature needs to be compared regardless of how it was invoked). Feel free to correct me on this and I am open to a better naming.

presto-spi/src/main/java/com/facebook/presto/spi/function/StatsPropagationBehavior.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+package com.facebook.presto.spi.function;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+import static java.util.Collections.unmodifiableSet;
+
+public enum StatsPropagationBehavior
```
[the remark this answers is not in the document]
ScrapCodes: Seems like a very long name (I am okay with this too), can we achieve the same effect via package renaming? for example
`package com.facebook.presto.spi.function.scalar`

PR #23546, opened 2024-08-29
Add support for AuthorizedIdentity credentials proxying

prithvip:
## Description
This change allows for AuthorizedIdentity to be included as a claim in a JWT token. This is required for https://github.com/prestodb/rfcs/pull/23, however this contribution is valuable on its own. 

## Motivation and Context
Today, Presto supports credentials proxying through means of a JWT token. The proxy authenticates the client, and constructs a JWT bearer token. The token is then signed and sent to the Presto coordinator in the "AUTHORIZATION" HTTP Header. The coordinator trusts the proxy if the token was signed using a shared secret (HMAC), or RSA key. 

"Principal" can be proxied through a JWT token, by means of the "subject" claim. However, the AccessControl::selectAuthorizedIdentity method requires submittal of all the X509 certificates that the client presented. This is a problem in a proxy setup, because JWT credentials could only proxy the principal, and the certificates themselves would not be forwarded. To fully support proxying, this change allows for AuthorizedIdentity to be included as a claim in the same JWT token. 

An alternative approach could be to forward the X509 certificates in a client-cert header, and implement IETF RFC 9440 (https://datatracker.ietf.org/doc/rfc9440/). However, this has two major drawbacks: 1) the proxy server would only be able to authenticate clients using certificate-based authentication 2) there would be issues with header size limitations. Thus, we chose to use the JWT mechanism instead. 

## Impact
No user-facing changes or performance impact. 

## Test Plan
Unit testing 

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Add support to proxy AuthorizedIdentity via JWT :pr:`23546`
```

### DIFF
--- a/presto-main/pom.xml
+++ b/presto-main/pom.xml
@@ -369,7 +369,6 @@
         <dependency>
             <groupId>io.jsonwebtoken</groupId>
             <artifactId>jjwt-jackson</artifactId>
-            <scope>runtime</scope>
         </dependency>
 
         <dependency>
--- a/presto-main/pom.xml
+++ b/presto-main/pom.xml
@@ -138,6 +138,11 @@
 
         <dependency>
             <groupId>com.facebook.airlift</groupId>
+            <artifactId>security</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>com.facebook.airlift</groupId>
             <artifactId>jaxrs</artifactId>
         </dependency>
 
--- a/presto-main/src/main/java/com/facebook/presto/server/security/JsonWebTokenAuthenticator.java
+++ b/presto-main/src/main/java/com/facebook/presto/server/security/JsonWebTokenAuthenticator.java
@@ -1,0 +1,252 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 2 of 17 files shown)


presto-main/src/main/java/com/facebook/presto/server/security/JsonWebTokenAuthenticator.java
```diff
@@ -0,0 +1,252 @@
...
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Strings.nullToEmpty;
+import static com.google.common.io.Files.asCharSource;
+import static com.google.common.net.HttpHeaders.AUTHORIZATION;
+import static java.lang.String.format;
+import static java.nio.charset.StandardCharsets.US_ASCII;
+import static java.util.Base64.getMimeDecoder;
+import static java.util.Objects.requireNonNull;
+
+public class JsonWebTokenAuthenticator
```
abhiseksaikia, line 59: I am wondering if we can add the support of passing AuthorizedIdentity in the airlift code itself instead of moving the class into presto-main and adding this functionality.

prithvip: I considered this approach, but it would require airlift to understand Presto-specific concepts such as AuthorizedIdentity. There are some authenticators in airlift and some authenticators in Presto, so I think it is reasonable to move this authenticator to Presto.

abhiseksaikia (dismissed):
LGTM!

presto-main/pom.xml
```diff
@@ -369,7 +369,7 @@
         <dependency>
             <groupId>io.jsonwebtoken</groupId>
             <artifactId>jjwt-jackson</artifactId>
-            <scope>runtime</scope>
+            <scope>compile</scope>
```
rschlussel, line 372: you should be able to remove this line.  compile is the default scope.

rschlussel (approved):
looks good, just small comment about removing the unneeded compile scope declaration.

presto-main/pom.xml
prithvip: done

rschlussel (approved):

PR #23572, opened 2024-09-02
[native] Require semicolon after velox macros, avoid triggering `-Wextra-semi`

dtolnay:
https://github.com/facebookincubator/velox/pull/10914 is going to wrap macros like `VELOX_CHECK` and `VELOX_UNREACHABLE` in `do { ... } while (0)` to enforce the use of a trailing semicolon. Before that change, writing velox macro calls with a trailing semicolon would trigger an needless semicolon warning in some circumstances.


dtolnay:
Sorting out CLA stuff. I'll resubmit this not via facebook-github-bot as soon as I can.

zpao:
@dtolnay I added you to our CCLA so FWIW you should be good on this PR - commit is authored by you (with public email) so the facebook bot export shouldn't impact anything. EasyCLA does require re-flowing through what you've already done to re-trigger the check (and PR needs to be open)

PR #23581, opened 2024-09-04
[native] Fix cmake target name

lingbin:
All targets under presto_cpp should use "presto_" as the prefix, so
rename the "prometheus_reporter" target to
"presto_prometheus_reporter".

No functional change.

```
== NO RELEASE NOTE ==
```


amitkdutta (approved):
Looks good. Thanks @lingbin

PR #23584, opened 2024-09-04
[parquet] Support 64-bit RLE-encoded ShortDecimal

ZacBlanco:
## Description

Previously, in the parquet writer short decimals could be written as RLE-encoded with an Int64 primitive type. However we lacked support in the reader to decode this type properly back into a short decimal.

This commit adds support for the RLE-encoded 64-bit short decimals.

## Motivation and Context

Decimals can be stored in three different formats: INT32 (P <= 9), INT64 (9 < P <= 18), and FIXED_LEN_BYTE_ARRAY (P > 18). We were missing read support for RLE-encoded short decimals with 9 < P <= 18. The following sequence of actions leads to failure.

```sql
SET SESSION iceberg.parquet_batch_read_optimization_enabled = true;
SET SESSION iceberg.parquet_writer_version = 'PARQUET_2_0';
CREATE table tmp AS SELECT * FROM (VALUES (1, CAST(15.2 as decimal(15, 2))), (2, 15.2), (3, 15.2), (4, 15.2)) as t(i, j);

presto:tpch> SELECT * FROM tmp;

Query 20240904_171754_00022_5rgwk, FAILED, 2 nodes
Splits: 5 total, 0 done (0.00%)
[Latency: client-side: 0:03, server-side: 0:03] [0 rows, 0B] [0 rows/s, 0B/s]

Query 20240904_171754_00022_5rgwk failed: com.facebook.presto.parquet.batchreader.decoders.rle.Int64RLEDictionaryValuesDecoder cannot be cast to com.facebook.presto.parquet.batchreader.decoders.ValuesDecoder$ShortDecimalValuesDecoder
```

After these changes, this error no longer appears.

## Impact

N/A

## Test Plan

- tests in parquet module

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/decoders/Decoders.java
+++ b/presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/decoders/Decoders.java
@@ -179,6 +179,9 @@ public class Decoders
                     if (isTimeStampMicrosType(columnDescriptor) || isTimeMicrosType(columnDescriptor)) {
                         return new Int64TimeAndTimestampMicrosRLEDictionaryValuesDecoder(bitWidth, inputStream, (LongDictionary) dictionary);
                     }
+                    if (isDecimalType(columnDescriptor) && isShortDecimalType(columnDescriptor)) {
+                        return new Int64RLEDictionaryValuesDecoder(bitWidth, inputStream, (LongDictionary) dictionary);
+                    }
                 }
                 case DOUBLE: {
                     return new Int64RLEDictionaryValuesDecoder(bitWidth, inputStream, (LongDictionary) dictionary);
--- a/presto-hive/src/test/java/com/facebook/presto/hive/parquet/AbstractTestParquetReader.java
+++ b/presto-hive/src/test/java/com/facebook/presto/hive/parquet/AbstractTestParquetReader.java
@@ -939,6 +939,23 @@ public abstract class AbstractTestParquetReader
         }
     }
 
+    @Test
+    public void testRLEDecimalBackedByINT64()
+            throws Exception
+    {
+        int[] scales = {9, 9, 9, 9, 9, 9, 9, 9, 9};
+        for (int precision = MAX_PRECISION_INT32 + 1; precision <= MAX_PRECISION_INT64; precision++) {
+            int scale = scales[precision - MAX_PRECISION_INT32 - 1];
+            MessageType parquetSchema = parseMessageType(format("message hive_decimal { optional INT64 test (DECIMAL(%d, %d)); }", precision, scale));
+            ContiguousSet<Long> longValues = longsBetween(1, 1_000);
+            ImmutableList.Builder<SqlDecimal> expectedValues = new ImmutableList.Builder<>();
+            for (Long value : longValues) {
+                expectedValues.add(SqlDecimal.of(value, precision, scale));
+            }
+            tester.testRoundTrip(javaLongObjectInspector, longValues, expectedValues.build(), createDecimalType(precision, scale), Optional.of(parquetSchema));
+        }
+    }
+
     private void testDecimal(int precision, int scale, Optional<MessageType> parquetSchema)
             throws Exception
     {
--- a/presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/decoders/rle/Int64RLEDictionaryValuesDecoder.java
+++ b/presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/decoders/rle/Int64RLEDictionaryValuesDecoder.java
@@ -14,6 +14,7 @@
 package com.facebook.presto.parquet.batchreader.decoders.rle;
 
 import com.facebook.presto.parquet.batchreader.decoders.ValuesDecoder.Int64ValuesDecoder;
+import com.facebook.presto.parquet.batchreader.decoders.ValuesDecoder.ShortDecimalValuesDecoder;
 import com.facebook.presto.parquet.dictionary.LongDictionary;
 import org.apache.parquet.io.ParquetDecodingException;
 import org.openjdk.jol.info.ClassLayout;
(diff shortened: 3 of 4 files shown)


elharo (approved):

presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/decoders/Decoders.java
```diff
@@ -178,6 +179,9 @@ private static ValuesDecoder createValuesDecoder(ColumnDescriptor columnDescript
                     if (isTimeStampMicrosType(columnDescriptor)) {
                         return new Int64TimestampMicrosRLEDictionaryValuesDecoder(bitWidth, inputStream, (LongDictionary) dictionary);
                     }
+                    if (isDecimalType(columnDescriptor) && isShortDecimalType(columnDescriptor)) {
```
hantangwangd, line 182: Can we omit the first condition? Is there a scenario where the `columnDescriptor` is a `ShortDecimalType` but not a `DecimalType`?

hantangwangd (commented):
Thanks for this supplement. Change looks good to me, just one little thing.

presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/decoders/Decoders.java
ZacBlanco: Confusing, but these technically, these check for two separate things

- `isDecimalType` checks if the primitive type is `DECIMAL`
- `isShortDecimal` actually checks the logical type annotation is decimal and then gets the precision parameter from the logical type to check if it is a short decimal

From my understanding `isShortDecimal` should always be true if `isDecimalType` returns true. However, this is the same check as in the previous block for `FLOAT`. I would prefer to stay consistent. Also, it is my understanding that Presto may not properly write logical type annotations yet according to #23388 -- so maybe let's just keep this now for consistency's sake?

hantangwangd: If we look deeper at `isDecimalType`, we will find that it ultimately also verifies that the logical type annotation is decimal. So as I understand, if a `columnDescriptor` is of type `ShortDecimalType`, it must first be of type `DecimalType`. Furthermore, if the primitive type's logical type annotation is null, it can neither be a `DecimalType` nor a `ShortDecimalType`, so it won't break the conclusion above (The current experiments in my local show the same behavior). And we also can see that in `encoding == PLAIN` clause, the checks in `INT32` and `INT64` are all `isShortDecimalType` only.

I'm OK for now to keep it as is, since I am not very sure if there are still some special scenarios present. But once it is completely checked and confirmed, I think it would be better to delete the first condition (so as to the previous block you mentioned) because it would cause a lot of confusion for future readers.

hantangwangd (dismissed):

presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/decoders/rle/Int64ShortDecimalRLEDictionaryValuesDecoder.java
```diff
@@ -0,0 +1,41 @@
...
+{
+    public Int64ShortDecimalRLEDictionaryValuesDecoder(int bitWidth, InputStream in, LongDictionary dictionary)
+    {
+        super(bitWidth, in, dictionary);
+    }
+
+    @Override
+    public void readNext(long[] values, int offset, int length) throws IOException
+    {
+        long[] tempValues = new long[length];
```
yingsu00, line 35: Allocating a buffer in every readNext call is bad for reliability and performance. There is no need to create a separate decoder and add unnecessary memory allocation and memory copies. I know you're copying existing code, but I suggest just use Int64RLEDictionaryValuesDecoder. If you have time, it would be nice to send a new PR to unify existing RLE dictionary decoders. After all, dictionary IDs can only be RLE/BP encoded, and is not relevant to the data column types. 

Ref: https://parquet.apache.org/docs/file-format/data-pages/encodings/
"Data page format: the bit width used to encode the entry ids stored as 1 byte (max bit width = 32), followed by the values encoded using RLE/Bit packed described above (with the given bit width)."

ZacBlanco: I updated this to just use the Int64 decoder

ZacBlanco:
I will look into improving the other bits in the batch reader to reduce allocations on the read path. It is probably also a good issue for new folks getting started with Presto.

presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/decoders/rle/Int64ShortDecimalRLEDictionaryValuesDecoder.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+package com.facebook.presto.parquet.batchreader.decoders.rle;
+
+import com.facebook.presto.parquet.batchreader.decoders.ValuesDecoder.ShortDecimalValuesDecoder;
+import com.facebook.presto.parquet.dictionary.LongDictionary;
+
+import java.io.InputStream;
+
+public class Int64ShortDecimalRLEDictionaryValuesDecoder
```
yingsu00, line 22: This class is not doing anything different than Int64RLEDictionaryValuesDecoder. Can we just use Int64RLEDictionaryValuesDecoder directly?

ZacBlanco: I've updated this to add `ShortDecimalValuesDecoder` to the `implements` list on `Int64RLEDictionaryValuesDecoder`  instead. I personally prefer having the explicit reader class as it makes the code a bit easier to follow IMO

ScrapCodes (approved):

yingsu00 (approved):

PR #23588, opened 2024-09-05
[native] Terminate all Velox macro calls with semicolon

dtolnay:
An upcoming change in Velox (https://github.com/facebookincubator/velox/pull/10914) is going to enforce the use of a semicolon in statement-like macros such as `VELOX_CHECK` and `VELOX_UNREACHABLE`.

Velox's current implementation of these macros, for which a trailing semicolon at the call site is optional, is problematic because the semicolons can trip the C++ compiler's `-Wextra-semi` lint in some circumstances. The lint is technically unimpeachable because the macro calls would have continued to compile with the caller's semicolon removed.

The overwhelming majority of Velox macro calls (in Velox as well as Presto) were already written with a trailing semicolon before this PR.

## Release Notes

== NO RELEASE NOTE ==


amitkdutta (approved):
Looks great. Thanks @dtolnay

PR #21495, opened 2023-12-07
Add ALTER TABLE SET PROPERTIES statement

pratyakshsharma:
Cherry-pick of https://github.com/trinodb/trino/pull/9401

Co-authored-by: ebyhr

## Motivation and Context
Setting table properties can help in modifying critical table properties such as the one which controls concurrency behaviour for an iceberg table.

## Impact
ALTER TABLE ... SET PROPERTIES

## Test Plan
Manual testing
Unit testing

## Contributor checklist

- [x] Please make sure your submission complies with our [development](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#development), [formatting](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#formatting), [commit message](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#commit-formatting-and-pull-requests), and [attribution guidelines](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#attribution).
- [x] PR description addresses the issue accurately and concisely.  If the change is non-trivial, a GitHub Issue is referenced.
- [x] Documented new properties (with its default value), SQL syntax, functions, or other functionality.
- [x] If release notes are required, they follow the [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines).
- [x] Adequate tests were added if applicable.
- [x] CI passed.

## Release Notes
Please follow [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines) and fill in the release notes below.

```
== RELEASE NOTES ==

General Changes
* Add :doc:`/sql/alter-table` SET PROPERTIES statement :pr:`21495`
```

PR #23592, opened 2024-09-05
Make Iceberg's procedures uniformly support named arguments

hantangwangd:
## Description

This PR enables existing procedures which do not yet support named arguments to support named arguments, ensures consistency in the naming of arguments across procedures, and supplements the document with explanations for named arguments and sequential arguments.

Besides, this PR removes a procedure `set_table_property` which is completely unused.

## Motivation and Context

Make Iceberg's procedures uniformly support named arguments

## Impact

After this change, Iceberg procedures `register_table` and `unregister_table` support named arguments as well

## Test Plan

 - Newly added test cases to show the support of using named arguments in procedure `register_table` and `unregister_table`

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

Iceberg Connector Changes
* Support using named arguments in procedure `register_table` and `unregister_table` :pr:`12345`
```

### DIFF
--- a/presto-docs/src/main/sphinx/connector/iceberg.rst
+++ b/presto-docs/src/main/sphinx/connector/iceberg.rst
@@ -732,7 +732,7 @@ example uses the earliest snapshot ID: ``2423571386296047175``
 Procedures
 ----------
 
-Use the CALL statement to perform data manipulation or administrative tasks. Procedures are available in the ``system`` schema of the catalog.
+Use the :doc:`/sql/call` statement to perform data manipulation or administrative tasks. Procedures are available in the ``system`` schema of the catalog.
 
 Register Table
 ^^^^^^^^^^^^^^
--- a/presto-docs/src/main/sphinx/connector/iceberg.rst
+++ b/presto-docs/src/main/sphinx/connector/iceberg.rst
@@ -739,20 +739,41 @@ Register Table
 
 Iceberg tables for which table data and metadata already exist in the
 file system can be registered with the catalog. Use the ``register_table``
-procedure on the catalog's ``system`` schema and supply the target schema,
-desired table name, and the location of the table metadata::
+procedure on the catalog's ``system`` schema to register a table which
+already exists but does not known by the catalog.
+
+The following arguments are available:
+
+===================== ========== =============== =======================================================================
+Argument Name         required   type            Description
+===================== ========== =============== =======================================================================
+``schema``            ✔️         string          Schema of the table to register
+
+``table_name``        ✔️         string          Name of the table to register
+
+``metadata_location`` ✔️         string          The location of the table metadata which is to be registered
+
+``metadata_file``                string          An optionally specified metadata file which is to be registered
+===================== ========== =============== =======================================================================
+
+Examples:
+
+* Register a table through supplying the target schema, desired table name, and the location of the table metadata::
 
     CALL iceberg.system.register_table('schema_name', 'table_name', 'hdfs://localhost:9000/path/to/iceberg/table/metadata/dir')
+
+    CALL iceberg.system.register_table(table_name => 'table_name', schema => 'schema_name', metadata_location => 'hdfs://localhost:9000/path/to/iceberg/table/metadata/dir')
 
 .. note::
 
     If multiple metadata files of the same version exist at the specified
     location, the most recently modified one is used.
 
-A metadata file can optionally be included as an argument to ``register_table``
-where a specific metadata file contains the targeted table state::
+* Register a table through additionally supplying a specific metadata file::
 
     CALL iceberg.system.register_table('schema_name', 'table_name', 'hdfs://localhost:9000/path/to/iceberg/table/metadata/dir', '00000-35a08aed-f4b0-4010-95d2-9d73ef4be01c.metadata.json')
+
+    CALL iceberg.system.register_table(table_name => 'table_name', schema => 'schema_name', metadata_location => 'hdfs://localhost:9000/path/to/iceberg/table/metadata/dir', metadata_file => '00000-35a08aed-f4b0-4010-95d2-9d73ef4be01c.metadata.json')
 
 .. note::
 
@@ -775,9 +796,23 @@ Unregister Table
 ^^^^^^^^^^^^^^^^
 
 Iceberg tables can be unregistered from the catalog using the ``unregister_table``
-procedure on the catalog's ``system`` schema::
+procedure on the catalog's ``system`` schema.
+
+The following arguments are available:
+
+===================== ========== =============== ===================================
+Argument Name         required   type            Description
+===================== ========== =============== ===================================
+``schema``            ✔️         string          Schema of the table to unregister
+
+``table_name``        ✔️         string          Name of the table to unregister
+===================== ========== =============== ===================================
+
+Examples::
 
     CALL iceberg.system.unregister_table('schema_name', 'table_name')
+
+    CALL iceberg.system.unregister_table(table_name => 'table_name', schema => 'schema_name')
 
 .. note::
 
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/procedure/RegisterTableProcedure.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/procedure/RegisterTableProcedure.java
@@ -84,9 +84,9 @@ public class RegisterTableProcedure
                 "register_table",
                 ImmutableList.of(
                         new Procedure.Argument("schema", VARCHAR),
-                        new Procedure.Argument("table", VARCHAR),
-                        new Procedure.Argument("metadataLocation", VARCHAR),
-                        new Procedure.Argument("metadataFile", VARCHAR, false, null)),
+                        new Procedure.Argument("table_name", VARCHAR),
+                        new Procedure.Argument("metadata_location", VARCHAR),
+                        new Procedure.Argument("metadata_file", VARCHAR, false, null)),
                 REGISTER_TABLE.bindTo(this));
     }
 
--- a/presto-docs/src/main/sphinx/sql/call.rst
+++ b/presto-docs/src/main/sphinx/sql/call.rst
@@ -18,6 +18,22 @@ Procedures can be provided by connectors to perform data manipulation or
 administrative tasks. For example, the :doc:`/connector/system` defines a
 procedure for killing a running query.
 
+CALL statement supports passing arguments by name or by position. Mixing position and named arguments is not supported.
+
+* Named arguments
+
+Procedure arguments all have a name. When passing arguments by name, arguments can be in any order
+and any optional argument can be omitted::
+
+    CALL catalog_name.system.procedure_name(arg_name_2 => arg_2, arg_name_1 => arg_1);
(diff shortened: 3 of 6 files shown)


presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -734,26 +734,63 @@ Procedures
...
+
+Procedure arguments all have a name. When passing arguments by name, arguments can be in any order
+and any optional argument can be omitted.
+
+    CALL iceberg.system.procedure_name(arg_name_2 => arg_2, arg_name_1 => arg_1);
+
+* Positional arguments
+
+When passing arguments by position, only the ending arguments may be omitted if they are optional.
+The intermediate parameters cannot be omitted whether optional or not.
```
steveburnett, line 749:
```suggestion
Intermediate parameters cannot be omitted.
```
Suggest considering shortening the sentence as shown, for clarity.

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -734,26 +734,63 @@ Procedures
...
 Use the CALL statement to perform data manipulation or administrative tasks. Procedures are available in the ``system`` schema of the catalog.
 
+CALL statement supports passing arguments by name or by position. Mixing position and named arguments is not supported.
+
+* Named arguments
+
+Procedure arguments all have a name. When passing arguments by name, arguments can be in any order
+and any optional argument can be omitted.
+
+    CALL iceberg.system.procedure_name(arg_name_2 => arg_2, arg_name_1 => arg_1);
```
steveburnett, line 744: Formatting nit: suggest fixing the indent of this line and of line 751 so the vertical lines visible in a local doc build screenshot are not created. 

Screenshot 2024-09-05 at 12 53 59 PM

steveburnett (changes requested):
Thanks for the doc! A formatting nit and a suggested rephrasing to consider.

presto-docs/src/main/sphinx/connector/iceberg.rst
hantangwangd: Thanks for pointing this out, fixed! Please take a look when available.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/procedure/SetTablePropertyProcedure.java
```diff
@@ -1,98 +0,0 @@
...
-        ConnectorMetadata metadata = metadataFactory.create();
-        IcebergTableName tableName = IcebergTableName.from(table);
-        SchemaTableName schemaTableName = new SchemaTableName(schema, tableName.getTableName());
-        Table icebergTable = IcebergUtil.getIcebergTable(metadata, session, schemaTableName);
-
-        icebergTable.updateProperties()
-                .set(key, value)
-                .commit();
-    }
-}
```
ZacBlanco, line 98: Is there a good reason to remove it? If I recall, the reason I added this was because iceberg's table properties are generic string-string key-value pairs. Presto doesn't allow setting any arbitrary properties through regular SQL syntax though. So this adds some useful functionality for users who might have some arbitrary table-level metadata to store.

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -734,26 +734,63 @@ Procedures
 
 Use the CALL statement to perform data manipulation or administrative tasks. Procedures are available in the ``system`` schema of the catalog.
 
+CALL statement supports passing arguments by name or by position. Mixing position and named arguments is not supported.
+
+* Named arguments
+
+Procedure arguments all have a name. When passing arguments by name, arguments can be in any order
+and any optional argument can be omitted::
```
ZacBlanco, line 742: IMO this should be documented more at the level of the `CALL` syntax rather than in the iceberg connector documentation. We shouldn't need it here as it isn't really iceberg-specific. Same with the positional argument syntax block below.

@steveburnett thoughts?

ZacBlanco (changes requested):
Thank you for these changes! I appreciate consistency 😄

presto-docs/src/main/sphinx/connector/iceberg.rst
steveburnett: That's a good point @ZacBlanco! If it's not Iceberg-specific and is relevant to all `CALL` statements, then I agree this content should be moved to the [CALL](https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/sql/call.rst) doc. 

I would think it appropriate to keep the Procedures heading, and the sentence "Use the CALL statement to perform data manipulation or administrative tasks. Procedures are available in the ``system`` schema of the catalog.", and add "See :doc:`/sql/call`." Does that work?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/procedure/SetTablePropertyProcedure.java
imjalpreet: > Besides, this PR removes a procedure set_table_property which is completely unused.

@hantangwangd It was unused as it's missing binding in IcebergCommonModule, so the Procedure wasn't even enabled till now. 

We can add the binding and enable it as it should be useful as Zac mentioned. What do you think?

presto-docs/src/main/sphinx/connector/iceberg.rst
imjalpreet: +1

presto-iceberg/src/main/java/com/facebook/presto/iceberg/procedure/RegisterTableProcedure.java
```diff
@@ -115,6 +115,7 @@ private void doRegisterTable(ConnectorSession clientSession, String schema, Stri
                         getFileSystem(clientSession, hdfsEnvironment, schemaTableName, metadataDirectory),
                         metadataDirectory));
 
+        System.out.println("==[RegisterTableProcedure]= " + metadataPath);
```
imjalpreet, line 118: nit: is this required?

presto-docs/src/main/sphinx/connector/iceberg.rst
steveburnett: Even better, delete my added "See :doc:`/sql/call`." and use the link in the sentence itself, like this: 

```
Use the :doc:`/sql/call` statement to perform data manipulation or administrative tasks. Procedures are available in the ``system`` schema of the catalog.
```

presto-iceberg/src/main/java/com/facebook/presto/iceberg/procedure/SetTablePropertyProcedure.java
hantangwangd: I just saw that this procedure was not used at all, not even registered as @imjalpreet mentioned. I do not have a strong feeling to remove it, since the PR #21495 has been blocked for quite some time, I would like to register and enable this procedure if you think it's necessary. Let me know your thoughts @ZacBlanco @imjalpreet.

presto-docs/src/main/sphinx/connector/iceberg.rst
hantangwangd: Thanks all for the suggestion, that looks wonderful, fixed.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/procedure/RegisterTableProcedure.java
hantangwangd: Nice catch, forgot to delete it, fixed.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/procedure/TestIcebergRegisterProcedure.java
```diff
@@ -373,6 +400,17 @@ public void testUnregisterTable()
...
+    public void testUnregisterTableWithNamedArguments()
+    {
+        String tableName = "unregister";
+        assertUpdate("CREATE TABLE " + tableName + " (id integer, value integer)");
+
+        // Unregister table with procedure
+        assertUpdate("CALL system.unregister_table(table_name => '" + tableName + "', schema => '" + TEST_SCHEMA + "')");
+        assertFalse(getQueryRunner().tableExists(getSession(), tableName));
+    }
+
```
agrawalreetika, line 413: Would it be better to rename either the this test class or Add a new Test class to incorporate `unregister_table` procedure related tests?

Also test classes for `SetTablePropertyProcedure` & `RollbackToSnapshotProcedure` procedures are missing, should we add few tests for that as well Or may be handle tests in future PR? WDYT?

steveburnett (dismissed):
LGTM! (docs)

presto-iceberg/src/test/java/com/facebook/presto/iceberg/procedure/TestIcebergRegisterProcedure.java
hantangwangd: > Would it be better to rename either the this test class or Add a new Test class to incorporate unregister_table procedure related tests?

Good suggestion, I have renamed the test class.

> Also test classes for SetTablePropertyProcedure & RollbackToSnapshotProcedure procedures are missing, should we add few tests for that as well Or may be handle tests in future PR?

Under your reminder, I realized that there are still some different kinds of work to be done. I think it's better to leave the work of registering and documenting `SetTablePropertyProcedure`, adding test cases for both `SetTablePropertyProcedure` and `RollbackToSnapshotProcedure` to subsequent separated PRs. And in this PR, we focus on support named arguments uniformly for existing procedures, and unifying the naming conventions for their arguments. Do you think this makes sense?

agrawalreetika: Sounds good to me 👍🏻

presto-iceberg/src/main/java/com/facebook/presto/iceberg/procedure/SetTablePropertyProcedure.java
hantangwangd: Now keep `SetTablePropertyProcedure` as it is, maybe it's better to handle it in a later dedicated PR.

ZacBlanco (approved):

agrawalreetika (approved):

presto-iceberg/src/main/java/com/facebook/presto/iceberg/procedure/SetTablePropertyProcedure.java
imjalpreet: Sounds good, we can enable the procedure and add some tests in a follow-up PR.

imjalpreet (approved):
LGTM, thank you.

steveburnett (approved):
LGTM! (docs)

presto-docs/src/main/sphinx/connector/iceberg.rst
steveburnett: Looks good in a new build, thanks!

Issue #22028, opened 2024-02-27
Add DDL statements to drop branches and tags


tdcmeehan:
To support WAP workflows, Presto should support the deletion of branches and tags.  These are metadata-only operations which can be accomplished through a new `ALTER TABLE` syntax.

Part of #22025 

## Expected Behavior or Use Case
`ALTER TABLE iceberg.default.mytable DROP BRANCH 'audit-branch'`

`ALTER TABLE iceberg.default.mytable DROP TAG 'audit-tag'`

## Presto Component, Service, or Connector
Iceberg connector, with additional syntax support from the parser, and additional connector functionality to support branching and tagging.  Only Iceberg will initially have this support.

PR #23614, opened 2024-09-10
feat(plugin-iceberg): Add DDL statements to drop branches and tags

agrawalreetika:
## Description
Add DDL statements to drop branches and tags

## Motivation and Context
Resolves https://github.com/prestodb/presto/issues/22028

## Impact
Resolves https://github.com/prestodb/presto/issues/22028

SQL support for dropping a branch from a table :

```
ALTER TABLE users DROP BRANCH 'branch1';
```

SQL support for dropping a tag from a table : 
```
ALTER TABLE users DROP TAG 'tag1';
```

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Add DDL support for dropping a branch from a table. 
* Add DDL support for dropping a tag from a table.

Iceberg Connector Changes
* Add support for dropping a branch from an Iceberg table.
* Add support for dropping a tag from an Iceberg table.
```

### DIFF
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
@@ -926,6 +926,22 @@ public abstract class IcebergAbstractMetadata
     }
 
     @Override
+    public void dropTag(ConnectorSession session, ConnectorTableHandle tableHandle, String tagName, boolean tagExists)
+    {
+        IcebergTableHandle icebergTableHandle = (IcebergTableHandle) tableHandle;
+        verify(icebergTableHandle.getIcebergTableName().getTableType() == DATA, "only the data table can have tag dropped");
+        Table icebergTable = getIcebergTable(session, icebergTableHandle.getSchemaTableName());
+        if (icebergTable.refs().containsKey(tagName) && icebergTable.refs().get(tagName).isTag()) {
+            icebergTable.manageSnapshots().removeTag(tagName).commit();
+        }
+        else {
+            if (!tagExists) {
+                throw new PrestoException(NOT_FOUND, format("Tag %s doesn't exist in table %s", tagName, icebergTableHandle.getSchemaTableName().getTableName()));
+            }
+        }
+    }
+
+    @Override
     public void addColumn(ConnectorSession session, ConnectorTableHandle tableHandle, ColumnMetadata column)
     {
         if (!column.isNullable()) {
--- a/presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedTestBase.java
+++ b/presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedTestBase.java
@@ -2071,6 +2071,33 @@ public abstract class IcebergDistributedTestBase
     }
 
     @Test
+    public void testDropTag()
+    {
+        assertUpdate("CREATE TABLE test_table_tag (id1 BIGINT, id2 BIGINT)");
+        assertUpdate("INSERT INTO test_table_tag VALUES (0, 00), (1, 10)", 2);
+
+        Table icebergTable = loadTable("test_table_tag");
+        icebergTable.manageSnapshots().createTag("testTag1", icebergTable.currentSnapshot().snapshotId()).commit();
+
+        assertUpdate("INSERT INTO test_table_tag VALUES (2, 30), (3, 30)", 2);
+        icebergTable.manageSnapshots().createTag("testTag2", icebergTable.currentSnapshot().snapshotId()).commit();
+        assertUpdate("INSERT INTO test_table_tag VALUES (4, 40), (5, 50)", 2);
+        assertEquals(icebergTable.refs().size(), 3);
+
+        assertQuery("SELECT count(*) FROM test_table_tag FOR SYSTEM_VERSION AS OF 'testTag1'", "VALUES 2");
+        assertQuery("SELECT count(*) FROM test_table_tag FOR SYSTEM_VERSION AS OF 'testTag2'", "VALUES 4");
+        assertQuery("SELECT count(*) FROM test_table_tag FOR SYSTEM_VERSION AS OF 'main'", "VALUES 6");
+
+        assertQuerySucceeds("ALTER TABLE test_table_tag DROP TAG 'testTag1'");
+        icebergTable = loadTable("test_table_tag");
+        assertEquals(icebergTable.refs().size(), 2);
+        assertQueryFails("ALTER TABLE test_table_tag DROP TAG 'testTagNotExist'", "Tag testTagNotExist doesn't exist in table test_table_tag");
+        assertQuerySucceeds("ALTER TABLE test_table_tag DROP TAG IF EXISTS 'testTag2'");
+        assertQuerySucceeds("ALTER TABLE test_table_tag DROP TAG IF EXISTS 'testTagNotExist'");
+        assertQuerySucceeds("DROP TABLE test_table_tag");
+    }
+
+    @Test
     public void testRefsTable()
     {
         assertUpdate("CREATE TABLE test_table_references (id1 BIGINT, id2 BIGINT)");
--- a/presto-docs/src/main/sphinx/connector/iceberg.rst
+++ b/presto-docs/src/main/sphinx/connector/iceberg.rst
@@ -1484,6 +1484,8 @@ Alter table operations are supported in the Iceberg connector::
 
      ALTER TABLE iceberg.web.page_views DROP BRANCH 'branch1';
 
+     ALTER TABLE iceberg.web.page_views DROP TAG 'tag1';
+
 To add a new column as a partition column, identify the transform functions for the column.
 The table is partitioned by the transformed value of the column::
 


steveburnett (approved):
LGTM! (docs)

Pull branch, local docs build, looks good. Thanks!

presto-hive/src/main/java/com/facebook/presto/hive/security/SqlStandardAccessControl.java
```diff
@@ -269,6 +271,42 @@ public void checkCanDropColumn(ConnectorTransactionHandle transaction, Connector
...
+                HiveColumnConverterProvider.DEFAULT_COLUMN_CONVERTER_PROVIDER,
+                context.getWarningCollector(),
+                context.getRuntimeStats());
+        if (!isTableOwner(transaction, identity, metastoreContext, tableName)) {
+            denyDropBranch(tableName.toString());
+        }
+    }
+
+    @Override
+    public void checkCanDropTag(ConnectorTransactionHandle transaction, ConnectorIdentity identity, AccessControlContext context, SchemaTableName tableName)
```
ZacBlanco, line 293: I wonder about the granularity of these methods. In other implementation (e.g. spark?) at what granularity to they enforce the ability to do CRUD operations on tags and branches? 

I'm thinking about a few cases

1. A group or user(s) can only access a certain set of branches or tags
2. A group or user(s) can only create branches starting from a specific branch
3. A group or user(s) can create tags

I know we're only implementing `DROP` but I want to understand the whole story for access control around branches and tags. Would we ever need to pass the branch/tag to these methods?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
```diff
@@ -632,6 +633,36 @@ public void rollback()
         // TODO: cleanup open transaction
     }
 
+    @Override
+    public void dropBranch(ConnectorSession session, ConnectorTableHandle tableHandle, Optional<String> branchName)
+    {
+        String branch = branchName.get().replace("'", "");
```
ZacBlanco, line 639: Instead of `get()` maybe do `getOrElseThrow` and include a helpful error message?

ZacBlanco, line 639: Why are we doing `.replace()` here?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
```diff
@@ -632,6 +633,36 @@ public void rollback()
...
+        }
+        else {
+            throw new PrestoException(NOT_FOUND, format("Branch %s doesn't exist in table %s", branchName.get(), icebergTableHandle.getSchemaTableName().getTableName()));
+        }
+    }
+
+    @Override
+    public void dropTag(ConnectorSession session, ConnectorTableHandle tableHandle, Optional<String> tagName)
+    {
+        String tag = tagName.get().replace("'", "");
```
ZacBlanco, line 654: similar comments/questions here

presto-spi/src/main/java/com/facebook/presto/spi/connector/ConnectorMetadata.java
```diff
@@ -837,6 +837,22 @@ default TableLayoutFilterCoverage getTableLayoutFilterCoverage(ConnectorTableLay
...
+    {
+        throw new PrestoException(NOT_SUPPORTED, "This connector does not support dropping table branch");
+    }
+
+    /**
+     * Drop the specified tag
+     */
+    default void dropTag(ConnectorSession session, ConnectorTableHandle tableHandle, Optional<String> tagName)
+    {
+        throw new PrestoException(NOT_SUPPORTED, "This connector does not support dropping table tag");
```
ZacBlanco, line 853:
```suggestion
        throw new PrestoException(NOT_SUPPORTED, "This connector does not support dropping table tags");
```

presto-spi/src/main/java/com/facebook/presto/spi/connector/ConnectorMetadata.java
```diff
@@ -837,6 +837,22 @@ default TableLayoutFilterCoverage getTableLayoutFilterCoverage(ConnectorTableLay
         return NOT_APPLICABLE;
     }
 
+    /**
+     * Drop the specified branch
+     */
+    default void dropBranch(ConnectorSession session, ConnectorTableHandle tableHandle, Optional<String> branchName)
+    {
+        throw new PrestoException(NOT_SUPPORTED, "This connector does not support dropping table branch");
```
ZacBlanco, line 845:
```suggestion
        throw new PrestoException(NOT_SUPPORTED, "This connector does not support dropping table branches");
```

presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedTestBase.java
```diff
@@ -1622,6 +1622,56 @@ public void testDecimal(boolean decimalVectorReaderEnabled)
         }
     }
 
+    @Test
+    public void testDropBranch()
```
ZacBlanco, line 1626: Can we add some tests using the `IF EXISTS` syntax?

presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedTestBase.java
```diff
@@ -1622,6 +1622,56 @@ public void testDecimal(boolean decimalVectorReaderEnabled)
...
+
+        assertQuerySucceeds("ALTER TABLE test_table_branch DROP BRANCH 'testBranch1'");
+        icebergTable = loadTable("test_table_branch");
+        assertEquals(icebergTable.refs().size(), 2);
+        assertQueryFails("ALTER TABLE test_table_branch DROP BRANCH 'testBranchNotExist'", "Branch 'testBranchNotExist' doesn't exist in table test_table_branch");
+        assertQuerySucceeds("DROP TABLE test_table_branch");
+    }
+
+    @Test
+    public void testDropTag()
```
ZacBlanco, line 1651: same comment about adding cases for `IF EXISTS`

ZacBlanco (changes requested):
Some high level feedback. Most of the code looks good. I will take a closer look on 2nd pass.

Also, I wanted to bring a thought I had about parser extensions. Since many connectors will not support this branching and tagging, I was thinking that maybe we ought to consider designing a SQL syntax plugin extension interface. Spark allows custom syntax extensions through implementing some set of interfaces or bringing your own parser. The upstream iceberg project now maintains their own spark SQL syntax extensions.

I'm not proposing we need that for this PR, but maybe it's something we should start thinking about if connectors start adding more radically different features that would be best left to some syntax extensions/optional plugins, especially for things outside the SQL specification.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
agrawalreetika: Here the branch String would come something like `'<branch-name>'` so stripping that before accessing

presto-hive/src/main/java/com/facebook/presto/hive/security/SqlStandardAccessControl.java
agrawalreetika: I think the granularity of access control methods in Spark for CRUD operations on Iceberg tags and branches is limited by Spark's integration with external systems (such as file systems, catalogs, and security frameworks like Apache Ranger). 
For example, Ranger policies can define access controls at the table level, which could be extended to manage specific branches or tag-based access.
And like for cloud-based catalogs like AWS Glue, you can control access to Iceberg metadata (branches and tags) via IAM policies that grant or restrict specific operations.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
ZacBlanco: Is there a reason it's quoted? I think the value passed here should not include the quotes. You might need to edit the parser grammar to strip any quotes

presto-hive/src/main/java/com/facebook/presto/hive/security/SqlStandardAccessControl.java
ZacBlanco: Thanks for the info. Would the parameters passed here as context have enough information for us to act at a similar level? I don't see anything in the method parameters that contains the branch name which I assume we would need to perform access control at a similar level.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
agrawalreetika: This is something open for discussion. As of now I made this quote based on the syntax in the [issue](https://github.com/prestodb/presto/issues/22028). We can discuss if we want to have grammar like other ALTER TABLE operations https://prestodb.io/docs/current/sql/alter-table.html#examples ?

presto-hive/src/main/java/com/facebook/presto/hive/security/SqlStandardAccessControl.java
agrawalreetika: I would like to discuss around this, Systems like Ranger can define access controls at the table level, column level. So in this case I think handle the same access of drop branch & tags could be table based. As per I can think of branch & tag level policies then has to be maintained on engine side if we introduce that in here?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
agrawalreetika: Sounds good to me. Updated the PR

agrawalreetika:
@ZacBlanco Can you please take another pass?

presto-hive/src/main/java/com/facebook/presto/hive/security/SqlStandardAccessControl.java
agrawalreetika: @tdcmeehan What do you think about access control for branches and tags? Would be based on the parent table itself or based on the tags/branches?
My thinking was since there is no policies enforced based on branch/tags via security frameworks so this should owner the same policies as table?

presto-parser/src/main/java/com/facebook/presto/sql/tree/DropBranch.java
```diff
@@ -0,0 +1,116 @@
...
+    {
+        return branchName;
+    }
+
+    public boolean isTableExists()
+    {
+        return tableExists;
+    }
+
+    public boolean isbranchExists()
```
hantangwangd, line 67:
```suggestion
    public boolean isBranchExists()
```

presto-parser/src/main/java/com/facebook/presto/sql/tree/DropBranch.java
```diff
@@ -0,0 +1,116 @@
...
+        return Objects.equals(tableName, that.tableName) &&
+                Objects.equals(branchName, that.branchName) &&
+                Objects.equals(tableExists, that.tableExists) &&
+                Objects.equals(branchExists, branchExists);
+    }
+
+    @Override
+    public int hashCode()
+    {
+        return Objects.hash(tableName, tableExists, branchName);
```
hantangwangd, line 103: Did you miss `branchExists`?

presto-parser/src/main/java/com/facebook/presto/sql/tree/DropBranch.java
```diff
@@ -0,0 +1,116 @@
...
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        DropBranch that = (DropBranch) o;
+        return Objects.equals(tableName, that.tableName) &&
+                Objects.equals(branchName, that.branchName) &&
+                Objects.equals(tableExists, that.tableExists) &&
+                Objects.equals(branchExists, branchExists);
```
hantangwangd, line 97:
```suggestion
                Objects.equals(branchExists, that.branchExists);
```

presto-parser/src/main/java/com/facebook/presto/sql/tree/DropTag.java
```diff
@@ -0,0 +1,116 @@
...
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        DropTag that = (DropTag) o;
+        return Objects.equals(tableName, that.tableName) &&
+                Objects.equals(tagName, that.tagName) &&
+                Objects.equals(tableExists, that.tableExists) &&
+                Objects.equals(tagExists, tagExists);
```
hantangwangd, line 97:
```suggestion
                Objects.equals(tagExists, that.tagExists);
```

presto-parser/src/main/java/com/facebook/presto/sql/tree/DropTag.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    @Override
+    public int hashCode()
+    {
+        return Objects.hash(tableName, tableExists, tagName);
```
hantangwangd, line 103: Missed `tagExists`?

presto-spi/src/main/java/com/facebook/presto/spi/connector/ConnectorMetadata.java
```diff
@@ -837,6 +837,22 @@ default TableLayoutFilterCoverage getTableLayoutFilterCoverage(ConnectorTableLay
         return NOT_APPLICABLE;
     }
 
+    /**
+     * Drop the specified branch
+     */
+    default void dropBranch(ConnectorSession session, ConnectorTableHandle tableHandle, Optional<String> branchName)
```
hantangwangd, line 843: Is it necessary to declare `branchName` as `Optional`?  Is it possible for it to be empty?

presto-spi/src/main/java/com/facebook/presto/spi/connector/ConnectorMetadata.java
```diff
@@ -837,6 +837,22 @@ default TableLayoutFilterCoverage getTableLayoutFilterCoverage(ConnectorTableLay
...
+    {
+        throw new PrestoException(NOT_SUPPORTED, "This connector does not support dropping table branches");
+    }
+
+    /**
+     * Drop the specified tag
+     */
+    default void dropTag(ConnectorSession session, ConnectorTableHandle tableHandle, Optional<String> tagName)
```
hantangwangd, line 851: The same as above, is it necessary to declare `tagName` as `Optional`?

presto-main/src/main/java/com/facebook/presto/execution/DropBranchTask.java
```diff
@@ -0,0 +1,77 @@
...
+                throw new SemanticException(NOT_SUPPORTED, statement, "'%s' is a materialized view, and drop branch is not supported", tableName);
+            }
+            return immediateFuture(null);
+        }
+
+        ConnectorId connectorId = metadata.getCatalogHandle(session, tableName.getCatalogName())
+                .orElseThrow(() -> new PrestoException(NOT_FOUND, "Catalog does not exist: " + tableName.getCatalogName()));
+        accessControl.checkCanDropBranch(session.getRequiredTransactionId(), session.getIdentity(), session.getAccessControlContext(), tableName);
+
+        metadata.dropBranch(session, tableHandleOptional.get(), Optional.of(statement.getBranchName().toString()));
```
hantangwangd, line 74: Seems we did not consider the `if exists` flag for `branch`. As I understand, we need to pass the flag value to `metadata.dropBranch(...)` and ultimately handle it in `ConnectorMetadata.dropBranch(...)`.

presto-main/src/main/java/com/facebook/presto/execution/DropTagTask.java
```diff
@@ -0,0 +1,77 @@
...
+                throw new SemanticException(NOT_SUPPORTED, statement, "'%s' is a materialized view, and drop tag is not supported", tableName);
+            }
+            return immediateFuture(null);
+        }
+
+        ConnectorId connectorId = metadata.getCatalogHandle(session, tableName.getCatalogName())
+                .orElseThrow(() -> new PrestoException(NOT_FOUND, "Catalog does not exist: " + tableName.getCatalogName()));
+        accessControl.checkCanDropTag(session.getRequiredTransactionId(), session.getIdentity(), session.getAccessControlContext(), tableName);
+
+        metadata.dropTag(session, tableHandleOptional.get(), Optional.of(statement.getTagName().toString()));
```
hantangwangd, line 74: The same as above, did not consider the `if exists` flag for `tag`.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
```diff
@@ -632,6 +633,34 @@ public void rollback()
...
+    {
+        IcebergTableHandle icebergTableHandle = (IcebergTableHandle) tableHandle;
+        verify(icebergTableHandle.getIcebergTableName().getTableType() == DATA, "only the data table can have branch dropped");
+        Table icebergTable = getIcebergTable(session, icebergTableHandle.getSchemaTableName());
+        if (icebergTable.refs().containsKey(branchName.get()) && icebergTable.refs().get(branchName.get()).isBranch()) {
+            icebergTable.manageSnapshots().removeBranch(branchName.get()).commit();
+        }
+        else {
+            throw new PrestoException(NOT_FOUND, format("Branch %s doesn't exist in table %s", branchName.get(), icebergTableHandle.getSchemaTableName().getTableName()));
+        }
```
hantangwangd, line 647: Should we take the value of `if exists` into consider here?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
```diff
@@ -632,6 +633,34 @@ public void rollback()
...
+    {
+        IcebergTableHandle icebergTableHandle = (IcebergTableHandle) tableHandle;
+        verify(icebergTableHandle.getIcebergTableName().getTableType() == DATA, "only the data table can have tag dropped");
+        Table icebergTable = getIcebergTable(session, icebergTableHandle.getSchemaTableName());
+        if (icebergTable.refs().containsKey(tagName.get()) && icebergTable.refs().get(tagName.get()).isTag()) {
+            icebergTable.manageSnapshots().removeTag(tagName.get()).commit();
+        }
+        else {
+            throw new PrestoException(NOT_FOUND, format("Tag %s doesn't exist in table %s", tagName.get(), icebergTableHandle.getSchemaTableName().getTableName()));
+        }
```
hantangwangd, line 661: The same as above, should we take the value of `if exists` into consider here?

presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedTestBase.java
```diff
@@ -1622,6 +1622,60 @@ public void testDecimal(boolean decimalVectorReaderEnabled)
...
+        assertQuery("SELECT count(*) FROM test_table_branch FOR SYSTEM_VERSION AS OF 'testBranch1'", "VALUES 2");
+        assertQuery("SELECT count(*) FROM test_table_branch FOR SYSTEM_VERSION AS OF 'testBranch2'", "VALUES 4");
+        assertQuery("SELECT count(*) FROM test_table_branch FOR SYSTEM_VERSION AS OF 'main'", "VALUES 6");
+
+        assertQuerySucceeds("ALTER TABLE test_table_branch DROP BRANCH 'testBranch1'");
+        icebergTable = loadTable("test_table_branch");
+        assertEquals(icebergTable.refs().size(), 2);
+        assertQueryFails("ALTER TABLE test_table_branch DROP BRANCH 'testBranchNotExist'", "Branch testBranchNotExist doesn't exist in table test_table_branch");
+        assertQuerySucceeds("ALTER TABLE test_table_branch DROP BRANCH IF EXISTS 'testBranch2'");
+        assertQueryFails("ALTER TABLE test_table_branch DROP BRANCH IF EXISTS 'testBranchNotExist'", "Branch testBranchNotExist doesn't exist in table test_table_branch");
```
hantangwangd, line 1648: Is this a correct behavior? As I understand, this statement should not fail.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedTestBase.java
```diff
@@ -1622,6 +1622,60 @@ public void testDecimal(boolean decimalVectorReaderEnabled)
...
+        assertQuery("SELECT count(*) FROM test_table_tag FOR SYSTEM_VERSION AS OF 'testTag1'", "VALUES 2");
+        assertQuery("SELECT count(*) FROM test_table_tag FOR SYSTEM_VERSION AS OF 'testTag2'", "VALUES 4");
+        assertQuery("SELECT count(*) FROM test_table_tag FOR SYSTEM_VERSION AS OF 'main'", "VALUES 6");
+
+        assertQuerySucceeds("ALTER TABLE test_table_tag DROP TAG 'testTag1'");
+        icebergTable = loadTable("test_table_tag");
+        assertEquals(icebergTable.refs().size(), 2);
+        assertQueryFails("ALTER TABLE test_table_tag DROP TAG 'testTagNotExist'", "Tag testTagNotExist doesn't exist in table test_table_tag");
+        assertQuerySucceeds("ALTER TABLE test_table_tag DROP TAG IF EXISTS 'testTag2'");
+        assertQueryFails("ALTER TABLE test_table_tag DROP TAG IF EXISTS 'testTagNotExist'", "Tag testTagNotExist doesn't exist in table test_table_tag");
```
hantangwangd, line 1675: The same as above, should this statement fail?

hantangwangd (changes requested):
The whole change overall looks good to me. Some little nits, and one problem for discussing about the behavior of `if exists` on `branch` and `tag`.

presto-main/src/main/java/com/facebook/presto/execution/DropBranchTask.java
agrawalreetika: Ohh Good catch, looks like I missed this. And this should not eventually handle the test case when `IF EXISTS` is supplied in the SQL. Thanks.

agrawalreetika:
@hantangwangd Thanks for you review. I have addressed your comments, please review.

hantangwangd (dismissed):
Thanks for the fix. LGTM!

ZacBlanco (dismissed):

agrawalreetika:
Hi @tdcmeehan, This is now waiting for the final committer review. Please review whenever you have time.

presto-main/src/main/java/com/facebook/presto/execution/DropBranchTask.java
```diff
@@ -0,0 +1,77 @@
...
+        if (!tableHandleOptional.isPresent()) {
+            if (!statement.isTableExists()) {
+                throw new SemanticException(MISSING_TABLE, statement, "Table '%s' does not exist", tableName);
+            }
+            return immediateFuture(null);
+        }
+
+        Optional<MaterializedViewDefinition> optionalMaterializedView = metadata.getMetadataResolver(session).getMaterializedView(tableName);
+        if (optionalMaterializedView.isPresent()) {
+            if (!statement.isTableExists()) {
```
pratyakshsharma, line 64: should we throw an error in cases when `if exists` is present in the query and it is a materialized view? If drop branch is just not supported for MVs, ideally error should be thrown irrespective of exists check?

presto-main/src/main/java/com/facebook/presto/execution/DropBranchTask.java
```diff
@@ -0,0 +1,77 @@
...
+                throw new SemanticException(NOT_SUPPORTED, statement, "'%s' is a materialized view, and drop branch is not supported", tableName);
+            }
+            return immediateFuture(null);
+        }
+
+        ConnectorId connectorId = metadata.getCatalogHandle(session, tableName.getCatalogName())
```
pratyakshsharma, line 70: nit: the variable connectorId is not getting used and can be removed. Also I have refactored and extracted this piece of code in MetadataUtil class in this commit - https://github.com/prestodb/presto/pull/21495/commits/dec29526b8ac710aded57e6d82211a841d7006ac. 
So we can probably do a similar change here as well so that all the Task classes use the same methods?i

presto-main/src/main/java/com/facebook/presto/execution/DropTagTask.java
```diff
@@ -0,0 +1,77 @@
...
+
+        Optional<MaterializedViewDefinition> optionalMaterializedView = metadata.getMetadataResolver(session).getMaterializedView(tableName);
+        if (optionalMaterializedView.isPresent()) {
+            if (!statement.isTableExists()) {
+                throw new SemanticException(NOT_SUPPORTED, statement, "'%s' is a materialized view, and drop tag is not supported", tableName);
+            }
+            return immediateFuture(null);
+        }
+
+        ConnectorId connectorId = metadata.getCatalogHandle(session, tableName.getCatalogName())
```
pratyakshsharma, line 70: nit: same as above.

presto-main/src/main/java/com/facebook/presto/execution/DropTagTask.java
```diff
@@ -0,0 +1,77 @@
...
+        if (!tableHandleOptional.isPresent()) {
+            if (!statement.isTableExists()) {
+                throw new SemanticException(MISSING_TABLE, statement, "Table '%s' does not exist", tableName);
+            }
+            return immediateFuture(null);
+        }
+
+        Optional<MaterializedViewDefinition> optionalMaterializedView = metadata.getMetadataResolver(session).getMaterializedView(tableName);
+        if (optionalMaterializedView.isPresent()) {
+            if (!statement.isTableExists()) {
```
pratyakshsharma, line 64: nit: same as above. What about the case when `if exists` is present and it is MV?

presto-main/src/main/java/com/facebook/presto/execution/DropBranchTask.java
pratyakshsharma: Alternate way of handling this can be introducing an enum Type in DropBranch and have only TABLE as the supported type.

pratyakshsharma (commented):
Thank you for the PR, minor comments, and rest looks good.

steveburnett:
Thanks for the release note! Nit on phrasing to follow the [Order of changes](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines#order-of-changes) in the Release Notes Guidelines.

```
== RELEASE NOTES ==

General Changes
* Add DDL support for dropping a branch from a table. 
* Add DDL support for dropping a tag from a table.

Iceberg Connector Changes
* Add support for dropping a branch from an Iceberg table.
* Add support for dropping a branch from an Iceberg table.
```

tdcmeehan:
I'm comfortable with the idea of this change.  While we should be cautious around the standard eventually adopting this type of syntax, the pace at which is the standard moves is far slower than even the Iceberg spec moves.  It is too slow for us to block progress on this PR.

I agree with the idea of customizing the parser and planning in general though, along the same lines as Spark.  However, given we have already broken standards compliance with time travel (on `SYSTEM_VERSION`), I'm comfortable pushing ahead with this.

tdcmeehan (approved):

PR #23617, opened 2024-09-10
Modify CREATE TABLE sql in iceberg tests

agrawalreetika:
## Description
Modify CREATE TABLE sql in iceberg tests

https://github.com/prestodb/presto/actions/runs/10786215241/attempts/3?pr=23584

## Motivation and Context
https://github.com/prestodb/presto/actions/runs/10786215241/attempts/3?pr=23584

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-iceberg/src/test/java/com/facebook/presto/iceberg/procedure/TestExpireSnapshotProcedure.java
+++ b/presto-iceberg/src/test/java/com/facebook/presto/iceberg/procedure/TestExpireSnapshotProcedure.java
@@ -79,7 +79,7 @@ public class TestExpireSnapshotProcedure
     public void testExpireSnapshotsInEmptyTable()
     {
         String tableName = "default_empty_table";
-        assertUpdate("CREATE TABLE " + tableName + " (id integer, value integer)");
+        assertUpdate("CREATE TABLE IF NOT EXISTS " + tableName + " (id integer, value integer)");
         Table table = loadTable(tableName);
         assertHasSize(table.snapshots(), 0);
 
@@ -106,7 +106,7 @@ public class TestExpireSnapshotProcedure
         Session session = sessionForTimezone(zoneId, legacyTimestamp);
         String tableName = "positional_args_table";
         try {
-            assertUpdate(session, "CREATE TABLE " + tableName + " (id integer, value varchar)");
+            assertUpdate(session, "CREATE TABLE IF NOT EXISTS " + tableName + " (id integer, value varchar)");
             assertUpdate(session, "INSERT INTO " + tableName + " VALUES(1, 'a')", 1);
 
             Table table = loadTable(tableName);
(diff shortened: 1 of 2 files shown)


presto-iceberg/src/test/java/com/facebook/presto/iceberg/procedure/TestRollbackToTimestampProcedure.java
```diff
@@ -71,7 +71,7 @@ public Object[][] timezones()
 
     private void createTable()
     {
-        assertUpdate("CREATE TABLE " + ICEBERG_CATALOG + "." + TEST_SCHEMA + "." + TEST_TABLE + " (id integer, value varchar)");
+        assertUpdate("CREATE TABLE IF NOT EXISTS " + ICEBERG_CATALOG + "." + TEST_SCHEMA + "." + TEST_TABLE + " (id integer, value varchar)");
```
hantangwangd, line 74: I think the core reason for this issue is that we operated on tables with the same table name in a non single threaded environment. So it seems that we cannot solve the issue in this way, perhaps we should either mark the test class as single threaded or make sure that all concurrently executed test cases use different table names.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/procedure/TestExpireSnapshotProcedure.java
```diff
@@ -200,7 +200,7 @@ public void testExpireSnapshotUsingNamedArgsOfSnapshotIds()
         Session session = getSession();
         String tableName = "named_args_snapshot_ids_table";
         try {
-            assertUpdate(session, "CREATE TABLE " + tableName + " (id integer, data varchar)");
+            assertUpdate(session, "CREATE TABLE IF NOT EXISTS " + tableName + " (id integer, data varchar)");
```
hantangwangd, line 203: Do we need these changes in `TestExpireSnapshotProcedure`? All tests in this class operate on tables with different table names, so it seems unlikely to meet the same failures as `TestRollbackToTimestampProcedure`. Any misunderstanding or overlook please let me know.

hantangwangd (changes requested):

presto-iceberg/src/test/java/com/facebook/presto/iceberg/procedure/TestRollbackToTimestampProcedure.java
agrawalreetika: > I think the core reason for this issue is that we operated on tables with the same table name in a non single threaded environment.
I missed the point about we are using same table name in this particular test. Would it be better to use different table names instead like other tests instead of making class single threaded?

presto-iceberg/src/test/java/com/facebook/presto/iceberg/procedure/TestExpireSnapshotProcedure.java
agrawalreetika: Apologies for the confusion, this is not related to failure. Just made this change to made consistency in the tests

presto-iceberg/src/test/java/com/facebook/presto/iceberg/procedure/TestRollbackToTimestampProcedure.java
hantangwangd: > Would it be better to use different table names instead like other tests instead of making class single threaded?

Agree, seems this would be better.

hantangwangd (approved):
Thanks for the fix, lgtm.

yingsu00 (approved):

PR #23622, opened 2024-09-11
[native] Fix recording of SUM type in PrometheusStatsReporter

lingbin:
For `SUM` type worker metrics, the corresponding type in 
`PrometheusStatsReporter` is `prometheus::Gauge`.

For these metrics, each time they are recorded (via 
`RECORD_METRIC_VALUE`), a "delta" is passed in, so `Gauge::Increment()` 
should be used in `PrometheusStatsReporter` instead of `Gauge::Set()`
(which overwrites the old value).

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/runtime-metrics/PrometheusStatsReporter.cpp
+++ b/presto-native-execution/presto_cpp/main/runtime-metrics/PrometheusStatsReporter.cpp
@@ -169,17 +169,23 @@ void PrometheusStatsReporter::addMetricValue(const char* key, size_t value)
   auto statsInfo = metricIterator->second;
   switch (statsInfo.statType) {
     case velox::StatType::COUNT: {
-      auto counter =
+      auto* counter =
           reinterpret_cast<::prometheus::Counter*>(statsInfo.metricPtr);
-      counter->Increment(value);
-    } break;
-    case velox::StatType::SUM:
+      counter->Increment(static_cast<double>(value));
+      break;
+    }
+    case velox::StatType::SUM: {
+      auto* gauge = reinterpret_cast<::prometheus::Gauge*>(statsInfo.metricPtr);
+      gauge->Increment(static_cast<double>(value));
+      break;
+    }
     case velox::StatType::AVG:
     case velox::StatType::RATE: {
       // Overrides the existing state.
-      auto gauge = reinterpret_cast<::prometheus::Gauge*>(statsInfo.metricPtr);
-      gauge->Set(value);
-    } break;
+      auto* gauge = reinterpret_cast<::prometheus::Gauge*>(statsInfo.metricPtr);
+      gauge->Set(static_cast<double>(value));
+      break;
+    }
     default:
       VELOX_UNSUPPORTED(
           "Unsupported metric type {}",
--- a/presto-native-execution/presto_cpp/main/runtime-metrics/tests/PrometheusReporterTest.cpp
+++ b/presto-native-execution/presto_cpp/main/runtime-metrics/tests/PrometheusReporterTest.cpp
@@ -22,6 +22,7 @@ class PrometheusReporterTest : public testing::Test {
   void SetUp() override {
     reporter = std::make_shared<PrometheusStatsReporter>(testLabels);
   }
+
   void verifySerializedResult(
       const std::string& fullSerializedResult,
       std::vector<std::string>& expected) {
@@ -32,6 +33,7 @@ class PrometheusReporterTest : public testing::Test {
       EXPECT_EQ(line, expected[i++]);
     }
   }
+
   const std::map<std::string, std::string> testLabels = {
       {"cluster", "test_cluster"},
       {"worker", "test_worker_pod"}};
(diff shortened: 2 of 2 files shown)


lingbin:
@majetideepak  Could you please help to take a look? Thanks.

majetideepak:
@lingbin thanks for the fix!
@karteekmurthys can you help review this?

aditi-pandit:
@lingbin : Thanks for this fix. Would it be possible to write a test for it ?

lingbin:
@aditi-pandit  A unit test has been added. And already rebased. Please take a look at it again, thanks.

presto-native-execution/presto_cpp/main/runtime-metrics/PrometheusStatsReporter.cpp
```diff
@@ -169,17 +169,23 @@ void PrometheusStatsReporter::addMetricValue(const char* key, size_t value)
...
           reinterpret_cast<::prometheus::Counter*>(statsInfo.metricPtr);
-      counter->Increment(value);
-    } break;
-    case velox::StatType::SUM:
+      counter->Increment(static_cast<double>(value));
+      break;
+    }
+    case velox::StatType::SUM: {
+      auto* gauge = reinterpret_cast<::prometheus::Gauge*>(statsInfo.metricPtr);
+      gauge->Increment(static_cast<double>(value));
```
karteekmurthys, line 179: The intent is to capture the change in value since the last time it was recorded. You can compute the cumulative sum in the visualization tool like Grafana. If we increment the value, we are as good as COUNT type. Please refer this doc that explains SUM capture Delta vs Cumulative SUM: https://opentelemetry.io/docs/specs/otel/metrics/data-model/#sums

We are capturing the delta and we don't want to increment here.

lingbin: @karteekmurthys  
Thank you for your explanation, I benefited a lot.

However, for Prometheus, I feel that when we use the "**pull mode**"(through the `fetchMetrics()` interface), it is impossible to accurately express the semantics of "**SUM capture Delta**".

I can give an example. If a metric is updated every `1` second, but we pull the full amount of metrics every `5` seconds,  if we use "SUM capture Delta" in Prometheus, we will lose information.  
For simplicity, suppose the delta value generated every second is `2`, 
```
  t0   t1    t2    t3    t4    t5    ...
  2,    2,    2,    2,    2,    2    ...
  |                             |
  v                             v
pull_1                        pull_2
```

As a result, only 2 points `(t0, 2)` and `(t5, 2)` are saved to Prometheus, that is, `t1`, `t2`, `t3`, `t4` are lost, so we are NOT able to compute the "cumulative sum" anymore.  

So, in my opinion, in Prometheus's "pull mode", we can only use the "Cumulative SUM" method.

Maybe I missed something, please correct me if I am wrong.

karteekmurthys: That is a good point @lingbin. We can minimize the data loss by setting you scrape interval by prometheus to 1s or 2s. Of course, this doesn't guarantee no data loss.

karteekmurthys: The scrape interval must be aware of your system as well. In our case, it is presto, which is meant to run analytic workloads where a single query can run for several seconds/minutes. So scraping values every 2/3s should give a decent picture of what is happening.

lingbin: @karteekmurthys  Thanks for your further explanation.

Some metrics are updated much more frequently than the "scrape interval", right? For example, they are updated many times per second (depending on the number of events).  I think there must be such metrics for those metrics that are not updated periodically in `PeriodicStatsReporter`.

Even more unfortunately, for those metrics with longer update periods, **reducing the "scrape interval" to a very small value will only give more wrong results**.

For example, let's assume a scenario where we scrape every `1` second, but the metric is updated every `5` seconds.

```
Time:              0 | 1 | 2 | 3 | 4 | 5 | 6 ...
Event:             2 | - | - | - | 5 | - | - ...
Cumulative:        2 | 2 | 2 | 2 | 7 | 7 | 7 ...
scraped-value:     2 | 2 | 2 | 2 | 5 | 5 | 5 ...

In this case, we can no longer even calculate the "Cumulative SUM" correctly any more 
based on the values ​​scraped, and still cannot accurately express "dalta". 

- At time t6, we expect the SUM to be 7 (2+5), but we get 23(2+2+2+2+5+5+5).  
- At time t2, t3, t4, we expect the DELTA to be 0, but we get 2.

If we see such a chart in the monitoring system, it will definitely be very misleading.
```

We expect to use Prometheus as a monitoring tool for online production systems, but the serious error mentioned above determines that we can only use the "**Cumulative SUM**" method.

lingbin: After using the "Cumulative SUM" method, if we need to calculate "delta", we can use [the `rate()` function](https://prometheus.io/docs/prometheus/latest/querying/functions/#rate)  provided by Prometheus, which calculates the **per-second** average rate of increase of the time series.  
Yes, it really cannot express the delta since the last time it was recorded.

karteekmurthys (approved):

majetideepak (approved):

aditi-pandit (approved):
Thanks @lingbin

PR #23627, opened 2024-09-11
[iceberg] Add UUID type support

ZacBlanco:
## Description

This PR adds support for reading and writing UUIDs on Iceberg tables with all available catalogs. In order to support this we also needed improvements to the parquet reader and writer for [Parquet's UUID logical type](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#uuid).

## Motivation and Context

Presto has type support for UUIDs. We should support reading and writing them in some of the connectors.

## Impact

- Iceberg tables can now be created with UUID types.

## Test Plan

-  Basic tests inside of the Iceberg module for round-trip UUID reading and writing.
- Additional tests in the parquet module for reading and writing UUID values

I also added a benchmark for reading and writing UUID types and compared it to our current LongDecimal benchmark to see the performance difference for another type which uses `FIXED_LENGTH_BYTE_ARRAY` as the underlying physical type.

These were the microbenchmarks from my local machine on ARM using a build of Corretto JDK11 and with reader verification disabled

```
Benchmark                       (batchReaderEnabled)  (parquetEncoding)   Mode  Cnt    Score   Error  Units
BenchmarkUuidColumnReader.read                  true              PLAIN  thrpt   20  265.542 ± 3.891  ops/s
BenchmarkUuidColumnReader.read                  true   DELTA_BYTE_ARRAY  thrpt   20   35.511 ± 0.598  ops/s
BenchmarkUuidColumnReader.read                 false              PLAIN  thrpt   20   27.316 ± 1.222  ops/s
BenchmarkUuidColumnReader.read                 false   DELTA_BYTE_ARRAY  thrpt   20   20.883 ± 0.521  ops/s


Benchmark                              (enableOptimizedReader)  (parquetEncoding)   Mode  Cnt   Score   Error  Units
BenchmarkLongDecimalColumnReader.read                     true              PLAIN  thrpt   20  10.926 ± 0.178  ops/s
BenchmarkLongDecimalColumnReader.read                     true   DELTA_BYTE_ARRAY  thrpt   20   8.948 ± 0.180  ops/s
BenchmarkLongDecimalColumnReader.read                    false              PLAIN  thrpt   20   8.632 ± 0.129  ops/s
BenchmarkLongDecimalColumnReader.read                    false   DELTA_BYTE_ARRAY  thrpt   20   7.771 ± 0.066  ops/s
```

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Add UUID type support to the Parquet reader and writer. :pr:`23627`

Iceberg Connector Changes
* Add support of UUID-typed columns :pr:`23627`
```

### DIFF
--- a/presto-hive-metastore/src/main/java/com/facebook/presto/hive/HiveType.java
+++ b/presto-hive-metastore/src/main/java/com/facebook/presto/hive/HiveType.java
@@ -20,6 +20,7 @@ import com.facebook.presto.common.type.Type;
 import com.facebook.presto.common.type.TypeManager;
 import com.facebook.presto.common.type.TypeSignature;
 import com.facebook.presto.common.type.TypeSignatureParameter;
+import com.facebook.presto.common.type.UuidType;
 import com.facebook.presto.spi.PrestoException;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
@@ -47,6 +48,7 @@ import static com.facebook.presto.common.type.DoubleType.DOUBLE;
 import static com.facebook.presto.common.type.IntegerType.INTEGER;
 import static com.facebook.presto.common.type.RealType.REAL;
 import static com.facebook.presto.common.type.SmallintType.SMALLINT;
+import static com.facebook.presto.common.type.StandardTypes.UUID;
 import static com.facebook.presto.common.type.TimestampType.TIMESTAMP;
 import static com.facebook.presto.common.type.TinyintType.TINYINT;
 import static com.facebook.presto.common.type.VarbinaryType.VARBINARY;
@@ -58,6 +60,7 @@ import static java.lang.String.format;
 import static java.util.Objects.requireNonNull;
 import static java.util.stream.Collectors.toList;
 import static org.apache.hadoop.hive.serde2.objectinspector.ObjectInspector.Category;
+import static org.apache.hadoop.hive.serde2.objectinspector.ObjectInspector.Category.PRIMITIVE;
 import static org.apache.hadoop.hive.serde2.typeinfo.TypeInfoFactory.binaryTypeInfo;
 import static org.apache.hadoop.hive.serde2.typeinfo.TypeInfoFactory.booleanTypeInfo;
 import static org.apache.hadoop.hive.serde2.typeinfo.TypeInfoFactory.byteTypeInfo;
@@ -87,6 +90,47 @@ public final class HiveType
     public static final HiveType HIVE_TIMESTAMP = new HiveType(timestampTypeInfo);
     public static final HiveType HIVE_DATE = new HiveType(dateTypeInfo);
     public static final HiveType HIVE_BINARY = new HiveType(binaryTypeInfo);
+    public static final HiveType HIVE_UUID = new HiveType(new TypeInfo()
+    {
+        @Override
+        public Category getCategory()
+        {
+            return PRIMITIVE;
+        }
+
+        @Override
+        public String getTypeName()
+        {
+            return UUID;
+        }
+
+        @Override
+        public boolean equals(Object other)
+        {
+            if (this == other) {
+                return true;
+            }
+            if (other == null || getClass() != other.getClass()) {
+                return false;
+            }
+
+            TypeInfo ti = (TypeInfo) other;
+
+            return UUID.equals(ti.getTypeName());
+        }
+
+        @Override
+        public int hashCode()
+        {
+            return UUID.hashCode();
+        }
+
+        @Override
+        public String toString()
+        {
+            return UUID;
+        }
+    });
 
     private final HiveTypeName hiveTypeName;
     private final TypeInfo typeInfo;
@@ -223,6 +267,9 @@ public final class HiveType
         switch (typeInfo.getCategory()) {
             case PRIMITIVE:
                 Type primitiveType = getPrimitiveType((PrimitiveTypeInfo) typeInfo);
+                if (primitiveType == null && typeInfo.getTypeName().equals(UUID)) {
+                    return UuidType.UUID.getTypeSignature();
+                }
                 if (primitiveType == null) {
                     break;
                 }
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/ExpressionConverter.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/ExpressionConverter.java
@@ -31,6 +31,7 @@ import com.facebook.presto.common.type.RowType;
 import com.facebook.presto.common.type.TimeType;
 import com.facebook.presto.common.type.TimestampType;
 import com.facebook.presto.common.type.Type;
+import com.facebook.presto.common.type.UuidType;
 import com.facebook.presto.common.type.VarbinaryType;
 import com.facebook.presto.common.type.VarcharType;
 import com.google.common.base.VerifyException;
@@ -41,6 +42,7 @@ import java.math.BigDecimal;
 import java.nio.ByteBuffer;
 import java.util.List;
 import java.util.Map;
+import java.util.UUID;
 
 import static com.facebook.presto.common.predicate.Marker.Bound.ABOVE;
 import static com.facebook.presto.common.predicate.Marker.Bound.BELOW;
@@ -220,6 +222,12 @@ public final class ExpressionConverter
             return new BigDecimal(Decimals.decodeUnscaledValue((Slice) value), decimalType.getScale());
         }
 
+        if (type instanceof UuidType) {
+            UuidType uuidType = (UuidType) type;
+            return marker.getValueBlock()
+                    .map(block -> UUID.fromString((String) uuidType.getObjectValue(null, block, 0)))
+                    .orElseThrow(NullPointerException::new);
+        }
         return marker.getValue();
     }
 }
--- a/presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedTestBase.java
+++ b/presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedTestBase.java
@@ -88,6 +88,7 @@ import java.io.Closeable;
 import java.io.File;
 import java.io.IOException;
 import java.lang.reflect.Field;
+import java.nio.ByteBuffer;
 import java.nio.file.Path;
 import java.time.LocalDateTime;
 import java.time.LocalTime;
@@ -95,6 +96,7 @@ import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
@@ -134,6 +136,7 @@ import static com.facebook.presto.testing.assertions.Assert.assertEquals;
 import static com.facebook.presto.tests.sql.TestTable.randomTableSuffix;
 import static java.lang.String.format;
 import static java.util.Objects.requireNonNull;
+import static java.util.UUID.randomUUID;
 import static org.apache.iceberg.SnapshotSummary.TOTAL_DATA_FILES_PROP;
 import static org.apache.iceberg.SnapshotSummary.TOTAL_DELETE_FILES_PROP;
 import static org.apache.parquet.column.ParquetProperties.WriterVersion.PARQUET_1_0;
@@ -1596,20 +1599,20 @@ public abstract class IcebergDistributedTestBase
         }
     }
 
-    @DataProvider(name = "decimalVectorReader")
-    public Object[] decimalVectorReader()
+    @DataProvider(name = "batchReadEnabled")
+    public Object[] batchReadEnabledReader()
     {
         return new Object[] {true, false};
     }
 
-    private Session decimalVectorReaderEnabledSession(boolean decimalVectorReaderEnabled)
+    private Session batchReadEnabledEnabledSession(boolean batchReadEnabled)
     {
         return Session.builder(getQueryRunner().getDefaultSession())
-                .setCatalogSessionProperty(ICEBERG_CATALOG, PARQUET_BATCH_READ_OPTIMIZATION_ENABLED, String.valueOf(decimalVectorReaderEnabled))
+                .setCatalogSessionProperty(ICEBERG_CATALOG, PARQUET_BATCH_READ_OPTIMIZATION_ENABLED, String.valueOf(batchReadEnabled))
                 .build();
     }
 
-    @Test(dataProvider = "decimalVectorReader")
+    @Test(dataProvider = "batchReadEnabled")
     public void testDecimal(boolean decimalVectorReaderEnabled)
     {
         String tableName = "test_decimal_vector_reader";
@@ -1625,7 +1628,7 @@ public abstract class IcebergDistributedTestBase
             // Insert data to table
             assertUpdate("INSERT INTO " + tableName + values, 4);
 
-            Session session = decimalVectorReaderEnabledSession(decimalVectorReaderEnabled);
+            Session session = batchReadEnabledEnabledSession(decimalVectorReaderEnabled);
             assertQuery(session, "SELECT * FROM " + tableName, values);
         }
         finally {
@@ -1701,6 +1704,7 @@ public abstract class IcebergDistributedTestBase
                     "   c_timestamp TIMESTAMP, " +
                     "   c_varchar VARCHAR, " +
                     "   c_varbinary VARBINARY, " +
+                    "   c_uuid UUID, " +
                     "   c_array ARRAY(BIGINT), " +
                     "   c_map MAP(VARCHAR, INT), " +
                     "   c_row ROW(a INT, b VARCHAR) " +
@@ -1708,16 +1712,16 @@ public abstract class IcebergDistributedTestBase
 
             assertUpdate(format("" +
                     "INSERT INTO %s " +
-                    "SELECT c_boolean, c_int, c_bigint, c_double, c_real, c_date, c_timestamp, c_varchar, c_varbinary, c_array, c_map, c_row " +
+                    "SELECT c_boolean, c_int, c_bigint, c_double, c_real, c_date, c_timestamp, c_varchar, c_varbinary, c_uuid, c_array, c_map, c_row " +
                     "FROM ( " +
                     "  VALUES " +
-                    "    (null, null, null, null, null, null, null, null, null, null, null, null), " +
-                    "    (true, INT '1245', BIGINT '1', DOUBLE '2.2', REAL '-24.124', DATE '2024-07-29', TIMESTAMP '2012-08-08 01:00', CAST('abc1' AS VARCHAR), to_ieee754_64(1), sequence(0, 10), MAP(ARRAY['aaa', 'bbbb'], ARRAY[1, 2]), CAST(ROW(1, 'AAA') AS ROW(a INT, b VARCHAR)))," +
-                    "    (false, INT '-1245', BIGINT '-1', DOUBLE '2.3', REAL '243215.435', DATE '2024-07-29', TIMESTAMP '2012-09-09 00:00', CAST('cba2' AS VARCHAR), to_ieee754_64(4), sequence(30, 35), MAP(ARRAY['ccc', 'bbbb'], ARRAY[-1, -2]), CAST(ROW(-1, 'AAA') AS ROW(a INT, b VARCHAR))) " +
-                    ") AS x (c_boolean, c_int, c_bigint, c_double, c_real, c_date, c_timestamp, c_varchar, c_varbinary, c_array, c_map, c_row)", tmpTableName), 3);
+                    "    (null, null, null, null, null, null, null, null, null, null, null, null, null), " +
+                    "    (true, INT '1245', BIGINT '1', DOUBLE '2.2', REAL '-24.124', DATE '2024-07-29', TIMESTAMP '2012-08-08 01:00', CAST('abc1' AS VARCHAR), to_ieee754_64(1), CAST('4ae71336-e44b-39bf-b9d2-752e234818a5' as UUID), sequence(0, 10), MAP(ARRAY['aaa', 'bbbb'], ARRAY[1, 2]), CAST(ROW(1, 'AAA') AS ROW(a INT, b VARCHAR)))," +
+                    "    (false, INT '-1245', BIGINT '-1', DOUBLE '2.3', REAL '243215.435', DATE '2024-07-29', TIMESTAMP '2012-09-09 00:00', CAST('cba2' AS VARCHAR), to_ieee754_64(4), CAST('4ae71336-e44b-39bf-b9d2-752e234818a5' as UUID), sequence(30, 35), MAP(ARRAY['ccc', 'bbbb'], ARRAY[-1, -2]), CAST(ROW(-1, 'AAA') AS ROW(a INT, b VARCHAR))) " +
+                    ") AS x (c_boolean, c_int, c_bigint, c_double, c_real, c_date, c_timestamp, c_varchar, c_varbinary, c_uuid, c_array, c_map, c_row)", tmpTableName), 3);
 
-            Session decimalVectorReaderEnabled = decimalVectorReaderEnabledSession(true);
-            Session decimalVectorReaderDisable = decimalVectorReaderEnabledSession(false);
+            Session decimalVectorReaderEnabled = batchReadEnabledEnabledSession(true);
+            Session decimalVectorReaderDisable = batchReadEnabledEnabledSession(false);
             assertQueryWithSameQueryRunner(decimalVectorReaderEnabled, "SELECT * FROM " + tmpTableName, decimalVectorReaderDisable);
         }
         finally {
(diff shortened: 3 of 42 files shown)


steveburnett:
Nit suggestion for the release note entry to follow the [Order of changes](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines#order-of-changes) in the Release Notes Guidelines: 

```
== RELEASE NOTES ==

General Changes
* Add UUID type support to the Parquet reader and writer. :pr:`23627`

Iceberg Connector Changes
* Add support of UUID-typed columns :pr:`23627`
```

ZacBlanco:
Fixed, thanks Steve!

presto-common/src/main/java/com/facebook/presto/common/type/Uuids.java
```diff
@@ -0,0 +1,43 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.common.type;
+
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+
+public class Uuids
```
elharo, line 19: Do we as a project have a strong opinion about UUID vs Uuid in class and method names? That is, do we capitalize acronyms like the JDK (URL) or like Google (Url)?

presto-common/src/main/java/com/facebook/presto/common/type/Uuids.java
```diff
@@ -0,0 +1,43 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.common.type;
+
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+
```
elharo, line 18: please add javadoc

presto-common/src/main/java/com/facebook/presto/common/type/Uuids.java
```diff
@@ -0,0 +1,43 @@
...
+public class Uuids
+{
+    private Uuids() {}
+
+    public static long[] rawUuidValuesFromBigEndian(byte[] bytes)
```
elharo, line 23: API comment

elharo, line 23: Returning two longs is weird compared to a byte array. It's a bit of an accident that a UUID has the same number of bytes as two longs, and most importantly these aren't really numbers. Doing arithmetic on them means nothing.

presto-common/src/main/java/com/facebook/presto/common/type/Uuids.java
```diff
@@ -0,0 +1,43 @@
...
+    {
+        long[] result = new long[2];
+        rawUuidValuesFromBigEndian(result, bytes, 0);
+        return result;
+    }
+
+    public static void rawUuidValuesFromBigEndian(long[] buffer, byte[] bytes, int dstIndex)
```
elharo, line 30: why not return the bytes?

presto-common/src/main/java/com/facebook/presto/common/type/Uuids.java
```diff
@@ -0,0 +1,43 @@
...
+    {
+        rawUuidValuesFromBigEndian(buffer, bytes, 0, dstIndex);
+    }
+
+    public static void rawUuidValuesFromBigEndian(long[] buffer, byte[] bytes, int srcIndex, int dstIndex)
```
elharo, line 35: API comment

presto-hive-metastore/src/main/java/com/facebook/presto/hive/HiveType.java
```diff
@@ -87,6 +90,50 @@ public final class HiveType
...
+            return UUID.equals(ti.getTypeName());
+        }
+
+        /**
+         * Generate the hashCode for this TypeInfo.
+         */
+        @Override
+        public int hashCode()
+        {
+            return UUID.hashCode();
```
elharo, line 128: so this is a singleton? If so the equals method can be simpler. Otherwise this hashCode method is incorrect

presto-iceberg/src/main/java/com/facebook/presto/iceberg/ExpressionConverter.java
```diff
@@ -220,6 +221,12 @@ private static Object getIcebergLiteralValue(Type type, Marker marker)
             return new BigDecimal(Decimals.decodeUnscaledValue((Slice) value), decimalType.getScale());
         }
 
+        if (type instanceof UuidType) {
+            UuidType uuidType = (UuidType) type;
+            return marker.getValueBlock()
+                    .map(block -> uuidType.getObjectValue(null, block, 0))
+                    .orElseThrow(() -> new IllegalArgumentException("UUID value block is null"));
```
elharo, line 228: NullPointerException?

presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedTestBase.java
```diff
@@ -1886,6 +1888,26 @@ public void testStatisticsFileCache()
         getQueryRunner().execute("DROP TABLE test_statistics_file_cache");
     }
 
+    @Test
+    public void testUuidRoundTrip()
+    {
+        assertQuerySucceeds("CREATE TABLE uuid_roundtrip(u uuid)");
+        UUID uuid = randomUUID();
```
elharo, line 1895: don't use random values in unit tests

presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedTestBase.java
```diff
@@ -1886,6 +1888,26 @@ public void testStatisticsFileCache()
...
+        assertUpdate(format("INSERT INTO uuid_roundtrip VALUES CAST('%s' as uuid)", uuid), 1);
+        assertQuery("SELECT CAST(u as varchar) FROM uuid_roundtrip", format("VALUES '%s'", uuid));
+    }
+
+    @Test
+    public void testUuidFilters()
+    {
+        assertQuerySucceeds("CREATE TABLE uuid_filters(u uuid)");
+        List<UUID> uuids = IntStream.range(0, 100).mapToObj(unused -> randomUUID())
```
elharo, line 1904: no random so the test is reproducible

presto-parquet/src/test/java/com/facebook/presto/parquet/batchreader/decoders/TestParquetUtils.java
```diff
@@ -144,6 +144,16 @@ public static byte[] generatePlainValuesPage(int valueCount, int valueSizeBits,
                 }
                 break;
             }
+            case 128:
+                for (int i = 0; i < valueCount; i++) {
+                    long value = random.nextLong();
+                    writer.writeLong(value);
+                    addedValues.add(value);
+                    value = random.nextLong();
```
elharo, line 152: no random values please; pick an arbitrary one

elharo (changes requested):

presto-common/src/main/java/com/facebook/presto/common/type/Uuids.java
ZacBlanco: The presto implementation of the UUID type specifies that a [UUID is represented by two long values](https://github.com/prestodb/presto/blob/2f4ea64d3bf3fa8c89b3e805d5d0987ac6d971c2/presto-common/src/main/java/com/facebook/presto/common/type/UuidType.java#L156-L158). parquet's spec imposes that a UUID is a 16-byte long array. We have to convert between storage and in-memory format in order to execute on the data properly. There is no way around this.

presto-common/src/main/java/com/facebook/presto/common/type/Uuids.java
ZacBlanco: This is to prevent memory allocations so that a single buffer may be re-used for multiple invocations of this method.

presto-common/src/main/java/com/facebook/presto/common/type/Uuids.java
ZacBlanco: I found within the Presto codebase we don't use full capitalization in most cases. Additionally we call the UUID type class `UuidType` so I opted for consistency with the existing naming.

presto-hive-metastore/src/main/java/com/facebook/presto/hive/HiveType.java
ZacBlanco: `TypeInfo` is a hive-specific class that I am implementing because it does not natively support UUIDs. This implementation follows the same primitive paradigm as the other `PrimitiveTypeInfo` which exist in the Apache Hive's implementation

https://github.com/apache/hive/blame/13ec7c3ab47001df25e7be87739731192075b0a7/serde/src/java/org/apache/hadoop/hive/serde2/typeinfo/PrimitiveTypeInfo.java#L93-L112

presto-parquet/src/test/java/com/facebook/presto/parquet/batchreader/decoders/TestParquetUtils.java
ZacBlanco: The entire test class revolves around generating random values for different primitive types. I don't think we should go against the grain here. Changing this entire class to not use random values is out of scope for this PR.

presto-hive-metastore/src/main/java/com/facebook/presto/hive/HiveType.java
elharo: If I'm reading this right the hashCode is a constant, which only makes sense if the object is a singleton. Am I missing something? (always possible)

elharo (changes requested):

presto-parquet/src/test/java/com/facebook/presto/parquet/batchreader/decoders/TestParquetUtils.java
elharo: I'm not asking for the entire class to change. I'm asking that this PR not make the problem worse. Random values in unit tests are a major source of flakiness. Please don't do this.

elharo (changes requested):

hantangwangd (commented):
Thanks for support `UUID` type in Iceberg connector. Should we add the mapping of `UUID` between PrestoDB type and Iceberg type to the chapter `Type mapping` in iceberg document?

steveburnett (approved):
LGTM! (docs)

presto-parquet/src/main/java/com/facebook/presto/parquet/cache/MetadataReader.java
```diff
@@ -373,7 +373,11 @@ private static void readTypeSchema(Types.GroupBuilder<?> builder, Iterator<Schem
                 }
                 typeBuilder = primitiveBuilder;
             }
-            if (element.isSetConverted_type()) {
+
+            if (element.isSetLogicalType()) {
+                getLogicalTypeAnnotationWithParameters(element.getLogicalType()).ifPresent(typeBuilder::as);
+            }
```
hantangwangd, line 379: nit: after this refactor, the `if(element.isSetLogicalType())` block below is redundant, we can remove it.

presto-parquet/src/main/java/com/facebook/presto/parquet/writer/valuewriter/UuidValuesWriter.java
```diff
@@ -0,0 +1,50 @@
...
+import org.apache.parquet.column.values.ValuesWriter;
+import org.apache.parquet.io.api.Binary;
+import org.apache.parquet.schema.PrimitiveType;
+
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+
+import static io.airlift.slice.SizeOf.SIZE_OF_LONG;
+
+public class UuidValuesWriter
```
hantangwangd, line 26: Should we mark this class as a `@NotThreadSafe`?

presto-parquet/src/main/java/com/facebook/presto/parquet/writer/valuewriter/UuidValuesWriter.java
```diff
@@ -0,0 +1,50 @@
...
+
+    @Override
+    public void write(Block block)
+    {
+        for (int i = 0; i < block.getPositionCount(); i++) {
+            if (!block.isNull(i)) {
+                writeBuffer.rewind();
+                writeBuffer.putLong(block.getLong(i, 0));
+                writeBuffer.putLong(block.getLong(i, SIZE_OF_LONG));
+                writeBuffer.rewind();
```
hantangwangd, line 45: Just bring for discussing, do you think it's better to use `writeBuffer.clear()` and `writeBuffer.flip()` here, although the effect seems to be the same in this scenario?

presto-parquet/src/main/java/com/facebook/presto/parquet/ColumnReaderFactory.java
```diff
@@ -96,6 +99,10 @@ public static ColumnReader createReader(RichColumnDescriptor descriptor, boolean
                     return isNested ? new BinaryNestedBatchReader(descriptor) : new BinaryFlatBatchReader(descriptor);
                 case FIXED_LEN_BYTE_ARRAY:
                     if (!isNested) {
+                        if (uuidType().equals(descriptor.getPrimitiveType().getLogicalTypeAnnotation())) {
```
hantangwangd, line 102: nit: we can use the newly added `ParquetTypeUtils.isUuidType()` here.

presto-parquet/src/main/java/com/facebook/presto/parquet/ColumnReaderFactory.java
```diff
@@ -126,10 +133,13 @@ public static ColumnReader createReader(RichColumnDescriptor descriptor, boolean
             case BINARY:
                 return createDecimalColumnReader(descriptor).orElse(new BinaryColumnReader(descriptor));
             case FIXED_LEN_BYTE_ARRAY:
+                if (uuidType().equals(descriptor.getPrimitiveType().getLogicalTypeAnnotation())) {
```
hantangwangd, line 136: The same as above, we can use `ParquetTypeUtils.isUuidType()`.

presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/decoders/plain/FixedLenByteArrayUuidPlainValuesDecoder.java
```diff
@@ -0,0 +1,74 @@
...
+    private final int typeLength;
+    private final byte[] inputBytes;
+    private final byte[] byteBuffer;
+    private final int bufferEnd;
+
+    private int bufferOffset;
+
+    public FixedLenByteArrayUuidPlainValuesDecoder(int typeLength, byte[] byteBuffer, int bufferOffset, int length)
+    {
+        checkArgument(typeLength > 0 && typeLength <= 16, "typeLength %s should be in range (1-16) for a UUID", typeLength);
```
hantangwangd, line 38: Could we just check `typeLength == 16` here?

presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/decoders/rle/UuidRLEDictionaryValuesDecoder.java
```diff
@@ -0,0 +1,63 @@
...
+        super(Integer.MAX_VALUE, bitWidth, inputStream);
+        requireNonNull(dictionary, "dictionary is null");
+        delegate = new BinaryRLEDictionaryValuesDecoder(bitWidth, inputStream, dictionary);
+    }
+
+    @Override
+    public void readNext(long[] values, int offset, int length)
+            throws IOException
+    {
+        BinaryValuesDecoder.ValueBuffer valueBuffer = delegate.readNext(length);
```
hantangwangd, line 42: nit: can use a static import here

presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/decoders/rle/UuidRLEDictionaryValuesDecoder.java
```diff
@@ -0,0 +1,63 @@
...
+        int bufferSize = valueBuffer.getBufferSize();
+        byte[] byteBuffer = new byte[bufferSize];
+        int[] offsets = new int[bufferSize + 1];
```
hantangwangd, line 45:
```suggestion
        int[] offsets = new int[length + 1];
```

I know you copy this code from other decoders like `LongDecimalRLEDictionaryValuesDecoder`. However,  I believe the length of offsets should be `length + 1`, and I will rise a PR to fix the other decoders. Any misunderstanding please let me know.

hantangwangd (commented):
The change overall looks good to me. Some nits and little problems.

presto-parquet/src/main/java/com/facebook/presto/parquet/writer/valuewriter/UuidValuesWriter.java
ZacBlanco: `clear()` and `flip()` are probably better. I think they make the intended use of the buffer more clear

hantangwangd (dismissed):

agrawalreetika (changes requested):

presto-parquet/src/main/java/com/facebook/presto/parquet/writer/valuewriter/UuidValuesWriter.java
```diff
@@ -0,0 +1,53 @@
...
+@NotThreadSafe
+public class UuidValuesWriter
+        extends PrimitiveValueWriter
+{
+    private final ByteBuffer writeBuffer = ByteBuffer.allocate(2 * SIZE_OF_LONG)
+            .order(ByteOrder.BIG_ENDIAN);
+
+    public UuidValuesWriter(ValuesWriter valuesWriter, PrimitiveType parquetType)
+    {
+        super(parquetType, valuesWriter);
```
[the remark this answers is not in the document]
ZacBlanco: Actually, I think the `BooleanValueWriter` is wrong because the superclass `PrimitivateValueWriter` already stores the reference to the `valuesWriter` which is accessed through `getValueWriter()`. Storing it again seems to just be a redundant reference.

https://github.com/prestodb/presto/blob/152e64dbebf2d8c9deedaae833d4ebfcaa4a531f/presto-parquet/src/main/java/com/facebook/presto/parquet/writer/valuewriter/PrimitiveValueWriter.java#L33-L43

In the UUID writer implementation, we just use `getValueWriter()`

presto-parquet/src/main/java/com/facebook/presto/parquet/writer/valuewriter/UuidValuesWriter.java
```diff
@@ -0,0 +1,53 @@
...
+    @Override
+    public void write(Block block)
+    {
+        for (int i = 0; i < block.getPositionCount(); i++) {
+            if (!block.isNull(i)) {
+                writeBuffer.clear();
+                writeBuffer.putLong(block.getLong(i, 0));
+                writeBuffer.putLong(block.getLong(i, SIZE_OF_LONG));
+                writeBuffer.flip();
+                getValueWriter().writeBytes(Binary.fromReusedByteBuffer(writeBuffer));
```
agrawalreetika, line 49: We don't have to update stats for this?

ZacBlanco: Fixed, good catch

ZacBlanco:
FYI @elharo since https://github.com/prestodb/presto/pull/23699 was merged I needed to re-add a profile which uses the previously removed maven plugin to generate the batch readers which is now reflected in the 1st commit of this PR

presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/BooleanFlatBatchReader.java
```diff
@@ -255,3 +255,4 @@ private void seek()
         }
     }
 }
+
```
elharo, line 258: adding these blank lines should be a separate PR, if it's needed at all. (If checkstyle doesn't complain it probably isn't)

elharo (changes requested):

presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/BooleanFlatBatchReader.java
ZacBlanco: They are added automatically by the templating tool which generates the files, but weren't caught before because they existed in `generated-sources`. I tried a large combination of [whitespace-control directives from freemarker](https://freemarker.apache.org/docs/dgui_misc_whitespace.html) but was unsuccessful in removing them. I can split these changes into a separate commit on this PR. The commit will be preserved when this PR merges

hantangwangd (dismissed):

agrawalreetika (dismissed):

presto-parquet/src/main/java/com/facebook/presto/parquet/reader/BinaryColumnReader.java
```diff
@@ -52,6 +55,9 @@ protected void readValue(BlockBuilder blockBuilder, Type type)
             if (isCharType(type)) {
                 value = truncateToLengthAndTrimSpaces(value, type);
             }
+            if (type instanceof UuidType) {
+                value = wrappedLongArray(rawUuidValuesFromBigEndian(binary.getBytes()));
```
yingsu00, line 59: There are a few opportunities to improve the perf and reliability here:
1. The memory for `value` was pre-allocated above on line 50 `value = wrappedBuffer(binary.getBytes());`. In this line `binary.getBytes()` allocates the memory and copy the data once already. But your `rawUuidValuesFromBigEndian` was not making use of the memory in `binary` object, but instead, it re-allocates two longs inside. Note that for char and varchar types, the memory was re-used and no new allocation was done. For UUID type, the output value is always a fixed length of 2 longs.  This is different than char and varchar types because the output length can only be known at run time for these two types. So for UUID, we can at least avoid the double allocation. I would suggest handle UUID separately and do not allocate the `Slice value` above.
2. It is not good and not necessary to wrap the 2 longs as a Slice at all. `wrappedLongArray` allocates a new Slice object. Although the base array was not copied, new memory needs to be allocate to hold the Slice object itself. Then `type.writeSlice(blockBuilder, value);` peels the Slice just to get the 2 longs we just wrapped up. I've seen many cases where allocating so many Slices objects caused GC pressure and perf degradation. It would be better to just use `writeLong` twice here. I know this is a pre-existing design problem, and would recommend fix this in a separate commit.

presto-parquet/.gitignore
```diff
@@ -0,0 +1,2 @@
+src/main/java/ParquetFlatColumnReaderGenerator.tdd
```
yingsu00, line 1: Can you please explain why we need to re-generate the readers? It'll be nice to add a commit message to explain it.

yingsu00 (commented):
Haven't finished reviewing, will continue later tonight.

presto-parquet/.gitignore
ZacBlanco: Previously during the build of this module we would always generate them. A recent PR committed the generated files (#23699) by copying them into the main source tree but removed the code generation mechanism. This PR made use of the generation mechanism to create the batch reader for the UUID type. I also think it's fine to keep the generation disabled behind a profile if we want to keep the files checked into the source tree. But we should entirely remove the mechanism which created the files in the first place. 

I will add a comment to the commit message to clarify this.

presto-parquet/src/main/java/com/facebook/presto/parquet/reader/BinaryColumnReader.java
ZacBlanco: Thanks for your suggestions. I tried to reduce the allocations and copying by writing directly to the block builder as I pulled values from the buffer. It increased the performance a measurable amount for the non-batchreader implementations

old perf results
```
Benchmark                       (enableOptimizedReader)  (parquetEncoding)   Mode  Cnt    Score   Error  Units
BenchmarkUuidColumnReader.read                     true              PLAIN  thrpt   10  158.193 ± 1.058  ops/s
BenchmarkUuidColumnReader.read                     true   DELTA_BYTE_ARRAY  thrpt   10   32.211 ± 0.345  ops/s
BenchmarkUuidColumnReader.read                    false              PLAIN  thrpt   10   13.030 ± 0.463  ops/s
BenchmarkUuidColumnReader.read                    false   DELTA_BYTE_ARRAY  thrpt   10   10.121 ± 0.377  ops/s
```

new perf results
```
Benchmark                       (enableOptimizedReader)  (parquetEncoding)   Mode  Cnt    Score   Error  Units
BenchmarkUuidColumnReader.read                     true              PLAIN  thrpt   20  161.496 ± 1.250  ops/s
BenchmarkUuidColumnReader.read                     true   DELTA_BYTE_ARRAY  thrpt   20   33.104 ± 0.257  ops/s
BenchmarkUuidColumnReader.read                    false              PLAIN  thrpt   20   27.478 ± 1.132  ops/s
BenchmarkUuidColumnReader.read                    false   DELTA_BYTE_ARRAY  thrpt   20   20.779 ± 0.564  ops/s
```

ZacBlanco: The updated code is here:

- [BinaryColumnReader](https://github.com/prestodb/presto/blob/891af15747ea2f43f9f2e6a65968a0c5ea6e73d3/presto-parquet/src/main/java/com/facebook/presto/parquet/reader/BinaryColumnReader.java#L46-L48)
- [Uuids](https://github.com/prestodb/presto/blob/891af15747ea2f43f9f2e6a65968a0c5ea6e73d3/presto-common/src/main/java/com/facebook/presto/common/type/Uuids.java#L76-L92)

presto-parquet/src/test/java/com/facebook/presto/parquet/reader/BenchmarkUuidColumnReader.java
```diff
@@ -0,0 +1,104 @@
...
+    protected void writeValue(ValuesWriter writer, long[] batch, int index)
+    {
+        Slice slice = Slices.wrappedLongArray(batch, index * 2, 2);
+        writer.writeBytes(Binary.fromConstantByteArray(slice.getBytes()));
+    }
+
+    @Override
+    protected boolean getEnableOptimizedReader()
+    {
+        return enableOptimizedReader;
```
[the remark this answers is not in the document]
ZacBlanco: new commit added

presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/decoders/rle/UuidRLEDictionaryValuesDecoder.java
```diff
@@ -0,0 +1,64 @@
...
+        ValueBuffer valueBuffer = delegate.readNext(length);
+        int bufferSize = valueBuffer.getBufferSize();
+        byte[] byteBuffer = new byte[bufferSize];
+        int[] offsets = new int[length + 1];
+        delegate.readIntoBuffer(byteBuffer, 0, offsets, 0, valueBuffer);
+
+        for (int i = 0; i < length; i++) {
+            int positionOffset = offsets[i];
+            int positionLength = offsets[i + 1] - positionOffset;
+            byte[] temp = new byte[positionLength];
```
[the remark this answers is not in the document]
ZacBlanco: I fixed this by removing the `Uuids` class altogether and using `BytesUtil#getLong` from the parquet module

presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/decoders/plain/FixedLenByteArrayUuidPlainValuesDecoder.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    @Override
+    public void readNext(long[] values, int offset, int length)
+    {
+        int localBufferOffset = bufferOffset;
+        int endOffset = (offset + length) * 2;
+
+        for (int currentOutputOffset = offset * 2; currentOutputOffset < endOffset; currentOutputOffset += 2) {
+            System.arraycopy(byteBuffer, localBufferOffset, inputBytes, 0, typeLength);
```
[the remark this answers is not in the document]
ZacBlanco: This was fixed by removing the copy and using `BytesUtil#getLong`

ZacBlanco:
Thanks for your detailed review @yingsu00 . The new performance numbers from the benchmark

original
```
Benchmark                       (enableOptimizedReader)  (parquetEncoding)   Mode  Cnt    Score   Error  Units
BenchmarkUuidColumnReader.read                     true              PLAIN  thrpt   10  158.193 ± 1.058  ops/s
BenchmarkUuidColumnReader.read                     true   DELTA_BYTE_ARRAY  thrpt   10   32.211 ± 0.345  ops/s
BenchmarkUuidColumnReader.read                    false              PLAIN  thrpt   10   13.030 ± 0.463  ops/s
BenchmarkUuidColumnReader.read                    false   DELTA_BYTE_ARRAY  thrpt   10   10.121 ± 0.377  ops/s
```

previous update (+50-100% in non-batch reader from original)
```
Benchmark                       (enableOptimizedReader)  (parquetEncoding)   Mode  Cnt    Score   Error  Units
BenchmarkUuidColumnReader.read                     true              PLAIN  thrpt   20  161.496 ± 1.250  ops/s
BenchmarkUuidColumnReader.read                     true   DELTA_BYTE_ARRAY  thrpt   20   33.104 ± 0.257  ops/s
BenchmarkUuidColumnReader.read                    false              PLAIN  thrpt   20   27.478 ± 1.132  ops/s
BenchmarkUuidColumnReader.read                    false   DELTA_BYTE_ARRAY  thrpt   20   20.779 ± 0.564  ops/s
```

newest update (+62% improvement in plain batch reader)
```
Benchmark                       (batchReaderEnabled)  (parquetEncoding)   Mode  Cnt    Score   Error  Units
BenchmarkUuidColumnReader.read                  true              PLAIN  thrpt   20  265.542 ± 3.891  ops/s
BenchmarkUuidColumnReader.read                  true   DELTA_BYTE_ARRAY  thrpt   20   35.511 ± 0.598  ops/s
BenchmarkUuidColumnReader.read                 false              PLAIN  thrpt   20   27.316 ± 1.222  ops/s
BenchmarkUuidColumnReader.read                 false   DELTA_BYTE_ARRAY  thrpt   20   20.883 ± 0.521  ops/s
```

presto-parquet/src/test/java/com/facebook/presto/parquet/batchreader/decoders/TestParquetUtils.java
ZacBlanco: I am not going to change this because it requires re-writing a lots of boilerplate of code to perform the same assertions. I filed an issue for someone to fix this class in the future: https://github.com/prestodb/presto/issues/23840

yingsu00 (approved):
Looks good!

hantangwangd (approved):

presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/BooleanFlatBatchReader.java
elharo: I'd go even further and make this a separate PR that goes in first since there are no other changes in this file.

presto-parquet/src/test/java/com/facebook/presto/parquet/batchreader/decoders/TestParquetUtils.java
elharo: Instead of random values why not just i, -79L*i, 237L*i?

agrawalreetika (approved):

presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/BooleanFlatBatchReader.java
ZacBlanco: What's the benefit of making a separate PR? Commits in a PR are preserved after the merge

PR #23628, opened 2024-09-11
[native] Reduce stuck drivers threshold to detach worker.

spershin:
## Description
In Prestissimo we detect stuck operators (drivers) and when it reaches a certain number we detach the worker from the cluster (announce to the coordinator that it is unavailable for new queries). This is in attempt to keep the cluster running even if few workers are in trouble.
Initially it was just one stuck driver and we saw that in some loads the cluster losing most of the workers.
Then we changed it to the half of the available driver threads and now we are seeing that the queries start queueing up on the coordinator before we reach that threshold on any worker.
This change is to use the driver thread HW multiplier for that threshold. While this will not fix the problem in all cases, it should make it better.
We are also thinking/planning to roll out the change to fail Tasks when stuck drivers are detected for some time to free slots for new queries. But that would be another change.

## Contributor checklist

- [ ] CI passed.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
+++ b/presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
@@ -469,10 +469,13 @@ void PeriodicTaskManager::addWatchdogTask() {
         }
         RECORD_METRIC_VALUE(kCounterNumStuckDrivers, stuckOpCalls.size());
 
-        // Detach worker from the cluster if more than half of driver threads
-        // are blocked by stuck operators (one unique operator can only get
-        // stuck on one unique thread).
-        if (stuckOpCalls.size() > numDriverThreads_ / 2) {
+        // Detach worker from the cluster if more than a certain number of
+        // driver threads are blocked by stuck operators (one unique operator
+        // can only get stuck on one unique thread).
+        const auto numStuckOperatorsToDetachWorker = std::min(
+            SystemConfig::instance()->driverNumStuckOperatorsToDetachWorker(),
+            numDriverThreads_);
+        if (stuckOpCalls.size() >= numStuckOperatorsToDetachWorker) {
           detachWorker("detected stuck operators");
         } else if (!deadlockTasks.empty()) {
           detachWorker("starving or deadlocked task");
--- a/presto-native-execution/presto_cpp/main/common/Configs.cpp
+++ b/presto-native-execution/presto_cpp/main/common/Configs.cpp
@@ -159,6 +159,7 @@ SystemConfig::SystemConfig() {
           NUM_PROP(kDriverStuckOperatorThresholdMs, 30 * 60 * 1000),
           NUM_PROP(
               kDriverCancelTasksWithStuckOperatorsThresholdMs, 40 * 60 * 1000),
+          NUM_PROP(kDriverNumStuckOperatorsToDetachWorker, 8),
           NUM_PROP(kSpillerNumCpuThreadsHwMultiplier, 1.0),
           STR_PROP(kSpillerFileCreateConfig, ""),
           NONE_PROP(kSpillerSpillPath),
@@ -382,6 +383,11 @@ size_t SystemConfig::driverStuckOperatorThresholdMs() const {
 size_t SystemConfig::driverCancelTasksWithStuckOperatorsThresholdMs() const {
   return optionalProperty<size_t>(
              kDriverCancelTasksWithStuckOperatorsThresholdMs)
+      .value();
+}
+
+size_t SystemConfig::driverNumStuckOperatorsToDetachWorker() const {
+  return optionalProperty<size_t>(kDriverNumStuckOperatorsToDetachWorker)
       .value();
 }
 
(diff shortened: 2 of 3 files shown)


presto-native-execution/presto_cpp/main/PeriodicTaskManager.h
```diff
@@ -149,6 +149,7 @@ class PeriodicTaskManager {
   folly::FunctionScheduler oneTimeRunner_;
   folly::ThreadedRepeatingFunctionRunner repeatedRunner_;
   size_t numDriverThreads_{0};
+  const size_t numStuckDriversToDetachWorker;
```
xiaoxmeng, line 152: Put const member first
```
s/numStuckDriversToDetachWorker/numStuckDriversToDetachWorker_/
```

presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
```diff
@@ -108,7 +108,10 @@ PeriodicTaskManager::PeriodicTaskManager(
       asyncDataCache_(asyncDataCache),
       arbitrator_(velox::memory::memoryManager()->arbitrator()),
       connectors_(connectors),
-      server_(server) {}
+      server_(server),
+      numStuckDriversToDetachWorker(
```
xiaoxmeng, line 112: Why use hw multiplier for this?

xiaoxmeng (commented):
@spershin thanks!

presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
spershin: Not too large number and not too big. Also this would tie the threshold to the number of driver threads.
We could use any % - I'm open for other ideas.
As I mentioned in the description, this is not a bullet proof strategy, but  better than the one we have now.

presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
```diff
@@ -108,6 +108,8 @@ PeriodicTaskManager::PeriodicTaskManager(
       asyncDataCache_(asyncDataCache),
       arbitrator_(velox::memory::memoryManager()->arbitrator()),
       connectors_(connectors),
+      numStuckDriversToDetachWorker_(
```
xiaoxmeng, line 111: I think it would better to have a separate config, such as double type hw multiplier for detach. It could be either based on the hardware concurrency or a factor of driverNumCpuThreadsHwMultiplier? such as
numStuckDriversToDetachWorker_(hw concurrency * driverNumCpuThreadsHwMultiplier * stuckDriversToDetachWorkerFactor)?

spershin: @xiaoxmeng 
So, you think we should have a separate config property for this?

What you describe looks like simple %:

pctStuckDriversToDetachWorker_
0 - never detach
1.0 - detach if all of them are blocked.
0.5 - half.
etc.

Or we can have a simple numStuckDriversToDetachWorker - a number of drivers, regardless of HW concurrency and number of cores.

Or we can have what is in this PR to not spawn yet another 'almost never used' config property.

I don't have a strong preference, just general strive for simplicity.

What would you prefer here?

spershin: Chatted offline, will add plain number parameter.

presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
```diff
@@ -469,10 +469,11 @@ void PeriodicTaskManager::addWatchdogTask() {
...
 
-        // Detach worker from the cluster if more than half of driver threads
-        // are blocked by stuck operators (one unique operator can only get
-        // stuck on one unique thread).
-        if (stuckOpCalls.size() > numDriverThreads_ / 2) {
+        // Detach worker from the cluster if more than a certain number of
+        // driver threads are blocked by stuck operators (one unique operator
+        // can only get stuck on one unique thread).
+        if (stuckOpCalls.size() >=
+            SystemConfig::instance()->driverNumStuckOperatorsToDetachWorker()) {
```
xiaoxmeng, line 476: do you want to take
```
std::max(SystemConfig::instance()->driverNumStuckOperatorsToDetachWorker(), max number of driver threads?);
```

xiaoxmeng (commented):
@spershin LGTM % one comment. Thanks!

presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
spershin: @xiaoxmeng 
What do you mean by "max number of driver threads"?

xiaoxmeng (approved):

presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
xiaoxmeng: yeah, std::min. "max number of driver threads" the number of driver threads

PR #20501, opened 2023-08-05
Add Iceberg Filter Pushdown Optimizer Rule for execution with Velox

imjalpreet:
Co-authors: @imjalpreet, @agrawalreetika and @tdcmeehan 

## Description
This PR introduces the below changes in Iceberg Connector in an ongoing effort to make it compatible with execution on Velox:
* New config property and session property to control the behavior of certain features like Filter Pushdown based on Java or Native Worker
* Add new fields in IcebergTableLayoutHandle to implement filter pushdown
* Add Iceberg Filter Pushdown Optimizer Rule for execution with Velox

## Motivation and Context
https://github.com/facebookincubator/velox/issues/5977

## Test Plan

These changes have been tested with Iceberg Catalog type Hive/Glue and Hadoop.

## Contributor checklist

- [x] Please make sure your submission complies with our [development](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#development), [formatting](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#formatting), [commit message](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#commit-formatting-and-pull-requests), and [attribution guidelines](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#attribution).
- [x] PR description addresses the issue accurately and concisely.  If the change is non-trivial, a GitHub Issue is referenced.
- [ ] Documented new properties (with its default value), SQL syntax, functions, or other functionality.
- [x] If release notes are required, they follow the [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines).
- [x] Adequate tests were added if applicable.
- [x] CI passed.

## Release Notes
Please follow [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines) and fill in the release notes below.

```
== RELEASE NOTES ==

Iceberg Changes
* Add ``iceberg.pushdown-filter-enabled`` config property to Iceberg Connector. This config property will control the behaviour of Filter Pushdown in the iceberg connector.
* Add Iceberg Filter Pushdown Optimizer Rule for execution with Velox
```

PR #23645, opened 2024-09-13
[WIP]Make partitions loading for iceberg table lazy to avoid unnecessary loading

hantangwangd:
## Description

Make partitions loading for iceberg table lazy to avoid unnecessary loading

## Motivation and Context

Make partitions loading for iceberg table lazy to avoid unnecessary loading

## Impact

N/A

## Test Plan

N/A

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```


ZacBlanco (changes requested):
I understand the motivation for this, but I am a little concerned about the other parts which implement the limiting of the partition loading. It feels a little convoluted, especially requiring checking the instance type of the iterable to set the max number of partitions to iterate. I feel like it could be simplified by just creating an iterable implementation which accepts the limit in the first place, rather than updating it after the fact. I may be missing something in my understanding if that isn't a feasible solution

The current implementation of partition loading in `IcebergPartitionLoader` looks good though.

presto-spi/src/main/java/com/facebook/presto/spi/LazyIterable.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.spi;
+
+public interface LazyIterable<T>
```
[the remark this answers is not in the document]
hantangwangd: Thanks for the review and feedback, the entire thinking is described as follows, please let me know if there is anything unreasonable. 

There are several points to consider when implementing setting max threshold for partitions lazy loading:

1. If any delete files are found during the file scanning process, the scanning and loading of partition values would be no longer continue, but instead a single value representing unpartitioned should be returned. So we cannot just scan a part of files to get a specified number of available partition values. So it seems that we cannot set the max threshold and terminate the scan in advance through methods such like `Stream.of(new Iterable<>(...){}).limit(<N>)...`.

2. The max threshold can be passed into `IcebergPartitionLoader` and used there during the loading to terminate the entire scanning process. Because during the loading process, even if no delete files are encountered, when the number of loaded available partitions exceeds the max threshold, we can terminate directly and return a value representing unpartitioned. So we have to figure out a way to pass `max threshold` into `IcebergPartitionLoader`.

3. The construction of the table layout, that is, the lazy partitions and `DiscretePredicates` information is built earlier than the actual loading. Meanwhile, the value of max threshold can only be determined when actual loading is required. For example, in `IcebergMetadataOptimizer`, if further reducible optimization is available, we do not limit the max number of partitions loaded; otherwise, the max number of loaded partitions is set 1000 by default. So it seems that we need to support this delay setting of maximum threshold for the partition loader.

4. `LazyLoadedPartitions` and `PartitionLoader` are both defined in `presto-hive-common`, while `DiscretePredicates` is defined in `presto-spi`, which do not depend on `presto-hive-common`. As mentioned above, we need a way to set the max threshold into `PartitionLoader` through `DiscretePredicates` before actual loading. So we define an interface in `presto-spi` and let `LazyLoadedPartitions` in `presto-hive-common` to implement it, in this way, max threshold can be finally passed into `PartitionLoader` through `LazyLoadedPartitons`.

presto-hive-common/src/main/java/com/facebook/presto/hive/LazyLoadedPartitions.java
```diff
@@ -0,0 +1,91 @@
...
+ */
+package com.facebook.presto.hive;
+
+import com.facebook.presto.spi.LazyIterable;
+
+import java.util.List;
+
+import static java.util.Objects.requireNonNull;
+
+public class LazyLoadedPartitions
```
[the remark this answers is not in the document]
hantangwangd: Good suggestion, after recheck the code I think it's feasible and reasonable. Fixed!

steveburnett:
Thanks for the release note entry! Minor suggestion to help follow the [Order of changes](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines#order-of-changes) in the Release Note Guidelines. 

```
== RELEASE NOTES ==

Iceberg Connector Changes
* Improve partitions loading by making them lazy for Iceberg tables to avoid unnecessary loading. :pr:`23645`
```

steveburnett:
> @steveburnett Thanks for your suggestion, fixed! Please take a look when convenient.

@hantangwangd LGTM, thanks!

steveburnett:
Hi, new release note guidelines. Please remove the manual PR link in the following format from the release note entries for this PR.

```

:pr:`12345`

```

hantangwangd:
> Hi, new release note guidelines. Please remove the manual PR link in the following format from the release note entries for this PR.

Oh sorry forgot that. Thanks for your reminder @steveburnett, fixed!

steveburnett:
@hantangwangd, if you rebase this PR that should pick up fixes to some tests that were made in the last few weeks, and should help with the blocked tests.

hantangwangd:
@steveburnett Sure, thanks for the reminder, done!

ZacBlanco:
I was playing around with this because I am not happy with the way the abstraction exposes a setMaxCount to change the iterable. It just doesn't sit right with me. I made a patch to slightly change the current approach. Could you take a look and tell me what you think?

[change-lazy.patch](https://github.com/user-attachments/files/19812972/change-lazy.patch)

hantangwangd:
> I made a patch to slightly change the current approach. Could you take a look and tell me what you think?

@ZacBlanco Sure, thank you very much for your perspective and the effort you have put into this. I'll check the patch and do some benchmark as soon as possible.

hantangwangd:
Hi @ZacBlanco, after checking the changes in the patch, I believe there is a slight misunderstanding about the use of threshold limit in the following code:

```
    public DiscreteLazyPredicates(DiscretePredicates discretePredicates, int limit)
    {
        super(discretePredicates.getColumns(), () -> StreamSupport.stream(discretePredicates.getPredicates().spliterator(), false).limit(limit).iterator());
    }
```

The `limit` here is supposed to serve as a threshold ---- once the number of loaded partitions exceeds it, we should stop the loading as soon as possible and return an `unpartitioned` flag to skip this optimization rule.

However, in this patch, the use of `limit` seems to load all partitions first, and then truncate the result to at most `limit` values. This can lead to incorrect result when the number of partitions exceeds the threshold. It would fail the unit test `TestIcebergLogicalPlanner.testMetadataQueryOptimizerWithMaxPartitionThreshold(...)`.

But if you're still unsure about the max load limit change, I'm OK to split this PR into two separated PRs. We can go ahead to pass the lazy loading part first, and then discuss the max load limit part of change in detail separately. What's your opinion?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
```diff
@@ -435,7 +437,11 @@ public ConnectorTableLayout getTableLayout(ConnectorSession session, ConnectorTa
 
         Map<String, ColumnHandle> predicateColumns = icebergTableLayoutHandle.getPredicateColumns().entrySet()
                 .stream().collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
-        Optional<TupleDomain<ColumnHandle>> predicate = partitions.map(parts -> getPredicate(icebergTableLayoutHandle, partitionColumns, parts, predicateColumns));
+        Optional<TupleDomain<ColumnHandle>> predicate = partitions
+                .map(parts ->
+                        getPredicate(icebergTableLayoutHandle, partitionColumns,
+                                StreamSupport.stream(parts.spliterator(), false).toList(),
```
PingLiuPing, line 443: When pushdown filter is enabled, it forces loading of all partitions, defeating the lazy loading. Is this work as design?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergPartitionLoader.java
```diff
@@ -0,0 +1,210 @@
...
+    private final FileFormat fileFormat;
+    private final ConnectorTableHandle tableHandle;
+    private final Constraint<ColumnHandle> constraint;
+    private final List<IcebergColumnHandle> partitionColumns;
+    private final boolean isEmptyTable;
+    private final TableScan tableScan;
+    private final RuntimeStats runtimeStats;
+    private final String loadingCountKey;
+    private final String loadingTimeKey;
+    public IcebergPartitionLoader(
```
PingLiuPing, line 78: nit: empty line.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
hantangwangd: As I see, "pushdown-filter-enabled" is a specific optimization path for native worker only. It's behavior is not exactly the same as that for java worker: filters may be thoroughly pushed down even if they do not fully comply with the partition boundary guarantees, with subsequent processing handled by the native worker.

Within this flow, the optimization logic in `BaseSubfieldExtractionRewriter` appears to intentionally accept the TupleDomain constructed from partitions value. Its read logic seems tightly coupled with Velox's relevant handling (see #20501 for reference). So I have chosen not to modify this section for now. 

Looking ahead, if feasible, we could consider further investigation to decouple this logic and refactor it towards a lazy loading approach. What are your thoughts on this?

hantangwangd:
@PingLiuPing thanks for the review and suggestion. I've supplemented some content about the queries that can be optimized using partitions information, along with the description of the benchmark selected queries.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
PingLiuPing: Thanks for providing the background information.

`Looking ahead, if feasible, we could consider further investigation to decouple this logic and refactor it towards a lazy loading approach.`
Yes, agree, I will create an issue to follow up.

presto-hive-common/src/main/java/com/facebook/presto/hive/MetadataUtils.java
```diff
@@ -57,15 +58,26 @@ public final class MetadataUtils
...
+            Iterable<TupleDomain<ColumnHandle>> partitionDomains = Iterables.transform(partitions, (hivePartition) -> {
+                if (hivePartition.getPartitionId().equals(UNPARTITIONED_ID)) {
+                    isUnPartitioned.set(true);
+                    return TupleDomain.all();
+                }
+                else {
+                    return TupleDomain.fromFixedValues(hivePartition.getKeys());
+                }
+            });
+            if (!isUnPartitioned.get()) {
```
PingLiuPing: Seems `isUnpartitioned` is used before setting the value, meaning `isUnpartitioned` is always false here.
The `isUnPartitioned` flag is set inside a lazy Iterables.transform() lambda, but checked before the iterable is consumed.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergPartitionLoader.java
```diff
@@ -0,0 +1,211 @@
...
+    @Override
+    public synchronized List<HivePartition> loadPartitions()
+    {
+        if (isEmptyTable) {
+            return ImmutableList.of();
+        }
+
+        long startTime = System.nanoTime();
+        Set<HivePartition> partitions = new HashSet<>();
+        try (CloseableIterable<FileScanTask> fileScanTasks = tableScan.planFiles()) {
```
PingLiuPing, line 122: Is `tableScan.planFiles()` an expensive operation? If yes, we might need to consider to cache the result since we do it twice (another one in line 203) given a non-empty table.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
PingLiuPing: Created a follow-up issue https://github.com/prestodb/presto/issues/26880

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergPartitionLoader.java
hantangwangd: After checking Iceberg's code in detail, I believe that `tableScan.planFiles()` is a low-overhead operation. It mainly reads the manifest-list file to obtain metadata of manifests, and for the same `tableScan` instance, the result is cached, meaning repeated calls incur even less overhead. The actual expensive operation is iterating through the returned `CloseableIterable`, which requires reading each manifest file. Additionally, the `CloseableIterable` returned by `tableScan.planFiles()` is a resource that needs to be explicitly closed. It's more suitable for one-time use and immediate release rather than being cached and consumed multiple times. Therefore, I think maintaining the `tableScan` instance itself is sufficient. Do you think this makes sense?

presto-hive-common/src/main/java/com/facebook/presto/hive/MetadataUtils.java
hantangwangd: Nice catch! Upon re-examining the relevant code, I think this logic can be removed directly. Simply returning a `TupleDomain.all` is sufficient. Later, it will be interpreted as `Unpartitioned`. Seeing the code here: https://github.com/prestodb/presto/blob/master/presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergMetadataOptimizer.java#L239-L249

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
hantangwangd: That would be great, thanks!

presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergLogicalPlanner.java
```diff
@@ -207,6 +214,90 @@ public void testMetadataQueryOptimizer(boolean enabled)
...
+                            ImmutableList.of("a", "b"),
+                            ImmutableList.of(
+                                    ImmutableList.of(new LongLiteral("1"), new StringLiteral("1001")),
+                                    ImmutableList.of(new LongLiteral("2"), new StringLiteral("1001")),
+                                    ImmutableList.of(new LongLiteral("3"), new StringLiteral("1002")),
+                                    ImmutableList.of(new LongLiteral("4"), new StringLiteral("1002"))))));
+            assertEquals(partitionsLazyLoadingCountMetric.getCount(), 2);
+
+            // Do not perform metadata optimization when the number of partitions exceeds the threshold
+            Session sessionWithThresholdSmaller = getSessionWithOptimizeMetadataQueriesAndThreshold(false, 3);
```
PingLiuPing, line 265: false to `enabled` ?

hantangwangd: The boolean flag is intentionally set to false to disable `pushdown_filter_enabled`. This is because when the number of partitions exceeds the threshold, metadata optimization is skipped and the `tableScanNode` remains in the plan tree. Enabling `pushdown_filter_enabled` in this scenario would cause queries to fail directly with the error:

```
com.facebook.presto.spi.PrestoException: Filter Pushdown not supported for Iceberg Java Connector
```

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergPartitionLoader.java
```diff
@@ -0,0 +1,211 @@
...
+                    partitions.add(newPartition);
+                }
+            }
+            runtimeStats.addMetricValue(loadingCountKey, NONE, 1);
+            runtimeStats.addMetricValue(loadingTimeKey, NANO, System.nanoTime() - startTime);
+        }
+        catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+        return new ArrayList<>(partitions);
```
tdcmeehan:
```suggestion
        return ImmutableList.copyOf(partitions);
```

presto-hive-common/src/main/java/com/facebook/presto/hive/LazyLoadedPartitions.java
```diff
@@ -0,0 +1,117 @@
...
+
+    private void tryFullyLoad()
+    {
+        if (!this.fullyLoaded) {
+            synchronized (this) {
+                if (!this.fullyLoaded) {
+                    this.partitions = ImmutableList.copyOf(this.partitionLoader.loadPartitions());
+                    this.fullyLoaded = true;
+                    this.partitionLoader = null;
+                }
```
tdcmeehan: It seems like there is a lot of redundant usage of `this`.  I think we can remove them to simplify the code a bit.

presto-hive-common/src/main/java/com/facebook/presto/hive/BaseHiveTableLayoutHandle.java
```diff
@@ -91,7 +91,7 @@ public TupleDomain<ColumnHandle> getPartitionColumnPredicate()
      * @return list of partitions if available, {@code Optional.empty()} if dropped
      */
     @JsonIgnore
-    public Optional<List<HivePartition>> getPartitions()
+    public Optional<? extends Iterable<HivePartition>> getPartitions()
```
tdcmeehan: I think it would be better to simply return `LazyLoadedPartitions` here, as the generics are a bit awkward.  If we do this, I would rename `LazyLoadedPartitions` to something simpler, maybe `PartitionSet`, as there's an eager constructor on that class already.

hantangwangd: Thanks for the suggestion, sounds great to me. I've refactored the code accordingly. Please take a look when you get a chance.

presto-hive-common/src/main/java/com/facebook/presto/hive/LazyLoadedPartitions.java
hantangwangd: Sure, done

tdcmeehan (approved):
Thanks @hantangwangd

ZacBlanco (approved):
LGTM!

PR #23649, opened 2024-09-13
Add property to enable eager building of plan as part of new native plan checker SPI

BryanCutler:
## Description

This is a prerequisite for the native plan checker SPI being worked on in https://github.com/prestodb/presto/pull/23596 and discussed in https://github.com/prestodb/rfcs/blob/main/RFC-0008-plan-checker.md.

The new property `eager-plan-validation-enabled` will enable the eager building and validation of a logical plan so that any errors or incompatibilities in the plan will cause the query to fail quickly, before queueing and cluster resources are assigned to keep queries with invalid plans from holding slots in the queue. 

## Motivation and Context

The new SPI for a native plan checker, being added in https://github.com/prestodb/presto/pull/23596,  would use this property so that a plan can be sent to the native sidecar and incompatibles with velox would be caught before actually sending the plan to a native worker to start execution. However, there is also benefit in enabling this property for non-native clusters since all plan checkers are run once the plan is built.

## Test Plan

Manual testing has been done with the property enabled and verified that queries that get queued up will be allowed to be validated and, if fail, removed from the queue before execution begins.

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Add new property `eager-plan-validation-enabled` for eager building of validation of a logical plan before queuing. :pr:`23649`

### DIFF
--- a/presto-docs/src/main/sphinx/admin/properties.rst
+++ b/presto-docs/src/main/sphinx/admin/properties.rst
@@ -72,6 +72,17 @@ Number of local parallel table writer threads per worker for partitioned writes.
 set, the number set by ``task_writer_count`` will be used. It is required to be a power
 of two for a Java query engine.
 
+``eager-plan-validation-enabled``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``boolean``
+* **Default value:** ``false``
+
+This property enables the eager building and validation of a logical plan.
+When enabled, the logical plan will begin to be built and validated before
+queueing and allocation of cluster resources so that any errors or
+incompatibilities in the query plan will fail quickly and inform the user.
+
 .. _tuning-memory:
 
 Memory Management Properties
--- a/presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
+++ b/presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
@@ -331,6 +331,7 @@ public final class SystemSessionProperties
     public static final String REWRITE_EXPRESSION_WITH_CONSTANT_EXPRESSION = "rewrite_expression_with_constant_expression";
     public static final String PRINT_ESTIMATED_STATS_FROM_CACHE = "print_estimated_stats_from_cache";
     public static final String REMOVE_CROSS_JOIN_WITH_CONSTANT_SINGLE_ROW_INPUT = "remove_cross_join_with_constant_single_row_input";
+    public static final String EAGER_PLAN_VALIDATION_ENABLED = "eager_plan_validation_enabled";
 
     // TODO: Native execution related session properties that are temporarily put here. They will be relocated in the future.
     public static final String NATIVE_SIMPLIFIED_EXPRESSION_EVALUATION_ENABLED = "native_simplified_expression_evaluation_enabled";
@@ -2018,6 +2019,11 @@ public final class SystemSessionProperties
                         "If one input of the cross join is a single row with constant value, remove this cross join and replace with a project node",
                         featuresConfig.isRemoveCrossJoinWithSingleConstantRow(),
                         false),
+                booleanProperty(
+                        EAGER_PLAN_VALIDATION_ENABLED,
+                        "Enable eager building and validation of logical plan before queueing",
+                        featuresConfig.isEagerPlanValidationEnabled(),
+                        false),
                 new PropertyMetadata<>(
                         DEFAULT_VIEW_SECURITY_MODE,
                         format("Set default view security mode. Options are: %s",
@@ -3342,6 +3348,11 @@ public final class SystemSessionProperties
     public static boolean isRewriteExpressionWithConstantEnabled(Session session)
     {
         return session.getSystemProperty(REWRITE_EXPRESSION_WITH_CONSTANT_EXPRESSION, Boolean.class);
+    }
+
+    public static boolean isEagerPlanValidationEnabled(Session session)
+    {
+        return session.getSystemProperty(EAGER_PLAN_VALIDATION_ENABLED, Boolean.class);
     }
 
     public static CreateView.Security getDefaultViewSecurityMode(Session session)
--- a/presto-main/src/main/java/com/facebook/presto/execution/SqlQueryExecution.java
+++ b/presto-main/src/main/java/com/facebook/presto/execution/SqlQueryExecution.java
@@ -75,13 +75,16 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 
 import static com.facebook.presto.SystemSessionProperties.getExecutionPolicy;
 import static com.facebook.presto.SystemSessionProperties.getQueryAnalyzerTimeout;
+import static com.facebook.presto.SystemSessionProperties.isEagerPlanValidationEnabled;
 import static com.facebook.presto.SystemSessionProperties.isLogInvokedFunctionNamesEnabled;
 import static com.facebook.presto.SystemSessionProperties.isSpoolingOutputBufferEnabled;
 import static com.facebook.presto.common.RuntimeMetricName.FRAGMENT_PLAN_TIME_NANOS;
@@ -141,6 +144,8 @@ public class SqlQueryExecution
     private final PlanCanonicalInfoProvider planCanonicalInfoProvider;
     private final QueryAnalysis queryAnalysis;
     private final AnalyzerContext analyzerContext;
+    private final CompletableFuture<PlanRoot> planFuture;
+    private final AtomicBoolean planFutureLocked = new AtomicBoolean();
 
     private SqlQueryExecution(
             QueryAnalyzer queryAnalyzer,
@@ -159,6 +164,7 @@ public class SqlQueryExecution
             ExecutorService queryExecutor,
             ScheduledExecutorService timeoutThreadExecutor,
             SectionExecutionFactory sectionExecutionFactory,
+            ExecutorService eagerPlanValidationExecutor,
             InternalNodeManager internalNodeManager,
             ExecutionPolicy executionPolicy,
             SplitSchedulerStats schedulerStats,
@@ -243,6 +249,10 @@ public class SqlQueryExecution
                     }
                 }
             }
+
+            // Optionally build and validate plan immediately, before execution begins
+            planFuture = isEagerPlanValidationEnabled(getSession()) ?
+                    CompletableFuture.supplyAsync(this::runCreateLogicalPlanAsync, eagerPlanValidationExecutor) : null;
         }
     }
 
@@ -460,8 +470,13 @@ public class SqlQueryExecution
                         Thread.currentThread(),
                         timeoutThreadExecutor,
                         getQueryAnalyzerTimeout(getSession()))) {
-                    // create logical plan for the query
-                    plan = createLogicalPlanAndOptimize();
+                    // If planFuture has not started, cancel and build plan in current thread
+                    if (planFuture != null && !planFutureLocked.compareAndSet(false, true)) {
+                        plan = planFuture.get();
+                    }
+                    else {
+                        plan = createLogicalPlanAndOptimize();
+                    }
                 }
 
                 metadata.beginQuery(getSession(), plan.getConnectors());
@@ -587,6 +602,21 @@ public class SqlQueryExecution
         }
         catch (StackOverflowError e) {
             throw new PrestoException(NOT_SUPPORTED, "statement is too large (stack overflow during analysis)", e);
+        }
+    }
+
+    private PlanRoot runCreateLogicalPlanAsync()
+    {
+        try {
+            // Check if creating plan async has been cancelled
+            if (planFutureLocked.compareAndSet(false, true)) {
+                return createLogicalPlanAndOptimize();
+            }
+            return null;
+        }
+        catch (Throwable e) {
+            fail(e);
+            throw e;
         }
     }
 
@@ -862,6 +892,7 @@ public class SqlQueryExecution
         private final ScheduledExecutorService timeoutThreadExecutor;
         private final ExecutorService queryExecutor;
         private final SectionExecutionFactory sectionExecutionFactory;
+        private final ExecutorService eagerPlanValidationExecutor;
         private final InternalNodeManager internalNodeManager;
         private final Map<String, ExecutionPolicy> executionPolicies;
         private final StatsCalculator statsCalculator;
@@ -883,6 +914,7 @@ public class SqlQueryExecution
                 @ForQueryExecution ExecutorService queryExecutor,
                 @ForTimeoutThread ScheduledExecutorService timeoutThreadExecutor,
                 SectionExecutionFactory sectionExecutionFactory,
+                @ForEagerPlanValidation ExecutorService eagerPlanValidationExecutor,
                 InternalNodeManager internalNodeManager,
                 Map<String, ExecutionPolicy> executionPolicies,
                 SplitSchedulerStats schedulerStats,
@@ -904,6 +936,7 @@ public class SqlQueryExecution
             this.queryExecutor = requireNonNull(queryExecutor, "queryExecutor is null");
             this.timeoutThreadExecutor = requireNonNull(timeoutThreadExecutor, "timeoutThreadExecutor is null");
             this.sectionExecutionFactory = requireNonNull(sectionExecutionFactory, "sectionExecutionFactory is null");
+            this.eagerPlanValidationExecutor = requireNonNull(eagerPlanValidationExecutor, "eagerPlanValidationExecutor is null");
             this.internalNodeManager = requireNonNull(internalNodeManager, "internalNodeManager is null");
             this.executionPolicies = requireNonNull(executionPolicies, "schedulerPolicies is null");
             this.planOptimizers = planOptimizers.getPlanningTimeOptimizers();
@@ -946,6 +979,7 @@ public class SqlQueryExecution
                     queryExecutor,
                     timeoutThreadExecutor,
                     sectionExecutionFactory,
+                    eagerPlanValidationExecutor,
                     internalNodeManager,
                     executionPolicy,
                     schedulerStats,
--- a/presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java
@@ -296,6 +296,9 @@ public class FeaturesConfig
 
     private boolean isInlineProjectionsOnValuesEnabled;
 
+    private boolean eagerPlanValidationEnabled;
+    private int eagerPlanValidationThreadPoolSize = 20;
+
     public enum PartitioningPrecisionStrategy
     {
         // Let Presto decide when to repartition
@@ -2968,5 +2971,31 @@ public class FeaturesConfig
     {
         this.isInlineProjectionsOnValuesEnabled = isInlineProjectionsOnValuesEnabled;
         return this;
+    }
+
+    @Config("eager-plan-validation-enabled")
+    @ConfigDescription("Enable eager building and validation of logical plan before queueing")
+    public FeaturesConfig setEagerPlanValidationEnabled(boolean eagerPlanValidationEnabled)
+    {
+        this.eagerPlanValidationEnabled = eagerPlanValidationEnabled;
+        return this;
+    }
+
+    public boolean isEagerPlanValidationEnabled()
+    {
+        return this.eagerPlanValidationEnabled;
+    }
+
+    @Config("eager-plan-validation-thread-pool-size")
+    @ConfigDescription("Size of thread pool to use when eager plan validation is enabled")
+    public FeaturesConfig setEagerPlanValidationThreadPoolSize(int eagerPlanValidationThreadPoolSize)
+    {
+        this.eagerPlanValidationThreadPoolSize = eagerPlanValidationThreadPoolSize;
+        return this;
+    }
+
+    public int getEagerPlanValidationThreadPoolSize()
+    {
+        return this.eagerPlanValidationThreadPoolSize;
     }
 }
--- a/presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/PlanChecker.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/PlanChecker.java
@@ -16,12 +16,17 @@ package com.facebook.presto.sql.planner.sanity;
 import com.facebook.presto.Session;
 import com.facebook.presto.metadata.Metadata;
 import com.facebook.presto.spi.WarningCollector;
+import com.facebook.presto.spi.plan.PlanCheckerProvider;
 import com.facebook.presto.spi.plan.PlanNode;
+import com.facebook.presto.spi.plan.SimplePlanFragment;
 import com.facebook.presto.sql.analyzer.FeaturesConfig;
+import com.facebook.presto.sql.planner.PlanFragment;
 import com.google.common.collect.ImmutableListMultimap;
 import com.google.common.collect.Multimap;
 
 import javax.inject.Inject;
+
+import static java.util.Objects.requireNonNull;
 
 /**
  * Perform checks on the plan that may generate warnings or errors.
@@ -29,15 +34,17 @@ import javax.inject.Inject;
 public final class PlanChecker
 {
     private final Multimap<Stage, Checker> checkers;
+    private final PlanCheckerProviderManager planCheckerProviderManager;
 
     @Inject
-    public PlanChecker(FeaturesConfig featuresConfig)
+    public PlanChecker(FeaturesConfig featuresConfig, PlanCheckerProviderManager planCheckerProviderManager)
     {
-        this(featuresConfig, false);
+        this(featuresConfig, false, planCheckerProviderManager);
     }
 
-    public PlanChecker(FeaturesConfig featuresConfig, boolean forceSingleNode)
+    public PlanChecker(FeaturesConfig featuresConfig, boolean forceSingleNode, PlanCheckerProviderManager planCheckerProviderManager)
     {
+        this.planCheckerProviderManager = requireNonNull(planCheckerProviderManager, "planCheckerProviderManager is null");
         ImmutableListMultimap.Builder<Stage, Checker> builder = ImmutableListMultimap.builder();
         builder.putAll(
                         Stage.INTERMEDIATE,
@@ -78,25 +85,58 @@ public final class PlanChecker
     public void validateFinalPlan(PlanNode planNode, Session session, Metadata metadata, WarningCollector warningCollector)
     {
         checkers.get(Stage.FINAL).forEach(checker -> checker.validate(planNode, session, metadata, warningCollector));
+        for (PlanCheckerProvider provider : planCheckerProviderManager.getPlanCheckerProviders()) {
+            for (com.facebook.presto.spi.plan.PlanChecker checker : provider.getFinalPlanCheckers()) {
+                checker.validate(planNode, warningCollector);
+            }
+        }
     }
 
     public void validateIntermediatePlan(PlanNode planNode, Session session, Metadata metadata, WarningCollector warningCollector)
     {
         checkers.get(Stage.INTERMEDIATE).forEach(checker -> checker.validate(planNode, session, metadata, warningCollector));
+        for (PlanCheckerProvider provider : planCheckerProviderManager.getPlanCheckerProviders()) {
+            for (com.facebook.presto.spi.plan.PlanChecker checker : provider.getIntermediatePlanCheckers()) {
+                checker.validate(planNode, warningCollector);
+            }
+        }
     }
 
-    public void validatePlanFragment(PlanNode planNode, Session session, Metadata metadata, WarningCollector warningCollector)
+    public void validatePlanFragment(PlanFragment planFragment, Session session, Metadata metadata, WarningCollector warningCollector)
     {
-        checkers.get(Stage.FRAGMENT).forEach(checker -> checker.validate(planNode, session, metadata, warningCollector));
+        checkers.get(Stage.FRAGMENT).forEach(checker -> checker.validateFragment(planFragment, session, metadata, warningCollector));
+        for (PlanCheckerProvider provider : planCheckerProviderManager.getPlanCheckerProviders()) {
+            for (com.facebook.presto.spi.plan.PlanChecker checker : provider.getFragmentPlanCheckers()) {
+                checker.validateFragment(toSimplePlanFragment(planFragment), warningCollector);
+            }
+        }
     }
 
     public interface Checker
     {
         void validate(PlanNode planNode, Session session, Metadata metadata, WarningCollector warningCollector);
+
+        default void validateFragment(PlanFragment planFragment, Session session, Metadata metadata, WarningCollector warningCollector)
+        {
+            validate(planFragment.getRoot(), session, metadata, warningCollector);
+        }
     }
 
     private enum Stage
     {
         INTERMEDIATE, FINAL, FRAGMENT
+    }
+
+    private static SimplePlanFragment toSimplePlanFragment(PlanFragment planFragment)
+    {
+        return new SimplePlanFragment(
+                planFragment.getId(),
+                planFragment.getRoot(),
+                planFragment.getVariables(),
+                planFragment.getPartitioning(),
+                planFragment.getTableScanSchedulingOrder(),
+                planFragment.getPartitioningScheme(),
+                planFragment.getStageExecutionDescriptor(),
+                planFragment.isOutputTableWriterFragment());
     }
 }
--- a/presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/PlanCheckerProviderManager.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/PlanCheckerProviderManager.java
@@ -1,0 +1,59 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.sql.planner.sanity;
+
+import com.facebook.presto.spi.plan.PlanCheckerProvider;
+import com.facebook.presto.spi.plan.PlanCheckerProviderFactory;
+import com.facebook.presto.spi.plan.SimplePlanFragmentSerde;
+import com.google.inject.Inject;
+
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.stream.Collectors;
+
+import static java.lang.String.format;
+import static java.util.Objects.requireNonNull;
+
+public class PlanCheckerProviderManager
+{
+    private final SimplePlanFragmentSerde simplePlanFragmentSerde;
+    private final Map<String, PlanCheckerProviderFactory> providerFactories = new ConcurrentHashMap<>();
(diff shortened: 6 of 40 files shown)

### CODE REVIEW

**Comment by BryanCutler** [comment]
cc @tdcmeehan @rschlussel

**Review by steveburnett — DISMISSED** [review]
LGTM! (docs)

**Comment by jaystarshot** [comment]
The plan is always built and validated before execution

**Comment by BryanCutler** [comment]
> The plan is always built and validated before execution

@jaystarshot yes they are - I suppose I worded the description poorly. The goal here is to make sure the plan is validated before cluster resources are assigned to run the query. Specifically, with a native cluster when there are plans that are incompatible with Velox, they do not fail until evaluated on a worker. Eagerly checking the plan will allow such failures to happen before running at the point where resources are assigned and also help remove these queries from holding slots in the queue.

File: presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java [hunk 1]
```diff
@@ -2969,4 +2971,17 @@ public FeaturesConfig setInlineProjectionsOnValues(boolean isInlineProjectionsOn
         this.isInlineProjectionsOnValuesEnabled = isInlineProjectionsOnValuesEnabled;
         return this;
     }
+
+    @Config("query-execution-fail-fast-enabled")
```
rschlussel, line 2975: maybe call this `validate-plan-before-queueing`?  Is that what this is for?

presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
```diff
@@ -2010,6 +2011,11 @@ public SystemSessionProperties(
                         "If one input of the cross join is a single row with constant value, remove this cross join and replace with a project node",
                         featuresConfig.isRemoveCrossJoinWithSingleConstantRow(),
                         false),
+                booleanProperty(
+                        QUERY_EXECUTION_FAIL_FAST_ENABLED,
+                        "Enable eager building and validation of logical plan before execution",
```
rschlussel, line 2016:
```suggestion
                        "Enable eager building and validation of logical plan before queueing",
```

presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java
BryanCutler: Yeah, it could be named better. However, it isn't specific for only queries that have been queued. It starts building/validating the plan as soon as `SqlQueryExecution` is constructed instead of waiting until `SqlQueryExecution.start()` is invoked, where `ManagedQueryExecution.startWaitingForResources()` has already completed.

Perhaps something like `eagerly-validate-plan-before-resources` ?

presto-main/src/main/java/com/facebook/presto/execution/SqlQueryExecution.java
```diff
@@ -243,6 +246,9 @@ private SqlQueryExecution(
                     }
                 }
             }
+
+            // Optionally build and validate plan immediately, before execution begins
+            planFuture = isQueryExecutionFailFastEnabled(getSession()) ? createLogicalPlanAsync() : null;
```
tdcmeehan, line 251: I'm wondering if we can just inline the `createLogicalPlanAsync` method, and instead, switch the executor based on the session property.

* If the flag is disabled, use `MoreExecutors.sameThreadExecutor()`
* If the flag is enabled, use a new executor which is injected into this class, and which we have metrics for.  Note we define the executor currently used [HERE](https://github.com/tdcmeehan/presto/blob/6072d373184ba802eb0a3fd6a4e0d211114c6956/presto-main/src/main/java/com/facebook/presto/server/CoordinatorModule.java#L263-L265), we could define a new one with more restricted paralellism, and which we could add dedicated metrics for (using a `QueryExecutionMBean` as you see in the link).  One thing we'd need to figure out is how to start the planner in the current thread if the queue were full, without a race, but I think this should be doable.

presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java
tdcmeehan: IIUC, I think queueing will be a more accurate term, because checking for resources happens after queueing.

tdcmeehan (changes requested):
We could try to write a test for this.

* We can use the blackhole connector to queue up queries that will never finish, without using a lot resources.
* One we hit the predetermined queue depth, we then check to see that the next query we send immediately fails.

presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java
BryanCutler: Sure, my point was that not all queries end up on the queue and might be started right away. In that case this property would still be used and could validate before waiting for resources completes. So I didn't want it to sound like this is only useful if workload has heavy queuing, but if `validate-plan-before-queueing` is the consensus, it's fine with me.

presto-main/src/main/java/com/facebook/presto/execution/SqlQueryExecution.java
BryanCutler: When flag is enabled, sure I can add a new executor

For when it's disabled, IIUC, using something like `MoreExecutors.directExecutor()` would cause the plan to be validated immediately in the constructor, which would change the default behavior - I don't think we want to do this. I could still inline `createLogicalPlanAsync()` if you want to make it more compact?

BryanCutler:
> We could try to write a test for this.
We can use the blackhole connector to queue up queries that will never finish, without using a lot resources.
One we hit the predetermined queue depth, we then check to see that the next query we send immediately fails.

Thanks @tdcmeehan, I can add a test, but is there a way to determine if a query has been queued or what the current queue size is?

tdcmeehan:
@BryanCutler take a look at `TestQueues`, which uses the BlackHole connector to deterministically queue, and which uses resource group configuration upfront.  If you know that your queries will hang (as the BlackHole connector does), and know the resource group configuration, you can know deterministically when your queries will start to queue.

presto-main/src/main/java/com/facebook/presto/execution/SqlQueryExecution.java
tdcmeehan: That's a good point, we wouldn't want that!  I suppose it should be as you said, if the flag is enabled, submit it to the new executor with restricted parallelism.  When we get to line 470, we should try to cancel it without interrupting it--if the cancellation succeeded or if the feature is not enabled, do it in the current thread, otherwise, `get()` the future.

presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java
tdcmeehan: Just explaining my thinking a bit here, rather than thinking of queueing as something that happens to a query, I think of it as a component in Presto.  But I see your point as this phrase may not be familiar to folks who are not Presto developers.

What about `eager-plan-validation-enabled`, and in the documentation we describe it more precisely?

BryanCutler: `eager-plan-validation-enabled` sgtm

presto-main/src/main/java/com/facebook/presto/execution/SqlQueryExecution.java
BryanCutler: So when you say try to cancel the future at line 470, is it just in case the ExecutorService hasn't started running the thread, and doing it the current thread would be quicker?

tdcmeehan: Yes, both quicker, and also it won't be subject to the concurrency restraints of the executor it would be running under.

rschlussel: also, let's make the executor service parallelism configurable.

BryanCutler: ok, good idea

presto-main/src/main/java/com/facebook/presto/execution/SqlQueryExecution.java
```diff
@@ -243,6 +250,17 @@ private SqlQueryExecution(
                     }
                 }
             }
+
+            // Optionally build and validate plan immediately, before execution begins
+            if (isEagerPlanValidationEnabled(getSession())) {
+                if (eagerPlanValidationExecutorInstance.get() == null) {
+                    eagerPlanValidationExecutorInstance.set(newFixedThreadPool(getEagerPlanValidationThreadPoolSize(getSession()), threadsNamed("presto-eager-plan-validation-%s")));
```
BryanCutler, line 257: I know this probably isn't the right way to setup the threadpool, but not sure the best way to create one where the size is taken from session property. Also, it should probably be conditional on `eager-plan-validation-enabled` being set. Any suggestions?

BryanCutler:
@tdcmeehan I had some trouble with the unit test, still working on that..

presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
```diff
@@ -3336,6 +3348,16 @@ public static boolean isRewriteExpressionWithConstantEnabled(Session session)
         return session.getSystemProperty(REWRITE_EXPRESSION_WITH_CONSTANT_EXPRESSION, Boolean.class);
     }
 
+    public static boolean isEagerPlanValidationEnabled(Session session)
+    {
+        return session.getSystemProperty(EAGER_PLAN_VALIDATION_ENABLED, Boolean.class);
+    }
+
+    public static int getEagerPlanValidationThreadPoolSize(Session session)
```
tdcmeehan, line 3356: We can keep this as a config property, since it should be static throughout the cluster's uptime.

presto-main/src/main/java/com/facebook/presto/execution/SqlQueryExecution.java
tdcmeehan: Let's not configure it from the session property at all.  We can always create the threadpool required for this feature.  For deployments that don't use it, to eliminate any overhead, we can just configure the threadpool to start at 0 (the core pool size).    We can configure the max pool size.

BryanCutler: Gotcha, I'm working on fixing it up

presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
BryanCutler: Ok, so I kept the config in  FeaturesConfig, let me know if there is a better place for it

presto-docs/src/main/sphinx/admin/properties.rst
```diff
@@ -53,6 +53,17 @@ output data set is not skewed in order to avoid the overhead of hashing and
 redistributing all the data across the network. This can also be specified
 on a per-query basis using the ``redistribute_writes`` session property.
 
+``eager-plan-validation-enabled``
+^^^^^^^^^^^^^^^^^^^^^^^
```
steveburnett, line 57:
```suggestion
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Formatting line should be the same number of characters as the line above that it is formatting. 

Local doc build shows the following new warnings (one misformatted line produces two warnings) as follows: 

```
/Users/steveburnett/Documents/GitHub/presto/presto-docs/src/main/sphinx/admin/properties.rst:57: WARNING: Title underline too short.

``eager-plan-validation-enabled``
^^^^^^^^^^^^^^^^^^^^^^^
/Users/steveburnett/Documents/GitHub/presto/presto-docs/src/main/sphinx/admin/properties.rst:57: WARNING: Title underline too short.

``eager-plan-validation-enabled``
^^^^^^^^^^^^^^^^^^^^^^^
```

steveburnett (changes requested):
Thanks for the doc! Nit of formatting, otherwise looks good.

tdcmeehan (changes requested):
Just some nits

presto-docs/src/main/sphinx/admin/properties.rst
BryanCutler: Thanks @steveburnett for catching that

steveburnett (dismissed):
LGTM! (docs)

Pull updated branch, new local doc build, `build succeeded, 13 warnings.` (as expected). 
Looks good, thanks!

BryanCutler:
I'm updating with another commit from https://github.com/prestodb/presto/pull/23596 which adds the SPI changes required for a custom plan checker. This is so that I can add a unit test with an added PlanChecker that can be set to fail plan validation and verify that it is done on a queued query.

presto-main/src/main/java/com/facebook/presto/execution/SqlQueryExecution.java
```diff
@@ -460,8 +471,13 @@ public void start()
                         Thread.currentThread(),
                         timeoutThreadExecutor,
                         getQueryAnalyzerTimeout(getSession()))) {
-                    // create logical plan for the query
-                    plan = createLogicalPlanAndOptimize();
+                    // If planFuture has not started, cancel and build plan in current thread
+                    if (planFuture != null && !planFutureLocked.compareAndSet(false, true)) {
```
BryanCutler, line 475: From the discussion at https://github.com/prestodb/presto/pull/23649#discussion_r1765583206, trying to call `planFuture.cancel(false)` will mark the future as cancelled, but the thread will continue to run. Using the AtomicBoolean `planFutureLocked` will effectively cancel the planFuture if set here first, then the future result (null) is discarded.

presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/PlanChecker.java
```diff
@@ -67,36 +80,88 @@ public PlanChecker(FeaturesConfig featuresConfig, boolean forceSingleNode)
...
         if (featuresConfig.isNativeExecutionEnabled() && (featuresConfig.isDisableTimeStampWithTimeZoneForNative() ||
                 featuresConfig.isDisableIPAddressForNative())) {
-            builder.put(Stage.INTERMEDIATE, new CheckUnsupportedPrestissimoTypes(featuresConfig));
+            checkers.put(Stage.INTERMEDIATE, new CheckUnsupportedPrestissimoTypes(featuresConfig));
         }
-        checkers = builder.build();
     }
 
-    public void validateFinalPlan(PlanNode planNode, Session session, Metadata metadata, WarningCollector warningCollector)
+    public synchronized void update(PlanCheckerProvider provider)
```
BryanCutler, line 90: @tdcmeehan comments from https://github.com/prestodb/presto/pull/23596/commits/486737b438220a1d4fa954e8a054e457bf77ad96#r1765240968

> Let's do the following:
>1. Let's make checkers volatile. This will be accessed in a multithreaded environment, and the volatile keyword will ensure visibility after updates across multiple threads.
>2. Let's remove providedInitialized, and instead add a method update which takes in a PlanCheckerProvider, which is called by the PlanCheckerProviderManager once the manager is initialized.
>3. Now that the flow of communication is reversed, we don't need ensureProvidedInitialized.
>4. Add a @GuardedBy("this") to the class to document the synchronization of this class.

My initial thoughts against this were because it leads to being able to create a new arbitrary `PlanChecker` or `PlanFragmenter` that could easily not be updated with the plugin provider. It's also where and how to call `update`, since you need to get an instance of the `PlanCheckerProviderManager` first. The changes in `LocalQueryRunner` show these extra steps.

presto-main/src/test/java/com/facebook/presto/cost/TestCostCalculator.java
```diff
@@ -151,7 +151,7 @@ public void setUp()
                 new SimpleTtlNodeSelectorConfig());
         PartitioningProviderManager partitioningProviderManager = new PartitioningProviderManager();
         nodePartitioningManager = new NodePartitioningManager(nodeScheduler, partitioningProviderManager, new NodeSelectionStats());
-        planFragmenter = new PlanFragmenter(metadata, nodePartitioningManager, new QueryManagerConfig(), new FeaturesConfig());
+        planFragmenter = new PlanFragmenter(metadata, nodePartitioningManager, new QueryManagerConfig(), new FeaturesConfig(), null);
```
rschlussel: we shouldn't allow PlanCheckerProviderManager to be null (would break things in a real environment). Lets add a requireNonNull check to PlanFragmenter, and update teh tests to create an empty PlanCheckerProviderManager.

presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/PlanChecker.java
```diff
@@ -17,27 +17,41 @@
...
-    public PlanChecker(FeaturesConfig featuresConfig)
+    public PlanChecker(FeaturesConfig featuresConfig, PlanCheckerProviderManager providerManager)
     {
-        this(featuresConfig, false);
+        this(featuresConfig, false, providerManager);
     }
 
     public PlanChecker(FeaturesConfig featuresConfig, boolean forceSingleNode)
     {
+        this(featuresConfig, forceSingleNode, null);
```
rschlussel: This should never be allowed to be null. We avoid nulls whenever possible in Presto. Instead have it be empty.

presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/PlanChecker.java
```diff
@@ -77,26 +91,95 @@ public PlanChecker(FeaturesConfig featuresConfig, boolean forceSingleNode)
...
+    private static List<Checker> fromSpi(List<com.facebook.presto.spi.plan.PlanChecker> checkers)
+    {
+        return checkers.stream().map(SpiCheckerAdapter::new).collect(Collectors.toList());
+    }
+
+    private void ensureProvidedInitialized()
     {
-        checkers.get(Stage.FRAGMENT).forEach(checker -> checker.validate(planNode, session, metadata, warningCollector));
+        if (!providedInitialized) {
+            if (providerMgr != null) {
```
rschlussel: ProviderManager should never be allowed to be null, and then you won't need this check.

presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/plancheckerprovidermanagers/PlanCheckerProviderManager.java
```diff
@@ -0,0 +1,82 @@
...
+        }
+        this.providerFactory = requireNonNull(providerFactory, "providerFactory is null");
+    }
+
+    public PlanCheckerProvider getPlanCheckerProvider()
+    {
+        return provider;
+    }
+
+    public void loadPlanCheckerProvider()
```
rschlussel, line 60: probably the config will be specific to each plan checker we support (similar to how it works for connectors or function registry), and then adding the configuration file is how you say you want to enable the plugin, and also how you set any configs for it.

presto-tests/src/test/java/com/facebook/presto/execution/TestQueues.java
```diff
@@ -374,6 +376,41 @@ public void testQueuedQueryInteraction()
...
+
+        Session firstSession = builder.setQueryId(QueryId.valueOf("20240930_203743_00001_11111")).build();
+        QueryId firstQuery = createQuery(queryRunner, firstSession, LONG_LASTING_QUERY);
+        waitForQueryState(queryRunner, firstQuery, RUNNING);
+
+        Session secondSession = builder.setQueryId(QueryId.valueOf("20240930_203743_00002_22222")).build();
+        QueryId secondQuery = createQuery(queryRunner, secondSession, LONG_LASTING_QUERY);
+        waitForQueryState(queryRunner, secondQuery, QUEUED);
+
+        Session thridSession = builder.setQueryId(QueryId.valueOf("20240930_203743_00003_33333")).build();
```
rschlussel, line 411: typo: thrid -> third

presto-tests/src/test/java/com/facebook/presto/execution/TestPlanCheckerProviderPlugin.java
```diff
@@ -0,0 +1,60 @@
...
+import com.facebook.presto.spi.plan.PlanCheckerProvider;
+import com.facebook.presto.spi.plan.PlanCheckerProviderFactory;
+import com.facebook.presto.spi.plan.SimplePlanFragmentSerde;
+import com.google.common.collect.ImmutableList;
+
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+public class TestPlanCheckerProviderPlugin
```
rschlussel, line 27: call this TestingPlanCheckerProviderPlugin (our classes containing tests are named TestXXX), and classes that are objects used in tests are called (TestingXXX)

rschlussel (changes requested):
i think some of my comments are spread across two PRs for the same changes.  sorry about that.

presto-main/src/test/java/com/facebook/presto/cost/TestCostCalculator.java
elharo: Do we routinely use Mockito or EasyMock for stubs like this?

rschlussel: No we don't use mocking in Presto.

presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/plancheckerprovidermanagers/PlanCheckerProviderManager.java
BryanCutler: I thought that specifying the plugin in the coordinator config was to enable it. This will currently load config from "etc/plan-checker-provider.properties" and provide those properties to the loaded plugin. That seemed to be how other plugin providers have the config setup. Are you saying you want `PlanCheckerProviderManager` to check the plan-checker-provider.properties to enable/disable the provided plugin?

BryanCutler:
Thanks for reviewing @rschlussel . The last commit removed the null checks with `PlanCheckerProviderManager` you brought up, let me know if this way looks better.

presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/plancheckerprovidermanagers/PlanCheckerProviderManager.java
rschlussel: i think i'm suggesting a variation, where we can have multiple plan checker providers, and then have configs like etc/plan-checker-providers/native-plan-checker.properties to add specific plan checker providers.  That way, for example, i could add the native plan checker, and also some custom plan checkers that are relevant only to my company's needs.

presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/PlanChecker.java
```diff
@@ -34,42 +36,39 @@
  */
 public final class PlanChecker
 {
-    private final PlanCheckerProviderManager providerMgr;
-    private Multimap<Stage, Checker> checkers;
-    private boolean providedInitialized;
+    @GuardedBy("this")
+    private volatile Multimap<Stage, Checker> checkers;
```
rschlussel, line 40: Synchronization here can be a performance problem when there are a lot of concurrent queries running. Since we're only updating once, make this a concurrentHashMap with a CopyOnWriteArrayList for the values, so you don't need to synchronize access. Also, make this final, and then you can remove the volatile keyword.

as an alternative, you could skip adding any plugin plan checkers to this map completely, and instead call to the PlanCheckerProvidermanager when you do the checks to get any extra plan checkers that they provide. (the suggestion from https://github.com/prestodb/presto/pull/23596#discussion_r1783477665)

presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/plancheckerprovidermanagers/PlanCheckerProviderManager.java
BryanCutler: Ah yes, I was wondering about having multiple plan checkers, so this sounds good to me. I'll fix this up so that can be supported.

BryanCutler: @rschlussel just want to make sure I have the semantics of loading multiple plan checker providers right:

1. For each plan checker provider plugin: multiple `PlanCheckerProviderFactory`s can be defined
2. For each `PlanCheckerProviderFactory`: the factory is added to the PlanCheckerProviderManager through `addPlanCheckerProviderFactory`, 
3. Calling `PlanCheckerProviderFactory.create()` the factory is responsible for loading configuration and creating a `PlanCheckerProvider`
4. The provider can use the configuration to determine if specific `PlanChecker`s are enabled/disabled etc.

Sound ok?

presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/PlanChecker.java
BryanCutler: I do prefer the second option, since it would just be iterating over the list of provided checkers so it wouldn't be adding overhead and it would certainly be cleaner. If I understand correctly this would make a check look something like this:

```java
    public synchronized void validateFinalPlan(PlanNode planNode, Session session, Metadata metadata, WarningCollector warningCollector)
    {
        checkers.get(Stage.FINAL).forEach(checker -> checker.validate(planNode, session, metadata, warningCollector));
        for (PlanCheckerProvider provider : planCheckerProviderManager.getPlanCheckerProviders()) {
            for (com.facebook.presto.spi.plan.PlanChecker checker : provider.getPlanCheckersFinal()) {
                checker.validate(planNode, warningCollector);
            }
        }
    }
``` 

That would require the `PlanChecker` keep the instance to the `PlanCheckerProviderManager`. I think @tdcmeehan preferred to not have that, so I'd like him to weigh in.

presto-spark-base/src/test/java/com/facebook/presto/spark/planner/TestIterativePlanFragmenter.java
```diff
@@ -224,7 +224,7 @@ private Void runTestIterativePlanFragmenter(PlanNode node, Plan plan, SubPlan fu
                 plan,
                 testingFragmentTracker::isFragmentFinished,
                 metadata,
-                new PlanChecker(new FeaturesConfig()),
+                new PlanChecker(new FeaturesConfig(), false),
```
BryanCutler, line 227: TODO: this could be reverted now

presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/PlanChecker.java
tdcmeehan: I don't think it occurred to me when I made my suggestion that even the read-only methods would need to be synchronized.

Re-reading this code, I'm wondering if we even need `PlanCheckerManager`, and if instead we should just do option 1 and have the existing `PlanChecker` class in `presto-main` subsume all of its responsibilities.  It seems the `PlanCheckerManager` is just an intermediary between the two, and it seems better consolidate it.

presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/plancheckerprovidermanagers/PlanCheckerProviderManager.java
tdcmeehan: This seems reasonable to me.

presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/PlanChecker.java
tdcmeehan: I just now remembered that there's a single node and distributed instance of `PlanChecker`, and that depends on a toggle that's known at runtime.  Agreed that option 2 looks like the best approach!

presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/PlanCheckerProviderManager.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import static java.lang.String.format;
+import static java.util.Objects.requireNonNull;
+
+public class PlanCheckerProviderManager
+{
+    private final SimplePlanFragmentSerde simplePlanFragmentSerde;
+    private final Map<String, PlanCheckerProviderFactory> providerFactories = new HashMap<>();
```
rschlussel, line 35: Our other discussion made me realize that this and the below need to use ConcurrentHashMap/CopyOnWriteArrayList.

presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/PlanCheckerProviderManager.java
```diff
@@ -0,0 +1,85 @@
...
+    private List<PlanCheckerProvider> providers = new ArrayList<>();
```
rschlussel, line 36: this should be final

presto-main/src/main/java/com/facebook/presto/sql/planner/sanity/PlanCheckerProviderManager.java
BryanCutler: Ok, but from what I understand the plugins are all installed when `PrestoServer` calls `PluginManager.installPlugins()`, is there a way that it's done from multiple threads?

rschlussel: I don't think it's ever updated from multiple threads, though we do read from it from multiple threads, so we care about visibility of the writes. And we don't make any guarantees about the writes so it's good practice to be safe.

BryanCutler: Sounds good, thanks for clarifying

presto-tests/src/test/java/com/facebook/presto/execution/TestQueues.java
```diff
@@ -374,6 +376,42 @@ public void testQueuedQueryInteraction()
...
 
+    @Test(timeOut = 240_000)
+    public void testEagerPlanValidation()
+            throws Exception
+    {
+        AtomicBoolean triggerValidationFailure = new AtomicBoolean();
+
+        queryRunner.installPlugin(new ResourceGroupManagerPlugin());
+        queryRunner.installPlugin(new TestingPlanCheckerProviderPlugin(triggerValidationFailure));
+        queryRunner.getPlanCheckerProviderManager().loadPlanCheckerProviders();
```
BryanCutler, line 387: I'm not sure if there is a better way to call `loadPlanCheckerProviders()`. In `PrestoServer` it is called during `run()`, but for `TestingPrestoServer` there isn't really an equivalent. It could just be loaded as needed though..

rschlussel: this seems fine

steveburnett (approved):
LGTM! (docs)

presto-main/src/main/java/com/facebook/presto/execution/SqlQueryExecution.java
```diff
@@ -538,7 +554,7 @@ private PlanRoot createLogicalPlanAndOptimize()
                     .getRuntimeStats()
                     .profileNanos(
                             LOGICAL_PLANNER_TIME_NANOS,
-                            () -> queryAnalyzer.plan(this.analyzerContext, queryAnalysis));
+                            () -> queryAnalyzer.plan(this.analyzerContext, queryAnalysis));;
```
rschlussel: typo: 2 semicolons

presto-main/src/main/java/com/facebook/presto/execution/SqlQueryExecution.java
```diff
@@ -554,7 +554,7 @@ private PlanRoot createLogicalPlanAndOptimize()
                     .getRuntimeStats()
                     .profileNanos(
                             LOGICAL_PLANNER_TIME_NANOS,
-                            () -> queryAnalyzer.plan(this.analyzerContext, queryAnalysis));;
```
rschlussel: this fix should be in the previous commit

presto-spi/src/main/java/com/facebook/presto/spi/plan/PlanCheckerProvider.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+package com.facebook.presto.spi.plan;
+
+import java.util.Collections;
+import java.util.List;
+
+public interface PlanCheckerProvider
+{
+    default List<PlanChecker> getPlanCheckersIntermediate()
```
rschlussel, line 22: nit: I think the below sounds more natural

```suggestion
    default List<PlanChecker> getIntermediatePlanCheckers()
```

presto-spi/src/main/java/com/facebook/presto/spi/plan/PlanCheckerProvider.java
```diff
@@ -0,0 +1,36 @@
...
+    {
+        return Collections.emptyList();
+    }
+
+    default List<PlanChecker> getPlanCheckersFinal()
```
rschlussel, line 27:
```suggestion
    default List<PlanChecker> getFinalPlanCheckers()
```

presto-spi/src/main/java/com/facebook/presto/spi/plan/PlanCheckerProvider.java
```diff
@@ -0,0 +1,36 @@
...
+    {
+        return Collections.emptyList();
+    }
+
+    default List<PlanChecker> getPlanCheckersFragment()
```
rschlussel, line 32:
```suggestion
    default List<PlanChecker> getFragmentPlanCheckers()
```

presto-main/src/main/java/com/facebook/presto/execution/SqlQueryExecution.java
```diff
@@ -590,6 +606,24 @@ private PlanRoot createLogicalPlanAndOptimize()
...
+                return createLogicalPlanAndOptimize();
+            }
+            return null;
+        }
+        catch (Throwable e) {
+            fail(e);
+            if (e instanceof Error) {
+                throw (Error) e;
+            }
+            throw new GenericInternalException(e);
```
rschlussel, line 623: why throw a GenericInternalException here?
1. that's really just for presto-common where PrestoExceptions aren't available. Otherwise prefer throwing a PrestoException with appropriate error code.
2. why isn't fail() sufficient to fail the query? And could throwing this exception get in the way here of getting the right error code?

presto-tests/src/test/java/com/facebook/presto/execution/TestQueues.java
```diff
@@ -374,6 +376,42 @@ public void testQueuedQueryInteraction()
...
+        Session secondSession = builder.setQueryId(QueryId.valueOf("20240930_203743_00002_22222")).build();
+        QueryId secondQuery = createQuery(queryRunner, secondSession, LONG_LASTING_QUERY);
+        waitForQueryState(queryRunner, secondQuery, QUEUED);
+
+        Session thirdSession = builder.setQueryId(QueryId.valueOf("20240930_203743_00003_33333")).build();
+        QueryId thirdQuery = createQuery(queryRunner, thirdSession, LONG_LASTING_QUERY);
+
+        // Force failure during plan validation after queuing has begun
+        triggerValidationFailure.set(true);
+        waitForQueryState(queryRunner, thirdQuery, FAILED);
```
rschlussel, line 409: maybe also check the error code to make sure the error handling is correct.

rschlussel (commented):
A few comments, but the new changes look good.  

Can you clean up the commits. right now the later commits fix things for earlier commits, and we want the commits to each be independently good (e.g. earlier commit uses volatile synchronized map, and later commit does the new approach, earlier commit supports one plan checker, later commit supports multiple).  Also, make sure the commit messages follow our guidelnes (titles in the imperative mood, and body with more description and any context)

presto-main/src/main/java/com/facebook/presto/execution/SqlQueryExecution.java
BryanCutler: Yes, fail() is enough to fail the query. We need to throw again in case `planFuture.get()` is called during`SqlQueryExecution.start()`, then it will stop with the exception. We don't need `GenericInternalException`, I can clean this up.

BryanCutler:
Thanks @rschlussel , I'll work on getting this fixed up and cleanup the commits today.

tdcmeehan (commented):
`Add property for eager building of plan` ✅

tdcmeehan (commented):
`SPI to support plugin for custom plan checkers` ✅

tdcmeehan (approved):
`Unit test to check eager plan validation` ✅

rschlussel (commented):
two small things about the commits. otherwise looks good.

1. commit titles should be in the imperative
SPI to support plugin for custom plan checkers -> Add SPI to support plugin for custom plan checkers.

2. Tests should be added with the code they are testing, so squash the unit test commit into the previous one that adds the spi.

rschlussel (approved):

BryanCutler:
Thanks @rschlussel @tdcmeehan and @steveburnett !

PR #23663, opened 2024-09-16
Add QueryType to AccessControlContext

RohanSidhu:
## Description
Adding a QueryType field into the AccessControlContext object to allow for this information to be available when making access control related decisions.

## Motivation and Context
The addition of QueryType allows for downstream function calls to access this context which can be used to alter decision making if needed.

## Impact
Minor performance enhancement as rebuilding Session is no longer needed. If additional context needs to be added into Session, we can utilize SessionBuilder to add additional/edit material.

## Test Plan
Tested code change by running sample queries and followed the call stack to ensure queryType was present.
Screenshot 2024-09-16 at 13 10 19
Also verified the context was present in the AccessControlContext through unit tests under AbstractTestQueries.

## Contributor checklist

- [X] CI passed.

## Release Notes
```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/Session.java
+++ b/presto-main/src/main/java/com/facebook/presto/Session.java
@@ -15,6 +15,7 @@ package com.facebook.presto;
 
 import com.facebook.presto.common.RuntimeStats;
 import com.facebook.presto.common.function.SqlFunctionProperties;
+import com.facebook.presto.common.resourceGroups.QueryType;
 import com.facebook.presto.common.transaction.TransactionId;
 import com.facebook.presto.common.type.TimeZoneKey;
 import com.facebook.presto.cost.PlanCostEstimate;
@@ -99,6 +100,7 @@ public final class Session
     private final Optional<Tracer> tracer;
     private final WarningCollector warningCollector;
     private final RuntimeStats runtimeStats;
+    private final Optional<QueryType> queryType;
 
     private final OptimizerInformationCollector optimizerInformationCollector = new OptimizerInformationCollector();
     private final OptimizerResultCollector optimizerResultCollector = new OptimizerResultCollector();
@@ -131,7 +133,8 @@ public final class Session
             Map<SqlFunctionId, SqlInvokedFunction> sessionFunctions,
             Optional<Tracer> tracer,
             WarningCollector warningCollector,
-            RuntimeStats runtimeStats)
+            RuntimeStats runtimeStats,
+            Optional<QueryType> queryType)
     {
         this.queryId = requireNonNull(queryId, "queryId is null");
         this.transactionId = requireNonNull(transactionId, "transactionId is null");
@@ -172,7 +175,8 @@ public final class Session
         this.tracer = requireNonNull(tracer, "tracer is null");
         this.warningCollector = requireNonNull(warningCollector, "warningCollector is null");
         this.runtimeStats = requireNonNull(runtimeStats, "runtimeStats is null");
-        this.context = new AccessControlContext(queryId, clientInfo, clientTags, source, warningCollector, runtimeStats);
+        this.queryType = requireNonNull(queryType, "queryType is null");
+        this.context = new AccessControlContext(queryId, clientInfo, clientTags, source, warningCollector, runtimeStats, queryType);
     }
 
     public QueryId getQueryId()
@@ -353,6 +357,11 @@ public final class Session
         return planNodeCostMap;
     }
 
+    public Optional<QueryType> getQueryType()
+    {
+        return queryType;
+    }
+
     public Session beginTransactionId(TransactionId transactionId, TransactionManager transactionManager, AccessControl accessControl)
     {
         requireNonNull(transactionId, "transactionId is null");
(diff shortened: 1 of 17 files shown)


ghelmling (commented):
Can you add test coverage for DispatchManager.createQuery(Internal) to validate that the session is getting populated correctly?  I think you could plug in a stubbed out DispatchQueryFactory to validate the Session instance passed to createDispatchQuery().

presto-main/src/main/java/com/facebook/presto/dispatcher/DispatchManager.java
```diff
@@ -303,6 +306,11 @@ private <C> void createQueryInternal(QueryId queryId, String slug, int retryCoun
                     Optional.ofNullable(sessionContext.getSchema()),
                     sessionContext.getIdentity().getPrincipal().map(Principal::getName)));
 
+            session = sessionBuilder.build();
+            if (sessionContext.getTransactionId().isPresent()) {
```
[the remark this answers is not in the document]
RohanSidhu: I reviewed the tasks that happen within beginTransactionId and I don't believe there is any dependency between what happens in there and the work done to create preparedQuery. Implying that the order does not matter on if we prepare the query or if we begin the transaction. From testing, I also found no impact in performance or functionality.

presto-main/src/main/java/com/facebook/presto/dispatcher/DispatchManager.java
```diff
@@ -268,16 +270,17 @@ private <C> void createQueryInternal(QueryId queryId, String slug, int retryCoun
...
 
             // decode session
-            session = sessionSupplier.createSession(queryId, sessionContext, warningCollectorFactory);
+            sessionBuilder = sessionSupplier.createSessionBuilder(queryId, sessionContext, warningCollectorFactory);
 
             // prepare query
-            AnalyzerOptions analyzerOptions = createAnalyzerOptions(session, session.getWarningCollector());
-            QueryPreparerProvider queryPreparerProvider = queryPreparerProviderManager.getQueryPreparerProvider(getAnalyzerType(session));
-            preparedQuery = queryPreparerProvider.getQueryPreparer().prepareQuery(analyzerOptions, query, session.getPreparedStatements(), session.getWarningCollector());
+            AnalyzerOptions analyzerOptions = createAnalyzerOptions(sessionBuilder.build(), sessionBuilder.getWarningCollector());
```
ghelmling, line 276: We're creating new Session instances here and  on the next line.  I think it would be better to use a local variable:
Session tempSession = sessionBuilder.build() 

and pass to both.

ghelmling (commented):
I noticed another bit in DispatchManager where we should consolidate the Session creation.  Sorry for the add-on, but would be good to clean up some of the duplicate creation that existed.

presto-main/src/main/java/com/facebook/presto/Session.java
```diff
@@ -131,7 +133,8 @@ public Session(
             Map<SqlFunctionId, SqlInvokedFunction> sessionFunctions,
             Optional<Tracer> tracer,
             WarningCollector warningCollector,
-            RuntimeStats runtimeStats)
+            RuntimeStats runtimeStats,
+            Optional<QueryType> queryType)
```
NikhilCollooru, line 137: why is this optional ? can a session have this empty ?

presto-main/src/main/java/com/facebook/presto/dispatcher/DispatchManager.java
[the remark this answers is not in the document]
NikhilCollooru: why are we starting a transaction here ?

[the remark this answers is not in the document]
RohanSidhu: The transaction was never started after I refactored the code to use SessionBuilder instead of Session. In the original construction of Session, the transaction was started once Session was built. Since we need queryType in the session, I moved out the transaction piece out so we could utilize SessionBuilder to add the necessary field. Since the transaction hasn't started, I moved it out and placed it as early as I could while including the addition of queryType.

presto-main/src/main/java/com/facebook/presto/Session.java
RohanSidhu: I don't believe there is an instance where queryType can be empty. If my understanding is correct, all queries that make it to DispatchManager will have some enum of QueryType. However, as a precaution, I wrapped QueryType in an optional.

NikhilCollooru: so then lets have it as NOT optional. And fix any issues that my arise.

RohanSidhu: After making the changes, I realized that there are some areas where I put an empty optional instead of null.
Example would be here: https://github.com/prestodb/presto/blob/master/presto-main/src/main/java/com/facebook/presto/SessionRepresentation.java#L308

Since the [session ctor](https://github.com/prestodb/presto/blob/33ff9521dbd25e2b7b1181cf01391cb899857d65/presto-main/src/main/java/com/facebook/presto/Session.java#L109) requires all fields to be non-null, I believe it may be the design appropriate choice to keep the queryType as an Optional.

NikhilCollooru: Why are we not adding QueryType to SessionRepresentation as well ? Then we do need NOT send Optional.empty() or null right ?

presto-main/src/main/java/com/facebook/presto/dispatcher/DispatchManager.java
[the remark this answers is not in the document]
NikhilCollooru: i think the transaction has to begin before you call analyzer ..so this is too late

presto-main/src/main/java/com/facebook/presto/dispatcher/DispatchManager.java
```diff
@@ -268,16 +270,18 @@ private <C> void createQueryInternal(QueryId queryId, String slug, int retryCoun
...
-            AnalyzerOptions analyzerOptions = createAnalyzerOptions(session, session.getWarningCollector());
+            AnalyzerOptions analyzerOptions = createAnalyzerOptions(session, sessionBuilder.getWarningCollector());
             QueryPreparerProvider queryPreparerProvider = queryPreparerProviderManager.getQueryPreparerProvider(getAnalyzerType(session));
-            preparedQuery = queryPreparerProvider.getQueryPreparer().prepareQuery(analyzerOptions, query, session.getPreparedStatements(), session.getWarningCollector());
+            preparedQuery = queryPreparerProvider.getQueryPreparer().prepareQuery(analyzerOptions, query, sessionBuilder.getPreparedStatements(), sessionBuilder.getWarningCollector());
             query = preparedQuery.getFormattedQuery().orElse(query);
 
             // select resource group
             Optional<QueryType> queryType = preparedQuery.getQueryType();
+            sessionBuilder.setQueryType(queryType);
```
NikhilCollooru, line 284: why make all these changes instead when you can just do 
`session = Session.builder(session).setQueryType(queryType).build();`

RohanSidhu: This is because of this [line](https://github.com/prestodb/presto/blob/2f4ea64d3bf3fa8c89b3e805d5d0987ac6d971c2/presto-main/src/main/java/com/facebook/presto/Session.java#L644). My change makes it so we still have access to the SessionBuilder object before we begin the transaction. After we start the transaction, we can't use the ctor that passes in Session. I could use the other [ctor](https://github.com/prestodb/presto/blob/2f4ea64d3bf3fa8c89b3e805d5d0987ac6d971c2/presto-main/src/main/java/com/facebook/presto/Session.java#L598) but would need to refill in all the Session fields.

presto-main/src/main/java/com/facebook/presto/Session.java
RohanSidhu: After making the change in SessionRepresentation and other places, there is another issue if we don't use an Optional. queryType is not defined when we build session. We pass in session in a few [places](https://github.com/prestodb/presto/blob/2f4ea64d3bf3fa8c89b3e805d5d0987ac6d971c2/presto-main/src/main/java/com/facebook/presto/dispatcher/DispatchManager.java#L274) before we even have queryType. So if we want to maintain the non-null rule in session ctor, I believe using an Optional is the best bet.

RohanSidhu: It is a little confusing but I am more than happy to discuss the thought process!

ghelmling: @NikhilCollooru the challenge here is the Session usage in DispatchManager#createQueryInternal:
1. Session is instantiated and passed to createAnalyzerOptions()
2. Same session is passed to getAnalyzerType()
3. The SQL string is only parsed on the following line in QueryPreparer#prepareQuery

So currently session is being used before we know what the QueryType actually is.  Now the Session usage in (1) and (2) is only to extract system property values, so we could try to separate that concern from the Session representation itself, pulling out SystemPropertyManager and the Map<String, String> of system properties.  But that looks like a big refactoring as the use of Session as an interface to system properties seems pretty deeply embedded.

Any thoughts on how we can better clean this up?  The current use of Session in DispatchManager#createQueryInternal is a bit wonky to begin with as Session is first created from SessionSupplier (not the builder for some reason), then recreated by SessionPropertyDefaults#newSessionQithDefaultProperties.  I don't really understand why the SessionPropertyDefaults part happens _after_ the analyzer and query preparer setup.

NikhilCollooru:
I took a closer look and the changes look okay to me. 

1. please squash all commits into one commit 
2. Please click "Please click here to be authorized" to get CLA authorized
3. Please build a custom presto version and run verifier tests to confirm the changes run fine for a bigger set of queries

RohanSidhu:
I have completed the verifier tests and did not find any significant failures. I ran over 10000 queries and only 12 have failed. Reasons for their failures are environment errors (i.e. com.facebook.presto.ExceededMemoryLimitException). If there is anything else that I needed or proof for the test, please let me know!

NikhilCollooru (approved):

NikhilCollooru (approved):

PR #23671, opened 2024-09-18
test(native): Add `AbstractTestQueries` to native tests

pramodsatya:
## Description
Adds support to run testcases from `AbstractTestQueries` with the C++ worker in `presto-native-tests`. These testcases are run with Presto's iterative optimizer disabled in `TestNonIterativeDistributedQueries` and with the default native hive query runner in `TestTpchDistributedQueries`. 

## Motivation and Context
Improves Presto C++ e2e test coverage with sidecar by extending production tests from `presto-tests` module to use the native worker.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Add testcases from `AbstractTestQueries` to `presto-native-tests`.
```

### DIFF
--- a/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/NativeQueryRunnerUtils.java
+++ b/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/NativeQueryRunnerUtils.java
@@ -24,12 +24,14 @@ import static com.facebook.presto.nativeworker.PrestoNativeQueryRunnerUtils.ICEB
 
 public class NativeQueryRunnerUtils
 {
+    private static final String DEFAULT_STORAGE_FORMAT = "DWRF";
+
     private NativeQueryRunnerUtils() {}
 
     public static Map<String, String> getNativeWorkerHiveProperties()
     {
         return ImmutableMap.of("hive.parquet.pushdown-filter-enabled", "true",
-                "hive.orc-compression-codec", "ZSTD", "hive.storage-format", "DWRF");
+                "hive.orc-compression-codec", "ZSTD", "hive.storage-format", DEFAULT_STORAGE_FORMAT);
     }
 
     public static Map<String, String> getNativeWorkerIcebergProperties()
@@ -78,12 +80,32 @@ public class NativeQueryRunnerUtils
      */
     public static void createAllTables(QueryRunner queryRunner)
     {
-        createAllTables(queryRunner, "DWRF");
+        createAllTables(queryRunner, DEFAULT_STORAGE_FORMAT, false);
     }
 
     public static void createAllTables(QueryRunner queryRunner, String storageFormat)
     {
-        createLineitem(queryRunner, storageFormat);
+        createAllTables(queryRunner, storageFormat, false);
+    }
+
+    /**
+     * Creates all tables for local testing, except for bench tables.
+     *
+     * @param queryRunner the QueryRunner instance used to execute table creation queries.
+     * @param storageFormat File format to use for the tables. If "DWRF" is specified, date columns are cast to
+     *                      VARCHAR, since DWRF does not support DATE type.
+     * @param useTpchStandardSchema when true, lineitem table is created with 16 columns as specified in the TPC-H
+     *                              standard; otherwise, the lineitem table is created with additional columns for
+     *                              testing purposes.
+     */
+    public static void createAllTables(QueryRunner queryRunner, String storageFormat, boolean useTpchStandardSchema)
+    {
+        if (useTpchStandardSchema) {
+            createLineitemStandard(queryRunner, storageFormat);
+        }
+        else {
+            createLineitem(queryRunner, storageFormat);
+        }
         createOrders(queryRunner, storageFormat);
         createOrdersEx(queryRunner);
         createOrdersHll(queryRunner);
@@ -107,7 +129,7 @@ public class NativeQueryRunnerUtils
      */
     public static void createAllIcebergTables(QueryRunner queryRunner)
     {
-        createLineitemStandard(queryRunner);
+        createLineitemStandard(queryRunner, ICEBERG_DEFAULT_STORAGE_FORMAT);
         createOrders(queryRunner);
         createNationWithFormat(queryRunner, ICEBERG_DEFAULT_STORAGE_FORMAT);
         createCustomer(queryRunner);
@@ -119,7 +141,7 @@ public class NativeQueryRunnerUtils
 
     public static void createLineitem(QueryRunner queryRunner)
     {
-        createLineitem(queryRunner, "DWRF");
+        createLineitem(queryRunner, DEFAULT_STORAGE_FORMAT);
     }
 
     public static void createLineitem(QueryRunner queryRunner, String storageFormat)
@@ -157,16 +179,32 @@ public class NativeQueryRunnerUtils
 
     public static void createLineitemStandard(QueryRunner queryRunner)
     {
-        createLineitemStandard(queryRunner.getDefaultSession(), queryRunner);
+        createLineitemStandard(queryRunner.getDefaultSession(), queryRunner, DEFAULT_STORAGE_FORMAT);
     }
 
     public static void createLineitemStandard(Session session, QueryRunner queryRunner)
     {
+        createLineitemStandard(session, queryRunner, DEFAULT_STORAGE_FORMAT);
+    }
+
+    public static void createLineitemStandard(QueryRunner queryRunner, String storageFormat)
+    {
+        createLineitemStandard(queryRunner.getDefaultSession(), queryRunner, storageFormat);
+    }
+
+    public static void createLineitemStandard(Session session, QueryRunner queryRunner, String storageFormat)
+    {
         if (!queryRunner.tableExists(session, "lineitem")) {
+            boolean castDateToVarchar = storageFormat.equals("DWRF");
+            String shipDate = castDateToVarchar ? "cast(shipdate as varchar) as shipdate" : "shipdate";
+            String commitDate = castDateToVarchar ? "cast(commitdate as varchar) as commitdate" : "commitdate";
+            String receiptDate = castDateToVarchar ? "cast(receiptdate as varchar) as receiptdate" : "receiptdate";
+
+            queryRunner.execute("DROP TABLE IF EXISTS lineitem");
             queryRunner.execute(session, "CREATE TABLE lineitem AS " +
                     "SELECT orderkey, partkey, suppkey, linenumber, quantity, extendedprice, discount, tax, " +
-                    "   returnflag, linestatus, cast(shipdate as varchar) as shipdate, cast(commitdate as varchar) as commitdate, " +
-                    "   cast(receiptdate as varchar) as receiptdate, shipinstruct, shipmode, comment " +
+                    "   returnflag, linestatus, " + shipDate + ", " + commitDate + ", " + receiptDate + ", " +
+                    "   shipinstruct, shipmode, comment " +
                     "FROM tpch.tiny.lineitem");
         }
     }
--- a/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/NativeQueryRunnerUtils.java
+++ b/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/NativeQueryRunnerUtils.java
@@ -200,7 +200,6 @@ public class NativeQueryRunnerUtils
             String commitDate = castDateToVarchar ? "cast(commitdate as varchar) as commitdate" : "commitdate";
             String receiptDate = castDateToVarchar ? "cast(receiptdate as varchar) as receiptdate" : "receiptdate";
 
-            queryRunner.execute("DROP TABLE IF EXISTS lineitem");
             queryRunner.execute(session, "CREATE TABLE lineitem AS " +
                     "SELECT orderkey, partkey, suppkey, linenumber, quantity, extendedprice, discount, tax, " +
                     "   returnflag, linestatus, " + shipDate + ", " + commitDate + ", " + receiptDate + ", " +
--- a/presto-native-tests/src/test/java/com/facebook/presto/nativetests/AbstractTestQueriesNative.java
+++ b/presto-native-tests/src/test/java/com/facebook/presto/nativetests/AbstractTestQueriesNative.java
@@ -1,0 +1,1172 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.nativetests;
+
+import com.facebook.presto.Session;
+import com.facebook.presto.testing.MaterializedResult;
+import com.facebook.presto.testing.MaterializedRow;
+import com.facebook.presto.tests.AbstractTestQueries;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableMultimap;
+import com.google.common.collect.Maps;
+import com.google.common.collect.Multimaps;
+import org.intellij.lang.annotations.Language;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+import java.util.List;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+import static com.facebook.presto.SystemSessionProperties.EXPRESSION_OPTIMIZER_NAME;
+import static com.facebook.presto.SystemSessionProperties.FIELD_NAMES_IN_JSON_CAST_ENABLED;
+import static com.facebook.presto.SystemSessionProperties.JOIN_PREFILTER_BUILD_SIDE;
+import static com.facebook.presto.SystemSessionProperties.MERGE_AGGREGATIONS_WITH_AND_WITHOUT_FILTER;
+import static com.facebook.presto.SystemSessionProperties.REMOVE_MAP_CAST;
+import static com.facebook.presto.SystemSessionProperties.REMOVE_REDUNDANT_CAST_TO_VARCHAR_IN_JOIN;
+import static com.facebook.presto.SystemSessionProperties.REWRITE_MIN_MAX_BY_TO_TOP_N;
+import static com.facebook.presto.testing.assertions.Assert.assertEquals;
+import static com.facebook.presto.tests.QueryAssertions.assertEqualsIgnoreOrder;
+import static org.testng.Assert.assertNotEquals;
+import static org.testng.Assert.assertTrue;
+import static org.testng.Assert.fail;
+
+public abstract class AbstractTestQueriesNative
+        extends AbstractTestQueries
+{
+    private boolean sidecarEnabled;
+
+    private @Language("RegExp") String applyNotRegisteredError;
+    private @Language("RegExp") String createHllFunctionUnsupportedError;
+    private @Language("RegExp") String charTypeUnsupportedError;
+    private @Language("RegExp") String unnestRowsInvalidFieldError;
+
+    public void init(boolean sidecarEnabled)
+    {
+        this.sidecarEnabled = sidecarEnabled;
+
+        if (sidecarEnabled) {
+            applyNotRegisteredError = ".*Function native.default.apply not registered*";
+            charTypeUnsupportedError = ".*Unknown type: char.*";
+            createHllFunctionUnsupportedError = ".*Function native.default.create_hll not registered.*";
+            unnestRowsInvalidFieldError = "Field not found: field(?:_\\d+)?. Available fields are: field.*";
+        }
+        else {
+            applyNotRegisteredError = ".*Scalar function name not registered: presto.default.apply.*";
+            charTypeUnsupportedError = ".*Failed to parse type \\[char\\(.*\\)].*";
+            createHllFunctionUnsupportedError = ".*Scalar function name not registered: presto.default.create_hll, called with arguments.*";
+            unnestRowsInvalidFieldError = "Field not found: field(?:_\\d+)?. Available fields are: field.*";
+        }
+    }
+
+    @DataProvider(name = "use_default_literal_coalesce")
+    public static Object[][] useDefaultLiteralCoalesce()
+    {
+        return new Object[][] {{true}};
+    }
+
+    @Override
+    @DataProvider(name = "optimize_hash_generation")
+    public Object[][] optimizeHashGeneration()
+    {
+        return new Object[][] {{"false"}};
+    }
+
+    /// This test is not applicable in Presto C++ since there is no bytecode IR as with JVM.
+    @Override
+    @Test(enabled = false)
+    public void testLargeBytecode() {}
+
+    /// TODO: The decimal coercion queries are flaky in Presto C++.
+    @Override
+    @Test(enabled = false)
+    public void testCoercions() {}
+
+    /// TODO: Enabling session property merge_aggregations_with_and_without_filter in Presto C++ clusters without the
+    /// sidecar results in incorrect results. See issue https://github.com/prestodb/presto/issues/26323.
+    @Override
+    @Test
+    public void testSameAggregationWithAndWithoutFilter()
+    {
+        if (sidecarEnabled) {
+            super.testSameAggregationWithAndWithoutFilter();
+        }
+        else {
+            Session enableOptimization = Session.builder(getSession())
+                    .setSystemProperty(MERGE_AGGREGATIONS_WITH_AND_WITHOUT_FILTER, "true")
+                    .build();
+            Session disableOptimization = Session.builder(getSession())
+                    .setSystemProperty(MERGE_AGGREGATIONS_WITH_AND_WITHOUT_FILTER, "false")
+                    .build();
+
+            computeActual(getQueryRunner(), enableOptimization, "select regionkey, count(name) filter (where name like '%N%') n_nations, count(name) all_nations from nation group by regionkey");
+            assertQuery(enableOptimization, "select count(name) filter (where name like '%N%') n_nations, count(name) all_nations from nation", "values (15,25)");
+            assertQuery(enableOptimization, "select count(1), count(1) filter (where k > 5) from (values 1, null, 3, 5, null, 8, 10) t(k)", "values (7, 2)");
+
+            String sql = "select regionkey, count(name) filter (where name like '%N%') n_nations, count(name) all_nations from nation group by regionkey";
+            // MaterializedResult resultWithOptimization = computeActual(enableOptimization, sql);
+            // MaterializedResult resultWithoutOptimization = computeActual(disableOptimization, sql);
+            // assertEqualsIgnoreOrder(resultWithOptimization, resultWithoutOptimization);
+
+            sql = "select count(name) filter (where name like '%N%') n_nations, count(name) all_nations from nation";
+            MaterializedResult resultWithOptimization = computeActual(enableOptimization, sql);
+            MaterializedResult resultWithoutOptimization = computeActual(disableOptimization, sql);
+            assertEqualsIgnoreOrder(resultWithOptimization, resultWithoutOptimization);
+
+            sql = "select partkey, sum(quantity), sum(quantity) filter (where discount > 0.1) from lineitem group by grouping sets((), (partkey))";
+            resultWithOptimization = computeActual(enableOptimization, sql);
+            resultWithoutOptimization = computeActual(disableOptimization, sql);
+            assertEqualsIgnoreOrder(resultWithOptimization, resultWithoutOptimization);
+
+            // multiple aggregations in query
+            sql = "select partkey, sum(quantity), sum(quantity) filter (where discount < 0.05), sum(linenumber), sum(linenumber) filter (where discount < 0.05) from lineitem group by partkey";
+            resultWithOptimization = computeActual(enableOptimization, sql);
+            resultWithoutOptimization = computeActual(disableOptimization, sql);
+            assertEqualsIgnoreOrder(resultWithOptimization, resultWithoutOptimization);
+            // aggregations in multiple levels
+            sql = "select partkey, avg(sum), avg(sum) filter (where tax < 0.05), avg(filtersum) from (select partkey, suppkey, sum(quantity) sum, sum(quantity) filter (where discount > 0.05) filtersum, max(tax) tax from lineitem where partkey=1598 group by partkey, suppkey) t group by partkey";
+            resultWithOptimization = computeActual(enableOptimization, sql);
+            resultWithoutOptimization = computeActual(disableOptimization, sql);
+            assertEqualsIgnoreOrder(resultWithOptimization, resultWithoutOptimization);
+            // global aggregation
+            sql = "select sum(quantity), sum(quantity) filter (where discount < 0.05) from lineitem";
+            resultWithOptimization = computeActual(enableOptimization, sql);
+            resultWithoutOptimization = computeActual(disableOptimization, sql);
+            assertEqualsIgnoreOrder(resultWithOptimization, resultWithoutOptimization);
+            // order by
+            sql = "select partkey, array_agg(suppkey order by suppkey), array_agg(suppkey order by suppkey) filter (where discount > 0.05) from lineitem group by partkey";
+            resultWithOptimization = computeActual(enableOptimization, sql);
+            resultWithoutOptimization = computeActual(disableOptimization, sql);
+            assertEqualsIgnoreOrder(resultWithOptimization, resultWithoutOptimization);
+            // grouping sets
+            sql = "SELECT partkey, suppkey, sum(quantity), sum(quantity) filter (where discount > 0.05) from lineitem group by grouping sets((), (partkey), (partkey, suppkey))";
+            resultWithOptimization = computeActual(enableOptimization, sql);
+            resultWithoutOptimization = computeActual(disableOptimization, sql);
+            assertEqualsIgnoreOrder(resultWithOptimization, resultWithoutOptimization);
+            // aggregation over union
+            sql = "SELECT partkey, sum(quantity), sum(quantity) filter (where orderkey > 0) from (select quantity, orderkey, partkey from lineitem union all select totalprice as quantity, orderkey, custkey as partkey from orders) group by partkey";
+            resultWithOptimization = computeActual(enableOptimization, sql);
+            resultWithoutOptimization = computeActual(disableOptimization, sql);
+            assertEqualsIgnoreOrder(resultWithOptimization, resultWithoutOptimization);
+            // aggregation over join
+            sql = "select custkey, sum(quantity), sum(quantity) filter (where tax < 0.05) from lineitem l join orders o on l.orderkey=o.orderkey group by custkey";
+            resultWithOptimization = computeActual(enableOptimization, sql);
+            resultWithoutOptimization = computeActual(disableOptimization, sql);
+            assertEqualsIgnoreOrder(resultWithOptimization, resultWithoutOptimization);
+        }
+    }
+
+    /// TODO: The last query in this testcase is flaky in Presto C++.
+    @Override
+    @Test
+    public void testLambdaInAggregation()
+    {
+        assertQuery("SELECT id, reduce_agg(value, 0, (a, b) -> a + b+0, (a, b) -> a + b) FROM ( VALUES (1, 2), (1, 3), (1, 4), (2, 20), (2, 30), (2, 40) ) AS t(id, value) GROUP BY id", "values (1, 9), (2, 90)");
+        assertQuery("SELECT id, 's' || reduce_agg(value, '', (a, b) -> concat(a, b, 's'), (a, b) -> concat(a, b, 's')) FROM ( VALUES (1, '2'), (1, '3'), (1, '4'), (2, '20'), (2, '30'), (2, '40') ) AS t(id, value) GROUP BY id",
+                "values (1, 's2s3ss4ss'), (2, 's20s30ss40ss')");
+        // assertQuery("SELECT id, reduce_agg(value, array[id, value], (a, b) -> a || b, (a, b) -> a || b) FROM ( VALUES (1, 2), (1, 3), (1, 4), (2, 20), (2, 30), (2, 40) ) AS t(id, value) GROUP BY id");
+    }
+
+    /// TODO: Check why Presto C++ doesn't throw an error for certain queries.
+    @Override
+    @Test
+    public void testCorrelatedNonAggregationScalarSubqueries()
+    {
+        String subqueryReturnedTooManyRows = ".*Scalar sub-query has returned multiple rows.*";
+
+        assertQuery("SELECT (SELECT 1 WHERE a = 2) FROM (VALUES 1) t(a)", "SELECT null");
+        assertQuery("SELECT (SELECT 2 WHERE a = 1) FROM (VALUES 1) t(a)", "SELECT 2");
+        assertQueryFails(
+                "SELECT (SELECT 2 FROM (VALUES 3, 4) WHERE a = 1) FROM (VALUES 1) t(a)",
+                subqueryReturnedTooManyRows);
+
+        // multiple subquery output projections
+        // TODO: Check why Presto C++ doesn't throw an error for below queries.
+        /*assertQueryFails(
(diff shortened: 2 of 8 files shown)


steveburnett (dismissed):
LGTM! (docs)

Reviewed README.md, no concerns. Thanks for the doc!

mknegi (dismissed):

tdcmeehan:
Are these tests being run as part of the CI?

pramodsatya:
They are not being run as part of CI currently, but we plan to setup a pipeline to run these tests on a regular cadence subsequently. Will also add a CI job to run these tests in a follow-up PR.

presto-native-tests/src/test/java/com.facebook.presto.nativetests/TestAggregations.java
```diff
@@ -0,0 +1,43 @@
...
+package com.facebook.presto.nativetests;
+
+import com.facebook.presto.nativeworker.NativeQueryRunnerUtils;
+import com.facebook.presto.nativeworker.PrestoNativeQueryRunnerUtils;
+import com.facebook.presto.testing.QueryRunner;
+
+public class TestAggregations
+        extends AbstractTestAggregationsNative
+{
+    private static final String storageFormat = "PARQUET";
```
jaystarshot, line 23: This is just testing parquet

pramodsatya: Yes the intention was to primarily test with parquet files. Most of the current e2e tests in `presto-native-execution` test with DWRF format. We can extend this to test with other file formats going forward.

pramodsatya:
> Just a suggestions for the release note: Consider using the description of the PR in the release note entry to help a release note reader understand what this module does.
> 
> ```
> == RELEASE NOTES ==
> General Changes
> * Add presto-native-tests module to run end-to-end tests with Presto native workers. :pr:`23671`
> ```

Thanks for the suggestion @steveburnett, updated accordingly.

zuyu:
Fyi, build error on your branch, after resolving conflicts w/ the latest `master` branch.

```
...
[INFO] presto-native-tests 0.290-SNAPSHOT ................. FAILURE [  7.832 s]
...
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  19.526 s (Wall Clock)
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.0.0:check (checkstyle) on project presto-native-tests: Failed during checkstyle execution: Unable to find configuration file at location: src/checkstyle/presto-checks.xml: Could not find resource 'src/checkstyle/presto-checks.xml'. -> [Help 1]
[ERROR]
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR]
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
[ERROR]
[ERROR] After correcting the problems, you can resume the build with the command
[ERROR]   mvn <args> -rf :presto-native-tests
```

presto-native-tests/src/test/java/com.facebook.presto.nativetests/TestAggregations.java
pramodsatya: > This is just testing parquet

Tests with `DWRF` are added here: https://github.com/prestodb/presto/pull/24809.

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java
aditi-pandit, line 1: @pramodsatya : Can you move the changes in this file to a separate PR ? Then it would leave the rest of the code very specific to native engines.

aditi-pandit (commented):
@pramodsatya @soumiiow : Can we move the custom functions test to a separate PR ? Don't think it is depending on any other changes in this PR is it ?

pramodsatya:
> @pramodsatya @soumiiow : Can we move the custom functions test to a separate PR ? Don't think it is depending on any other changes in this PR is it ?

@aditi-pandit, the concern in moving this to a separate PR (and test class) is that there are 41 other test classes in different modules that run the testcases from `AbstractTestQueries`, which include the e2e tests for Presto custom functions. So we did not want to break down `AbstractTestQueries`. Would it be fine to keep it in this PR?

aditi-pandit:
> > @pramodsatya @soumiiow : Can we move the custom functions test to a separate PR ? Don't think it is depending on any other changes in this PR is it ?
> 
> @aditi-pandit, the concern in moving this to a separate PR (and test class) is that there are 41 other test classes in different modules that run the testcases from `AbstractTestQueries`, which include the e2e tests for Presto custom functions. So we did not want to break down `AbstractTestQueries`. Would it be fine to keep it in this PR?

@pramodsatya : I see. We could do another approach then : Write a new test just for custom udf functions... that would be more useful for engineers wanting to understand custom udf anyways. And then have a second PR for `AbstractTestQueries`. wdyt ?

pramodsatya:
> > > @pramodsatya @soumiiow : Can we move the custom functions test to a separate PR ? Don't think it is depending on any other changes in this PR is it ?
> > 
> > 
> > @aditi-pandit, the concern in moving this to a separate PR (and test class) is that there are 41 other test classes in different modules that run the testcases from `AbstractTestQueries`, which include the e2e tests for Presto custom functions. So we did not want to break down `AbstractTestQueries`. Would it be fine to keep it in this PR?
> 
> @pramodsatya : I see. We could do another approach then : Write a new test just for custom udf functions... that would be more useful for engineers wanting to understand custom udf anyways. And then have a second PR for `AbstractTestQueries`. wdyt ?

Thanks for the suggestion @aditi-pandit, we went with this approach initially by moving the custom udf testcases to a separate test class, `AbstractTestCustomFunctions` (prototyped [here](https://github.com/pramodsatya/presto/commits/dyfunc_native_tests/)). However, we were concerned moving these testcases to `AbstractTestCustomFunctions` would result in them not being run by the other test suites that derive from `AbstractTestQueries`. We could duplicate the custom udf tests and copy them from `AbstractTestQueries` to `AbstractTestCustomFunctions` instead of moving them, not sure if that would be preferable as it would result in code duplication. Could you please share how to proceed?

pramodsatya:
@aditi-pandit, I missed that `AbstractTestCustomFunctions` was a part of the custom function tests added here and this wasn't the intention. I have reverted this now and moved custom function tests back to `AbstractTestQueries`. Could you please take another look at the custom function tests commit?

aditi-pandit:
@pramodsatya : I would be fine to create a separate test class 'AbstractTestCustomFunctions' with repetition for presto-native-tests only.

> Thanks for the suggestion @aditi-pandit, we went with this approach initially by moving the custom udf testcases to a separate test class, `AbstractTestCustomFunctions` (prototyped [here](https://github.com/pramodsatya/presto/commits/dyfunc_native_tests/)). However, we were concerned moving these testcases to `AbstractTestCustomFunctions` would result in them not being run by the other test suites that derive from `AbstractTestQueries`. We could duplicate the custom udf tests and copy them from `AbstractTestQueries` to `AbstractTestCustomFunctions` instead of moving them, not sure if that would be preferable as it would result in code duplication. Could you please share how to proceed?

pramodsatya:
> @pramodsatya : I would be fine to create a separate test class 'AbstractTestCustomFunctions' with repetition for presto-native-tests only.
> 
> > Thanks for the suggestion @aditi-pandit, we went with this approach initially by moving the custom udf testcases to a separate test class, `AbstractTestCustomFunctions` (prototyped [here](https://github.com/pramodsatya/presto/commits/dyfunc_native_tests/)). However, we were concerned moving these testcases to `AbstractTestCustomFunctions` would result in them not being run by the other test suites that derive from `AbstractTestQueries`. We could duplicate the custom udf tests and copy them from `AbstractTestQueries` to `AbstractTestCustomFunctions` instead of moving them, not sure if that would be preferable as it would result in code duplication. Could you please share how to proceed?

Thanks @aditi-pandit, opened a separate PR for custom function native tests: https://github.com/prestodb/presto/pull/25480. Could you please take a look?

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java
pramodsatya: Thanks for the suggestion @aditi-pandit, moved these changes to a separate PR: https://github.com/prestodb/presto/pull/25486. Could you please help review this change?

pramodsatya:
@aditi-pandit, could you please help review this PR?

presto-native-tests/pom.xml
aditi-pandit, line 1: @pramodsatya : It would be good to combine the code in this commit with the one where the build started breaking. Committing this separately would cause a temporary Presto build breakage.

presto-native-tests/src/test/java/com/facebook/presto/nativetests/AbstractTestQueriesNative.java
```diff
@@ -0,0 +1,1174 @@
...
+    /// Custom window functions are unsupported in Presto C++.
+    @Override
+    @Test(enabled = false)
+    public void testCustomRank() {}
+
+    /// Queries in this testcase use the apply function, which is used to test lambda expressions, and is currently
+    /// unsupported in Presto C++. See issue: https://github.com/prestodb/presto/issues/20741.
+    @Override
+    @Test(enabled = false)
+    public void testApplyLambdaRepeated() {}
```
aditi-pandit: Rather than disable this test, we should change it to look for relevant error messages instead.

presto-native-tests/src/test/java/com/facebook/presto/nativetests/AbstractTestQueriesNative.java
```diff
@@ -0,0 +1,1174 @@
...
+
+    /// Queries in this testcase use the apply function, which is used to test lambda expressions, and is currently
+    /// unsupported in Presto C++. See issue: https://github.com/prestodb/presto/issues/20741.
+    @Override
+    @Test(enabled = false)
+    public void testLambdaCapture() {}
```
aditi-pandit: Same as above. Can we check for error messages instead ?

presto-native-tests/src/test/java/com/facebook/presto/nativetests/AbstractTestQueriesNative.java
```diff
@@ -0,0 +1,1174 @@
...
+
+    /// Queries in this testcase use the apply function, which is used to test lambda expressions, and is currently
+    /// unsupported in Presto C++. See issue: https://github.com/prestodb/presto/issues/20741.
+    @Override
+    @Test(enabled = false)
+    public void testLambdaInAggregationContext() {}
```
aditi-pandit: same as above... check for error messages instead.

presto-native-tests/src/test/java/com/facebook/presto/nativetests/AbstractTestQueriesNative.java
```diff
@@ -0,0 +1,1174 @@
...
+
+    /// Queries in this testcase use the apply function, which is used to test lambda expressions, and is currently
+    /// unsupported in Presto C++. See issue: https://github.com/prestodb/presto/issues/20741.
+    @Override
+    @Test(enabled = false)
+    public void testLambdaInSubqueryContext() {}
```
aditi-pandit: Check for error messages instead.

presto-native-tests/src/test/java/com/facebook/presto/nativetests/AbstractTestQueriesNative.java
```diff
@@ -0,0 +1,1174 @@
...
+
+    /// Queries in this testcase use the apply function, which is used to test lambda expressions, and is currently
+    /// unsupported in Presto C++. See issue: https://github.com/prestodb/presto/issues/20741.
+    @Override
+    @Test(enabled = false)
+    public void testNonDeterministicInLambda() {}
```
aditi-pandit: Check for error messages instead.

presto-native-tests/src/test/java/com/facebook/presto/nativetests/AbstractTestQueriesNative.java
```diff
@@ -0,0 +1,1174 @@
...
+
+    /// Queries in this testcase use the apply function, which is used to test lambda expressions, and is currently
+    /// unsupported in Presto C++. See issue: https://github.com/prestodb/presto/issues/20741.
+    @Override
+    @Test(enabled = false)
+    public void testRowSubscriptInLambda() {}
```
aditi-pandit: Check for error messages instead.

presto-native-tests/src/test/java/com/facebook/presto/nativetests/AbstractTestQueriesNative.java
```diff
@@ -0,0 +1,1174 @@
...
+
+    /// Queries in this testcase use the apply function, which is used to test lambda expressions, and is currently
+    /// unsupported in Presto C++. See issue: https://github.com/prestodb/presto/issues/20741.
+    @Override
+    @Test(enabled = false)
+    public void testTryWithLambda() {}
```
aditi-pandit: Check for error messages instead.

presto-native-tests/src/test/java/com/facebook/presto/nativetests/AbstractTestQueriesNative.java
```diff
@@ -0,0 +1,1174 @@
...
+    /// This test is not applicable in Presto C++ as there is no bytecode IR as with JVM.
+    @Override
+    @Test(enabled = false)
+    public void testLargeBytecode() {}
+
+    /// The functions returned by SHOW FUNCTIONS and their order differ in Presto and Presto C++. Exact comparison is
+    /// not needed and this test is disabled in Presto C++.
+    @Override
+    @Test(enabled = false)
+    public void testShowFunctions() {}
```
aditi-pandit: This is an interesting one because we can compute which functions are not supported either side with SHOW FUNCTIONS. But I don't really have a concrete idea on what we can do with it.

presto-native-tests/src/test/java/com/facebook/presto/nativetests/AbstractTestQueriesNative.java
```diff
@@ -0,0 +1,1174 @@
...
+
+    /// This test uses char type which is not supported in Presto C++. See issue:
+    /// https://github.com/prestodb/presto/issues/21332.
+    @Override
+    @Test(enabled = false)
+    public void testLikePrefixAndSuffixWithChars() {}
```
aditi-pandit: Check for error messages here.

presto-native-tests/src/test/java/com/facebook/presto/nativetests/AbstractTestQueriesNative.java
```diff
@@ -0,0 +1,1174 @@
...
+
+    /// array_cum_sum function in Presto C++ does not support decimal array inputs, see issue:
+    /// https://github.com/prestodb/presto/issues/24402.
+    @Override
+    @Test(enabled = false)
+    public void testArrayCumSumDecimals() {}
```
aditi-pandit: I have seen a PR from Joe for this. It will be good to enable this test and assert the error messages so that he will have to explicitly make changes when his PR makes it to Velox.

presto-native-tests/src/test/java/com/facebook/presto/nativetests/AbstractTestQueriesNative.java
```diff
@@ -0,0 +1,1174 @@
...
+    /// See issue for more details: https://github.com/facebookincubator/velox/issues/10338.
+    @Override
+    @Test(enabled = false)
+    public void testPreserveAssignmentsInJoin() {}
+
+    /// TODO: This test is pending on Varchar(n) support in Presto C++.
+    /// Planner error: type of variable 'expr' is expected to be varchar(5), but the actual type is varchar.
+    @Override
+    @Test(enabled = false)
+    public void testCaseToMapOptimization() {}
```
aditi-pandit: Check error messages in test instead.

presto-docs/src/main/sphinx/presto_cpp/limitations.rst
aditi-pandit, line 1: You could move the changes in limitations.rst and TestNonIterativeDistributedQueries and TestTpchDistributedQueries into a separate PR as that will be much easier to review and submit than this PR.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/NativeQueryRunnerUtils.java
```diff
@@ -282,7 +282,7 @@ public static void createCustomer(Session session, QueryRunner queryRunner)
     {
         if (!queryRunner.tableExists(session, "customer")) {
             queryRunner.execute(session, "CREATE TABLE customer AS " +
-                    "SELECT custkey, name, address, nationkey, phone, acctbal, comment, mktsegment " +
+                    "SELECT custkey, name, address, nationkey, phone, acctbal, mktsegment, comment " +
```
aditi-pandit: I don't follow why this fixed something. Please can you elaborate.

presto-native-tests/src/test/java/com/facebook/presto/nativetests/NativeTestsUtils.java
```diff
@@ -47,10 +47,10 @@ public static void createTables(String storageFormat)
                     .setAddStorageFormatToPath(true)
                     .build();
             if (storageFormat.equals("DWRF")) {
-                NativeQueryRunnerUtils.createAllTables(javaQueryRunner, true);
+                NativeQueryRunnerUtils.createAllTablesStandard(javaQueryRunner, true);
```
aditi-pandit: Don't understand why you changed this to standard.. Seems non-intuitive as the calling method is called createTables.

presto-main-base/src/main/java/com/facebook/presto/sql/planner/PlannerUtils.java
```diff
@@ -480,7 +482,16 @@ public static boolean isSupportedArrayContainsFilter(FunctionResolution function
             if (functionResolution.isArrayContainsFunction(callExpression.getFunctionHandle())) {
                 RowExpression array = callExpression.getArguments().get(0);
                 RowExpression element = callExpression.getArguments().get(1);
-                checkState(array.getType() instanceof ArrayType && ((ArrayType) array.getType()).getElementType().equals(element.getType()));
+                // varchar(n) is not supported in Presto C++, compare base types of array and element for varchar to
```
aditi-pandit: Can you move this commit outside of this PR and write a separate individual test for it.  You can start with the original test you fixed, and then we might want to add more variations to it.

presto-main-base/src/main/java/com/facebook/presto/metadata/FunctionSignatureMatcher.java
```diff
@@ -71,7 +71,7 @@ public Optional<Signature> match(Collection<? extends SqlFunction> candidates, L
                 .filter(function -> !function.getSignature().getTypeVariableConstraints().isEmpty())
                 .collect(Collectors.toList());
 
-        match = matchFunctionExact(genericCandidates, parameterTypes);
+        match = matchFunctionGeneric(genericCandidates, parameterTypes);
```
aditi-pandit, line 74: Separate this into a PR also for just this condition and write an individual test for it.

aditi-pandit (commented):
@pramodsatya : Thanks for the iterations of this PR. Have the next round of comments.

presto-native-tests/src/test/java/com/facebook/presto/nativetests/AbstractTestQueriesNative.java
```diff
@@ -0,0 +1,1174 @@
...
+
+    /// TODO: Fix result mismatch between Presto and Presto C++.
+    /// TODO: Check difference in behavior with sidecar enabled and disabled.
+    /// no sidecar
+    /// VeloxUserError: numInput == 1 (6000 vs. 1) Expected single row of input. Received 6000 rows.
+    /// sidecar
+    /// VeloxUserError: numInput == 1 (7500 vs. 1) Expected single row of input. Received 7500 rows.
+    @Override
+    @Test(enabled = false)
+    public void testScalarSubquery() {}
```
aditi-pandit: @pramodsatya : Create an issue for this test.

presto-native-tests/src/test/java/com/facebook/presto/nativetests/AbstractTestQueriesNative.java
```diff
@@ -0,0 +1,1174 @@
...
+    @Test(enabled = false)
+    public void testCoercions() {}
+
+    /// TODO: Fix result mismatch between Presto and Presto C++.
+    /// java.lang.AssertionError: Iterators differ at element [0]: [1, 1, O] != [5, 5, F]
+    /// Expected :[1, 1, O]
+    /// Actual   :[5, 5, F]
+    @Override
+    @Test(enabled = false)
+    public void testTopNUnpartitionedWindow() {}
```
aditi-pandit: Create an issue for this.

presto-native-tests/src/test/java/com/facebook/presto/nativetests/AbstractTestQueriesNative.java
```diff
@@ -0,0 +1,1174 @@
...
+    /*
+    not equal
+    Actual rows (1 of 1 extra rows shown, 1 rows in total):
+        [1500]
+    Expected rows (1 of 1 missing rows shown, 1 rows in total):
+        [500]
+    */
+    @Override
+    @Test
+    public void testCorrelatedExistsSubqueriesWithEqualityPredicatesInWhere()
```
aditi-pandit: @pramodsatya : Do we understand why this happens ? This shouldn't fail in Presto C++ without side-car. Please create an issue for it. 

Same for the next 2 as well.

presto-main-base/src/main/java/com/facebook/presto/metadata/FunctionSignatureMatcher.java
pramodsatya: Moved to https://github.com/prestodb/presto/pull/25640, thanks!

steveburnett:
Suggested changes for the release note. 

```
== RELEASE NOTES ==

* Add testcases from ``AbstractTestQueries`` to ``presto-native-tests``.
```

presto-native-tests/src/test/java/com/facebook/presto/nativetests/AbstractTestQueriesNative.java
pramodsatya: This test checks metadata for certain functions and would need to be modified for native functions, I figured since validation with Presto java is not necessary for output of SHOW FUNCTIONS this test could be disabled. I have now modified the test to validate native function signatures instead of disabling the test, would this be fine?

presto-native-tests/src/test/java/com/facebook/presto/nativetests/AbstractTestQueriesNative.java
pramodsatya: The PR from Joe landed and this testcase is passing now, reverted this change.

presto-native-tests/src/test/java/com/facebook/presto/nativetests/AbstractTestQueriesNative.java
pramodsatya: This issue is not seen in the latest run, reverted the overriden test.

presto-docs/src/main/sphinx/presto_cpp/limitations.rst
pramodsatya: The changes in TestNonIterativeDistributedQueries and TestTpchDistributedQueries depend on AbstractTestQueriesNative, and the comment in limitations.rst explains the difference in `approx_set` results for which the test is added here. Is it fine to retain these changes here?

presto-main-base/src/main/java/com/facebook/presto/sql/planner/PlannerUtils.java
pramodsatya: This is no longer needed with the fix in https://github.com/prestodb/presto/pull/26003, reverted this change and other overriden Varchar(N) tests.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/NativeQueryRunnerUtils.java
pramodsatya: Thanks for pointing it out, the `select *` in testcase was resulting in the column order being swapped for these two columns, modifying the query in testcase instead fixed this error.

presto-native-tests/src/test/java/com/facebook/presto/nativetests/AbstractTestQueriesNative.java
pramodsatya: Upon further investigation the issue was just mismatch of error message in Presto and Velox, modified the native testcase to check for the right error message.

presto-native-tests/src/test/java/com/facebook/presto/nativetests/AbstractTestQueriesNative.java
pramodsatya: Upon checking further, the results were same in Presto and PrestoC++ but the order of rows was different. Modifying the query in base testcase to enforce ordering by `rank` ensures a deterministic order of rows and fixes this error, thanks.

presto-native-tests/src/test/java/com/facebook/presto/nativetests/NativeTestsUtils.java
pramodsatya: Thanks for pointing it out, reverted this change and added an arg instead to `createTables` for creating lineitem table as per Tpch spec (as in `createLineitemStandard`) without the additional columns.

pramodsatya (commented):
Thanks for the feedback @aditi.pandit, addressed the comments. Could you PTAL?

pramodsatya:
@aditi-pandit, could you please help review this PR? I addressed the previous review comments and opened issues for failures identified.

presto-native-tests/src/test/java/com/facebook/presto/nativetests/TestTpchDistributedQueries.java
```diff
@@ -0,0 +1,118 @@
...
+    {
+        //  Generate a super-long query: SELECT x,x,x,x,x,... FROM (VALUES 1,2,3,4,5) t(x)
+        @Language("SQL") String longQuery = "SELECT x" + Strings.repeat(",x", 500_000) + " FROM (VALUES 1,2,3,4,5) t(x)";
+        assertQueryFails(longQuery, "Query text length \\(1000037\\) exceeds the maximum length \\(1000000\\)");
+    }
+
+    @Test
+    public void testAnalyzePropertiesSystemTable()
+    {
+        assertQuery("SELECT COUNT(*) FROM system.metadata.analyze_properties WHERE catalog_name = 'tpch'", "SELECT 0");
```
aditi-pandit, line 82: @pramodsatya : Why does this fail on native ?

pramodsatya: @aditi-pandit, this does not fail on native, it is added here to maintain consistency with the `TestTpchDistributedQueries` from `presto-tests`:
https://github.com/prestodb/presto/blob/9ebd1f2a97256b9f0e5e77c5722bd39045034778/presto-tests/src/test/java/com/facebook/presto/tests/TestTpchDistributedQueries.java#L43

Ideally the tests here should be moved to a separate abstract class, I can take this up in a subsequent PR?

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeHiveExternalTableTpchQueries.java
```diff
@@ -109,7 +109,7 @@ protected void createTables()
         createSchemaIfNotExist(javaQueryRunner, TPCH_SCHEMA);
         Session session = Session.builder(super.getSession()).setCatalog(HIVE).setSchema(TPCH_SCHEMA).build();
         createOrders(session, javaQueryRunner, true);
-        createLineitemStandard(session, javaQueryRunner);
+        createLineitemStandard(session, javaQueryRunner, true);
```
aditi-pandit: For Parquet, we support DATE type. Do we really need this conversion ?

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeIcebergTpchQueries.java
```diff
@@ -32,7 +32,7 @@ public abstract class AbstractTestNativeIcebergTpchQueries
     protected void createTables()
     {
         QueryRunner queryRunner = (QueryRunner) getExpectedQueryRunner();
-        createLineitemStandard(queryRunner);
+        createLineitemStandard(queryRunner, true);
```
aditi-pandit: The casting should depend on storageFormat as some types like DWRF don't support DATE, but Parquet does. Can you check ?

aditi-pandit (commented):
@pramodsatya : Another high level comment: Whether we cast Date depends on the storage format as DWRF doesn't support DATE but Parquet does. Ideally we should use that to guide the casting and avoid the extra parameter as much as possible. Lets cleanup guided by that.

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java
```diff
@@ -7563,9 +7563,9 @@ public void testRemoveRedundantCastToVarcharInJoinClause()
                 .setSystemProperty(REMOVE_REDUNDANT_CAST_TO_VARCHAR_IN_JOIN, "true")
                 .build();
         // Trigger optimization
-        assertQuery(session, "select * from orders o join customer c on cast(o.custkey as varchar) = cast(c.custkey as varchar)");
+        assertQuery(session, "select o.custkey, c.name from orders o join customer c on cast(o.custkey as varchar) = cast(c.custkey as varchar)");
         assertQuery(session, "select o.orderkey, c.name from orders o join customer c on cast(o.custkey as varchar) = cast(c.custkey as varchar)");
-        assertQuery(session, "select *, cast(o.custkey as varchar), cast(c.custkey as varchar) from orders o join customer c on cast(o.custkey as varchar) = cast(c.custkey as varchar)");
+        assertQuery(session, "select c.name, cast(o.custkey as varchar), cast(c.custkey as varchar) from orders o join customer c on cast(o.custkey as varchar) = cast(c.custkey as varchar)");
```
aditi-pandit, line 7948: Don't follow why these columns are added. Why did * not work ?

presto-native-tests/src/test/java/com/facebook/presto/nativetests/TestTpchDistributedQueries.java
aditi-pandit: Yeah... you can move the tests to a separate abstract class. Lets avoid rewriting the tests.

presto-native-tests/src/test/java/com/facebook/presto/nativetests/AbstractTestQueriesNative.java
```diff
@@ -0,0 +1,1158 @@
...
+    public void testCoercions() {}
+
+    /// With sidecar disabled, the expression is constant-folded on the coordinator.
+    /// TODO: Depends on now function in Velox: https://github.com/facebookincubator/velox/pull/14139.
+    @Override
+    @Test
+    public void testValuesWithTimestamp()
+    {
+        if (sidecarEnabled) {
+            @Language("RegExp") String nowFunctionUnsupportedError = ".*Function native.default.now not registered.*";
```
aditi-pandit: Think now() is supported in Velox recently.

presto-native-tests/src/test/java/com/facebook/presto/nativetests/AbstractTestQueriesNative.java
```diff
@@ -0,0 +1,1158 @@
...
+    /// Color functions are not supported in Presto C++.
+    @Override
+    @Test
+    public void testFunctionArgumentTypeConstraint()
+    {
+        @Language("RegExp") String errorMessage = sidecarEnabled ? ".*Function native.default.rgb not registered.*" : ".*Unexpected parameters \\(color\\) for function greatest.*";
+        assertQueryFails("SELECT greatest(rgb(255, 0, 0))", errorMessage);
+    }
+
+    /// The output JSON formatted string is different in Presto C++.
```
aditi-pandit, line 872: @amitkdutta : Can you review these tests for json_format usage ?

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeHiveExternalTableTpchQueries.java
pramodsatya: Reverted and using storage format now.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeIcebergTpchQueries.java
pramodsatya: Overloaded the function to take in storageFormat, thanks.

presto-native-tests/src/test/java/com/facebook/presto/nativetests/TestTpchDistributedQueries.java
pramodsatya: I tried moving these tests out of `TestTpchDistributedQueries` and into an abstract class, but because we are overriding several tests from the base class `AbstractTestQueries` for native testing, this refactor does not help avoid rewriting the tests.
Could I add a TODO comment for this?

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java
pramodsatya: The market segment and comment columns in customer table are swapped when select * is used (this happens without the config being tested here as well).

pramodsatya (commented):
Thanks @aditi-pandit, I have reverted the `castDateToVarchar` parameter to use `storageFormat` instead, and separated this change to https://github.com/prestodb/presto/pull/26955.

Could you please take another look?

steveburnett:
Consider rebasing the PR to rerun the tests.

pramodsatya:
@aditi-pandit, @czentgr, could you please help review these tests?

presto-native-tests/src/test/java/com/facebook/presto/nativetests/AbstractTestQueriesNative.java
```diff
@@ -0,0 +1,1172 @@
...
+            assertEquals(row.getField(0), row.getField(1));
+        }
+        else {
+            super.testValuesWithTimestamp();
+        }
+    }
+
+    @Override
+    @Test
+    public void testMinMaxByToWindowFunction()
```
aditi-pandit, line 984: Why are these rewrites needed ?

presto-native-tests/src/test/java/com/facebook/presto/nativetests/AbstractTestQueriesNative.java
```diff
@@ -0,0 +1,1172 @@
...
+                        "SELECT ARRAY[null, CAST(row(1, null) AS ROW(INTEGER, INTEGER))] " +
+                        "UNION ALL " +
+                        "SELECT ARRAY[CAST(row(null, 2) AS ROW(INTEGER, INTEGER))])) " +
+                        "CROSS JOIN unnest(agg_result) as r(c1, c2)",
+                unnestRowsInvalidFieldError, true);
+    }
+
+    /// The error message for invalid map_agg function differs in Presto and Velox.
+    @Override
+    @Test(enabled = false)
```
aditi-pandit: Why is this test disabled ?

presto-native-tests/src/test/java/com/facebook/presto/nativetests/AbstractTestQueriesNative.java
```diff
@@ -0,0 +1,1172 @@
...
+        assertQueryFails(sql, arrayCumSumFunctionSignatureUnsupportedError, true);
+
+        sql = "select array_cum_sum(k) from (values (array[cast(null as varchar), '6', '0'])) t(k)";
+        assertQueryFails(sql, arrayCumSumFunctionSignatureUnsupportedError, true);
+    }
+
+    /// The error message for the case where subquery returns multiple rows differs in Presto and Velox.
+    @Override
+    @Test
+    public void testScalarSubquery()
```
aditi-pandit, line 719: Is there a way to avoid this test repetition ? It might be okay to refactor testScalarSubquery with a common function that takes the error strings as a parameter.

presto-native-tests/src/test/java/com/facebook/presto/nativetests/AbstractTestQueriesNative.java
```diff
@@ -0,0 +1,1172 @@
...
+        }
+        else {
+            super.testShowSession();
+        }
+    }
+
+    /// The integer overflow error message differs in Presto and Velox.
+    @Override
+    @Test
+    public void testRemoveMapCastFailure()
```
aditi-pandit, line 609: It would be better to abstract a common function that takes as input a parameter for the error string and then change this function implementation to call the common function rather than repeating all the SQL.

presto-native-tests/src/test/java/com/facebook/presto/nativetests/AbstractTestQueriesNative.java
```diff
@@ -0,0 +1,1172 @@
...
+    @Override
+    @Test
+    public void testTryWithLambda()
+    {
+        assertQueryFails("SELECT TRY(apply(5, x -> x + 1) / 0)", applyNotRegisteredError);
+        assertQueryFails("SELECT TRY(apply(5 + RANDOM(1), x -> x + 1) / 0)", applyNotRegisteredError);
+        assertQueryFails("SELECT apply(5 + RANDOM(1), x -> x + TRY(1 / 0))", applyNotRegisteredError);
+    }
+
+    /// The functions are fetched from sidecar when it is enabled so the output of SHOW FUNCTIONS and the order of
```
aditi-pandit, line 453: Is there a way to change the test to continue to validate the results without the ordering ?

presto-native-tests/src/test/java/com/facebook/presto/nativetests/AbstractTestQueriesNative.java
```diff
@@ -0,0 +1,1172 @@
...
+    public void testLambdaInAggregation()
+    {
+        assertQuery("SELECT id, reduce_agg(value, 0, (a, b) -> a + b+0, (a, b) -> a + b) FROM ( VALUES (1, 2), (1, 3), (1, 4), (2, 20), (2, 30), (2, 40) ) AS t(id, value) GROUP BY id", "values (1, 9), (2, 90)");
+        assertQuery("SELECT id, 's' || reduce_agg(value, '', (a, b) -> concat(a, b, 's'), (a, b) -> concat(a, b, 's')) FROM ( VALUES (1, '2'), (1, '3'), (1, '4'), (2, '20'), (2, '30'), (2, '40') ) AS t(id, value) GROUP BY id",
+                "values (1, 's2s3ss4ss'), (2, 's20s30ss40ss')");
+        // assertQuery("SELECT id, reduce_agg(value, array[id, value], (a, b) -> a || b, (a, b) -> a || b) FROM ( VALUES (1, 2), (1, 3), (1, 4), (2, 20), (2, 30), (2, 40) ) AS t(id, value) GROUP BY id");
+    }
+
+    /// TODO: Check why Presto C++ doesn't throw an error for certain queries.
+    @Override
```
aditi-pandit, line 191: Lets get into the details of this. Please create issues if needed.

presto-native-tests/src/test/java/com/facebook/presto/nativetests/AbstractTestQueriesNative.java
```diff
@@ -0,0 +1,1172 @@
...
+            assertEqualsIgnoreOrder(resultWithOptimization, resultWithoutOptimization);
+            // aggregation over join
+            sql = "select custkey, sum(quantity), sum(quantity) filter (where tax < 0.05) from lineitem l join orders o on l.orderkey=o.orderkey group by custkey";
+            resultWithOptimization = computeActual(enableOptimization, sql);
+            resultWithoutOptimization = computeActual(disableOptimization, sql);
+            assertEqualsIgnoreOrder(resultWithOptimization, resultWithoutOptimization);
+        }
+    }
+
+    /// TODO: The last query in this testcase is flaky in Presto C++.
```
aditi-pandit, line 178: Can we investigate the flakiness ? Please create an issue for this.

aditi-pandit (commented):
Thanks @pramodsatya. Have bunch of comments.

Issue #23660, opened 2024-09-16
Query with scalar subquery having mathematical function returns 0 rows


pratyakshsharma:
In case when a query involves scalar subquery with mathematical function and the subquery returns no rows, the entire query returns no rows. This only happens in cases with mathematical functions present in the subquery. 
 

## Your Environment

* Presto version used: 0.290-SNAPSHOT (latest master)
* Storage (HDFS/S3/GCS..):
* Data source and connector used: hive connector
* Deployment (Cloud or On-prem): local machine
* [Pastebin](https://pastebin.com/) link to the complete debug logs:

## Expected Behavior
The scalar subquery should return null and other selected columns should return proper values.

## Current Behavior
When the scalar subquery returns no rows, the entire query returns 0 rows.

## Possible Solution
The query plan in such cases is introducing an empty Values Node in case `simplify_plan_with_empty_input` is set to true (It is true by default). If it is false, PickTableLayoutForPredicate rule removes TableScan node for the columns not in the subquery. Also an additional FilterNode with false predicate gets introduced. 

## Steps to Reproduce
Run the below simple query - 
```
select
  comment,
  (
    select
      name
    from
      tpch.tiny.nation
    where
      nationkey = 1
      and regionkey = 1
      and width_bucket(nationkey, 1, 100, 10) = 2
  )
from
  tpch.tiny.nation;
```

## Screenshots (if appropriate)
Outputs
<img width="1598" alt="Screenshot 2024-09-16 at 10 15 32 PM" src="https://github.com/user-attachments/assets/0d17bdf3-9c6f-43af-92cb-f7b073fe8e9b">

When `simplify_plan_with_empty_input` is true
<img width="1672" alt="Screenshot 2024-09-16 at 10 10 17 PM" src="https://github.com/user-attachments/assets/47db3db2-df43-44e0-8797-c27aed58068c">

When `simplify_plan_with_empty_input` is false
<img width="1710" alt="Screenshot 2024-09-16 at 10 10 02 PM" src="https://github.com/user-attachments/assets/705b3c60-bf56-4a1e-a52a-74b4ce7d58ca">

PR #23674, opened 2024-09-18
Fix correctness bug in EffectivePredicateExtractor

aaneja:
## Description
Fixes https://github.com/prestodb/presto/issues/23660

During predicate pushdown through a Join, we were 
- Pulling up a predicate expression that only refers to constant values
- Incorrectly pushing down this predicate between join sides
- This predicate expression gets optimized to `FALSE`, but now is pushed down to both sides of the join

This results in the optimizer Incorrectly assuming no results will be produced

## Fix
Pulling up expressions happens in`EffectivePredicateExtractor#pullExpressionThroughVariables`. This was modified to stop pulling up expressions that only refer to constants since this is logically incorrect

## Test Plan
- New unit and E2E tests added

## Contributor checklist

- [X] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/sql/planner/EffectivePredicateExtractor.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/planner/EffectivePredicateExtractor.java
@@ -60,6 +60,7 @@ import java.util.function.Predicate;
 
 import static com.facebook.presto.common.function.OperatorType.EQUAL;
 import static com.facebook.presto.common.type.BooleanType.BOOLEAN;
+import static com.facebook.presto.expressions.LogicalRowExpressions.FALSE_CONSTANT;
 import static com.facebook.presto.expressions.LogicalRowExpressions.TRUE_CONSTANT;
 import static com.facebook.presto.expressions.LogicalRowExpressions.extractConjuncts;
 import static com.facebook.presto.spi.relation.SpecialFormExpression.Form.IS_NULL;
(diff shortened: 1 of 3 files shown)


elharo (approved):
Nice discovery. Thanks!

jaystarshot:
Should we add a test from https://github.com/prestodb/presto/issues/23660?

presto-main/src/main/java/com/facebook/presto/sql/planner/EffectivePredicateExtractor.java
```diff
@@ -418,15 +419,21 @@ private RowExpression pullExpressionThroughVariables(RowExpression expression, C
             for (RowExpression conjunct : new EqualityInference.Builder(functionManger).nonInferableConjuncts(expression)) {
                 if (determinismEvaluator.isDeterministic(conjunct)) {
                     RowExpression rewritten = equalityInference.rewriteExpression(conjunct, in(variables));
-                    if (rewritten != null) {
+                    if (rewritten != null && (hasVariableReferences(rewritten) || rewritten.equals(FALSE_CONSTANT))) {
                         effectiveConjuncts.add(rewritten);
                     }
+                    // If equality inference has reduced the predicate to an expression referring to only constants, it does not make sense to pull this predicate up
```
jaystarshot, line 425: Bit confused here, as to why do we have the condition to skip if rewritten has variable references? What if the variable reference is pointing to a constant in some project below?

aaneja: If `rewritten` has variable references that are `in(variables)` we do *not* skip them and pull them up (by adding them to `effectiveConjuncts`. The problem here is that `EqaulityInference` is designed to rewrite expressions to just constants (see [test](https://github.com/prestodb/presto/blob/185876fab2f8aa2ff5caf2a9da6284a9c227c9c4/presto-main/src/test/java/com/facebook/presto/sql/planner/TestEqualityInference.java#L302-L303)), even though, technically this expression does not refer to any `variables`

However, if the `rewritten` expression has no variable references, le.g `mod(10,5)=1` - this predicate should not be pulled up, to say, a Join node, because it has no references in the outputs of the Join and cannot impact any predicate inferencing there

jaystarshot (approved):

PR #23722, opened 2024-09-25
Fix REST session bug for Iceberg REST catalogs

kiersten-stokes:
## Description
By relying on the Iceberg `loadCatalog` utility function for `RESTCatalog`s, there was no way to specify options for a `RESTSessionCatalog`. The result of this is that Iceberg REST user sessions were not being honored. This PR changes the flow for `RESTCatalog`s only, manually creating a `RESTCatalog` instance with the correct `RESTSessionCatalog` context.

## Motivation and Context
Fixes a known bug with REST sessions not working as expected (no related issue).

## Test Plan
Manual testing was done. Working on adding a CI test, which requires changes to the test REST server. I plan on further investigating options for this.

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/rest/IcebergRestCatalogFactory.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/rest/IcebergRestCatalogFactory.java
@@ -20,21 +20,35 @@ import com.facebook.presto.iceberg.IcebergCatalogName;
 import com.facebook.presto.iceberg.IcebergConfig;
 import com.facebook.presto.iceberg.IcebergNativeCatalogFactory;
 import com.facebook.presto.spi.ConnectorSession;
+import com.facebook.presto.spi.PrestoException;
+import com.facebook.presto.spi.security.ConnectorIdentity;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.util.concurrent.UncheckedExecutionException;
 import io.jsonwebtoken.Jwts;
-import org.apache.iceberg.rest.auth.OAuth2Properties;
+import org.apache.iceberg.CatalogProperties;
+import org.apache.iceberg.catalog.Catalog;
+import org.apache.iceberg.catalog.SessionCatalog.SessionContext;
+import org.apache.iceberg.rest.HTTPClient;
+import org.apache.iceberg.rest.RESTCatalog;
 
 import javax.inject.Inject;
 
 import java.util.Date;
 import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.ExecutionException;
 
 import static com.facebook.presto.iceberg.rest.AuthenticationType.OAUTH2;
 import static com.facebook.presto.iceberg.rest.SessionType.USER;
+import static com.google.common.base.Throwables.throwIfInstanceOf;
+import static com.google.common.base.Throwables.throwIfUnchecked;
 import static java.lang.String.format;
 import static java.util.Objects.requireNonNull;
+import static java.util.UUID.randomUUID;
 import static org.apache.iceberg.CatalogProperties.URI;
+import static org.apache.iceberg.CatalogUtil.configureHadoopConf;
 import static org.apache.iceberg.rest.auth.OAuth2Properties.CREDENTIAL;
+import static org.apache.iceberg.rest.auth.OAuth2Properties.JWT_TOKEN_TYPE;
 import static org.apache.iceberg.rest.auth.OAuth2Properties.OAUTH2_SERVER_URI;
 import static org.apache.iceberg.rest.auth.OAuth2Properties.TOKEN;
 
@@ -43,6 +57,7 @@ public class IcebergRestCatalogFactory
 {
     private final IcebergRestConfig catalogConfig;
     private final NodeVersion nodeVersion;
+    private final String catalogName;
 
     @Inject
     public IcebergRestCatalogFactory(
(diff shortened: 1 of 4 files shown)


presto-iceberg/src/main/java/com/facebook/presto/iceberg/rest/IcebergRestCatalogFactory.java
```diff
@@ -55,6 +69,28 @@ public IcebergRestCatalogFactory(
...
+    }
+
+    @Override
+    public Catalog getCatalog(ConnectorSession session)
+    {
+        try {
+            return catalogCache.get(getCacheKey(session), () -> {
+                RESTCatalog catalog = new RESTCatalog(
+                        convertSession(session),
+                        config -> HTTPClient.builder(config).uri(config.get(CatalogProperties.URI)).build());
```
kiersten-stokes, line 82: Need to change the import for `CatalogProperties.URI`

agrawalreetika (commented):
Overall looks good to me. Can we add some tests around before and after making changes in session properties for REST catalog?

presto-iceberg/src/test/java/org/apache/iceberg/rest/IcebergRestCatalogServlet.java
```diff
@@ -94,6 +96,16 @@ protected void execute(ServletRequestContext context, HttpServletResponse respon
         response.setStatus(HttpServletResponse.SC_OK);
         responseHeaders.forEach(response::setHeader);
 
+        String jwt = context.headers.get("Authorization");
+        if (jwt != null) {
+            if (!isAuthorizedUserToken(jwt)) {
```
hantangwangd, line 101:
```suggestion
        if (jwt != null && !isAuthorizedUserToken(jwt)) {
```

hantangwangd (commented):
Overall looks good to me, one little nit.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/rest/IcebergRestCatalogFactory.java
```diff
@@ -77,22 +126,99 @@ protected Map<String, String> getCatalogProperties(ConnectorSession session)
...
+        });
 
+        return properties.build();
+    }
+
+    protected SessionContext convertSession(ConnectorSession session)
+    {
+        RestSessionBuilder sessionContextBuilder = builder(session);
+        catalogConfig.getSessionType().ifPresent(type -> {
+            if (type.equals(USER)) {
```
ZacBlanco, line 140: to stick with the functional style of the `Optional` type, I would recommend instead to use `.filter(type -> type.equals(USER))`, then you can remove the conditional here in the `ifPresent` block

presto-iceberg/src/main/java/com/facebook/presto/iceberg/rest/IcebergRestCatalogFactory.java
```diff
@@ -77,22 +126,99 @@ protected Map<String, String> getCatalogProperties(ConnectorSession session)
...
+    {
+        private String sessionId;
+        private String identity;
+        private Map<String, String> properties;
+        private Map<String, String> credentials;
+        private final ConnectorIdentity wrappedIdentity;
+
+        private RestSessionBuilder(ConnectorSession session)
+        {
+            sessionId = randomUUID().toString();
```
ZacBlanco, line 178: Wondering if we should default to using the query ID instead here. Though I do notice we never actually leave the default in place. The only caller of `setSessionId` always uses the pattern `$user-$source`. Do you think using query ID might be easier to trace issues in the future? Or was there a specific reason to use the `$user-$source` pattern?

ZacBlanco (commented):
Minor nits, otherwise lgtm

presto-iceberg/src/main/java/com/facebook/presto/iceberg/rest/IcebergRestCatalogFactory.java
kiersten-stokes: Ah thank you! I completely forgot about `filter` (clearly)

presto-iceberg/src/main/java/com/facebook/presto/iceberg/rest/IcebergRestCatalogFactory.java
kiersten-stokes: This raises a few good points. As far as the default session ID, it does remain in place when the REST session type is `NONE` (which is the default config value and likely the more common situation at the moment). That being said, perhaps I should be more explicit about setting it in an `orElse` block in the `convertSession` function for clarity.

I'm not tied to using a random UUID for the session ID here, but I'm also not sure query ID would have the desired affect either. Essentially, we want REST user sessions to persist across queries (again, assuming `SessionType` is set to `USER`). This is where the session Id comes in - it's purpose is for server-side state caching. I figured the combo of 'user' and 'source' would be the most useful for persisting state. That being said, I'm not very well-versed in server-side handling of REST sessions in general so I'd be happy to change this if you think there's a better way. Perhaps paring it down to just user would work.

All this to say that session Id is somewhat useless if user sessions aren't active, hence the use of a random UUID. I don't think it would help much for troubleshooting either way because session Id will really only be set when the catalog is created (or, more generally, on cache miss).

Let me know your thoughts - will push a fix for your other review comment shortly as well.

agrawalreetika (approved):

hantangwangd (approved):

presto-iceberg/src/main/java/com/facebook/presto/iceberg/rest/IcebergRestCatalogFactory.java
ZacBlanco: I think we can leave it as a random UUID. I was curious about the reasoning behind it. I also didn't realize this wouldn't be set for non-`USER` session types. I think it's fine to leave as is. Thanks for the explanation!

ZacBlanco (approved):

PR #5835, opened 2016-08-10
fix for non power of two task_writer_count conflicting with hash partitioning

sopel39:
POC fix for: https://github.com/prestodb/presto/issues/5834

PR #23744, opened 2024-09-28
Allow task_partitioned_writer_count to be non-power of two for native engine

xiaoxmeng:
task_partitioned_writer_count is used to configure the number of drivers at worker size
which doesn't have to be power of two. In Prestissimo we need to configure this to avoid
the number of table writer workers is a multiple of the number of table writer threads per
worker. As Presto java engine still requires table writer threads to be power of two so
keep the check for non-native engine scenario.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-docs/src/main/sphinx/admin/properties.rst
+++ b/presto-docs/src/main/sphinx/admin/properties.rst
@@ -53,6 +53,25 @@ output data set is not skewed in order to avoid the overhead of hashing and
 redistributing all the data across the network. This can also be specified
 on a per-query basis using the ``redistribute_writes`` session property.
 
+``task_writer_count``
+^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``integer``
+* **Default value:** ``1``
+
+Default number of local parallel table writer threads per worker. It is required
+to be a power of two for a Java query engine.
+
+``task_partitioned_writer_count``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``integer``
+* **Default value:** ``task_writer_count``
+
+Number of local parallel table writer threads per worker for partitioned writes. If not
+set, the number set by ``task_writer_count`` will be used. It is required to be a power
+of two for a Java query engine.
+
 .. _tuning-memory:
 
 Memory Management Properties
--- a/presto-main/src/main/java/com/facebook/presto/execution/TaskManagerConfig.java
+++ b/presto-main/src/main/java/com/facebook/presto/execution/TaskManagerConfig.java
@@ -428,14 +428,14 @@ public class TaskManagerConfig
     }
 
     @Min(1)
-    @PowerOfTwo
     public int getWriterCount()
     {
         return writerCount;
     }
 
+    // NOTE: writer count needs to be a power of two for java query engine.
     @Config("task.writer-count")
-    @ConfigDescription("Number of writers per task")
+    @ConfigDescription("Number of writer threads per task")
     public TaskManagerConfig setWriterCount(int writerCount)
     {
         this.writerCount = writerCount;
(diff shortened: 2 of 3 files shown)


kewang1024 (dismissed):
Have no context of why task_partitioned_writer_count has to be power of two in the first place, but we do need to avoid having number of table writer workers (eg 300) be the multiple of task_partitioned_writer_count (for example 4) which would cause skew

rschlussel:
Context for why power of two is required is here: https://github.com/prestodb/presto/pull/5835.  I think we should only make this change for native execution because right now, apparently Java cannot handle non-power of 2 writer counts.  Also, you should remove the annotation TaskManagerConfig.getPartitionedWriterCount(). And might make sense to do a similar change for TaskManagerConfig.getWriterCount and the corresponding session property.

steveburnett:
I didn't find documentation for this property, consider adding doc for it.

xiaoxmeng:
> I didn't find documentation for this property, consider adding doc for it.

Document added.

presto-docs/src/main/sphinx/admin/properties.rst
```diff
@@ -53,6 +53,25 @@ output data set is not skewed in order to avoid the overhead of hashing and
...
 on a per-query basis using the ``redistribute_writes`` session property.
 
+``task_writer_count``
+^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``integer``
+* **Default value:** ``1``
+
+Default number of local parallel table writer threads per worker. It is required
+to be power of two for java query engine.
```
steveburnett, line 63:
```suggestion
to be a power of two for a Java query engine.
```

presto-docs/src/main/sphinx/admin/properties.rst
```diff
@@ -53,6 +53,25 @@ output data set is not skewed in order to avoid the overhead of hashing and
...
+
+``task_partitioned_writer_count``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``integer``
+* **Default value:** ``task_writer_count``
+
+Number of local parallel table writer threads per worker for partitioned writes. If not
+set, the number set by ``task_writer_count`` will be used. It is required to be power
```
steveburnett, line 72:
```suggestion
set, the number set by ``task_writer_count`` will be used. It is required to be a power
```

presto-docs/src/main/sphinx/admin/properties.rst
```diff
@@ -53,6 +53,25 @@ output data set is not skewed in order to avoid the overhead of hashing and
...
+of two for java query engine.
```
steveburnett, line 73:
```suggestion
of two for a Java query engine.
```

steveburnett (changes requested):
Thanks for the doc! A few minor nits, looks good!

xiaoxmeng:
> Context for why power of two is required is here: #5835. I think we should only make this change for native execution because right now, apparently Java cannot handle non-power of 2 writer counts. Also, you should remove the annotation TaskManagerConfig.getPartitionedWriterCount(). And might make sense to do a similar change for TaskManagerConfig.getWriterCount and the corresponding session property.

@rschlussel thanks for suggestions. Updated. Please help take a second look.

xiaoxmeng (commented):
@steveburnett document comments addressed. Thank you!

steveburnett (dismissed):
LGTM! (docs) 

Thanks for the quick update! Pull updated branch, new local doc build, looks good.

presto-main/src/main/java/com/facebook/presto/execution/TaskManagerConfig.java
```diff
@@ -428,29 +428,27 @@ public TaskManagerConfig setInfoMaxAge(Duration infoMaxAge)
...
     @Min(1)
-    @PowerOfTwo
     public int getWriterCount()
     {
         return writerCount;
     }
 
     @Config("task.writer-count")
-    @ConfigDescription("Number of writers per task")
+    @ConfigDescription("Number of writer threads per worker")
```
rschlussel, line 437: i would leave this description as per-task, as this is the TaskManagerConfig, and the control is in the task. (it just is also true that there is also one task per worker).

presto-main/src/main/java/com/facebook/presto/execution/TaskManagerConfig.java
```diff
@@ -428,29 +428,27 @@ public TaskManagerConfig setInfoMaxAge(Duration infoMaxAge)
     }
 
     @Min(1)
-    @PowerOfTwo
```
rschlussel, line 431: This change fully removes it from the power of two check from the config checking. I don't see an easy way to gate any kind of check based on whether native execution is disabled, but this should maybe have at least a code comment explaining the constraint.

presto-main/src/main/java/com/facebook/presto/execution/TaskManagerConfig.java
xiaoxmeng: Updated.

xiaoxmeng (commented):
@rschlussel comments addressed. Thanks!

rschlussel (approved):

PR #23149, opened 2024-07-09
[native] Fix driver count reporting from worker

tanjialiang:
We were reporting split counts to driver count. This PR fixes it.
```
== NO RELEASE NOTE ==
```

PR #23755, opened 2024-10-01
[native] Push down FilterNode into TableScan

Yuhta:
Sometimes the planner put a `FilterNode` right after `TableScanNode`. This change merges these 2 nodes by moving the filter expression into remaining filter of the table scan when possible. This results in fewer IO and CPU because table scan can leverage the information in remaining filter to skip file stripes in case of random sampling.

Also add `$path` and `$bucket` in split info columns and fix split counts in coordinator UI.

### DIFF
--- a/presto-native-execution/presto_cpp/main/PrestoTask.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoTask.cpp
@@ -590,10 +590,20 @@ void PrestoTask::updateExecutionInfoLocked(
   prestoTaskStats.outputPositions = 0;
   prestoTaskStats.outputDataSizeInBytes = 0;
 
-  prestoTaskStats.queuedDrivers = veloxTaskStats.numQueuedDrivers;
-  prestoTaskStats.totalDrivers = veloxTaskStats.numTotalDrivers;
+  // Presto Java reports number of drivers to number of splits in Presto UI
+  // because split and driver are 1 to 1 mapping relationship. This is not true
+  // in Prestissimo where 1 driver handles many splits. In order to quickly
+  // unblock developers from viewing the correct progress of splits in
+  // Prestissimo's coordinator UI, we put number of splits in total, queued, and
+  // finished to indicate the progress of the query. Number of running drivers
+  // are passed as it is to have a proper running drivers count in UI.
+  //
+  // TODO: We should really extend the API (protocol::TaskStats and Presto
+  // coordinator UI) to have splits information as a proper fix.
+  prestoTaskStats.totalDrivers = veloxTaskStats.numTotalSplits;
+  prestoTaskStats.queuedDrivers = veloxTaskStats.numQueuedSplits;
   prestoTaskStats.runningDrivers = veloxTaskStats.numRunningDrivers;
-  prestoTaskStats.completedDrivers = veloxTaskStats.numCompletedDrivers;
+  prestoTaskStats.completedDrivers = veloxTaskStats.numFinishedSplits;
 
   prestoTaskStats.pipelines.resize(veloxTaskStats.pipelineStats.size());
   for (int i = 0; i < veloxTaskStats.pipelineStats.size(); ++i) {
--- a/presto-native-execution/presto_cpp/main/types/PrestoToVeloxConnector.cpp
+++ b/presto-native-execution/presto_cpp/main/types/PrestoToVeloxConnector.cpp
@@ -1109,13 +1109,15 @@ HivePrestoToVeloxConnector::toVeloxSplit(
   for (const auto& [key, value] : hiveSplit->storage.serdeParameters) {
     serdeParameters[key] = value;
   }
-  std::unordered_map<std::string, std::string> infoColumns;
-  infoColumns.reserve(2);
-  infoColumns.insert(
-      {"$file_size", std::to_string(hiveSplit->fileSplit.fileSize)});
-  infoColumns.insert(
+  std::unordered_map<std::string, std::string> infoColumns = {
+      {"$path", hiveSplit->fileSplit.path},
+      {"$file_size", std::to_string(hiveSplit->fileSplit.fileSize)},
       {"$file_modified_time",
-       std::to_string(hiveSplit->fileSplit.fileModifiedTime)});
+       std::to_string(hiveSplit->fileSplit.fileModifiedTime)},
+  };
+  if (hiveSplit->tableBucketNumber) {
+    infoColumns["$bucket"] = std::to_string(*hiveSplit->tableBucketNumber);
+  }
   auto veloxSplit =
       std::make_unique<velox::connector::hive::HiveConnectorSplit>(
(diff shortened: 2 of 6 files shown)


presto-native-execution/presto_cpp/main/PrestoTask.cpp
```diff
@@ -590,10 +590,20 @@ void PrestoTask::updateExecutionInfoLocked(
...
-  prestoTaskStats.totalDrivers = veloxTaskStats.numTotalDrivers;
+  // Presto Java reports number of drivers to number of splits in Presto UI
+  // because split and driver are 1 to 1 mapping relationship. This is not true
+  // in Prestissimo where 1 driver handles many splits. In order to quickly
+  // unblock developers from viewing the correct progress of splits in
+  // Prestissimo's coordinator UI, we put number of splits in total, queued, and
+  // finished to indicate the progress of the query. Number of running drivers
+  // are passed as it is to have a proper running drivers count in UI.
+  //
+  // TODO: We should really extend the API (protocol::TaskStats and Presto
```
aditi-pandit, line 601: This is quite hacky and will be confusing for Prestissimo engineers who don't come from Presto Java.

We should ideally extend protocol::TaskStats and extend the API. How expensive is this ? At the very least can we create a Jira issue ? I'll get it triaged to someone in our team if you'll don't have the cycles.

@tdcmeehan @amitkdutta

presto-native-execution/presto_cpp/main/types/PrestoToVeloxConnector.cpp
```diff
@@ -1109,13 +1109,15 @@ HivePrestoToVeloxConnector::toVeloxSplit(
   for (const auto& [key, value] : hiveSplit->storage.serdeParameters) {
     serdeParameters[key] = value;
   }
-  std::unordered_map<std::string, std::string> infoColumns;
-  infoColumns.reserve(2);
-  infoColumns.insert(
-      {"$file_size", std::to_string(hiveSplit->fileSplit.fileSize)});
-  infoColumns.insert(
+  std::unordered_map<std::string, std::string> infoColumns = {
```
aditi-pandit, line 1112: We could follow this up with some code cleanup in Velox which uses constant strings for path and bucket synthesized columns. Is that the intention going forward ?

The info columns are used by Gluten as well. Might be worth checking with them.

aditi-pandit (commented):
Thanks @Yuhta. Had few questions:

presto-native-execution/presto_cpp/main/PrestoTask.cpp
Yuhta: This was the situation from the start of Prestissimo and we made a regression during a fix (#23149).  Just put it back to what it was.

presto-native-execution/presto_cpp/main/types/PrestoToVeloxConnector.cpp
Yuhta: Yes that is why I add these.  Gluten can add any info columns they want in their split creation.

Yuhta:
@aditi-pandit Local planner has no dependency on Hive so it does not have knowledge about remaining filter.  I guess this was the same reason why Presto planner does not put it in remaining filter as well (did not verify).  Seems Prestissimo is a reasonable place to fix this, as we have all the dependency needed and we have the knowledge that this will benefit the execution layer (Velox).

presto-native-execution/presto_cpp/main/PrestoTask.cpp
aditi-pandit: Ah I see. Do you want to create a github issue for the protocol::TaskStats cleanup ? There are engineers on our team who want to fix that issue.

presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp
```diff
@@ -607,6 +607,27 @@ core::PlanNodePtr VeloxQueryPlanConverterBase::toVeloxQueryPlan(
...
+  // For ScanFilter and ScanFilterProject, the planner sometimes put the
+  // remaining filter in a FilterNode after the TableScan.  We need to put it
+  // back to TableScan so that Velox can leverage it to do stripe level
+  // skipping.  Otherwise we only get row level skipping and lose some
+  // optimization opportunity in case of very low selectivity.
+  if (auto tableScan = std::dynamic_pointer_cast<const protocol::TableScanNode>(
+          node->source)) {
+    if (auto* tableLayout = dynamic_cast<protocol::HiveTableLayoutHandle*>(
+            tableScan->table.connectorTableLayout.get())) {
+      auto remainingFilter =
```
aditi-pandit, line 619: This code looks a bit strange in that tableLayout->remainingPredicate is a boolean expression evaluating to true, and its reset to the filter predicate instead. So you are suggesting the Presto planner first moves this around and then we want to move it back ?

aditi-pandit:
> @aditi-pandit Local planner has no dependency on Hive so it does not have knowledge about remaining filter. I guess this was the same reason why Presto planner does not put it in remaining filter as well (did not verify). Seems Prestissimo is a reasonable place to fix this, as we have all the dependency needed and we have the knowledge that this will benefit the execution layer (Velox).

Hmm... Yeah, Velox Local Planner doesn't have dependency on Hive. Alright, your approach is reasonable.

presto-native-execution/presto_cpp/main/PrestoTask.cpp
Yuhta: Sure #23758

presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp
Yuhta: Yes we only reset it when it is a constant true (no-op).  This is also the pattern I observed in query.

presto-native-execution/presto_cpp/main/PrestoTask.cpp
aditi-pandit: Thanks

presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp
Yuhta: Yes it should be empty values node if the condition is constant false

aditi-pandit (approved):
Thanks @Yuhta for this code, and answering all my questions.

jaystarshot:
Please consider adding release notes following our release notes guide - [link](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines).

Using below for now

`*Merged FilterNode into TableScanNode where possible, reducing I/O and CPU; added $path and $bucket to split info, and fixed split counts in coordinator UI. :pr:`23755``

PR #16891, opened 2021-10-19
Rename array_dupes function to array_duplicates

mbasmanova:
Also, rename array_has_dupes function to array_has_duplicates.

The new names are consistent with the rest of the Presto functions.

Kept the old names as aliases, but removed these from the documentation
to discourage usage.

```
== RELEASE NOTES ==

General Changes
* Rename array_dupes and array_has_dupes functions to :func:`array_duplicates` and :func:`array_has_duplicates`.

PR #23762, opened 2024-10-02
Remove array_dupes and array_has_dupes alias.

amitkdutta:
## Description
Removing non-standard alias names array_dupes and array_has_dupes. [These were moved to alias and removed from documentation in 2021](https://github.com/prestodb/presto/pull/16891).

## Motivation and Context
Keeping function names standard.

## Impact
If there is any legacy usage of these aliases, those needs to be migrated to standard names. Standard name `array_duplicates` and `array_has_duplicates` are already available and no change is done there.

## Test Plan
Existing tests and removed test with alias.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Remove array_dupes and array_has_dupes alias names from functions :func:`array_duplicates` and :func:`array_has_duplicates`.

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/operator/scalar/sql/ArraySqlFunctions.java
+++ b/presto-main/src/main/java/com/facebook/presto/operator/scalar/sql/ArraySqlFunctions.java
@@ -69,7 +69,7 @@ public class ArraySqlFunctions
                 "m -> m)";
     }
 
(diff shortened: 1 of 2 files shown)


NikhilCollooru (approved):
If there are any pipelines using these aliases, they will fail. We should make a post in the Presto FYI group to inform the users

PR #23782, opened 2024-10-07
Add build param to skip creating hardlinks

singcha:
## Description
Presto build fails when a filesystem with no support for hardlinks is used. Add new param to specify hardlink include jars.

## Motivation and Context
Fixes build issue when underlying filesystem doesn't support hardlinks

## Test Plan
Built on devserver

== NO RELEASE NOTE ==


README.md
```diff
@@ -33,6 +33,10 @@ Presto has a comprehensive set of unit tests that can take several minutes to ru
 
     ./mvnw clean install -DskipTests
 
+Presto uses provisio plugin for packaging using hard links. If your filesystem doesn't support hardlinks, you can bypass hardlinks using:
```
steveburnett, line 36:
```suggestion
Presto uses the [Provisio](https://github.com/jvanzyl/provisio) plugin for packaging using hard links. If your filesystem doesn't support hard links, you can bypass hard links using:
```
Suggest adding a link to Provisio. Nit for consistency in "hard link".

steveburnett (changes requested):
Thanks for the doc!

steveburnett (approved):
LGTM! (docs)

README.md
```diff
@@ -33,6 +33,10 @@ Presto has a comprehensive set of unit tests that can take several minutes to ru
 
     ./mvnw clean install -DskipTests
 
+Presto uses the [Provisio](https://github.com/jvanzyl/provisio) plugin for packaging using hard links. If your filesystem doesn't support hard links, you can bypass hard links using:
+    
+    ./mvnw clean install -presto.hard-link-includes=''
+
```
tdcmeehan, line 39: My sense is that the vast majority of filesystems folks use do support hardlinks.  If that's the case, this is a very prominent callout in the README for relatively esoteric information.  I would remove it, and perhaps add a comment in the POM instead.

singcha: IMO - Anyone setting up project & building it, is going to look in README file for build instructions.

tdcmeehan: Maybe let's create a section toward the bottom for common issues with the build.  For now, we can just put this there, but add other stuff as needed (incorrect JDK, Maven timeouts, etc).

singcha: Updated

tdcmeehan (approved):

tdcmeehan:
@singcha for some reason this keeps failing the Maven checks.  Can you look into why?

PR #23794, opened 2024-10-09
Upgraded commons-io to version 2.16.0

infvg:
## Description
Upgraded commons-io to version 2.16.0
## Motivation and Context
This upgrade was created to deal with CVEs found in lower versions
## Impact
None

## Release Notes
```
== RELEASE NOTES ==

General Changes
* Upgraded commons-io to version 2.16.0 :pr:`23794`
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -304,6 +304,12 @@
             </dependency>
 
             <dependency>
+                <groupId>commons-io</groupId>
+                <artifactId>commons-io</artifactId>
+                <version>2.16.0</version>
+            </dependency>
+
+            <dependency>
                 <groupId>com.facebook.presto</groupId>
                 <artifactId>presto-parquet</artifactId>
                 <version>${project.version}</version>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -2357,6 +2363,17 @@
                     </configuration>
                 </plugin>
 
+                <plugin>
+                    <groupId>org.basepom.maven</groupId>
+                    <artifactId>duplicate-finder-maven-plugin</artifactId>
+                    <configuration>
+                        <ignoredClassPatterns combine.children="append">
```
agrawalreetika, line 2370: Why do we need `combine.children="append"` here?

infvg: Instead of adding this plugin to several child modules, we can add this in the root and this will cause it to be inherited.

agrawalreetika (approved):

ZacBlanco (approved):

tdcmeehan (approved):

Issue #23798, opened 2024-10-09
EXPLAIN ANALYZE fails on queries with CTE materialization


ZacBlanco:
When running a simple query that uses CTE materialization, `EXPLAIN ANALYZE` on the query will fail

## Your Environment

- HiveQueryRunner on MacOS

Session configuration
```
SET SESSION cte_partitioning_provider_catalog='hive';
SET SESSION hive.temporary_table_storage_format = 'PARQUET';
SET SESSION hive.temporary_table_schema = '__temp_ctes__';
```

## Expected Behavior
We should get a valid `EXPLAIN ANALYZE` output

## Steps to Reproduce

1.
2.
3.
4.

PR #23824, opened 2024-10-14
Modified EXPLAIN ANALYZE output

infvg:
## Description
The EXPLAIN ANALYZE operator only supports one substage when returning the output.
When outputting in TEXT format, modify it to loop through the substages and return
the substage ID and its plan.
For JSON format, return a list of plans.

## Motivation and Context
Multiple substages were not previously supported in the output
Resolves: #23798

## Impact
Modifies the EXPLAIN ANALYZE output

## Test Plan
Tested locally using the HiveQueryRunner

```
== RELEASE NOTES ==

General Changes
* Modified EXPLAIN ANALYZE output for TEXT and JSON format :pr:``

```
== NO RELEASE NOTE ==
Modified the EXPLAIN ANALYZE output for TEXT and JSON formats. Text now returns each stage ID followed by its plan. JSON returns a list of plans.
```

### CODE REVIEW

File: presto-main/src/main/java/com/facebook/presto/operator/ExplainAnalyzeOperator.java [hunk 1]
```diff
@@ -149,18 +150,28 @@ public Page getOutput()
...
+                for (StageInfo stageInfo : queryInfo.getOutputStage().get().getSubStages()) {
+                    planStringBuilder.append("Stage ID: ").append(stageInfo.getStageId().toString()).append('\n')
+                            .append(textDistributedPlan(stageInfo, functionAndTypeManager, operatorContext.getSession(),
+                                    verbose));
+                }
+                plan = planStringBuilder.toString();
                 break;
             case JSON:
-                plan = jsonDistributedPlan(queryInfo.getOutputStage().get().getSubStages().get(0), functionAndTypeManager, operatorContext.getSession());
+                StringJoiner planStringJoiner = new StringJoiner(",", "[", "]");
```
ZacBlanco, line 169: Since the output needs to be valid JSON, we should probably use a JSON codec for lists and render it through that instead. It likely will require some more refactoring of the `PlanPrinter` class to achieve that. Joining strings like this is prone to errors/generating invalid JSON

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestDistributedQueries.java
```diff
@@ -380,12 +380,18 @@ private static void assertJsonNodesHaveStats(JsonRenderer.JsonRenderedNode node)
...
+            fragments.values().forEach(planFragment -> assertJsonNodesHaveStats(planFragment.getPlan()));
+        }
+        fragmentsList = renderer.deserialize((String) computeActual("EXPLAIN ANALYZE (format JSON) SELECT rank() OVER (PARTITION BY orderkey ORDER BY clerk DESC) FROM orders").getOnlyValue());
+        for (Map<PlanFragmentId, JsonRenderer.JsonPlan> fragments : fragmentsList) {
+            fragments.values().forEach(planFragment -> assertJsonNodesHaveStats(planFragment.getPlan()));
+        }
+        fragmentsList = renderer.deserialize((String) computeActual("EXPLAIN ANALYZE (format JSON) SELECT rank() OVER (PARTITION BY orderkey ORDER BY clerk DESC) FROM orders WHERE orderkey < 0").getOnlyValue());
+        for (Map<PlanFragmentId, JsonRenderer.JsonPlan> fragments : fragmentsList) {
+            fragments.values().forEach(planFragment -> assertJsonNodesHaveStats(planFragment.getPlan()));
+        }
```
ZacBlanco, line 394: we should add another test where we enable CTEs and verify that all stages exist in the plan output

ZacBlanco (changes requested):

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestDistributedQueries.java
infvg: added

presto-hive/src/test/java/com/facebook/presto/hive/TestHiveDistributedQueries.java
```diff
@@ -116,5 +120,31 @@ public void testTrackMaterializedCTEs()
...
+    {
+        Session materializedSession = Session.builder(getSession())
+                .setSystemProperty(VERBOSE_OPTIMIZER_INFO_ENABLED, "true")
+                .setSystemProperty(CTE_MATERIALIZATION_STRATEGY, "ALL")
+                .setSystemProperty(CTE_PARTITIONING_PROVIDER_CATALOG, "hive")
+                .build();
+
+        JsonRenderer renderer = new JsonRenderer(getQueryRunner().getMetadata().getFunctionAndTypeManager());
+
+        List<Map<PlanFragmentId, JsonRenderer.JsonPlan>> fragmentsList = renderer.deserialize((String) computeActual(materializedSession, "EXPLAIN ANALYZE (format JSON) SELECT * FROM orders").getOnlyValue());
```
ZacBlanco, line 134: So I see you copied some queries from the other test case. We actually need a new query which executes with some CTE materialization.

Instead of all these, let's just have one query:

```sql
EXPLAIN ANALYZE with x as (SELECT * from orders WHERE totalprice < 1000) SELECT avg(totalprice) FROM x;
```

This should generate a query with 1 CTE which splits the execution into two stages and should fail without your changes

ZacBlanco (changes requested):
Thanks for the updates! A few more comments

presto-hive/src/test/java/com/facebook/presto/hive/TestHiveDistributedQueries.java
```diff
@@ -116,5 +120,31 @@ public void testTrackMaterializedCTEs()
         checkCTEInfo(explain, "tbl2", 1, false, true);
     }
 
+    @Test
+    public void testExplainAnalyzeMaterializedCTEs()
+    {
+        Session materializedSession = Session.builder(getSession())
+                .setSystemProperty(VERBOSE_OPTIMIZER_INFO_ENABLED, "true")
```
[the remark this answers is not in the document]
infvg: Resolved

presto-hive/src/test/java/com/facebook/presto/hive/TestHiveDistributedQueries.java
```diff
@@ -116,5 +120,31 @@ public void testTrackMaterializedCTEs()
...
+            fragments.values().forEach(planFragment -> assertJsonNodesHaveStats(planFragment.getPlan()));
+        });
+
+        fragmentsList = renderer.deserialize((String) computeActual(materializedSession, "EXPLAIN ANALYZE (format JSON) SELECT rank() OVER (PARTITION BY orderkey ORDER BY clerk DESC) FROM orders").getOnlyValue());
+        fragmentsList.forEach(fragments -> {
+            fragments.values().forEach(planFragment -> assertJsonNodesHaveStats(planFragment.getPlan()));
+        });
+
+        fragmentsList = renderer.deserialize((String) computeActual(materializedSession, "EXPLAIN ANALYZE (format JSON) SELECT rank() OVER (PARTITION BY orderkey ORDER BY clerk DESC) FROM orders WHERE orderkey < 0").getOnlyValue());
+        fragmentsList.forEach(fragments -> {
```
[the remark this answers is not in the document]
infvg: Fixed

presto-hive/src/test/java/com/facebook/presto/hive/TestHiveDistributedQueries.java
infvg: Changed the query

steveburnett:
Thanks for the release note entry! Nit suggestion to rephrase to follow the [Order of changes](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines#order-of-changes) in the Release Note Guidelines. 

```
== RELEASE NOTES ==

General Changes
* Improve the EXPLAIN ANALYZE output for TEXT and JSON formats, so that TEXT now returns each stage ID followed by its plan. JSON returns a list of plans. :pr:`23824`
```

ZacBlanco (approved):

aaneja:
I think the root cause here is not with a failure in rendering, but with how the `EXPLAIN ANALYZE` is executing the query. PFA, two QueryInfo JSON's -

- [plain_materializedcte_queryinfo.json](https://github.com/user-attachments/files/18319085/plain_materializedcte_queryinfo.json) : Is the execution of `WITH t as (VALUES 1, 2, 3) SELECT * FROM t`. On this I notice that `outputStage.subStages` is an array with only 1 element

- [explain_analyze_materialized_cte_query_info.json](https://github.com/user-attachments/files/18319083/explain_analyze_materialized_cte_query_info.json) : Is the execution of `EXPLAIN ANALYZE WITH t as (VALUES 1, 2, 3) SELECT * FROM t`. On this I notice that `outputStage.subStages` is an array with 2 elements. This is the reason we observe the [failed assertion](https://github.com/prestodb/presto/blob/f661a1ce171a58697030f09c870c013a252ff55a/presto-main/src/main/java/com/facebook/presto/operator/ExplainAnalyzeOperator.java#L152). The 'extra' sub-stage is a `TableScanNode` for the temp materialized CTE table. It's stageId differs from the 'regular' query run as well

To me, this look like a bug with how the plan got fragmented. I need to dive deeper to confirm this

PR #23832, opened 2024-10-15
[native] Add the native isBlockedTiming stat to operator stats

zuyu:
Exposes the `isBlockedTiming` stat from native worker to coordinator:
1. add the stat in `OperatorStats` and update presto protocol.
2. Presto task collects the stat from velox per https://github.com/facebookincubator/velox/pull/10923.

```
== NO RELEASE NOTE ==
```


presto-main/src/main/java/com/facebook/presto/operator/OperatorContext.java
```diff
@@ -536,6 +538,10 @@ public OperatorStats getOperatorStats()
 
                 1,
 
+                isBlockedTiming.getCalls(),
+                succinctNanos(isBlockedTiming.getWallNanos()),
+                succinctNanos(isBlockedTiming.getCpuNanos()),
```
yingsu00, line 543: getAllocationBytes()

yingsu00 (approved):

PR #23845, opened 2024-10-16
Fix ARRAYS_OVERLAP function bug

### LINKED ISSUE #23730: arrays_overlap return different results after swapping parameters for array types
```
SELECT
    ARRAYS_OVERLAP(ARRAY[ARRAY[1, 2], ARRAY[1, NULL]], ARRAY[ARRAY[1, 2]])
=> true
SELECT
    ARRAYS_OVERLAP(ARRAY[ARRAY[1, NULL], ARRAY[1, 2]], ARRAY[ARRAY[1, 2]])
=> throw
```

## Expected Behavior
Both queries should return the same result

## Current Behavior
They return different results

## Possible Solution
Switch completely to use "[set based implementation](https://github.com/prestodb/presto/blob/14da6b7754aba39b5aae3203c94b1e0dabedfba3/presto-main/src/main/java/com/facebook/presto/operator/scalar/ArraysOverlapFunction.java#L73)" for arrays_overlap

infvg:
## Description

In some arrays containing arrays that contain null values, arrays_overlap was not properly comparing values. Switch array_overlap to set based implementation.

## Motivation and Context

Fixes  #23730

## Impact
None

## Test Plan
Added the failing functions to unit tests and ran testArraysOverlap to confirm the functionality for the other queries remains the same.

Screenshot 2024-10-16 at 5 12 35 PM

== RELEASE NOTES ==
```
General Changes
* Fix a bug where a mirrored :func:`arrays_overlap(x, y) -> boolean` function does not return the correct value.
```

### DIFF
--- a/presto-main-base/src/main/java/com/facebook/presto/operator/scalar/ArraysOverlapFunction.java
+++ b/presto-main-base/src/main/java/com/facebook/presto/operator/scalar/ArraysOverlapFunction.java
@@ -18,10 +18,18 @@ import com.facebook.presto.common.type.StandardTypes;
 import com.facebook.presto.common.type.Type;
 import com.facebook.presto.operator.aggregation.TypedSet;
 import com.facebook.presto.spi.function.Description;
+import com.facebook.presto.spi.function.OperatorDependency;
 import com.facebook.presto.spi.function.ScalarFunction;
 import com.facebook.presto.spi.function.SqlNullable;
 import com.facebook.presto.spi.function.SqlType;
 import com.facebook.presto.spi.function.TypeParameter;
+
+import java.lang.invoke.MethodHandle;
+
+import static com.facebook.presto.common.function.OperatorType.IS_DISTINCT_FROM;
+import static com.facebook.presto.common.type.TypeUtils.readNativeValue;
+import static com.facebook.presto.util.Failures.internalError;
+import static com.google.common.base.Defaults.defaultValue;
 
 @ScalarFunction("arrays_overlap")
 @Description("Returns true if arrays have common elements")
@@ -36,6 +44,7 @@ public final class ArraysOverlapFunction
     @SqlNullable
     public static Boolean arraysOverlap(
             @TypeParameter("T") Type elementType,
+            @OperatorDependency(operator = IS_DISTINCT_FROM, argumentTypes = {"T", "T"}) MethodHandle elementIsDistinctFrom,
             @SqlType("array(T)") Block leftArray,
             @SqlType("array(T)") Block rightArray)
     {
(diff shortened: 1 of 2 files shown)


presto-main/src/test/java/com/facebook/presto/type/TestArrayOperators.java
```diff
@@ -1295,6 +1295,8 @@ public void testArraysOverlap()
         assertFunction("ARRAYS_OVERLAP(ARRAY [NULL, 3], ARRAY [2, 1])", BooleanType.BOOLEAN, null);
         assertFunction("ARRAYS_OVERLAP(ARRAY [3, NULL], ARRAY [2, 1])", BooleanType.BOOLEAN, null);
         assertFunction("ARRAYS_OVERLAP(ARRAY [3, NULL], ARRAY [2, 1, NULL])", BooleanType.BOOLEAN, null);
+        assertFunction("ARRAYS_OVERLAP(ARRAY[ARRAY[1, 2], ARRAY[1, NULL]], ARRAY[ARRAY[1, 2]])", BooleanType.BOOLEAN, true);
```
kewang1024, line 1298: Keep the same format as above?

kewang1024 (commented):
Thanks @infvg for the fix, this makes sense to me, @rschlussel was looking at this issue before, let's have her review this PR as well

presto-main/src/test/java/com/facebook/presto/type/TestArrayOperators.java
infvg: Fixed

presto-common/src/main/java/com/facebook/presto/common/type/ArrayType.java
```diff
@@ -86,10 +86,14 @@ public boolean equalTo(Block leftBlock, int leftPosition, Block rightBlock, int
...
-            checkElementNotNull(rightArray.isNull(i), ARRAY_NULL_ELEMENT_MSG);
-            if (!elementType.equalTo(leftArray, i, rightArray, i)) {
-                return false;
+            if (leftArray.isNull(i) == rightArray.isNull(i)) {
+                if (!elementType.equalTo(leftArray, i, rightArray, i)) {
+                    return false;
+                }
+            }
+            else {
+                checkElementNotNull(leftArray.isNull(i), ARRAY_NULL_ELEMENT_MSG);
```
rschlussel, line 95: this is not right. we intentionally throw for equals comparison with nulls for arrays and maps (and if we did change it, we would want to return false rather than throw).

presto-main/src/main/java/com/facebook/presto/operator/scalar/ArraysOverlapFunction.java
```diff
@@ -59,8 +60,13 @@ public static Boolean arraysOverlap(
                     hasNull = true;
                     continue;
                 }
-                if (elementType.equalTo(leftArray, i, rightArray, j)) {
-                    return true;
+                try {
+                    if (elementType.equalTo(leftArray, i, rightArray, j)) {
```
rschlussel, line 64: equalTo should be throwing on null input.  Instead we should pass the IS_DISTINCT_FROM method handle and call that instead of equal to (and additionally pass it to the TypedSet for the set based implementation). You can look at TypedSet.isContainedAt for how it calls the distinct from method handle.  Additionally, you can look at ArrayIntersectFunction for how the method handle gets passed in.

steveburnett:
@infvg , are you available to work on this?

presto-main/src/main/java/com/facebook/presto/operator/scalar/ArraysOverlapFunction.java
infvg: Added the implementation

rschlussel (approved):

yingsu00 (approved):

Issue #23849, opened 2024-10-17
Internal Error in Presto When Querying Elastic search Index with No Mapped Columns


K-Naga-Vivek:
When querying an Elasticsearch index that has no defined columns, Presto throws an internal error. Elasticsearch allows the creation of indices without any mapped columns, which causes Presto to fail when attempting to retrieve data from such indices. 

## Your Environment

* Presto version used: latest
* Data source and connector used: ElasticSearch
* Debug logs :

> Query 20241017_031656_00002_9vw2m failed: Internal error
> java.util.NoSuchElementException
>         at java.base/java.util.LinkedHashMap$LinkedHashIterator.nextNode(LinkedHashMap.java:721)
>         at java.base/java.util.LinkedHashMap$LinkedValueIterator.next(LinkedHashMap.java:746)
>         at com.facebook.presto.elasticsearch.client.ElasticsearchClient.lambda$getIndexMetadata$8(ElasticsearchClient.java:495)
>         at com.facebook.presto.elasticsearch.client.ElasticsearchClient.doRequest(ElasticsearchClient.java:726)
>         at com.facebook.presto.elasticsearch.client.ElasticsearchClient.getIndexMetadata(ElasticsearchClient.java:485)
>         at com.facebook.presto.elasticsearch.ElasticsearchMetadata.makeInternalTableMetadata(ElasticsearchMetadata.java:202)
>         at com.facebook.presto.elasticsearch.ElasticsearchMetadata.makeInternalTableMetadata(ElasticsearchMetadata.java:197)
>         at com.facebook.presto.elasticsearch.ElasticsearchMetadata.getColumnHandles(ElasticsearchMetadata.java:377)
>         at com.facebook.presto.metadata.MetadataManager.getColumnHandles(MetadataManager.java:507)
>         at com.facebook.presto.metadata.MetadataManager$1.getColumnHandles(MetadataManager.java:1383)
>         at com.facebook.presto.util.MetadataUtils.lambda$getTableColumnMetadata$3(MetadataUtils.java:97)
>         at com.facebook.presto.common.RuntimeStats.profileNanos(RuntimeStats.java:136)
>         at com.facebook.presto.util.MetadataUtils.getTableColumnMetadata(MetadataUtils.java:95)
>         at com.facebook.presto.util.MetadataUtils.getTableColumnsMetadata(MetadataUtils.java:54)
[...]

PR #23850, opened 2024-10-17
Handle exception for empty tables in elasticsearch

K-Naga-Vivek:
Add a catch block to handle the exception when a table in Elasticsearch does not have any columns. Previously, this situation would cause an internal error in Presto.

The code iterates through the mappings (columns) but lacks an exception handler for cases where an element is not found. So, catching the NoSuchElementException to throw the appropriate error.

## Motivation and Context
https://github.com/prestodb/presto/issues/23849

## Test Plan
Added the related testcase to the code change.

Tested and working as excepted.
image

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

Elasticsearch Changes
* Improve handling of exceptions for empty tables in Elasticsearch :pr:`23850`

### DIFF
--- a/presto-elasticsearch/src/main/java/com/facebook/presto/elasticsearch/client/ElasticsearchClient.java
+++ b/presto-elasticsearch/src/main/java/com/facebook/presto/elasticsearch/client/ElasticsearchClient.java
@@ -492,6 +492,10 @@ public class ElasticsearchClient
                     // Older versions of ElasticSearch supported multiple "type" mappings
                     // for a given index. Newer versions support only one and don't
                     // expose it in the document. Here we skip it if it's present.
+
+                    if (!mappings.elements().hasNext()) {
+                        return new IndexMetadata(new IndexMetadata.ObjectType(ImmutableList.of()));
+                    }
                     mappings = mappings.elements().next();
                 }
 
--- a/presto-elasticsearch/src/test/java/com/facebook/presto/elasticsearch/TestElasticsearchIntegrationSmokeTest.java
+++ b/presto-elasticsearch/src/test/java/com/facebook/presto/elasticsearch/TestElasticsearchIntegrationSmokeTest.java
@@ -805,4 +805,12 @@ public class TestElasticsearchIntegrationSmokeTest
         client.getLowLevelClient()
                 .performRequest("PUT", "/" + indexName, ImmutableMap.of(), new NStringEntity(mapping, ContentType.APPLICATION_JSON));
     }
+
+    @Test
+    public void testEmptyIndexNoMappings()
+            throws IOException
+    {
+        client.getLowLevelClient().performRequest("PUT", "/emptyindex");
+        assertQueryFails("SELECT * FROM emptyindex", "line 1:8: SELECT \\* not allowed from relation that has no columns");
+    }
 }

### CODE REVIEW

**Comment by steveburnett** [comment]
Thanks for the release note entry! Nit suggestion to help it follow the [Order of changes](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines#order-of-changes) phrasing in the Release Notes Guidelines. 

```
== RELEASE NOTES ==

Elasticsearch Changes
* Improve handling of exceptions for empty tables in Elasticsearch :pr:`23850`
```

File: presto-elasticsearch/src/main/java/com/facebook/presto/elasticsearch/client/ElasticsearchClient.java [hunk 1]
```diff
@@ -502,6 +503,9 @@ public IndexMetadata getIndexMetadata(String index)
             catch (IOException e) {
                 throw new PrestoException(ELASTICSEARCH_INVALID_RESPONSE, e);
             }
+            catch (NoSuchElementException e) {
+                throw new PrestoException(ELASTICSEARCH_QUERY_FAILURE, "Relation does not have any columns");
```
tdcmeehan, line 507:
```suggestion
                throw new PrestoException(ELASTICSEARCH_QUERY_FAILURE, "Relation does not have any columns", e);
```

presto-elasticsearch/src/main/java/com/facebook/presto/elasticsearch/client/ElasticsearchClient.java
```diff
@@ -502,6 +503,9 @@ public IndexMetadata getIndexMetadata(String index)
             catch (IOException e) {
                 throw new PrestoException(ELASTICSEARCH_INVALID_RESPONSE, e);
             }
+            catch (NoSuchElementException e) {
```
tdcmeehan, line 506: Can you explain where this is thrown?  I'm just wondering if it's because of the iteration on line 496.

tdcmeehan: So instead of catching an exception, can you instead check if `hasNext()` and throw?

K-Naga-Vivek:
@tdcmeehan, as suggested, I’ve updated the code to return an empty list instead of throwing an error when encountering tables without columns. This change resolves the failing test case and functions as expected. Now, if a `SELECT * ` query is run on a table with no columns, the **StatementAnalyzer** throws an error: `SELECT * not allowed from relation that has no columns` as intended.

Could you please review the changes?

tdcmeehan:
Why does this connector throw at all if there are no columns?  Can't it return a table with no columns?

K-Naga-Vivek:
The connector is currently set to return an empty list, but Presto has a check in the [StatementAnalyzer](https://github.com/prestodb/presto/blob/f02bf22c5b202cebde459c07024b295877f54ca1/presto-main/src/main/java/com/facebook/presto/sql/analyzer/StatementAnalyzer.java#L2729) that throws an error if a table has no columns.

tdcmeehan:
Can we let the exception occur in the analyzer, instead of throwing within the Elastic Search connector?

tdcmeehan (approved):

PR #23855, opened 2024-10-17
[native] Revert remaining filter pushdown from FilterNode

Yuhta:
If there is column renaming happening in-between, this will not work. Handling projection is too complex at worker level, we need to fix this in optimizer.

### DIFF
--- a/presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp
+++ b/presto-native-execution/presto_cpp/main/types/PrestoToVeloxQueryPlan.cpp
@@ -606,29 +606,6 @@ core::PlanNodePtr VeloxQueryPlanConverterBase::toVeloxQueryPlan(
             left->outputType()));
   }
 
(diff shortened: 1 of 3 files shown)


amitkdutta (approved):
@Yuhta Can this PR create correctness issues and is it already running in production environment?

Yuhta:
@amitkdutta In theory it could cause correctness issue, but in practice it is unlikely, because the optimizer generated name always has a underscore number suffix and it is unlikely we happen to have a similar name with same underscore number suffix with same type in the same table.

jaystarshot:
@Yuhta is this a revert of an existing PR? Just checking to determine if release notes are necessary for this PR."

Yuhta:
@jaystarshot Yes it's partial revert of #23755

jaystarshot:
Okay thanks! I am adding this in release notes, please let me know if its inaccurate

`* Reverted merging of FilterNode into TableScanNode done in pr `23755`.`

PR #23857, opened 2024-10-18
Introduce Plugin interface for REST function server

abevk2023:
Introduce function plugin interface to load existing scalar functions implemented on top of presto SPI plugin interface.

Depend on https://github.com/prestodb/presto/pull/23687

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-function-server/pom.xml
+++ b/presto-function-server/pom.xml
@@ -110,6 +110,26 @@
         </dependency>
 
         <dependency>
+            <groupId>com.google.code.findbugs</groupId>
+            <artifactId>jsr305</artifactId>
+        </dependency>
+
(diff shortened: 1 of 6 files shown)


presto-function-server/pom.xml
```diff
@@ -129,6 +149,30 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
```
tdcmeehan, line 152: Instead of using the Pinot plugin and bringing in those dependencies, can you just create a simple local function and register it?

tdcmeehan (changes requested):

tdcmeehan (approved):

PR #23929, opened 2024-10-31
[native] Use row wise encoding in partitioned remote exchanges for wide data sets

arhimondr:
## Description

Pick row wise exchange encoding when columnar encoding becomes inefficient.

This PR only contains planner changes. Execution path will be provided as a separate PR and will be provided for native execution only.

## Motivation and Context

Columnar partitioning can get inefficient when the number of output partitions and the number of columns is high.

For example an exchange of 1000 VARCHAR columns into 300 partitions requires 900,000 buffers to be allocated. Even when the allocations are as small as 128B per buffer the amount of memory allocated can be considerable (110MB) and can be far exceeding the PartitionedOutput operator budget (usually ~32MB for ~300 partitions).

CPU wise it can also be extremely inefficient. For the example above, for every record 300,000 buffers have to be updated.

## Impact

Improved efficiency for exchanges of more than 500 columns.

## Test Plan

Unit, Integration and e2e tests

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

* Improve partitioned remote exchanges for wide data sets (more than 500 columns) to use row wise encoding. :pr:`23929 `
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
+++ b/presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
@@ -332,6 +332,7 @@ public final class SystemSessionProperties
     private static final String NATIVE_EXECUTION_EXECUTABLE_PATH = "native_execution_executable_path";
     private static final String NATIVE_EXECUTION_PROGRAM_ARGUMENTS = "native_execution_program_arguments";
     public static final String NATIVE_EXECUTION_PROCESS_REUSE_ENABLED = "native_execution_process_reuse_enabled";
+    public static final String NATIVE_MIN_COLUMNAR_ENCODING_CHANNELS_TO_PREFER_ROW_WISE_ENCODING = "native_min_columnar_encoding_channels_to_prefer_row_wise_encoding";
 
     private final List<PropertyMetadata<?>> sessionProperties;
 
@@ -1824,6 +1825,11 @@ public final class SystemSessionProperties
                 booleanProperty(INLINE_PROJECTIONS_ON_VALUES,
                         "Whether to evaluate project node on values node",
                         featuresConfig.getInlineProjectionsOnValues(),
+                        false),
+                integerProperty(
+                        NATIVE_MIN_COLUMNAR_ENCODING_CHANNELS_TO_PREFER_ROW_WISE_ENCODING,
+                        "Minimum number of columnar encoding channels to consider row wise encoding for partitioned exchange. Native execution only",
+                        queryManagerConfig.getMinColumnarEncodingChannelsToPreferRowWiseEncoding(),
                         false));
     }
 
@@ -3099,5 +3105,10 @@ public final class SystemSessionProperties
     public static boolean isInlineProjectionsOnValues(Session session)
     {
         return session.getSystemProperty(INLINE_PROJECTIONS_ON_VALUES, Boolean.class);
+    }
+
+    public static int getMinColumnarEncodingChannelsToPreferRowWiseEncoding(Session session)
+    {
+        return session.getSystemProperty(NATIVE_MIN_COLUMNAR_ENCODING_CHANNELS_TO_PREFER_ROW_WISE_ENCODING, Integer.class);
     }
 }
--- a/presto-main/src/main/java/com/facebook/presto/sql/planner/PlanOptimizers.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/planner/PlanOptimizers.java
@@ -35,6 +35,7 @@ import com.facebook.presto.sql.planner.iterative.rule.CrossJoinWithArrayNotConta
 import com.facebook.presto.sql.planner.iterative.rule.CrossJoinWithOrFilterToInnerJoin;
 import com.facebook.presto.sql.planner.iterative.rule.DesugarLambdaExpression;
 import com.facebook.presto.sql.planner.iterative.rule.DetermineJoinDistributionType;
+import com.facebook.presto.sql.planner.iterative.rule.DetermineRemotePartitionedExchangeEncoding;
 import com.facebook.presto.sql.planner.iterative.rule.DetermineSemiJoinDistributionType;
 import com.facebook.presto.sql.planner.iterative.rule.EliminateCrossJoins;
 import com.facebook.presto.sql.planner.iterative.rule.EvaluateZeroLimit;
@@ -929,7 +930,14 @@ public class PlanOptimizers
 
         // Precomputed hashes - this assumes that partitioning will not change
         builder.add(new HashGenerationOptimizer(metadata.getFunctionAndTypeManager()));
-
+        builder.add(new IterativeOptimizer(
+                metadata,
+                ruleStats,
+                statsCalculator,
+                costCalculator,
+                ImmutableSet.of(new DetermineRemotePartitionedExchangeEncoding(
+                        featuresConfig.isNativeExecutionEnabled(),
+                        featuresConfig.isPrestoSparkExecutionEnvironment()))));
         builder.add(new MetadataDeleteOptimizer(metadata));
 
         // TODO: consider adding a formal final plan sanitization optimizer that prepares the plan for transmission/execution/logging
--- a/presto-main/src/main/java/com/facebook/presto/sql/planner/SystemPartitioningHandle.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/planner/SystemPartitioningHandle.java
@@ -101,13 +101,25 @@ public final class SystemPartitioningHandle
     @Override
     public boolean isSingleNode()
     {
-        return partitioning == SystemPartitioning.COORDINATOR_ONLY || partitioning == SystemPartitioning.SINGLE;
+        return function == SystemPartitionFunction.SINGLE;
     }
 
     @Override
     public boolean isCoordinatorOnly()
     {
         return partitioning == SystemPartitioning.COORDINATOR_ONLY;
+    }
+
+    @Override
+    public boolean isBroadcast()
+    {
+        return function == SystemPartitionFunction.BROADCAST;
+    }
+
+    @Override
+    public boolean isArbitrary()
+    {
+        return function == SystemPartitionFunction.ROUND_ROBIN;
     }
 
     @Override
--- a/presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/DetermineRemotePartitionedExchangeEncoding.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/DetermineRemotePartitionedExchangeEncoding.java
@@ -1,0 +1,137 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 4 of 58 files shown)


steveburnett:
Thanks for the release note entry! Suggest rephrasing to follow the [Order of changes](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines#order-of-changes) in the Release Notes Guidelines, and some nit formatting fixes. 

Maybe something like this? Please revise if my draft doesn't accurately describe the PR. 

```
== RELEASE NOTES ==

* Improve exchange encoding in the planner by falling back from columnar encoding to row wise encoding for wide (500 or more columns) data sets. :pr:`23929`
```

presto-main/src/main/java/com/facebook/presto/sql/planner/PlanOptimizers.java
```diff
@@ -890,13 +891,13 @@ public PlanOptimizers(
...
         builder.add(new IterativeOptimizer(
-                metadata,
-                ruleStats,
-                statsCalculator,
-                costCalculator,
-                ImmutableSet.of(
-                        new PushPartialAggregationThroughJoin(),
-                        new PushPartialAggregationThroughExchange(metadata.getFunctionAndTypeManager(), featuresConfig.isNativeExecutionEnabled()))),
+                        metadata,
+                        ruleStats,
```
rschlussel, line 895: why does autoreformatting change this? it doesn't line up with anything else in the class.

presto-main/src/main/java/com/facebook/presto/sql/planner/SystemPartitioningHandle.java
```diff
@@ -110,6 +110,18 @@ public boolean isCoordinatorOnly()
...
+    @Override
+    public boolean isBroadcast()
+    {
+        return function == SystemPartitionFunction.BROADCAST;
+    }
+
+    @Override
+    public boolean isArbitrary()
+    {
+        return function == SystemPartitionFunction.ROUND_ROBIN;
```
rschlussel, line 122: Does this intentionally exclude ARBITRARY_DISTRIBUTION which has SystemPartitioningFunction.UNKNOWN?  makes the name a bit confusing.

presto-main/src/test/java/com/facebook/presto/server/remotetask/TestHttpRemoteTask.java
```diff
@@ -245,12 +246,14 @@ public void testHTTPRemoteBadTaskSize()
         httpRemoteTaskFactory.stop();
 
         assertTrue(remoteTask.getTaskStatus().getState().isDone(), format("TaskStatus is not in a done state: %s", remoteTask.getTaskStatus()));
-        assertEquals(getOnlyElement(remoteTask.getTaskStatus().getFailures()).getMessage(), "TaskUpdate size of 1.97kB has exceeded the limit of 1kB");
+        assertThat(getOnlyElement(remoteTask.getTaskStatus().getFailures()).getMessage())
+                .matches("TaskUpdate size of .+? has exceeded the limit of 1kB");
```
rschlussel, line 250: do we expect this change to impact the taskUpdate size? by how much (might it cause problems)?

presto-main/src/test/java/com/facebook/presto/sql/planner/iterative/rule/TestDetermineRemotePartitionedExchangeEncoding.java
```diff
@@ -0,0 +1,240 @@
...
+
+    @AfterClass(alwaysRun = true)
+    public void tearDown()
+    {
+        tester.close();
+        tester = null;
+    }
+
+    @Test
+    public void testSapphire()
```
rschlussel, line 86: nit: call this testPrestoOnSpark (and ditto for assertForSapphire)

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/DetermineRemotePartitionedExchangeEncoding.java
```diff
@@ -0,0 +1,136 @@
...
+    {
+        return nativeExecution || prestoSparkExecutionEnvironment;
+    }
+
+    @Override
+    public Result apply(ExchangeNode node, Captures captures, Context context)
+    {
+        if (prestoSparkExecutionEnvironment) {
+            // In Presto on Spark, row-wise encoding is always used for non-special shuffles (i.e., excluding broadcast, single, and arbitrary shuffles).
+            // To accurately reflect this in the plan, the exchange encoding is set here.
```
rschlussel, line 67: maybe note that presto on spark doesn't actually use the fields from the plan, though.

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/DetermineRemotePartitionedExchangeEncoding.java
```diff
@@ -0,0 +1,136 @@
...
+            return Result.empty();
+        }
+        // otherwise switch to row-wise
+        return Result.ofPlanNode(node.withRowWiseEncoding());
+    }
+
+    private Result determineForNativeExecution(Session session, ExchangeNode node)
+    {
+        // keep columnar for special cases
+        if (node.getPartitioningScheme().isSingleOrBroadcastOrArbitrary()) {
```
rschlussel, line 94: I'm curious why we check that it's not single or broadcast or arbitrary instead of checking that it is a fixed hash distribution? What other things are we trying to include here?

jaystarshot:
> `For example an exchange of 1000 VARCHAR columns into 300 partitions requires 900,000 buffers to be allocated. Even when the allocations are as small as 128B per buffer the amount of memory allocated can be considerable (110MB) and can be far exceeding the PartitionedOutput operator budget (usually ~32MB for ~300 partitions).`

Sorry a dumb question but if the number of rows are 1000 per batch won't we need the same amount of buffer size on the receiver and sender?

presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
```diff
@@ -3100,4 +3106,9 @@ public static boolean isInlineProjectionsOnValues(Session session)
     {
         return session.getSystemProperty(INLINE_PROJECTIONS_ON_VALUES, Boolean.class);
     }
+
+    public static int getMinColumnarEncodingChannelsToPreferRowWiseEncoding(Session session)
+    {
+        return session.getSystemProperty(MIN_COLUMNAR_ENCODING_CHANNELS_TO_PREFER_ROW_WISE_ENCODING, Integer.class);
```
xiaoxmeng, line 3112: What's the default value for this? How to disable this? by setting it to an extreme large value? thanks!

xiaoxmeng (commented):
@arhimondr looks good. Thanks!

presto-main/src/main/java/com/facebook/presto/sql/planner/PlanOptimizers.java
arhimondr: Good question. Something must be wrong with my formatter. Reverted it for now.

presto-main/src/main/java/com/facebook/presto/sql/planner/SystemPartitioningHandle.java
arhimondr: `ARBITRARY_DISTRIBUTION` looks like a hack used by the `AssignUniqueId`. Basically it is used to say `uniqueId`'s are grouped together, but the function used to group them together is "unknown".

This partitioning is never used to actually partition data, as the rule how to partition is not defined. It is used only to state as a matter of fact that something is partitioned based on a certain set of keys, but partitioning function is unknown.

presto-main/src/test/java/com/facebook/presto/server/remotetask/TestHttpRemoteTask.java
arhimondr: The new field increases the task update size by a couple of bytes so the test fails.

I was trying to understand the intention of this test, and it seems like it was added to check if the error message is properly formatted. 

There's a test down below that injects a JSON of a certain size and checks if the size matches exactly. It felt like checking here for the exact size may not be necessary.

@misterjpapa could you please take a look and advice?

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/DetermineRemotePartitionedExchangeEncoding.java
arhimondr: Technically a connector can provide any partitioning function. In most cases it is hash based, but it can be anything. The engine provides a special case shortcuts for certain partitioning functions it knows the semantics of.

For example the engine knows that for "single", "arbitrary" or "broadcast" partitioning can be done at the page level and doesn't necessarily need to be done at the row level. This allows the engine to avoid additional cost associated with splitting pages. Hence these partitioning functions are rather an exception than a rule, hence it felt it makes more sense to single them out.

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/AddExchanges.java
```diff
@@ -1250,6 +1253,7 @@ public PlanWithProperties visitUnion(UnionNode node, PreferredProperties parentP
                                                 source.getNode().getOutputVariables(),
                                                 Optional.empty(),
                                                 nullsAndAnyReplicated,
+                                                COLUMNAR,
```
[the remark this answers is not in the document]
arhimondr: At the step when exchanges are added the optimizer sets the encoding to COLUMNAR. There's a rule downstream checking what would be the optimal encoding and switching to ROW_WISE is necessary. 

At the point when exchanges are added the number of columns exchanged may not be final, hence it might be too early to pick an optimal one at this step.

presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
arhimondr: > What's the default value for this?

1000 columnar streams. For example VARCHAR - 3 streams (nulls, values, offsets). BIGINT - 2 streams (nulls, values. ARRAY<VARCHAR> - 5 streams, etc.

> How to disable this? by setting it to an extreme large value?

Yes

arhimondr:
Thank you for the suggestions @steveburnett . @xiaoxmeng landed the execution change first so I was able to simplify the PR description and release notes.

arhimondr:
@jaystarshot 

The number of allocated buffers scales with the number of partitions and the number of columnar streams.

For example, let's assume there's an exchange of a data set consisting of a single BIGINT column.

BIGINT  data is represented as 2 columnar streams: `is_null`, `data`.

When partitioning data into 2 partitions the implementation needs to allocate 4 arrays:

- `is_null` - Partition 1
- `data` - Partition 1
- `is_null` - Partition 2
- `data` - Partition 2

If the number of streams and partitions is high the number of buffers allocated can easily reach a million (3000 streams * 333 partitions).

Buffers are usually allocated with a reasonable headroom to avoid re-allocation for every row. In Velox the minimum size for a single buffer is 128B. With 1M buffers the total buffer space allocated will be equal to 128MB what is significantly higher than the default buffer size limit of 32MB.

Another problem is CPU. A columnar implementation has to update each of the `3000` streams for every row, making it very inefficient.

arhimondr:
@xiaoxmeng @rschlussel Thank you for the review. Updated. Could you please take an another look?

xiaoxmeng (approved):

rschlussel (approved):

Issue #23338, opened 2024-07-30
PrestoExchangeSource HttpClient histogram metric is blocking query


karteekmurthys:
[PrometheusStatsReporter](https://github.com/prestodb/presto/pull/22360) updates worker stats synchronously. This is fine as most of the stats are updated through PeriodicTaskManager in a separate thread periodically without blocking queries.

But [this specific metric](https://github.com/prestodb/presto/blob/master/presto-native-execution/presto_cpp/main/PrestoExchangeSource.cpp#L117-L119) is updated in the query path and is updated way too frequently.
```
        RECORD_HISTOGRAM_METRIC_VALUE(
            kCounterHttpClientPrestoExchangeOnBodyBytes, bufferBytes);
      });
```
The Q67 of SF10K TPCDS workload takes several hours to finish since we merged PrometheusStatsReporter. We are working on how to improve this in the stats reporter but we should also avoid reporting metrics in the query path if possible.

Request to review if `presto_cpp.http.client.presto_exchange_source.on_body_bytes` can be removed or made optional for tracking.

We tested Q67 at SF10K by removing this metric and the query completes successfully.

We also have a workaround of updating [histogram asynchronously](https://github.com/prestodb/presto/commit/437a333fd9b024e9c66efd9c84de94a35663e3e9) in PrometheusStatsReporter. This solution was tested successfully as well against Q67

karteekmurthys:
cc: @majetideepak @amitkdutta

karteekmurthys:
@pramodsatya gathered some stats total time spent different locations in Presto CPP where RECORD_HISTOGRAM is called. The Q67 query was run for over an hour and `PrestoExchangeSource.cpp:117`  took total of 3256278906 ms in hour of query run.


```
reg/add   function                               line num
Add       addRootPool                            Memory.cpp:248                          76
          clearThread                            Driver.h:153                        166326
          operator()                             ExchangeClient.cpp:163               87166
                                                 ExchangeClient.cpp:167               61760
                                                 ExchangeClient.cpp:170               31759
                                                 FileHandle.cpp:64                       45
                                                 PrestoExchangeSource.cpp:117    3256278906
          runInternal                            Driver.cpp:522                      348599
          updatePrestoExchangeSourceMemoryStats  PeriodicTaskManager.cpp:268          18196
          ~MemoryPoolImpl                        MemoryPool.cpp:462                      45
          ~ScopedArbitration                     SharedArbitrator.cpp:880              1110
[...]

PR #2506, opened 2015-03-11
Merge new array/map/row memory format from Treasure Data

cberner:
A few modifications on top of https://github.com/facebook/presto/pull/2240

PR #23948, opened 2024-11-04
[native] Report different thread pool executors' stats

tanjialiang:
folly::ThreadPoolExecutor::getPoolStats() provides better and more detailed thread pool stats for both cpu thread pool and io thread pool. We shall leverage that API to provide with more accurate metrics instead of having our own estimation by enqueue tasks to the queue.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
+++ b/presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
@@ -69,8 +69,142 @@ folly::StringPiece getCounterForBlockingReason(
     default:
       return {};
   }
-  return {};
 }
+
+class ThreadPoolExecutorStatsReporter {
+ public:
+  ThreadPoolExecutorStatsReporter(
+      folly::ThreadPoolExecutor* executor,
+      const std::string& poolName,
+      uint32_t estimatedMaxNumTasks)
+      : executor_(executor),
+        numThreadsMetricName_(
+            fmt::format(kCounterThreadPoolNumThreadsFormat, poolName)),
+        numActiveThreadsMetricName_(
+            fmt::format(kCounterThreadPoolNumActiveThreadsFormat, poolName)),
+        numPendingTasksMetricName_(
+            fmt::format(kCounterThreadPoolNumPendingTasksFormat, poolName)),
+        numTotalTasksMetricName_(
+            fmt::format(kCounterThreadPoolNumTotalTasksFormat, poolName)),
+        maxIdleTimeNsMetricName_(
+            fmt::format(kCounterThreadPoolMaxIdleTimeNsFormat, poolName)) {
+    VELOX_CHECK_NOT_NULL(executor_);
+    const auto numThreads = executor_->numThreads();
+    const auto numHistogramBuckets = 100;
+    DEFINE_METRIC(numThreadsMetricName_, facebook::velox::StatType::AVG);
+    DEFINE_HISTOGRAM_METRIC(
+        numActiveThreadsMetricName_, 1, 0, numThreads, 50, 90, 100);
+    DEFINE_HISTOGRAM_METRIC(
+        numPendingTasksMetricName_,
+        estimatedMaxNumTasks / numHistogramBuckets,
+        0,
+        estimatedMaxNumTasks,
+        50,
+        90,
+        100);
+    DEFINE_HISTOGRAM_METRIC(
+        numTotalTasksMetricName_,
+        estimatedMaxNumTasks / numHistogramBuckets,
+        0,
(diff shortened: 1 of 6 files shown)


presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
```diff
@@ -186,19 +326,40 @@ void PeriodicTaskManager::updateExecutorStats() {
...
+  std::vector<ThreadPoolExecutorStatsReporter> reporters;
+  auto addExecutorFunc = [&](folly::ThreadPoolExecutor* executor,
+                             const std::string& executorName,
+                             uint32_t estimatedMaxNumTasks) {
+    if (executor != nullptr) {
+      reporters.push_back(ThreadPoolExecutorStatsReporter(
+          executor, executorName, estimatedMaxNumTasks));
+    }
+  };
+  addExecutorFunc(driverCPUExecutor_, "driver_cpu_executor", 1'000);
```
xiaoxmeng, line 349: This is the estimated max? Can we bump up a bit? thanks!

presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
```diff
@@ -71,6 +71,140 @@ folly::StringPiece getCounterForBlockingReason(
...
+        30'000'000'000 /* 30s */,
+        0,
+        300'000'000'000 /* 300s */,
+        50,
+        90,
+        100);
+  }
+
+  void report() const {
+    auto poolStats = executor_->getPoolStats();
```
xiaoxmeng, line 124:
```
const auto poolStats = ...
```

xiaoxmeng (approved):
@tanjialiang do we want to remove the old metrics? Thanks!

tanjialiang:
> @tanjialiang do we want to remove the old metrics? Thanks!

Sure, removed.

xiaoxmeng (dismissed):

xiaoxmeng (approved):

presto-native-execution/presto_cpp/main/PeriodicTaskManager.cpp
```diff
@@ -69,8 +69,142 @@ folly::StringPiece getCounterForBlockingReason(
...
+        90,
+        100);
+    DEFINE_HISTOGRAM_METRIC(
+        maxIdleTimeNsMetricName_,
+        10'000'000'000 /* 10s */,
+        0,
+        300'000'000'000 /* 300s */,
+        50,
+        90,
+        100);
```
lingbin, line 120: @tanjialiang  I have a question and would appreciate your guidance:

Why are the metrics other than `numThreadsMetricName_` (AVG type) declared as `StatType::HISTOGRAM` type instead of also being declared as `StatType::AVG` type? Is there a specific consideration or benefit to this?

In Meta's internal monitoring system, is there a difference in the monitoring graphs for the `numThreadsMetricName_` and `numActiveThreadsMetricName_` metrics?

lingbin: @tanjialiang  The reason I'm asking this is because updating `StatType::HISTOGRAM` type metrics is significantly **more expensive** than updating `StatType::AVG` type metrics.   This is because it not only updates the internal count by bucket (the more buckets, the greater the computational overhead), but also calculates various quantile values ​​(which are very costly).   

I've seen **several performance issues**(#25058  #23338 ) related to the Histogram type, which are **still unresolved**, and many people are even considering completely disabling Histograms(#2506 ), despite this resulting in missing metrics. (We currently use a similar approach, disabling all Histogram types, which causes many metrics to be unavailable, including several thread pool metrics here).

Although the above problem was found in `PrometheusStatsExporter`(the **only one** available in the community), I believe it is likely a widespread issue (including the metric library used internally by Meta), as the computational cost of the Histogram type is undoubtedly greater than that of the Avg type.

For the metrics here, I feel that declaring them as `AVG` type would be sufficient, but I'm unsure about their internal usage within Meta, so I'd like to consult you. Looking forward to your reply, thanks.

lingbin: @tanjialiang   I think these thread pool-related metrics are very important, and thank you for adding them.  However, it's unfortunate that they cannot be enabled because they are Histogram types.  

If I could get your approval, I would like to submit a PR to change them to `StatType::AVG` type.

lingbin: @xiaoxmeng  Could you please take a look at this problem as well?

PR #23882, opened 2024-10-24
Added catalog and schema level access checks in USE statement

annmegha:
Added catalog and schema level access checks when a user tries USE statement

## Motivation and Context

A user with any role who has not been granted access to a specific catalog or schema is still able to execute the USE command. The expected behavior is that such users should receive an "Access Denied" error when attempting to access a catalog or schema for which they lack permission.
Refer : https://github.com/prestodb/presto/issues/23881

## Test Plan

Added the related testcases to the code change

## Contributor checklist

- [ ] Please make sure your submission complies with our [development](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#development), [formatting](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#formatting), [commit message](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#commit-formatting-and-pull-requests), and [attribution guidelines](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#attribution).
- [ ] PR description addresses the issue accurately and concisely.  If the change is non-trivial, a GitHub Issue is referenced.
- [ ] Documented new properties (with its default value), SQL syntax, functions, or other functionality.
- [ ] If release notes are required, they follow the [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines).
- [ ] Adequate tests were added if applicable.
- [ ] CI passed.

## Release Notes
Please follow [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines) and fill in the release notes below.

```
== RELEASE NOTES ==

General Changes
* Added catalog and schema level access checks in USE statement :pr:`23882 `
```

PR #23965, opened 2024-11-06
Revert "Added catalog and schema level access checks in USE statement"

tdcmeehan:
Reverts prestodb/presto#23882

This is causing failures on CI trunk.


ZacBlanco:
Thanks for opening this - I think the tests did run in the original PR, but the first action that ran the test ended up failing. I think it's just a flaky test? But I definitely prefer reverting and fixing the flakiness first.

rschlussel (approved):

PR #23990, opened 2024-11-08
Allow hbo stats for failed stages with local memory

jaystarshot:
## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* ... :pr:`12345`
* ... :pr:`12345`

Hive Connector Changes
* ... :pr:`12345`
* ... :pr:`12345`
```

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/cost/HistoryBasedPlanStatisticsTracker.java
+++ b/presto-main/src/main/java/com/facebook/presto/cost/HistoryBasedPlanStatisticsTracker.java
@@ -67,6 +67,7 @@ import static com.facebook.presto.common.resourceGroups.QueryType.INSERT;
 import static com.facebook.presto.common.resourceGroups.QueryType.SELECT;
 import static com.facebook.presto.cost.HistoricalPlanStatisticsUtil.updatePlanStatistics;
 import static com.facebook.presto.cost.HistoryBasedPlanStatisticsManager.historyBasedPlanCanonicalizationStrategyList;
+import static com.facebook.presto.spi.StandardErrorCode.EXCEEDED_LOCAL_MEMORY_LIMIT;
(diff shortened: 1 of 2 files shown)


jaystarshot:
@feilong-liu What do you think? any particular reason to not do this before?

feilong-liu:
> @feilong-liu What do you think? any particular reason to not do this before?

Why EXCEEDED_LOCAL_MEMORY_LIMIT is special, I mean will the statistics be accurate for this error?

jaystarshot:
The failed stage will likely have metrics of the worst offending one so should help in joins?

feilong-liu:
> The failed stage when there is a oom error will likely have metrics of the worst offending task so should help in joins/aggs?

HBO considers the statistics recorded accurate, and it will be shared among queries which has the same plan hash (not necessarily the same query). 
Also this change will record statistics for all plan nodes, rather than the specific join/aggregation which has OOM. I think a better way is to record the statistics only for this failed node and add new type of statistics saying it's at least this number, rather than an accurate estimation.

jaystarshot:
Yes filtering by plannode is better but its a bigger change than this. I think using OOM stats as accurate won't have an adverse effect since OOM is a special case because If the current plan is seen again in any query it will again oom.

feilong-liu:
> Yes filtering by plannode is better but its a bigger change than this. Also operator stast are not always awailable for all operators within a stage. I have typically seen only 1-2 operators per stage having runtime stats. I think using OOM stats as accurate won't have an adverse effect since OOM is a special case because If the current plan is seen again in any query it will again oom.

If you want to try, I suggest to add a session property to gate it.

jaystarshot:
Closing for now, will check if considering the individual operator stats can be filtered

PR #23991, opened 2024-11-09
[jdk upgrade] maven checks for presto-orc

ZacBlanco:
pass maven-checks for presto-orc

- fixes 1 checkstyle and 288 modernizer errors


ZacBlanco:
successful checks went from 49 -> 50 and presto-orc passed.

Issue #23998, opened 2024-11-11
Support Google compressed string polylines 


jquirke:
## Expected Behavior or Use Case

Add geospatial functions to work with (i.e. transform) the popular Google polyline string encoding as used in Google Maps and other products to and from the existing Presto geometry constructs like ST_Polyline, ST_Point.

These are efficiently encoded as per:
https://developers.google.com/maps/documentation/utilities/polylinealgorithm

## Presto Component, Service, or Connector

Geospatial functions

## Possible Implementation

A possible implementation for decoding might look like as suggested in PR:  #23999 , where a google maps string is decoded to an array of ST_Points. Alternatively, we could transform it into an ESRI polyline.

```
presto> SELECT google_polyline_decode('_p~iF~ps|U_ulLnnqC_mqNvxq`@')
                                _col0
----------------------------------------------------------------------
 [POINT (38.5 -120.2), POINT (40.7 -120.95), POINT (43.252 -126.453)]
(1 row)

```

A possible implementation for encoding might look like as suggested in PR:  #23999 , where an array of ST_Points is encoded to a google polyline string.

```
presto> SELECT google_polyline_encode(ARRAY[ST_Point (38.5, -120.2), ST_Point (40.7, -120.95), ST_Point (43.252, -126.453)])

            _col0
-----------------------------
 _p~iF~ps|U_ulLnnqC_mqNvxq`@
(1 row)

```

## Context

This is an increasingly popular format [1], [2] for storing complex maps routes efficiently.

For example routes of Uber trips are efficiently stored using this format. 

[1] https://github.com/frederickjansen/polyline/stargazers
[2] https://github.com/urschrei/pypolyline/stargazers 
[3] https://stackoverflow.com/search?q=google+polyline

jquirke:
Implemented in #23999

PR #23999, opened 2024-11-11
Add functions to support Google polylines

jquirke:
Support the popular and efficient Google polyline format by adding a function "gmaps_decode_polyline" that decodes a polyline string into an array of ST_Points, and a "gmaps_encode_polyline" that does the reverse.

Implements #23998  

## Description

- Adds google_decode_polyline function (VARCHAR -> ARRAY of ST_Point), with optional precision argument
- Adds google_encode_polyline function (ARRAY of ST_Point -> VARCHAR), with optional precision argument

## Motivation and Context

This is an increasingly popular format [1], [2], [3] for storing complex maps routes efficiently. 
Implementing functionality to convert to the native Presto ST_Geometry types would be incredibly useful for working with this compressed route data.

For example routes of Uber trips are efficiently stored using this format.

[1] https://github.com/frederickjansen/polyline/stargazers
[2] https://github.com/urschrei/pypolyline/stargazers
[3] https://stackoverflow.com/search?q=google+polyline

## Test Plan

Tests to cover successful decode and corrupted polyline string.

```
mvnw test
```

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Add :func:`google_polyline_decode` function to convert Google polyline to Presto ST_Geometry types. :pr:`23999`
* Add :func:`google_polyline_encode` function to convert Presto ST_Geometry to Google polyline types. :pr:`23999`

```


steveburnett:
Thanks! A couple of formatting nits in the release note entry. 

```
== RELEASE NOTES ==

General Changes
* Add :func:`gmaps_polyline_decode` function to convert Google polyline to Presto ST_Geometry types. :pr:`23999`
```

presto-main/src/main/java/com/facebook/presto/geospatial/GeoFunctions.java
```diff
@@ -1341,4 +1343,46 @@ public Slice next()
             }
         };
     }
+
+    @SqlNullable
+    @Description("Decodes a GMaps-Polyline string into an array of Points")
+    @ScalarFunction("gmaps_polyline_decode")
+    @SqlType("array(" + GEOMETRY_TYPE_NAME + ")")
+    public static Block gMapsPolylineDecode(@SqlType(StandardTypes.VARCHAR) Slice polyline) throws PrestoException
```
tdcmeehan, line 1351: I believe this is implemented according to a spec or documentation.  If so, it would be helpful to link that documentation here.

presto-main/src/main/java/com/facebook/presto/geospatial/GeoFunctions.java
```diff
@@ -1341,4 +1343,46 @@ public Slice next()
...
+    {
+        final double precision = Math.pow(10.0, (double) DEFAULT_POLYLINE_PRECISION_EXPONENT);
```
tdcmeehan, line 1353: Does this presume the underlying polyline has this precision?  i.e. if a polyline was encoded using higher precision, would this result in information loss?

presto-main/src/main/java/com/facebook/presto/geospatial/GeoFunctions.java
```diff
@@ -1341,4 +1343,46 @@ public Slice next()
             }
         };
     }
+
+    @SqlNullable
```
tdcmeehan, line 1347: When does the method return null?

presto-main/src/test/java/com/facebook/presto/geospatial/TestGeoFunctions.java
```diff
@@ -1371,4 +1371,13 @@ private void assertInvalidGeometryJson(String json)
     {
         assertInvalidFunction("geometry_from_geojson('" + json + "')", "Invalid GeoJSON:.*");
     }
+
+    @Test
+    public void testGMapsPolyline()
```
tdcmeehan, line 1376: Two test cases seems sparse.  Are you sure this is sufficient test coverage?  If we look at polyline libraries, how extensive are their test cases and is there anything we can learn from them?

presto-main/src/main/java/com/facebook/presto/geospatial/GeoFunctions.java
jquirke: It doesn't. Good call. fixed.

presto-main/src/main/java/com/facebook/presto/geospatial/GeoFunctions.java
jquirke: Yes. I have added a link to https://developers.google.com/maps/documentation/utilities/polylinealgorithm (this was in the Feature Request issue, but I agree that involves too much code archaeology to find via git blame -> git hub.

presto-main/src/main/java/com/facebook/presto/geospatial/GeoFunctions.java
jquirke: Yes. The standard and almost all implementations and use cases use 1e5. There are a couple of cases out there that use a precision of 6.

Ordinarily, an optional argument overriding the precision in the function might be worthwhile ([this is how the python library does it](https://pypi.org/project/polyline/)). 

However, please correct me if I'm wrong; so I have specified a separate function that specifies precision.

presto-main/src/test/java/com/facebook/presto/geospatial/TestGeoFunctions.java
jquirke: I tend to agree. Ordinarily I would just grab some actual trip data and use that to form test cases, but on the public internet that reveals PII.

Instead I've created a case manually - a case with 

(1) precision in the input floats higher than supported
(2) duplicate points in a row
(3) line crosses back over itself (not very significant, but hey)
(4) line terminates at the origin (non consecutive duplicate points)

jquirke:
> Thanks! A couple of formatting nits in the release note entry.
> 
> ```
> == RELEASE NOTES ==
> 
> General Changes
> * Add :func:`gmaps_polyline_decode` function to convert Google polyline to Presto ST_Geometry types. :pr:`23999`
> ```

Also, re naming of this function, strictly speaking this function is not a google maps only tool, but "Google" and "Google Maps" is a registered trade mark in the United States and elsewhere. Please advise.

presto-main/src/main/java/com/facebook/presto/geospatial/GeoFunctions.java
tdcmeehan: To me this seems like [`approx_percentile`](https://prestodb.io/docs/current/functions/aggregate.html#approx_percentile-x-percentage-same-as-x), which has two forms: one in which you can specify accuracy, and one in which we use a default accuracy and you don't specify it.  I think we can take the same approach here.

tdcmeehan (commented):
Just some nits.

Is there any plan to backport this to Velox?  If so, I think it would be worthwhile to work on that in parallel with this PR so we can use the fuzzer to work out any potential bugs.  (Presto does not have a fuzzer.)

steveburnett:
> > Thanks! A couple of formatting nits in the release note entry.
> > ```
> > == RELEASE NOTES ==
> > 
> > General Changes
> > * Add :func:`gmaps_polyline_decode` function to convert Google polyline to Presto ST_Geometry types. :pr:`23999`
> > ```
> 
> Also, re naming of this function, strictly speaking this function is not a google maps only tool, but "Google" and "Google Maps" is a registered trade mark in the United States and elsewhere. Please advise.

I don't have any specific response. @tdcmeehan?

tdcmeehan:
I think this falls under fair use.  We already have functions and reference [Bing tiles](https://prestodb.io/docs/current/functions/geospatial.html#bing-tiles), this seems like the same sort of thing, we can use this name because it is necessary to describe the format.

jaystarshot:
@tdcmeehan what is the policy for presto c++ support?  When should we start enforcing required c++ implementation?

tdcmeehan:
There's no need to enforce that they be consistent.

What often happens is, because Velox has a sophisticated fuzzer, when the C++ version is developed, the fuzzer will surface bugs in the original Java implementation.  (Even what seems like simple functions have very hard to catch edge cases that are difficult to catch with example based testing.) So if there is an intention to add a Velox counterpart, I'd recommend we add it now so we can use the fuzzer test to work out the bugs now.

jquirke:
> > > > Thanks! A couple of formatting nits in the release note entry.
> > > > ```
> > > > == RELEASE NOTES ==
> > > > 
> > > > General Changes
> > > > * Add :func:`gmaps_polyline_decode` function to convert Google polyline to Presto ST_Geometry types. :pr:`23999`
> > > > ```
> > > 
> > > 
> > > Also, re naming of this function, strictly speaking this function is not a google maps only tool, but "Google" and "Google Maps" is a registered trade mark in the United States and elsewhere. Please advise.
> > 
> > 
> > I don't have any specific response. @tdcmeehan?
> 
> I think this falls under [fair use](https://en.wikipedia.org/wiki/Nominative_use). We already have functions and reference [Bing tiles](https://prestodb.io/docs/current/functions/geospatial.html#bing-tiles), this seems like the same sort of thing, we can use this name because it is necessary to describe the format.

Great, I renamed it: s/gmaps/google/

jquirke:
> @jquirke LGTM, please squash commits

Done.

jquirke:
> > Thanks! A couple of formatting nits in the release note entry.
> > ```
> > == RELEASE NOTES ==
> > 
> > General Changes
> > * Add :func:`gmaps_polyline_decode` function to convert Google polyline to Presto ST_Geometry types. :pr:`23999`
> > ```
> 
> Also, re naming of this function, strictly speaking this function is not a google maps only tool, but "Google" and "Google Maps" is a registered trade mark in the United States and elsewhere. Please advise.

jquirke:
Just a quick follow up - are we moving ahead with incorporating this as a geospatial pair of functions?

tdcmeehan (approved):

tdcmeehan:
One final nit, looks like the commit message is referencing an unrelated Github issue, can you please fix that?

jquirke:
> One final nit, looks like the commit message is referencing an unrelated Github issue, can you please fix that?

Done, although this re-triggered all 50+ CI/CD checks and one of the flaky ones ("smoke") failed. Not sure how to re-run (or even if I can re-run) a single check in isolation.

tdcmeehan:
Thanks for the contribution @jquirke !

PR #24004, opened 2024-11-11
Fix http splitting vulnerabilities

auden-woolfson:
## Description
Dealing with vulnerabilities found in static scan labeled http splitting vulnerability

[https://cwe.mitre.org/data/definitions/113.html](https://cwe.mitre.org/data/definitions/113.html)

```
== RELEASE NOTES ==

General Changes
* Throw exception on invalid http headers in async page transport servlet. :pr:`24004`
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/server/AsyncPageTransportServlet.java
+++ b/presto-main/src/main/java/com/facebook/presto/server/AsyncPageTransportServlet.java
@@ -43,6 +43,7 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
 import java.io.IOException;
+import java.util.Enumeration;
 import java.util.List;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ScheduledExecutorService;
@@ -127,6 +128,20 @@ public class AsyncPageTransportServlet
         TaskId taskId = null;
         OutputBufferId bufferId = null;
         long token = 0;
+
+        if (request != null) {
+            Enumeration<String> headerNames = request.getHeaderNames();
+            while (headerNames.hasMoreElements()) {
+                String headerName = headerNames.nextElement();
+                String headerValue = request.getHeader(headerName);
+                if (headerName.contains("\r") || headerName.contains("\n")) {
+                    throw new IllegalArgumentException(format("Invalid header name: %s", headerName));
+                }
+                if (headerValue.contains("\r") || headerValue.contains("\n")) {
+                    throw new IllegalArgumentException(format("Invalid header value: %s", headerValue));
+                }
+            }
+        }
 
         int previousIndex = -1;
         for (int part = 0; part < 8; part++) {
--- a/presto-tests/src/test/java/com/facebook/presto/server/TestAsyncPageTransportServlet.java
+++ b/presto-tests/src/test/java/com/facebook/presto/server/TestAsyncPageTransportServlet.java
@@ -15,6 +15,10 @@ package com.facebook.presto.server;
 
 import com.facebook.presto.execution.TaskId;
 import com.facebook.presto.execution.buffer.OutputBuffers.OutputBufferId;
+import com.google.common.collect.ImmutableListMultimap;
+import com.google.common.collect.ImmutableMap;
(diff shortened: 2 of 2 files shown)


steveburnett:
Thanks for the release note entry! Suggest revising to follow the [Order of changes](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines#order-of-changes) in the Release Notes Guidelines. Maybe something like? 

```
== RELEASE NOTES ==

General Changes
* Improve HTTP responses from async page transport servlet, by sanitizing the URI before use. :pr:`24004`
```

Please revise my suggestion if my phrasing doesn't correctly describe the work you've done.

presto-main/src/main/java/com/facebook/presto/server/AsyncPageTransportServlet.java
```diff
@@ -128,12 +128,14 @@ protected void parseURI(String requestURI, HttpServletRequest request, HttpServl
         OutputBufferId bufferId = null;
         long token = 0;
 
+        String sanitizedRequestURI = requestURI.replaceAll("[\\r\\n]", "");
```
czentgr, line 131: So this removes \r and \n from the URI. But having these chars would not be a valid URI in the first place. The server listener doesn't filter this out already?
The vulnerability talks about having these values in the header - because these values are used to separate header and body in HTTP.

Validating the input is fine. Just surprised this isn't already taken care of.

pdabre12: +1. 
We need to sanitize the request headers in order to handle this vulnerability.

ZacBlanco (changes requested):
Is it possible to write a test which can verify the new behavior?

presto-main/src/main/java/com/facebook/presto/server/AsyncPageTransportServlet.java
auden-woolfson: So just to clarify, you are suggesting I get all headers from the request variable and sanitize each one using the same method here?

presto-main/src/main/java/com/facebook/presto/server/AsyncPageTransportServlet.java
```diff
@@ -128,6 +129,18 @@ protected void parseURI(String requestURI, HttpServletRequest request, HttpServl
         OutputBufferId bufferId = null;
         long token = 0;
 
+        Enumeration<String> headerNames = request.getHeaderNames();
+        while (headerNames.hasMoreElements()) {
```
pdabre12, line 133: I think it would be better to not throw if the header contains a "\r" or "\n". Instead remove these characters, and proceed with the sanitized header string.

pdabre12 (commented):
Thanks @auden-woolfson.
I have an initial comment.

presto-main/src/main/java/com/facebook/presto/server/AsyncPageTransportServlet.java
auden-woolfson: `HttpServletRequest` is read only. I can only make a new request wrapper that returns the header values that have "\r" and "\n" removed. Not sure this is actually what we want.

ZacBlanco: [According to the grammar in RFC 7230](https://www.rfc-editor.org/rfc/rfc7230#section-3.2.6) `\r` and `\n` are not valid characters at all in header field names or in header values. I think it is fair to just throw an exception. It will notify the upstream request/proxy they've done something invalid to the request.

ZacBlanco (approved):
I think this approach is fine, but I wonder how much overhead this has. I'm curious about the impact on request throughput now that we're doing this additional header validation. I'm sure it won't be too high, but if we somehow happen to be sending tons of headers, it could imagine it might slow things down. A microbenchmark would be nice, but not necessary

pdabre12 (dismissed):
LGTM, please squash the commits.

ZacBlanco (approved):

pdabre12 (approved):

tdcmeehan (approved):

PR #24025, opened 2024-11-12
[jdk upgrade] maven checks on presto-base-jdbc

auden-woolfson:
Build success on base jdbc module

### DIFF
--- a/presto-base-jdbc/pom.xml
+++ b/presto-base-jdbc/pom.xml
@@ -180,6 +180,12 @@
         </dependency>
 
         <dependency>
+            <groupId>com.facebook.presto</groupId>
+            <artifactId>presto-analyzer</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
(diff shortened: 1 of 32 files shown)


ZacBlanco:
For your commit messages, can you update them to [follow the pattern](https://github.com/prestodb/presto/pull/23979/commits) we've been doing on the `upgradeJava17Bytecode` branch. It will help if/when we need to cherry pick bits down the road. Something like

- `modernizer for <module>`
- `dependency-scope for <module>`

auden-woolfson:
> For your commit messages, can you update them to [follow the pattern](https://github.com/prestodb/presto/pull/23979/commits) we've been doing on the `upgradeJava17Bytecode` branch. It will help if/when we need to cherry pick bits down the road. Something like
> 
>     * `modernizer for <module>`
> 
>     * `dependency-scope for <module>`

Ok, I just did it as one commit for this one since they weren't separate but I will follow the convention for the next one.

ZacBlanco (approved):

PR #24035, opened 2024-11-13
Upgrade cli

auden-woolfson:
## Description
presto-cli checks all passing


steveburnett:
Any documentation impact? Probably not to https://prestodb.io/docs/current/installation/cli.html , but maybe to https://prestodb.io/docs/current/clients/presto-cli.html ?

auden-woolfson:
> Any documentation impact? Probably not to https://prestodb.io/docs/current/installation/cli.html , but maybe to https://prestodb.io/docs/current/clients/presto-cli.html ?

There shouldn't be... all I did was change the scope of a dependency and refactor a few functions that the modernizer recommends. No actual change to functionality or docs

presto-cli/pom.xml
```diff
@@ -30,6 +30,12 @@
         <dependency>
             <groupId>com.facebook.presto</groupId>
             <artifactId>presto-client</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>com.facebook.airlift</groupId>
+                    <artifactId>json</artifactId>
+                </exclusion>
+            </exclusions>
```
ZacBlanco, line 38: why add this exclusion?

auden-woolfson: There is a dependency scope error as follows...

```
[ERROR] Found a problem with test-scoped dependency com.facebook.airlift:json
[ERROR] Scope compile was expected by artifact com.facebook.presto:presto-client:0.291-SNAPSHOT
[ERROR]
[ERROR] Dependency chain:
[ERROR] com.facebook.presto:presto-cli:jar:0.291-SNAPSHOT
[ERROR] \- com.facebook.presto:presto-client:jar:0.291-SNAPSHOT:compile
[ERROR]    \- com.facebook.airlift:json:jar:0.215:compile
```

It is including json as a transitive dependency in the compile scope when it is only needed in the test scope

ZacBlanco: Since the the json dependency is likely required by the `presto-client` jar we don't want to exclude it. For this case and any others that you come across in the future, the solution is to just add it to the set of ignored dependencies like in this commit: https://github.com/prestodb/presto/pull/23979/commits/024e1dbeba54cccd8d97526238d8f39b8aa05f38

PR #24040, opened 2024-11-14
[native] Disable plan validation failure for timestamp with timezone and ipaddress types for native execution.

amitkdutta:
## Description
As native engine now supports timestamp with timezone and ipaddress types, no need to make queries fail with plan validation by default. This change disables such validation failures by default.

## Motivation and Context
Keeping timestamp and ipaddress by default enabled like java engine.

## Impact
No new impact. If a production system explicitly set these configs, now they can remove those as default is enabled.

## Test Plan
Updated existing tests.

```
== NO RELEASE NOTE ==
```


tdcmeehan:
Should we also delete the code `CheckUnsupportedPrestissimoTypes`?

feilong-liu (approved):

aditi-pandit:
> Should we also delete the code `CheckUnsupportedPrestissimoTypes`?

Agree. We should remove the code as well. It doesn't provide any value anymore. We can re-add it if we want to bring the config back.

pramodsatya (commented):
Thanks @amitkdutta! The failing e2e tests with `IpAddress` and `TimestampWithTimeZone` types in `presto-native-tests` are running successfully with this fix.

amitkdutta:
@tdcmeehan @aditi-pandit Yes, will delete `CheckUnsupportedPrestissimoTypes`. We just launched timestamp and ipaddress support a week ago, and as the features are new, it could be possible to hit issues in next few weeks. If anything observed, we will have the opportunity to disable timestamp but keep the build with other fixes as each release contains many new correctness fixes with velox. Henece, I want to keep the check alive but disable by default. Once we pass through a few releases, we will be more confident to delete it completely.

feilong-liu (approved):

aditi-pandit (approved):
Thanks @amitkdutta

amitkdutta:
Now we have enabled all the timestamp and ipaddress tests in CI, and found [one result mismatch issue with checksum and ipaddress](https://github.com/prestodb/presto/blob/b823353e6d8986f95492832dba75ad61357e7432/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeAggregations.java#L331-L332)

Disabled the test for now and created an issue
https://github.com/prestodb/presto/issues/24130
CC: @kgpai @pramodsatya

PR #24041, opened 2024-11-14
maven checks for remaining dependencies of presto-hive

ZacBlanco:
maven checks passing for remaining dependencies before tackling presto-hive


ZacBlanco:
affected module all have passing maven checks

- presto-password-authenticators
- presto-session-property-managers
- presto-cache
- presto-memory
- presto-benchmark
- presto-hive-common

PR #24042, opened 2024-11-14
Changes to normalize the string to improve security

bibith4:
## Description
Normalized the unnormalized string to increase security
Made the changes in
presto-hive/src/main/java/com/facebook/presto/hive/HiveZeroRowFileCreator.java
presto-main/src/main/java/com/facebook/presto/storage/TempStorageManager.java

## Motivation and Context
When a system accepts external inputs usually uses a filtering or validation technique to check if the input is safe or a malicious input. When filtering techniques like blacklisting are used, certain characters whose use can be malicious can skip this filtering if they are encoded in other different format than the expected.Java API provides the Normalizer class along with its normalize method to transform a string into a proper format.

## Impact
no impact

## Test Plan
Verified existing Unit test case for above fix :
(1)presto-main/src/main/java/com/facebook/presto/testing/TestingTempStorageManager.java

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-hive/src/main/java/com/facebook/presto/hive/HiveZeroRowFileCreator.java
+++ b/presto-hive/src/main/java/com/facebook/presto/hive/HiveZeroRowFileCreator.java
@@ -34,6 +34,7 @@ import javax.inject.Inject;
 import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.nio.file.Paths;
+import java.text.Normalizer;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Properties;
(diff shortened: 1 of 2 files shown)


presto-hive/src/main/java/com/facebook/presto/hive/HiveZeroRowFileCreator.java
```diff
@@ -107,10 +108,12 @@ private byte[] generateZeroRowFile(
     {
         String tmpDirectoryPath = System.getProperty("java.io.tmpdir");
         String tmpFileName = format("presto-hive-zero-row-file-creator-%s-%s", session.getQueryId(), randomUUID().toString());
-        java.nio.file.Path tmpFilePath = Paths.get(tmpDirectoryPath, tmpFileName);
+        String normalizedDirPath = Normalizer.normalize(tmpDirectoryPath, Normalizer.Form.NFKC);
+        String normalizedFileName = Normalizer.normalize(tmpFileName, Normalizer.Form.NFKC);
```
aaneja, line 112: `tmpFileName` shouldn't need normalization since the the query-id and randomUUID should be regular ASCII characters as per design

bibith4: @aaneja As discussed marking the issue as false postive and closing this PR/

PR #24048, opened 2024-11-14
Fix regex vulnerabilites

auden-woolfson:
## Description
Check for matching regex in a different thread in query state info resource in order to prevent resource over usage

## Motivation and Context
Regex DOS vulnerability found in static CVE scan

[CVE info](https://cwe.mitre.org/data/definitions/400.html)

```
== RELEASE NOTES ==

Security Changes
*  Add multi-threaded regex matching function in QueryStateInfoResource to protect from ReDoS attacks in response to `CVE-2024-45296 <https://www.cve.org/CVERecord?id=CVE-2024-45296>`_. :pr:`24048`
 ```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/server/QueryStateInfoResource.java
+++ b/presto-main/src/main/java/com/facebook/presto/server/QueryStateInfoResource.java
@@ -20,6 +20,8 @@ import com.facebook.presto.metadata.InternalNodeManager;
 import com.facebook.presto.resourcemanager.ResourceManagerProxy;
 import com.facebook.presto.spi.QueryId;
 import com.facebook.presto.spi.resourceGroups.ResourceGroupId;
+import com.google.re2j.Pattern;
+import io.airlift.slice.Slices;
 
 import javax.annotation.security.RolesAllowed;
 import javax.inject.Inject;
@@ -45,7 +47,6 @@ import java.util.List;
 import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.OptionalInt;
-import java.util.regex.Pattern;
 
 import static com.facebook.presto.execution.QueryState.QUEUED;
 import static com.facebook.presto.server.QueryStateInfo.createQueryStateInfo;
@@ -107,7 +108,7 @@ public class QueryStateInfoResource
 
             List<QueryStateInfo> queryStateInfos = queryInfos.stream()
                     .filter(queryInfo -> includeAllQueries || !queryInfo.getState().isDone())
-                    .filter(queryInfo -> userPattern.map(pattern -> pattern.matcher(queryInfo.getSession().getUser()).matches()).orElse(true))
+                    .filter(queryInfo -> userPattern.map(pattern -> pattern.matcher(Slices.utf8Slice(queryInfo.getSession().getUser())).matches()).orElse(true))
                     .map(queryInfo -> getQueryStateInfo(
                             queryInfo,
                             includeAllQueryProgressStats,
--- a/presto-main/src/test/java/com/facebook/presto/server/TestQueryStateInfoResource.java
+++ b/presto-main/src/test/java/com/facebook/presto/server/TestQueryStateInfoResource.java
@@ -40,6 +40,7 @@ import static com.facebook.presto.execution.QueryState.RUNNING;
 import static com.facebook.presto.testing.assertions.Assert.assertEquals;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.testng.Assert.assertNotNull;
+import static org.testng.Assert.assertThrows;
 import static org.testng.Assert.assertTrue;
 
 @Test(singleThreaded = true)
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by auden-woolfson** [comment]
Will squash commits once approved

**Comment by BryanCutler** [comment]
Can you add a link to the CVE info?

**Comment by BryanCutler** [comment]
This looks like the related CVE https://www.cve.org/CVERecord?id=CVE-2024-45296

File: presto-main/src/main/java/com/facebook/presto/server/QueryStateInfoResource.java [hunk 1]
```diff
@@ -119,6 +124,23 @@ public void getQueryStateInfos(
         }
     }
 
+    private static boolean matchesWithTimeout(Optional<Pattern> pattern, String input, long timeoutMillis)
```
BryanCutler, line 127: Can the CVE be resolved by doing some kind of sanitation on the regex input instead of running in a thread for a set amount of time?

auden-woolfson: Not entirely. Java already contains some built in checking for evil regex, but there are infinite possible evil regex patterns and they can be very tricky. It is impossible to know if a regex match will hang forever unless actually running the matching algorithm. Cutting off resources at a certain point is the only _100%_ effective way to protect against this kind of vulnerability.

tdcmeehan (changes requested):
Perhaps we can modify the endpoint to instead take in a simplified syntax to match users.  e.g. `*xyz*` instead of `.*xyz.*`.  I don't think it's a good idea to take in arbitrary regex in a RESTful endpoint.

auden-woolfson:
> Perhaps we can modify the endpoint to instead take in a simplified syntax to match users. e.g. `*xyz*` instead of `.*xyz.*`. I don't think it's a good idea to take in arbitrary regex in a RESTful endpoint.

So do you mean something closer to a normal search function than regex?

imsayari404:
Hi @auden-woolfson ,
One minute suggestion , I think  It would be beneficial to make the ``timeoutMillis`` value configurable via a property or environment variable, allowing dynamic adjustment based on deployment needs. Additionally, we should ensure proper resource management by shutting down the ``ExecutorService`` after use, such as calling ``executor.shutdown()`` or ``executor.awaitTermination()`` to prevent potential resource leaks.
will that be okay ? @tdcmeehan @jp-sivaprasad @BryanCutler

steveburnett:
> This looks like the related CVE https://www.cve.org/CVERecord?id=CVE-2024-45296

Assuming that this is the correct CVE to link to, a revised release note entry could be similar to this: 

```
== RELEASE NOTES ==

Security Changes
*  Add multi-threaded regex matching function in QueryStateInfoResource to protect from ReDoS attacks in response to `CVE-2024-45296 <https://www.cve.org/CVERecord?id=CVE-2024-45296>`_. :pr:`24048`
```

jp-sivaprasad (commented):
Please add test case in TestQueryStateInfoResource

auden-woolfson:
> Please add test case in TestQueryStateInfoResource

I had test cases before, I could not find an evil regex that actually caused the timeout to happen. Java has built in measures against evil regexes in the native matching library and I was unable to find any workarounds online. Please let me know if you are able to find a better way to do this... if you would like to see my previous tests you can look at the older commits.

imsayari404 (commented):
Hi @auden-woolfson,
Just a quick suggestion: I think the ``timeoutMillis`` value could be configured via a property or environment variable, so it can be dynamically adjusted based on deployment needs. Additionally, we should ensure proper resource management by shutting down the ``ExecutorService`` after use, such as by calling ``executor.shutdown()`` or ``executor.awaitTermination()``, to avoid potential resource leaks.

@tdcmeehan @jp-sivaprasad @BryanCutler, would this approach work for everyone?

auden-woolfson:
Addressed all comments, will clean commit history once approved

jp-sivaprasad (approved):

auden-woolfson:
Current failures:

```
[INFO] 
[INFO] Results:
[INFO] 
Error:  Failures: 
Error:    TestHiveCommitHandleOutput.testCommitOutputForPartitions:224 expected [1734386810] but found [1734386811]
Error:    TestHiveHistoryBasedStatsTracking.testHistoryBasedStatsCalculatorMultipleStrategies:207->AbstractTestQueryFramework.assertPlan:499->AbstractTestQueryFramework.assertPlan:507->AbstractTestQueryFramework.lambda$assertPlan$7:509 Plan does not match, expected [
```

and...

```
2024-12-16T21:56:33.2377274Z [INFO] 
2024-12-16T21:56:33.2377593Z [INFO] Results:
2024-12-16T21:56:33.2377906Z [INFO] 
2024-12-16T21:56:33.2379911Z [ERROR] Failures: 
2024-12-16T21:56:33.2380737Z [ERROR]   TestEventListener.testOutputStats:314 expected object to not be null
```

All three of these passed in my local environment. Rebasing onto master and pushing again.

@tdcmeehan

tdcmeehan (approved):

PR #24063, opened 2024-11-17
[native] Allow incoming Velox abfs namespace change

majetideepak:
abfs namespace is being removed in Velox for the AbfsFileSystem to be consistent with other File Systems.
Velox PR: https://github.com/facebookincubator/velox/pull/11419

```
== NO RELEASE NOTE ==
```


aditi-pandit (approved):
Thanks @majetideepak

Issue #23754, opened 2024-10-01
ClassCastException in testArrayCumSum: Double cannot be cast to Float


athmaja-n:
## Bug Description
The testArrayCumSum test in the presto-native-tests module fails with a ClassCastException due to type mismatch between Double and Float. The test expects Float values (corresponding to the SQL REAL type), but the array_cum_sum function returns Double values (corresponding to the SQL DOUBLE type) instead, causing the exception.

### Test Case Location:

**File:** presto-native-execution/presto-native-tests/src/test/java/com/facebook/presto/nativetests/AbstractTestQueriesNative.java

**Test Method:** testArrayCumSum

Code Snippet (from AbstractTestQueriesNative.java):

```
sql = "select array_cum_sum(k) from (values (array[cast(5.1 as real), 6.1, 0.5]), (ARRAY[]), (CAST(NULL AS array(real))), (ARRAY [cast(null as real), 6.1, 0.5]), (ARRAY [cast(2.5 as real), 6.1, null, 3.2])) t(k)";
MaterializedResult raw = computeActual(sql);
List<MaterializedRow> rowList = raw.getMaterializedRows();
List<Float> actualFloat = (List<Float>) rowList.get(0).getField(0);  // Fails here due to Double-to-Float cast
```
Error Message:

`java.lang.ClassCastException: class java.lang.Double cannot be cast to class java.lang.Float (java.lang.Double and java.lang.Float are in module java.base of loader 'bootstrap')`

## Expected Behavior
The array_cum_sum function should return cumulative sums for REAL values as Float. The test should pass without encountering any ClassCastException.

## Current Behavior
The array_cum_sum function promotes REAL (floating-point Float in Java) values to DOUBLE during execution, leading to a ClassCastException when the test expects Float results.
Error Message:
`java.lang.ClassCastException: class java.lang.Double cannot be cast to class java.lang.Float
`

## Possible Solution
Ensure that array_cum_sum maintains the type consistency and does not promote REAL values to DOUBLE unless explicitly requested.

zuyu:
@pramodsatya Regarding #23671, I do NOT think velox has a bug here.

Prestissimo result:
```SQL
presto> select array_cum_sum(k) from (values (array[cast(5.1 as real), 6.1, 0.5]), (ARRAY[]), (CAST(NULL AS array(real))), (ARRAY [cast(null as real), 6.1, 0.5]), (ARRAY [cast(2.5 as real), 6.1, null, 3.2])) t(k);
[...]

PR #24065, opened 2024-11-17
[native] Disable Parsing Decimal as Double

zuyu:
Turn off this knob, make it consistent w/ Presto Java since 0.198 to avoid #23754.

https://github.com/prestodb/presto/blob/f72d863cd9cbd17ed1ec2b4bdd3b358998f14449/presto-docs/src/main/sphinx/language/types.rst?plain=1#L107

```
== NO RELEASE NOTE ==
```


yingsu00 (commented):
@zuyu You need to squash the 3 commits together. Otherwise the first commit is going to break CI.

yingsu00 (approved):

PR #24104, opened 2024-11-20
Add input distribution to delete node

feilong-liu:
## Description
Add an optional field to DeleteNode so that DeleteNode can specify the input data distributions for a DeleteNode.

This PR also include changes which moved the DeleteNode to SPI

## Motivation and Context
To enable specify input data properties for DeleteNode. Default is empty which means no requirement.

## Impact
Allow to specify input data properties for DeleteNode.

## Test Plan
Existing tests.

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Add an optional input distribution constraint to DeleteNode :pr:`24104`
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/AddLocalExchanges.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/AddLocalExchanges.java
@@ -22,6 +22,7 @@ import com.facebook.presto.spi.SortingProperty;
 import com.facebook.presto.spi.VariableAllocator;
 import com.facebook.presto.spi.WarningCollector;
 import com.facebook.presto.spi.plan.AggregationNode;
+import com.facebook.presto.spi.plan.DeleteNode;
 import com.facebook.presto.spi.plan.DistinctLimitNode;
 import com.facebook.presto.spi.plan.EquiJoinClause;
 import com.facebook.presto.spi.plan.JoinNode;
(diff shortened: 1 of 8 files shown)


presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/AddLocalExchanges.java
```diff
@@ -435,6 +436,31 @@ public PlanWithProperties visitWindow(WindowNode node, StreamPreferredProperties
...
+        public PlanWithProperties visitDelete(DeleteNode node, StreamPreferredProperties parentPreferences)
+        {
+            if (!node.getInputDistribution().isPresent()) {
+                return visitPlan(node, parentPreferences);
+            }
+            DeleteNode.InputDistribution inputDistribution = node.getInputDistribution().get();
+            StreamPreferredProperties childRequirements = parentPreferences
+                    .constrainTo(node.getSource().getOutputVariables())
+                    .withDefaultParallelism(session)
+                    .withPartitioning(inputDistribution.getPartitionBy());
```
rschlussel, line 449: do we need to worry about order sensitivity here?

feilong-liu: I think not. The AddLocalExchanges is to add local exchanges and we do not expect it add sort node to enforce the input order (this will be in a separate optimizer rule, let's say it's DeleteSortRule).
Order property can be used in two ways here:
* if the input is already partially sorted or even totally sorted, we can record it in the DeleteNode so as to instruct the DeleteSortRule to simplify the sort process, if this DeleteSortRule runs after AddLocalExchanges. For this part we need to add more attribute to DeleteNode and I consider it as a future optimization to avoid introducing more complex changes before we see more usage of this new input distribution field
* assert that the input is sorted here if the DeleteSortRule is before AddLocalExchanges. Again I plan to do this assertion in DeleteNode during runtime (and we need it anyway for robustness even if we assert the data is sorted in the plan here, since we may introduce other changes later which break this property) and avoid introducing more complex changes.

PR #24108, opened 2024-11-21
Add enhanced cte scheduling mode

jaystarshot:
## Description

Fixes #22205 

Enhance the scheduler to selectively block only the TableScan stages that depend on incomplete CTE TableWriter stages, rather than blocking all dependent sections.

Previously, all sections relying on a CTE writer were blocked until the writer completed. 
With this change, only the specific TableScans referencing the CTE are delayed, allowing other stages to proceed. This optimization can significantly improve query latency.

For example for the materialized CTE T:

```
WITH T AS (SELECT * FROM tpch.orders) 
SELECT * 
FROM T 
JOIN (SELECT * FROM customer) b 
ON t.uuid = b.uuid;

```
In this query, the right side of the join can be executed and kept in memory concurrently while the CTE write operation completes.

2 commits
1. Add the CTE used info to tablescans and table finish 
2. Use this info in scheduling by adding a central manager per scheduler and then using this to unblock/wake up tasks and to maintain completed ctes.

Can cause more resource utilization in some cases where intermediate results get blocked due to written ctes  but that may also happen without any materialization

## Test Plan
Existing UTs + prod queries

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Improve scheduling for CTE materialization: Now, only the stages containing CTE table scans that reference CTE table write stages are blocked till the write is complete, instead of the entire query being blocked as was the case previously. This is controlled by the session property ``enhanced_cte_scheduling_enabled`` (on by default) :pr:`24108`

```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/execution/SqlStageExecution.java
+++ b/presto-main/src/main/java/com/facebook/presto/execution/SqlStageExecution.java
@@ -26,10 +26,15 @@ import com.facebook.presto.metadata.RemoteTransactionHandle;
 import com.facebook.presto.metadata.Split;
 import com.facebook.presto.server.remotetask.HttpRemoteTask;
 import com.facebook.presto.spi.PrestoException;
+import com.facebook.presto.spi.plan.CteMaterializationInfo;
 import com.facebook.presto.spi.plan.PlanFragmentId;
+import com.facebook.presto.spi.plan.PlanNode;
 import com.facebook.presto.spi.plan.PlanNodeId;
+import com.facebook.presto.spi.plan.TableFinishNode;
+import com.facebook.presto.spi.plan.TableScanNode;
 import com.facebook.presto.split.RemoteSplit;
 import com.facebook.presto.sql.planner.PlanFragment;
+import com.facebook.presto.sql.planner.optimizations.PlanNodeSearcher;
 import com.facebook.presto.sql.planner.plan.RemoteSourceNode;
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.ImmutableMap;
@@ -60,8 +65,10 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
+import java.util.stream.Collectors;
 
 import static com.facebook.presto.SystemSessionProperties.getMaxFailedTaskPercentage;
+import static com.facebook.presto.SystemSessionProperties.isEnhancedCTESchedulingEnabled;
 import static com.facebook.presto.failureDetector.FailureDetector.State.GONE;
 import static com.facebook.presto.operator.ExchangeOperator.REMOTE_CONNECTOR_ID;
 import static com.facebook.presto.spi.StandardErrorCode.GENERIC_INTERNAL_ERROR;
@@ -557,7 +564,6 @@ public final class SqlStageExecution
             // stage finished while we were scheduling this task
             task.abort();
         }
-
         return task;
     }
 
@@ -592,6 +598,59 @@ public final class SqlStageExecution
         // Fetch the results from the buffer assigned to the task based on id
         String splitLocation = remoteSourceTaskLocation.toASCIIString() + "/results/" + taskId.getId();
         return new Split(REMOTE_CONNECTOR_ID, new RemoteTransactionHandle(), new RemoteSplit(new Location(splitLocation), remoteSourceTaskId));
+    }
+
+    private static String getCteIdFromSource(PlanNode source)
+    {
+        // Traverse the plan node tree to find a TableWriterNode with TemporaryTableInfo
+        return PlanNodeSearcher.searchFrom(source)
+                .where(planNode -> planNode instanceof TableFinishNode)
+                .findFirst()
+                .flatMap(planNode -> ((TableFinishNode) planNode).getCteMaterializationInfo())
+                .map(CteMaterializationInfo::getCteId)
+                .orElseThrow(() -> new IllegalStateException("TemporaryTableInfo has no CTE ID"));
+    }
+
+    public boolean isCTETableFinishStage()
+    {
+        return PlanNodeSearcher.searchFrom(planFragment.getRoot())
+                .where(planNode -> planNode instanceof TableFinishNode &&
+                        ((TableFinishNode) planNode).getCteMaterializationInfo().isPresent())
+                .findSingle()
+                .isPresent();
+    }
+
+    public String getCTEWriterId()
+    {
+        // Validate that this is a CTE TableFinish stage and return the associated CTE ID
+        if (!isCTETableFinishStage()) {
+            throw new IllegalStateException("This stage is not a CTE writer stage");
+        }
+        return getCteIdFromSource(planFragment.getRoot());
+    }
+
+    public boolean requiresMaterializedCTE()
+    {
+        if (!isEnhancedCTESchedulingEnabled(session)) {
+            return false;
+        }
+        // Search for TableScanNodes and check if they reference TemporaryTableInfo
+        return PlanNodeSearcher.searchFrom(planFragment.getRoot())
+                .where(planNode -> planNode instanceof TableScanNode)
+                .findAll().stream()
+                .anyMatch(planNode -> ((TableScanNode) planNode).getCteMaterializationInfo().isPresent());
+    }
+
+    public List<String> getRequiredCTEList()
+    {
+        // Collect all CTE IDs referenced by TableScanNodes with TemporaryTableInfo
+        return PlanNodeSearcher.searchFrom(planFragment.getRoot())
+                .where(planNode -> planNode instanceof TableScanNode)
+                .findAll().stream()
+                .map(planNode -> ((TableScanNode) planNode).getCteMaterializationInfo()
+                        .orElseThrow(() -> new IllegalStateException("TableScanNode has no TemporaryTableInfo")))
+                .map(CteMaterializationInfo::getCteId)
+                .collect(Collectors.toList());
     }
 
     private void updateTaskStatus(TaskId taskId, TaskStatus taskStatus)
--- a/presto-main/src/main/java/com/facebook/presto/execution/scheduler/CTEMaterializationTracker.java
+++ b/presto-main/src/main/java/com/facebook/presto/execution/scheduler/CTEMaterializationTracker.java
@@ -1,0 +1,63 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.execution.scheduler;
+
+import com.google.common.base.Preconditions;
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.SettableFuture;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+/*
+ * Tracks the completion status of table-finish nodes that write temporary tables for CTE materialization.
+ * CTEMaterializationTracker manages a map of materialized CTEs and their associated materialization futures.
+ * When a stage includes a CTE table finish, it marks the corresponding CTE as materialized and completes
+ * the associated future.
+ * This signals the scheduler that some dependency has been resolved, prompting it to resume/continue scheduling.
+ */
+public class CTEMaterializationTracker
+{
+    private final Map<String, SettableFuture<Void>> materializationFutures = new ConcurrentHashMap<>();
+
+    public ListenableFuture<Void> getFutureForCTE(String cteName)
+    {
+        return Futures.nonCancellationPropagating(
+                materializationFutures.compute(cteName, (key, existingFuture) -> {
+                    if (existingFuture == null) {
+                        // Create a new SettableFuture and store it internally
+                        return SettableFuture.create();
+                    }
+                    Preconditions.checkArgument(!existingFuture.isCancelled(),
+                            String.format("Error: Existing future was found cancelled in CTEMaterializationTracker for cte", cteName));
+                    return existingFuture;
+                }));
+    }
+
+    public void markCTEAsMaterialized(String cteName)
+    {
+        materializationFutures.compute(cteName, (key, existingFuture) -> {
+            if (existingFuture == null) {
+                SettableFuture<Void> completedFuture = SettableFuture.create();
+                completedFuture.set(null);
+                return completedFuture;
+            }
+            Preconditions.checkArgument(!existingFuture.isCancelled(),
+                    String.format("Error: Existing future was found cancelled in CTEMaterializationTracker for cte", cteName));
+            existingFuture.set(null); // Notify all listeners
+            return existingFuture;
+        });
+    }
+}
--- a/presto-main/src/main/java/com/facebook/presto/execution/scheduler/FixedSourcePartitionedScheduler.java
+++ b/presto-main/src/main/java/com/facebook/presto/execution/scheduler/FixedSourcePartitionedScheduler.java
@@ -74,6 +74,8 @@ public class FixedSourcePartitionedScheduler
 
     private final Queue<Integer> tasksToRecover = new ConcurrentLinkedQueue<>();
 
+    private final CTEMaterializationTracker cteMaterializationTracker;
+
     @GuardedBy("this")
     private boolean closed;
 
@@ -87,13 +89,15 @@ public class FixedSourcePartitionedScheduler
             int splitBatchSize,
             OptionalInt concurrentLifespansPerTask,
             NodeSelector nodeSelector,
-            List<ConnectorPartitionHandle> partitionHandles)
+            List<ConnectorPartitionHandle> partitionHandles,
+            CTEMaterializationTracker cteMaterializationTracker)
     {
         requireNonNull(stage, "stage is null");
         requireNonNull(splitSources, "splitSources is null");
         requireNonNull(bucketNodeMap, "bucketNodeMap is null");
         checkArgument(!requireNonNull(nodes, "nodes is null").isEmpty(), "nodes is empty");
         requireNonNull(partitionHandles, "partitionHandles is null");
+        this.cteMaterializationTracker = cteMaterializationTracker;
 
         this.stage = stage;
         this.nodes = ImmutableList.copyOf(nodes);
@@ -179,6 +183,29 @@ public class FixedSourcePartitionedScheduler
     {
         // schedule a task on every node in the distribution
         List<RemoteTask> newTasks = ImmutableList.of();
+
+        // CTE Materialization Check
+        if (stage.requiresMaterializedCTE()) {
+            List<ListenableFuture<?>> blocked = new ArrayList<>();
+            List<String> requiredCTEIds = stage.getRequiredCTEList();
+            for (String cteId : requiredCTEIds) {
+                ListenableFuture<Void> cteFuture = cteMaterializationTracker.getFutureForCTE(cteId);
+                if (!cteFuture.isDone()) {
+                    // Add CTE materialization future to the blocked list
+                    blocked.add(cteFuture);
+                }
+            }
+            // If any CTE is not materialized, return a blocked ScheduleResult
+            if (!blocked.isEmpty()) {
+                return ScheduleResult.blocked(
+                        false,
+                        newTasks,
+                        whenAnyComplete(blocked),
+                        BlockedReason.WAITING_FOR_CTE_MATERIALIZATION,
+                        0);
+            }
+        }
+        // schedule a task on every node in the distribution
         if (!scheduledTasks) {
             newTasks = Streams.mapWithIndex(
                     nodes.stream(),
@@ -191,9 +218,8 @@ public class FixedSourcePartitionedScheduler
             // notify listeners that we have scheduled all tasks so they can set no more buffers or exchange splits
             stage.transitionToFinishedTaskScheduling();
         }
-
-        boolean allBlocked = true;
         List<ListenableFuture<?>> blocked = new ArrayList<>();
+        boolean allBlocked = true;
         BlockedReason blockedReason = BlockedReason.NO_ACTIVE_DRIVER_GROUP;
 
         if (groupedLifespanScheduler.isPresent()) {
--- a/presto-main/src/main/java/com/facebook/presto/execution/scheduler/ScheduleResult.java
+++ b/presto-main/src/main/java/com/facebook/presto/execution/scheduler/ScheduleResult.java
@@ -57,6 +57,11 @@ public class ScheduleResult
          * grouped execution where there are multiple lifespans per task).
          */
         MIXED_SPLIT_QUEUES_FULL_AND_WAITING_FOR_SOURCE,
+
+        /**
+         * Waiting for the completion of CTE materialization by the table writer.
+         */
+        WAITING_FOR_CTE_MATERIALIZATION,
         /**/;
 
         public BlockedReason combineWith(BlockedReason other)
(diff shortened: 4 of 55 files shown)


presto-main/src/main/java/com/facebook/presto/execution/scheduler/FixedSourcePartitionedScheduler.java
```diff
@@ -179,6 +183,29 @@ public ScheduleResult schedule()
     {
         // schedule a task on every node in the distribution
         List<RemoteTask> newTasks = ImmutableList.of();
+        List<ListenableFuture<?>> blocked = new ArrayList<>();
+
+        // CTE Materialization Check
+        if (stage.requiresMaterializedCTE()) {
```
jaystarshot, line 189: main change

presto-main/src/main/java/com/facebook/presto/execution/scheduler/SqlQueryScheduler.java
```diff
@@ -278,6 +280,17 @@ else if (state == CANCELED) {
 
         for (StageExecutionAndScheduler stageExecutionInfo : stageExecutions.values()) {
             SqlStageExecution stageExecution = stageExecutionInfo.getStageExecution();
+            // Add a listener for state changes
```
jaystarshot, line 283: Add a listener to stages with tablefinish which updates our tracker

presto-main/src/main/java/com/facebook/presto/execution/SqlStageExecution.java
```diff
@@ -594,6 +600,59 @@ private static Split createRemoteSplitFor(TaskId taskId, URI remoteSourceTaskLoc
...
     }
 
+    private String getCteIdFromSource(PlanNode source)
+    {
+        // Traverse the plan node tree to find a TableWriterNode with TemporaryTableInfo
+        return PlanNodeSearcher.searchFrom(source)
+                .where(planNode -> planNode instanceof TableFinishNode)
+                .findFirst()
+                .map(planNode -> ((TableFinishNode) planNode).getTemporaryTableInfo().orElseThrow(
+                        () -> new IllegalStateException("TableFinishNode has no TemporaryTableInfo")))
```
ZacBlanco, line 610:
```suggestion
                .flatMap(planNode -> ((TableFinishNode) planNode).getTemporaryTableInfo())
```

presto-main/src/main/java/com/facebook/presto/execution/SqlStageExecution.java
```diff
@@ -594,6 +600,59 @@ private static Split createRemoteSplitFor(TaskId taskId, URI remoteSourceTaskLoc
...
+                .map(TemporaryTableInfo::getCteId)
+                .orElseThrow(() -> new IllegalStateException("TemporaryTableInfo has no CTE ID"));
+    }
+
+    public boolean isCTETableFinishStage()
+    {
+        return PlanNodeSearcher.searchFrom(planFragment.getRoot())
+                .where(planNode -> (planNode instanceof TableFinishNode))
+                .findAll().stream()
```
ZacBlanco, line 619: do you think we should verify there is only one `TableFinishNode` using something like `getOnlyElement`?

presto-main/src/main/java/com/facebook/presto/execution/scheduler/FixedSourcePartitionedScheduler.java
```diff
@@ -179,6 +183,29 @@ public ScheduleResult schedule()
...
+                    ListenableFuture<Void> materializationFuture = cteMaterializationTracker.getFutureForCTE(cteId);
+                    blocked.add(materializationFuture);
+                }
+            }
+            // If any CTE is not materialized, return a blocked ScheduleResult
+            if (!blocked.isEmpty()) {
+                return ScheduleResult.blocked(
+                        false,
+                        newTasks,
+                        whenAnyComplete(blocked),
```
ZacBlanco, line 203: Trying to wrap my head around the logic here:

> If any CTE is not materialized, return a blocked ScheduleResult

So the logic should be, in order to return a result which isn't blocked, then the `blocked` list should be empty. This makes sense based on the `blocked.isEmpty()` condition.

However, the `blocked` future argument is `whenAnyComplete(blocked)`. Shouldn't this be something like `whenAllComplete`? Or is there some nuance to what the `blocked` argument should be to the `ScheduleResult.blocked` function? I am not very familiar with this portion of the codebase.

presto-main/src/main/java/com/facebook/presto/execution/scheduler/SqlQueryScheduler.java
```diff
@@ -258,7 +259,8 @@ private SqlQueryScheduler(
         stageExecutions.stream()
                 .forEach(execution -> this.stageExecutions.put(execution.getStageExecution().getStageExecutionId().getStageId(), execution));
 
-        this.maxConcurrentMaterializations = getMaxConcurrentMaterializations(session);
+//        this.maxConcurrentMaterializations = getMaxConcurrentMaterializations(session);
+        this.maxConcurrentMaterializations = 100;
```
ZacBlanco, line 263: remove and uncomment?

presto-main/src/main/java/com/facebook/presto/execution/scheduler/FixedSourcePartitionedScheduler.java
jaystarshot: I think any or all don't matter since blocked list just contains one future here

presto-main/src/main/java/com/facebook/presto/execution/SqlStageExecution.java
rschlussel: you can use findSingle() for this. Also, instead of streaming again and then anyMatch, you can move all of that into the where condition above.

presto-main/src/main/java/com/facebook/presto/execution/scheduler/ScheduleResult.java
```diff
@@ -57,13 +57,16 @@ public enum BlockedReason
          * grouped execution where there are multiple lifespans per task).
          */
         MIXED_SPLIT_QUEUES_FULL_AND_WAITING_FOR_SOURCE,
+
+        WAITING_FOR_CTE_MATERIALIZATION,
```
rschlussel, line 61: add a comment explaining what this means

presto-spi/src/main/java/com/facebook/presto/spi/plan/TemporaryTableInfo.java
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.spi.plan;
+
+public class TemporaryTableInfo
```
rschlussel, line 16: add some comments explaining what this class is for

rschlussel (commented):
can you add some tests for this?

jaystarshot:
Thanks for the review. Actually we have a lot of good cte materialization tests which are now using this new flow [here](https://github.com/prestodb/presto/blob/e416ff11482b92f812c2c47ef5ea0f0b19d85393/presto-hive/src/test/java/com/facebook/presto/hive/TestCteExecution.java#L50) (over 50) so I didn’t add a new one.

hantangwangd (commented):
Thanks for the work, should we also supplement the document for configuration property `enhanced-cte-blocking`?

presto-main/src/main/java/com/facebook/presto/execution/scheduler/FixedSourcePartitionedScheduler.java
jaystarshot: so FixedSourcePartitionedScheduler is just for the stages reading from bucketed tables. As far as I know presto doesn't allow multiple tablescans in one stage. But this code is still resilient

presto-main/src/main/java/com/facebook/presto/execution/SqlStageExecution.java
jaystarshot: Better to throw exp

presto-main/src/main/java/com/facebook/presto/execution/scheduler/CTEMaterializationTracker.java
```diff
@@ -0,0 +1,59 @@
...
+    public ListenableFuture<Void> getFutureForCTE(String cteName)
+    {
+        if (materializationFutures.containsKey(cteName)) {
+            if (!materializationFutures.get(cteName).isCancelled()) {
+                return materializationFutures.get(cteName);
+            }
+        }
+        materializationFutures.put(cteName, SettableFuture.create());
+        return materializationFutures.get(cteName);
+    }
```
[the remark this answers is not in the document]
jaystarshot: Yeah maybe its a good point to make it atomic, I added this in the compute() method of concurrenthashmap which is atomic

steveburnett (approved):
LGTM! (docs)

kaikalur:
> In this query, the right side of the join can be executed and kept in memory concurrently while the CTE write operation completes.

Won't that increase memory for the query?

jaystarshot:
Yes unfortunately no way to avoid that till we read concurrently while we write maybe with a new exchange

presto-main/src/main/java/com/facebook/presto/execution/scheduler/FixedSourcePartitionedScheduler.java
hantangwangd: Oh yes, for hive connector, it's the bucketed tables that will expose their partitioning property into `HivePartitioningHandle` when considering colocated joins. So when left and right tables in a join have compatible bucket definition exactly on the join keys, they will be left into the same stage.

Currently, the tableScanNode for the temporary table caused by a CTE does not inherit the bucket feature of its source table, so seems that it won't stay in the same stage with other tableScanNodes. Do we have any plan to support inheriting the bucket feature of the source table in future?

jaystarshot:
@rschlussel Can you please take a look

presto-main/src/main/java/com/facebook/presto/execution/scheduler/CTEMaterializationTracker.java
```diff
@@ -0,0 +1,58 @@
...
+    {
+        if (materializedCtes.putIfAbsent(cteName, true) == null) {
+            SettableFuture<Void> future = materializationFutures.get(cteName);
+            if (future != null && !future.isCancelled()) {
+                future.set(null); // Notify all listeners
+            }
+        }
+    }
+
+    public synchronized boolean hasBeenMaterialized(String cteName)
```
rschlussel, line 54: why is this synchronized? don't think it should be necessary with the ConcurrentHashMap, but if it is necessary, you'd need to synchronize the writes too.

presto-main/src/main/java/com/facebook/presto/execution/scheduler/CTEMaterializationTracker.java
```diff
@@ -0,0 +1,58 @@
...
+public class CTEMaterializationTracker
+{
+    private final Map<String, SettableFuture<Void>> materializationFutures = new ConcurrentHashMap<>();
+
+    private final Map<String, Boolean> materializedCtes = new ConcurrentHashMap<>();
+
+    public ListenableFuture<Void> getFutureForCTE(String cteName)
+    {
+        return materializationFutures.compute(cteName, (key, existingFuture) -> {
+            if (existingFuture == null || existingFuture.isCancelled()) {
```
rschlussel, line 37: why do we change isCancelled? Seems like that could be a race condition anyway (gets cancelled by another thread after we check the condition).

presto-main/src/main/java/com/facebook/presto/execution/scheduler/CTEMaterializationTracker.java
```diff
@@ -0,0 +1,58 @@
...
+            }
+            return existingFuture;
+        });
+    }
+
+    public void markCTEAsMaterialized(String cteName)
+    {
+        if (materializedCtes.putIfAbsent(cteName, true) == null) {
+            SettableFuture<Void> future = materializationFutures.get(cteName);
+            if (future != null && !future.isCancelled()) {
```
rschlussel, line 48: same question about checking isCancelled().

presto-main/src/main/java/com/facebook/presto/execution/scheduler/CTEMaterializationTracker.java
```diff
@@ -0,0 +1,58 @@
...
+
+import javax.annotation.concurrent.ThreadSafe;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+/*
+ *    Tracks whether tablefinish nodes writing temporary tables for CTE Materialization are complete
+ */
+@ThreadSafe
```
rschlussel, line 27: I think in Presto we expect that everything is thread safe unless it's specifically noted as not thread safe.

presto-main/src/main/java/com/facebook/presto/execution/scheduler/FixedSourcePartitionedScheduler.java
```diff
@@ -179,6 +183,29 @@ public ScheduleResult schedule()
     {
         // schedule a task on every node in the distribution
         List<RemoteTask> newTasks = ImmutableList.of();
+        List<ListenableFuture<?>> blocked = new ArrayList<>();
```
rschlussel, line 186: move this inside the if(stage.requiresMaterializedCTE()) condition.

presto-main/src/main/java/com/facebook/presto/execution/scheduler/FixedSourcePartitionedScheduler.java
rschlussel: I'm still a bit confused here. A stage can have multiple CTE dependencies if they have the same bucketing, so seems like we shouldn wait for all blocked futures to complete, right? (this isn't broken now, as if we unblock when some complete, we'll just do this check again, just want to make sure i'm understanding the flow properly)

presto-main/src/main/java/com/facebook/presto/execution/scheduler/CTEMaterializationTracker.java
jaystarshot: yes correct this is not needed

presto-main/src/main/java/com/facebook/presto/execution/scheduler/CTEMaterializationTracker.java
jaystarshot: As we know, the Scheduler’s [schedule()](https://github.com/prestodb/presto/commit/baf737b12240068a54aef16a2990ee624821519e) method is single-threaded. When the CTE writer stage (or another blocked stage) becomes unblocked [here](https://github.com/prestodb/presto/blob/85cc36af2cca5be4b15d17884f3f6a01bd107781/presto-main/src/main/java/com/facebook/presto/execution/scheduler/SqlQueryScheduler.java#L514), the schedule() method resumes and cancels all pending futures [here](https://github.com/prestodb/presto/blob/85cc36af2cca5be4b15d17884f3f6a01bd107781/presto-main/src/main/java/com/facebook/presto/execution/scheduler/SqlQueryScheduler.java#L517). Then it will try to schedle all blocked stages again and we will call FixedSourcePartitionedScheduler’s schedule() method, at which point a cancelled future will appear in the CTEMaterializationMap.

I think you are correct when you say that a race condition can appear in some extremely unlikely case, will need to take a deeper dive on the fix here because it doesn't seem trivial. We could check and not cancel the cte futures in the schedule() method but then it would need the info passed to the future

jaystarshot: I think I can add make all methods of CTEMaterializationTracker synchronized and remove all futures from the materializationFutures maps when the shedule() method resumes and cancels all pending futures [here](https://github.com/prestodb/presto/blob/85cc36af2cca5be4b15d17884f3f6a01bd107781/presto-main/src/main/java/com/facebook/presto/execution/scheduler/SqlQueryScheduler.java#L517). Then it should be threadsafe.

rschlussel: can we ignore the fact that it's been canceled?

jaystarshot: Since our map in CTEMaterializationTracker is storing the cancelled future, the [FixedSourceCountScheduler](https://github.com/prestodb/presto/blob/85cc36af2cca5be4b15d17884f3f6a01bd107781/presto-main/src/main/java/com/facebook/presto/execution/scheduler/FixedSourcePartitionedScheduler.java#L178) schedule method will return this cancelled future (which was cancelled by a previous loop of the scheduling process). I think that we need to make sure that the  FixedSourcePartitionedScheduler returns a fresh non cancelled future

jaystarshot: https://github.com/prestodb/presto/pull/24108/files#diff-bbcd68b946ac8590a0a1d1371f51e32687b0061d914689eb6fb3b852b69d3ac1R534 (main change clearing all futues)

presto-main/src/main/java/com/facebook/presto/execution/scheduler/CTEMaterializationTracker.java
```diff
@@ -0,0 +1,68 @@
...
+
+    private final Map<String, Boolean> materializedCtes = new HashMap<>();
+
+    public synchronized ListenableFuture<Void> getFutureForCTE(String cteName)
+    {
+        return materializationFutures.compute(cteName, (key, existingFuture) -> {
+            if (existingFuture == null) {
+                return SettableFuture.create();
+            }
+            checkArgument(!existingFuture.isCancelled(), "CTE future was found in a cancelled state");
```
rschlussel, line 43: this is still not thread safe, as the cancellation isn't atomic with clearing the futures, so it's possible to hit this condition (future cancelled, something tries to get the future before clearAllFutures is called).  I think it would be better to have the code consuming the futures gracefully handle receiving a cancelled future and not have to worry about synchronization.

jaystarshot: Part 1 `(future canceled, something tries to get the future before clearAllFutures is called)` can be fixed by clearing all before canceling. 

For part 2 `It's also possible for the future to be canceled after it's returned, but before the consumer checks it or even before it's returned, but after this argument check` 
I think this will never happen as blocked.cancel() is invoked in SqlQueryScheduler only after all stages are scheduled. At that point, all futures have already been returned to the SqlQueryScheduler, so the consumer will have completed scheduling (and return the blocked result) before this happens.


> I think it would be better to have the code consuming the futures gracefully handle receiving a canceled future and not have to worry about synchronization.

I think the code consuming the future can be other Schedulers and making them handle canceled futures might be challenging (not 100% sure if that is even needed yet)

rschlussel: that's how it works today, but it's not really thread safe, as there's no guarantee that some other thread holding the future won't cancel it or that someone won't reorder the calling code without realizing the dependency and break the assumptions that clearAllFutures is always called before cancelling any future in the map. At the very least this needs some very clear code comments about the assumptions this class is making and requirements for anything using it.

presto-main/src/main/java/com/facebook/presto/execution/scheduler/CTEMaterializationTracker.java
```diff
@@ -0,0 +1,70 @@
...
+ * the associated future. This signals the scheduler that some dependency has been resolved, prompting it
+ * to resume scheduling. The scheduler can call the clearAllFutures() method once it starts scheduling again
+ */
+public class CTEMaterializationTracker
+{
+    private final Map<String, SettableFuture<Void>> materializationFutures = new HashMap<>();
+
+    private final Map<String, Boolean> materializedCtes = new HashMap<>();
+
+    public synchronized ListenableFuture<Void> getFutureForCTE(String cteName)
```
rschlussel, line 35: it would be nice to avoid all this locking. it seems like right now it's only needed to make the materializationFutures and materializedCTE calls atomic (because aside from that we still make assumptions about when cancel is or is not called, so it's not providing any thread safety with regard to the cancellation calls).

  I'm thinking it would be better to use a single ConcurrentHashMap<>, and then not clear the futures before cancellation and instead check in getFutureForCTE() that it hasn't been cancelled, and if it has create a new one (like the code you had before).  something like the following.

```
private final Map<String, SettableFuture<Void>> materializationFutures = new ConcurrentHashMap<>();

public ListenableFuture<Void> getFutureForCTE(String cteName)
    {
         return materializationFutures.compute(cteName, (key, existingFuture) -> {
            if (existingFuture == null || existingFuture.iscancelled()) {
                return SettableFuture.create();
            }
            return existingFuture;
        });

   public void markCTEAsMaterialized(String cteName)
    {
         materializationFutures.compute(cteName, (key, existingFuture) -> {
            if (existingFuture == null || existingFuture.iscancelled()) {
                SettableFuture completedFuture = SettableFuture.create();
                completedFuture.set(null);
                return completedFuture;
            }
            existingFuture.set(null); // notify all listeners
            return existingFuture;
        });


```

jaystarshot: Not happy with the assumptions that we are making, will brainstorm for a better way but it won't be trivial

jaystarshot:
```
public void markCTEAsMaterialized(String cteName)
    {
        materializationFutures.compute(cteName, (key, existingFuture) -> {
            if (existingFuture == null || existingFuture.isCancelled()) {
                SettableFuture completedFuture = SettableFuture.create();
                completedFuture.set(null);
                return completedFuture;
            }
            existingFuture.set(null); // notify all listeners
            return existingFuture;
        });
    }
```
Also here completing the future when its cancelled
```
  if (existingFuture == null || existingFuture.isCancelled()) {
                SettableFuture completedFuture = SettableFuture.create();
                completedFuture.set(null);
                return completedFuture;
            }
```
is not really useful since there will be no listeners on the new future. the entire goal of the markCTEAsMaterialized() is to unblock the stages which were waiting on this future

rschlussel: the point of setting a new completed future if the original one was canceled is in order to add it to the map. that way when someone calls getFutureForCTE() they will get a future that is done.

jaystarshot: Got it, first race condition is fixed.  For the second condition

> For part 2 It's also possible for the future to be canceled after it's returned, but before the consumer checks it or even before it's returned, but after this argument check

If we see the FixedSourcePartitionScheduler code 
```
 // If any CTE is not materialized, return a blocked ScheduleResult
            if (!blocked.isEmpty()) {
                return ScheduleResult.blocked(
                        false,
                        newTasks,
                        whenAnyComplete(blocked),
                        BlockedReason.WAITING_FOR_CTE_MATERIALIZATION,
                        0);
            }
```
Actually I don't think we are even cancelling the future in the CteMaterializationTracker. We are cancelling the future  which was wrapped by the `whenAnyComplete` so that problem seems non exisitent

jaystarshot: Okay I think I found a good fix. If we check the [FixedSourcePatitionScheduler](https://github.com/prestodb/presto/pull/24108/files#diff-1090adf62744bbb6103f894606a0574e2d8b040779e2279e9846a820c2d6a118R200) code, we can see that it actually returns a wrapped future in whenAnyComplete. This wrapped future which can be cancelled. I also debugged and verified that if this is cancelled the inner ones are also cancelled (maybe because we cancelled with mayInterruptIfRunning?) 
However if we make sure that the ctematerialization tracker always returns a `Futures.nonCancellationPropagating` then the future in the tracker can never be cancelled so in a way we have handled cancellations. 

```
if (!blocked.isEmpty()) {
                return ScheduleResult.blocked(
                        false,
                        newTasks,
                        whenAnyComplete(blocked),
                        BlockedReason.WAITING_FOR_CTE_MATERIALIZATION,
                        0);
            }
```

jaystarshot:
@rschlussel can you please take a look again? Thanks

rschlussel (dismissed):

jaystarshot:
Rebased to resolve conflicts and added part of https://github.com/prestodb/presto/pull/21626 which was inadvertently removed. Detail [changelog](https://github.com/prestodb/presto/compare/563330c55f67791737dba82e010a2a712a05bf43..2a41b8349c4fa5792eed1a0f01ebacc38aaa3) since review

rschlussel (approved):

PR #24000, opened 2024-11-11
Removing class objects comparison using getName(),getSimpleName()

bibith4:
## Description
Changes to remove class objects comparison using the methods getName(),getSimpleName() to improve security.
Made changes in
presto-hive/src/main/java/com/facebook/presto/hive/HiveUtil.java
presto-main/src/main/java/com/facebook/presto/execution/QueryStats.java
presto-main/src/main/java/com/facebook/presto/execution/executor/TaskExecutor.java
presto-main/src/main/java/com/facebook/presto/operator/AggregationOperator.java
presto-main/src/main/java/com/facebook/presto/operator/HashAggregationOperator.java
presto-main/src/main/java/com/facebook/presto/operator/HashBuilderOperator.java
presto-main/src/main/java/com/facebook/presto/operator/TableWriterOperator.java
presto-main/src/main/java/com/facebook/presto/operator/TaskContext.java
presto-main/src/main/java/com/facebook/presto/operator/scalar/JoniRegexpFunctions.java

## Motivation and Context
Do not compare class objects using the getName() or getSimpleName() method which may causes multiple security vulnerabilities includes String manipulations,class spoofing and namespace collisions

## Impact
No impact

## Test Plan
Verified existing Unit test case for above fix
presto-hive/src/test/java/com/facebook/presto/hive/TestHiveUtil.java
presto-main/src/test/java/com/facebook/presto/execution/TestQueryStats.java
presto-main/src/test/java/com/facebook/presto/execution/executor/TestTaskExecutor.java
presto-main/src/main/java/com/facebook/presto/testing/TestingTaskContext.java

## Release Notes

```
== NO RELEASE NOTE ==
```

PR #24109, opened 2024-11-21
Removing class objects comparison using the getName(),getSimpleName()

bibith4:
## Description
Changes to remove class objects comparison using the methods getName(),getSimpleName() to improve security.
Follow up PR of https://github.com/prestodb/presto/pull/24000
Made changes in
presto-hive/src/main/java/com/facebook/presto/hive/orc/OrcPageSourceFactoryUtils.java

## Motivation and Context
Do not compare class objects using the getName() or getSimpleName() method which may causes multiple security vulnerabilities includes String manipulations,class spoofing and namespace collisions

## Impact
No impact

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-hive/src/main/java/com/facebook/presto/hive/orc/OrcPageSourceFactoryUtils.java
+++ b/presto-hive/src/main/java/com/facebook/presto/hive/orc/OrcPageSourceFactoryUtils.java
@@ -37,6 +37,7 @@ import io.airlift.units.DataSize;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FSDataInputStream;
 import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.hdfs.BlockMissingException;
 
 import java.io.FileNotFoundException;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


presto-hive/src/main/java/com/facebook/presto/hive/orc/OrcPageSourceFactoryUtils.java
```diff
@@ -111,14 +112,14 @@ public static OrcReader getOrcReader(OrcEncoding orcEncoding, List<HiveColumnHan
     public static PrestoException mapToPrestoException(Exception e, Path path, HiveFileSplit fileSplit)
     {
         // instanceof and class comparison do not work here since they are loaded by different class loaders.
```
tdcmeehan, line 114: Isn't this comment still relevant

bibith4: @tdcmeehan Shall I replace all occurrences of "instanceof" with "isAssignableFrom" in the mapToPrestoException method?

tdcmeehan: I'm just thinking that the problem is still present, because you're comparing potentially the same class but with two different classloaders.

sdruzkin: 1. isAssignableFrom won't work for classes loaded with different classloaders. 
2. I don't see any security threats posed by this class name comparison logic. I'd discard this PR.

bibith4: @tdcmeehan @sdruzkin Ok. Then i will mark these issues as false positive and close this PR.

Issue #23152, opened 2024-07-09
Implement Jdbc join pushdown capabilities in presto


Ajas-Mangal:
## Expected Behavior or Use Case

When presto identifies a join query which is specific to a Jdbc remote datasource, it split the join query into multiple select query based on the number of tables involved in the join query and select all records from each tables using the jdbc connector without passing the join condition. 

If we  "Push down" or send these joins which involves same catalog/remote datasource as part of the  SQL  to the remote data source it increase the performance 3x to 10x.

## Presto Component, Service, or Connector

Jdbc Connector and spi module which create presto PlanNode

## Possible Implementation

Begin to look at Presto analyze component that breaks down a multi join query into multiple subqueries and do the following actions

- Retain the join for the same catalog (Jdbc connector) tables instead of breaks to multiple select query
-  Pass predicates that are related to that query before you pass it to JDBC.

## Context
This git issue is to address a performance limitation of Presto federation of SQLs of JDBC connector to remote data sources such as DB2, Postgres, Oracle etc.

What is the limitation:
For a given query that might involve different remote data sources, the query is divided into sub-queries that can be sent to the respective data source, results of these sub-queries are fetched into the presto workers, additional operations such as filters, joins, sorts are applied before sending back to the user.

Operations such as filters, joins involve tables from the same remote data source, it is best to "Push down" or send these joins as part of the SQL query to the remote data source and it can be 10x to 100x faster than fetch all data of the table into presto workers and then apply these operators.

zhangbutao:
I think this issue is similar to https://github.com/prestodb/presto/pull/16583

Thanzeel-Hassan-IBM:
An RFC is raised for this : https://github.com/prestodb/rfcs/pull/32

PR #24115, opened 2024-11-22
Introduce SPI and Core Support for JDBC Join Pushdown Optimization

Haritha-Koloth:
## Description
This change introduces foundational support for join pushdown in JDBC connectors. By delegating join operations, directly to the remote datasource, it unlocks significant performance gains by leveraging the processing capabilities of the underlying datasource.

## Motivation and Context
Fixes [23152](https://github.com/prestodb/presto/issues/23152)

## Impact
This update includes only the SPI and core changes for the feature. Users will see its impact once the JDBC modifications to leverage the pushdown optimizer are also integrated. The primary benefit is enhanced performance for INNER join queries involving large tables with substantial row counts.

Detailed RFC - [https://github.com/prestodb/rfcs/pull/32](https://github.com/prestodb/rfcs/pull/32)

## Test Plan

- Unit tests added for newly added optimizer

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Add configuration property ``optimizer.inner-join-pushdown-enabled`` to support inner join pushdown. Defaults to ``false``. Session property to override this ``optimizer_inner_join_pushdown_enabled``. :pr:`24115`
* Add configuration property ``optimizer.inequality-join-pushdown-enabled`` to support inequality join pushdown. Defaults to ``false``. Session property to override this ``optimizer_inequality_join_pushdown_enabled``. :pr:`24115`

```

```
== NO RELEASE NOTE ==
```


Haritha-Koloth:
The jdbc code to leverage the changes in this PR could be found here - [https://github.com/Haritha-Koloth/presto/pull/2](https://github.com/Haritha-Koloth/presto/pull/2)

Haritha-Koloth:
Test failure seems to be unrelated to the code changes in this PR. Raised a sample PR with ReadMe change and it fails with the same error too - [https://github.com/prestodb/presto/pull/24118](https://github.com/prestodb/presto/pull/24118)

Will force-push after some time to see if the error resolves.

lschampion:
this feature is very useful, and significant when presto as virtual data engine.

jaystarshot:
`By delegating join operations, directly to the remote datasource, it unlocks significant performance gains by leveraging the processing capabilities of the underlying datasource.`

Just curious, why will delegating joins to the datasource result in performance gains?

aaneja:
@jaystarshot Join pushdown will help when -

1. We have a reducing join - data that Presto reads from the 'join' source will be much smaller than streaming the two join sources
2. The underlying relational system can do more joins more efficiently than Presto - say use index lookups instead of having to do a hash-join 

See [these examples](https://github.com/prestodb/rfcs/blob/main/RFC-0009-jdbc-join-push-down.md#individual-connectors) listed in the RFC

presto-main/src/test/java/com/facebook/presto/sql/planner/optimizations/TestGroupInnerJoinsByConnector.java
```diff
@@ -0,0 +1,531 @@
...
+        ConnectorId connectorId1 = new ConnectorId(catalog1);
+        ConnectorId connectorId2 = new ConnectorId(catalog2);
+
+        MockConnectorFactory.Builder builder = MockConnectorFactory.builder();
+        MockConnectorFactory mockConnectorFactory = builder.withListSchemaNames(connectorSession -> ImmutableList.of("test_schema"))
+                .withListTables((connectorSession, schemaNameOrNull) ->
+                        ImmutableList.of(
+                                new SchemaTableName("test_schema", "test_view"),
+                                new SchemaTableName("test_schema", "another_table")))
+                .withConnectorCapabilities(ImmutableSet.of()).build();
```
aaneja, line 184: Let's have all the connectors SUPPORTS_JOIN_PUSHDOWN. The test should work with or without the connectors supporting it

presto-main/src/test/java/com/facebook/presto/sql/planner/optimizations/TestGroupInnerJoinsByConnector.java
```diff
@@ -0,0 +1,531 @@
...
+
+public class TestGroupInnerJoinsByConnector
+{
+    private static final Metadata METADATA = MetadataManager.createTestMetadataManager();
+    private static final PlanBuilder PLAN_BUILDER = new PlanBuilder(TEST_SESSION, new PlanNodeIdAllocator(), METADATA);
+
+    @Test
+    public void testDoesNotPushDownOuterJoin()
+    {
+        String catalogName = "test_catalog";
```
aaneja, line 116: Please try refactoring this setup code into smaller methods that can be reused across tests

presto-main/src/test/java/com/facebook/presto/sql/planner/optimizations/TestGroupInnerJoinsByConnector.java
```diff
@@ -0,0 +1,531 @@
...
+                createNewSession(transactionId),
+                new InformationSchemaTableHandle(catalogName, "information_schema", "views"),
+                constraint,
+                Optional.empty());
+
+        VariableReferenceExpression left = newBigintVariable("a1");
+        VariableReferenceExpression right = newBigintVariable("a2");
+        EquiJoinClause joinClause = new EquiJoinClause(left, right);
+
+        PlanNode plan = output(join(FULL,
```
aaneja, line 161: Join pushdown will also not work for LEFT and RIGHT joins, can we add those assertions too ?

presto-main/src/test/java/com/facebook/presto/sql/planner/optimizations/TestGroupInnerJoinsByConnector.java
```diff
@@ -0,0 +1,531 @@
...
+        Constraint<ColumnHandle> constraint = new Constraint<>(TupleDomain.withColumnDomains(domains.build()));
+
+        InformationSchemaMetadata informationSchemaMetadata = new InformationSchemaMetadata(catalogName, metadata);
+        informationSchemaMetadata.getTableLayouts(
+                createNewSession(transactionId),
+                new InformationSchemaTableHandle(catalogName, "information_schema", "views"),
+                constraint,
+                Optional.empty());
+
+        PlanBuilder planBuilder = new PlanBuilder(session, new PlanNodeIdAllocator(), metadata);
```
aaneja, line 369: Unused

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/GroupInnerJoinsByConnector.java
```diff
@@ -0,0 +1,663 @@
...
+ *   `-- TableScanNode (with all the details of the three TableScanNodes)
+ *       `-- Set<ConnectorTableHandle> (ConnectorHandleSet)
+ *          `-- TableHandle1
+ *          `-- TableHandle2
+ *          `-- TableHandle3
+ */
+public class GroupInnerJoinsByConnector
+        implements PlanOptimizer
+{
+    private static final Logger logger = Logger.get(GroupInnerJoinsByConnector.class);
```
aaneja, line 119: Unused

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/GroupInnerJoinsByConnector.java
```diff
@@ -0,0 +1,663 @@
...
+        this.determinismEvaluator = new RowExpressionDeterminismEvaluator(metadata.getFunctionAndTypeManager());
+        this.metadata = metadata;
+    }
+
+    @Override
+    public PlanOptimizerResult optimize(PlanNode plan, Session session, TypeProvider types, VariableAllocator variableAllocator, PlanNodeIdAllocator idAllocator, WarningCollector warningCollector)
+    {
+        if (isEnabled(session)) {
+            PlanNode rewrittenPlan = SimplePlanRewriter.rewriteWith(new Rewriter(functionResolution, determinismEvaluator, idAllocator, metadata, session), plan);
+            return PlanOptimizerResult.optimizerResult(rewrittenPlan, !rewrittenPlan.equals(plan));
```
aaneja, line 137: We've been preferring the use of the IterativeOptimizer over PlanOptimizer.  

Since your rewrite can very explicitly determine if the plan is changed (by using `rewrittenPlan.equals(plan)`) modifying this class to instead implement an iterative optimizer rule (i.e `implements Rule<JoinNode>`) should be pretty straightforward.

This should reduce the test burden as well - you will be able to use a `RuleTester` and its associated helpers directly

aaneja (commented):
Did a quick first pass. I think we should switch `GroupInnerJoinsByConnector` to be an iterative optimizer rule.

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/GroupInnerJoinsByConnector.java
Ajas-Mangal: Hi @aaneja , @Haritha-Koloth 

Below are the reasons to use PlanOptimizer instead of Iterative optimizer Rule.

1) Need to visit Multiple nodes (visitJoin, visitFilter, etc.). Using a Rule we may able to perform operations on a single node only, I guess.

2) Repeated application of JoinNode Rule, due to child join.

Not sure these cases can resolved through a Single Iterative Optimizer Rule. For handling filter node, required a Rule<FilterNode> and for join required another Rule<JoinNode>.

Haritha-Koloth: @aaneja Thanks! We are exploring the potential and viability of the above suggestions.

Ajas-Mangal: Hi @Haritha-Koloth and @aaneja 

I think we already did some good test coverage for this using PlanOptimizer itself.

Do we have any additional benefit for changing to  optimizer Rule? If we use optimizer Rule how we could handlw non-equijoin scenario where we have to visit FilterNode and JoinNode

Haritha-Koloth: @Ajas-Mangal As per the discussion with @aaneja, the usage of IterativeOptimizer is being preferred over PlanOptimizer implementations and there is a potential advantage of opening up more nodes for pushdown. 

As for the non-equijoin scenario, I guess we can have multiple rule matchers to handle the filter and join nodes.

I am trying out the suggested changes, will update here if met with some blockers/concerns.

ethanyzhang:
Hi @aaneja , any next step on this?

aaneja:
> Hi @aaneja , any next step on this?

A rewrite of the existing rule as an IterativeOptimizer is being explored, see https://github.com/prestodb/presto/pull/24115#discussion_r1883736093.

Haritha-Koloth:
> Please add documentation for the configuration properties `optimizer.inner-join-pushdown-enabled` and `optimizer.inequality-join-pushdown-enabled`, and the session properties `optimizer_inner_join_pushdown_enabled` and `optimizer_inequality_join_pushdown_enabled`.

@steveburnett - Is this the correct file to add documentation for the above configurations?
 - [https://github.com/prestodb/presto/blob/05bc56ceddaf7debed9f5fd7aa20e07093aea969/presto-docs/src/main/sphinx/admin/properties.rst#optimizer-properties](https://github.com/prestodb/presto/blob/05bc56ceddaf7debed9f5fd7aa20e07093aea969/presto-docs/src/main/sphinx/admin/properties.rst#optimizer-properties)

steveburnett:
> > Please add documentation for the configuration properties `optimizer.inner-join-pushdown-enabled` and `optimizer.inequality-join-pushdown-enabled`, and the session properties `optimizer_inner_join_pushdown_enabled` and `optimizer_inequality_join_pushdown_enabled`.
> 
> @steveburnett - Is this the correct file to add documentation for the above configurations?
> 
>     * https://github.com/prestodb/presto/blob/05bc56ceddaf7debed9f5fd7aa20e07093aea969/presto-docs/src/main/sphinx/admin/properties.rst#optimizer-properties

Yes, for the configuration properties. Please also document the session properties by updating https://github.com/prestodb/presto/blob/05bc56ceddaf7debed9f5fd7aa20e07093aea969/presto-docs/src/main/sphinx/admin/properties-session.rst.

Haritha-Koloth:
Hi @aaneja 

We explored your suggestion to convert GroupInnerJoinsByConnector into an IterativeOptimizer. Below are our observations and findings:

- Equality Joins: For equality joins, we are able to construct a single table scan node based on the current logic. However, this approach fails because the IterativeOptimizer flow requires the input and output variables to remain consistent. When building the single table scan node, we need to include output variables from all sources to enable predicate pushdown. Below is the specific error encountered after the optimizer is run once.
`java.lang.IllegalArgumentException: com.facebook.presto.sql.planner.optimizations.GroupInnerJoinsByConnector: transformed expression doesn't produce same outputs: [intcolumn2, varcharcolumn1_2] vs [intcolumn1, intcolumn2, intcolumn1_0, varcharcolumn1_2] for node: com.facebook.presto.spi.plan.FilterNode@d85792e3`

- Inequality Joins: To handle inequality joins, we may need to create a separate rule that can accommodate FilterNodes. Alternatively, we may need to enhance the logic in [Pattern.java](https://github.com/prestodb/presto/blob/f2b7f88404f5d87ef016372de508b96800cf4482/presto-matching/src/main/java/com/facebook/presto/matching/Pattern.java) to return a generic node type(PlanNode) after evaluating multiple pattern matches and implement Rule<PlanNode> from GroupInnerJoinsByConnector.

- Exit Condition: Additionally, we may need to introduce an exit condition to skip over processed join nodes. This would address scenarios where nodes are being iteratively processed and prevent redundant processing.

The modified code is available here - [iterative_opt_pushdown](https://github.com/Haritha-Koloth/presto/tree/iterative_opt_pushdown)

Please let us know your thoughts on these.

CC: @Ajas-Mangal

aaneja:
1. Yes, the `IterativeOptimizer` has an invariant that a transformation cannot produce more or less output variables than the input plan nodes. A simple way to restrict the ouputs is to add a `ProjectNode` that constrains the output to the required variables, e.g https://github.com/aaneja/presto/commit/dc2eff4fed15c02bfeee5053ff50d540ea0632f3. If there is scope for pushing down this ProjectNode, it will be handled by later rules.
2. The pattern matcher allows for creating multiple nested matchers and has the ability to extract any specific node during the matching process. See [this matcher](https://github.com/prestodb/presto/blob/84e3c18b651d2c5cd95cd62999963ef78bc18ba9/presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/TransformDistinctInnerJoinToLeftEarlyOutJoin.java#L78-L84) for example. If you can describe the kind of matcher you need, I can help you build it
3. Can you describe an example of redundant processing ? There can be cases where we reapply the same rule more than once on the same already optimized node. This is by design. The 'exit' condition we want for an IterativeOptimizer rule is that there are no plan modifications after rule re-application

Reg 1 - I did a couple of join pushdown tests with the outputs restricted [(see aaneja/iterative_opt_pushdown)](https://github.com/prestodb/presto/compare/master...aaneja:presto:iterative_opt_pushdown). I did not find any errors here. Please feel to try it out with your setup and let me know if you spot any failures

presto-docs/src/main/sphinx/admin/properties-session.rst
```diff
@@ -336,7 +336,28 @@ The corresponding configuration property is :ref:`admin/properties:\`\`optimizer
...
+
+The corresponding configuration property is :ref:`admin/properties:\`\`optimizer.inner-join-pushdown-enabled\`\``.
+
+``optimizer_inequality_join_pushdown_enabled``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``boolean``
+* **Default value:** ``false``
+
+Enable push down inner join inequality predicates to database. For this configuration to be enabled, :ref:`admin/properties-session:\`\`optimizer_inner_join_pushdown_enabled\`\`` should be set to true.
```
steveburnett, line 358:
```suggestion
Enable push down inner join inequality predicates to database. For this configuration to be enabled, :ref:`admin/properties-session:\`\`optimizer_inner_join_pushdown_enabled\`\`` should be set to ``true``.
```

presto-docs/src/main/sphinx/admin/properties.rst
```diff
@@ -900,7 +900,28 @@ The corresponding session property is :ref:`admin/properties-session:\`\`treat-l
...
+
+The corresponding session property is :ref:`admin/properties-session:\`\`optimizer_inner_join_pushdown_enabled\`\``.
+
+``optimizer.inequality-join-pushdown-enabled``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``boolean``
+* **Default value:** ``false``
+
+Enable push down inner join inequality predicates to database. For this configuration to be enabled, :ref:`admin/properties:\`\`optimizer.inner-join-pushdown-enabled\`\`` should be set to true.
```
steveburnett, line 922:
```suggestion
Enable push down inner join inequality predicates to database. For this configuration to be enabled, :ref:`admin/properties:\`\`optimizer.inner-join-pushdown-enabled\`\`` should be set to ``true``.
```

steveburnett (changes requested):
Thanks for the doc! Formatting nits, all the links and everything else looks good.

presto-docs/src/main/sphinx/admin/properties.rst
Haritha-Koloth: Thanks @steveburnett . Updated with suggested changes.

steveburnett (dismissed):
LGTM! (docs)

Haritha-Koloth:
> * The pattern matcher allows for creating multiple nested matchers and has the ability to extract any specific node during the matching process. See [this matcher](https://github.com/prestodb/presto/blob/84e3c18b651d2c5cd95cd62999963ef78bc18ba9/presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/TransformDistinctInnerJoinToLeftEarlyOutJoin.java#L78-L84) for example. If you can describe the kind of matcher you need, I can help you build it

Hi @aaneja , regarding the point 2):

We need to build a matcher that can accept a join OR a filter. But I could not find any examples where 2 different types of nodes can be handled via a single rule. The kind of node we are trying to accept should be mentioned on the implementation also, right? How can we accommodate 2 different types there? Did you mean that we can have 2 different rules and patterns for these 2 use cases? Or is there a way we could use a generic PlanNode for this?

aaneja:
@Haritha-Koloth What you have is a case where you want to match on `JoinNode` or `Filter<-JoinNode` and in the latter case, use the filter as an extra filter while building the 'combined' MultiJoinNode.
You can use a rule-set to define the relevant patterns and call a base-class for the actual rewriting
I rewrote your code here - https://github.com/aaneja/presto/commit/ed62d8b812a3dfdb1fa7dc1417b49b780f77458d

Haritha-Koloth:
@aaneja After the above suggested changes, the pushdown feature is working as expected and the modified code is [here](https://github.com/Haritha-Koloth/presto/pull/3). But for the unit tests for the optimizer, the current RuleTester flow cannot be reused as there are connector capability checks which requires a MockConnector. We attempted to use mock connector along with the RuleTester, but the RuleTester flow failed due to session id mismatch. We are looking into this. Please let us know your insights on this.

Haritha-Koloth:
@aaneja Regarding your suggestion to make all connectors support Pushdown(connector capability that restricts the feature) [https://github.com/prestodb/presto/pull/24115#discussion_r1883708429](https://github.com/prestodb/presto/pull/24115#discussion_r1883708429), how will we limit this feature for non-jdbc connectors? There should be some form of control on which connectors can have this feature, right? We added this capability and certain implementations to fetch connector id and translation logic only for JDBC connectors in the assumption that the feature will only benefit them. Please let us know your thoughts on this.

presto-main/src/test/java/com/facebook/presto/sql/planner/iterative/rule/TestGroupInnerJoinsByConnectorRuleSet.java
```diff
@@ -0,0 +1,365 @@
...
+import static com.facebook.presto.sql.planner.assertions.PlanMatchPattern.node;
+import static com.facebook.presto.sql.planner.assertions.PlanMatchPattern.project;
+import static com.facebook.presto.sql.planner.iterative.rule.test.PlanBuilder.expression;
+import static com.google.common.base.MoreObjects.toStringHelper;
+import static com.google.common.base.Preconditions.checkState;
+import static com.google.common.collect.ImmutableList.toImmutableList;
+import static java.util.function.Function.identity;
+import static java.util.stream.Collectors.toMap;
+
+public class TestGroupInnerJoinsByConnectorRuleSet
```
aaneja, line 98: Please add tests for -
1. We have four tables from two distinct connectors and the input join-graph is of different shapes, say `(conn1Table1 IJ con2Table2) IJ (conn1Table3 IJ con2Table4)`, and we can end up with the right table groupings
2. After the rule is applied, join conditions and filters are pulled up

presto-main/src/main/java/com/facebook/presto/metadata/Metadata.java
```diff
@@ -517,4 +519,8 @@ default TableLayoutFilterCoverage getTableLayoutFilterCoverage(Session session,
     void dropConstraint(Session session, TableHandle tableHandle, Optional<String> constraintName, Optional<String> columnName);
 
     void addConstraint(Session session, TableHandle tableHandle, TableConstraint<String> tableConstraint);
+    default boolean isPushdownSupportedForFilter(Session session, TableHandle tableHandle, RowExpression filter, Map<VariableReferenceExpression, ColumnHandle> symbolToColumnHandleMap)
```
aaneja, line 522: Add a new line before the method. Can you add a Javadoc explaining what this method is for, why it is needed and when can connector's return `true` as the result ?

presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java
```diff
@@ -2884,6 +2886,31 @@ public int getEagerPlanValidationThreadPoolSize()
         return this.eagerPlanValidationThreadPoolSize;
     }
 
+    @Config("optimizer.inner-join-pushdown-enabled")
+    @ConfigDescription("push down inner join predicates to database")
```
aaneja, line 2890: nit: Capitalize `Push`

presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java
```diff
@@ -2884,6 +2886,31 @@ public int getEagerPlanValidationThreadPoolSize()
...
+        return this;
+    }
+
+    public boolean isInnerJoinPushdownEnabled()
+    {
+        return innerJoinPushdownEnabled;
+    }
+
+    @Config("optimizer.inequality-join-pushdown-enabled")
+    @ConfigDescription("push down inner join inequality predicates to database")
```
aaneja, line 2903: nit: Capitalize `Push`

presto-main/src/main/java/com/facebook/presto/connector/informationSchema/InformationSchemaMetadata.java
```diff
@@ -251,6 +251,32 @@ public ConnectorTableLayoutResult getTableLayoutForConstraint(ConnectorSession s
         return new ConnectorTableLayoutResult(layout, constraint.getSummary());
     }
 
+    @Override
+    public List<ConnectorTableLayoutResult> getTableLayouts(ConnectorSession session, ConnectorTableHandle table, Constraint<ColumnHandle> constraint, Optional<Set<ColumnHandle>> desiredColumns)
```
aaneja, line 255: Is this related to this PR ?

presto-spi/pom.xml
```diff
@@ -88,7 +88,7 @@
         <dependency>
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
-            <scope>test</scope>
+            <scope>provided</scope>
```
aaneja, line 91: Is this needed ?

aaneja (changes requested):

presto-main/src/main/java/com/facebook/presto/sql/planner/PlanOptimizers.java
```diff
@@ -770,6 +771,15 @@ public PlanOptimizers(
         // After this step, nodes with same `statsEquivalentPlanNode` will share same history based statistics.
         builder.add(new StatsRecordingPlanOptimizer(optimizerStats, new HistoricalStatisticsEquivalentPlanMarkingOptimizer(statsCalculator)));
 
+        builder.add(new IterativeOptimizer(
+                metadata,
+                ruleStats,
+                statsCalculator,
+                estimatedExchangesCostCalculator,
+                new GroupInnerJoinsByConnectorRuleSet(metadata).rules()));
+        builder.add(new ApplyConnectorOptimization(() -> planOptimizerManager.getOptimizers(STRUCTURAL)));
```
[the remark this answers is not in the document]
Haritha-Koloth: You mean adding the connector optimizer? Or adding it as a STRUCTURAL one?

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/GroupInnerJoinsByConnectorRuleSet.java
```diff
@@ -0,0 +1,734 @@
...
+                    ImmutableSet.Builder<VariableReferenceExpression> builder)
+            {
+                builder.add(variable);
+                return null;
+            }
+        }
+
+        protected PlanNode getCombinedJoin(JoinNode node, FunctionResolution functionResolution, DeterminismEvaluator determinismEvaluator, Metadata metadata, Session session, Lookup lookup, PlanNodeIdAllocator idAllocator)
+        {
+            if (node.getType() == INNER) {
```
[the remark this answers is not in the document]
Haritha-Koloth: We do not have the plan available when isEnabled() [invoked](https://github.com/prestodb/presto/blob/e884a445f27fb86337bdb9e5a9ea8f4e413fb5ba/presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/IterativeOptimizer.java#L182), right? 
Or do you mean since we have the [below check in getPattern()](https://github.com/Haritha-Koloth/presto/blob/b268dd9577af27b2deeb8ab8500a5d65e86f3769/presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/GroupInnerJoinsByConnectorRuleSet.java#L191), the above check is redundant?

presto-spi/pom.xml
Haritha-Koloth: yes, we have compile time dependency on the library

presto-main/src/main/java/com/facebook/presto/connector/informationSchema/InformationSchemaMetadata.java
Haritha-Koloth: This seems to be a delta from an internal working fork and differs from the upstream. I am checking what exactly was breaking without this change and will update here.

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/GroupInnerJoinsByConnectorRuleSet.java
```diff
@@ -0,0 +1,734 @@
...
+        @Override
+        public Pattern<JoinNode> getPattern()
+        {
+            return join().matching(
+                    joinNode -> joinNode.getType() == INNER
+                            && determinismEvaluator.isDeterministic(joinNode.getFilter().orElse(TRUE_CONSTANT)));
+        }
+
+        @Override
+        public boolean isEnabled(Session session)
```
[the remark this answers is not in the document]
aaneja: You can refactor the base class to do this. See https://github.com/aaneja/presto/commit/eb1f14e71798d1b65e57ded77cdbcb64d4777023

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/GroupInnerJoinsByConnectorRuleSet.java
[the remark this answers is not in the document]
aaneja: Sorry, I meant the rule pattern check ensures that the plan matched only Join nodes with an `INNER` join. You can remoe the `if` check

presto-spi/pom.xml
aaneja: We want to avoid bringing in Guava as a compile time dependency in the SPI. What classes are used ? We will need to rewrite these classes

presto-main/src/main/java/com/facebook/presto/sql/planner/PlanOptimizers.java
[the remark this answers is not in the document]
aaneja: 1. We did not use `STRUCTURAL` in the internal PR
2. We are not adding the `JdbcJoinPushdown` connector plan optimizer in this PR. Let's add this change in your next PR that will need it

presto-main/src/test/java/com/facebook/presto/sql/planner/iterative/rule/TestGroupInnerJoinsByConnectorRuleSet.java
```diff
@@ -0,0 +1,365 @@
...
+        {
+            checkState(shapeMatches(node), "Plan testing framework error: shapeMatches returned false in detailMatches in %s", this.getClass().getName());
+            TableScanNode tableScanNode = (TableScanNode) node;
+            ConnectorTableHandle connectorHandle = tableScanNode.getTable().getConnectorHandle();
+
+            if (connectorId.equals(tableScanNode.getTable().getConnectorId()) &&
+                    connectorHandle instanceof JoinTableSet && connectorHandle.equals(tableScanNode.getTable().getConnectorHandle())) {
+                return MatchResult.match(SymbolAliases.builder().putAll(Arrays.stream(columns).collect(toMap(identity(), SymbolReference::new))).build());
+            }
+
```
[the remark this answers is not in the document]
Haritha-Koloth: The above check 
image (14)
would always fail as there is a transaction field which is also compared in the TableHandle.equals() and will always be different. But if we just look at the connectorHandle inside tablescan and the output tablehandle, it would match. 
image (13)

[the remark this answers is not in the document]
aaneja: Okay, so lets write a custom comparer without the transaction handle

presto-main/src/main/java/com/facebook/presto/sql/analyzer/FeaturesConfig.java
Haritha-Koloth: Done

presto-main/src/main/java/com/facebook/presto/metadata/Metadata.java
Haritha-Koloth: Done

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/GroupInnerJoinsByConnectorRuleSet.java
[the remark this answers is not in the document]
Haritha-Koloth: Done

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/GroupInnerJoinsByConnectorRuleSet.java
```diff
@@ -0,0 +1,734 @@
...
+ * <p>
+ * Suppose that TableScanNode1, TableScanNode2 and TableScanNode3 are from the same catalog.
+ * <p>
+ * After Transformation:
+ * --OutputNode
+ * `-- TableScanNode (with all the details of the three TableScanNodes)
+ * `-- Set<ConnectorTableHandle> (ConnectorHandleSet)
+ * `-- TableHandle1
+ * `-- TableHandle2
+ * `-- TableHandle3
```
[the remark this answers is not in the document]
Haritha-Koloth: Done

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/GroupInnerJoinsByConnectorRuleSet.java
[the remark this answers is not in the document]
Haritha-Koloth: Done

presto-main/src/main/java/com/facebook/presto/sql/planner/PlanOptimizers.java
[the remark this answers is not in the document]
Haritha-Koloth: Done

presto-main/src/test/java/com/facebook/presto/sql/planner/iterative/rule/TestGroupInnerJoinsByConnectorRuleSet.java
```diff
@@ -0,0 +1,365 @@
...
+        VariableReferenceExpression right = newBigintVariable("a2");
+        EquiJoinClause joinClause = new EquiJoinClause(left, right);
+
+        Set<JoinTableInfo> joinTableInfos = new HashSet<>();
+        Map<VariableReferenceExpression, ColumnHandle> assignments = ImmutableMap.of(newBigintVariable("a1"), new TestingColumnHandle("a1"), newBigintVariable("a2"),
+                new TestingColumnHandle("a2"));
+        List<VariableReferenceExpression> outputVariables = ImmutableList.of(newBigintVariable("a1"), newBigintVariable("a2"));
+        JoinTableInfo joinTableInfo = new JoinTableInfo(new TestingMetadata.TestingTableHandle(new SchemaTableName("table_schema", "test-table")),
+                assignments, outputVariables);
+        joinTableInfos.add(joinTableInfo);
```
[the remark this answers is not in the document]
Haritha-Koloth: Updated

presto-main/src/test/java/com/facebook/presto/sql/planner/iterative/rule/TestGroupInnerJoinsByConnectorRuleSet.java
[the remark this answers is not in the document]
Haritha-Koloth: Done

presto-main/src/test/java/com/facebook/presto/sql/planner/iterative/rule/TestGroupInnerJoinsByConnectorRuleSet.java
Haritha-Koloth: Done

presto-main/src/test/java/com/facebook/presto/sql/planner/iterative/rule/TestGroupInnerJoinsByConnectorRuleSet.java
```diff
@@ -0,0 +1,490 @@
...
+                tableHandleSet2,
+                TestingTransactionHandle.create(),
+                Optional.empty());
+
+        assertGroupInnerJoinsByConnectorRuleSet()
+                .on(p ->
+                        p.join(
+                                INNER,
+                                p.join(
+                                        INNER,
```
aaneja, line 328: Can we mix this up - instead of having a first level joins between the same catalog table, create a first level join between different catalogs.

Use a join clause that allows for a JOIN between all tables like `table1.col1 = table2.col1 AND table2.col1 = table3.col1 and table3.col1 = table4.col1 and table4.col1 = table1.col1`

presto-main/src/test/java/com/facebook/presto/sql/planner/iterative/rule/TestGroupInnerJoinsByConnectorRuleSet.java
```diff
@@ -0,0 +1,490 @@
...
+                        joinClause))
+                .matches(project(filter(
+                        "a1 = a2 and a1 > b1 and true",
+                        JoinTableScanMatcher.tableScan(CATALOG_SUPPORTING_JOIN_PUSHDOWN, tableHandle, "a1", "a2", "b1"))));
+    }
+
+    @Test
+    public void testPushDownWithTwoDifferentConnectors()
+    {
+        VariableReferenceExpression leftJoinKey1 = newBigintVariable("a1");
```
aaneja, line 261: It looks like a lot of these variables can be inlined. Please do that to make this code more readable

presto-main/src/test/java/com/facebook/presto/sql/planner/iterative/rule/TestGroupInnerJoinsByConnectorRuleSet.java
```diff
@@ -0,0 +1,490 @@
...
+                tableScan(CATALOG_SUPPORTING_JOIN_PUSHDOWN, "a2", "b2"),
+                joinClause)).matches(project(filter(
+                "a1 = a2 and true",
+                JoinTableScanMatcher.tableScan(CATALOG_SUPPORTING_JOIN_PUSHDOWN, tableHandle, "a1", "a2"))));
+    }
+
+    @Test
+    public void testJoinPushDownHappenedWithFilters()
+    {
+        VariableReferenceExpression left = newBigintVariable("a1");
```
aaneja, line 219: Same comment as below, inline variables that are only used once

presto-main/src/test/java/com/facebook/presto/sql/planner/iterative/rule/TestGroupInnerJoinsByConnectorRuleSet.java
```diff
@@ -0,0 +1,490 @@
...
+        EquiJoinClause joinClause = new EquiJoinClause(left, right);
+
+        assertGroupInnerJoinsByConnectorRuleSet().on(p -> p.join(FULL,
+                tableScan(connectorName, "a1", "b1"),
+                tableScan(connectorName, "a2", "b2"),
+                joinClause)).doesNotFire();
+    }
+
+    @Test
+    public void testDoesNotPushDownTwoDifferentConnectors()
```
aaneja, line 168: Let's enhance this test so oin 3 tables `(t1 IJ t2) IJ t3`, and `t1` and `t3` are from a catalog that can be joined, while `t2` is from a different catalog

aaneja (changes requested):

jaystarshot (commented):
Can you please clarify what changes are made in ReorderJoins rule? Since that code path will be common with other sources.
I only see that MultiJoinNode is moved out, if that is needed it should be in a separate commit.
Also Please cleanup commits so that it is easier for review.

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/GroupInnerJoinsByConnectorRuleSet.java
```diff
@@ -0,0 +1,726 @@
...
+ * |   `-- TableScanNode2
+ * |       `-- TableHandle2
+ * `-- TableScanNode3
+ * `-- TableHandle3
+ * <p>
+ * Suppose that TableScanNode1, TableScanNode2 and TableScanNode3 are from the same catalog.
+ * <p>
+ * After Transformation:
+ * --OutputNode
+ * `--FilterNode(pulled up equi-join clause + join filters)
```
jaystarshot, line 117: The transformation looks like its a FilterNode -> TableScan Node.

Where is the filter pushdown into the tablescan expected to happen?

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/GroupInnerJoinsByConnectorRuleSet.java
```diff
@@ -0,0 +1,726 @@
...
+ * |       `-- TableScanNode4
+ * |           `-- TableHandle4
+ * `-- TableScanNode3
+ *     `-- TableHandle3
+ * <p>
+ * Suppose that TableScanNode1, TableScanNode2 and TableScanNode3 are from the same catalog, but TableScanNode4 is from another catalog.
+ * <p>
+ * After Transformation:
+ * --OutputNode
+ * `--FilterNode(pulled up equi-join clause + join filters)
```
jaystarshot, line 143: Why does the filter of the 4th join needs to be pulled up?
Keeping it as an inner join should be more efficient?

jaystarshot (commented):
Can you please clarify what changes are made in ReorderJoins rule? Since that code path will be common with other sources.
I only see that MultiJoinNode is moved out, if that is needed it should be in a separate commit.
Also Please cleanup commits so that it is easier for review.

jaystarshot:
Can we add some warnings through warning collector or populate some runtime metrics when joins are pushed down to the participating db?

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/GroupInnerJoinsByConnectorRuleSet.java
aaneja: [Predicate pushdown](https://github.com/prestodb/presto/blob/9d0e329ead6faae8c466ecc8b2319a64cab67e1f/presto-main/src/main/java/com/facebook/presto/sql/planner/PlanOptimizers.java#L781) is run immediately after this ruleset is applied and rewrites the cross-joins back to inner-joins.

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/GroupInnerJoinsByConnectorRuleSet.java
aaneja: The filters that will be pulled up iff they can later be pushed down by the connector in question; e.g the [JDBC filter pushdown rule](https://github.com/prestodb/presto/blob/9d0e329ead6faae8c466ecc8b2319a64cab67e1f/presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/optimization/JdbcComputePushdown.java) will do this for JDBC based connectors

presto-main/src/main/java/com/facebook/presto/sql/planner/plan/MultiJoinNode.java
```diff
@@ -0,0 +1,180 @@
...
+
+import static com.facebook.presto.expressions.LogicalRowExpressions.extractConjuncts;
+import static com.facebook.presto.spi.plan.JoinType.INNER;
+import static com.google.common.collect.ImmutableList.toImmutableList;
+import static java.util.Objects.requireNonNull;
+
+/**
+ * This class represents a set of inner joins that can be executed in any order.
+ */
+public class MultiJoinNode
```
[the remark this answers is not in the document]
Haritha-Koloth: Yes, this is a refactoring change as we are reusing the MultiJoinNode for the new optimizer. Will raise a new commit for this change.

presto-spi/src/main/java/com/facebook/presto/spi/connector/ConnectorPlanOptimizerProvider.java
```diff
@@ -29,4 +30,12 @@ public interface ConnectorPlanOptimizerProvider
      * The plan will be only executed on a single node.
      */
     Set<ConnectorPlanOptimizer> getPhysicalPlanOptimizers();
+
+    /**
+     * The optimizers that ensure the plan is structured for execution.
+     */
+    default Set<ConnectorPlanOptimizer> getStructuralPlanOptimizers()
```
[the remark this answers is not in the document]
Haritha-Koloth: This change is for the subsequent PR [Haritha-Koloth#2](https://github.com/Haritha-Koloth/presto/pull/2) that leverages the SPI-level updates in this PR. We will be adding a structural plan optimizer to modify the plan at the connector level. We're including it here since this PR contains all the necessary SPI-level changes.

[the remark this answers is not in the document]
jaystarshot: lets move it to the subsequent PR, its not clear whats this supposed to do here + semantics might change

presto-main/src/main/java/com/facebook/presto/sql/planner/plan/MultiJoinNode.java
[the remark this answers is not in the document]
Haritha-Koloth: Done

presto-main-base/src/main/java/com/facebook/presto/sql/planner/optimizations/GroupInnerJoinsByConnectorRuleSet.java
```diff
@@ -0,0 +1,874 @@
...
+import static com.facebook.presto.sql.planner.plan.Patterns.join;
+import static com.facebook.presto.sql.planner.plan.Patterns.source;
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.collect.ImmutableList.toImmutableList;
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
+import static com.google.common.collect.Iterables.filter;
+import static java.util.Objects.requireNonNull;
+
+/**
+ * This optimizer attempts to group TableScanNode's of an inner-join graph that belong to the same connector
```
jaystarshot, line 97: Can we add some comment here on what role determinism plays because I see that rules filter deterministic join filters

presto-main-base/src/main/java/com/facebook/presto/sql/planner/PlanOptimizers.java
```diff
@@ -770,6 +771,15 @@ public PlanOptimizers(
...
         builder.add(new StatsRecordingPlanOptimizer(optimizerStats, new HistoricalStatisticsEquivalentPlanMarkingOptimizer(statsCalculator)));
 
+        builder.add(new IterativeOptimizer(
+                metadata,
+                ruleStats,
+                statsCalculator,
+                estimatedExchangesCostCalculator,
+                new GroupInnerJoinsByConnectorRuleSet(metadata).rules()));
+        // GroupInnerJoinsByConnectorRuleSet pulls up filters, we need to push these down again
+        builder.add(predicatePushDown, simplifyRowExpressionOptimizer);
```
jaystarshot, line 781: @aaneja This will add the predicate pushdown for all queries and can potentially change plans. Is there a way to only add this when this feature is turned on with configs?

presto-main-base/src/main/java/com/facebook/presto/sql/planner/optimizations/GroupInnerJoinsByConnectorRuleSet.java
Haritha-Koloth: Done

presto-main-base/src/main/java/com/facebook/presto/sql/planner/PlanOptimizers.java
jaystarshot: I think its better to gate since all the places are chosen specially and the order of optimizatiins can matter. Also it might add unwanted latency

aaneja: We do add predicate pushdown [speculatively in other places](https://github.com/prestodb/presto/blob/5dae81da37541684a25c92ea293eaabb02a14582/presto-main-base/src/main/java/com/facebook/presto/sql/planner/PlanOptimizers.java#L749-L758). I do have a proof-of-concept for [gating it though](https://github.com/aaneja/presto/commit/5dae81da37541684a25c92ea293eaabb02a14582). @Haritha-Koloth can you try this out

Haritha-Koloth: Sure, I'll test and update.

presto-main-base/src/main/java/com/facebook/presto/sql/planner/optimizations/GroupInnerJoinsByConnectorRuleSet.java
```diff
@@ -203,9 +207,9 @@ public static class FilterOnJoinRule
     {
         private static final Capture<JoinNode> JOIN = newCapture();
 
-        public FilterOnJoinRule(Metadata metadata)
+        public FilterOnJoinRule(Metadata metadata, PlanOptimizer predicatePushdownOptimizer)
```
aaneja, line 210: @Haritha-Koloth Can you share a query where we see a `FilterNode<-JoinNode` at this point in the planner ? I'm wondering if this rule variant is actually needed

presto-main-base/src/main/java/com/facebook/presto/sql/planner/PlanOptimizers.java
Haritha-Koloth: Hi @jaystarshot,
The above change is done. Please have a look and let us know your thoughts.

presto-main-base/src/main/java/com/facebook/presto/sql/planner/optimizations/GroupInnerJoinsByConnectorRuleSet.java
```diff
@@ -187,7 +191,7 @@ public Pattern<JoinNode> getPattern()
         @Override
         public Result apply(JoinNode node, Captures captures, Context context)
         {
-            PlanNode rewrittenPlan = getCombinedJoin(node, functionResolution, determinismEvaluator, metadata, context.getSession(), context.getLookup(), (PlanNodeIdAllocator) context.getIdAllocator());
+            PlanNode rewrittenPlan = getCombinedJoin(node, functionResolution, determinismEvaluator, metadata, context);
```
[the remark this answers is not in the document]
aaneja: We only need to do the predicate pushdown iff the rule actually applies, which we check via [getContainsCombinedSources](https://github.com/prestodb/presto/blob/ee02f8c31dfa7ac60eee7b0ce6ccf8dc78b09db8/presto-main-base/src/main/java/com/facebook/presto/sql/planner/optimizations/GroupInnerJoinsByConnectorRuleSet.java#L350C40-L350C66), so it makes sense to have it there.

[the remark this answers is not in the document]
jaystarshot: You already know if the rule applies here (as you are checking it below if rewrittenPlan.equals(node)). You can also pass the info if rule applied from there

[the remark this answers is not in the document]
aaneja: We need to perform predicate pushdown also for the `FilterOnJoinRule`. `getCombinedJoin` is the common method between the two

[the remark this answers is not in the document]
jaystarshot: Isn't this apply also common across them?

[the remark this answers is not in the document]
jaystarshot: okay there are two places so maybe its fine

jaystarshot (dismissed):
This will apply the predicate pushdown each time the rules are triggered but i think its fine since this change is gated

Haritha-Koloth:
> Please clean up the commits
> 
> ```
> Commits on Apr 1, 2025
> [Add comment on filter determinism check](https://github.com/prestodb/presto/pull/24115/commits/5759cd5084a652a8377c1527761cc77d22e26731)
> [Haritha-Koloth](https://github.com/prestodb/presto/commits?author=Haritha-Koloth)
> Haritha-Koloth
> committed
> 2 weeks ago
> ·
> Commits on Apr 9, 2025
> [Gating the predicatepushdown for Join Pushdown](https://github.com/prestodb/presto/pull/24115/commits/ee02f8c31dfa7ac60eee7b0ce6ccf8dc78b09db8)
> ```
> 
> seem unnecessary. I think one for spi changes, one for main changes and one refactoring the reorder joins is fine

Thanks @jaystarshot! I have cleaned up the commits.

jaystarshot (dismissed):
Nvm i see the problem with that you are using that in your optimizer.
Okay please separate the commits for spi changes

jaystarshot (approved):

aaneja (approved):

PR #24128, opened 2024-11-23
Add Hive "orc_max_coalesced_distance_bytes" session property

spershin:
## Description
Adding "orc_max_coalesced_distance_bytes" session property to the Hive connector.
This property would be used in Presto C++ to control read coalesce.
It can be increased for very wide tables and sparse reads to reduce number of storage calls.

## Motivation and Context
It can be increased for very wide tables and sparse reads to reduce number of storage calls.

## Contributor checklist

- [ ] CI passed.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-hive-common/src/main/java/com/facebook/presto/hive/HiveCommonSessionProperties.java
+++ b/presto-hive-common/src/main/java/com/facebook/presto/hive/HiveCommonSessionProperties.java
@@ -32,6 +32,7 @@ import static com.facebook.presto.common.type.VarcharType.VARCHAR;
 import static com.facebook.presto.common.type.VarcharType.createUnboundedVarcharType;
 import static com.facebook.presto.spi.StandardErrorCode.INVALID_SESSION_PROPERTY;
 import static com.facebook.presto.spi.session.PropertyMetadata.booleanProperty;
+import static com.facebook.presto.spi.session.PropertyMetadata.longProperty;
 import static com.facebook.presto.spi.session.PropertyMetadata.stringProperty;
 import static com.google.common.base.Preconditions.checkArgument;
 import static java.lang.String.format;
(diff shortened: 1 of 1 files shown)


arhimondr (approved):

presto-hive-common/src/main/java/com/facebook/presto/hive/HiveCommonSessionProperties.java
```diff
@@ -57,6 +58,7 @@ public class HiveCommonSessionProperties
     private static final String ORC_STREAM_BUFFER_SIZE = "orc_stream_buffer_size";
     private static final String ORC_TINY_STRIPE_THRESHOLD = "orc_tiny_stripe_threshold";
     private static final String ORC_ZSTD_JNI_DECOMPRESSION_ENABLED = "orc_zstd_jni_decompression_enabled";
+    private static final String ORC_MAX_COALESCED_DISTANCE_BYTES = "orc_max_coalesced_distance_bytes";
```
arhimondr, line 61: There is `orc_max_merge_distance` session property for Presto that does exactly the same. I wonder if it would be better to rename the property in Prestissimo / Velox to match the Presto name to avoid unnecessary confusion?

spershin: @arhimondr 
Great, I didn't know we have that already.
We should definitely use it.

We don't need another one.
It is a different name and of DataSize type, so it will need some translation in Presto C++ layer.

Yuhta: @spershin In Velox we would like to keep it as format agnostic, is it possible to do the translation `orc_max_merge_distance -> max-coalesced-distance-bytes` in Presto CPP layer?

PR #24144, opened 2024-11-25
Add SPI for delegating row expression optimizer

tdcmeehan:
## Description
As part of [RFC-0006](https://github.com/prestodb/rfcs/blob/main/RFC-0006-expression-eval.md), we need to support out of process expression evaluation.  Add support for pluggable expression optimization and planner support to utilize the new SPI.

## Motivation and Context
[RFC-0006](https://github.com/prestodb/rfcs/blob/main/RFC-0006-expression-eval.md). See https://github.com/prestodb/presto/pull/24126 for larger changes that include the Presto sidecar as described in the RFC.

## Impact
No impact by default as the old in-memory evaluation is the default.

## Test Plan
Tests have been added.

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

SPI Changes
* Add ``CoordinatorPlugin#getExpressionOptimizerFactories`` to customize expression evaluation in the Presto coordinator. :pr:`24144`
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
+++ b/presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
@@ -327,6 +327,7 @@ public final class SystemSessionProperties
     public static final String INLINE_PROJECTIONS_ON_VALUES = "inline_projections_on_values";
     public static final String INCLUDE_VALUES_NODE_IN_CONNECTOR_OPTIMIZER = "include_values_node_in_connector_optimizer";
     public static final String SINGLE_NODE_EXECUTION_ENABLED = "single_node_execution_enabled";
+    public static final String EXPRESSION_OPTIMIZER_NAME = "expression_optimizer_name";
 
     // TODO: Native execution related session properties that are temporarily put here. They will be relocated in the future.
     public static final String NATIVE_AGGREGATION_SPILL_ALL = "native_aggregation_spill_all";
@@ -1852,7 +1853,12 @@ public final class SystemSessionProperties
                 booleanProperty(NATIVE_EXECUTION_SCALE_WRITER_THREADS_ENABLED,
                         "Enable automatic scaling of writer threads",
                         featuresConfig.isNativeExecutionScaleWritersThreadsEnabled(),
-                        !featuresConfig.isNativeExecutionEnabled()));
+                        !featuresConfig.isNativeExecutionEnabled()),
+                stringProperty(
+                        EXPRESSION_OPTIMIZER_NAME,
+                        "Configure which expression optimizer to use",
+                        featuresConfig.getExpressionOptimizerName(),
+                        false));
     }
 
     public static boolean isSpoolingOutputBufferEnabled(Session session)
@@ -3152,5 +3158,10 @@ public final class SystemSessionProperties
     public static boolean isNativeExecutionScaleWritersThreadsEnabled(Session session)
     {
         return session.getSystemProperty(NATIVE_EXECUTION_SCALE_WRITER_THREADS_ENABLED, Boolean.class);
+    }
+
+    public static String getExpressionOptimizerName(Session session)
+    {
+        return session.getSystemProperty(EXPRESSION_OPTIMIZER_NAME, String.class);
     }
 }
--- a/presto-main/src/main/java/com/facebook/presto/server/testing/TestingPrestoServer.java
+++ b/presto-main/src/main/java/com/facebook/presto/server/testing/TestingPrestoServer.java
@@ -74,6 +74,7 @@ import com.facebook.presto.spi.security.AccessControl;
 import com.facebook.presto.split.PageSourceManager;
 import com.facebook.presto.split.SplitManager;
 import com.facebook.presto.sql.analyzer.FeaturesConfig;
+import com.facebook.presto.sql.expressions.ExpressionOptimizerManager;
 import com.facebook.presto.sql.parser.SqlParser;
 import com.facebook.presto.sql.parser.SqlParserOptions;
 import com.facebook.presto.sql.planner.ConnectorPlanOptimizerManager;
@@ -171,6 +172,7 @@ public class TestingPrestoServer
     private final TaskManager taskManager;
     private final GracefulShutdownHandler gracefulShutdownHandler;
     private final ShutdownAction shutdownAction;
+    private final ExpressionOptimizerManager expressionManager;
     private final RequestBlocker requestBlocker;
     private final boolean resourceManager;
     private final boolean catalogServer;
@@ -373,6 +375,7 @@ public class TestingPrestoServer
         procedureTester = injector.getInstance(ProcedureTester.class);
         splitManager = injector.getInstance(SplitManager.class);
         pageSourceManager = injector.getInstance(PageSourceManager.class);
+        expressionManager = injector.getInstance(ExpressionOptimizerManager.class);
         if (coordinator) {
             dispatchManager = injector.getInstance(DispatchManager.class);
             queryManager = injector.getInstance(QueryManager.class);
@@ -390,6 +393,7 @@ public class TestingPrestoServer
             eventListenerManager = ((TestingEventListenerManager) injector.getInstance(EventListenerManager.class));
             clusterStateProvider = null;
             planCheckerProviderManager = injector.getInstance(PlanCheckerProviderManager.class);
+            expressionManager.loadExpressionOptimizerFactories();
         }
         else if (resourceManager) {
             dispatchManager = null;
@@ -708,6 +712,11 @@ public class TestingPrestoServer
     public ShutdownAction getShutdownAction()
     {
         return shutdownAction;
+    }
+
+    public ExpressionOptimizerManager getExpressionManager()
+    {
+        return expressionManager;
     }
 
     public boolean isCoordinator()
--- a/presto-main/src/main/java/com/facebook/presto/sql/expressions/ExpressionOptimizerManager.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/expressions/ExpressionOptimizerManager.java
@@ -1,0 +1,137 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.sql.expressions;
+
+import com.facebook.presto.FullConnectorSession;
+import com.facebook.presto.Session;
+import com.facebook.presto.metadata.FunctionAndTypeManager;
+import com.facebook.presto.nodeManager.PluginNodeManager;
+import com.facebook.presto.spi.ConnectorSession;
+import com.facebook.presto.spi.NodeManager;
+import com.facebook.presto.spi.relation.ExpressionOptimizer;
+import com.facebook.presto.spi.sql.planner.ExpressionOptimizerContext;
+import com.facebook.presto.spi.sql.planner.ExpressionOptimizerFactory;
+import com.facebook.presto.sql.relational.FunctionResolution;
+import com.facebook.presto.sql.relational.RowExpressionOptimizer;
+import com.google.common.collect.ImmutableList;
+
+import javax.inject.Inject;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+import static com.facebook.presto.SystemSessionProperties.getExpressionOptimizerName;
+import static com.facebook.presto.util.PropertiesUtil.loadProperties;
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Strings.isNullOrEmpty;
+import static com.google.common.io.Files.getNameWithoutExtension;
+import static java.util.Objects.requireNonNull;
+
+public class ExpressionOptimizerManager
+{
+    public static final String DEFAULT_EXPRESSION_OPTIMIZER_NAME = "default";
+    private static final File EXPRESSION_MANAGER_CONFIGURATION_DIRECTORY = new File("etc/expression-manager/");
+    private static final String EXPRESSION_MANAGER_FACTORY_NAME = "expression-manager-factory.name";
+
+    private final NodeManager nodeManager;
+    private final FunctionAndTypeManager functionAndTypeManager;
+    private final FunctionResolution functionResolution;
+    private final File configurationDirectory;
+
+    private final Map<String, ExpressionOptimizerFactory> expressionOptimizerFactories = new ConcurrentHashMap<>();
+    private final Map<String, ExpressionOptimizer> expressionOptimizers = new ConcurrentHashMap<>();
+
+    @Inject
+    public ExpressionOptimizerManager(PluginNodeManager nodeManager, FunctionAndTypeManager functionAndTypeManager)
+    {
+        this(nodeManager, functionAndTypeManager, EXPRESSION_MANAGER_CONFIGURATION_DIRECTORY);
+    }
+
+    public ExpressionOptimizerManager(PluginNodeManager nodeManager, FunctionAndTypeManager functionAndTypeManager, File configurationDirectory)
+    {
+        requireNonNull(nodeManager, "nodeManager is null");
+        this.nodeManager = requireNonNull(nodeManager, "nodeManager is null");
+        this.functionAndTypeManager = requireNonNull(functionAndTypeManager, "functionAndTypeManager is null");
+        this.functionResolution = new FunctionResolution(functionAndTypeManager.getFunctionAndTypeResolver());
+        this.configurationDirectory = requireNonNull(configurationDirectory, "configurationDirectory is null");
+        expressionOptimizers.put(DEFAULT_EXPRESSION_OPTIMIZER_NAME, new RowExpressionOptimizer(functionAndTypeManager));
+    }
+
+    public void loadExpressionOptimizerFactories()
+    {
+        try {
+            for (File file : listFiles(configurationDirectory)) {
+                if (file.isFile() && file.getName().endsWith(".properties")) {
+                    loadExpressionOptimizerFactory(file);
+                }
+            }
+        }
+        catch (IOException e) {
+            throw new UncheckedIOException("Failed to load expression manager configuration", e);
+        }
+    }
+
+    private void loadExpressionOptimizerFactory(File configurationFile)
+            throws IOException
+    {
+        String name = getNameWithoutExtension(configurationFile.getName());
+        checkArgument(!isNullOrEmpty(name), "File name is empty, full path: %s", configurationFile.getAbsolutePath());
+        checkArgument(!name.equals(DEFAULT_EXPRESSION_OPTIMIZER_NAME), "Cannot name an expression optimizer instance %s", DEFAULT_EXPRESSION_OPTIMIZER_NAME);
+
+        Map<String, String> properties = new HashMap<>(loadProperties(configurationFile));
+        String factoryName = properties.remove(EXPRESSION_MANAGER_FACTORY_NAME);
+        checkArgument(!isNullOrEmpty(factoryName), "%s does not contain %s", configurationFile, EXPRESSION_MANAGER_FACTORY_NAME);
+        checkArgument(expressionOptimizerFactories.containsKey(factoryName),
+                "ExpressionOptimizerFactory %s is not registered, registered factories: ", factoryName, expressionOptimizerFactories.keySet());
+
+        ExpressionOptimizer optimizer = expressionOptimizerFactories.get(factoryName).createOptimizer(
+                properties,
+                new ExpressionOptimizerContext(nodeManager, functionAndTypeManager, functionResolution));
+        expressionOptimizers.put(name, optimizer);
+    }
+
+    public void addExpressionOptimizerFactory(ExpressionOptimizerFactory expressionOptimizerFactory)
+    {
+        String name = expressionOptimizerFactory.getName();
+        checkArgument(
+                expressionOptimizerFactories.putIfAbsent(name, expressionOptimizerFactory) == null,
+                "ExpressionOptimizerFactory %s is already registered", name);
+    }
+
+    public ExpressionOptimizer getExpressionOptimizer(ConnectorSession connectorSession)
+    {
+        // TODO: Remove this check once we have a more appropriate abstraction for session properties retrieved from plugins
+        checkArgument(connectorSession instanceof FullConnectorSession, "connectorSession is not an instance of FullConnectorSession");
+        Session session = ((FullConnectorSession) connectorSession).getSession();
+        String expressionOptimizerName = getExpressionOptimizerName(session);
+        checkArgument(expressionOptimizers.containsKey(expressionOptimizerName), "ExpressionOptimizer '%s' is not registered", expressionOptimizerName);
+        return expressionOptimizers.get(expressionOptimizerName);
+    }
+
+    private static List<File> listFiles(File directory)
+    {
+        if (directory.isDirectory()) {
+            File[] files = directory.listFiles();
+            if (files != null) {
+                return ImmutableList.copyOf(files);
+            }
+        }
+        return ImmutableList.of();
+    }
+}
--- a/presto-main/src/main/java/com/facebook/presto/sql/expressions/ExpressionOptimizerManager.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/expressions/ExpressionOptimizerManager.java
@@ -20,6 +20,7 @@ import com.facebook.presto.nodeManager.PluginNodeManager;
 import com.facebook.presto.spi.ConnectorSession;
 import com.facebook.presto.spi.NodeManager;
 import com.facebook.presto.spi.relation.ExpressionOptimizer;
+import com.facebook.presto.spi.relation.ExpressionOptimizerProvider;
 import com.facebook.presto.spi.sql.planner.ExpressionOptimizerContext;
 import com.facebook.presto.spi.sql.planner.ExpressionOptimizerFactory;
 import com.facebook.presto.sql.relational.FunctionResolution;
@@ -44,6 +45,7 @@ import static com.google.common.io.Files.getNameWithoutExtension;
 import static java.util.Objects.requireNonNull;
 
 public class ExpressionOptimizerManager
+        implements ExpressionOptimizerProvider
 {
     public static final String DEFAULT_EXPRESSION_OPTIMIZER_NAME = "default";
     private static final File EXPRESSION_MANAGER_CONFIGURATION_DIRECTORY = new File("etc/expression-manager/");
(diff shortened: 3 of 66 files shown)


presto-main/src/main/java/com/facebook/presto/server/testing/TestingPrestoServer.java
```diff
@@ -704,6 +706,11 @@ public ShutdownAction getShutdownAction()
         return shutdownAction;
     }
 
+    public ExpressionOptimizerManager getExpressionManager()
+    {
+        return injector.getInstance(ExpressionOptimizerManager.class);
```
pdabre12, line 711: Can we make this a member variable?

presto-main/src/main/java/com/facebook/presto/sql/expressions/ExpressionOptimizerManager.java
```diff
@@ -0,0 +1,105 @@
...
+    private final AtomicReference<ExpressionOptimizer> rowExpressionInterpreter = new AtomicReference<>();
+    private final NodeManager nodeManager;
+    private final FunctionAndTypeManager functionAndTypeManager;
+    private final FunctionResolution functionResolution;
+    private final ExpressionOptimizer defaultExpressionOptimizer;
+
+    @Inject
+    public ExpressionOptimizerManager(InternalNodeManager nodeManager, FunctionAndTypeManager functionAndTypeManager, NodeInfo nodeInfo)
+    {
+        requireNonNull(nodeManager, "nodeManager is null");
```
pdabre12, line 59: Is there a specific need for the `nodeManager` to be of type `InternalNodeManager`. It seems like it gets used for `pluginNodeManager`, why not inject an instance of  `pluginNodeManager` instead?

pdabre12 (commented):
Thanks @tdcmeehan. 
Just some initial comments

presto-tests/src/test/java/com/facebook/presto/memory/TestMemoryManager.java
```diff
@@ -512,7 +512,7 @@ public void clusterPoolsMultiCoordinatorCleanup()
         queryRunner2.close();
     }
 
-    @Test(timeOut = 60_000, groups = {"clusterPoolsMultiCoordinator"})
+    @Test(timeOut = 600_000, groups = {"clusterPoolsMultiCoordinator"})
```
ZacBlanco, line 515: why is this so much higher now?

presto-tests/src/test/java/com/facebook/presto/memory/TestMemoryManager.java
```diff
@@ -544,6 +544,7 @@ public void testClusterPoolsMultiCoordinator()
             generalPool = memoryManager.getClusterInfo(GENERAL_POOL);
             reservedPool = memoryManager.getClusterInfo(RESERVED_POOL);
             MILLISECONDS.sleep(10);
+            System.out.println("waiting");
```
ZacBlanco, line 547: remove

presto-main/src/main/java/com/facebook/presto/server/testing/TestingPrestoServer.java
```diff
@@ -385,6 +388,7 @@ public TestingPrestoServer(
             eventListenerManager = ((TestingEventListenerManager) injector.getInstance(EventListenerManager.class));
             clusterStateProvider = null;
             planCheckerProviderManager = injector.getInstance(PlanCheckerProviderManager.class);
+            expressionManager.loadExpressionOptimizerFactory();
```
ZacBlanco, line 391: I feel like having an additional method to configure/load the expression optimizer is a bit of an anti-pattern. It creates the potential for bugs if someone is trying to properly create the expression manager by requiring this additional method to be called. I haven't dug much deeper, but is there a reason we can't perform the loading of the factory in the constructor of the `expressionManager`?

presto-main/src/main/java/com/facebook/presto/sql/expressions/ExpressionOptimizerManager.java
```diff
@@ -0,0 +1,103 @@
...
+    public void addExpressionOptimizerFactory(ExpressionOptimizerFactory expressionOptimizerFactory)
+    {
+        String name = expressionOptimizerFactory.getName();
+        checkArgument(
+                this.expressionOptimizerFactories.putIfAbsent(name, expressionOptimizerFactory) == null,
+                "ExpressionOptimizerFactory %s is already registered", name);
+    }
+
+    @Override
+    public ExpressionOptimizer getExpressionOptimizer()
```
ZacBlanco, line 99: A thought I had while looking at the structure of the expression optimizer manager:

```suggestion
    public ExpressionOptimizer getExpressionOptimizer(Session session)
```


Do you think it would be beneficial to add the ability to pass session properties here? I could imagine a case where expression optimization could be configured in the future by the session or potentially switch/update configurations by the end user.

The current implementation requires the user to update config files and restart the server if there were other configs outside of just enabling/disabling the delegating optimizer. It might also allow you to encapsulate the configuration logic for returning the `DelegatingExpressionOptimizer` entirely in this class rather than having the current conditional inside of the `SimplifyRowExpressions` class.

presto-main/src/main/java/com/facebook/presto/sql/relational/DelegatingRowExpressionOptimizer.java
```diff
@@ -0,0 +1,90 @@
...
+            requireNonNull(rowExpression, "optimized expression is null");
+            if (originalExpression.equals(rowExpression)) {
+                break;
+            }
+        }
+        return rowExpression;
+    }
+
+    @Override
+    public Object optimize(RowExpression rowExpression, Level level, ConnectorSession session, Function<VariableReferenceExpression, Object> variableResolver)
```
ZacBlanco, line 72: not a big fan of having the duplicated code in these methods. Consider refactoring to a private method?

presto-main/src/main/java/com/facebook/presto/sql/expressions/ExpressionOptimizerManager.java
```diff
@@ -0,0 +1,103 @@
...
+
+import static com.facebook.presto.util.PropertiesUtil.loadProperties;
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Strings.isNullOrEmpty;
+import static java.util.Objects.requireNonNull;
+
+public class ExpressionOptimizerManager
+        implements ExpressionOptimizerProvider
+{
+    private static final File EXPRESSION_MANAGER_CONFIGURATION = new File("etc/expression-manager.properties");
```
ZacBlanco, line 44: Do we really need a separate config file for this? Is there a reason it can't just be rolled into the existing coordinator properties file?

Additionally, I feel like rather than having a flag which explicitly enables this, then having to configure the specific expression optimizer factory is complex. It would be simpler to have a single flag which specifies the `expression-manager-factory.name` and the default for the flag should be the current default behavior. From what I understand it should default to the name for the in-memory expression optimizer.

ZacBlanco (commented):
At a high-level, the changes make sense. I had a few questions about the configuration and how we might be able to simplify it to be more straightforward for users

presto-main/src/main/java/com/facebook/presto/sql/expressions/ExpressionOptimizerManager.java
rschlussel: I've been thinking about this for other plugins as well- of having the ability to pass session properties to plugins. Right now we only support session properties for connectors, but we should support them for other plugins too.  When I looked into this before, I think we needed to basically create a connectorSession for these plugins as well and be able to get session properties based on the name prefix of the plugin.  The only thing is that it means you can't e.g. give your catalog and function namespace manager the same name.

presto-main/src/main/java/com/facebook/presto/sql/expressions/ExpressionOptimizerManager.java
tdcmeehan: It's a separate config because the implementation will define properties that the Presto runtime can't know in advance.

I think this should be modeled as a plugin because it introduces a degree of separation between Presto and Velox, and also opens the door for third party integrations.

presto-main/src/main/java/com/facebook/presto/sql/expressions/ExpressionOptimizerManager.java
tdcmeehan: I think I can get this to work for now by passing in a ConnectorSession, and casting to FullConnectorSession inside of ExpressionOptimizerManager.  I agree this will reduce the branching that both of you have pointed out.  I agree we should have a design around plugin sessions as well.

presto-main/src/main/java/com/facebook/presto/sql/planner/iterative/rule/SimplifyRowExpressions.java
```diff
@@ -39,44 +44,52 @@
...
 
-        private RowExpression rewrite(RowExpression expression, ConnectorSession session)
+        private RowExpression rewrite(RowExpression expression, Session session)
         {
             // Rewrite RowExpression first to reduce depth of RowExpression tree by balancing AND/OR predicates.
             // It doesn't matter whether we rewrite/optimize first because this will be called by IterativeOptimizer.
             RowExpression rewritten = RowExpressionTreeRewriter.rewriteWith(logicalExpressionRewriter, expression, true);
-            RowExpression optimizedRowExpression = optimizer.optimize(rewritten, SERIALIZABLE, session);
-            return optimizedRowExpression;
+            if (isDelegatingRowExpressionOptimizerEnabled(session)) {
```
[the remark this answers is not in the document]
tdcmeehan: I think I can do that by following @ZacBlanco's feedback: https://github.com/prestodb/presto/pull/24144#discussion_r1890842187

presto-main/src/main/java/com/facebook/presto/server/testing/TestingPrestoServer.java
tdcmeehan: I believe the reason for this convention is these loading methods might entail something expensive, or that could fail, and it's helpful to do such things outside of Guice's bootstrapping process.  Such failures are often very verbose because Guice returns the error in the context of where it failed during the bootstrapping process, and I think it's cleaner to do it outside of the Guice dependency graph so that the failure is clear and isolated.

presto-main/src/main/java/com/facebook/presto/sql/expressions/ExpressionOptimizerManager.java
tdcmeehan: Yes, this will work.  Thanks for the suggestion @ZacBlanco.

It can't be a `Session` because these optimizers are used in connectors as well, and those connectors won't have access to `Session` (see `RowExpressionService`).  But we can pass in a `ConnectorSession`, and convert it back to a `FullConnectorSession` as a compromise.

This eliminates the branch in `SimplifyRowExpressions` and centralizes where one toggles the expression evaluation code.  I think it's cleaner this way in spite of the hack to convert `ConnectorSession` to `FullConnectorSession`.  It also guarantees that you use the right `ExpressionOptimizer` as long as you retrieve it from the appropriate interfaces (`RowExpressionService` in a connector, `ExpressionOptimizerManager` in the planner).  Let me know what you think of this approach.

presto-main/src/main/java/com/facebook/presto/sql/relational/DelegatingRowExpressionOptimizer.java
tdcmeehan: It's difficult as it stands because these two methods return different types.  I raised https://github.com/prestodb/presto/pull/24287 which would make it easier to share code between the two methods.

presto-tests/src/test/java/com/facebook/presto/tests/expressions/TestExpressionOptimizers.java
```diff
@@ -0,0 +1,133 @@
...
+        CallExpression predicate = call(
+                "LIKE",
+                resolution.likeVarcharFunction(),
+                BOOLEAN,
+                ImmutableList.of(
+                        new ConstantExpression(wrappedBuffer(value), VARCHAR),
+                        new ConstantExpression(castVarcharToLikePattern(utf8Slice(pattern)), LIKE_PATTERN)));
+        assertEquals(optimizeRowExpression(predicate, EVALUATED), expected);
+    }
+    @Override
```
pdabre12, line 74: Nit: Add an empty line.

presto-main/src/main/java/com/facebook/presto/sql/relational/ConnectorRowExpressionService.java
```diff
@@ -54,9 +55,9 @@ public DomainTranslator getDomainTranslator()
     }
 
     @Override
-    public ExpressionOptimizer getExpressionOptimizer()
+    public ExpressionOptimizer getExpressionOptimizer(ConnectorSession session)
```
pdabre12, line 58: It seems like this change is made in a later commit (https://github.com/prestodb/presto/pull/24144/commits/2d609d0aace7246a922029cf11ca16c72cc1f01b ), whereas the function signature in `RowExpressionService` was changed in prior commit. This will cause the build to fail.
Please make this change in a prior commit.

pdabre12 (commented):
Some comments,

presto-tests/src/test/java/com/facebook/presto/tests/expressions/TestExpressionOptimizers.java
```diff
@@ -0,0 +1,133 @@
...
+import static com.facebook.presto.type.LikePatternType.LIKE_PATTERN;
+import static io.airlift.slice.Slices.utf8Slice;
+import static io.airlift.slice.Slices.wrappedBuffer;
+import static org.testng.Assert.assertEquals;
+
+public class TestExpressionOptimizers
+        extends TestExpressions
+{
+    private final FunctionResolution resolution = new FunctionResolution(getMetadata().getFunctionAndTypeManager().getFunctionAndTypeResolver());
+    private ExpressionOptimizer expressionOptimizer;
```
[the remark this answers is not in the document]
tdcmeehan: This is constructed in the `setUp` method, so it can't be final.

presto-main/src/main/java/com/facebook/presto/sql/relational/DelegatingRowExpressionOptimizer.java
```diff
@@ -0,0 +1,80 @@
...
+    {
+        ExpressionOptimizer delegate = expressionOptimizerManager.getExpressionOptimizer(session);
+        Object currentExpression = rowExpression;
+        Object originalExpression;
+        for (int i = 0; i < maxOptimizationAttempts; i++) {
+            originalExpression = currentExpression;
+            currentExpression = delegate.optimize(toRowExpression(currentExpression, rowExpression.getType()), level, session, variableResolver);
+            if (currentExpression == null || currentExpression.equals(originalExpression)) {
```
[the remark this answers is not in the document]
tdcmeehan: It won't, it's removed in the recent refactoring.

presto-main/src/main/java/com/facebook/presto/sql/relational/DelegatingRowExpressionOptimizer.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.function.Function;
+
+import static com.facebook.presto.sql.planner.LiteralEncoder.toRowExpression;
+import static com.google.common.base.Preconditions.checkArgument;
+import static java.util.Objects.requireNonNull;
+
+public final class DelegatingRowExpressionOptimizer
+        implements ExpressionOptimizer
+{
+    private static final int DEFAULT_MAX_OPTIMIZATION_ATTEMPTS = 10;
```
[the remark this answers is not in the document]
tdcmeehan: There is a session property.  We actually don't need this constant because the session property's default is 10, I'll remove this.

tdcmeehan:
Updates:

1. Removed `DelegatingExpressionOptimizer` and all related tests.
2. Added the ability to configure multiple expression optimizers and switch them based on a session property.  Should be useful for fuzzer testing.

Thanks for the comments, please take another @ZacBlanco @rschlussel @aaneja

presto-main/src/main/java/com/facebook/presto/sql/relational/DelegatingRowExpressionOptimizer.java
```diff
@@ -0,0 +1,80 @@
...
+        return rowExpression;
+    }
+
+    @Override
+    public Object optimize(RowExpression rowExpression, Level level, ConnectorSession session, Function<VariableReferenceExpression, Object> variableResolver)
+    {
+        ExpressionOptimizer delegate = expressionOptimizerManager.getExpressionOptimizer(session);
+        Object currentExpression = rowExpression;
+        Object originalExpression;
+        for (int i = 0; i < maxOptimizationAttempts; i++) {
```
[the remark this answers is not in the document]
tdcmeehan: This was to guard against a poorly written expression optimizer, but I removed it, as in principle we can leave that responsibility to the plugin.

presto-main/src/main/java/com/facebook/presto/sql/expressions/ExpressionOptimizerManager.java
```diff
@@ -0,0 +1,155 @@
...
+    {
+        String name = getNameWithoutExtension(configurationFile.getName());
+        checkArgument(!isNullOrEmpty(name), "File name is empty, full path: %s", configurationFile.getAbsolutePath());
+        checkArgument(!name.equals(DEFAULT_OPTIMIZER_KEY), "Cannot name an expression optimizer instance default");
+
+        Map<String, String> properties = new HashMap<>(loadProperties(configurationFile));
+        String factoryName = properties.remove(EXPRESSION_MANAGER_FACTORY_NAME);
+        checkArgument(!isNullOrEmpty(factoryName), "%s does not contain %s", configurationFile, EXPRESSION_MANAGER_FACTORY_NAME);
+        checkArgument(expressionOptimizerFactories.containsKey(factoryName),
+                "ExpressionOptimizerFactory %s is not registered, registered factories: ", factoryName, expressionOptimizerFactories.keySet());
```
ZacBlanco, line 108: This code looks correct, but can we not re-use any of the existing tools from airlift's configuration module to load this? I feel like doing this manually seems more prone to error.

ZacBlanco (commented):
Two comments, otherwise lgtm

presto-main/src/main/java/com/facebook/presto/sql/expressions/ExpressionOptimizerManager.java
tdcmeehan: We do leverage airlift configuration tools, but within the plugin.  At this layer, we need more flexibility than I think those tools can provide: we need to check if the plugin is the default plugin, and we need to check the factory name to see if it matches with any of the plugins that have been discovered.  These are removed before being sent to the plugin because they're not plugin-specific properties, but once passed to the plugin they will be loaded via airlift as that's the convention established by other plugins (including connectors).

I agree it's a bit brittle with all the `checkArgument`s, but I feel like the unit tests are the best way to ensure this is right, while making sure we don't pass useless properties to the plugins.  Let me know if you can think of a better way to go about it that I didn't think of.

presto-main/src/test/java/com/facebook/presto/sql/expressions/TestExpressionOptimizerManager.java
```diff
@@ -0,0 +1,184 @@
...
+
+        assertOptimizedExpression("1+1", "2", ImmutableMap.of());
+
+        // Override the default optimizer based on session property
+        assertOptimizedExpression("1+1", "'foo'", ImmutableMap.of("expression_optimizer_instance", "foo"));
+        assertOptimizedExpression("1+1", "'bar'", ImmutableMap.of("expression_optimizer_instance", "bar"));
+    }
+
+    @Test
+    public void testDefaultMayBeConfigured()
```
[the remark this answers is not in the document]
tdcmeehan: Thanks, added a test showing that it's not allowed.

presto-main/src/main/java/com/facebook/presto/sql/expressions/ExpressionOptimizerManager.java
```diff
@@ -0,0 +1,152 @@
...
+                expressionOptimizerFactories.putIfAbsent(name, expressionOptimizerFactory) == null,
+                "ExpressionOptimizerFactory %s is already registered", name);
+    }
+
+    public ExpressionOptimizer getExpressionOptimizer(ConnectorSession connectorSession)
+    {
+        checkArgument(connectorSession instanceof FullConnectorSession, "connectorSession is not an instance of FullConnectorSession");
+        Session session = ((FullConnectorSession) connectorSession).getSession();
+        String expressionOptimizerInstance = getExpressionOptimizerInstance(session);
+        if (isNullOrEmpty(expressionOptimizerInstance)) {
```
rschlussel, line 138: I feel like this logic is unnecessarily fragmented. Can we instead have the session property backed by a config property (like other session properties/configs) and have "default" as the default value for the config? That will also let people set desired configurations for the whole cluster.

presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
```diff
@@ -327,6 +327,7 @@ public final class SystemSessionProperties
     public static final String INLINE_PROJECTIONS_ON_VALUES = "inline_projections_on_values";
     public static final String INCLUDE_VALUES_NODE_IN_CONNECTOR_OPTIMIZER = "include_values_node_in_connector_optimizer";
     public static final String SINGLE_NODE_EXECUTION_ENABLED = "single_node_execution_enabled";
+    public static final String EXPRESSION_OPTIMIZER_INSTANCE = "expression_optimizer_instance";
```
rschlussel, line 330: maybe call this "expression_optimizer_name"? instance makes it sound like we're providing the object itself.

presto-main/src/main/java/com/facebook/presto/sql/expressions/ExpressionOptimizerManager.java
```diff
@@ -0,0 +1,152 @@
...
+    {
+        String name = expressionOptimizerFactory.getName();
+        checkArgument(
+                expressionOptimizerFactories.putIfAbsent(name, expressionOptimizerFactory) == null,
+                "ExpressionOptimizerFactory %s is already registered", name);
+    }
+
+    public ExpressionOptimizer getExpressionOptimizer(ConnectorSession connectorSession)
+    {
+        checkArgument(connectorSession instanceof FullConnectorSession, "connectorSession is not an instance of FullConnectorSession");
```
rschlussel, line 135: i don't like assuming that a session is a connectorSession. it breaks the abstractions.  In this case all we care about is the session property. can we instead add a getSystemProperty() method to ConnectorSession?

tdcmeehan: While I agree this is not ideal, I'm of the opinion that it's worse to expose `getSystemProperty` to the SPI, because this allows all plugins to access system properties that were previously only known to the engine or runtime.  While not ideal, I think this approach is a bit less worse because it's within `presto-main`, and we have examples of this being done for system connector and metadata tables provided to the JDBC driver.  I do think that we should probably think of a better abstraction for how we provide session properties to plugins which are not connectors, and I've added a TODO and will open a Github issue to track it.  Let me know if you think this is reasonable.

presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
tdcmeehan: Agreed, done.

presto-main/src/main/java/com/facebook/presto/sql/expressions/ExpressionOptimizerManager.java
tdcmeehan: Good suggestion, done.

presto-main/src/main/java/com/facebook/presto/sql/expressions/ExpressionOptimizerManager.java
rschlussel: Fair criticism and suggestion to punt on resolving this.  i do feel like the problem i'm talking about isn't about plugins that aren't connectors, it's easy enough to implement passing them ConnectorSessions. The thing that makes this require a ConnectorSession vs. a Session is that it gets called from connector optimizers.  and we had similar issue with stats stuff a while back.  The plugin separation break down when connectors call into stuff from presto-main.

presto-main/src/main/java/com/facebook/presto/sql/relational/RowExpressionOptimizer.java
```diff
@@ -24,31 +26,38 @@
...
     {
-        this.metadata = requireNonNull(metadata, "metadata is null");
+        this(requireNonNull(metadata, "metadata is null").getFunctionAndTypeManager());
+    }
+
+    public RowExpressionOptimizer(FunctionMetadataManager functionMetadataManager)
+    {
+        checkArgument(functionMetadataManager instanceof FunctionAndTypeManager, "Expected functionMetadataManager to be instance of FunctionAndTypeManager");
+        this.functionAndTypeManager = (FunctionAndTypeManager) requireNonNull(functionMetadataManager, "functionMetadataManager is null");
     }
```
pdabre12, line 46: This check seems unnecessary,

```suggestion

    public RowExpressionOptimizer(FunctionAndTypeManager functionAndTypeManager)
    {
        this.functionAndTypeManager =  requireNonNull(functionAndTypeManager, "functionAndTypeManager is null");
    }
```

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/CteProjectionAndPredicatePushDown.java
```diff
@@ -87,10 +89,14 @@ public class CteProjectionAndPredicatePushDown
         implements PlanOptimizer
 {
     private final Metadata metadata;
+    private final ExpressionOptimizerManager expressionOptimizerManager;
+    private final FeaturesConfig featuresConfig;
 
-    public CteProjectionAndPredicatePushDown(Metadata metadata)
```
pdabre12, line 91: `featuresConfig` does not seem to be used, can we remove it?

pdabre12 (commented):
Just two nits, otherwise LGTM.

presto-main/src/main/java/com/facebook/presto/sql/relational/RowExpressionOptimizer.java
tdcmeehan: Good catch, done.

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/CteProjectionAndPredicatePushDown.java
tdcmeehan: Yes, thanks!

presto-main/src/main/java/com/facebook/presto/sql/expressions/ExpressionOptimizerManager.java
```diff
@@ -0,0 +1,156 @@
...
+        String defaultPropertyValue = properties.remove(DEFAULT_PROPERTY);
+        checkArgument(defaultPropertyValue == null || "true".equalsIgnoreCase(defaultPropertyValue) || "false".equalsIgnoreCase(defaultPropertyValue),
+                "Only true or false is allowed for %s, found: %s", DEFAULT_PROPERTY, defaultPropertyValue);
+        boolean isDefault = parseBoolean(defaultPropertyValue);
+
+        ExpressionOptimizer optimizer = expressionOptimizerFactories.get(factoryName).createOptimizer(
+                properties,
+                new ExpressionOptimizerContext(nodeManager, functionAndTypeManager, functionResolution));
+        expressionOptimizers.put(name, optimizer);
+        if (isDefault) {
```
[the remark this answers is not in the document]
tdcmeehan: Originally I was trying to make it so just installing the plugin could easily configure the right default value.  But I'll take that up in a new PR, I've removed the `default` flag.

rschlussel (approved):

aditi-pandit (approved):
+1 for presto-native-execution changes.

PR #24149, opened 2024-11-26
[native] Add LinuxMemoryChecker check/warning to ensure system-mem-limit-gb is reasonably set

minhancao:
## Description

Add additional checks and warnings to ensure 
system-memory-gb <= system-mem-limit-gb < available machine memory of deployment.

For cgroup v1:
Set available machine memory of deployment to be the smaller number
between /proc/meminfo and memory.limit_in_bytes.

For cgroup v2:
Set available machine memory of deployment to be the smaller number
between /proc/meminfo and memory.max.
If memory.max contains "max" string, then look at
/proc/meminfo for the MemTotal, otherwise use the
value in memory.max.

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

Prestissimo (Native Execution)
* Replace default values for configuration properties ``system-memory-gb`` to 57 
  and ``system-mem-limit-gb`` to 60.

Documentation
* Replace default values for configuration properties ``system-memory-gb`` to 57 
  and ``system-mem-limit-gb`` to 60.
```


presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp
```diff
@@ -49,6 +49,63 @@ class LinuxMemoryChecker : public PeriodicMemoryChecker {
     LOG(INFO) << fmt::format("Changed to using memory stat file {}", statFile_);
   }
 
+  void start() {
+    // Set memMaxFile to "/sys/fs/cgroup/memory/memory.limit_in_bytes" for
+    // cgroup v1 or "sys/fs/cgroup/memory.max" for cgroup v2.
+    struct stat buffer;
```
czentgr, line 55: Do we need to re-declare this as a struct here?

presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp
```diff
@@ -49,6 +49,63 @@ class LinuxMemoryChecker : public PeriodicMemoryChecker {
...
+    // Check system-memory-gb < system-mem-limit-gb < actual total memory
+    // capacity.
+
+    // For cgroup v1:
+    // Set actual total memory to be the smaller number between /proc/meminfo
+    // and memory.limit_in_bytes
+    int64_t actualTotalMemory = 0;
+    folly::gen::byLine("/proc/meminfo") |
+        [&](const folly::StringPiece& line) -> void {
+      if (actualTotalMemory == 0) {
```
czentgr, line 75: We don't need this check because we return immediately if it is set on line 79.

presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp
```diff
@@ -49,6 +49,63 @@ class LinuxMemoryChecker : public PeriodicMemoryChecker {
...
+        if (line == "max") {
+          return;
+        }
+        actualTotalMemory =
+            std::min(actualTotalMemory, folly::to<int64_t>(line));
+        return;
+      };
+    }
+
+    VELOX_CHECK_LT(
```
czentgr, line 95: We should allow less or equal.

presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp
```diff
@@ -49,6 +49,63 @@ class LinuxMemoryChecker : public PeriodicMemoryChecker {
...
+        [&](const folly::StringPiece& line) -> void {
+      if (actualTotalMemory == 0) {
+        actualTotalMemory = static_cast<int64_t>(
+            extractNumericConfigValueWithRegex(line, kMemTotalRegex) * 1024);
+      }
+      if (actualTotalMemory != 0) {
+        return;
+      }
+    };
+    if (memMaxFile != "None") {
```
czentgr, line 83: We might need to add a comment here as to what the content could be for the maxMemFile. Since we check for the "max" keyword too.

czentgr (commented):
Can we add a test with fake files again just like we did with the original tests for this class?
That way we can try the "max" value for cgv2, and gigantic values and reasonable values. Basically testing the various situations we saw when investigating this.

presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp
minhancao: Tried declaring it as `stat buffer;` and got error:
```
/root/presto/presto-native-execution/./presto_cpp/main/LinuxMemoryChecker.cpp:30:9: error: expected ‘;’ before ‘buffer’
   30 |     stat buffer;
```

presto-native-execution/presto_cpp/main/tests/LinuxMemoryCheckerTest.cpp
```diff
@@ -33,10 +33,20 @@ class LinuxMemoryCheckerTest : public testing::Test {
 
   LinuxMemoryCheckerTest() : memChecker(PeriodicMemoryChecker::Config{}) {}
```
pramodsatya, line 34: Nit: not related to this PR but `memChecker` should probably be renamed to `memChecker_` as per the coding conventions.

presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp
```diff
@@ -49,6 +49,78 @@ class LinuxMemoryChecker : public PeriodicMemoryChecker {
...
+    if (config_.systemMemLimitBytes < systemConfig->systemMemoryGb()) {
+      LOG(WARNING) << "system-mem-limit-gb is smaller than system-memory-gb. "
+                   << "Expected: system-mem-limit-gb >= system-memory-gb.";
+    }
+
+    PeriodicMemoryChecker::start();
+  }
+
+  int64_t getActualTotalMemory() {
+    // Set actual total memory to be the smaller number between /proc/meminfo
```
pramodsatya, line 95: Should we add `// For cgroup v1:` in the comment as before?

presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp
```diff
@@ -143,10 +215,15 @@ class LinuxMemoryChecker : public PeriodicMemoryChecker {
...
   const boost::regex kMemAvailableRegex{R"!(MemAvailable:\s*(\d+)\s*kB)!"};
-  const boost::regex kMemTotalRegex{R"!(MemTotal:\s*(\d+)\s*kB)!"};
+  const boost::regex kMemTotalRegex{R"!(MemTotal:\s*(\d+)\s+kB)!"};
   const char* kCgroupV1Path = "/sys/fs/cgroup/memory/memory.stat";
   const char* kCgroupV2Path = "/sys/fs/cgroup/memory.stat";
+  const char* kCgroupV1MaxMemFilePath =
+      "/sys/fs/cgroup/memory/memory.limit_in_bytes";
+  const char* kCgroupV2MaxMemFilePath = "/sys/fs/cgroup/memory.max";
   std::string statFile_;
+  std::string memInfoFile_ = "/proc/meminfo";
```
pramodsatya, line 225: Nit: `const std::string kMemInfoFile_`

pramodsatya (commented):
Thanks @minhancao, could you please squash the commits?

presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp
minhancao: Both cgroup v1 and cgroup v2 will use the MemTotal number in /proc/meminfo if it is smaller than their respective memory.limit_in_bytes or memory.max value. 

Added more comments in the function for cgroup v1 to explain what's going on when extracting the value from memory.limit_in_bytes.

presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp
minhancao: This has to stay as a non-const variable since I need to change its path to point to the meminfo test file when I am running the LinuxMemoryCheckerTests.

majetideepak (changes requested):
It is ideal if we can avoid checking in data files for testing.
We only need a few fields from the file for testing.
Can we write these required fields to a temporary file as part of the testing?

presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp
```diff
@@ -49,6 +62,73 @@ class LinuxMemoryChecker : public PeriodicMemoryChecker {
...
+    LOG(INFO) << fmt::format("Changed to using meminfo file {}", memInfoFile_);
+  }
+
+  void start() {
+    // Check system-memory-gb < system-mem-limit-gb < actual total memory
+    // capacity.
+    VELOX_CHECK_LE(
+        config_.systemMemLimitBytes,
+        getActualTotalMemory(),
+        "system-mem-limit-gb is higher than the actual total memory capacity.");
```
czentgr, line 82: Can we add the actual numbers that were found if this fails? Especially because the total memory capacity doesn't come from the config but is determined.

presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp
```diff
@@ -49,6 +62,73 @@ class LinuxMemoryChecker : public PeriodicMemoryChecker {
...
+        return;
+      }
+      actualTotalMemory = static_cast<int64_t>(
+          extractNumericConfigValueWithRegex(line, kMemTotalRegex) * 1024);
+    };
+
+    // For cgroup v1, memory.limit_in_bytes can default to a really big numeric
+    // value in bytes like 9223372036854771712 to represent that
+    // memory.limit_in_bytes is not set to a value. The default value here is
+    // set to PAGE_COUNTER_MAX, which is LONG_MAX/PAGE_SIZE on 64-bit platform.
```
czentgr, line 109: Nit: "on the 64-bit platform".

presto-native-execution/presto_cpp/main/tests/examples/cgroupV1memoryNotSet.limit_in_bytes
```diff
@@ -0,0 +1 @@
+9223372036854771712
```
czentgr, line 1: As discussed, lets not add example files but generate the content in memory.
The logic relies on reading files. So we could write to the temp dir path which is used in other tests and use those files for the test.

presto-native-execution/presto_cpp/main/tests/LinuxMemoryCheckerTest.cpp
```diff
@@ -60,28 +131,201 @@ TEST_F(LinuxMemoryCheckerTest, basic) {
       LinuxMemoryChecker(PeriodicMemoryChecker::Config{
           1'000, true, 1024, 32, true, 5, "/path/to/dir", "", 5, 512}),
       "heapDumpFilePrefix cannot be empty when heap dump is enabled.");
-  LinuxMemoryChecker memChecker(PeriodicMemoryChecker::Config{
+  LinuxMemoryChecker memChecker_(PeriodicMemoryChecker::Config{
```
czentgr, line 134: Nit: this is a local variable and not a member so should be `memChecker`.

presto-native-execution/presto_cpp/main/tests/LinuxMemoryCheckerTest.cpp
```diff
@@ -17,29 +17,100 @@
...
-    ASSERT_EQ(memChecker.getUsedMemory(), expectedMemoryUsage);
+    auto tempTestFile = exec::test::TempFilePath::create();
+    tempTestFile->append(content);
+    auto testFilePath = tempTestFile->getPath();
+
+    memChecker_.setStatFile(testFilePath);
+    ASSERT_EQ(memChecker_.getUsedMemory(), expectedMemoryUsage);
   }
+
+  std::string memInfoText =
```
czentgr, line 56: Nit: this is a const member so it should be named: `const std::string kMemInfoText_`.

minhancao:
@czentgr @majetideepak @pramodsatya 
I have addressed all the PR comments, please review this PR when you can, thank you!

presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp
```diff
@@ -49,6 +73,87 @@ class LinuxMemoryChecker : public PeriodicMemoryChecker {
     LOG(INFO) << fmt::format("Changed to using memory stat file {}", statFile_);
   }
 
+  void setMemMaxFile(std::string memMaxFile) {
```
czentgr, line 76: Use const & for the parameter.

presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp
```diff
@@ -49,6 +73,87 @@ class LinuxMemoryChecker : public PeriodicMemoryChecker {
...
+    memMaxFile_ = memMaxFile;
+    LOG(INFO) << fmt::format(
+        "Changed to using memory max file {}", memMaxFile_);
+  }
+
+  void setMemInfo(std::string memInfoFile) {
```
czentgr, line 82: Use const & for the parameter.

presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp
```diff
@@ -49,6 +73,87 @@ class LinuxMemoryChecker : public PeriodicMemoryChecker {
...
+    // Check system-memory-gb < system-mem-limit-gb < memory limit for process.
+    auto* systemConfig = SystemConfig::instance();
+    int64_t systemMemoryInBytes = systemConfig->systemMemoryGb() << 30;
+    PRESTO_STARTUP_LOG(INFO)
+        << fmt::format("System memory in bytes: {}", systemMemoryInBytes);
+
+    PRESTO_STARTUP_LOG(INFO) << fmt::format(
+        "System memory limit in bytes: {}", config_.systemMemLimitBytes);
+
+    int64_t memoryLimitForProcess = getMemoryLimitForProcess();
```
czentgr, line 97: We can use `auto` here.

presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp
```diff
@@ -83,6 +188,10 @@ class LinuxMemoryChecker : public PeriodicMemoryChecker {
     if (statFile_ != "None") {
       folly::gen::byLine(statFile_.c_str()) |
           [&](const folly::StringPiece& line) -> void {
+        if (activeAnon != 0 && inactiveAnon != 0) {
```
czentgr, line 191: Why did we move this up? This should be false on the first try because the variables are just initialized above to 0.

presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp
```diff
@@ -92,19 +201,19 @@ class LinuxMemoryChecker : public PeriodicMemoryChecker {
...
 
       // Unit is in bytes.
       return inactiveAnon + activeAnon;
     }
 
     // Last resort use host machine info.
-    folly::gen::byLine("/proc/meminfo") |
+    folly::gen::byLine(memInfoFile_.c_str()) |
         [&](const folly::StringPiece& line) -> void {
+      if (memAvailable != 0 && memTotal != 0) {
```
czentgr, line 213: Same question, why was this moved here?

presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp
```diff
@@ -30,12 +31,35 @@ class LinuxMemoryChecker : public PeriodicMemoryChecker {
...
+    }
+
+    // Set memMaxFile_ to:
+    // "/sys/fs/cgroup/memory/memory.limit_in_bytes" for cgroup v1
+    // or
+    // "/sys/fs/cgroup/memory.max" for cgroup v2.
+    if ((stat(kCgroupV1MaxMemFilePath, &buffer) == 0)) {
+      memMaxFile_ = kCgroupV1MaxMemFilePath;
+      PRESTO_STARTUP_LOG(INFO)
+          << fmt::format("Using cgroup v1 memory max file {}", memMaxFile_);
```
majetideepak, line 53: Is it possible to have cgroup v2 memory stat file along with group v1 memory max file?

presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp
majetideepak, line 82: setMemInfoFile

presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp
```diff
@@ -49,6 +73,87 @@ class LinuxMemoryChecker : public PeriodicMemoryChecker {
...
+    memInfoFile_ = memInfoFile;
+    LOG(INFO) << fmt::format("Changed to using meminfo file {}", memInfoFile_);
+  }
+
+  void start() override {
+    // Check system-memory-gb < system-mem-limit-gb < memory limit for process.
+    auto* systemConfig = SystemConfig::instance();
+    int64_t systemMemoryInBytes = systemConfig->systemMemoryGb() << 30;
+    PRESTO_STARTUP_LOG(INFO)
+        << fmt::format("System memory in bytes: {}", systemMemoryInBytes);
```
majetideepak, line 92: Are these not logged during startup from the config.properties file?

presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp
```diff
@@ -49,6 +73,87 @@ class LinuxMemoryChecker : public PeriodicMemoryChecker {
...
+
+    PRESTO_STARTUP_LOG(INFO) << fmt::format(
+        "System memory limit in bytes: {}", config_.systemMemLimitBytes);
+
+    int64_t memoryLimitForProcess = getMemoryLimitForProcess();
+    PRESTO_STARTUP_LOG(INFO) << fmt::format(
+        "Memory limit for process in bytes: {}", memoryLimitForProcess);
+
+    VELOX_CHECK_LE(
```
majetideepak, line 101: I think these checks are sufficient. No need to log above.

presto-native-execution/presto_cpp/main/tests/LinuxMemoryCheckerTest.cpp
```diff
@@ -17,29 +17,102 @@
...
+    auto tempTestFile = exec::test::TempFilePath::create();
+    tempTestFile->append(content);
+    auto testFilePath = tempTestFile->getPath();
+
+    memChecker_.setStatFile(testFilePath);
+    ASSERT_EQ(memChecker_.getUsedMemory(), expectedMemoryUsage);
   }
+
+  const std::string kMemInfoText_ =
+      "MemTotal:       129294272 kB\n"
```
majetideepak, line 59: Do we need all these entries?

presto-native-execution/presto_cpp/main/tests/LinuxMemoryCheckerTest.cpp
```diff
@@ -68,20 +141,193 @@ TEST_F(LinuxMemoryCheckerTest, basic) {
...
+      "pgpgin 97614\n"
+      "pgpgout 87091\n"
+      "pgfault 55869\n"
+      "pgmajfault 132\n"
+      "inactive_anon 1486848\n"
+      "active_anon 3649536\n"
+      "inactive_file 20410368\n"
+      "active_file 11894784\n"
+      "unevictable 5406720\n"
+      "hierarchical_memory_limit 9223372036854771712\n"
```
majetideepak, line 251: Do we need all these entries?

presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp
minhancao: Moved back down to be the last if condition

presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp
minhancao: Moved back down to be the last if condition

presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp
minhancao: No, that combination is not possible.

presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp
minhancao: These are indeed logged by showing what's in the config.properties file during startup. 

I still think having this log here in LinuxMemoryChecker.cpp is beneficial since it'd be easier to visually see the configs in bytes and make sure it follows this condition: system-memory-gb < system-mem-limit-gb < memory limit for process.

presto-native-execution/presto_cpp/main/tests/LinuxMemoryCheckerTest.cpp
minhancao: Yes, I think these entries are necessary since we are trying to simulate actually reading through the /proc/meminfo file and parsing out the necessary values from it.

presto-native-execution/presto_cpp/main/tests/LinuxMemoryCheckerTest.cpp
minhancao: Yes, I think these entries are necessary since we are trying to simulate actually reading through the memory.stat file and parsing out the necessary values from it.

presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp
minhancao: I still think having logging here of these 3 configs in LinuxMemoryChecker.cpp is beneficial since it'd be easier to visually see the configs in bytes and make sure it follows this condition: system-memory-gb < system-mem-limit-gb < memory limit for process.

@czentgr Wdyt?

czentgr: @majetideepak I've asked Minhan to add these. They are only logged once at the beginning and gives us a better view of the environment in case we do see OOMs. So I think this will be beneficial in case we need to look at the logs.

presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp
majetideepak: I feel we could then combine both V1 parts in one block and V2 parts in another block.

minhancao: Combined them as suggested:
```
if ((stat(kCgroupV1Path, &buffer) == 0)) {
      statFile_ = kCgroupV1Path;
      memMaxFile_ = kCgroupV1MaxMemFilePath;
      PRESTO_STARTUP_LOG(INFO) << "Using cgroup v1.";
    } else if ((stat(kCgroupV2Path, &buffer) == 0)) {
      statFile_ = kCgroupV2Path;
      memMaxFile_ = kCgroupV2MaxMemFilePath;
      PRESTO_STARTUP_LOG(INFO) << "Using cgroup v2.";
    } else {
      statFile_ = "None";
      memMaxFile_ = "None";
      PRESTO_STARTUP_LOG(INFO)
          << "Could not find cgroup in the Linux system.";
    }

    PRESTO_STARTUP_LOG(INFO)
        << fmt::format("Memory stat file: {}", statFile_);
    PRESTO_STARTUP_LOG(INFO)
        << fmt::format("Memory max file {}", memMaxFile_);
        ```

Comment (minhancao) [reply presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp:53]: I think we should leave this as it is. 

The reason being this scenario is possible. In a cgroup v2 Linux machine it was able to find the memory.stat file but unable to find the memory.max file due to "slices" existing on the host machine:
statFile_ = "/sys/fs/cgroup/memory.stat"
memMaxFile_ = "None"

When you're inside a container using cgroup v1 or v2, you won't see "slices" directly because each container gets its own isolated cgroup hierarchy, meaning the slices defined on the host system aren't visible within the container; instead, the container will have its own "scope" which acts as a separate cgroup structure under the parent container slice, effectively creating a nested hierarchy where you manage resources within the container's isolated space. This is why we're able to find it easily in a container with the constant paths that we defined:
```
const char* kCgroupV1MaxMemFilePath = "/sys/fs/cgroup/memory/memory.limit_in_bytes";
const char* kCgroupV2MaxMemFilePath = "/sys/fs/cgroup/memory.max";
```

However, on a host machine notice the "system.slice" and "user.slice" directories below. In cgroup, slices are a way to group resources for memory management.
```
root@biday3ember11:/sys/fs/cgroup# ls
cgroup.controllers      cpu.stat             memory.pressure
cgroup.max.depth        dev-hugepages.mount  memory.stat
cgroup.max.descendants  dev-mqueue.mount     misc.capacity
cgroup.procs            init.scope           proc-sys-fs-binfmt_misc.mount
cgroup.stat             io.cost.model        sys-fs-fuse-connections.mount
cgroup.subtree_control  io.cost.qos          sys-kernel-config.mount
cgroup.threads          io.pressure          sys-kernel-debug.mount
cpu.pressure            io.prio.class        sys-kernel-tracing.mount
cpuset.cpus.effective   io.stat              system.slice
cpuset.mems.effective   memory.numa_stat     user.slice
```
```
root@biday3ember11:/sys/fs/cgroup/system.slice# ls | grep "memory.max"
memory.max
```

**Comment by minhancao** [comment]
@majetideepak I have updated the PR with some new changes, please review when you can, thank you!

File: presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp (hunk 18 above)
Comment (minhancao) [reply presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp]: Setting the memMaxFile_ and memStatFile_ should both have their own if block to check because it's possible for one of them to exist and the other does not on a Linux machine.

File: presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp [hunk 23]
```diff
@@ -45,8 +65,90 @@ class LinuxMemoryChecker : public PeriodicMemoryChecker {
...
 
   void setStatFile(std::string statFile) {
-    statFile_ = statFile;
-    LOG(INFO) << fmt::format("Changed to using memory stat file {}", statFile_);
+    memStatFile_ = statFile;
+    LOG(INFO) << fmt::format(
+        "Changed to using memory stat file {}", memStatFile_);
+  }
+
+  void setMemMaxFile(const std::string& memMaxFile) {
```
majetideepak, line 73: Ideally, this is only used for testing. Let's comment that this is used for testing.

presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp
```diff
@@ -45,8 +65,90 @@ class LinuxMemoryChecker : public PeriodicMemoryChecker {
...
+    memMaxFile_ = memMaxFile;
+    LOG(INFO) << fmt::format(
```
majetideepak, line 75: Do we need this log for tests?

presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp
```diff
@@ -45,8 +65,90 @@ class LinuxMemoryChecker : public PeriodicMemoryChecker {
...
+        "Changed to using memory max file {}", memMaxFile_);
+  }
+
+  void setMemInfoFile(const std::string& memInfoFile) {
```
majetideepak, line 79: same as above. Mark as testing API.

presto-native-execution/presto_cpp/main/tests/LinuxMemoryCheckerTest.cpp
```diff
@@ -17,29 +17,102 @@
...
+  void checkMemoryLimitForProcess(
+      const std::string& content,
+      int64_t expectedMemoryMax) {
+    auto tempMemInfoFile = exec::test::TempFilePath::create();
+    tempMemInfoFile->append(kMemInfoText_);
+    auto memInfoPath = tempMemInfoFile->getPath();
 
-  LinuxMemoryCheckerTest() : memChecker(PeriodicMemoryChecker::Config{}) {}
+    auto tempTestFile = exec::test::TempFilePath::create();
+    tempTestFile->append(content);
```
majetideepak, line 39: tempMemMaxFile

presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp
```diff
@@ -45,8 +65,90 @@ class LinuxMemoryChecker : public PeriodicMemoryChecker {
...
+        [&](const folly::StringPiece& line) -> void {
+      if (memoryLimitForProcess != 0) {
+        return;
+      }
+      memoryLimitForProcess = static_cast<int64_t>(
+          extractNumericConfigValueWithRegex(line, kMemTotalRegex) * 1024);
+    };
+
+    // For cgroup v1, memory.limit_in_bytes can default to a really big numeric
+    // value in bytes like 9223372036854771712 to represent that
```
majetideepak, line 128: can we not use `int128_t` for these big values?

presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp
```diff
@@ -45,8 +65,90 @@ class LinuxMemoryChecker : public PeriodicMemoryChecker {
...
+    // memory.limit_in_bytes is not set to a value. The default value here is
```
majetideepak, line 129: Where is this default value of `PAGE_COUNTER_MAX` set?

presto-native-execution/presto_cpp/main/tests/LinuxMemoryCheckerTest.cpp
```diff
@@ -68,20 +141,192 @@ TEST_F(LinuxMemoryCheckerTest, basic) {
...
 }
 
+TEST_F(LinuxMemoryCheckerTest, sysMemLimitBytesCheck) {
+  auto tempMemInfoFile = exec::test::TempFilePath::create();
+  tempMemInfoFile->append(kMemInfoText_);
+  auto memInfoPath = tempMemInfoFile->getPath();
+
+  auto tempTestFile = exec::test::TempFilePath::create();
+  tempTestFile->append("131000000000\n");
+  auto testFilePath = tempTestFile->getPath();
```
majetideepak, line 151: memMaxFilePath?

presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp
minhancao: In the Linux kernel source code.

https://github.com/torvalds/linux/blob/master/include/linux/page_counter.h#L47

presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp
minhancao: Changing to use int128_t instead of int64_t

minhancao:
@czentgr @majetideepak Just updated the PR, please review when you can!

steveburnett (dismissed):
LGTM! (docs)

Pull branch, new doc build, looks good. Thanks!

minhancao:
Confirmed works on cgroup v1 machine:
```
I0211 09:34:08.711416 489817 LinuxMemoryChecker.cpp:35] [PRESTO_STARTUP] Using cgroup v1.
I0211 09:34:08.711459 489817 LinuxMemoryChecker.cpp:55] [PRESTO_STARTUP] Using memory stat file: /sys/fs/cgroup/memory/memory.stat
I0211 09:34:08.711473 489817 LinuxMemoryChecker.cpp:58] [PRESTO_STARTUP] Using memory max file /sys/fs/cgroup/memory/memory.limit_in_bytes
I0211 09:34:08.711670 489817 LinuxMemoryChecker.cpp:90] [PRESTO_STARTUP] System memory in bytes: 2147483648
I0211 09:34:08.711715 489817 LinuxMemoryChecker.cpp:93] [PRESTO_STARTUP] System memory limit in bytes: 64424509440
I0211 09:34:08.711890 489817 LinuxMemoryChecker.cpp:97] [PRESTO_STARTUP] Available machine memory of deployment in bytes: 8331362304
E0211 09:34:08.711936 489817 Exceptions.h:66] Line: /root/presto/presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp:101, Function:start, Expression: config_.systemMemLimitBytes <= availableMemoryOfDeployment (64424509440 vs. 8331362304) system memory limit = 64424509440 bytes is higher than the available machine memory of deployment = 8331362304 bytes., Source: RUNTIME, ErrorCode: INVALID_STATE
terminate called after throwing an instance of 'facebook::velox::VeloxRuntimeError'
  what():  Exception: VeloxRuntimeError
Error Source: RUNTIME
Error Code: INVALID_STATE
Reason: (64424509440 vs. 8331362304) system memory limit = 64424509440 bytes is higher than the available machine memory of deployment = 8331362304 bytes.
Retriable: False
Expression: config_.systemMemLimitBytes <= availableMemoryOfDeployment
Function: start
File: /root/presto/presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp
Line: 101
Stack trace:
# 0  _ZN8facebook5velox7process10StackTraceC1Ei
# 1  _ZN8facebook5velox14VeloxException5State4makeIZNS1_C4EPKcmS5_St17basic_string_viewIcSt11char_traitsIcEES9_S9_S9_bNS1_4TypeES9_EUlRT_E_EESt10shared_ptrIKS2_ESA_SB_
# 2  _ZN8facebook5velox14VeloxExceptionC1EPKcmS3_St17basic_string_viewIcSt11char_traitsIcEES7_S7_S7_bNS1_4TypeES7_
# 3  _ZN8facebook5velox17VeloxRuntimeErrorC1EPKcmS3_St17basic_string_viewIcSt11char_traitsIcEES7_S7_S7_bS7_
# 4  _ZN8facebook5velox6detail14veloxCheckFailINS0_17VeloxRuntimeErrorERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEEEvRKNS1_18VeloxCheckFailArgsET0_
# 5  _ZN8facebook6presto18LinuxMemoryChecker5startEv
# 6  _ZN8facebook6presto12PrestoServer28addMemoryCheckerPeriodicTaskEv
# 7  _ZN8facebook6presto12PrestoServer3runEv
# 8  main
# 9  __libc_start_main
# 10 _start

*** Aborted at 1739295248 (Unix time, try 'date -d @1739295248') ***
*** Signal 6 (SIGABRT) (0x77959) received by PID 489817 (pthread TID 0x7fda627b8e80) (linux TID 489817) (maybe from PID 489817, UID 0) (code: -6), stack trace: ***
    @ 000000000aca5db1 _ZN5folly10symbolizer12_GLOBAL__N_113signalHandlerEiP9siginfo_tPv
                       /root/presto/presto-native-execution/dependencies/deps-download/folly/folly/experimental/symbolizer/SignalHandler.cpp:453
    @ 000000000001441f (unknown)
    @ 000000000004300b gsignal
    @ 0000000000022858 abort
    @ 00000000000a4ee5 (unknown)
    @ 00000000000b6f8b (unknown)
    @ 00000000000b6ff6 _ZSt9terminatev
    @ 00000000000b7257 __cxa_throw
    @ 000000000ac4809a __cxa_throw
                       /root/presto/presto-native-execution/dependencies/deps-download/folly/folly/debugging/exception_tracer/ExceptionTracerLib.cpp:159
    @ 000000000ab4659d _ZN8facebook5velox6detail14veloxCheckFailINS0_17VeloxRuntimeErrorERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEEEvRKNS1_18VeloxCheckFailArgsET0_
                       /root/presto/presto-native-execution/velox/velox/common/base/Exceptions.h:74
                       -> /root/presto/presto-native-execution/velox/velox/common/base/Exceptions.cpp
    @ 000000000099ec70 _ZN8facebook6presto18LinuxMemoryChecker5startEv
                       /root/presto/presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp:101
    @ 0000000000ccb692 _ZN8facebook6presto12PrestoServer28addMemoryCheckerPeriodicTaskEv
                       /root/presto/presto-native-execution/presto_cpp/main/PrestoServer.cpp:1044
    @ 0000000000cc6fd2 _ZN8facebook6presto12PrestoServer3runEv
                       /root/presto/presto-native-execution/presto_cpp/main/PrestoServer.cpp:552
    @ 00000000009e137c main
                       /root/presto/presto-native-execution/presto_cpp/main/PrestoMain.cpp:30
    @ 0000000000024082 __libc_start_main
    @ 00000000007c909d _start
Fatal signal handler. ThreadDebugInfo object not found.
Aborted (core dumped)
```

minhancao:
Confirmed works on cgroup v2 machine:
```
I0211 09:40:50.499085  4661 LinuxMemoryChecker.cpp:46] [PRESTO_STARTUP] Using cgroup v2.
I0211 09:40:50.499154  4661 LinuxMemoryChecker.cpp:55] [PRESTO_STARTUP] Using memory stat file: /sys/fs/cgroup/memory.stat
I0211 09:40:50.499213  4661 LinuxMemoryChecker.cpp:58] [PRESTO_STARTUP] Using memory max file /proc/meminfo
I0211 09:40:50.499545  4661 LinuxMemoryChecker.cpp:89] [PRESTO_STARTUP] System memory in bytes: 2147483648
I0211 09:40:50.499578  4661 LinuxMemoryChecker.cpp:92] [PRESTO_STARTUP] System memory limit in bytes: 4294967296
I0211 09:40:50.499729  4661 LinuxMemoryChecker.cpp:96] [PRESTO_STARTUP] Available machine memory of deployment in bytes: 67421741056
I0211 09:40:50.499758  4661 PeriodicMemoryChecker.cpp:48] [PRESTO_STARTUP] Creating server memory pushback checker, memory check interval 1000ms, system memory limit: 4.00GB, memory shrink size: 20.00GB
I0211 09:40:50.499864  4661 PeriodicMemoryChecker.cpp:57] [PRESTO_STARTUP] Malloc memory heap dumper is not enabled
```

majetideepak (approved):
Thanks, @minhancao

czentgr (approved):

minhancao:
@jaystarshot Hi Jay! Can you help review this PR when you get the chance please?

presto-native-execution/presto_cpp/main/tests/LinuxMemoryCheckerTest.cpp
```diff
@@ -68,20 +141,196 @@ TEST_F(LinuxMemoryCheckerTest, basic) {
...
+      "total_inactive_file 165715968\n"
+      "total_active_file 65421312\n"
+      "total_unevictable 18653184\n";
+  checkMemoryUsage(content, 5136384);
 }
 
 TEST_F(LinuxMemoryCheckerTest, memoryStatFileV2) {
   // Testing cgroup v2 memory.stat file.
-  checkMemoryUsage("cgroupV2memory.stat", 274713448448);
+  const std::string content =
```
jaystarshot, line 277: Previously this test seems to be reading from the .stat file. Isn't that better?

presto-native-execution/presto_cpp/main/PeriodicMemoryChecker.h
```diff
@@ -71,10 +71,10 @@ class PeriodicMemoryChecker {
 
   /// Starts the 'PeriodicMemoryChecker'. A background scheduler will be
   /// launched to perform the checks. This should only be called once.
-  void start();
+  virtual void start();
 
   /// Stops the 'PeriodicMemoryChecker'.
-  void stop();
+  virtual void stop();
```
jaystarshot, line 77: Stop is not overriden so this is not needed?

presto-native-execution/presto_cpp/main/tests/LinuxMemoryCheckerTest.cpp
minhancao: We wanted to avoid checking in data files for testing so that is why I provided its content to be written to a temporary file in the tests.

presto-native-execution/presto_cpp/main/PeriodicMemoryChecker.h
minhancao: Removing virtual for stop()

jaystarshot (dismissed):

majetideepak (approved):

majetideepak:
@minhancao CI is failing. Please verify before merging.

minhancao:
@czentgr @majetideepak @jaystarshot 
I have fixed the CI failure. Please review this PR one more time, thank you!

presto-native-execution/presto_cpp/main/LinuxMemoryChecker.cpp
```diff
@@ -45,8 +67,94 @@ class LinuxMemoryChecker : public PeriodicMemoryChecker {
...
+    memMaxFile_ = memMaxFile;
+  }
+
+  // This function is used for testing.
+  void setMemInfoFile(const std::string& memInfoFile) {
+    memInfoFile_ = memInfoFile;
+  }
+
+  void start() override {
+    if (config_.systemMemPushbackEnabled) {
```
czentgr, line 86: Lets move this check before the checker isd created and return null around line 175. We don't need to call the constructor in this case at all, and then no start function would be called.

minhancao: @czentgr Did you maybe mean a different line number in this file? Line 175 is part of a big paragraph comment:
```
// However we noticed that this value was not closely related to the
```

minhancao: Did you mean to move the check to the singleton code block on line 282?

czentgr: Yes, the line was from the unchanged LinuxMemoryChecker.cpp but with your additions it is the code section you highlighted.

majetideepak (approved):

majetideepak:
@jaystarshot can you please approve this PR again? CI is now passing.

jaystarshot (approved):

czentgr (approved):

PR #24172, opened 2024-11-29
Add single worker execution mode for native execution

kewang1024:
## Description

```
== RELEASE NOTES ==

General Changes
* ... :pr:`12345`
* ... :pr:`12345`
```


tdcmeehan:
Can you help to explain why this feature is or has to be tied to native execution?  Perhaps describe the background and motivation?

arhimondr (commented):
High level comments


1. Do we want to support single node execution on per-query basis?

This can be useful to improve latency of tiny queries running on a large cluster. For example a user may know that a query is small and may decide to run it on a multi node cluster in a single node mode. 

If decided to support it is necessary to make sure the session property is used consistently through the code.

If decided not to support it for now I think the session property should be removed and a configuration property should only be used.

2. Should the single node execution mode be native specific?

When running a Java cluster deployment with a dedicated coordinator and a dedicated worker (workers) additional exchanges at worker - coordinator boundary are necessary.

I'm thinking if a simpler mental model would be to always add coordinator-to-worker exchanges when single node execution is requested?

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/TestPrestoNativeWriterSingleWorker.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    @Override
+    protected ExpectedQueryRunner createExpectedQueryRunner() throws Exception
+    {
+        return PrestoNativeQueryRunnerUtils.createJavaQueryRunner();
+    }
+
+    @Override
+    public void testScaleWriters() {
```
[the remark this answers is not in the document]
kewang1024: To test scaled-writer under single worker execution mode, it wouldn't scale to multiple worker tasks

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/AddExchangeForNativeSingleWorker.java
```diff
@@ -0,0 +1,105 @@
...
+            this.session = session;
+        }
+
+        public boolean isPlanChanged()
+        {
+            return planChanged;
+        }
+
+        @Override
+        public PlanNode visitTableFinish(TableFinishNode node, RewriteContext<Void> context)
```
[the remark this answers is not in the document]
kewang1024: setCoordinatorOnlyDistribution currently works for ExplainAnalyze, TableFinish, MetadataDelete and StatisticsWriterNode

MetadataDelete we don't need to add exchange, I have added the exchange for the rest

presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
```diff
@@ -328,6 +328,7 @@ public final class SystemSessionProperties
     // TODO: Native execution related session properties that are temporarily put here. They will be relocated in the future.
     public static final String NATIVE_AGGREGATION_SPILL_ALL = "native_aggregation_spill_all";
     private static final String NATIVE_EXECUTION_ENABLED = "native_execution_enabled";
+    private static final String NATIVE_SINGLE_WORKER_EXECUTION = "native_single_worker_execution";
```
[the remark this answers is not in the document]
kewang1024: I tried singleNodeExecutionEnabled, but then realized it would cause confusion with [forceSingleNode](https://github.com/prestodb/presto/blob/1327a3e6c2025659760b8f94142dec0edc663485/presto-main/src/main/java/com/facebook/presto/sql/planner/PlanOptimizers.java#L257)

Node can either be worker or coordinator, but what we want is explicitly worker. So `singleWorkerExecutionEnabled` makes more sense

presto-main/src/main/java/com/facebook/presto/sql/planner/PlanOptimizers.java
```diff
@@ -813,7 +814,11 @@ public PlanOptimizers(
                         costCalculator,
                         ImmutableSet.of(new ScaledWriterRule())));
 
-        if (!forceSingleNode) {
+        if (featuresConfig.isNativeExecutionEnabled() && featuresConfig.isNativeSingleWorkerExecution()) {
```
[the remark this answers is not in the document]
kewang1024: But for some cases, it would be exchange from coordinator to worker? For example, scanning system table
Aggregation [Worker]
|
Exchange
|
TableScan (system table) [Coordinator]

presto-main/src/main/java/com/facebook/presto/SystemSessionProperties.java
```diff
@@ -324,6 +324,7 @@ public final class SystemSessionProperties
     public static final String OPTIMIZER_USE_HISTOGRAMS = "optimizer_use_histograms";
     public static final String WARN_ON_COMMON_NAN_PATTERNS = "warn_on_common_nan_patterns";
     public static final String INLINE_PROJECTIONS_ON_VALUES = "inline_projections_on_values";
+    public static final String SINGLE_WORKER_EXECUTION_ENABLED = "single_worker_execution_enabled";
```
arhimondr, line 327: Let's call it `single_node_execution_enabled`. When running on Java and `schedule-on-coordinator` is enabled the query may run completely on coordinator.

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/AddExchangeForSingleWorkerExecution.java
```diff
@@ -0,0 +1,160 @@
...
+import static com.facebook.presto.sql.planner.PlannerUtils.containsSystemTableScan;
+import static com.facebook.presto.sql.planner.SystemPartitioningHandle.SINGLE_DISTRIBUTION;
+import static com.facebook.presto.sql.planner.iterative.rule.PickTableLayout.pushPredicateIntoTableScan;
+import static com.facebook.presto.sql.planner.plan.ExchangeNode.Scope.REMOTE_STREAMING;
+import static com.facebook.presto.sql.planner.plan.ExchangeNode.Type.GATHER;
+import static com.facebook.presto.sql.planner.plan.ExchangeNode.ensureSourceOrderingGatheringExchange;
+import static com.facebook.presto.sql.planner.plan.ExchangeNode.gatheringExchange;
+import static java.util.Objects.requireNonNull;
+
+public class AddExchangeForSingleWorkerExecution
```
arhimondr, line 49: nit: `AddExchangesForSingleNodeExecution`

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/AddExchangeForSingleWorkerExecution.java
```diff
@@ -0,0 +1,160 @@
...
+        {
+            return addGatherExchange(node);
+        }
+
+        private PlanNode addGatherExchange(InternalPlanNode node)
+        {
+            PlanNode child = node.getSources().get(0);
+
+            // if the child is already a gathering exchange
+            if ((child instanceof ExchangeNode) && ((ExchangeNode) child).getType() == ExchangeNode.Type.GATHER) {
```
arhimondr, line 152: nit: Maybe also check if child is an exchange and replace it with `GATHER`. Then you can use it in `visitTableFinish` as well.

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/MergeJoinForSortedInputOptimizer.java
```diff
@@ -55,6 +58,9 @@ public void setEnabledForTesting(boolean isSet)
     @Override
     public boolean isEnabled(Session session)
     {
+        if (!isSingleWorkerExecutionEnabled(session)) {
```
arhimondr, line 61: Should it be `if (isSingleWorkerExecutionEnabled(session)) ...`

arhimondr (commented):
LGTM % nits and fixing test failures

steveburnett:
Thanks for the release note entry! Some formatting nits. 

```
== RELEASE NOTES ==

General Changes
* Add single worker execution. To improve latency of tiny queries running on a large cluster, we introduce single worker execution mode: query will only use one node to execute and plan would be optimized accordingly. This feature can be turned on by the configuration property ``single-node-execution-enabled`` or session property ``single_node_execution_enabled``. :pr:`24172`
```

Also, consider adding documentation for the new configuration property and session property to either the Presto [Configuration, Session] Properties pages, or the Presto C++ pages, as appropriate.

arhimondr (approved):

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/AddExchangesForSingleNodeExecution.java
```diff
@@ -0,0 +1,150 @@
...
+
+        @Override
+        public PlanNode visitTableScan(TableScanNode node, RewriteContext<Void> context)
+        {
+            PlanNode plan = pushPredicateIntoTableScan(node, new ConstantExpression(true, BOOLEAN), true, session, idAllocator, metadata);
+            // Presto Java and Presto Native use different hash functions for partitioning
+            // An additional exchange makes sure the data flows through a native worker in case it need to be partitioned for downstream processing
+            if (containsSystemTableScan(plan)) {
+                plan = gatheringExchange(idAllocator.getNextId(), REMOTE_STREAMING, plan);
+            }
```
tdcmeehan, line 102: Are these three lines of code the reason we've extended so many test cases to apply to Presto - single node - native?  If so, I'm wondering if a few, targetted example based tests are more appropriate.  I'm concerned the additional tests don't add value and will make our CI slower and more expensive.

tdcmeehan (changes requested):

presto-main/src/main/java/com/facebook/presto/sql/planner/optimizations/AddExchangesForSingleNodeExecution.java
kewang1024: Not only so, essentially under single-node mode, we will not use addExchange optimizer. Theoretically we should test many cases that originally have remote exchange (especially partitioned), would they still return correct result under single-node mode.
 
And also scheduling has changes accordingly, one of the test cases actually caught a issue for it.

I understand your concern, I remove some of test (that I think could potentially be redundant in terms of exchange pattern, for example window function VS aggregation function)

kewang1024:
It won't let me rerun some sporadic flaky test, have to push to force rerun :(

tdcmeehan:
@kewang1024 this feature could be useful for lower latency deployments to support canned or bounded queries, where latency is expected to be low.  Could you please introduce some documentation for it as @steveburnett requested?

tdcmeehan:
> > High level comments
> > 
> > 1. Do we want to support single node execution on per-query basis?
> > 
> > This can be useful to improve latency of tiny queries running on a large cluster. For example a user may know that a query is small and may decide to run it on a multi node cluster in a single node mode.
> 
> We can also potentially use HBO/CBO to decide to run some in single node mode

@kewang1024 /@kaikalur can you create an issue for this so we don't lose track of this suggestion?

I think we could also consider to toggle this feature via a resource group, similar to [per-query-limits in resource groups](https://prestodb.io/docs/current/admin/resource-groups.html#resource-group-properties).  This might make the feature more convenient to toggle, and in a multinode cluster, this would more accurate value for the hard/soft concurrency limit (and perhaps make it safer to run in a multitenant deployment).  I can create an issue for that, I don't think it needs to be added here.

kewang1024:
@steveburnett  the Presto C++ pages is not a good place since this is not limited to c++, I failed to find the [Configuration, Session] one you're referring to, can u give me a pointer

steveburnett:
> @steveburnett the Presto C++ pages is not a good place since this is not limited to c++, I failed to find the [Configuration, Session] one you're referring to, can u give me a pointer

Of course! I was referring the to Presto Configuration Properties page
https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/admin/properties.rst

or the Presto Session Properties page
https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/admin/properties-session.rst

kewang1024:
Thanks @steveburnett for the prompt response, updated the doc.
cc: @tdcmeehan

presto-docs/src/main/sphinx/admin/properties.rst
```diff
@@ -70,6 +70,15 @@ When enabled, the logical plan will begin to be built and validated before
 queueing and allocation of cluster resources so that any errors or
 incompatibilities in the query plan will fail quickly and inform the user.
 
+``single-node-execution-enabled``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``boolean``
+* **Default value:** ``false``
+
+This property enables running the query on a single node under a distributed cluster setup.
```
tdcmeehan, line 79:
```suggestion
This property ensures that queries scheduled in this cluster use only a single node for execution, which may improve performance for small queries which can be executed within a single node.
```

The above is just a suggestion.  Avoid referring to "the query", because a config property is global and affects "all queries".  I also wouldn't presume the reader has knowledge on what is considered to be a low latency use case, so I'd drop the definite article, and I'd try to qualify what counts as a small/tiny query in some way.

kewang1024:
@tdcmeehan Updated accordingly, can you help take another look? Thanks!

presto-main/src/main/java/com/facebook/presto/execution/scheduler/SectionExecutionFactory.java
```diff
@@ -326,7 +327,58 @@ else if (partitioningHandle.equals(SCALED_WRITER_DISTRIBUTION)) {
...
+                        return false;
+                    }
+                    @Override
+                    public void assignOrUpdateBucketToNode(int bucketedId, InternalNode node, boolean cacheable)
+                    {
+                    }
+                    @Override
+                    public boolean isDynamic()
+                    {
+                        return true;
```
tdcmeehan, line 356: Just curious, why is this true?  Shouldn't this be false, since I think there is a single node and a single task?

presto-main/src/main/java/com/facebook/presto/execution/scheduler/SectionExecutionFactory.java
```diff
@@ -326,7 +327,58 @@ else if (partitioningHandle.equals(SCALED_WRITER_DISTRIBUTION)) {
...
+                    }
+                    @Override
+                    public boolean hasInitialMap()
+                    {
+                        return false;
```
tdcmeehan, line 361: Can't this be true?  Since there's only one node being returned in `getBucketToNode`?

tdcmeehan (commented):
Overall, wondering what's preventing us from using `FixedBucketNodeMap`, since it seems that `BucketNodeMap` is aligned with this use case (only using a single node-bucket).

presto-main/src/main/java/com/facebook/presto/execution/scheduler/SectionExecutionFactory.java
arhimondr: It is only applicable for grouped execution. Grouped execution is currently not supported for single node mode. It can be supported if needed, but generally the idea is that only small queries should run single node, while grouped execution is generally applicable for very large queries.

presto-main/src/main/java/com/facebook/presto/execution/scheduler/SectionExecutionFactory.java
arhimondr: This is true. @kewang1024 I wonder if instead we can simply use a `DynamicBucketNodeMap((split) -> 0, 1)`. Basically pretending there's only a single bucket for all splits to avoid a custom override?

kewang1024: Sure, can make that change now

arhimondr (approved):

Issue #21332, opened 2023-11-07
[Native] - Support for char(n) is missing in velox


mknegi:
It seems currently we don't support char(n) type in velox. One can create a table with this type but can't do any operations on the table.

## Current Behavior

```
presto:tpch> describe tab2;
 Column |   Type   | Extra | Comment
--------+----------+-------+---------
 col1   | integer  |       |
 col2   | char(20) |       |
(2 rows)

Query 20231106_204830_00050_qqghb, FINISHED, 2 nodes
Splits: 3 total, 3 done (100.00%)
[Latency: client-side: 94ms, server-side: 77ms] [2 rows, 117B] [25 rows/s, 1.48KB/s]

presto:tpch> select * from tab2;
Query 20231106_204836_00051_qqghb failed:  Failed to parse type [char(20)]:

presto:tpch> insert into tab2 values (1, 'aaa');
Query 20231106_205006_00052_qqghb failed:  Failed to parse type [char(20)]:
```

mknegi:
Another issue https://github.com/facebookincubator/velox/discussions/6888 is opened for varchar(n) support.

pdabre12:
Remove the char to varchar mapping in https://github.com/prestodb/presto/pull/24179 when this is implemented.

PR #24179, opened 2024-12-02
Add a native type manager

pdabre12:
## Description
With the introduction of the presto-native-execution module, Presto now has forked the execution engine into a separate process. This allows for the execution engine to be written in C++ and to be more efficient than the Java execution engine. However, this also means that the execution engine code is now entirely separate from the coordinator code, and that the types supported by the coordinator might not necessarily be supported by the execution engine.

This change adds a new native type manager which allows the types supported by the C++ engine to be known to the Java coordinator. 

For more context: [RFC-0003](https://github.com/prestodb/rfcs/blob/main/RFC-0003-native-spi.md)

## Test Plan
Unit and end-to-end tests. More comprehensive end-to-end tests will be written in the future.

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

* Add a native type manager
```


presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/typemanager/NativeTypeManager.java
```diff
@@ -0,0 +1,195 @@
...
+    private static final Set<String> NATIVE_ENGINE_SUPPORTED_PARAMETRIC_TYPES =
+            ImmutableSet.of(
+                    ARRAY,
+                    DECIMAL,
+                    MAP,
+                    ROW,
+                    FUNCTION);
+
+    private final TypeManager typeManager;
+    private final LoadingCache<ExactTypeSignature, Type> parametricTypeCache;
```
tdcmeehan, line 108: Just copy the `ExactTypeSignature` here as well.  It's a bit too specific to be useful as a general API in common.

presto-spi/src/main/java/com/facebook/presto/spi/type/TypesProvider.java
```diff
@@ -0,0 +1,28 @@
...
+package com.facebook.presto.spi.type;
+
+import com.facebook.presto.common.type.ParametricType;
+import com.facebook.presto.common.type.Type;
+import com.facebook.presto.common.type.TypeSignature;
+
+import java.util.Map;
+import java.util.Optional;
+
+public interface TypesProvider
```
tdcmeehan, line 23: I'm wondering if we can avoid this new interface and instead use the existing `TypeManager` interface.

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/typemanager/NativeTypeManager.java
pdabre12: The problem is `ExactTypeSignature` is used in `FunctionAndTypeManager` too.

presto-main/src/main/java/com/facebook/presto/metadata/FunctionAndTypeManager.java
```diff
@@ -301,6 +325,17 @@ public FunctionMetadata getFunctionMetadata(FunctionHandle functionHandle)
         return functionNamespaceManager.get().getFunctionMetadata(functionHandle);
     }
 
+    @Override
+    public Type instantiateParametricType(ExactTypeSignature exactSignature)
```
tdcmeehan, line 329: Can this take in a `TypeSignature`, and we manually create an `ExactTypeSignature` inside this method, so we can avoiding moving `ExactTypeSignature` into the SPI?

pdabre12: That would mean not using the `ExactTypeSignature` here: https://github.com/prestodb/presto/blob/fa3a49b8d6df5625865506df8d176cb457de2796/presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/typemanager/NativeTypeManager.java#L142,  which makes the usage of `ExactTypeSignature` redundant.

auden-woolfson (commented):
LGTM

presto-main/src/main/java/com/facebook/presto/metadata/FunctionAndTypeManager.java
pdabre12: Done.

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/typemanager/NativeTypeManager.java
pdabre12: Done.

steveburnett:

```

:pr:`12345`

```

pdabre12:
@steveburnett Thank you. Made the change.

pdabre12:
@tdcmeehan @rschlussel This PR is now ready for review, PTAL. Thanks.

steveburnett:
Thanks for the release note entry! Minor change to follow the [Order of sections](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines#order-of-sections) in the Release Notes Guidelines: 

```
== RELEASE NOTES ==

Prestissimo (Native Execution) changes
* Add a native type manager
```

presto-main/src/main/java/com/facebook/presto/metadata/FunctionAndTypeManager.java
```diff
@@ -835,6 +893,19 @@ public CatalogSchemaName configureDefaultNamespace(String defaultNamespacePrefix
         return new CatalogSchemaName(catalogSchemaNameString[0], catalogSchemaNameString[1]);
     }
 
+    private TypeManager getServingTypeManager()
+    {
+        // If the NativeSidecarPlugin is loaded, typeManagerFactories is non-empty and holds an instance of NativeTypeManagerFactory.
+        // Check if the nativeTypeManager exists in typeManagers else use BuiltInTypeAndFunctionNamespaceManager.
```
tdcmeehan, line 899:
```suggestion
        // Check if a custom TypeManager has been registered, otherwise use BuiltInTypeAndFunctionNamespaceManager.
```

tdcmeehan (dismissed):

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/typemanager/NativeTypeManager.java
```diff
@@ -0,0 +1,303 @@
...
+        if (typeSignature.getBase().equals("char") || typeSignature.getBase().equals("varchar")) {
+            typeSignature = createUnboundedVarcharType().getTypeSignature();
+        }
+        Type type = types.get(typeSignature);
+        if (type != null) {
+            return type;
+        }
+        try {
+            return parametricTypeCache.getUnchecked(new ExactTypeSignature(typeSignature));
+        }
```
auden-woolfson, line 144: Can also make this more concise the same way

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/typemanager/NativeTypeManager.java
```diff
@@ -0,0 +1,303 @@
...
+    public boolean canCoerce(Type actualType, Type expectedType)
+    {
+        throw new UnsupportedOperationException();
+    }
+
+    private void addAllTypes(List<Type> typesList, List<ParametricType> parametricTypesList)
+    {
+        for (Type type : typesList) {
+            addType(type);
+        }
```
auden-woolfson, line 179:
```
typesList.stream().forEach(type -> addType(type));
```

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/typemanager/NativeTypeManager.java
```diff
@@ -0,0 +1,303 @@
...
+        // todo: Fix this hack
+        // Native engine does not support parameterized varchar, and varchar isn't in the lists of types returned from the engine
+        addType(VarcharType.VARCHAR);
+
+        for (ParametricType type : parametricTypesList) {
+            addParametricType(type);
+        }
```
auden-woolfson, line 186: Can use same functional pattern

auden-woolfson (changes requested):
LGTM, just a few best practices/nits.

presto-main/src/main/java/com/facebook/presto/metadata/BuiltInTypeAndFunctionNamespaceManager.java
```diff
@@ -1261,21 +1262,34 @@ public ScalarFunctionImplementation getScalarFunctionImplementation(Signature si
...
         }
         try {
-            return Optional.ofNullable(parametricTypeCache.getUnchecked(new ExactTypeSignature(typeSignature)));
+            return parametricTypeCache.getUnchecked(new ExactTypeSignature(typeSignature));
         }
         catch (UncheckedExecutionException e) {
             throwIfUnchecked(e.getCause());
             throw new RuntimeException(e.getCause());
         }
     }
```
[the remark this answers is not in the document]
pdabre12: Yes, it will be more concise, but I think the current code is much easier to read. What do you think?

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/typemanager/NativeTypeManager.java
pdabre12: Same as above

presto-main/src/main/java/com/facebook/presto/metadata/BuiltInTypeAndFunctionNamespaceManager.java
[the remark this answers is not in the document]
auden-woolfson: Ok, either way works

auden-woolfson (dismissed):

presto-common/src/main/java/com/facebook/presto/common/type/TypeManager.java
```diff
@@ -28,4 +29,13 @@ public interface TypeManager
...
     boolean canCoerce(Type actualType, Type expectedType);
+
+    default Type instantiateParametricType(TypeSignature typeSignature)
+    {
+        throw new UnsupportedOperationException();
+    }
+
+    List<Type> getTypes();
+
+    Map<String, ParametricType> getParametricTypes();
```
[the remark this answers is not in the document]
pdabre12: Made the change, thanks.

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/typemanager/NativeTypeManagerConfig.java
```diff
@@ -0,0 +1,38 @@
...
+import io.airlift.units.MinDuration;
+
+import static java.util.concurrent.TimeUnit.MINUTES;
+
+public class NativeTypeManagerConfig
+{
+    private Duration typesCacheExpiration = new Duration(5, MINUTES);
+
+    @MinDuration("0ns")
+    public Duration getTypesCacheExpiration()
```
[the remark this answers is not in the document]
pdabre12: Removed this.

presto-main/src/main/java/com/facebook/presto/metadata/FunctionAndTypeManager.java
```diff
@@ -835,6 +896,18 @@ public CatalogSchemaName configureDefaultNamespace(String defaultNamespacePrefix
...
+    private TypeManager getServingTypeManager()
+    {
+        // Check if a custom TypeManager has been registered, otherwise use BuiltInTypeAndFunctionNamespaceManager.
+        for (Map.Entry<String, TypeManagerFactory> entry : typeManagerFactories.entrySet()) {
+            if (!typeManagers.containsKey(entry.getKey())) {
+                throw new PrestoException(GENERIC_USER_ERROR, format("Type manager not loaded for factory: %s", entry.getKey()));
+            }
+            return typeManagers.get(entry.getKey());
+        }
+        return builtInTypeAndFunctionNamespaceManager;
```
ZacBlanco, line 908: Is this something we could use with `Suppliers.memoize`?

presto-main/src/main/java/com/facebook/presto/sql/analyzer/ExpressionAnalyzer.java
```diff
@@ -1342,6 +1349,12 @@ public Type visitCast(Cast node, StackableAstVisitorContext<Context> context)
             catch (IllegalArgumentException e) {
                 throw new SemanticException(TYPE_MISMATCH, node, "Unknown type: " + node.getType());
             }
+            catch (PrestoException e) {
+                if (e.getErrorCode() == UNKNOWN_TYPE.toErrorCode()) {
+                    throw new SemanticException(TYPE_MISMATCH, node, "Unknown type: " + node.getType());
+                }
```
ZacBlanco, line 1355: Were you supposed to delete the `IllegalArgumentException` block here?

Also The fact that this is pattern (catch PrestoException, check error code) is so common makes me think we might want to create something like `UnknownTypeException` as subclass of `PrestoException` that always has the `UNKNOWN_TYPE` error code. Then all these catch blocks can just check for `UnknownTypeException` instead. It seems like forgetting this conditional could be something we miss in the future.

Another option could be to just throw SemanticException directly from the type manager? It would prevent having to modify the catch/re-throw blocks in a lot of locations

ZacBlanco (commented):
Overall, looks good. I am concerned about the exception handling aspect though.

Also, please also update the release note section header to `Prestissimo (Native Execution) Changes`

presto-main/src/main/java/com/facebook/presto/metadata/FunctionAndTypeManager.java
pdabre12: Done, thanks.

presto-main/src/main/java/com/facebook/presto/sql/analyzer/ExpressionAnalyzer.java
pdabre12: Added a new `UnknownTypeException`, subclass of `PrestoException`.

No, the `IllegalArgumentException`  block is still required. 
This block catches `IllegalArgumentExceptions` from classes like https://github.com/prestodb/presto/blob/master/presto-main/src/main/java/com/facebook/presto/type/MapParametricType.java#L52

pdabre12:
@ZacBlanco Thanks for the quick review.
 Addressed your comments, can you please have another look?

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/typemanager/NativeTypeManager.java
```diff
@@ -0,0 +1,298 @@
...
+                filterSupportedTypes(
+                        NATIVE_ENGINE_SUPPORTED_PARAMETRIC_TYPES,
+                        new ArrayList<>(typeManager.getParametricTypes()),
+                        ParametricType::getName));
+    }
+
+    @Override
+    public Type getType(TypeSignature typeSignature)
+    {
+        // Todo: Fix this hack, native execution does not support parameterized char/varchar type signatures.
```
ZacBlanco, line 133: Do we have an open ticket for this so that we can track when the code needs to be removed?

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/typemanager/NativeTypeManager.java
```diff
@@ -0,0 +1,298 @@
...
+        if (typeSignature.getBase().equals("char") || typeSignature.getBase().equals("varchar")) {
```
ZacBlanco, line 134: nit: use the type name from the StandardTypes class
```suggestion
        if (typeSignature.getBase().equals(CHAR) || typeSignature.getBase().equals(VARCHAR)) {
```

ZacBlanco (changes requested):
Two minors things, otherwise LGTM

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/typemanager/NativeTypeManager.java
pdabre12: I was not able to find an open ticket, but this is the PR where support will be added : https://github.com/facebookincubator/velox/pull/10727, its still a WIP.

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/typemanager/NativeTypeManager.java
pdabre12: Done, thanks !

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/typemanager/NativeTypeManager.java
pdabre12: I found an issue: https://github.com/prestodb/presto/issues/21332. I mentioned there that the char to varchar mapping in this PR needs to be removed when the varchar(N) support lands. Will that work?

ZacBlanco: Yeah, that works

ZacBlanco (dismissed):

pdabre12:
@rschlussel gentle ping, can you please have a look?

majetideepak (commented):
I have one question. 
I cannot sign off on this since I do not have enough context on the changes made to the Java API.
We will need an approval from a Java committer.
CC: @jaystarshot

presto-common/src/main/java/com/facebook/presto/common/type/StandardTypes.java
```diff
@@ -70,6 +72,7 @@ private StandardTypes() {}
             ROW,
             ARRAY,
             MAP,
+            FUNCTION,
```
rschlussel, line 75: what is a function type?

jaystarshot:
I think both commits can be squashed into one, also please add some info in the commit message [guidelines](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#example-commit-message)

presto-common/src/main/java/com/facebook/presto/common/type/TypeManager.java
```diff
@@ -28,4 +29,19 @@ public interface TypeManager
     Type getParameterizedType(String baseTypeName, List<TypeSignatureParameter> typeParameters);
 
     boolean canCoerce(Type actualType, Type expectedType);
+
+    default Type instantiateParametricType(TypeSignature typeSignature)
```
rschlussel, line 33: can we extract a separate interface to split out the TypeManager from a TypeNamespaceManager. Currently the responsibilities of FunctionAndTypeManager and BuiltinTypeAndFunctionNamespaceManager are distinct, but the methods for both are all in one interface, and we throw unsupportedoperationexceptions for the things that we don't expect to be under the responsibility for the particular class, which is not very safe.

rschlussel:
> I think both commits can be squashed into one, also please add some info in the commit message [guidelines](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#example-commit-message)

either way is okay, but I like having the smaller commits

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/typemanager/NativeTypeManager.java
```diff
@@ -0,0 +1,299 @@
...
+import static com.google.common.base.Throwables.throwIfUnchecked;
+import static com.google.common.collect.ImmutableList.toImmutableList;
+import static java.util.Objects.requireNonNull;
+
+public class NativeTypeManager
+        implements TypeManager
+{
+    private static final Logger log = Logger.get(NativeTypeManager.class);
+
+    private static final Set<String> NATIVE_ENGINE_SUPPORTED_TYPES =
```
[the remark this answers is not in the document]
pdabre12: Ideally yes, but since types is a very slow moving thing, it makes sense to just add a newly added type to this list as the updates are not going to be very often. What do you think?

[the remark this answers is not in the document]
majetideepak: This is okay as a first step. But user-defined types would require querying the native engine.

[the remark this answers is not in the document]
aditi-pandit: @majetideepak : This came up in a conversation with @pdabre12.

For user defined types, in addition to the worker type, any functions it participates in, the java co-ordinator also needs to know how to serialize/deserialize the objects of the type in java... Unless we are always going to pass-through opaque/binary serialized versions of the user defined type from the co-ordinator to the client. This is still an open-ended issue to resolve before making the types available from the native engine.

Once we have a good design for it, we can make the types list more dynamically retrieved from the workers.

presto-common/src/main/java/com/facebook/presto/common/type/StandardTypes.java
pdabre12: It's used for lambda expressions. `FunctionType` is defined here: https://github.com/prestodb/presto/blob/f7103fe0c052460c38328f0708a1127aeccbf0c3/presto-common/src/main/java/com/facebook/presto/common/type/FunctionType.java#L30.
I do not think it should be added to `StandardTypes.java` , so I just added it to the supported parametric types list in `NativeTypeManager`.

presto-common/src/main/java/com/facebook/presto/common/type/TypeManager.java
pdabre12: This was a huge refactor, I have a WIP PR for it: https://github.com/prestodb/presto/pull/24640. 
Is it okay if we let this one pass and handle in the PR described above?

presto-main/src/main/java/com/facebook/presto/metadata/FunctionAndTypeManager.java
pdabre12: @ZacBlanco I had to revert the changes. We cannot use `Suppliers.memoize` here because the call to `getServingTypeManager()` happens before the plugins are loaded meaning the `typeManager` used will always be `BuiltInTypeAndFunctionNamespaceManager` even if the `NativeTypeManager` is loaded.

pdabre12:
@rschlussel Please have another look, thanks.

presto-main/src/main/java/com/facebook/presto/metadata/FunctionAndTypeManager.java
ZacBlanco: Since getting a reference to the type manager is done quite frequently, it would be nice if we could just have a single reference so we don't have to do this iteration so often. I know that `typeManagerFactories` is likely going to just be 1 or 2 entries, but it seems inefficient to have to do this for every access. Do you think there's any other way we could cache this since it should be the same for the lifetime of the coordinator?

pdabre12: Maybe using an `AtomicReference` would work here.
`private final AtomicReference<TypeManager> servingTypeManager;`
 
 In the constructor call, 
 // initialize it with the built-in type manager.
` servingTypeManager = new AtomicReference<>(builtInTypeANdFunctionNamespaceManager);
` 
 As soon as the `nativeTypeManager` is load, update the `servingTypeManager` to be the `nativeTypeManager`.
 
 ```
public void loadTypeManager(String typeManagerName)
    {
        requireNonNull(typeManagerName, "typeManagerName is null");
        TypeManagerFactory factory = typeManagerFactories.get(typeManagerName);
        checkState(factory != null, "No factory for type manager %s", typeManagerName);
        TypeManager typeManager = factory.create(new TypeManagerContext(this));

        if (typeManagers.putIfAbsent(typeManagerName, typeManager) != null) {
            throw new IllegalArgumentException(format("Type manager [%s] is already registered", typeManager));
        }
        servingTypeManager.compareAndSet(servingTypeManager.get(), typeManager):
    }
```
In this way, we can just use `servingTypeManager` directly, which should solve the inefficiency described above.
What do you think?

Comment (ZacBlanco) [reply presto-main/src/main/java/com/facebook/presto/metadata/FunctionAndTypeManager.java:908]: I think this approach sounds good

Comment (pdabre12) [reply presto-main/src/main/java/com/facebook/presto/metadata/FunctionAndTypeManager.java]: Made the changes, thanks.

**Comment by pdabre12** [comment]
@rschlussel 
Do you have any more comments on this? 
I addressed the previous comments. Thanks.

**Review by rschlussel — DISMISSED** [review]

**Comment by pdabre12** [comment]
@ZacBlanco Do you wanna take a look?

File: presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/typemanager/NativeTypeManager.java [hunk 18]
```diff
@@ -0,0 +1,299 @@
...
+                        NATIVE_ENGINE_SUPPORTED_PARAMETRIC_TYPES,
+                        new ArrayList<>(typeManager.getParametricTypes()),
+                        ParametricType::getName));
+    }
+
+    @Override
+    public Type getType(TypeSignature typeSignature)
+    {
+        // Todo: Fix this hack, native execution does not support parameterized char/varchar type signatures.
+        if (typeSignature.getBase().equals(CHAR) || typeSignature.getBase().equals(VARCHAR)) {
```
aditi-pandit, line 135: Trying to reduce the impact of this hack. What happens if we remove CHAR from this condition and fail all the queries with it.

presto-main/src/main/java/com/facebook/presto/metadata/FunctionAndTypeManager.java
```diff
@@ -319,6 +346,17 @@ public FunctionMetadata getFunctionMetadata(FunctionHandle functionHandle)
         return functionNamespaceManager.get().getFunctionMetadata(functionHandle);
     }
 
+    @Override
+    public Type instantiateParametricType(TypeSignature typeSignature)
+    {
+        Map<String, ParametricType> parametricTypes = servingTypeManager.get().getParametricTypes().stream()
```
ZacBlanco, line 352: This seems a little expensive every time we instantiate a parametric type. Can we do anything to avoid the loop+copy here?

pdabre12: Maybe we can use `Suppliers.cache()`.
The only problem seems to be that the `servingTypeManager` can change so we need to update the cache based on whether `servingTypeManager` has changed.

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/typemanager/NativeTypeManager.java
pdabre12: Done, removed `CHAR`.

presto-main/src/main/java/com/facebook/presto/metadata/FunctionAndTypeManager.java
pdabre12: Figured out a way , 
using 
`private final AtomicReference<Supplier<Map<String, ParametricType>>> servingTypeManagerParametricTypesSupplier;`

Updating the cache when we load the native type manager.

```
public void loadTypeManager(String typeManagerName)
    {
        requireNonNull(typeManagerName, "typeManagerName is null");
        TypeManagerFactory factory = typeManagerFactories.get(typeManagerName);
        checkState(factory != null, "No factory for type manager %s", typeManagerName);
        TypeManager typeManager = factory.create(new TypeManagerContext(this));

        if (typeManagers.putIfAbsent(typeManagerName, typeManager) != null) {
            throw new IllegalArgumentException(format("Type manager [%s] is already registered", typeManager));
        }
        servingTypeManager.compareAndSet(servingTypeManager.get(), typeManager);
        // Reset the parametric types cache
        servingTypeManagerParametricTypesSupplier.set(this::getServingTypeManagerParametricTypes);
    }
```

pdabre12:
@rschlussel 
Can you please have another look?
Native engine shouldn't allow creating tables with char column.
So any queries on the older tables constructed with char column will fail.
Is it safe to do so in your production environment?

libianoss:
@rschlussel can you help to review this PR?Thanks.

ZacBlanco (approved):

rschlussel:
> @rschlussel Can you please have another look? Native engine shouldn't allow creating tables with char column. Whenever the sidecar plugin is enabled, any queries on the older tables constructed with char column will fail. Is it safe to do so in your production environment?

i don't think i understand what this means. Why doesn't native engine allow creating tables with char column?  Currently without sidecar, what happens if someone runs a query on a table with a char column?

pdabre12:
> > @rschlussel Can you please have another look? Native engine shouldn't allow creating tables with char column. Whenever the sidecar plugin is enabled, any queries on the older tables constructed with char column will fail. Is it safe to do so in your production environment?
> 
> i don't think i understand what this means. Why doesn't native engine allow creating tables with char column? Currently without sidecar, what happens if someone runs a query on a table with a char column?

As of now, the native engine allows creating tables with char column and queries on a table with a char column should succeed. But with sidecar enabled, we are getting a list of supported types by the native engine and `CHAR` is not a part of it. As a result, if you have old tables created with the underlying column type as char and enable the sidecar, the queries on that table will fail.

rschlussel:
> > > @rschlussel Can you please have another look? Native engine shouldn't allow creating tables with char column. Whenever the sidecar plugin is enabled, any queries on the older tables constructed with char column will fail. Is it safe to do so in your production environment?
> > 
> > 
> > i don't think i understand what this means. Why doesn't native engine allow creating tables with char column? Currently without sidecar, what happens if someone runs a query on a table with a char column?
> 
> As of now, the native engine allows creating tables with char column and queries on a table with a char column should succeed. But with sidecar enabled, we are getting a list of supported types by the native engine and `CHAR` is not a part of it. As a result, if you have old tables created with the underlying column type as char and enable the sidecar, the queries on that table will fail.

if currently we can run queries on char columns, then why is it not supported?

pdabre12:
> > > > @rschlussel Can you please have another look? Native engine shouldn't allow creating tables with char column. Whenever the sidecar plugin is enabled, any queries on the older tables constructed with char column will fail. Is it safe to do so in your production environment?
> > > 
> > > 
> > > i don't think i understand what this means. Why doesn't native engine allow creating tables with char column? Currently without sidecar, what happens if someone runs a query on a table with a char column?
> > 
> > 
> > As of now, the native engine allows creating tables with char column and queries on a table with a char column should succeed. But with sidecar enabled, we are getting a list of supported types by the native engine and `CHAR` is not a part of it. As a result, if you have old tables created with the underlying column type as char and enable the sidecar, the queries on that table will fail.
> 
> if currently we can run queries on char columns, then why is it not supported?

I am not really sure what happens under the hood right now when you create a table with char columns in a native cluster. The native engine does not understand char type, maybe it gets casted to a varchar( not sure) ? cc: @aditi-pandit.
But with this change, we only report the types supported by the native engine to the coordinator and as of now char is not one of them, so the queries would fail. This isn't an exclusive case for the char type but all the older tables created with types not supported by the native engine would fail.

rschlussel:
@gggrace14 showed me that we actually are failing now in the workers when you try to query a char column, so this change should be okay.

rschlussel (approved):

aditi-pandit (approved):
Thanks @pdabre12

aditi-pandit:
> > > > > @rschlussel Can you please have another look? Native engine shouldn't allow creating tables with char column. Whenever the sidecar plugin is enabled, any queries on the older tables constructed with char column will fail. Is it safe to do so in your production environment?
> > > > 
> > > > 
> > > > i don't think i understand what this means. Why doesn't native engine allow creating tables with char column? Currently without sidecar, what happens if someone runs a query on a table with a char column?
> > > 
> > > 
> > > As of now, the native engine allows creating tables with char column and queries on a table with a char column should succeed. But with sidecar enabled, we are getting a list of supported types by the native engine and `CHAR` is not a part of it. As a result, if you have old tables created with the underlying column type as char and enable the sidecar, the queries on that table will fail.
> > 
> > 
> > if currently we can run queries on char columns, then why is it not supported?
> 
> I am not really sure what happens under the hood right now when you create a table with char columns in a native cluster. The native engine does not understand char type, maybe it gets casted to a varchar( not sure) ? cc: @aditi-pandit. But with this change, we only report the types supported by the native engine to the coordinator and as of now char is not one of them, so the queries would fail. This isn't an exclusive case for the char type but all the older and newer tables/queries with types not supported by the native engine would fail.

@rschlussel : Today C++ clusters would support CREATE TABLE with char columns, but SELECT from them would fail at the workers since the PrestoToVelox conversions would reject the CHAR column. With this change, CHAR would not be exposed as a type, so even CREATE TABLE with char columns would fail.

PR #24226, opened 2024-12-09
Fix strrpos function

feilong-liu:
## Description
The current strrpos function returns wrong result for multibyte characters

## Motivation and Context
bug fix

## Impact
bug fix

## Test Plan
Unit tests

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Fix bug in strrpos function for multibyte characters
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/operator/scalar/StringFunctions.java
+++ b/presto-main/src/main/java/com/facebook/presto/operator/scalar/StringFunctions.java
@@ -267,13 +267,15 @@ public final class StringFunctions
             return 1;
         }
 
+        String stringUtf8 = string.toStringUtf8();
+        String substringUtf8 = substring.toStringUtf8();
         int foundInstances = 0;
         // set the initial index just after the end of the string
         // this is to allow for the initial index decrement
-        int index = string.length();
+        int index = stringUtf8.length();
         do {
             // step backwards through string
-            index = string.toStringUtf8().lastIndexOf(substring.toStringUtf8(), index - 1);
+            index = stringUtf8.lastIndexOf(substringUtf8, index - 1);
             if (index < 0) {
                 return 0;
             }
@@ -281,7 +283,9 @@ public final class StringFunctions
         }
         while (foundInstances < instance);
 
-        return countCodePoints(string, 0, index) + 1;
+        // stringPositionFromStart function returns countCodePoints(string, 0, index) + 1 because it directly works on Slice and use indexOf function of Slice,
+        // Here we convert Slice to Utf8 string call lastIndexOf function of String to get index, hence directly return index+1
+        return index + 1;
     }
 
     @Description("suffix starting at given index")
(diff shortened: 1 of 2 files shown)


presto-main/src/main/java/com/facebook/presto/operator/scalar/StringFunctions.java
```diff
@@ -281,7 +281,7 @@ private static long stringPositionFromEnd(Slice string, Slice substring, long in
         }
         while (foundInstances < instance);
 
-        return countCodePoints(string, 0, index) + 1;
+        return index + 1;
```
feilong-liu, line 284: index is already calculated with lastIndexOf of String type, doe not need to call countCodePoints again.

presto-main/src/main/java/com/facebook/presto/operator/scalar/StringFunctions.java
```diff
@@ -281,7 +281,7 @@ private static long stringPositionFromEnd(Slice string, Slice substring, long in
         }
         while (foundInstances < instance);
 
-        return countCodePoints(string, 0, index) + 1;
```
rschlussel, line 284: I think this behavior is intentional: see the note at the beginning here: https://prestodb.io/docs/current/functions/string.html#string-functions

feilong-liu: I see, but this leads to unintuitive behavior, for example

`assertFunction("STRPOS('\u4FE1\u5FF5,\u7231,\u5E0C\u671B', '\u7231')", BIGINT, 4L);` // strpos returns 4
while `assertFunction("STRRPOS('\u4FE1\u5FF5,\u7231,\u5E0C\u671B', '\u7231')", BIGINT, 2L);` // strrpos returns 2

strrpos and strpos return different result for same input (here it all matches the only \u7231 here).
And users are using substr(input, strrpos(input, delim)+1) which does not return the expected result (they expect to get the substr after the last delim)

rschlussel: oh i see. Thanks for clarifying. It's definitely a bug.

It looks like the difference between strpos and strrpos is that strpos operates on a slice (Slice.indexOf()),which seems to be about bytes, so then it needs to count codepoints after getting the index. strrpos converts to utf8 strings (String.lastIndexOf()), so the index returned  has a different meaning from the one used in strpos/slices. It already refers to code points. Could use a comment to explain the difference between these two otherwise very similar looking functions, but otherwise this change looks good.

By the way, could also take the slice -> String conversion outside of the loop.

rschlussel (approved):

presto-main/src/main/java/com/facebook/presto/operator/scalar/StringFunctions.java
feilong-liu: Added comments, and also move the slice -> String conversion outside of the loop

PR #24228, opened 2024-12-10
[native] Add stage id to arbitration priority for exec::Task

tanjialiang:
Make stage id as arbitration priority so that stages higher-up (smaller id) gets reclaimed first. This is proved to have ~30% gains on queries having multiple tasks reclaimed.
```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/TaskManager.cpp
+++ b/presto-native-execution/presto_cpp/main/TaskManager.cpp
@@ -523,7 +523,9 @@ std::unique_ptr<TaskInfo> TaskManager::createOrUpdateTaskImpl(
           planFragment,
           prestoTask->id.id(),
           std::move(queryCtx),
-          exec::Task::ExecutionMode::kParallel);
+          exec::Task::ExecutionMode::kParallel,
+          static_cast<exec::Consumer>(nullptr),
+          prestoTask->id.stageId());
       // TODO: move spill directory creation inside velox task execution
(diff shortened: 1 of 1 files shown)


presto-native-execution/presto_cpp/main/TaskManager.cpp
```diff
@@ -523,7 +523,9 @@ std::unique_ptr<TaskInfo> TaskManager::createOrUpdateTaskImpl(
           planFragment,
           prestoTask->id.id(),
           std::move(queryCtx),
-          exec::Task::ExecutionMode::kParallel);
+          exec::Task::ExecutionMode::kParallel,
+          static_cast<exec::Consumer>(nullptr),
```
xiaoxmeng, line 527: hmm, we shall actually place this ahead of consumer and after execution mode.

xiaoxmeng (approved):

aditi-pandit:
Thanks @tanjialiang. I have always wondered about the interaction between spilling order of different fragments of the same query, and this definitely is a step in the right direction !

Would it be possible to write some unit tests for this behavior ?

presto-native-execution/presto_cpp/main/TaskManager.cpp
tanjialiang: Sure. Let's make it a separate refactoring after this is done.

PR #23923, opened 2024-10-31
[Native] Fix table writer to use the actual storage format

kewang1024:
Coordinator decides data storage format in following logic:
1. Unpartitioned table use table format
2. Partitioned table
     - use table format if respect-table-format is set to true
     - use hive.storage-format if respect-table-format is false

The final decision is stored in actualStorageFormat of 
HiveOutputTableHandle. Switch to use actualStorageFormat 
for prestissimo.

tableStorageFormat and partitionStorageFormat in 
HiveOutputTableHandle would be deprecated later.

```
== NO RELEASE NOTE ==
```

PR #24248, opened 2024-12-12
[native] Enhance e2e test for table format for insert

kewang1024:
## Description
There are two types of table write, one is create table as select,
the other is insert.
After fixing both:
https://github.com/prestodb/presto/pull/23923
https://github.com/prestodb/presto/pull/24230

Add e2e tests for both createAsSelect and Insert.

```
== NO RELEASE NOTE ==
```


xiaoxmeng (approved):
@kewang1024 thanks for the improvement!

PR #24283, opened 2024-12-19
Support Hudi merged view files for partition path updates without compaction

codope:
## Description

Support Hudi merged view files for partition path updates without compaction. This is needed with Merge-on-Read Hudi tables when partition has been updated for a record and the table has not been compacted yet.

- Add a client and session config to specify tables (comma-separated `schemaName.tableName`) where this support is needed.
- Based on config, if the current tables needs to be listed using merged view, then do so in `HudiDirectoryLister`.
- Add a test for Merge-on-Read table with a record being updated from partition p1 to p2 and no compaction.
- Add some test fixtures to support above test scenario.

## Motivation and Context

Support Hudi merged view files for partition path updates without compaction. This is needed with Merge-on-Read Hudi tables when partition has been updated for a record and the table has not been compacted yet.

## Impact

Support Hudi merged view files for partition path updates without compaction. This is needed with Merge-on-Read Hudi tables when partition has been updated for a record and the table has not been compacted yet. As a result, even the read-optimized view of Merge-on-Read tables under partition path updates without compaction should not return any duplicates.

## Test Plan

- Add a test for Merge-on-Read table with a record being updated from partition p1 to p2 and no compaction.
- Add some test fixtures to support above test scenario.

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```


presto-hive/src/main/java/com/facebook/presto/hive/HiveClientConfig.java
```diff
@@ -1647,6 +1648,19 @@ public boolean isHudiMetadataEnabled()
         return this.hudiMetadataEnabled;
     }
 
+    @Config("hive.hudi-tables-use-merged-view")
+    @ConfigDescription("For Hudi tables prefer to fetch the list of files from the merged file system view")
```
ZacBlanco, line 1652: I think the description should be more clear about the type of value it's looking for. The type is a string, so for users it should be more explicit what the value should be. When I read this I initially thought it should be a boolean.

Based on looking at the tests it looks like it needs a dot-separated name of schema and table? If so, it should be clear about which tables should be added here

```suggestion
    @ConfigDescription("For Hudi tables, A comma-separated list in the form of <schema>.<table> which should prefer to fetch the list of files from the merged file system view")
```

presto-hive/src/main/java/com/facebook/presto/hive/HiveSessionProperties.java
```diff
@@ -608,6 +609,11 @@ public HiveSessionProperties(HiveClientConfig hiveClientConfig, OrcFileWriterCon
                         "For Hudi tables prefer to fetch the list of file names, sizes and other metadata from the internal metadata table rather than storage",
                         hiveClientConfig.isHudiMetadataEnabled(),
                         false),
+                stringProperty(
+                        HUDI_TABLES_USE_MERGED_VIEW,
+                        "For Hudi tables, use merged view to read data",
```
ZacBlanco, line 614: please update the description here as well

nsivabalan (commented):
can we add tests please

presto-hive/src/main/java/com/facebook/presto/hive/HiveClientConfig.java
codope: Good suggestion, thanks. 
I have updated the config docs at both places

codope:
> can we add tests please

There is already a test in `TestHudiDirectoryLister` to cover this scenario specifically.

codope:
@ZacBlanco Feedback is addressed. Please take a look again.

presto-hive/src/main/java/com/facebook/presto/hive/HudiDirectoryLister.java
```diff
@@ -68,6 +80,22 @@ public HudiDirectoryLister(Configuration conf, ConnectorSession session, Table t
...
+        if (latestCompactionInstantTime != null) {
+            this.latestInstant = latestCompactionInstantTime;
+        }
+        else {
+            this.latestInstant = metaClient.getActiveTimeline()
+                    .getCommitsTimeline()
+                    .filterCompletedInstants()
+                    .lastInstant()
+                    .map(HoodieInstant::getTimestamp).orElseThrow(() -> new RuntimeException("No active instant found"));
+        }
```
ZacBlanco, line 98: I feel like we should go with the fully functional approach rather than mixing imperative styles here since you're already using the optional type.

```suggestion
this.latestInstant = metaClient.getActiveTimeline()
                .getCommitsAndCompactionTimeline()
                .filterCompletedInstants()
                .filter(instant -> HoodieTableType.MERGE_ON_READ.equals(metaClient.getTableType()) && instant.getAction().equals(HoodieTimeline.COMPACTION_ACTION))
                .lastInstant()
                .map(HoodieInstant::getTimestamp).orElseGet(() -> metaClient.getActiveTimeline()
                        .getCommitsTimeline()
                        .filterCompletedInstants()
                        .lastInstant()
                        .map(HoodieInstant::getTimestamp).orElseThrow(() -> new RuntimeException("No active instant found")));
```

presto-hive/src/main/java/com/facebook/presto/hive/HudiDirectoryLister.java
```diff
@@ -111,15 +142,26 @@ public HudiFileInfoIterator(
...
         {
             String partition = FSUtils.getRelativePartitionPath(new Path(tablePath), directory);
             if (fileStatuses.isPresent()) {
                 fileSystemView.addFilesToView(fileStatuses.get());
                 this.hoodieBaseFileIterator = fileSystemView.fetchLatestBaseFiles(partition).iterator();
             }
             else {
-                this.hoodieBaseFileIterator = fileSystemView.getLatestBaseFiles(partition).iterator();
+                if (shouldUseMergedView) {
+                    Stream<FileSlice> fileSlices = HoodieTableType.MERGE_ON_READ.equals(tableType) ?
```
ZacBlanco, line 157: nit: can we static import `MERGE_ON_READ`

presto-hive/src/test/java/com/facebook/presto/hive/TestHudiDirectoryLister.java
```diff
@@ -107,6 +112,71 @@ private Table getMockTable()
...
+                        ImmutableMap.of()),
+                ImmutableList.of(),
+                ImmutableList.of(),
+                ImmutableMap.of(),
+                Optional.empty(),
+                Optional.empty());
+    }
+
+    @Test
+    public void testDirectoryListerForMORTableWithPartitionUpdates()
```
ZacBlanco, line 141: Can we also add a negative test to verify when the merged view is disabled/doesn't include the table name?

ZacBlanco (changes requested):
Thanks for the updates. The configurations look better to me now. 

I have one more concern, and that's the added files for these tests take up about 8MB. Not a whole lot, but for the health and size of the repository I would either like to generate the Hudi tables during the test, or minimize the size of added files to be < 1MiB.

presto-hive/src/test/java/com/facebook/presto/hive/TestHudiDirectoryLister.java
codope: added a negative test `testDirectoryListerForMORTableWithoutTableNames`

codope:
@ZacBlanco I have addressed your feedback. On the added resource files for the tests, currently we don't have a hudi writer integration. One solution is to zip the hudi testing data and unzip in tests. Overall size on disk would be less,  with a little bit of overhead during tests (probably similar or lesser than writing hudi tables). If you think that's ok, I can do that in a separate PR. Please let me know your thoughts.

ZacBlanco:
Thank you for the updates. It looks good to me now. On the table size aspect:

> currently we don't have a hudi writer integration. One solution is to zip the hudi testing data and unzip in tests. Overall size on disk would be less, with a little bit of overhead during tests (probably similar or lesser than writing hudi tables).

Have you tested the size of the zipped table? Also, we might even avoid zipping if we perform the minimum number of operations and make the size of the table as small as possible.

Also, did you try to minimize the number of operations and records on this table before including it in the PR? I'm not sure how you generated this table, but ideally we would generate a table with the smallest amount of data and metadata. e.g. This table should have one int-sized column and only a small number (<10?) records across the smallest number of inserts to (1-2?) alongside the partition path update.

> If you think that's ok, I can do that in a separate PR. Please let me know your thoughts.

These files will get pulled on every clone of the repository. Per-day Presto is seeing 1-2K clones. With the current table it adds that's up to 4GB additional data per day being pulled. 

It needs to be done in the same commit that the table is introduced, or else the extra data will still exist in the git history and will affect non-shallow pulls.

codope:
@ZacBlanco , sorry I was on leave for a month. Back now. I will try to keep test artifacts smaller. I can create a table with smaller data and but I need at least two commits to cover the scenario which is getting fixed here.

codope:
@ZacBlanco I have updated the test artifacts and rebased the code. The total size for test artifacts including metadata is now 1.1 MB (50% of previous commit). The test data has two fields: `id INT, partition STRING` and 4 records 2 in each partition.

ZacBlanco (approved):
Thank you for the updates - I appreciate the effort to make the table size smaller. Can you also squash your changes into a single commit?

codope:
@ZacBlanco Thanks for the approval. I have squashed and rebased. I see one unrelated test failing - `TestBrutalShutdown.testQueryRetryOnShutdown`. It seems to be flaky in CI environment because I ran locally a few times and it passed.

rschlussel (approved):
approving based on @ZacBlanco' review

PR #24287, opened 2024-12-19
Refactor ExpressionOptimizer

tdcmeehan:
## Description
Refactor `ExpressionOptimizer` to consistently return `RowExpression`.  Before, `ExpressionOptimizer` would return `RowExpression` only for the method which did not take in a variable resolver function.  I cannot discern a good reason for this.  By refactoring this interface and making them both return `RowExpression`, and documenting that the two methods are equivalent when an identity function is used for the variable resolver, it becomes possible to share code that uses both methods.

## Motivation and Context
This is to help address PR review feedback here: https://github.com/prestodb/presto/pull/24144#discussion_r1890678899

## Impact
Code refactoring.  Implementations of `ExpressionOptimizer` in the SPI will need to update their plugins to accommodate the changed API.

## Test Plan
Unit tests

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

SPI Changes
* Improve ExpressionOptimizer#optimize method with a variable resolver to return ``RowExpression``. :pr:`24287`
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/sql/relational/RowExpressionOptimizer.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/relational/RowExpressionOptimizer.java
@@ -46,9 +46,9 @@ public final class RowExpressionOptimizer
     }
 
     @Override
-    public Object optimize(RowExpression expression, Level level, ConnectorSession session, Function<VariableReferenceExpression, Object> variableResolver)
+    public RowExpression optimize(RowExpression expression, Level level, ConnectorSession session, Function<VariableReferenceExpression, Object> variableResolver)
     {
         RowExpressionInterpreter interpreter = new RowExpressionInterpreter(expression, metadata.getFunctionAndTypeManager(), session, level);
-        return interpreter.optimize(variableResolver::apply);
+        return toRowExpression(expression.getSourceLocation(), interpreter.optimize(variableResolver::apply), expression.getType());
     }
 }
--- a/presto-hive-common/src/main/java/com/facebook/presto/hive/rule/BaseSubfieldExtractionRewriter.java
+++ b/presto-hive-common/src/main/java/com/facebook/presto/hive/rule/BaseSubfieldExtractionRewriter.java
@@ -436,7 +436,7 @@ public abstract class BaseSubfieldExtractionRewriter
 
             // Skip pruning if evaluation fails in a recoverable way. Failing here can cause
             // spurious query failures for partitions that would otherwise be filtered out.
-            Object optimized = null;
+            RowExpression optimized;
             try {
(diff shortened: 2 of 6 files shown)


ZacBlanco (approved):

ZacBlanco:
Don't forget to update the release note with the correct PR #

hantangwangd (approved):
Thanks for this refactor, LGTM!

jaystarshot (approved):

presto-main/src/main/java/com/facebook/presto/sql/relational/RowExpressionOptimizer.java
```diff
@@ -46,9 +46,9 @@ public RowExpression optimize(RowExpression rowExpression, Level level, Connecto
     }
 
     @Override
-    public Object optimize(RowExpression expression, Level level, ConnectorSession session, Function<VariableReferenceExpression, Object> variableResolver)
+    public RowExpression optimize(RowExpression expression, Level level, ConnectorSession session, Function<VariableReferenceExpression, Object> variableResolver)
     {
         RowExpressionInterpreter interpreter = new RowExpressionInterpreter(expression, metadata.getFunctionAndTypeManager(), session, level);
-        return interpreter.optimize(variableResolver::apply);
+        return toRowExpression(expression.getSourceLocation(), interpreter.optimize(variableResolver::apply), expression.getType());
```
rschlussel, line 52: is it possible to change RowExpressionInterpreter to return a RowExpression instead of an object (seems like basically it requires wrapping any "values" in constant expressions, and then changing anything that calls it to expect that to happen).  Returning an Object that could be anything and leaving it to the caller be able to handle it  seems like a bad pattern to begin with.

rschlussel: but this change is still an improvement, so i'm fine with merging this first.

aaneja (approved):

presto-main/src/main/java/com/facebook/presto/sql/relational/RowExpressionOptimizer.java
tdcmeehan: I think it is, but it's not a straightforward refactor, and I'd like to get this PR in separately so it can address the feedback in #24144.

The idea that `RowExpressionInterpreter` should return object is I think just a relic of the fact that `RowExpressionInterpreter` is largely copied from `ExpressionInterpreter`.  One problematic thing that's copied over is lambda optimizations--if the interpreter is set to `EVALUATED`, then it may return a method handle.  As best I can tell, users of `RowExpressionInterpreter` never expect nor know what to do with a method handle, and I think it's just a relic from `ExpressionInterpreter` which actually may return method handles.

rschlussel (approved):

PR #24288, opened 2024-12-19
Assign event loop to tasks and run a task's methods in the same loop

shangm2:
## Description
1. this pr proposed to use a event loop to run task's methods so that multiple tasks can be assigned to the same event loop and a task's methods will only be executed on the same thread

## Motivation and Context
1. we saw a big number of tasks were created for certain queries which causes contention on the underlying synchronous queue
2. this pr proposed to use a event loop to run task's method so that multiple tasks can be assigned to the same event loop and a task's method will be executed on the same thread
3. some methods within http remote task does not need the "synchronized" keywords if we are certain they will be executed in the same thread.

## Test Plan
1. strobelight seems fine https://fburl.com/scuba/strobelight_java_asyncprofiler/on_demand/3qo6j1h9
Screenshot 2024-12-22 at 16 34 47
2. ran verifier successfully (screenshot is coming)

## Contributor checklist

- [ x] Please make sure your submission complies with our [contributing guide](https://github.com/prestodb/presto/blob/master/CONTRIBUTING.md), in particular [code style](https://github.com/prestodb/presto/blob/master/CONTRIBUTING.md#code-style) and [commit standards](https://github.com/prestodb/presto/blob/master/CONTRIBUTING.md#commit-standards).
- [ x] PR description addresses the issue accurately and concisely.  If the change is non-trivial, a GitHub Issue is referenced.
- [ x] Adequate tests were added if applicable.
- [ x] CI passed.

```
== RELEASE NOTES ==

General Changes
* Improve efficiency of coordinator when running a large number of tasks. :pr:`24288`
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/server/remotetask/ContinuousTaskStatusFetcher.java
+++ b/presto-main/src/main/java/com/facebook/presto/server/remotetask/ContinuousTaskStatusFetcher.java
@@ -13,7 +13,6 @@
  */
 package com.facebook.presto.server.remotetask;
 
-import com.facebook.airlift.concurrent.SetThreadName;
 import com.facebook.airlift.http.client.HttpClient;
 import com.facebook.airlift.http.client.Request;
 import com.facebook.airlift.http.client.ResponseHandler;
@@ -38,13 +37,9 @@ import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import io.airlift.units.Duration;
+import io.netty.channel.EventLoop;
 
-import javax.annotation.concurrent.GuardedBy;
-
-import java.util.concurrent.Executor;
-import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Consumer;
 
 import static com.facebook.airlift.http.client.HttpUriBuilder.uriBuilderFrom;
@@ -60,6 +55,7 @@ import static com.facebook.presto.server.thrift.ThriftCodecWrapper.unwrapThriftC
 import static com.facebook.presto.spi.StandardErrorCode.REMOTE_TASK_ERROR;
 import static com.facebook.presto.spi.StandardErrorCode.REMOTE_TASK_MISMATCH;
 import static com.facebook.presto.util.Failures.REMOTE_TASK_MISMATCH_ERROR;
+import static com.google.common.base.Verify.verify;
 import static io.airlift.units.Duration.nanosSince;
 import static java.lang.String.format;
 import static java.util.Objects.requireNonNull;
@@ -75,20 +71,16 @@ class ContinuousTaskStatusFetcher
     private final Codec<TaskStatus> taskStatusCodec;
 
     private final Duration refreshMaxWait;
-    private final Executor executor;
+    private final EventLoop taskEventLoop;
     private final HttpClient httpClient;
     private final RequestErrorTracker errorTracker;
     private final RemoteTaskStats stats;
     private final boolean binaryTransportEnabled;
     private final boolean thriftTransportEnabled;
     private final Protocol thriftProtocol;
-
-    private final AtomicLong currentRequestStartNanos = new AtomicLong();
-
-    @GuardedBy("this")
+    private long currentRequestStartNanos;
     private boolean running;
 
-    @GuardedBy("this")
     private ListenableFuture<BaseResponse<TaskStatus>> future;
 
     public ContinuousTaskStatusFetcher(
@@ -97,10 +89,9 @@ class ContinuousTaskStatusFetcher
             TaskStatus initialTaskStatus,
             Duration refreshMaxWait,
             Codec<TaskStatus> taskStatusCodec,
-            Executor executor,
+            EventLoop taskEventLoop,
             HttpClient httpClient,
             Duration maxErrorDuration,
-            ScheduledExecutorService errorScheduledExecutor,
             RemoteTaskStats stats,
             boolean binaryTransportEnabled,
             boolean thriftTransportEnabled,
@@ -110,23 +101,25 @@ class ContinuousTaskStatusFetcher
 
         this.taskId = requireNonNull(taskId, "taskId is null");
         this.onFail = requireNonNull(onFail, "onFail is null");
-        this.taskStatus = new StateMachine<>("task-" + taskId, executor, initialTaskStatus);
+        this.taskStatus = new StateMachine<>("task-" + taskId, taskEventLoop, initialTaskStatus);
 
         this.refreshMaxWait = requireNonNull(refreshMaxWait, "refreshMaxWait is null");
         this.taskStatusCodec = requireNonNull(taskStatusCodec, "taskStatusCodec is null");
 
-        this.executor = requireNonNull(executor, "executor is null");
+        this.taskEventLoop = requireNonNull(taskEventLoop, "taskEventLoop is null");
         this.httpClient = requireNonNull(httpClient, "httpClient is null");
 
-        this.errorTracker = taskRequestErrorTracker(taskId, initialTaskStatus.getSelf(), maxErrorDuration, errorScheduledExecutor, "getting task status");
+        this.errorTracker = taskRequestErrorTracker(taskId, initialTaskStatus.getSelf(), maxErrorDuration, taskEventLoop, "getting task status");
         this.stats = requireNonNull(stats, "stats is null");
         this.binaryTransportEnabled = binaryTransportEnabled;
         this.thriftTransportEnabled = thriftTransportEnabled;
         this.thriftProtocol = requireNonNull(thriftProtocol, "thriftProtocol is null");
     }
 
-    public synchronized void start()
+    public void start()
     {
+        verify(taskEventLoop.inEventLoop());
+
         if (running) {
             // already running
             return;
@@ -135,8 +128,10 @@ class ContinuousTaskStatusFetcher
         scheduleNextRequest();
     }
 
-    public synchronized void stop()
+    public void stop()
     {
+        verify(taskEventLoop.inEventLoop());
+
         running = false;
         if (future != null) {
             // do not terminate if the request is already running to avoid closing pooled connections
@@ -145,8 +140,10 @@ class ContinuousTaskStatusFetcher
         }
     }
 
-    private synchronized void scheduleNextRequest()
+    private void scheduleNextRequest()
     {
+        verify(taskEventLoop.inEventLoop());
+
         // stopped or done?
         TaskStatus taskStatus = getTaskStatus();
         if (!running || taskStatus.getState().isDone()) {
@@ -163,7 +160,7 @@ class ContinuousTaskStatusFetcher
         // if throttled due to error, asynchronously wait for timeout and try again
         ListenableFuture<?> errorRateLimit = errorTracker.acquireRequestPermit();
         if (!errorRateLimit.isDone()) {
-            errorRateLimit.addListener(this::scheduleNextRequest, executor);
+            errorRateLimit.addListener(this::scheduleNextRequest, taskEventLoop);
             return;
         }
 
@@ -189,7 +186,7 @@ class ContinuousTaskStatusFetcher
 
         errorTracker.startRequest();
         future = httpClient.executeAsync(request, responseHandler);
-        currentRequestStartNanos.set(System.nanoTime());
+        currentRequestStartNanos = System.nanoTime();
         FutureCallback callback;
         if (thriftTransportEnabled) {
             callback = new ThriftHttpResponseHandler(this, request.getUri(), stats.getHttpResponseStats(), REMOTE_TASK_ERROR);
@@ -201,7 +198,7 @@ class ContinuousTaskStatusFetcher
         Futures.addCallback(
                 future,
                 callback,
-                executor);
+                taskEventLoop);
     }
 
     TaskStatus getTaskStatus()
@@ -212,59 +209,62 @@ class ContinuousTaskStatusFetcher
     @Override
     public void success(TaskStatus value)
     {
-        try (SetThreadName ignored = new SetThreadName("ContinuousTaskStatusFetcher-%s", taskId)) {
-            updateStats(currentRequestStartNanos.get());
-            try {
-                updateTaskStatus(value);
-                errorTracker.requestSucceeded();
-            }
-            finally {
-                scheduleNextRequest();
-            }
+        verify(taskEventLoop.inEventLoop());
+
+        updateStats(currentRequestStartNanos);
+        try {
+            updateTaskStatus(value);
+            errorTracker.requestSucceeded();
+        }
+        finally {
+            scheduleNextRequest();
         }
     }
 
     @Override
     public void failed(Throwable cause)
     {
-        try (SetThreadName ignored = new SetThreadName("ContinuousTaskStatusFetcher-%s", taskId)) {
-            updateStats(currentRequestStartNanos.get());
-            try {
-                // if task not already done, record error
-                TaskStatus taskStatus = getTaskStatus();
-                if (!taskStatus.getState().isDone()) {
-                    errorTracker.requestFailed(cause);
-                }
+        verify(taskEventLoop.inEventLoop());
+
+        updateStats(currentRequestStartNanos);
+        try {
+            // if task not already done, record error
+            TaskStatus taskStatus = getTaskStatus();
+            if (!taskStatus.getState().isDone()) {
+                errorTracker.requestFailed(cause);
             }
-            catch (Error e) {
-                onFail.accept(e);
-                throw e;
-            }
-            catch (RuntimeException e) {
-                onFail.accept(e);
-            }
-            finally {
-                scheduleNextRequest();
-            }
+        }
+        catch (Error e) {
+            onFail.accept(e);
+            throw e;
+        }
+        catch (RuntimeException e) {
+            onFail.accept(e);
+        }
+        finally {
+            scheduleNextRequest();
         }
     }
 
     @Override
     public void fatal(Throwable cause)
     {
-        try (SetThreadName ignored = new SetThreadName("ContinuousTaskStatusFetcher-%s", taskId)) {
-            updateStats(currentRequestStartNanos.get());
-            onFail.accept(cause);
-        }
+        verify(taskEventLoop.inEventLoop());
+
+        updateStats(currentRequestStartNanos);
+        onFail.accept(cause);
     }
 
     void updateTaskStatus(TaskStatus newValue)
     {
+        verify(taskEventLoop.inEventLoop());
+
         // change to new value if old value is not changed and new value has a newer version
         AtomicBoolean taskMismatch = new AtomicBoolean();
         taskStatus.setIf(newValue, oldValue -> {
             // did the task instance id change
-            boolean isEmpty = oldValue.getTaskInstanceIdLeastSignificantBits() == 0 && oldValue.getTaskInstanceIdMostSignificantBits() == 0;
+            boolean isEmpty = (oldValue.getTaskInstanceIdLeastSignificantBits() == 0 && oldValue.getTaskInstanceIdMostSignificantBits() == 0)
+                    || (newValue.getTaskInstanceIdLeastSignificantBits() == 0 && newValue.getTaskInstanceIdMostSignificantBits() == 0);
             if (!isEmpty &&
                     !(oldValue.getTaskInstanceIdLeastSignificantBits() == newValue.getTaskInstanceIdLeastSignificantBits() &&
                             oldValue.getTaskInstanceIdMostSignificantBits() == newValue.getTaskInstanceIdMostSignificantBits())) {
@@ -291,11 +291,6 @@ class ContinuousTaskStatusFetcher
         }
     }
 
-    public synchronized boolean isRunning()
-    {
-        return running;
-    }
-
     /**
      * Listener is always notified asynchronously using a dedicated notification thread pool so, care should
      * be taken to avoid leaking {@code this} when adding a listener in a constructor. Additionally, it is
@@ -308,6 +303,8 @@ class ContinuousTaskStatusFetcher
 
     private void updateStats(long currentRequestStartNanos)
     {
+        verify(taskEventLoop.inEventLoop());
+
         stats.statusRoundTripMillis(nanosSince(currentRequestStartNanos).toMillis());
     }
 }
--- a/presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTask.java
+++ b/presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTask.java
@@ -13,7 +13,6 @@
  */
 package com.facebook.presto.server.remotetask;
 
-import com.facebook.airlift.concurrent.SetThreadName;
 import com.facebook.airlift.http.client.HttpClient;
 import com.facebook.airlift.http.client.HttpUriBuilder;
 import com.facebook.airlift.http.client.Request;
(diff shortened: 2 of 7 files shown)


presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTaskFactory.java
```diff
@@ -101,6 +104,9 @@ public class HttpRemoteTaskFactory
     private final MetadataManager metadataManager;
     private final QueryManager queryManager;
     private final DecayCounter taskUpdateRequestSize;
+    // TODO: use config file to set this value
+    private final EventLoopGroup eventLoopGroup = new DefaultEventLoopGroup(2000,
```
arhimondr, line 108: Probably not worth setting it to a value higher then the number of CPU's available on the system

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTaskFactory.java
```diff
@@ -257,6 +264,7 @@ public RemoteTask createRemoteTask(
                 taskUpdateRequestSize,
                 handleResolver,
                 connectorTypeSerdeManager,
-                schedulerStatsTracker);
+                schedulerStatsTracker,
+                eventLoopGroup.next());
```
arhimondr, line 268: Does the task still need an `executor`? Also worth checking usages of `updateScheduledExecutor`/`errorScheduledExecutor` and see if all are proper

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTask.java
```diff
@@ -453,18 +459,20 @@ public URI getRemoteTaskLocation()
...
     {
-        try (SetThreadName ignored = new SetThreadName("HttpRemoteTask-%s", taskId)) {
-            // to start we just need to trigger an update
-            started.set(true);
-            scheduleUpdate();
-
-            taskStatusFetcher.start();
-            taskInfoFetcher.start();
-        }
+        taskEventLoop.execute(() -> {
```
arhimondr, line 462: Does this change remove all synchronized sections?

arhimondr (commented):
Nice!

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTaskFactory.java
shangm2: Good point!

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTaskFactory.java
arhimondr: The `executor` is used to run `StateChangeListener`s attached. This is another thing that seems to be broken. The `StateChangeListener` has to have an executor attached. It shouldn't be a responsibility of the event producer to provide a thread where to run a foreign callback. Instead a callback should be provided with an executor to run by the creator of a callback.

shangm2: Totally make sense. Learned a lot from your precious comment. Really appreciate it. Let me update the code.

tdcmeehan:
Very nice change.  So I understand, is the effect of this to reduce the number of IO threads related to task management from # of tasks to # of queries?

arhimondr:
@tdcmeehan Yeah, the idea is to reduce the number of event processing threads and remove synchronization with the eventual goal of coordinator supporting more tasks running concurrently. Currently we are seeing scalability issues when the number of tasks grows beyond 30-50K. Dispatching events on standard Java thread pools becomes a problem.

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTask.java
```diff
@@ -688,29 +693,33 @@ public void addFinalTaskInfoListener(StateChangeListener<TaskInfo> stateChangeLi
...
+                checkArgument(weightThreshold == whenSplitQueueHasSpaceThreshold.getAsLong(), "Multiple split queue space notification thresholds not supported");
+            }
+            else {
+                whenSplitQueueHasSpaceThreshold = OptionalLong.of(weightThreshold);
+                updateSplitQueueSpace();
+            }
+            if (splitQueueHasSpace) {
+                future.set(null);
+            }
+            whenSplitQueueHasSpace.createNewListener().addListener(() -> future.set(null), directExecutor());
```
shangm2, line 710: anchor

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTask.java
shangm2: @arhimondr yes, all synchronized sections are now async. The only thing I am not certain is this change https://github.com/prestodb/presto/pull/24288/files#r1894673304. The original synchronized implementation might return a future but now we will need to notify the listener in a async way. I wonder what your thought is on this. So far, I am testing the code on verifier and things look good.

amitkdutta:
@shangm2 This is a great change. Please add details in testing section if you have performed any shadow tests. Will be great if we can run it through problematic traffic and show before/after with this change.

shangm2:
> @shangm2 This is a great change. Please add details in testing section if you have performed any shadow tests. Will be great if we can run it through problematic traffic and show before/after with this change.

Yep. I am making screenshots about the result of running verifier agains this change. One issue is this pr changes the way we use thread pool and I am not sure which metrics we can use to compare before and after the change. What do you think @arhimondr ?  Can you suggest some metrics I can use to confirm the change? Thanks. I did run some strobelight and did not see anything suspicious: https://fburl.com/scuba/strobelight_java_asyncprofiler/on_demand/3qo6j1h9
Screenshot 2024-12-22 at 16 34 47

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTask.java
```diff
@@ -234,6 +230,8 @@ public final class HttpRemoteTask
     private final DecayCounter taskUpdateRequestSize;
     private final SchedulerStatsTracker schedulerStatsTracker;
 
+    private final EventLoop taskEventLoop;
```
arhimondr, line 233: - Atomic fields no longer need to be atomic (`nextSplitId`, `needsUpdate`, `sendPlan`, `started`, `aborted`, `outputBuffers`). It is better to remove the atomic as it may suggest that the variable is accesses from multiple threads concurrently.
- Remove `GuardedBy` annotations, as the accesses are no longer protected by a mutex. Instead you may want to add a comment at the beginning of the class and explain what is going on in this class (e.g.: that all interactions with the TaskState have to be done via an event loop)

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTask.java
```diff
@@ -367,10 +361,10 @@ public HttpRemoteTask(
                     initialTask.getTaskStatus(),
                     taskStatusRefreshMaxWait,
                     taskStatusCodec,
-                    executor,
+                    taskEventLoop,
```
arhimondr, line 364: Maybe just pass it once and call it `taskEventLoop`

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTask.java
```diff
@@ -386,9 +380,9 @@ public HttpRemoteTask(
                     metadataUpdatesCodec,
                     maxErrorDuration,
                     summarizeTaskInfo,
-                    executor,
-                    updateScheduledExecutor,
-                    errorScheduledExecutor,
+                    taskEventLoop,
```
arhimondr, line 383: ditto

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTask.java
```diff
@@ -604,7 +609,7 @@ public void onFailure(Throwable failedReason)
                     doRemoveRemoteSource(errorTracker, request, future);
                 }
                 else {
-                    errorRateLimit.addListener(() -> doRemoveRemoteSource(errorTracker, request, future), errorScheduledExecutor);
+                    errorRateLimit.addListener(() -> doRemoveRemoteSource(errorTracker, request, future), taskEventLoop);
```
arhimondr, line 612: Run the callback itself in the event loop as well

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTask.java
```diff
@@ -453,18 +448,20 @@ public URI getRemoteTaskLocation()
...
-        try (SetThreadName ignored = new SetThreadName("HttpRemoteTask-%s", taskId)) {
-            // to start we just need to trigger an update
-            started.set(true);
-            scheduleUpdate();
-
-            taskStatusFetcher.start();
-            taskInfoFetcher.start();
-        }
+        taskEventLoop.execute(() -> {
+            try (SetThreadName ignored = new SetThreadName("HttpRemoteTask-%s", taskId)) {
```
arhimondr, line 452: I would recommend not to override the thread name (here and in other places). It is easier to diagnose threading related issues when the threads are named consistently (e.g.: allowing you to find all threads belonging to a certain thread pool and all related events for it)

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTask.java
```diff
@@ -367,10 +361,10 @@ public HttpRemoteTask(
                     initialTask.getTaskStatus(),
                     taskStatusRefreshMaxWait,
                     taskStatusCodec,
-                    executor,
+                    taskEventLoop,
                     httpClient,
                     maxErrorDuration,
-                    errorScheduledExecutor,
+                    taskEventLoop,
```
arhimondr, line 367: Let's do the same exercise (removing synchronization) for the `ContinuousTaskStatusFetcher` and the `TaskInfoFetcher`

arhimondr:
It would be great if we could stress test this change (run many queries with many stages concurrently) and see if there are any major bottlenecks by profiling the event loop executor for both on-cpu and off-cpu activity.

arhimondr (commented):
Love the change. Thank you for working on it

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTask.java
```diff
@@ -832,103 +843,105 @@ private void onFailureTaskInfo(
...
+            taskEventLoop.schedule(() -> doScheduleAsyncCleanupRequest(cleanupBackoff, request, action), delayNanos, NANOSECONDS);
         }
     }
 
-    private synchronized void scheduleUpdate()
+    private void scheduleUpdate()
     {
         taskUpdateTimeline.add(System.nanoTime());
-        executor.execute(this::sendUpdate);
+        taskEventLoop.execute(this::sendUpdate);
```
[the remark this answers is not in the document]
shangm2: nice catch. thanks!

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTask.java
shangm2: Good point. Updating the code.

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTask.java
shangm2: Sure. I was trying not to modify the code in other files and keep this pr straightforward but looks like TaskInfoFetcher and ContinuousTaskStatusFetcher are only being used here. Let me update them as well.

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTask.java
shangm2: Good point. Thanks for catching this!

shangm2:
> It would be great if we could stress test this change (run many queries with many stages concurrently) and see if there are any major bottlenecks by profiling the event loop executor for both on-cpu and off-cpu activity.

Noted. Let me resolve the comments and run tableau queries against the change.

steveburnett:
I don't think this needs a release note - I'm making this statement for others to weigh in in case I'm overlooking something. 

If so, please update the Release Notes section of this PR to 

```
== NO RELEASE NOTE ==
```

presto-main/src/main/java/com/facebook/presto/server/remotetask/ContinuousTaskStatusFetcher.java
```diff
@@ -110,22 +100,22 @@ public ContinuousTaskStatusFetcher(
...
-        this.errorTracker = taskRequestErrorTracker(taskId, initialTaskStatus.getSelf(), maxErrorDuration, errorScheduledExecutor, "getting task status");
+        this.errorTracker = taskRequestErrorTracker(taskId, initialTaskStatus.getSelf(), maxErrorDuration, taskEventLoop, "getting task status");
         this.stats = requireNonNull(stats, "stats is null");
         this.binaryTransportEnabled = binaryTransportEnabled;
         this.thriftTransportEnabled = thriftTransportEnabled;
         this.thriftProtocol = requireNonNull(thriftProtocol, "thriftProtocol is null");
     }
 
-    public synchronized void start()
+    public void start()
```
arhimondr, line 118: I would recommend adding a `verify(taskEventLoop.inEventLoop());` check to all methods assumed to be run in event loop. This will help with prevention of hard to discover bugs and make code self documented.

presto-main/src/main/java/com/facebook/presto/server/remotetask/ContinuousTaskStatusFetcher.java
```diff
@@ -212,50 +202,44 @@ TaskStatus getTaskStatus()
...
     {
-        try (SetThreadName ignored = new SetThreadName("ContinuousTaskStatusFetcher-%s", taskId)) {
-            updateStats(currentRequestStartNanos.get());
-            onFail.accept(cause);
-        }
+        updateStats(currentRequestStartNanos);
+        onFail.accept(cause);
     }
 
     void updateTaskStatus(TaskStatus newValue)
```
arhimondr, line 245: I remember seeing a race condition when the `newValue` may have `taskInstanceIdLeastSignificantBits` and `taskInstanceIdMostSignificantBits` (for example when task updates fails at the very beginning) causing the tests sporadically fail on the CI. You may want to change the condition to 

```
boolean isEmpty = (oldValue.getTaskInstanceIdLeastSignificantBits() == 0 && oldValue.getTaskInstanceIdMostSignificantBits() == 0)
                    || (newValue.getTaskInstanceIdLeastSignificantBits() == 0 && newValue.getTaskInstanceIdMostSignificantBits() == 0);
```

presto-main/src/main/java/com/facebook/presto/server/remotetask/ContinuousTaskStatusFetcher.java
```diff
@@ -291,7 +275,7 @@ void updateTaskStatus(TaskStatus newValue)
         }
     }
 
-    public synchronized boolean isRunning()
+    public boolean isRunning()
```
arhimondr, line 278: nit: unused method, we can remove it

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTask.java
```diff
@@ -298,121 +292,116 @@ public HttpRemoteTask(
...
-            long pendingSourceSplitsWeight = 0;
-            for (PlanNodeId planNodeId : planFragment.getTableScanSchedulingOrder()) {
-                Collection<Split> tableScanSplits = initialSplits.get(planNodeId);
-                if (tableScanSplits != null && !tableScanSplits.isEmpty()) {
-                    pendingSourceSplitCount += tableScanSplits.size();
-                    pendingSourceSplitsWeight = addExact(pendingSourceSplitsWeight, SplitWeight.rawValueSum(tableScanSplits, Split::getSplitWeight));
-                }
+            else {
+                updateTaskStats();
+                taskEventLoop.execute(this::updateSplitQueueSpace);
```
arhimondr, line 398: This callback is run within the event loop. Maybe just add the `verify(taskEventLoop.inEventLoop());` check at the beginning of the method and replace the `taskEventLoop.execute(this::updateSplitQueueSpace)` with `updateSplitQueueSpace()` to make it easier to reason about.

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTask.java
```diff
@@ -688,29 +693,33 @@ public void addFinalTaskInfoListener(StateChangeListener<TaskInfo> stateChangeLi
...
+        SettableFuture<?> future = SettableFuture.create();
+        taskEventLoop.execute(() -> {
+            if (whenSplitQueueHasSpaceThreshold.isPresent()) {
+                checkArgument(weightThreshold == whenSplitQueueHasSpaceThreshold.getAsLong(), "Multiple split queue space notification thresholds not supported");
+            }
+            else {
+                whenSplitQueueHasSpaceThreshold = OptionalLong.of(weightThreshold);
+                updateSplitQueueSpace();
+            }
+            if (splitQueueHasSpace) {
```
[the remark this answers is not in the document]
arhimondr: You must've forgotten to address it / commit it. Since `splitQueueHasSpace` is now `volatile` please add a 

```
+        if (splitQueueHasSpace) {
+            return immediateVoidFuture();
+        }
```

at the beginning of the `whenSplitQueueHasSpace` method

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTask.java
```diff
@@ -729,40 +728,42 @@ private void updateTaskStats()
...
-            PlanNodeId planNodeId = source.getPlanNodeId();
-            boolean isTableScanSource = tableScanPlanNodeIds.contains(planNodeId);
-            int removed = 0;
-            long removedWeight = 0;
-            for (ScheduledSplit split : source.getSplits()) {
-                if (pendingSplits.remove(planNodeId, split)) {
-                    if (isTableScanSource) {
-                        removed++;
-                        removedWeight = addExact(removedWeight, split.getSplit().getSplitWeight().getRawValue());
+        taskEventLoop.execute(() -> {
```
arhimondr, line 733: nit: `processTaskUpdate` is always called from an event loop. Maybe replace the `taskEventLoop.execute(() -> {` with a simple assert `verify(taskEventLoop.inEventLoop());` at the beginning of the method

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTask.java
```diff
@@ -832,103 +833,105 @@ private void onFailureTaskInfo(
...
+                    outputBuffers,
+                    writeInfo);
+            long serializeStartCpuTimeNanos = THREAD_MX_BEAN.getCurrentThreadCpuTime();
+            byte[] taskUpdateRequestJson = taskUpdateRequestCodec.toBytes(updateRequest);
+            schedulerStatsTracker.recordTaskUpdateSerializedCpuTime(THREAD_MX_BEAN.getCurrentThreadCpuTime() - serializeStartCpuTimeNanos);
+
+            taskUpdateRequestSize.add(taskUpdateRequestJson.length);
+
+            if (taskUpdateRequestJson.length > maxTaskUpdateSizeInBytes) {
+                failTask(new PrestoException(EXCEEDED_TASK_UPDATE_SIZE_LIMIT, getExceededTaskUpdateSizeMessage(taskUpdateRequestJson)));
```
arhimondr, line 890: `return` after `failTask`

arhimondr (commented):
LGTM % CI failures, small comments

I checked CI failures and it looks like the tests are broken.

Maybe add an explicit wait instead of `waitUntilIdle` magic. 

For example you can explicitly discover when the task is finished by doing something like

```
+        SettableFuture<Void> finished = SettableFuture.create();
+        remoteTask.addStateChangeListener(status -> {
+            if (status.getState().isDone()) {
+                finished.set(null);
+            }
+        });
+        finished.get();
```

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTask.java
shangm2: Yep. Thanks a lot for catching this.

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTask.java
shangm2: OMG. I was trying to find all the places where I overuse the taskEventLoop.execute but still miss this one. Thanks a ton!

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTask.java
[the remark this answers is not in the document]
shangm2: Ahh, now I understand what you meant by "keep the shortcut". Learnt a lot. Really appreciate all the guidance.

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTask.java
shangm2: This totally makes sense.

presto-main/src/main/java/com/facebook/presto/server/remotetask/ContinuousTaskStatusFetcher.java
shangm2: Thanks for sharing.

presto-main/src/main/java/com/facebook/presto/server/remotetask/ContinuousTaskStatusFetcher.java
shangm2: Good point. Thanks a ton for the suggestion. Learned a lot.

arhimondr (approved):

tdcmeehan (commented):
Please add a release note as this is a relevant improvement that operators of Presto may want to know about.  Maybe something along the lines of

```
* Improve efficiency of coordinator when running a large number of tasks.
```

What do you think?

steveburnett:
Draft release note entry suggestion:

```
== RELEASE NOTES ==

General Changes
* * Improve efficiency of coordinator when running a large number of tasks. :pr:`24288`
```

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTaskFactory.java
```diff
@@ -101,6 +88,8 @@ public class HttpRemoteTaskFactory
     private final MetadataManager metadataManager;
     private final QueryManager queryManager;
     private final DecayCounter taskUpdateRequestSize;
+    private final EventLoopGroup eventLoopGroup = new DefaultEventLoopGroup(Runtime.getRuntime().availableProcessors(),
+            new ThreadFactoryBuilder().setNameFormat("task-event-loop-%s").setDaemon(true).build());
```
tdcmeehan, line 92: Can you explain the reasoning for choosing `DefaultEventLoopGroup` instead of other implementations of `EventLoopGroup`?

```suggestion
    private final EventLoopGroup eventLoopGroup = new DefaultEventLoopGroup(
            Runtime.getRuntime().availableProcessors(),
            new ThreadFactoryBuilder().setNameFormat("task-event-loop-%s").setDaemon(true).build());
```

shangm2: Absolutely. Netty provides a few event loop group implementations. A quick comparison among them could be: 

- NioEventLoopGroup: Most commonly used, best for TCP/IP socket communications
- EpollEventLoopGroup: Linux-specific implementation, better performance than NIO on Linux systems
- KQueueEventLoopGroup: MacOS/BSD-specific implementation, better performance on MacOS/BSD systems
- DefaultEventLoopGroup: For non-network operations, used for local event processing
- LocalEventLoopGroup: Optimized for local transport (in-VM pipe communication)

In our case, the network async calls are already being handled by http client and this event loop is primarily being used for task scheduling and coordination (method calls, state management, etc.). So defaultEventLoopGroup might be a better choice.  Please let me know if this makes sense and thanks a lot for the comments!

PR #24301, opened 2024-12-26
[native] Fix the shuffle hanging problem when fetch data from presto …

xiaoxmeng:
Presto java worker doesn't set next token in response for a get data size request
and it simply returns the buffered data from the source. The Prestissimo worker
updates 'sequence_' with ack sequence encoded by next token and uses it as data
stream offset to fetch data from the source. If the next token str is not set in
the data response handler header, then ack sequence is set to zero which resets the
'sequence_' to zero, and the rollback of 'sequence_' can cause data hanging problem
when Prestissimo worker fetch data from a Presto java worker (which can only happen
in the case that Presto java worker run by the coordinator as some metadata operation
can only execute on the coordinator) from the reset 'sequence_'.

This PR fixes the issue by avoiding updating 'sequence_' if next token is not set
in data response handler.
```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/PrestoExchangeSource.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoExchangeSource.cpp
@@ -265,7 +265,7 @@ void PrestoExchangeSource::processDataResponse(
   VELOX_CHECK(
       !headers->getIsChunked(),
       "Chunked http transferring encoding is not supported.");
-  uint64_t contentLength =
+  const uint64_t contentLength =
       atol(headers->getHeaders()
                .getSingleOrEmpty(proxygen::HTTP_HEADER_CONTENT_LENGTH)
                .c_str());
@@ -291,10 +291,19 @@ void PrestoExchangeSource::processDataResponse(
     }
   }
 
-  int64_t ackSequence =
-      atol(headers->getHeaders()
-               .getSingleOrEmpty(protocol::PRESTO_PAGE_NEXT_TOKEN_HEADER)
-               .c_str());
+  std::optional<int64_t> ackSequenceOpt;
+  const auto nextTokenStr = headers->getHeaders().getSingleOrEmpty(
+      protocol::PRESTO_PAGE_NEXT_TOKEN_HEADER);
+  if (!nextTokenStr.empty()) {
+    // NOTE: when get data size from Presto coordinator, it might not set next
+    // token so we shouldn't update 'sequence_' if it is empty. Otherwise,
+    // 'sequence_' gets reset and we can't fetch any data from the source with
+    // the rolled back 'sequence_'.
+    ackSequenceOpt = atol(nextTokenStr.c_str());
+  } else {
+    VELOX_CHECK_EQ(
+        contentLength, 0, "next token is not set in non-empty data response");
+  }
 
   std::unique_ptr<exec::SerializedPage> page;
   const bool empty = response->empty();
@@ -357,7 +366,9 @@ void PrestoExchangeSource::processDataResponse(
       queue_->enqueueLocked(nullptr, queuePromises);
     }
 
-    sequence_ = ackSequence;
+    if (ackSequenceOpt.has_value()) {
+      sequence_ = ackSequenceOpt.value();
+    }
     requestPending_ = false;
     requestPromise = std::move(promise_);
   }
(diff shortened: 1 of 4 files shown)


presto-native-execution/presto_cpp/main/PrestoExchangeSource.cpp
```diff
@@ -291,10 +291,19 @@ void PrestoExchangeSource::processDataResponse(
...
+      protocol::PRESTO_PAGE_NEXT_TOKEN_HEADER);
+  if (!nextTokenStr.empty()) {
+    // NOTE: when get data size from Presto coordinator, it might not set next
+    // token so we shouldn't update 'sequence_' if it is empty. Otherwise,
+    // 'sequence_' gets reset and we can't fetch any data from the source with
+    // the rolled back 'sequence_'.
+    ackSequenceOpt = atol(nextTokenStr.c_str());
+  } else {
+    VELOX_CHECK_EQ(
+        contentLength, 0, "next token is not set non-empty data response");
```
tanjialiang, line 305: ... is not set for non-empty ...

tanjialiang: By having this check, we are not supporting data fetching from java?

presto-native-execution/presto_cpp/main/PrestoExchangeSource.cpp
```diff
@@ -291,10 +291,19 @@ void PrestoExchangeSource::processDataResponse(
     }
   }
 
-  int64_t ackSequence =
-      atol(headers->getHeaders()
-               .getSingleOrEmpty(protocol::PRESTO_PAGE_NEXT_TOKEN_HEADER)
-               .c_str());
+  std::optional<int64_t> ackSequenceOpt;
```
tanjialiang, line 294: If we are not supporting data fetching from java, then we can get away by not using std::optional. The token check is performed anyways in the else block so whenever receiving data from coordinator it will not continue execution.

tanjialiang (approved):
Thanks % for some minor

presto-native-execution/presto_cpp/main/PrestoExchangeSource.cpp
```diff
@@ -357,7 +366,9 @@ void PrestoExchangeSource::processDataResponse(
       queue_->enqueueLocked(nullptr, queuePromises);
     }
 
-    sequence_ = ackSequence;
+    if (ackSequenceOpt.has_value()) {
+      sequence_ = ackSequenceOpt.value();
+    }
```
tanjialiang, line 371: Have an else block and perform some finish checks?

presto-native-execution/presto_cpp/main/PrestoExchangeSource.cpp
xiaoxmeng: We support and only java won't set next token str in get data size request with empty data.

presto-native-execution/presto_cpp/main/PrestoExchangeSource.cpp
xiaoxmeng: We also see empty sequence in case of get data size. Get data size and get data share the same response handler.

xiaoxmeng (commented):
@tanjialiang thanks!

tanjialiang (approved):

aditi-pandit:
@amitkdutta, @xiaoxmeng, @tanjialiang : The PR description says this fix is applicable from Prestissimo worker would fetch from a Java worker which could be present for some metadata operation at the co-ordinator.

Can you give an example of some SQL that does this ? 

I'm aware of SystemConnectorQueries, but unsure if they need a full Java worker.

amitkdutta:
@aditi-pandit Here is an example:

```
select 
  count(*), 
  checksum(data_ds) AS data_ds, 
  checksum(data_ts) AS data_ts, 
  checksum(platform_id), 
  checksum(resource_type_id), 
  checksum(agg_level) AS agg_level 
from 
  (
    SELECT 
      ds AS data_ds, 
      ts AS data_ts, 
      platform_id, 
      resource_type_id, 
      agg_level AS agg_level 
    FROM 
      "some_table$partitions" 
    WHERE 
      ds = '2024-12-14'
  )
```

It was with columns like `$partitions`, but a bit complex query. Simple `select * from "some_table$partitions"` was working though.

PR #24309, opened 2024-12-31
[Iceberg] Fix argument order in bucket transforms

ZacBlanco:
## Description

Swap arguments to their correct positions in the `bucketValues` function.

## Motivation and Context

Argument ordering is wrong and can cause incorrect data returned in queries on bucketed tables. This does not affect writing of the tables tables

## Impact

- No impact to existing users. Queries which triggered improper behavior before should now return data correctly. The codepath for this only occurred in the `IcebergPlanOptimizer`. It's unlikely that any users encountered this as it seems the codepath required to hit this error is gated by a conditional which prevents us from utilizing the "valueTransform" with bucketed partitioning functions. [code](https://github.com/prestodb/presto/blob/65d252c1d0d3cc8fb3977a197b726aa99df02bff/presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergPlanOptimizer.java#L335-L338).

## Test Plan

- Existing tests

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/PartitionTransforms.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/PartitionTransforms.java
@@ -249,7 +249,7 @@ public final class PartitionTransforms
 
     private static int bucketValueInteger(Block block, int position, int count)
     {
-        return bucketValue(block, count, position, pos -> bucketHash(INTEGER.getLong(block, pos)));
+        return bucketValue(block, position, count, pos -> bucketHash(INTEGER.getLong(block, pos)));
     }
 
     private static Block bucketBigint(Block block, int count)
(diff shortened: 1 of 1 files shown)


hantangwangd (approved):

ZacBlanco:
@hantangwangd FYI after further analysis I don't believe it is actually possible to trigger this codepath via a test because the error occurs only for the `valueTransform` function. The codepath which actually calls the `valueTransform` function prevents the bucket transform from being called [here](https://github.com/prestodb/presto/blob/65d252c1d0d3cc8fb3977a197b726aa99df02bff/presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergPlanOptimizer.java#L335-L338)

So we won't be able to write a test for this. It is more like a cosmetic change.

hantangwangd:
@ZacBlanco Thanks for the analysis. You are right that the constructed bucket `ValueTransform` is not called anywhere since enforced pushdown optimization is not applicable for bucket transform.

PR #24341, opened 2025-01-09
Use StringBuilder for some toString methods instead of string concatenation 

shangm2:
## Description
1. Passing a string builder along while creating ids (those with embedded ids) so that only one string builder will be created along the way

## Motivation and Context
1. While creating ids, say taskId, a string builder will be created to do so but this taskId will need a stageExecutionId which will trigger the creation of another string builder and stageExecutionId needs stagesId and the same will happen again.
2. We see some crazy cpu usage for toString method due to string concatenation, T212253302
image

## Test Plan
1. Doing verifier tests

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

Optimizations
* Improve toString method for a few Id classes using StringBuilder :pr:`24341`
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/execution/StageId.java
+++ b/presto-main/src/main/java/com/facebook/presto/execution/StageId.java
@@ -79,7 +79,15 @@ public class StageId
     @JsonValue
     public String toString()
     {
-        return queryId + "." + id;
+        StringBuilder builder = new StringBuilder();
+        appendString(builder);
+        return builder.toString();
+    }
+
+    public void appendString(StringBuilder builder)
+    {
+        queryId.appendString(builder);
+        builder.append(".").append(id);
     }
 
     @Override
--- a/presto-main/src/test/java/com/facebook/presto/execution/TestId.java
+++ b/presto-main/src/test/java/com/facebook/presto/execution/TestId.java
@@ -1,0 +1,36 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.execution;
+
+import com.facebook.presto.spi.QueryId;
+import org.testng.annotations.Test;
+
+import static org.testng.Assert.assertEquals;
+
+public class TestId
+{
+    @Test
+    public void testAppendString()
+    {
+        QueryId queryId = new QueryId("foo");
+        StageId stageId = new StageId(queryId, 1);
+        StageExecutionId stageExecutionId = new StageExecutionId(stageId, 2);
+        TaskId taskId = new TaskId(stageExecutionId, 3, 1);
+
+        assertEquals(queryId.toString(), "foo");
+        assertEquals(stageId.toString(), String.format("%s.%s", queryId.getId(), stageId.getId()));
+        assertEquals(stageExecutionId.toString(), String.format("%s.%s.%s", queryId.getId(), stageId.getId(), stageExecutionId.getId()));
+        assertEquals(taskId.toString(), String.format("%s.%s.%s.%s.%s", queryId.getId(), stageId.getId(), stageExecutionId.getId(), taskId.getId(), taskId.getAttemptNumber()));
+    }
+}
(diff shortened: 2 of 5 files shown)


Sullivan-Patrick:
I'm not sure, but I'd expect the compiler to implicitly use Stringbuilder for these cases. This seems to confirm that, at least after J9: 

https://stackoverflow.com/questions/1532461/stringbuilder-vs-string-concatenation-in-tostring-in-java

Did you do a perf test after your changes to see if it makes a difference?

kaikalur:
> I'm not sure, but I'd expect the compiler to implicitly use Stringbuilder for these cases. This seems to confirm that, at least after J9:
> 
> https://stackoverflow.com/questions/1532461/stringbuilder-vs-string-concatenation-in-tostring-in-java
> 
> Did you do a perf test after your changes to see if it makes a difference?

You can decompile the classfile and see! I'd be surprised if it's at JIT level

hantangwangd:
It seems that in `.class` flies these codes have all been turned into `StringBuilder.append()` by java compiler. Have you observed any performance differences before and after this change?

shangm2:
@Sullivan-Patrick @kaikalur @hantangwangd  you guys are absolutely correct. It is using the builder under the hood and there is no perf diff before and after the change. Sometimes, it can be worse depending on the perf time. Thanks folks.

tdcmeehan:
Perhaps the kernel of truth here is we are doing too many useless allocations to serialize TaskIds.  We serialize them by first converting them to String, then serializing the String.  Perhaps we could just directly serialize the TaskId and dependent structures like `StageExecutionId` and `StageId` with custom serializers.

For example, we have a custom serializer for Block:

```java
jsonBinder(binder).addSerializerBinding(Block.class).to(BlockJsonSerde.Serializer.class);
```

We could try adding three of them for the structures above and see if this reduces useless allocations.  Perhaps under heavy load there is a marginal performance improvement.

Long term, we should probably investigate using binary serialization for better scalability and efficiency.

presto-main/src/main/java/com/facebook/presto/execution/StageId.java
```diff
@@ -79,7 +79,13 @@ public int getId()
     @JsonValue
     public String toString()
     {
-        return queryId + "." + id;
+        StringBuilder builder = new StringBuilder();
+        return withStringBuilder(builder).toString();
+    }
+
+    public StringBuilder withStringBuilder(StringBuilder builder)
```
NikhilCollooru, line 86: how about renaming this to --> `toString(StringBuilder builder)`

shangm2: The thing is it is NOT returning a string. Maybe toStringBuilder?

sdruzkin: IMHO there is no point in returning a StringBuilder or applying a builder like method naming. 

I agree with Nikhil that an overloaded toString(StringBuilder) fits better.

tdcmeehan: +1 that this should not return `StringBuilder`.  Just make them append so there's only one option when you want to serialize them:

```suggestion
    public void toString(StringBuilder builder)
    {
        builder.append(queryId.getId()).append(".").append(id);
    }
```

The no-arg `toString` becomes:

```java
    @JsonValue
    public String toString()
    {
        StringBuilder builder = new StringBuilder();
        appendString(builder);
        return builder.toString();
    }

```

If `toString` sounds bad because it's not returning a string, what about `appendString`?

yhwang: +1 for `appendString` and passing `StringBuilder` for it makes sense.

shangm2: Thanks for all your precious comments. Really appreciate it. Feel free to take another look.

NikhilCollooru (dismissed):

NikhilCollooru:
Are there any unit tests that test this code ? If not can we add one ?

shangm2:
> Are there any unit tests that test this code ? If not can we add one ?

Good point. Checking.

shangm2:
@NikhilCollooru just added unit tests.

presto-main/src/test/java/com/facebook/presto/execution/TestId.java
```diff
@@ -0,0 +1,44 @@
...
+        QueryId queryId = new QueryId("foo");
+        StageId stageId = new StageId(queryId, 1);
+        StageExecutionId stageExecutionId = new StageExecutionId(stageId, 2);
+        TaskId taskId = new TaskId(stageExecutionId, 3, 1);
+
+        StringBuilder builder = new StringBuilder();
+        builder.append("foo");
+        assertEquals(builder.toString(), queryId.toString());
+
+        builder.append(".").append(stageId.getId());
```
sdruzkin, line 35: Nit: using String.format for the expected value would be more readable.

shangm2: Good point.

sdruzkin (approved):

NikhilCollooru (approved):
LGTM

Issue #20958, opened 2023-09-25
Clickhouse connector incorrectly transforms Timestamp constants to DateTime literals


Dnnd:
Presto-clickhouse connector transforms a `TIMESTAMP WITH TIME ZONE` constant like `from_iso8601_timestamp('2023-09-20T00:00:00+09:00')` to `TIMESTAMP '2023-09-19 15:00:00.000'` literal in native Clickhouse SQL. `TIMESTAMP` in Clickhouse is an alias to the `DateTime` type, which doesn't offer sub-second precision. This leads to a parsing error in Clickhouse.

## Your Environment

* Presto version used: 0.283
* Data source and connector used: presto-clickhouse
* Deployment (Cloud or On-prem): on-perm

A Clickhouse table with the following schema (native Clickhouse DDL syntax):
```sql
CREATE TABLE default.t
(
    `ts` DateTime,
    `value` Float32
)
ENGINE = MergeTree
ORDER BY ts
```

Presto-clickhouse  connector is configured like this:
```
connector.name=clickhouse
clickhouse.connection-url=jdbc:clickhouse://localhost:8123
clickhouse.connection-user=default
```

## Expected behavior
I expect the following query to complete without errors (Presto SQL):

```sql
select count(*) 
FROM clickhouse.default.t 
where ts >= from_iso8601_timestamp('2023-09-20T00:00:00+09:00');
```

## Current Behavior
Query returns an error:
```
Query 20230925_102132_00013_842zf failed: ClickHouse exception, code: 1002, host: localhost, port: 8123; Code: 6. DB::Exception: Cannot parse string '2023-09-19 15:00:00.000' as DateTime: syntax error at position 19 (parsed just '2023-09-19 15:00:00'): While processing SELECT count() FROM default.t WHERE ts >= toDateTime('2023-09-19 15:00:00.000'). (CANNOT_PARSE_TEXT) (version 23.6.2.18 (official build))
```

<details>
<summary> Stacktrace </summary>

```
[...]

PR #24344, opened 2025-01-09
[Clickhouse] add DateTime64 type support

Apidcloud:
## Description
This adds support to the missing type DateTime64 from ClickHouse.

## Motivation and Context
I noticed DateTime64 is not supported. As mentioned in https://github.com/prestodb/presto/issues/20958, DateTime64 supports milliseconds, whereas DateTime doesn't.

## Impact
We are currently blocked from using prestodb because every entry we have in our tables has a column createdAt with the type DateTime64. Milliseconds make sense when we are dealing with user/system events.

## Test Plan
Tests were added showing the creation of a table with a DateTime64 column (mapped from timestamp in prestodb), insertion of timestamps into a DateTime64 column with 0 or multiple millisecond digits, and finally the selection of them.

## Release Notes

```
Clickhouse Connector Changes
* Add ``DateTime64`` type support. :pr:`24344`
```

### DIFF
--- a/presto-clickhouse/src/main/java/com/facebook/presto/plugin/clickhouse/StandardReadMappings.java
+++ b/presto-clickhouse/src/main/java/com/facebook/presto/plugin/clickhouse/StandardReadMappings.java
@@ -48,6 +48,7 @@ import static com.facebook.presto.plugin.clickhouse.ClickHouseErrorCode.JDBC_ERR
 import static com.facebook.presto.plugin.clickhouse.DateTimeUtil.getMillisOfDay;
 import static com.facebook.presto.plugin.clickhouse.ReadMapping.longReadMapping;
 import static com.facebook.presto.plugin.clickhouse.ReadMapping.sliceReadMapping;
+import static com.facebook.presto.plugin.clickhouse.TimestampUtil.getMillisecondsFromTimestampString;
 import static io.airlift.slice.Slices.utf8Slice;
 import static io.airlift.slice.Slices.wrappedBuffer;
 import static java.lang.Float.floatToRawIntBits;
@@ -140,7 +141,8 @@ public final class StandardReadMappings
     {
         return longReadMapping(TIMESTAMP, (resultSet, columnIndex) -> {
             Timestamp timestamp = resultSet.getTimestamp(columnIndex);
-            return timestamp.getTime();
+            // getTimestamp loses the milliseconds, but we can get them from the getString
+            return timestamp.getTime() + getMillisecondsFromTimestampString(resultSet.getString(columnIndex));
         });
     }
 
@@ -163,6 +165,8 @@ public final class StandardReadMappings
                     return Optional.of(varcharReadMapping(createUnboundedVarcharType()));
                 }
                 return Optional.of(varbinaryReadMapping());
+            case "DateTime64": // DateTime64(n)
+                return Optional.of(timestampReadMapping());
             case "block":
                 return Optional.of(doubleReadMapping());
         }
--- a/presto-clickhouse/src/main/java/com/facebook/presto/plugin/clickhouse/TimestampUtil.java
+++ b/presto-clickhouse/src/main/java/com/facebook/presto/plugin/clickhouse/TimestampUtil.java
@@ -1,0 +1,35 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.plugin.clickhouse;
+
+public class TimestampUtil
+{
+    private TimestampUtil() {}
+
+    public static int getMillisecondsFromTimestampString(String timestampString)
+    {
+        int dotIndex = timestampString.indexOf('.');
+        if (dotIndex == -1) {
+            return 0;
+        }
+
+        String fraction = timestampString.substring(dotIndex + 1);
+        int nonNormalized = Integer.parseInt(fraction);
+        if (nonNormalized == 0 || fraction.length() == 3) {
+            return nonNormalized;
+        }
+        // this will make sure it's always 3 digits. e.g., 7 -> 700; 71 -> 710; 7591 -> 759
+        return (int) Math.round(nonNormalized * Math.pow(10, -(Math.floor(Math.log10(nonNormalized)) - 2)));
+    }
+}
--- a/presto-clickhouse/src/test/java/com/facebook/presto/plugin/clickhouse/TestingClickHouseServer.java
+++ b/presto-clickhouse/src/test/java/com/facebook/presto/plugin/clickhouse/TestingClickHouseServer.java
@@ -26,7 +26,7 @@ import static org.testcontainers.containers.ClickHouseContainer.HTTP_PORT;
 public class TestingClickHouseServer
         implements Closeable
 {
-    private static final String CLICKHOUSE_IMAGE = "yandex/clickhouse-server:20.8";
+    private static final String CLICKHOUSE_IMAGE = "clickhouse/clickhouse-server:23.12.2.59";
     private final ClickHouseContainer dockerContainer;
 
     public TestingClickHouseServer()
--- a/presto-docs/src/main/sphinx/connector/clickhouse.rst
+++ b/presto-docs/src/main/sphinx/connector/clickhouse.rst
@@ -89,6 +89,26 @@ Run ``SELECT`` to access the ``cks`` table in the ``tpch`` database::
     If you used a different name for your catalog properties file, use
     that catalog name instead of ``clickhouse`` in the above examples.
 
+PrestoDB to ClickHouse Type Mapping
+-----------------------------------
+
+========================================== ========================= =================================================================================
+**PrestoDB Type**                          **ClickHouse Type**       **Notes**
+========================================== ========================= =================================================================================
+BOOLEAN                                    UInt8                     ClickHouse uses UInt8 as boolean, restricted values to 0 and 1.
+TINYINT                                    Int8
+SMALLINT                                   Int16
+INTEGER                                    Int32
+BIGINT                                     Int64
+REAL                                       Float32
+DOUBLE                                     Float64
+DECIMAL                                    Decimal(precision, scale) The precision and scale are dynamic based on the PrestoDB type.
+CHAR / VARCHAR                             String                    The String type replaces VARCHAR, BLOB, CLOB, and related types from other DBMSs.
+VARBINARY                                  String
+DATE                                       Date
+TIMESTAMP                                  DateTime64(3)             Timestamp with 3 digits of millisecond precision.
+========================================== ========================= =================================================================================
+
 Table properties
 ----------------
 
--- a/presto-clickhouse/src/main/java/com/facebook/presto/plugin/clickhouse/ClickHouseClient.java
+++ b/presto-clickhouse/src/main/java/com/facebook/presto/plugin/clickhouse/ClickHouseClient.java
@@ -17,6 +17,7 @@ import com.facebook.airlift.log.Logger;
 import com.facebook.presto.common.predicate.TupleDomain;
 import com.facebook.presto.common.type.CharType;
 import com.facebook.presto.common.type.DecimalType;
+import com.facebook.presto.common.type.TimestampType;
 import com.facebook.presto.common.type.Type;
 import com.facebook.presto.common.type.VarbinaryType;
 import com.facebook.presto.common.type.VarcharType;
@@ -874,6 +875,9 @@ public class ClickHouseClient
         }
         if (type == DATE) {
             return "Date";
+        }
+        if (type instanceof TimestampType) {
+            return "DateTime64(3)";
         }
         throw new PrestoException(NOT_SUPPORTED, "Unsupported column type: " + type);
     }
--- a/presto-clickhouse/src/main/java/com/facebook/presto/plugin/clickhouse/ClickHousePageSink.java
+++ b/presto-clickhouse/src/main/java/com/facebook/presto/plugin/clickhouse/ClickHousePageSink.java
@@ -17,6 +17,7 @@ import com.facebook.airlift.log.Logger;
 import com.facebook.presto.common.Page;
 import com.facebook.presto.common.block.Block;
 import com.facebook.presto.common.type.DecimalType;
+import com.facebook.presto.common.type.TimestampType;
 import com.facebook.presto.common.type.Type;
 import com.facebook.presto.spi.ConnectorPageSink;
 import com.facebook.presto.spi.ConnectorSession;
@@ -162,6 +163,10 @@ public class ClickHousePageSink
         else if (DATE.equals(type)) {
             // convert to midnight in default time zone
             statement.setDate(parameter, convertZonedDaysToDate(type.getLong(block, position)));
+        }
+        else if (type instanceof TimestampType) {
+            // setTimestamp doesn't work, so we use setLong as described at https://github.com/ClickHouse/clickhouse-java/issues/608
+            statement.setLong(parameter, type.getLong(block, position));
         }
         else {
             throw new PrestoException(NOT_SUPPORTED, "Unsupported column type: " + type.getDisplayName());
--- a/presto-clickhouse/src/test/java/com/facebook/presto/plugin/clickhouse/TestClickHouseDistributedQueries.java
+++ b/presto-clickhouse/src/test/java/com/facebook/presto/plugin/clickhouse/TestClickHouseDistributedQueries.java
@@ -25,18 +25,23 @@ import org.testng.annotations.AfterClass;
 import org.testng.annotations.Test;
 
 import java.security.SecureRandom;
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
 
 import static com.facebook.presto.common.type.BigintType.BIGINT;
 import static com.facebook.presto.common.type.BooleanType.BOOLEAN;
 import static com.facebook.presto.common.type.VarcharType.VARCHAR;
 import static com.facebook.presto.plugin.clickhouse.ClickHouseQueryRunner.createClickHouseQueryRunner;
 import static com.facebook.presto.testing.MaterializedResult.resultBuilder;
+import static com.facebook.presto.testing.TestingSession.DEFAULT_TIME_ZONE_KEY;
 import static com.facebook.presto.testing.assertions.Assert.assertEquals;
 import static com.facebook.presto.tests.QueryAssertions.assertEqualsIgnoreOrder;
 import static java.lang.Character.MAX_RADIX;
 import static java.lang.Math.abs;
 import static java.lang.Math.min;
 import static java.lang.String.format;
+import static java.time.format.DateTimeFormatter.ISO_ZONED_DATE_TIME;
 import static java.util.Objects.requireNonNull;
 import static java.util.stream.Collectors.joining;
 import static java.util.stream.IntStream.range;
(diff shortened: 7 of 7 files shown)


ZacBlanco (changes requested):
You'll need to add some tests to verify the mapping works correctly.

Based on your PR description, it sounds like you may have been trying to create tables in clickhouse through presto? If so, you won't be able to use the same column type names that clickhouse uses. You'll need to use the mapped type. i.e. DateTime64 should map to timestamp. See `TestClickhouseDistributedQueries` for example tests.

Apidcloud:
I’ve based myself on that file and timestamp has the same issue. Invalid column type after I try to do an assertUpdate with a create table. Date works fine. Datetime, datetime64, or timestamp don’t.

On 9 Jan 2025, at 23:33, Zac Blanco ***@***.***> wrote:

﻿

@ZacBlanco requested changes on this pull request.

You'll need to add some tests to verify the mapping works correctly.

Based on your PR description, it sounds like you may have been trying to create tables in clickhouse through presto? If so, you won't be able to use the same column type names that clickhouse uses. You'll need to use the mapped type. i.e. DateTime64 should map to timestamp. See TestClickhouseDistributedQueries for example tests.

steveburnett:
Thanks for the release note! Minor rephrasing suggestion. 

```
== RELEASE NOTES ==

Clickhouse Connector Changes
* Add ``DateTime64`` type support. :pr:`24344`
```

Apidcloud:
Somehow in the tests, this timestamp string (`insert into t1 VALUES (timestamp '2025-01-08 12:35:59.079')`) becomes `1736292959079` (Tuesday, 7 January 2025 23:35:59.079), but when I do it on the UI at :8080 the same timestamp string becomes `1736336159079` (Wednesday, 8 January 2025 11:35:59.079). How come?

Apidcloud:
Ah, it seems the timezone in tests is totally off. How can i change that? I'm based in Vienna, whereas tests are running in Bahia Mexico lol
ui-timezone
test-timezone

Apidcloud:
I'm trying to convert the hour I want i UTC to the test timezone and use that in the insert statement. I assumed it would transform it to the UTC counterpart, but it went back almost 24h. What am I missing?

```
String tableT = "t1";

        // Get the current time zone
        String testTimeZone = TimeZone.getDefault().getID();

        ZonedDateTime originalTimestamp = ZonedDateTime.parse("2025-01-08T12:34:56.179Z", DateTimeFormatter.ISO_ZONED_DATE_TIME);
        ZonedDateTime adjustedTimestamp = originalTimestamp.withZoneSameInstant(ZoneId.of(testTimeZone));

        // Correct: 2025-01-08 06:34:56.179"
        String adjustedTimestampString = adjustedTimestamp.format(DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.SSS"));

        assertUpdate("CREATE TABLE " + tableT + " (ts timestamp not null)");
        assertUpdate("INSERT INTO " + tableT + " (ts) VALUES (timestamp '" + adjustedTimestampString + "')", 1);
        // wrong. 2025-01-07 17:34:56.179 
        // after insert it's wrong. in docker, the timezone is UTC within clickhouse and the entry shows 2025-01-07 17:34:56.179
```

Apidcloud:
I'm getting a bit confused... after further checking, the insert is somehow getting Pacific/Apia Samoa (UTC+13). It's using some legacy timestamp parsing, too. If I change the code from before to indeed use Pacific/Apia, the timestamp stored on clickhouse is finally correct. The question is now: why is it set to such timezone? or, more importantly, how can I go around it?
Screenshot 2025-01-10 at 22 25 47
Screenshot 2025-01-10 at 22 25 10

Apidcloud:
Finally seems to be working. Can you tell me the process of creating the tar.gz prestodb file available for download, so that we can start using this version tomorrow?

ZacBlanco (changes requested):
This is a great improvement to the connector. Mostly just style nits. Mainly, I am concerned about the performance. I think it would be great to implement this without any string parsing. Have you attempted any other solutions?

Also, our clickhouse-jdbc client dependency is quite out of date. We might be able to upgrade it to the latest `com.clickhouse:clickhouse-jdbc:0.7.2` which may provide more options for timestamps, though I haven't looked further into it.

presto-clickhouse/src/test/java/com/facebook/presto/plugin/clickhouse/TestingClickHouseServer.java
```diff
@@ -26,7 +26,7 @@
 public class TestingClickHouseServer
         implements Closeable
 {
-    private static final String CLICKHOUSE_IMAGE = "yandex/clickhouse-server:20.8";
```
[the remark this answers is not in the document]
Apidcloud: This was actually out of necessity, as older versions of clickhouse (also mentioned in that clickhouse issue) had some issues and the setLong wouldn't work. I would document it but couldn't find the right place to do so.

presto-clickhouse/src/main/java/com/facebook/presto/plugin/clickhouse/StandardReadMappings.java
```diff
@@ -139,6 +145,29 @@ public static ReadMapping timeReadMapping()
     public static ReadMapping timestampReadMapping()
     {
         return longReadMapping(TIMESTAMP, (resultSet, columnIndex) -> {
+            String timestampString = resultSet.getString(columnIndex);
+            Pattern fractionalPattern = Pattern.compile("\\.(\\d+)$");
+            Matcher matcher = fractionalPattern.matcher(timestampString);
```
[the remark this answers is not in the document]
Apidcloud: As far as I can tell it always comes as a string. I tried getLong again and got this exception:
Screenshot 2025-01-13 at 09 40 29

This is something that might change in a newer version of the clickhouse jdbc.

presto-clickhouse/src/main/java/com/facebook/presto/plugin/clickhouse/StandardReadMappings.java
```diff
@@ -163,6 +192,8 @@ public static Optional<ReadMapping> jdbcTypeToPrestoType(ClickHouseTypeHandle ty
                     return Optional.of(varcharReadMapping(createUnboundedVarcharType()));
                 }
                 return Optional.of(varbinaryReadMapping());
+            case "DateTime64": // DateTime64(n)
```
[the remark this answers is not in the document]
Apidcloud: DateTime64 can receive 'precision' on clickhouse side (similar to what the FixedString(n) means), but defaults to 3 which is what we are using

Apidcloud:
Thanks for the prompt review. Already pushed the changes.

I do agree the clickhouse jdbc is really outdated, and I tried clickhouse jdbc 0.7.2 at some point. But I can remember the tpch tables in TestClickHouseDistributedQueries would fail to be created because 'transactions are not supported'. Either way, I would try to update clickhouse jdbc in a subsequent MR, as it would most likely have other breaking changes.

presto-clickhouse/src/main/java/com/facebook/presto/plugin/clickhouse/StandardReadMappings.java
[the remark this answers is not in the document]
Apidcloud: But since DateTime64 will default to a precision of 3 I don't think we need the StringBuilder anymore. And if that's the case maybe we can drop the regex pattern as well?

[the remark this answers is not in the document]
Apidcloud: Just pushed a version of the combined approach (getTimestamp + getString)

presto-clickhouse/src/test/java/com/facebook/presto/plugin/clickhouse/TestingClickHouseServer.java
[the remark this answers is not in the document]
Apidcloud: Ah, i found some .rst files inside sphinx. Also added a table with prestodb type to clickhouse mapping

presto-docs/src/main/sphinx/connector/clickhouse.rst
```diff
@@ -89,6 +89,32 @@ Run ``SELECT`` to access the ``cks`` table in the ``tpch`` database::
     If you used a different name for your catalog properties file, use
     that catalog name instead of ``clickhouse`` in the above examples.
 
+PrestoDB to ClickHouse Type Mapping
+====================================
```
steveburnett, line 93:
```suggestion
-----------------------------------
```
When I built the doc locally, this heading `PrestoDB to ClickHouse Type Mapping` was missing completely. I replaced `====` with `----` and the heading displays in a local build. See screenshots. 

Before
Screenshot 2025-01-13 at 9 44 42 AM

After
Screenshot 2025-01-13 at 9 45 18 AM

Also, please make sure the formatting line (93) has exactly the same number of characters as in the line (92) being formatted.

steveburnett (changes requested):
Thanks for the doc! Nit of formatting in the header, the table looks great!

presto-docs/src/main/sphinx/connector/clickhouse.rst
Apidcloud: Ah, good catch. I used an only viewer and completely forgot the header should also show. Pushed the change

Apidcloud:
Can you retrigger the CI? It seems one of the test suites timed out

presto-clickhouse/src/main/java/com/facebook/presto/plugin/clickhouse/StandardReadMappings.java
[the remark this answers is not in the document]
Apidcloud: and actually the calendar (and thus the server time zone and reflection) is not needed. Getting the milliseconds from the string is enough

presto-clickhouse/src/main/java/com/facebook/presto/plugin/clickhouse/TimestampUtil.java
```diff
@@ -0,0 +1,39 @@
...
+        if (dotIndex != -1) {
+            String fractionalPart = timestampString.substring(dotIndex + 1);
+            int fractionalLength = fractionalPart.length();
+            // Normalize the fractional part to at least 3 digits by padding with zeros. e.g., 7 -> 700; 70 -> 700
+            if (fractionalLength == 1) {
+                fractionalPart += "00";
+            }
+            else if (fractionalLength == 2) {
+                fractionalPart += "0";
+            }
```
ZacBlanco, line 34: Do you think we should consider cases of 4, 5, 6,  etc decimal places? Also, rather than parsing a larger string, have you considered just parsing the value directly then normalizing at the end? Conditionals  are expensive and should ideally be avoided in the datatpath. It may be cheaper than the current solution, but you might need to write a microbenchmark.

Here's an example function that might get what you're going for:

https://www.desmos.com/calculator/2hheawsmd4

In case the link ever goes bad, this is the function:

$f\left(x\right)=\text{round}\left(x\cdot10^{\left(-\left(\text{floor}\left(\log10\left(x\right)\right)\ -\ 3\right)\right)}\right)$

ZacBlanco (changes requested):

presto-clickhouse/src/main/java/com/facebook/presto/plugin/clickhouse/TimestampUtil.java
Apidcloud: No because more than 3 get interpreted to seconds (as they should).
If clickhouse is storing 2025-01-08 12:34:56.7591 then the return when you do a select will be 2025-01-08 12:35:03.591 (note the 7 more seconds).

I will take a look into the rest of what you mentioned in a bit.

ZacBlanco: Wouldn't additional values after the decimal point refer to higher precision, not necessarily more seconds? I guess if we hard code 3-decimal place precision in the clickhouse type we wouldn't expect to see more decimal places though.

Apidcloud: The 1st decimal place in 4 digits of precision gets translated to seconds when it gets to Presto. We hardcode the 3 decimal places when we insert from Presto. But if people do that outside of Presto, then it might be possible to have more. It's unlikely, but the common scenarios are: 3 digits for milliseconds, 6 for microseconds, and 9 for nanoseconds. The precision values can range from 0 to 9. But in case people hardcode 1 or 2, then we need the padding on presto side, otherwise .7 gets interpreted as 007 instead of 700, and .71 as 071 instead of 710

Apidcloud: When you say 'Also, rather than parsing a larger string', are you suggesting parsing the string decimal places directly and adding the number afterwards? Like, if it is 1 digit, then we multiply it by 100, 2 digits by 10.

ZacBlanco: Yes because in your original code there are conditionals and allocations occurring just to create the string which is then parsed by `Integer.parseInt`. You could change the code to use mathematical operations instead to avoid allocations and conditionals.

```java
int nonNormalized = Integer.parseInt(fraction)
int normalized = (int) Math.round(fraction * Math.pow(10, -(Math.floor(Math.log10(fraction) - 3)))
return normalized;
```

steveburnett (dismissed):
LGTM! (docs)

presto-clickhouse/src/main/java/com/facebook/presto/plugin/clickhouse/TimestampUtil.java
Apidcloud: I must be missing something because desmos is off by a factor of 10 when I do the same manually. For instance, if the fraction is 75, then log10(75) is 1.875. Floored is 1. 10 to the power of 2 is 100, and multiplying that by 75 is 7500 instead of the 750 shown in desmos.

Apidcloud: I think what I mentioned before regarding the 4 digits is incorrect as you pointed out (ref https://clickhouse.com/docs/en/sql-reference/data-types/datetime64). It wouldn't really make sense for the digit to suddenly mean seconds. Precision refers to seconds, but it goes lower and lower

presto-docs/src/main/sphinx/connector/clickhouse.rst
```diff
@@ -89,6 +89,32 @@ Run ``SELECT`` to access the ``cks`` table in the ``tpch`` database::
...
+SMALLINT                                   Int16
+INTEGER                                    Int32
+BIGINT                                     Int64
+REAL                                       Float32
+DOUBLE                                     Float64
+DECIMAL                                    Decimal(precision, scale) The precision and scale are dynamic based on the PrestoDB type.
+CHAR / VARCHAR                             String                    The String type replaces VARCHAR, BLOB, CLOB, etc., from other DBMSs.
+VARBINARY                                  String
+DATE                                       Date
+TIMESTAMP                                  DateTime64(3)             Timestamp with 3 digits of millisecond precision. This is tested against Clickhouse-Server 23.12.2.59
```
ZacBlanco, line 109: I would recommend leaving out last sentence with the clickhouse server version. It is unlikely this doc will be updated if we upgrade the test server in the future.

presto-docs/src/main/sphinx/connector/clickhouse.rst
```diff
@@ -89,6 +89,32 @@ Run ``SELECT`` to access the ``cks`` table in the ``tpch`` database::
...
+========================================== ========================= =====================================================================================================
+
+
+Unsupported types
+-----------------
+If a column type is not supported, an exception is thrown with the message: `Unsupported column type`.
+
```
ZacBlanco, line 116: I don't think this is necessary, unless you want to specify a list of actual types

ZacBlanco (changes requested):
Code LGTM. two doc comments.

Apidcloud:
Rebased and squashed commits

presto-docs/src/main/sphinx/connector/clickhouse.rst
```diff
@@ -89,6 +89,26 @@ Run ``SELECT`` to access the ``cks`` table in the ``tpch`` database::
...
+========================================== ========================= =====================================================================
+BOOLEAN                                    UInt8                     ClickHouse uses UInt8 as boolean, restricted values to 0 and 1.
+TINYINT                                    Int8
+SMALLINT                                   Int16
+INTEGER                                    Int32
+BIGINT                                     Int64
+REAL                                       Float32
+DOUBLE                                     Float64
+DECIMAL                                    Decimal(precision, scale) The precision and scale are dynamic based on the PrestoDB type.
+CHAR / VARCHAR                             String                    The String type replaces VARCHAR, BLOB, CLOB, etc., from other DBMSs.
```
steveburnett, line 106:
```suggestion
CHAR / VARCHAR                             String                    The String type replaces VARCHAR, BLOB, CLOB, and related types from other DBMSs.
```
Avoid using Latin abbreviations. See the GitLab Developer Documentation Style Guide [Recommended Word List entry for etc.](https://docs.gitlab.com/ee/development/documentation/styleguide/word_list.html#etc) for discussion and suggestions. I have suggested one possibility but if this doesn't work for you please think of another.

steveburnett (changes requested):
One nit of phrasing. Thanks!

steveburnett (approved):
LGTM! (docs) 

ZacBlanco (approved):
Thanks for this great work!

presto-clickhouse/src/main/java/com/facebook/presto/plugin/clickhouse/StandardReadMappings.java
```diff
@@ -140,7 +141,8 @@ public static ReadMapping timestampReadMapping()
     {
         return longReadMapping(TIMESTAMP, (resultSet, columnIndex) -> {
             Timestamp timestamp = resultSet.getTimestamp(columnIndex);
-            return timestamp.getTime();
+            // getTimestamp loses the milliseconds, but we can get them from the getString
```
jaystarshot, line 144: What do you mean here? 


 /**
     * Returns the number of milliseconds since January 1, 1970, 00:00:00 GMT
     * represented by this {@code Timestamp} object.
     *
     * @return  the number of milliseconds since January 1, 1970, 00:00:00 GMT
     *          represented by this date.
     * @see #setTime
     */
    public long getTime() {
        long time = super.getTime();
        return (time + (nanos / 1000000));
    }

jaystarshot: do you mean for precision? can we use nanos then?

```
/**
     * Gets this {@code Timestamp} object's {@code nanos} value.
     *
     * @return this {@code Timestamp} object's fractional seconds component
     * @see #setNanos
     */
    public int getNanos() {
        return nanos;
    }
```

Apidcloud: That if there's a timestamp string like 2025-01-08 06:34:56.179, the .179 is lost (3 digits have a precision of milliseconds)

jaystarshot: even if we use nanos?

Apidcloud:
Yeah, Nanos from the timestamp is always 0

On 15 Jan 2025, at 22:03, Jay Narale ***@***.***> wrote:

﻿

@jaystarshot commented on this pull request.

In presto-clickhouse/src/main/java/com/facebook/presto/plugin/clickhouse/StandardReadMappings.java<https://github.com/prestodb/presto/pull/24344#discussion_r1917331371>:

> @@ -140,7 +141,8 @@ public static ReadMapping timestampReadMapping()
     {
         return longReadMapping(TIMESTAMP, (resultSet, columnIndex) -> {
             Timestamp timestamp = resultSet.getTimestamp(columnIndex);
-            return timestamp.getTime();
+            // getTimestamp loses the milliseconds, but we can get them from the getString


even if we use nanos?

presto-clickhouse/src/main/java/com/facebook/presto/plugin/clickhouse/StandardReadMappings.java
Apidcloud: Yeah, Nanos from the timestamp is always 0

jaystarshot (approved):

tdcmeehan (approved):

PR #24351, opened 2025-01-13
Fix ORC single rowgroup read to use StripeMetadataSource instead of M…

nmahadevuni:
## Description
This is a fix to use StripeMetadataSource instead of MetadataReader, even when there is only one row group in a stripe.

## Motivation and Context
Currently, when there are more than 1 rowgroups in a stripe, StripeMetadataSource is being used and when there is only 1 rowgroup, MetadataReader is used. This bypasses caching reader if enabled, losing the benefits of caching. Fixed to use StripeMetadataSource in both cases.

## Impact
No impact

## Test Plan
Current tests in AbstractTestOrcReader.java covers this scenario.

## Release Notes

```
== NO RELEASE NOTE ==
```


sdruzkin:
LGTM. Please fix up the description - there is a partially cut word before the description.

sdruzkin (approved):

sdruzkin:
I fixed up the description and restarted the failed test to get a clean signal.

nmahadevuni:
Thanks @sdruzkin

PR #24355, opened 2025-01-13
[native] Fix shuffle compression kind string in toVeloxConfigs()

spershin:
The current code sets integer to shuffle compression kind, while code expects a compression kind string.

```
== NO RELEASE NOTE ==
```


amitkdutta (approved):
Thanks @spershin

PR #24370, opened 2025-01-15
Add symlink path cache to storage partition loader

auden-woolfson:
## Description
Users have reported performance issues in certain scenarios that require frequent access to symlink files from the storage partition loader. This is likely due to a high volume of IO operations being performed on these files, frequent opening and closing. We can cache the file contents in order to prevent this from happening. 

```
== RELEASE NOTES ==

General Changes
* ... :pr:`12345`
* ... :pr:`12345`

Hive Connector Changes
* ... :pr:`12345`
* ... :pr:`12345`
```

```
== NO RELEASE NOTE ==
```


agrawalreetika:
@auden-woolfson I think this is not needed. Can we close this?

PR #24380, opened 2025-01-16
[native] Fix crash in PrestoTask::updateExecutionInfoLocked

arhimondr:
```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/PrestoTask.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoTask.cpp
@@ -795,43 +795,46 @@ void PrestoTask::updateExecutionInfoLocked(
   }
   for (int i = 0; i < veloxTaskStats.pipelineStats.size(); ++i) {
     auto& veloxPipeline = veloxTaskStats.pipelineStats[i];
-    if (veloxPipeline.inputPipeline) {
-      const auto& firstVeloxOpStats = veloxPipeline.operatorStats[0];
-      prestoTaskStats.rawInputPositions += firstVeloxOpStats.rawInputPositions;
-      prestoTaskStats.rawInputDataSizeInBytes +=
-          firstVeloxOpStats.rawInputBytes;
-      prestoTaskStats.processedInputPositions +=
-          firstVeloxOpStats.inputPositions;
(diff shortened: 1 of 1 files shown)


presto-native-execution/presto_cpp/main/PrestoTask.cpp
```diff
@@ -795,20 +795,26 @@ void PrestoTask::updateExecutionInfoLocked(
...
+        prestoTaskStats.processedInputPositions +=
+            firstVeloxOpStats.inputPositions;
+        prestoTaskStats.processedInputDataSizeInBytes +=
+            firstVeloxOpStats.inputBytes;
+      }
+      if (veloxPipeline.outputPipeline) {
+        const auto& lastVeloxOpStats = veloxPipeline.operatorStats.back();
+        prestoTaskStats.outputPositions += lastVeloxOpStats.outputPositions;
+        prestoTaskStats.outputDataSizeInBytes += lastVeloxOpStats.outputBytes;
+      }
```
spershin, line 817: Suggestion: we can include two following for loops under this condition as well up to line 833

spershin (approved):
Thanks for quickly fixing the issue!

arhimondr:
@spershin Thank you for a prompt review. Updated.

amitkdutta:
@arhimondr Thanks for the quick fix. Wondering why we didn't see the problem in OSS change. Do we need to add any e2e tests that we have in Meta internal system.

spershin (approved):

arhimondr:
@amitkdutta  From what I can see this can only happen when query:

1. uses grouped execution (for non grouped execution drivers are created eagerly: https://github.com/facebookincubator/velox/blob/main/velox/exec/Task.cpp#L683, for grouped execution lazily when splits are received: https://github.com/facebookincubator/velox/blob/main/velox/exec/Task.cpp#L1284, https://github.com/facebookincubator/velox/blob/main/velox/exec/Task.cpp#L1390)
3. Initial TaskUpdateRequest has no splits

I guess #1 is why I didn't catch it when shadowing thousands of queries. I assume I just didn't get to to run a lot of grouped execution queries.

And the #2 is why tests are not deterministically failing.

PR #24396, opened 2025-01-17
[native] Fix varchar cast for json

natashasehgal:
## Description
Resolve error in running query on Prestissimo not Presto - "Scalar function presto.default.substr not registered with arguments: (JSON, BIGINT, BIGINT)".

## Motivation and Context
This is not due to missing function in Prestissimo, as the function signature does not exist in Presto. It occurs when attempting to cast JSON as varchar of capped length.
Note: Exception is still raised for try_cast() behavior. Alignment is out of scope for this PR

## Impact
N/A

## Test Plan
- [x] Added unit test

## Contributor checklist

- [x] CI passed.

## Release Notes
```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/types/PrestoToVeloxExpr.cpp
+++ b/presto-native-execution/presto_cpp/main/types/PrestoToVeloxExpr.cpp
@@ -181,8 +181,13 @@ std::optional<TypedExprPtr> convertCastToVarcharWithMaxLength(
   VELOX_DCHECK(end == returnType.data() + returnType.size() - 1);
 
   VELOX_DCHECK_EQ(args.size(), 1);
-  const auto arg = args[0];
 
+  auto arg = args[0];
+  // If the argument is of JSON type, convert it to VARCHAR before applying
+  // substr.
+  if (velox::isJsonType(arg->type())) {
+    arg = std::make_shared<CastTypedExpr>(velox::VARCHAR(), arg, false);
+  }
   return std::make_shared<CallTypedExpr>(
       arg->type(),
       std::vector<TypedExprPtr>{
@@ -256,8 +261,8 @@ std::optional<TypedExprPtr> tryConvertCast(
   }
 
   // When the return type is varchar with max length, truncate if only the
-  // argument type is varchar (or varchar with max length). Non-varchar argument
-  // types are not truncated.
+  // argument type is varchar, or varchar with max length or json. Non-varchar
+  // argument types are not truncated.
   if (returnType.find(kVarchar) == 0 &&
       args[0]->type()->kind() == TypeKind::VARCHAR &&
       returnType.size() > strlen(kVarchar)) {
--- a/presto-native-execution/presto_cpp/main/types/tests/RowExpressionTest.cpp
+++ b/presto-native-execution/presto_cpp/main/types/tests/RowExpressionTest.cpp
@@ -18,6 +18,7 @@
 #include "presto_cpp/presto_protocol/core/presto_protocol_core.h"
 #include "velox/core/Expressions.h"
 #include "velox/type/Type.h"
+#include "velox/functions/prestosql/types/JsonType.h"
 
 using namespace facebook::presto;
 using namespace facebook::velox;
@@ -30,6 +31,7 @@ class RowExpressionTest : public ::testing::Test {
   }
 
   void SetUp() override {
+    registerJsonType();
     pool_ = memory::MemoryManager::getInstance()->addLeafPool();
     converter_ =
         std::make_unique<VeloxExprConverter>(pool_.get(), &typeParser_);
@@ -625,6 +627,29 @@ TEST_F(RowExpressionTest, castToVarchar) {
     ASSERT_NE(returnExpr, nullptr);
     ASSERT_TRUE(returnExpr->nullOnFailure());
     ASSERT_EQ(returnExpr->type()->toString(), "VARCHAR");
+  }
+  // CAST(json AS varchar(3))
+  {
+    std::shared_ptr<protocol::CallExpression> p =
+        json::parse(makeCastToVarchar(false, "json", "varchar(3)"));
+    auto expr = converter_->toVeloxExpr(p);
+    auto returnExpr = std::dynamic_pointer_cast<const CallTypedExpr>(expr);
+
+    ASSERT_NE(returnExpr, nullptr);
+    ASSERT_EQ(returnExpr->name(), "presto.default.substr");
+
+    auto returnArg1 = std::dynamic_pointer_cast<const CastTypedExpr>(
+        returnExpr->inputs()[0]);
+    auto returnArg2 = std::dynamic_pointer_cast<const ConstantTypedExpr>(
+        returnExpr->inputs()[1]);
+    auto returnArg3 = std::dynamic_pointer_cast<const ConstantTypedExpr>(
+        returnExpr->inputs()[2]);
+
+    ASSERT_EQ(returnArg1->type()->toString(), "VARCHAR");
+    ASSERT_EQ(returnArg2->type()->toString(), "BIGINT");
+    ASSERT_EQ(returnArg2->value().toJson(returnArg2->type()), "1");
+    ASSERT_EQ(returnArg3->type()->toString(), "BIGINT");
+    ASSERT_EQ(returnArg3->value().toJson(returnArg3->type()), "3");
   }
 }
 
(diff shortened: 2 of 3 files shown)


amitkdutta:
@natashasehgal Thanks for the fix. Please publish the PR, its in draft mode.

presto-native-execution/presto_cpp/main/types/tests/RowExpressionTest.cpp
```diff
@@ -18,13 +18,17 @@
...
 using namespace facebook::velox;
 using namespace facebook::velox::core;
 
 class RowExpressionTest : public ::testing::Test {
  public:
+  RowExpressionTest(){
+    registerJsonType();
+  }
   static void SetUpTestCase() {
     memory::MemoryManager::testingSetInstance({});
```
amitkdutta, line 33: May be we can register in setup and remove the constructor.

amitkdutta (commented):
Looks good. Lets update the commit message also.

presto-native-execution/presto_cpp/main/types/tests/RowExpressionTest.cpp
natashasehgal: Updated this

amitkdutta (dismissed):
Looks good. Thanks @natashasehgal

kgpai (approved):

presto-native-execution/presto_cpp/main/types/PrestoToVeloxExpr.cpp
```diff
@@ -181,8 +181,15 @@ std::optional<TypedExprPtr> convertCastToVarcharWithMaxLength(
...
   VELOX_DCHECK_EQ(args.size(), 1);
-  const auto arg = args[0];
 
+  auto arg = args[0];
+  // If the argument is of JSON type, convert it to VARCHAR before applying substr.
+  if (velox::isJsonType(arg->type())) {
+    arg = std::make_shared<CallTypedExpr>(
+        velox::VARCHAR(),
+        std::vector<TypedExprPtr>{arg},
+        "presto.default.json_format");
```
gggrace14, line 191: There's something more that we need to check here. Should we wrap first with CallTypedExpr "presto.default.json_format" (`json_format(json_input)`) here or with CastTypedExpr (`cast(json_input as varchar)`)? I'm not sure if the two return the same for all inputs.

@kgpai Krishna, do you know if json_format() and cast(json as varchar) return the same for all cases for *JAVA* (not Prestissimo)?

If not, we can pick the cases of input, try on Presto Java, and see whether cast(json as varchar(255)) return the same as json_format() or cast(json as varchar).

presto-native-execution/presto_cpp/main/types/PrestoToVeloxExpr.cpp
gggrace14, line 1: Now I see why it can get into the path of convertCastToVarcharWithMaxLength() today. It is because this condition

```
  // When the return type is varchar with max length, truncate if only the
  // argument type is varchar (or varchar with max length). Non-varchar argument
  // types are not truncated.
  if (returnType.find(kVarchar) == 0 &&
      args[0]->type()->kind() == TypeKind::VARCHAR &&
      returnType.size() > strlen(kVarchar)) {
    return convertCastToVarcharWithMaxLength(returnType, args, nullOnFailure);
  }
```

 `args[0]->type()->kind() == TypeKind::VARCHAR` is also satisfied if isJsonType(args[0]->type()). Would you mind changing the above comment accordingly and making it explicit "truncate if only the argument type is varchar, or varchar with max length or json"?

gggrace14 (changes requested):

gggrace14 (changes requested):
Request double check

presto-native-execution/presto_cpp/main/types/PrestoToVeloxExpr.cpp
natashasehgal: From documentation, json_format() makes sense to me

https://prestodb.io/docs/current/functions/json.html

> json_format() and CAST(json AS VARCHAR) have completely different semantics.
> 
> json_format() serializes the input JSON value to JSON text conforming to [RFC 7159](https://datatracker.ietf.org/doc/html/rfc7159.html). The JSON value can be a JSON object, a JSON array, a JSON string, a JSON number, true, false or null:
> 
> 
> SELECT json_format(JSON '{"a": 1, "b": 2}'); -- '{"a":1,"b":2}'
> SELECT json_format(JSON '[1, 2, 3]'); -- '[1,2,3]'
> SELECT json_format(JSON '"abc"'); -- '"abc"'
> SELECT json_format(JSON '42'); -- '42'
> SELECT json_format(JSON 'true'); -- 'true'
> SELECT json_format(JSON 'null'); -- 'null'
> CAST(json AS VARCHAR) casts the JSON value to the corresponding SQL VARCHAR value. For JSON string, JSON number, true, false or null, the cast behavior is same as the corresponding SQL type. JSON object and JSON array cannot be cast to VARCHAR:
> 
> 
> SELECT CAST(JSON '{"a": 1, "b": 2}' AS VARCHAR); -- ERROR!
> SELECT CAST(JSON '[1, 2, 3]' AS VARCHAR); -- ERROR!
> SELECT CAST(JSON '"abc"' AS VARCHAR); -- 'abc'; Note the double quote is gone
> SELECT CAST(JSON '42' AS VARCHAR); -- '42'
> SELECT CAST(JSON 'true' AS VARCHAR); -- 'true'
> SELECT CAST(JSON 'null' AS VARCHAR); -- NULL

kgpai: @gggrace14 cast(json_input as varchar)) is not correct, that will quote the values in json_input. json_format is correct option here and also is inexpensive (it virtually no-op).

gggrace14: Then for CAST(json AS VARCHAR(255)), is the behavior more like CAST(json AS VARCHAR) or JSON_FORMAT(json)? Esp. for the exception case and the JSON string case that has non-printable ASCII character and unicode character. You can check on a prod Java cluster.

kgpai: Thats a good question , @natashasehgal Can you validate this ? whether behavior is cast(json as varchar) and then substr on it or is it equivalent to json_format(json) and then substr on it ?

gggrace14: From the unit tests added in https://github.com/facebookincubator/velox/pull/11284, we can find the special cases that we need to test.

natashasehgal: @kgpai I noticed the behavior of Prestissimo JSON -> VARCHAR(255) is consistent with Presto JSON->VARCHAR(255) if I use cast not json_parse. 

json_parse() is causing extra string quotations to remain

I have made an update to this part.

kgpai (commented):
Looks good. Can you add an e2e test ?

natashasehgal:
> Looks good. Can you add an e2e test ?

Yes. @kgpai could you please point me to a suitable location to add an e2e test?  I wasn't able to find an e.g for `presto-native-execution` changes

kgpai (approved):
Thanks, LGTM.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeGeneralQueries.java
```diff
@@ -510,6 +510,7 @@ public void testCast()
         // Round-trip tests of casts for Json.
         assertQuery("SELECT cast(cast(name as JSON) as VARCHAR), cast(cast(size as JSON) as INTEGER), cast(cast(size + 0.01 as JSON) as DOUBLE), cast(cast(size > 5 as JSON) as BOOLEAN) FROM part");
         assertQuery("SELECT cast(cast(array[suppkey, nationkey] as JSON) as ARRAY(INTEGER)), cast(cast(map(array[name, address, phone], array[1.1, 2.2, 3.3]) as JSON) as MAP(VARCHAR(40), DOUBLE)), cast(cast(map(array[name], array[phone]) as JSON) as MAP(VARCHAR(25), JSON)), cast(cast(array[array[suppkey], array[nationkey]] as JSON) as ARRAY(JSON)) from supplier");
+        assertQuery("SELECT cast(json_extract(x, '$.a') AS varchar(255)) AS extracted_value FROM (VALUES ('{\"a\": \"Some long string\"}')) AS t(x)");
```
gggrace14, line 513: Would be better to add one more test case of assertQueryFails(). But I don't want to block the merging of the PR.

gggrace14 (approved):
Thank you @natashasehgal for iterating on this change and studying the behavior thoroughly! The impl looks right to me.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeGeneralQueries.java
natashasehgal: Great idea. I merged this already, will follow up with an appropriate failure test case

Issue #24383, opened 2025-01-17
Enabling Presto Iceberg to leverage the powerful capabilities of object storages


hantangwangd:
Currently, object stores become more and more common comparing with HDFS, they have higher scalability, better performance, and support for cloud services. So, enabling Presto Iceberg to leverage the powerful ability of object stores should be a good feature.

As discussed with @tdcmeehan in PR #24221, there may be issues with transaction atomicity and consistency when using HadoopCatalog to directly manage metadata on object stores. Although many people are trying to solve this problem (see: https://lists.apache.org/thread/kh4n98w4z22sc8h2vot4q8n44vdtnltg), especially after the emergence of S3 conditional writes, the current situation is that there are risks involved.

However, referring to the Iceberg community's discussion on the capabilities and limitations of HadoopCatalog  (see: https://lists.apache.org/thread/oohcjfp1vpo005h2r0f6gfpsp6op0qps and https://lists.apache.org/thread/v7x65kxrrozwlvsgstobm7685541lf5w), we know that HadoopCatalog is capable of managing metadata on HDFS, thanks to HDFS's strictly support for atomic non-rewritable rename operations. Moreover, some Iceberg users have already used HadoopCatalog to maintain metadata files on HDFS while store data files on object stores, see: https://lists.apache.org/thread/rkg1cnmcl102o8g9ko5l0o152jzgpglm. Therefore, this is a solution that theoretically does not have transaction consistency issues and has been validated in production environments which could leverage object stores.

We can expand the capabilities of our Iceberg Hadoop catalog and achieve the followings:
 - 1. Support for setting independent data write path for Iceberg tables, which is also a native capability supported by Iceberg lib. This way, when creating a table, we can specify an independent location for its actual data, such as a path on S3.
 - 2. Add a configuration parameter for Iceberg Hadoop catalog, such as `iceberg.catalog.warehouse.datadir`, which represents the default data writing root directory for newly created tables in the entire catalog. If this value is configured, all newly created tables will default to setting their data write path based on this root dir, unless explicitly specified the table's data write path in table creation statement.

After doing these, in production environment, we can configure "iceberg.catalog.warehouse" as a locally deployed HDFS path and "iceberg.catalog.warehouse.datadir" as an S3 path, to safely utilize the powerful storage capabilities of object stores.

[...]

PR #24354, opened 2025-01-13
Use jitpack snapshot release for release-notes.sh

ZacBlanco:
## Description

Changes the URL to download the release notes tool executable jar.

## Motivation and Context

There were updates added a while back in https://github.com/prestodb/presto-release-tools/commit/0194b7ac6e6aee4d176163fe7b2ef9e660ea2e14 which never got used because we are a bit slow on publishing releases. Using the jitpack repository will allow us to use the latest code without having to publish a release.

## Impact

N/A

## Test Plan

Verified download worked

## Contributor checklist

- [x] Please make sure your submission complies with our [contributing guide](https://github.com/prestodb/presto/blob/master/CONTRIBUTING.md), in particular [code style](https://github.com/prestodb/presto/blob/master/CONTRIBUTING.md#code-style) and [commit standards](https://github.com/prestodb/presto/blob/master/CONTRIBUTING.md#commit-standards).
- [x] PR description addresses the issue accurately and concisely.  If the change is non-trivial, a GitHub Issue is referenced.
- [x] Documented new properties (with its default value), SQL syntax, functions, or other functionality.
- [x] If release notes are required, they follow the [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines).
- [x] Adequate tests were added if applicable.
- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

PR #24397, opened 2025-01-18
[Iceberg]Support setting warehouse data directory for Hadoop catalog

hantangwangd:
## Description

This PR enable Presto Iceberg Hadoop catalog to specify an independent warehouse data directory to store table data files, in this way, we can manage metadata files on HDFS and store data files on Object Stores in a formal production environment.

See issue: #24383 

## Motivation and Context

Enabling Presto Iceberg to leverage the powerful capabilities of object storages

## Impact

Hadoop catalog has the capability of leveraging object stores

## Test Plan

 - Build an object storage environment base on MinIO docker, configure `iceberg.catalog.warehouse` to a local file path, and `iceberg.catalog.hadoop.warehouse.datadir` to a s3 path, fully run `IcebergDistributedTestBase`, `IcebergDistributedSmokeTestBase`, and `TestIcebergDistributedQueries` in CI test

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

Iceberg Connector Changes
* Add table properties ``write.data.path`` to specify independent data write paths for Iceberg tables
* Add connector configuration property ``iceberg.catalog.hadoop.warehouse.datadir`` for Hadoop catalog to specify root data write path for its new created tables
```


presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergUtil.java
```diff
@@ -1120,9 +1110,20 @@ public void close()
...
+
+        String writeDataLocation = IcebergTableProperties.getWriteDataLocation(tableMetadata.getProperties());
+        if (!Strings.isNullOrEmpty(writeDataLocation)) {
+            if (catalogType.equals(CatalogType.HADOOP)) {
+                propertiesBuilder.put(WRITE_DATA_LOCATION, writeDataLocation);
+            }
+            else {
+                throw new PrestoException(NOT_SUPPORTED, "Not support set write_data_path on catalog: " + catalogType);
+            }
+        }
```
agrawalreetika, line 1125: Would it better to have logic for setting `WRITE_DATA_LOCATION` here itself? https://github.com/prestodb/presto/pull/24397/files#diff-1bedde049adaf86f49d050ee5b7dfd4584eb9669321946ece016873970d4e9f1R324

hantangwangd: Before setting a warehouse data dir based location, We first need to check 'write_data_path' in the table creation statement, and ensure that non-Hadoop catalog including Hive catalog do not allow this property to be explicitly set. So I extracted this part of logic to `IcebergUtil.populateTableProperties(...)`, and invoked the it from both `IcebergNativeMetadata` and `IcebergHiveMetadata`. Seems we cannot just do it in `IcebergNativeMetadata`. Do you think this makes sense?

agrawalreetika: Ohh sorry I missed, I was suggesting it to be in `IcebergUtil.populateTableProperties(...)` so it should be `propertiesBuilder.put` not `tableProperties.put` (Edited the code suggestion above)

Also in the above since we are already checking if its not `HADOOP` catalog we are throwing exception. Do you think we need to check anything more here?  Or I missing something here?

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -260,11 +260,34 @@ Property Name                                           Description
                                                         This property is required if the ``iceberg.catalog.type`` is
                                                         ``hadoop``.
 
+``iceberg.catalog.warehouse.datadir``                   The catalog warehouse root data path for Iceberg tables.
+                                                        Example: ``s3://iceberg_bucket/warehouse``
+                                                        If set, all tables in this hadoop catalog will default to
```
steveburnett, line 265:
```suggestion
                                                        If set, all tables in this Hadoop catalog will default to
```

steveburnett (changes requested):
Thanks for the doc! Looks good, only a few nits only of punctuation and capitalization.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergUtil.java
hantangwangd: Good suggestion, got your point now. I moved the method `populateTableProperties(...)` from `IcebergUtil` to `IcebergAbstractMetadata`, and put the entire setting logic for `WRITE_DATA_LOCATION ` you suggested above into it,  so that there is no need to perform checks and settings in `IcebergNativeMetadata` again. Please take a look when available, thanks!

steveburnett (dismissed):
LGTM! (docs)

Pulled updated branch, new local doc build, looks good. Thanks!

hantangwangd:
Thanks @steveburnett for your suggestion, fixed!

agrawalreetika (dismissed):
Thanks for the change. LGTM

presto-iceberg/pom.xml
```diff
@@ -263,6 +263,28 @@
             </exclusions>
         </dependency>
 
+        <dependency>
```
imjalpreet, line 266: nit: Can we move this test dependency to follow the other test dependencies after the comment `<!-- for testing -->`? 

https://github.com/prestodb/presto/blob/5fd6b9cdedcbb1b9a8b05951e14273882d586938/presto-iceberg/pom.xml#L517C9-L517C29

imjalpreet (commented):
@hantangwangd Thanks for the PR! I took a first pass and had some minor comments.

ZacBlanco (commented):
I just have a few high level comments. At the core, I understand what we're trying to solve, but I'm not sure yet if this is the right solution. What happens when we have a table which already exists at the warehouse directory but has a data directory which doesn't align with the new datadir property? Should we respect the property or the table in the case of an insert? This could be confusing for users. 

And if the table already exists and doesn't doesn't have a metadata folder which is in a "safe" filesystem, should we error, or warn the user? Do we even have a way of knowing that a filesystem is "safe" (supports atomic renames)?

Correct me if I'm wrong, but in theory we could have already supported this use case within the iceberg connector by using SetTableProperty procedure and just setting "write.data.path" on the individual tables, right? From my understanding, all this change does is provide a default for new tables and makes it a viewable table property. I'm wondering if it might be better providing this as schema-level property that users can set, similar to how the hive connector has a schema-level "location" property. Then, we can set defaults for the data path on schema creation, but override it on a table-level if we prefer. 

However, I don't believe the metadata path could be set that way since the HadoopCatalog relies on listing the warehouse metadata directories to get the schemas and tables.

Just some thoughts for discussion. I think I just want to refine our approach and and make sure there isn't any ambiguous behavior from the user perspective.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergConfig.java
```diff
@@ -127,6 +128,19 @@ public IcebergConfig setCatalogWarehouse(String catalogWarehouse)
         return this;
     }
 
+    public String getCatalogWarehouseDataDir()
+    {
+        return catalogWarehouseDataDir;
+    }
+
+    @Config("iceberg.catalog.warehouse.datadir")
+    @ConfigDescription("Iceberg catalog default root data writing directory")
```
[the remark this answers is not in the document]
hantangwangd: > Does it make sense to add a check to throw an error at the server startup if this config is present in the config file when the catalog type is not Hadoop?

I added this check in `IcebergCommonModule`, but I'm not quite sure if we should throw an error or just ignore this property in non-Hadoop catalogs. Currently, `iceberg.catalog.warehouse` will be just ignored instead of throw an error. Perhaps we can listen to what other reviewers are thinking @tdcmeehan @ZacBlanco @agrawalreetika @kiersten-stokes?

> should we mention that it is only supported for Hadoop catalog in the ConfigDescription?

Sure, added!

presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedSmokeTestBase.java
```diff
@@ -150,6 +152,34 @@ public void testShowCreateTable()
                         ")", schemaName, getLocation(schemaName, "orders")));
     }
 
+    @Test
+    public void testTableWithSpecifiedWriteDataLocation()
```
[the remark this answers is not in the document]
hantangwangd: I think they look the same mainly because they are both implementations which implies not supported by default in the base test class. And in the relevant subclass of Hadoop catalog, they will be overwritten with different implementation logic.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedSmokeTestBase.java
```diff
@@ -150,6 +152,34 @@ public void testShowCreateTable()
...
+        assertQueryFails(String.format("create table test_table_with_specified_write_data_location2(a int, b varchar) with (write_data_path = '%s')", dataWriteLocation),
+                "Not support set write_data_path on catalog: " + catalogType);
+    }
+
+    @Test
+    public void testPartitionedTableWithSpecifiedWriteDataLocation()
+            throws IOException
+    {
+        String dataWriteLocation = Files.createTempDirectory("test_table_with_specified_write_data_location3").toAbsolutePath().toString();
+        assertQueryFails(String.format("create table test_table_with_specified_write_data_location3(a int, b varchar) with (write_data_path = '%s')", dataWriteLocation),
```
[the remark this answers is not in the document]
hantangwangd: Sure, done!

presto-iceberg/pom.xml
hantangwangd: Sure, done!

presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergQueryRunner.java
```diff
@@ -202,13 +203,11 @@ public static DistributedQueryRunner createIcebergQueryRunner(
...
-                .putAll(getConnectorProperties(CatalogType.valueOf(catalogType), icebergDataDirectory))
-                .putAll(extraConnectorProperties)
-                .build();
-
-        queryRunner.createCatalog(ICEBERG_CATALOG, "iceberg", icebergProperties);
+        Map<String, String> icebergProperties = new HashMap<>();
+        icebergProperties.put("iceberg.file-format", format.name());
+        icebergProperties.putAll(getConnectorProperties(CatalogType.valueOf(catalogType), icebergDataDirectory));
+        icebergProperties.putAll(extraConnectorProperties);
+        queryRunner.createCatalog(ICEBERG_CATALOG, "iceberg", ImmutableMap.copyOf(icebergProperties));
```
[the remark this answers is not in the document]
hantangwangd: Yes, we need a way to let the properties explicitly set in the test class to override the default properties set by the underlying query runner. For example, we need to explicitly specify `iceberg.catalog.warehouse` in the test class.

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -370,6 +393,9 @@ Property Name                             Description
 ``location``                               Optionally specifies the file system location URI for
                                            the table.
 
+``write_data_path``                        Optionally specifies the file system location URI for
```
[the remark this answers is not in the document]
hantangwangd: Agree, I think this makes sense. Then how do you think we should handle the previous table properties?

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -260,11 +260,34 @@ Property Name                                           Description
...
+                                                        data directory.
+
 ``iceberg.catalog.cached-catalog-num``                  The number of Iceberg catalogs to cache. This property is     ``10``
                                                         required if the ``iceberg.catalog.type`` is ``hadoop``.
                                                         Otherwise, it will be ignored.
 ======================================================= ============================================================= ============
 
+Configure the `Amazon S3 <https://prestodb.io/docs/current/connector/hive.html#amazon-s3-configuration>`_
+properties to specify a S3 location as the warehouse data directory for the Hadoop catalog. This way,
+the data and delete files of Iceberg tables are stored in S3. An example configuration includes:
```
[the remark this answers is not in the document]
hantangwangd: Sure, will supplement the document. To clarify, warehouse data dir needn't to be on a filesystem which supports atomic renames, only warehouse dir (maintaining metadata files) need.

hantangwangd:
@ZacBlanco Thanks for your feedback, very glad to be able to clarify and refine the approch through discussion. Overall, I think the core point of this change is to separate the data write path and metadata path of Iceberg tables in Hadoop catalog, and provide a simpler and more convenient way for users to specify the data write path of tables. I will reply item by item below. Please let me know if you have any comments.

> What happens when we have a table which already exists at the warehouse directory but has a data directory which doesn't align with the new datadir property? Should we respect the property or the table in the case of an insert?

Undoubtedly, when inserting data, the iceberg table's table property `write.data.path` will be respected.  In general, `iceberg.catalog.warehouse.datadir` is a Hadoop catalog level configuration property used to define the default data write path on table creating. It will not affect the data write path of existing tables.

 - We can explicitly specify the write data path of a table when creating it, and later change the write data path at any time through the `SetTableProperty` procedure, so the write data path of a table does not need to be aligned with the warehouse data dir.

 - We can change the location of the warehouse data dir at any time, which will only affect newly created tables, and will not affect the data write path of existing tables.

 - When we want to change the data write path of a table, we should explicitly use the SetTableProperty procedure (which needs to support property `write.data.location`). And the strategy of the Iceberg table itself is also similar, that is, the existing data files will not be affected by the later modified `write.data.location`.

Do you think the above behaviors make sense?

> And if the table already exists and doesn't doesn't have a metadata folder which is in a "safe" filesystem, should we error, or warn the user? Do we even have a way of knowing that a filesystem is "safe" (supports atomic renames)?

I think we could specify the requirements for the file system where the metadata is located in the document. But I don't think Presto engine is responsible for identifying and distinguishing specific file systems and reminding users. This is the user's own responsibility, because they have a better understanding of their own usage scenarios. For example, for a considerable number of users, they do not need to consider transaction consistency issues caused by cross engine interoperability.

> in theory we could have already supported this use case within the iceberg connector by using SetTableProperty procedure and just setting "write.data.path" on the individual tables

`SetTableProperty` procedure and `CreateTable` can both specify and set table properties, but their usage scenarios are different. I don't think `SetTableProperty` can replace `CreateTable` to specify table properties, its main purpose is to alter table properties. And I think users may also not be able to accept the requirement to execute set properties procedure after creating a table in order to specify the data write path. What do you think?

> I'm wondering if it might be better providing this as schema-level property that users can set, similar to how the hive connector has a schema-level "location" property. Then, we can set defaults for the data path on schema creation, but override it on a table-level if we prefer.

Firstly, for Hadoop catalog, we cannot arbitrarily set the value of `location` which also maintains metadata data like hive. Therefore, we should only set a property like `data.location` which only contains data files. Secondly, I don't think this schema-level concept you mentioned conflicts with the catalog level data dir. Their roles are similar, just at different level. As described above, they all used to define the default data write path on table creating, they will not affect the data write path of existing tables, and can be overridden on a table-level if we prefer.

Therefore, if necessary, we can add this schema-level data dir property in the future. I think it can work well with catalog-level data dir. What's your opinion?

steveburnett:
New release note guidelines as of last week: PR #24354 automatically adds links to this PR to the release notes. Please remove the manual PR link in the following format from the release note entries for this PR. 

```
:pr:`12345`
```

ZacBlanco:
Thanks for your response. I did some more thinking around this and concluded that we don't really have a consistent way of specifying location for each connector. But I think we should strive to do so.

Some examples:

1. In the hive connector, we allow schemas to set the `location` property. It can also be set through a `external_location` table property, otherwise it is controlled by the HMS configuration (external to Presto).
2. In the iceberg connector when using the Hive catalog, the location is determined by either the HMS configuration (external to presto), or configured by the user specifying a `location` property.
3. In the Iceberg REST connector, table location is specified via the catalog-level `iceberg.catalog.warehouse` property.


My thoughts are that 

1. We should definitely provide the ability to set both the iceberg `write.data.location` and `write.metadata.location` through the table properties in the `CREATE TABLE` statements. I think that we should use the iceberg property names directly rather than doing something like `location` or `write_data_location` and `write_metadata_location`.
2. I think providing this configuration in Presto's iceberg catalog configuration for the `HadoopCatalog` inthis PR is probably fine because the difference with the HMS and Hive connector is that it's an external process with its own configuration. With the hadoop catalog, we don't really have that option since it's all filesystem-based.
  a. Though this makes me wonder whether the RestCatalog should utilize this property as I might expect RestCatalog implementations to have some kind of similar default data and metadata locations. I looked at [Nessie's docs](https://projectnessie.org/nessie-latest/configuration/#warehouse-defaults) and saw that they have the ability to specify default locations. I think we might want to ensure that those configs are respected over any Presto configurations. We may even want to remove the ability to specify default locations for the Presto iceberg catalog when configured with Rest/Nessie

presto-docs/src/main/sphinx/connector/iceberg.rst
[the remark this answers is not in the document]
ZacBlanco: We don't have to do it in this PR but here's a solution I think would work for deprecation of the old names:

Inside of `IcebergTableProperties` create a map of (new property name -> old property name) say, `deprecatedProperties`. 

Then we update all existing table properties with iceberg-equivalents to the iceberg name (if one exists).

Inside of the constructor or `getTableProperties` method of `IcebergTableProperties`, we take the `deprecatedProperties` map and iterate over of the existing properties and if it has an entry in the `deprecatedProperties` map,  add a new hidden property with the old name and same parameters as the new property.

In each of the `getXXX` methods, we would also need to fall back to checking for the old property name in the map. We would also need a method to add a warning to the session that the table property name is deprecated. We can leave it like that for a few releases and eventually remove the deprecated property names.

presto-docs/src/main/sphinx/connector/iceberg.rst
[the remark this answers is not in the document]
ZacBlanco: Ah yes, mis-typed. I meant metadata :) . We should just be clear to users what is going to be safe/unsafe

presto-iceberg/src/test/java/com/facebook/presto/iceberg/hadoop/TestIcebergDistributedOnS3Hadoop.java
ZacBlanco, line 1: I'm a little concerned about the impact of runtime of the CI when adding this test class and the `TestIcebergSmokeOnS3Hadoop` test class. These classes take quite a bit of time to complete, so I think we should try to be prudent about adding new configurations permutations that run through the whole gamut of tests.

That being said, since the only real difference in the configuration is the warehouse default data dir, and the fact we're running on S3 I think we ought to enable just one version of this test. Either the default which operates directly on the local filesystem or this one. I would actually err towards using this class because it represents something closer to real-world use case while still exercising the Hadoop catalog.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/hadoop/TestIcebergDistributedOnS3Hadoop.java
```diff
@@ -0,0 +1,133 @@
...
+import org.testng.annotations.BeforeClass;
+
+import java.net.URI;
+
+import static com.facebook.presto.iceberg.CatalogType.HADOOP;
+import static com.facebook.presto.iceberg.container.IcebergMinIODataLake.ACCESS_KEY;
+import static com.facebook.presto.iceberg.container.IcebergMinIODataLake.SECRET_KEY;
+import static com.facebook.presto.testing.TestingConnectorSession.SESSION;
+import static com.facebook.presto.tests.sql.TestTable.randomTableSuffix;
+import static com.google.common.io.Files.createTempDir;
```
ZacBlanco, line 43: nit: can we use the java.nio method over the guava one?

presto-iceberg/src/test/java/com/facebook/presto/iceberg/hadoop/TestIcebergHadoopCatalogOnS3DistributedQueries.java
```diff
@@ -0,0 +1,91 @@
...
+import com.google.common.collect.ImmutableMap;
+import com.google.common.net.HostAndPort;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
+
+import static com.facebook.presto.iceberg.CatalogType.HADOOP;
+import static com.facebook.presto.iceberg.container.IcebergMinIODataLake.ACCESS_KEY;
+import static com.facebook.presto.iceberg.container.IcebergMinIODataLake.SECRET_KEY;
+import static com.facebook.presto.tests.sql.TestTable.randomTableSuffix;
+import static com.google.common.io.Files.createTempDir;
```
ZacBlanco, line 29: same thing on using jdk over guava for this. `createTempDir` is deprecated

presto-iceberg/src/test/java/com/facebook/presto/iceberg/hadoop/TestIcebergDistributedOnS3Hadoop.java
```diff
@@ -0,0 +1,133 @@
...
+    protected HdfsEnvironment getHdfsEnvironment()
+    {
+        HiveClientConfig hiveClientConfig = new HiveClientConfig();
+        MetastoreClientConfig metastoreClientConfig = new MetastoreClientConfig();
+        HiveS3Config hiveS3Config = new HiveS3Config()
+                .setS3AwsAccessKey(ACCESS_KEY)
+                .setS3AwsSecretKey(SECRET_KEY)
+                .setS3PathStyleAccess(true)
+                .setS3Endpoint(format("http://%s:%s", hostAndPort.getHost(), hostAndPort.getPort()))
+                .setS3UseInstanceCredentials(false);
```
ZacBlanco, line 123: is there any reason we need to set this? It's also a deprecated config

presto-iceberg/src/test/java/com/facebook/presto/iceberg/hadoop/TestIcebergSmokeOnS3Hadoop.java
```diff
@@ -0,0 +1,461 @@
...
+        IcebergConfig icebergConfig = new IcebergConfig();
+        icebergConfig.setCatalogType(HADOOP);
+        icebergConfig.setCatalogWarehouse(getCatalogDirectory().toString());
+
+        HiveS3Config hiveS3Config = new HiveS3Config()
+                .setS3AwsAccessKey(ACCESS_KEY)
+                .setS3AwsSecretKey(SECRET_KEY)
+                .setS3PathStyleAccess(true)
+                .setS3Endpoint(format("http://%s:%s", hostAndPort.getHost(), hostAndPort.getPort()))
+                .setS3UseInstanceCredentials(false);
```
ZacBlanco, line 435: any reason to set?

presto-iceberg/src/test/java/com/facebook/presto/iceberg/hadoop/TestIcebergSmokeOnS3Hadoop.java
```diff
@@ -0,0 +1,461 @@
...
+                "format = '" + fileFormat + "', " +
+                "partitioning = ARRAY['ORDER_STATUS', 'Ship_Priority', 'Bucket(order_key,9)']" +
+                ") " +
+                "AS " +
+                "SELECT orderkey AS order_key, shippriority AS ship_priority, orderstatus AS order_status " +
+                "FROM tpch.tiny.orders";
+
+        assertUpdate(session, createTable, "SELECT count(*) from orders");
+
+        String createTableSql = format("" +
```
ZacBlanco, line 185: It feels weird to use `format` but then also do string concatenation with `+` in the middle for `fileFormat`. Can we try to be consistent? I think it's fine for the line breaks though

presto-iceberg/src/test/java/com/facebook/presto/iceberg/hadoop/TestIcebergSmokeOnS3Hadoop.java
```diff
@@ -0,0 +1,461 @@
...
+                        "   location = '%s',\n" +
+                        "   metadata_delete_after_commit = false,\n" +
+                        "   metadata_previous_versions_max = 100,\n" +
+                        "   metrics_max_inferred_column = 100,\n" +
+                        "   partitioning = ARRAY['order_status','ship_priority','bucket(order_key, 9)'],\n" +
+                        "   write_data_path = '%s'\n" +
+                        ")",
+                getSession().getCatalog().get(),
+                getSession().getSchema().get(),
+                "test_create_partitioned_table_as_" + fileFormat.toString().toLowerCase(ENGLISH),
```
ZacBlanco, line 204: There's a lot of `fileFormat.toString().toLowerCase(ENGLISH)` in this method. Maybe just extract to a local variable? It might be a little easier to read

presto-iceberg/src/test/java/com/facebook/presto/iceberg/hadoop/TestIcebergSmokeOnS3Hadoop.java
```diff
@@ -0,0 +1,461 @@
...
+                .setS3PathStyleAccess(true)
+                .setS3Endpoint(format("http://%s:%s", hostAndPort.getHost(), hostAndPort.getPort()))
+                .setS3UseInstanceCredentials(false);
+
+        IcebergNativeCatalogFactory catalogFactory = new IcebergNativeCatalogFactory(icebergConfig,
+                new IcebergCatalogName(ICEBERG_CATALOG),
+                new PrestoS3ConfigurationUpdater(hiveS3Config),
+                new HiveGcsConfigurationInitializer(new HiveGcsConfig()));
+
+        return IcebergUtil.getNativeIcebergTable(catalogFactory,
```
ZacBlanco, line 442: static import?

ZacBlanco (commented):
Thanks for the great work - I do have a concern about the additional load on CI the two new test classes might have. Please let me know your thoughts

hantangwangd:
@ZacBlanco Thanks for your investigation and perspective. I think a significant portion of the current inconsistency is due to Iceberg connector with Hive catalog. In presto, its configuration is closer to Hive connector than Iceberg lib's Hive catalog, because it is a presto native implementation based on `presto-hive-metastore` rather than Iceberg lib's Hive catalog.

As far as I can see, in Iceberg lib, `warehouse` is a commonly used catalog properties, see [Iceberg catalog properties](https://iceberg.apache.org/docs/nightly/configuration/#catalog-properties).  It has a role in all catalogs: for Hadoop/Nessie/Jdbc catalogs, it is a mandatory property to be configured, while in Hive/Rest/InMemory catalogs, it is an optional property. So it seems that in presto, catalog property `iceberg.catalog.warehouse` (which maps to the catalog property `warehouse` of Iceberg lib) should also be a common catalog property in Iceberg connector and should not be removed when configured with Rest/Nessie.

As you said, the newly added configuration property is an external process with Iceberg's own configuration, and it should only be used for Hadoop catalog. Therefore, in order to avoid the confusion with property `iceberg.catalog.warehouse` and causing users to believe that it is also a commonly used catalog property, I think we can make it more clear by doing the following things:

 - 1. Name this newly added configuration property `iceberg.catalog.hadoop.warehouse.datadir` instead of `iceberg.catalog.warehouse.datadir`.
 - 2. Add a check to throw an error at the server startup if this config property is present when the catalog type is not Hadoop, as suggested by @imjalpreet .
 - 3. Clearly state in the document that this is a Hdoop catalog specific configuration property.

Do you think this makes sense?

presto-iceberg/src/test/java/com/facebook/presto/iceberg/hadoop/TestIcebergDistributedOnS3Hadoop.java
hantangwangd: No, maybe I did some experiments, but forgot to delete them. I have deleted them all.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/hadoop/TestIcebergSmokeOnS3Hadoop.java
hantangwangd: Sure, fixed!

presto-iceberg/src/test/java/com/facebook/presto/iceberg/hadoop/TestIcebergSmokeOnS3Hadoop.java
hantangwangd: Good suggestion, done.

hantangwangd:
> New release note guidelines as of last week: PR #24354 automatically adds links to this PR to the release notes. Please remove the manual PR link in the following format from the release note entries for this PR.

Thanks @steveburnett for your message. I have fixed the release note and supplemented some content to the document, please take a look when convenient.

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -260,11 +260,42 @@ Property Name                                           Description
                                                         This property is required if the ``iceberg.catalog.type`` is
                                                         ``hadoop``.
 
+``iceberg.catalog.hadoop.warehouse.datadir``            The catalog warehouse root data path for Iceberg tables.
+                                                        Example: ``s3://iceberg_bucket/warehouse``.
+                                                        This property is only supported with Hadoop catalog. If set,
+                                                        all tables in this Hadoop catalog will default to saving
```
steveburnett, line 266:
```suggestion
                                                        all tables in this Hadoop catalog default to saving
```

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -260,11 +260,42 @@ Property Name                                           Description
...
+
 ``iceberg.catalog.cached-catalog-num``                  The number of Iceberg catalogs to cache. This property is     ``10``
                                                         required if the ``iceberg.catalog.type`` is ``hadoop``.
                                                         Otherwise, it will be ignored.
 ======================================================= ============================================================= ============
 
+.. note::
+
+    Hadoop catalog requires a file system that supports atomic rename operation, such as HDFS,
+    to maintain metadata files in order to implement atomic transaction commit. It's unsafe to
```
steveburnett, line 278:
```suggestion
    to maintain metadata files in order to implement atomic transaction commits. Do not
```

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -260,11 +260,42 @@ Property Name                                           Description
...
+    set `iceberg.catalog.warehouse` to a path in object stores and local file systems. But it's
```
steveburnett, line 279:
```suggestion
    set `iceberg.catalog.warehouse` to a path in object stores and local file systems. 
```

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -260,11 +260,42 @@ Property Name                                           Description
...
+    safe to set `iceberg.catalog.hadoop.warehouse.datadir` to a path in the object stores.
```
steveburnett, line 280:
```suggestion
    Set `iceberg.catalog.hadoop.warehouse.datadir` to a path in the object stores.
```

steveburnett (changes requested):
Thanks for the expanded documentation! I made a few minor revision suggestions to use imperative, to emphasize what the user should and should not do.

presto-docs/src/main/sphinx/connector/iceberg.rst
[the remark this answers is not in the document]
hantangwangd: It sounds good. I have changed the newly added table property name to `data.write.path`, which is exactly the same as Iceberg's corresponding table property name. Please take a look when available, thanks!

presto-docs/src/main/sphinx/connector/iceberg.rst
hantangwangd: Thank you @steveburnett for the suggestions. One little difference here is that we want to express that it is safe for users to set `iceberg.catalog.hadoop.warehouse.datadir` to a path in object stores. That is to say, they can point it to an S3 path, an HDFS path, or not specify it at all. Can you give some more advice about it?

presto-iceberg/src/test/java/com/facebook/presto/iceberg/hadoop/TestIcebergDistributedOnS3Hadoop.java
ZacBlanco: have you tried just putting `@Test(enabled = false)` at the class-level?

ZacBlanco:
> @ZacBlanco Thanks for your investigation and perspective. I think a significant portion of the current inconsistency is due to Iceberg connector with Hive catalog. In presto, its configuration is closer to Hive connector than Iceberg lib's Hive 
> As you said, the newly added configuration property is an external process with Iceberg's own configuration, and it should only be used for Hadoop catalog. Therefore, in order to avoid the confusion with property `iceberg.catalog.warehouse` and causing users to believe that it is also a commonly used catalog property, I think we can make it more clear by doing the following things:
> 
>     * 1. Name this newly added configuration property `iceberg.catalog.hadoop.warehouse.datadir` instead of `iceberg.catalog.warehouse.datadir`.
> 
>     * 2. Add a check to throw an error at the server startup if this config property is present when the catalog type is not Hadoop, as suggested by @imjalpreet .
> 
>     * 3. Clearly state in the document that this is a Hdoop catalog specific configuration property.
> 
> 
> Do you think this makes sense?


This approach sounds good to me

presto-docs/src/main/sphinx/connector/iceberg.rst
hantangwangd: So much appreciate for your advice. I like the second option too. Made some little modifications to your wording, because object stores and HDFS are two different types of storage. Please take a look when available.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/hadoop/TestIcebergDistributedOnS3Hadoop.java
hantangwangd: It doesn't seem to work. I previously used @Ignore, but it also didn't work.

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -256,15 +256,50 @@ Property Name                                           Description
 ======================================================= ============================================================= ============
 ``iceberg.catalog.warehouse``                           The catalog warehouse root path for Iceberg tables.
 
+                                                        Hadoop catalog requires a file system that supports atomic
+                                                        rename operation, such as HDFS, to maintain metadata files
+                                                        in order to implement atomic transaction commit.
+
                                                         Example: ``hdfs://nn:8020/warehouse/path``
+
+                                                        Do not set `iceberg.catalog.warehouse` to a path in object
```
steveburnett, line 265:
```suggestion
                                                        Do not set ``iceberg.catalog.warehouse`` to a path in object
```

Formatting nit.

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -256,15 +256,50 @@ Property Name                                           Description
...
+                                                        stores or local file systems in the production environment.
+
                                                         This property is required if the ``iceberg.catalog.type`` is
                                                         ``hadoop``.
 
+``iceberg.catalog.hadoop.warehouse.datadir``            The catalog warehouse root data path for Iceberg tables.
+                                                        Example: ``s3://iceberg_bucket/warehouse``.
```
steveburnett, line 272:
```suggestion
                                                        Example: ``s3://iceberg_bucket/warehouse``.
                                                        
```

Suggest adding some blank lines to space the sentences the same way as in the Description for ``iceberg.catalog.warehouse``. Here's how the two rows look in a local doc build. The second one looks cramped when compared to the first. 

Screenshot 2025-02-05 at 1 26 36 PM

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -256,15 +256,50 @@ Property Name                                           Description
...
+                                                        This property is only supported with Hadoop catalog. If set,
```
steveburnett, line 273:
```suggestion
                                                        This property is only supported with Hadoop catalog. 
                                                        
                                                        If set,
```

steveburnett (changes requested):
Thanks for the quick revision! I'm glad you rephrased my suggestion for technical accuracy. Just a few minor nits of formatting and spacing, but everything looks good.

presto-docs/src/main/sphinx/connector/iceberg.rst
hantangwangd: Following your advice, I made some minor adjustments. Please take a look when available, thanks a lot.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
```diff
@@ -1076,4 +1148,20 @@ else if (tableVersion.getVersionExpressionType() instanceof VarcharType) {
         }
         throw new PrestoException(NOT_SUPPORTED, "Unsupported table version type: " + tableVersion.getVersionType());
     }
+
+    protected Optional<String> getWriteDataLocation(Table table)
+    {
+        try {
+            String writeDataLocation = table.properties().get(WRITE_DATA_LOCATION);
+            return Optional.of(Strings.isNullOrEmpty(writeDataLocation) ? table.location() : writeDataLocation);
```
ZacBlanco, line 1156: technically the default from the `LocationProviders` class in iceberg is `table.location() + "/data"

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergNativeMetadata.java
```diff
@@ -376,4 +379,13 @@ public void unregisterTable(ConnectorSession clientSession, SchemaTableName sche
...
         catalogFactory.getCatalog(clientSession).dropTable(toIcebergTableIdentifier(schemaTableName, catalogFactory.isNestedNamespaceEnabled()), false);
     }
+
+    protected Optional<String> getDataLocationBasedOnWarehouseDataDir(SchemaTableName schemaTableName)
+    {
+        if (!catalogType.equals(HADOOP)) {
+            return Optional.empty();
+        }
+        return Optional.ofNullable(warehouseDataDir.map(base -> base + schemaTableName.getSchemaName() + "/" + schemaTableName.getTableName())
+                .orElse(null));
```
ZacBlanco, line 389: `Optional.ofNullable` is redundant here

```suggestion
        return warehouseDataDir.map(base -> base + schemaTableName.getSchemaName() + "/" + schemaTableName.getTableName());
```

presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedSmokeTestBase.java
```diff
@@ -150,6 +152,34 @@ public void testShowCreateTable()
                         ")", schemaName, getLocation(schemaName, "orders")));
     }
 
+    @Test
+    public void testTableWithSpecifiedWriteDataLocation()
+            throws IOException
+    {
+        String dataWriteLocation = Files.createTempDirectory("test_table_with_specified_write_data_location2").toAbsolutePath().toString();
+        assertQueryFails(String.format("create table test_table_with_specified_write_data_location2(a int, b varchar) with (\"write.data.path\" = '%s')", dataWriteLocation),
+                "Table property write\\.data\\.path is not supported with catalog type: " + catalogType);
```
ZacBlanco, line 161: is the `\\` because the `assertQueryFails` argument thinks the `.` is a regex operator?

presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergSystemTablesHadoop.java
```diff
@@ -0,0 +1,50 @@
...
+{
+    @Override
+    protected QueryRunner createQueryRunner()
+            throws Exception
+    {
+        return IcebergQueryRunner.createIcebergQueryRunner(ImmutableMap.of(), HADOOP, ImmutableMap.of());
+    }
+
+    @Test
+    public void testPropertiesTableWithSpecifiedDataWriteLocation()
```
ZacBlanco, line 36: Hmm this is getting me thinking that maybe we ought to have a test class which creates a query runner configured with _all_ of the catalog types. Then we don't have to create new test classes which have the overhead of starting new query runners all the time.

I think this is fine for the PR, but I am concerned that we're creating a lot of overhead in testing for these different config scenarios. I think it might be worth an effort to try to consolidate to reduce the CI times.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/hadoop/TestIcebergSmokeHadoop.java
```diff
@@ -44,18 +46,85 @@ public TestIcebergSmokeHadoop()
...
+                            ")\n" +
+                            "WITH (\n" +
+                            "   delete_mode = 'merge-on-read',\n" +
+                            "   format = 'PARQUET',\n" +
+                            "   format_version = '2',\n" +
+                            "   location = '%s',\n" +
+                            "   metadata_delete_after_commit = false,\n" +
+                            "   metadata_previous_versions_max = 100,\n" +
+                            "   metrics_max_inferred_column = 100,\n" +
+                            "   \"write.data.path\" = '%s'\n" +
```
ZacBlanco, line 72: I think it's ok to leave for now, but I noticed adding a new table property is quite cumbersome because we have to update all of these tests which look for the exact SQL create statement. It's a bit of a burden for maintenance. Do you have any thoughts about how we might be able to make it less fragile?

ZacBlanco (commented):
Mostly nits. A couple questions for discussion

presto-iceberg/src/test/java/com/facebook/presto/iceberg/hadoop/TestIcebergSmokeHadoop.java
hantangwangd: Perhaps we can extract a method that can verify the result of the `show create table` statement, which takes the basic information, columns, and properties fo the table as parameters. All relevant tests can utilize this method, which may result in fewer changes to the test cases when adding new table property in the future.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedSmokeTestBase.java
hantangwangd: Yes, we need to escape `.` here.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergNativeMetadata.java
hantangwangd: Nice catch, fixed!

steveburnett (approved):
LGTM! (docs) 

Pull updated branch, new local doc build, looks good.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
hantangwangd: After rechecking this part of the code, I found that this method seems unnecessary. It's OK to just use `tableLocation` as before. Because when constructing the `LocationProvider` later, we passed both `tableLocation` and `tableProperties` into it. And the `LocationProvider` will find the appropriate write data location based on `tableProperties` and `tableLocation`.

I have refactored this part of code, please take a look when convenient, thanks.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
```diff
@@ -1016,6 +1032,62 @@ public void setTableProperties(ConnectorSession session, ConnectorTableHandle ta
...
+
+        String formatVersion = getFormatVersion(tableMetadata.getProperties());
+        verify(formatVersion != null, "Format version cannot be null");
+        propertiesBuilder.put(TableProperties.FORMAT_VERSION, formatVersion);
+
+        if (parseFormatVersion(formatVersion) < MIN_FORMAT_VERSION_FOR_DELETE) {
+            propertiesBuilder.put(TableProperties.DELETE_MODE, RowLevelOperationMode.COPY_ON_WRITE.modeName());
+        }
+        else {
+            RowLevelOperationMode deleteMode = IcebergTableProperties.getDeleteMode(tableMetadata.getProperties());
```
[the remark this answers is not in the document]
hantangwangd: Seems we should use `IcebergTableProperties.getDeleteMode(...)` here, since we have already imported `IcebergUtil.getDeleteMode`.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/hadoop/TestIcebergDistributedOnS3Hadoop.java
hantangwangd: A feasible way is to declare them as abstract classes, this works well in CI environment, see #24508. Not so sure if there is a better way.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
```diff
@@ -1016,6 +1038,62 @@ public void setTableProperties(ConnectorSession session, ConnectorTableHandle ta
...
+    {
+        ImmutableMap.Builder<String, String> propertiesBuilder = ImmutableMap.builderWithExpectedSize(16);
+
+        String writeDataLocation = getWriteDataLocation(tableMetadata.getProperties());
+        if (!isNullOrEmpty(writeDataLocation)) {
+            if (catalogType.equals(HADOOP)) {
+                propertiesBuilder.put(WRITE_DATA_LOCATION, writeDataLocation);
+            }
+            else {
+                throw new PrestoException(NOT_SUPPORTED, "Table property write.data.path is not supported with catalog type: " + catalogType);
```
ZacBlanco, line 1051: Is this actually the case? If the property is set on the table correctly it will be reflected in the location provider, right? I don't think we need to disable it for other catalogs?

ZacBlanco (commented):
Thanks for the updates. One final question

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
hantangwangd: It's a good question. Enabling this table property directly on all catalogs may cause unexpected problems. For example, if we specify this property in a table on Hive file catalog, when we trying to drop the table, we would fail with the following message:

```
Table test_table contains Iceberg path override properties and cannot be dropped from Presto
```

That would lead in a lot of troubles for test on Iceberg hive catalog.

So I choose not to enable it on other catalogs for now, and will enable them after confirming that there are no issues. Do you think it's OK?

ZacBlanco: I understand that for the Hive file catalog this could cause problems, but I also worry about adding in this restriction that did not exist before for other catalogs such as rest or nessie. It is possible someone could have relied on a separate data path previously, and then this restriction would prevent them from accessing their tables.

Also, the specific error referenced is only on `dropTable`. Read-based queries should still work fine, right?

hantangwangd: Sure, I have enabled this table property on other catalogs as well, and adjusted the test cases accordingly.  Thanks to the fact that property `write.data.path` is only set to the Iceberg's `Table` when explicitly specified, most test cases for hive catalog will not be affected, except the ones that create tables with explicitly specified write data path.

Besides, I have squashed the commits. Due to the many forth and back changes in the middle, I think it may be more concise after squashed. Please take a look when available, thanks.

steveburnett (dismissed):
LGTM! (docs)

ZacBlanco (dismissed):

presto-iceberg/src/test/java/com/facebook/presto/iceberg/hadoop/TestIcebergHadoopCatalogOnS3DistributedQueries.java
```diff
@@ -0,0 +1,95 @@
...
+        this.dockerizedS3DataLake = new IcebergMinIODataLake(bucketName, WAREHOUSE_DATA_DIR);
+        this.dockerizedS3DataLake.start();
+        hostAndPort = this.dockerizedS3DataLake.getMinio().getMinioApiEndpoint();
+        super.init();
+    }
+
+    @AfterClass
+    public void tearDown()
+    {
+        if (dockerizedS3DataLake != null) {
```
agrawalreetika, line 80: Same as above. Should this be marked as `@AfterClass(alwaysRun = true)`, considering the case when execution of `this.dockerizedS3DataLake.start();` fails?

agrawalreetika (changes requested):

presto-iceberg/src/test/java/com/facebook/presto/iceberg/hadoop/TestIcebergHadoopCatalogOnS3DistributedQueries.java
hantangwangd: Thanks for pointing out this @agrawalreetika, I have added `alwaysRun = true` to all the there newly added test classes that based on `dockerizedS3DataLake`. Please take a look when convenient.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/container/IcebergMinIODataLake.java
```diff
@@ -0,0 +1,117 @@
...
+        return minIOContainer;
+    }
+
+    @Override
+    public void close()
+            throws IOException
+    {
+        stop();
+    }
+}
```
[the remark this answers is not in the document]
hantangwangd: I think it's not very suitable to reuse `HiveMinIODataLake`, as they have many differences. `HiveMinIODataLake` needs to build a Hive on HDFS as a metastore in addition to an object storage, while `IcebergMinIODataLake` only builds an object storage and then creates a warehouse dir path on it. Moreover, the amount of code in the entire class is not very large, and there is not a lot of duplicate code between them. So I choose not to reuse it, do you think this makes sense?

presto-iceberg/src/test/java/com/facebook/presto/iceberg/container/IcebergMinIODataLake.java
```diff
@@ -0,0 +1,117 @@
...
+                            "us-east-1"))
+                    .withPathStyleAccessEnabled(true)
+                    .withCredentials(new AWSStaticCredentialsProvider(
+                            new BasicAWSCredentials(ACCESS_KEY, SECRET_KEY)))
+                    .build();
+            s3Client.createBucket(this.bucketName);
+            s3Client.putObject(this.bucketName, this.warehouseDir, "");
+        }
+        finally {
+            isStarted.set(true);
```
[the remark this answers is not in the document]
hantangwangd: > If an exception occurs before or on starting MinIO, do we still want to mark it as started?

I believe it's more suitable to mark `isStarted` as  true in the finally clause, it seems that it will not bring any averse effects, but just give `close()` a change to try to release the resources. For example, when `minIOContainer` successfully starts but bucket or directory creation fails on it. Do you think it makes sense?

> ... and we should catch the failure?

Considering `IcebergMinIODataLake` is only used in test, if there are any exceptions in `start()`, the initialization for the whole test class will fail and the actual exceptions will be thrown and printed. So is there a necessary to catch the exception and do other processing here? I'm open on it, please let me know your thoughts, thanks.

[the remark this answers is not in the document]
agrawalreetika: Thanks for the details. I think while setting it, I think its fine as it is.

agrawalreetika (approved):

ZacBlanco (approved):

PR #24411, opened 2025-01-21
Enhancements to the presto router ui

auden-woolfson:
Enhancements added to the presto-router. Brought from IBM internal router changes

```
== RELEASE NOTES ==

General Changes
* Enhance presto router UI
```

### DIFF
--- a/presto-router/src/main/resources/router_ui/index.html
+++ b/presto-router/src/main/resources/router_ui/index.html
@@ -27,6 +27,12 @@
     <!-- Sparkline -->
     <script type="text/javascript" src="vendor/jquery.sparkline/jquery.sparkline.min.js"></script>
 
+    <script type="text/javascript" >
+        jQuery.htmlPrefilter = function( html ) {
+            return html;
+        };
+    </script>
+
     <!-- CSS loader -->
     <link href="vendor/css-loaders/loader.css" rel="stylesheet">
 
--- a/presto-router/src/main/resources/router_ui/src/components/PageTitle.jsx
+++ b/presto-router/src/main/resources/router_ui/src/components/PageTitle.jsx
@@ -45,7 +45,9 @@ export class PageTitle extends React.Component<Props, State> {
     refreshLoop() {
         clearTimeout(this.timeoutId);
         fetch("/v1/info")
-            .then(response => response.json())
+            .then(response => {
+                return response.json()
+            })
             .then(info => {
                 this.setState({
                     info: info,
@@ -120,7 +122,7 @@ export class PageTitle extends React.Component<Props, State> {
                             </table>
                         </div>
                         <div id="navbar" className="navbar-collapse collapse">
-                            <ul className="nav navbar-nav navbar-right">
+                            <ul className="nav navbar-nav navbar-cluster-info-right">
                                 <li>
                                     <span className="navbar-cluster-info">
                                         <span className="uppercase">Environment</span><br/>
--- a/presto-router/src/main/java/com/facebook/presto/router/RouterResource.java
+++ b/presto-router/src/main/java/com/facebook/presto/router/RouterResource.java
@@ -20,6 +20,8 @@ import com.google.inject.Inject;
 
 import javax.annotation.security.PermitAll;
 import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import javax.ws.rs.GET;
 import javax.ws.rs.POST;
 import javax.ws.rs.Path;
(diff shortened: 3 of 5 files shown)


ethanyzhang:
Hi @yhwang @unidevel can you help look at the ui part of this PR? Thx!

ethanyzhang:
@auden-woolfson Is there any co-author you need to add to this?

presto-router/src/main/resources/router_ui/src/components/PageTitle.jsx
```diff
@@ -45,7 +46,14 @@ export class PageTitle extends React.Component<Props, State> {
     refreshLoop() {
         clearTimeout(this.timeoutId);
         fetch("/v1/info")
-            .then(response => response.json())
+            .then(response => {
+                if (response.status === 401) {
```
yhwang, line 50: Can you explain the logic of handling the 401 response code here?
Can this be done by sending a different URL in the fetch above?

presto-router/src/main/resources/router_ui/logout.html
```diff
@@ -0,0 +1,58 @@
+<!DOCTYPE html>
```
yhwang, line 1: Where is this file used? Any corresponding server-side code? I guess on the server-side, the session needs to be cleaned.

yhwang (commented):
UI changes look good to me. However, I left some questions.

presto-router/src/main/resources/router_ui/logout.html
auden-woolfson: Password authentication is added in a separate PR, can move this into that one

presto-router/src/main/resources/router_ui/src/components/PageTitle.jsx
auden-woolfson: This handles the case of expired Oauth2 tokens, since we are not adding the Oauth2 functionality yet I will remove this.

yhwang (dismissed):
LGTM!

However, still need to fix the CLA error.

ethanyzhang:
The first commit "[new resource method to redirect root path to UI](https://github.com/prestodb/presto/pull/24411/commits/67c02665228523856eb98e72f0fb4a25766833c4)" seems to be authored with a different email address that was not associated with your open-source github account, that's why the CLA cannot pass

unidevel (approved):
After removed should good to go, thanks.

auden-woolfson:
@ethanyzhang @yhwang, Thanks for catching the CLA error. I was planning on squashing commits and fixing CLA after the review process. I like to leave commits the way they are until I know the code has been verified in case I need to revert or anything like that.

presto-router/src/main/resources/router_ui/index.html
```diff
@@ -27,6 +27,12 @@
     <!-- Sparkline -->
     <script type="text/javascript" src="vendor/jquery.sparkline/jquery.sparkline.min.js"></script>
 
+    <script type="text/javascript" >
```
yhwang, line 30: This inline script won't cause any issues until CSP is enabled on the presto-router. I just had another PR to fix the CSP errors and found the same inline script. I moved it directly into the index.jsx. So it will be in the index.js directly.

You don't have to fix it now. Just share the info with you if the CSP header needs to be enabled in the presto-router in the future.

yhwang (dismissed):

BryanCutler (approved):
LGTM, is it possible to attach a screenshot of the UI with your changes?

auden-woolfson:
> LGTM, is it possible to attach a screenshot of the UI with your changes?

@BryanCutler actually there seem to be some issues with it, I will update and then request another review

auden-woolfson:
This PR is now ready for final review/merge, here is a screenshot of the UI as well @BryanCutler 

Screenshot 2025-03-12 at 12 34 55 PM

yhwang (approved):
LGTM for the UI changes

unidevel (approved):
LGTM

unidevel:
@auden-woolfson @yhwang I noticed the package `@dagrejs/graphlib` is difference with  `dagre-d3-es` which used by `presto-ui`, it's better to use the same package in future to reduce the maintaince effort.

BryanCutler (approved):
LGTM, but it would be good to show the login part you added in the screenshot if possible

auden-woolfson:
> LGTM, but it would be good to show the login part you added in the screenshot if possible

The PR no longer includes login

BryanCutler:
I see, ok can you update the description then?

yingsu00 (approved):

PR #24412, opened 2025-01-22
Use a wrapper for event loop to handle exceptions gracefully

shangm2:
## Description
1. we saw the following error
Screenshot 2025-01-21 at 16 26 07 and thanks to @arhimondr, the root cause is if running code on event loop and the code throws exception, the event loop thread will be killed but the event loop group will not create a new thread.

## Motivation and Context
1. Use a event loop wrapper to handle exceptions while running httpRemoteTask related jobs on event loop gracefully
2. Also update the code so that we only add listener after the object has been fully constructed.

## Test Plan
1. Local HiveQueryRunner works great
2. Internal verifier test passed

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

Fixs
* Fix how the exceptions thrown from event loop is being handled while performing HttpRemoteTask :pr:`24412`

```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTask.java
+++ b/presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTask.java
@@ -76,7 +76,6 @@ import com.google.common.util.concurrent.SettableFuture;
 import com.sun.management.ThreadMXBean;
 import io.airlift.units.DataSize;
 import io.airlift.units.Duration;
-import io.netty.channel.EventLoop;
 import it.unimi.dsi.fastutil.longs.LongArrayList;
 import org.joda.time.DateTime;
 
@@ -143,7 +142,7 @@ import static java.util.concurrent.TimeUnit.SECONDS;
  * This class now uses an event loop concurrency model to eliminate the need for explicit synchronization:
  * <ul>
  * <li>All mutable state access and modifications are performed on a single dedicated event loop thread</li>
- * <li>External threads submit operations to the event loop using {@code taskEventLoop.execute()}</li>
+ * <li>External threads submit operations to the event loop using {@code safeExecuteOnEventLoop()}</li>
  * <li>The event loop serializes all operations, eliminating race conditions without using locks</li>
  * </ul>
  * <p>
@@ -230,9 +229,9 @@ public final class HttpRemoteTask
     private final DecayCounter taskUpdateRequestSize;
     private final SchedulerStatsTracker schedulerStatsTracker;
 
-    private final EventLoop taskEventLoop;
+    private final HttpRemoteTaskFactory.SafeEventLoop taskEventLoop;
 
-    public HttpRemoteTask(
+    public static HttpRemoteTask createHttpRemoteTask(
             Session session,
             TaskId taskId,
             String nodeId,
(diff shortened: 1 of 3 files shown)


tdcmeehan (commented):
Please update the release notes to indicate that this is a bug fix per our [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines#order-of-changes)?

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTask.java
```diff
@@ -402,9 +484,6 @@ public HttpRemoteTask(
                 updateSplitQueueSpace();
             }
         });
-
-        updateTaskStats();
-        taskEventLoop.execute(this::updateSplitQueueSpace);
```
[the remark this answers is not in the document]
shangm2: Nice catch. Thanks!

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTaskFactory.java
```diff
@@ -89,7 +96,14 @@ public class HttpRemoteTaskFactory
     private final QueryManager queryManager;
     private final DecayCounter taskUpdateRequestSize;
     private final EventLoopGroup eventLoopGroup = new DefaultEventLoopGroup(Runtime.getRuntime().availableProcessors(),
-            new ThreadFactoryBuilder().setNameFormat("task-event-loop-%s").setDaemon(true).build());
+            new ThreadFactoryBuilder().setNameFormat("task-event-loop-%s").setDaemon(true).build())
+    {
+        @Override
+        protected EventLoop newChild(Executor executor, Object... args)
+        {
+            return new SafeEventLoop(this);
```
arhimondr, line 104: You need to pass the `executor` as in default implementation

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTaskFactory.java
```diff
@@ -229,6 +243,49 @@ public RemoteTask createRemoteTask(
...
+                    }
+                    catch (Throwable t) {
+                        log.error("Error running task on event loop. Failing the task.", t);
+                        if (onFail != null) {
+                            onFail.accept(t);
+                        }
+                    }
+                });
+            }
+            catch (Throwable t) {
```
tdcmeehan, line 278: Can you give an example of how this could happen?  My intuition is Netty should ensure it always succeeds and any failure would happen within the execute callback.

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTaskFactory.java
```diff
@@ -229,6 +243,49 @@ public RemoteTask createRemoteTask(
...
+                log.error("Error submitting task to event loop. Failing the task.", t);
+                if (onFail != null) {
+                    onFail.accept(t);
+                }
+            }
+        }
+
+        public void setOnFail(Consumer<Throwable> onFail)
+        {
+            this.onFail = onFail;
```
arhimondr, line 288: In this model a single `EventLoop` can be shared between multiple tasks. It may happen that the wrong task is failed based on what `onFail` handler is set. We may not be able to attach the failure handler on this level and may need to have an another wrapper inside the `HttpRemoteTask` that fails a task on a failure. We may not be able to use this wrapper for every operation (e.g.: when the executor is passed to a different object, say `StateMachine`), but that's okay. It always used to be like that.

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTaskFactory.java
```diff
@@ -229,6 +243,49 @@ public RemoteTask createRemoteTask(
...
+            extends DefaultEventLoop
+    {
+        private Consumer<Throwable> onFail;
+
+        public SafeEventLoop(EventLoopGroup parent)
+        {
+            super(parent);
+        }
+
+        public void execute(Runnable runnable)
```
arhimondr, line 263: `execute` is not the only method that takes a task. I would recommend overriding the `run` method, e.g.:

```
@Override
        public void run()
        {
            for(;;) {
                Runnable task = takeTask();
                if (task != null) {
                    try {
                        runTask(task);
                    }
                    catch (Throwable t) {
                        log.error("Error running task on event loop", t);
                    }
                    updateLastExecutionTime();
                }

                if (confirmShutdown()) {
                    break;
                }
            }
        }
```

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTask.java
```diff
@@ -296,6 +370,7 @@ public HttpRemoteTask(
         requireNonNull(taskEventLoop, "taskEventLoop is null");
 
         this.taskEventLoop = taskEventLoop;
+        this.taskEventLoop.setOnFail(this::failTask);
```
arhimondr, line 373: Instead of attaching an event handler here maybe have a method `onEventLoop`, e.g.:

```
private void safeExecuteOnEventLoop(Runnable r)
    {
        taskEventLoop.execute(() -> {
            try {
                r.run();
            }
            catch (Throwable t) {
                failTask(t);
            }
        });
    }
```

and used it everywhere instead of `taskEventLoop.execute` directly.

This method will not cover failures in objects where the `taskEventLoop` is passed directly, but it's always been that way. I checked a couple of objects (e.g.: `TaskInfoFetcher`, `TaskStatusFetcher`) and they designed to propagate exceptions correctly.

arhimondr (commented):
Tricky bug. Didn't expect the implementation to be not exception safe. It used to be that some executors implementations in Java were not thread safe (e.g.: ScheduledExecutorService) but I remember it was fixed a while ago. Thank you for working on the fix.

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTaskFactory.java
shangm2: Nice catch!

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTaskFactory.java
shangm2: Good point! Thanks for pointing it out.

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTaskFactory.java
shangm2: Great point!

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTask.java
shangm2: Yeah, this is much cleaner. Thanks for the suggestion!

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTaskFactory.java
shangm2: The issue we observed is that when  something bad happen on the event loop, the thread for that particular event loop gets terminated. And further submitting a task to that event loop will trigger the RejectedExecutionException [here](https://github.com/prestodb/presto/blob/bd4c541233d4ab750d89e38246c1b688cfaa5115/presto-main/src/main/java/com/facebook/presto/execution/StateMachine.java#L313), which within netty comes from [here](https://github.com/netty/netty/blob/4.1/common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java#L352).

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTaskFactory.java
```diff
@@ -229,6 +243,39 @@ public RemoteTask createRemoteTask(
...
+    }
+
+    /***
+     *  One observation about event loop is if submitted task fails, it could kill the thread but the event loop group will not create a new one.
+     *  Here, we wrap it as safe event loop so that if any submitted job fails, we chose to log the error and fail the entire task.
+     */
+    static class SafeEventLoop
+            extends DefaultEventLoop
+    {
+        private Consumer<Throwable> onFail;
```
arhimondr, line 256: nit: remove this field

arhimondr (approved):

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTaskFactory.java
shangm2: Ooops, thanks a lot for catching this.

NikhilCollooru (approved):

PR #24414, opened 2025-01-22
Optimize how we merge multiple operatorStats

shangm2:
## Description
1. Thanks to @arhimondr who observed that GC could take too much cpu to clean up memory during heavy load
Screenshot 2025-01-21 at 21 55 30

2. This pr will be the first of a series of optimization to improve how objects are being created along the critical path.
3. This pr optimizes how we merge multiple operatorStats without creating temporary/intermediate objects

## Motivation and Context
1. The original code will create temporary objects every time we add two OperatorStats together (with same id) using v.add(operatorStats) and this intermediate object will be discarded when it is used to merge with next OperatorStats object, This PR groups all operatorStats by their id and then merge them together in one go.
2. Refactoring the code by moving local variables into a dedicated class so that we can easily use one loop within the _create_ method to aggregate all necessary metrics.

## Test Plan
1. local hiveQueryRunner works fine
2. Internal verifier tests passed

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

Optimizations
* Improve how we merge multiple operator stats together. :pr:`24414`
* Improve metrics creation by refactoring local variables to a dedicated class. :pr:`24414`


```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/execution/StageExecutionInfo.java
+++ b/presto-main/src/main/java/com/facebook/presto/execution/StageExecutionInfo.java
@@ -31,6 +31,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 
@@ -42,6 +43,7 @@ import static com.facebook.presto.common.RuntimeMetricName.TASK_SCHEDULED_TIME_N
 import static com.facebook.presto.common.RuntimeUnit.NANO;
 import static com.facebook.presto.common.RuntimeUnit.NONE;
 import static com.facebook.presto.execution.StageExecutionState.FINISHED;
+import static com.google.common.collect.ImmutableList.toImmutableList;
 import static io.airlift.units.DataSize.succinctBytes;
 import static io.airlift.units.Duration.succinctDuration;
 import static java.lang.Math.max;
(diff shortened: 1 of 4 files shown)


presto-main/src/main/java/com/facebook/presto/execution/StageExecutionInfo.java
```diff
@@ -71,186 +71,87 @@ public static StageExecutionInfo create(
...
-        boolean fullyBlocked = true;
-        Set<BlockedReason> blockedReasons = new HashSet<>();
-
-        Map<String, OperatorStats> operatorToStats = new HashMap<>();
-        RuntimeStats mergedRuntimeStats = new RuntimeStats();
-        mergedRuntimeStats.mergeWith(stageRuntimeStats);
-
-        List<TaskStats> allTaskStats = new ArrayList<>();
+        TaskStatsAggregator taskStatsAggregator = new TaskStatsAggregator();
+        taskStatsAggregator.increaseTotalTaskCount(taskInfos.size());
```
arhimondr, line 75: nit: this can be a constructor parameter

presto-main/src/main/java/com/facebook/presto/execution/StageExecutionInfo.java
```diff
@@ -71,186 +71,87 @@ public static StageExecutionInfo create(
...
+        taskStatsAggregator.mergeRuntimeStats(stageRuntimeStats);
```
arhimondr, line 76: nit: this can also be a constructor parameters

presto-main/src/main/java/com/facebook/presto/execution/StageExecutionInfo.java
```diff
@@ -309,4 +210,172 @@ public static StageExecutionInfo unscheduledExecutionInfo(int stageId, boolean i
...
+            // First collect all operator stats by their ID
+            Map<String, List<OperatorStats>> operatorStatsById = new HashMap<>();
+            for (PipelineStats pipeline : taskStats.getPipelines()) {
+                for (OperatorStats operatorStats : pipeline.getOperatorSummaries()) {
+                    String id = pipeline.getPipelineId() + "." + operatorStats.getOperatorId();
+                    operatorStatsById.computeIfAbsent(id, k -> new ArrayList<>()).add(operatorStats);
+                }
+            }
+            // Merge each group into a single result
+            operatorStatsById.forEach((id, statsToMerge) -> statsToMerge.stream().reduce(OperatorStats::add).ifPresent(merged -> operatorToStats.put(id, merged)));
```
arhimondr, line 318: Ideally you would want to merge once operator stats are collected from all tasks (e.g.: in `getOperatorToStatsById`)

Basically the idea is to get rid of [this method](https://github.com/prestodb/presto/blob/c2b22c097aa366ed446caa84c7f83a8ffad1bb96/presto-main/src/main/java/com/facebook/presto/operator/OperatorStats.java#L719) and always use [this one](https://github.com/prestodb/presto/blob/c2b22c097aa366ed446caa84c7f83a8ffad1bb96/presto-main/src/main/java/com/facebook/presto/operator/OperatorStats.java#L724)

presto-main/src/main/java/com/facebook/presto/execution/StageExecutionInfo.java
```diff
@@ -309,4 +210,172 @@ public static StageExecutionInfo unscheduledExecutionInfo(int stageId, boolean i
...
+        {
+            return mergedRuntimeStats;
+        }
+
+        public Map<String, OperatorStats> getOperatorToStatsById()
+        {
+            return operatorToStats;
+        }
+
+        public void increaseTotalTaskCount(int count)
```
arhimondr, line 346: nit: consider passing `TaksInfo` to `processTask(Stats)` instead of having the `increase` methods

presto-main/src/main/java/com/facebook/presto/execution/StageExecutionInfo.java
```diff
@@ -309,4 +210,172 @@ public static StageExecutionInfo unscheduledExecutionInfo(int stageId, boolean i
...
+            updateRuntimeStats(taskStats);
+        }
+
+        private void updateOperatorStats(TaskStats taskStats)
+        {
+            // First collect all operator stats by their ID
+            Map<String, List<OperatorStats>> operatorStatsById = new HashMap<>();
+            for (PipelineStats pipeline : taskStats.getPipelines()) {
+                for (OperatorStats operatorStats : pipeline.getOperatorSummaries()) {
+                    String id = pipeline.getPipelineId() + "." + operatorStats.getOperatorId();
```
arhimondr, line 313: Consider having a struct `OperatorKey` with two `int` fields with `hashCode` and `equals` instead of creating a string

presto-main/src/main/java/com/facebook/presto/execution/StageExecutionInfo.java
```diff
@@ -309,4 +210,172 @@ public static StageExecutionInfo unscheduledExecutionInfo(int stageId, boolean i
...
+        {
+            mergedRuntimeStats.mergeWith(runtimeStats);
+        }
+
+        public RuntimeStats getMergedRuntimeStats()
+        {
+            return mergedRuntimeStats;
+        }
+
+        public Map<String, OperatorStats> getOperatorToStatsById()
```
arhimondr, line 341: consider returning a `List<OperatorStats>` and calling this method `getOperatorStats`

presto-main/src/main/java/com/facebook/presto/execution/StageExecutionInfo.java
shangm2: Passing it in the constructor.

presto-main/src/main/java/com/facebook/presto/execution/StageExecutionInfo.java
shangm2: Awesome suggestion. Thanks a lot!

presto-main/src/main/java/com/facebook/presto/execution/StageExecutionInfo.java
shangm2: Yep, yep.

presto-main/src/main/java/com/facebook/presto/execution/StageExecutionInfo.java
```diff
@@ -309,4 +208,198 @@ public static StageExecutionInfo unscheduledExecutionInfo(int stageId, boolean i
...
+
+        public RuntimeStats getMergedRuntimeStats()
+        {
+            return mergedRuntimeStats;
+        }
+
+        public List<OperatorStats> getOperatorSummaries()
+        {
+            operatorStatsByKey.forEach((key, statsToMerge)
+                    -> operatorSummaries.add(statsToMerge.get(0).add(statsToMerge.subList(1, statsToMerge.size()))));
```
arhimondr, line 371: Expanding `operatorSummaries` makes this field non idempotent. Every subsequent call of `getOperatorSummaries` will return an expanded list. I recommend dropping the `operatorSummaries` field and using `toImmutableList` collector instead, for example:

```
return operatorStatsByKey.entries().stream()
...
.collect(toImmutableList())
```

arhimondr (commented):
Thank you for following up. Looks good to me % comments

presto-main/src/main/java/com/facebook/presto/execution/StageExecutionInfo.java
shangm2: Absolutely good point. Thanks for catching this!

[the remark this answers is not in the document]
shangm2: Got it. Was too afraid of breaking anything so was trying to make minimal changes. Will start being brave a bit in future. Thanks a lot for all the guidance!

presto-main/src/main/java/com/facebook/presto/operator/PipelineContext.java
```diff
@@ -201,9 +202,9 @@ public void driverFinished(DriverContext driverContext)
 
         // merge the operator stats into the operator summary
         List<OperatorStats> operators = driverStats.getOperatorStats();
-        for (OperatorStats operator : operators) {
-            operatorSummaries.compute(operator.getOperatorId(), (operatorId, summaryStats) -> summaryStats == null ? operator : summaryStats.add(operator));
-        }
+        operators.stream().collect(Collectors.groupingBy(OperatorStats::getOperatorId))
```
arhimondr, line 205: nit: static import `groupingBy`. Consider using the same approach with `merge` + `toImmutableList`

presto-main/src/main/java/com/facebook/presto/operator/PipelineContext.java
```diff
@@ -418,24 +419,9 @@ public PipelineStats getPipelineStats()
...
-            }
-            else {
-                combined = runningStats.get(0);
-                if (runningStats.size() > 1) {
-                    combined = combined.add(runningStats.subList(1, runningStats.size()));
-                }
-            }
-            operatorSummaries.put(operatorId, combined);
-        }
+        runningOperators.asMap().forEach((operatorId, runningStats) -> {
```
arhimondr, line 422: ditto

shangm2: Good point. Did not realize this part is rather similar to StageExecutionStats

presto-main/src/main/java/com/facebook/presto/operator/OperatorStats.java
```diff
@@ -716,67 +717,77 @@ public DataSize getIsBlockedAllocation()
...
-    public OperatorStats add(Iterable<OperatorStats> operators)
-    {
-        long totalDrivers = this.totalDrivers;
+        OperatorStats first = operators.stream().findFirst().get();
+        int stageId = first.getStageId();
+        int operatorId = first.getOperatorId();
+        int stageExecutionId = first.getStageExecutionId();
+        int pipelineId = first.getPipelineId();
+        PlanNodeId planNodeId = first.getPlanNodeId();
+        String operatorType = first.getOperatorType();
```
shangm2, line 732: @arhimondr one issue with "0" based initial value is we dont have a starting value for those ids but need to grab them from the first item of the collections passed in. This is safe since we check its emptiness right above. Let me know what you think.

presto-main/src/main/java/com/facebook/presto/operator/PipelineContext.java
```diff
@@ -378,9 +380,11 @@ public PipelineStats getPipelineStats()
         boolean hasUnfinishedDrivers = false;
         boolean unfinishedDriversFullyBlocked = true;
 
-        TreeMap<Integer, OperatorStats> operatorSummaries = new TreeMap<>(this.operatorSummaries);
-        ListMultimap<Integer, OperatorStats> runningOperators = ArrayListMultimap.create();
         ImmutableList.Builder<DriverStats> drivers = ImmutableList.builderWithExpectedSize(driverContexts.size());
+        // Make deep copy of each list
+        ConcurrentMap<Integer, List<OperatorStats>> operatorStatsById = this.operatorStatsById.entrySet().stream()
```
arhimondr, line 385: Does it have to be  `ConcurrentMap`? Have you considered a simple `Map` and the `toMap` collector?

presto-main/src/main/java/com/facebook/presto/operator/PipelineContext.java
```diff
@@ -378,9 +380,11 @@ public PipelineStats getPipelineStats()
...
+                .collect(toConcurrentMap(Map.Entry::getKey, e -> new ArrayList<>(Arrays.asList(e.getValue()))));
```
arhimondr, line 386: Do you need to have an extra `new ArrayList<>(...)`? Would `Arrays.asList(e.getValue())` alone work?

arhimondr (dismissed):
Very good. Thank you for the follow up.  

Looks great to me % two small questions

presto-main/src/main/java/com/facebook/presto/operator/PipelineContext.java
shangm2: Yeah, it needs to be new ArrayList<> since Arrays.asList(e.getValue()) will create a immutable list but we need it be mutable so we can add item to it later on

presto-main/src/main/java/com/facebook/presto/operator/PipelineContext.java
shangm2: You are absolutely correct. It does not need to be concurrentMap. Regular map will do the job.

presto-main/src/main/java/com/facebook/presto/operator/PipelineContext.java
arhimondr: The standard JDK `Arrays.asList` returns a mutable ArrayList: https://github.com/openjdk/jdk21/blob/master/src/java.base/share/classes/java/util/Arrays.java#L4222

Is this the JDK `Arrays.asList` used here? Or is there one in Guava?

arhimondr (approved):

presto-main/src/main/java/com/facebook/presto/operator/PipelineContext.java
shangm2: It is from standard JDK. It gives the following exception because the call "Returns a fixed-size list backed by the specified array." (so basically a view of the array) We can modify existing elements but can not resize it.
Screenshot 2025-01-28 at 12 49 15

arhimondr: I didn't know that. Thank you for the explanation

Issue #22422, opened 2024-04-04
TestIcebergParquetMetadataCaching.testParquetMetadataCaching looks flaky


elharo:
Likely flaky; need to rerun to be sure

```
2024-04-04T16:12:10.2637984Z [ERROR] Tests run: 2084, Failures: 1, Errors: 0, Skipped: 23, Time elapsed: 1,512.985 s <<< FAILURE! - in TestSuite
2024-04-04T16:12:10.2640602Z [ERROR] com.facebook.presto.iceberg.TestIcebergParquetMetadataCaching.testParquetMetadataCaching  Time elapsed: 0.401 s  <<< FAILURE!
2024-04-04T16:12:10.2642202Z java.lang.AssertionError: expected [true] but found [false]
2024-04-04T16:12:10.2642946Z 	at org.testng.Assert.fail(Assert.java:110)
2024-04-04T16:12:10.2643619Z 	at org.testng.Assert.failNotEquals(Assert.java:1413)
2024-04-04T16:12:10.2644327Z 	at org.testng.Assert.assertTrue(Assert.java:56)
2024-04-04T16:12:10.2645010Z 	at org.testng.Assert.assertTrue(Assert.java:66)
2024-04-04T16:12:10.2646669Z 	at com.facebook.presto.iceberg.TestIcebergParquetMetadataCaching.testParquetMetadataCaching(TestIcebergParquetMetadataCaching.java:90)
2024-04-04T16:12:10.2648290Z 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
2024-04-04T16:12:10.2649495Z 	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
2024-04-04T16:12:10.2650778Z 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
2024-04-04T16:12:10.2651817Z 	at java.lang.reflect.Method.invoke(Method.java:498)
2024-04-04T16:12:10.2652984Z 	at org.testng.internal.invokers.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:135)
2024-04-04T16:12:10.2654377Z 	at org.testng.internal.invokers.TestInvoker.invokeMethod(TestInvoker.java:673)
2024-04-04T16:12:10.2655619Z 	at org.testng.internal.invokers.TestInvoker.invokeTestMethod(TestInvoker.java:220)
[...]

PR #24415, opened 2025-01-22
Fix flaky Iceberg parquet metadata cache test

nmahadevuni:
## Description
Fixes flaky test described in https://github.com/prestodb/presto/issues/22422

## Motivation and Context
It seems jmx metrics may get mixed up if we are using the query runner and multiple workers. The jmx metrics are bound to coordinator cache object only and owing to the very tiny data read in the test query, if the coordinator doesn't process any split, it will just report zero metrics. So, making the number of nodes to only 1, so this problem is eliminated.

## Impact
No impact

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergParquetMetadataCaching.java
+++ b/presto-iceberg/src/test/java/com/facebook/presto/iceberg/TestIcebergParquetMetadataCaching.java
@@ -45,7 +45,7 @@ public class TestIcebergParquetMetadataCaching
                 PARQUET,
                 false,
                 true,
-                OptionalInt.of(2),
+                OptionalInt.of(1),
                 Optional.empty());
     }
 


tdcmeehan (approved):

hantangwangd:
@nmahadevuni Sorry for the late response.  I took some time to figure out why coordinator and workers in the query runner will impact each other's jmx metrics (it seems that one will override the other). The core reason is that, all nodes in the query runner have the same MBeanServer instance and then registered their own parquet metadata `cacheStatsMBean` into this singleton MBeanServer.

Set the number of nodes to only 1 can definitely fix this flaky test. But if we want to completely solve the problem, we need to inject separate MBeanServers for different coordinators and workers, which may involve a small refactoring of `IcebergPlugin` and `IcebergConnectorFactory`. So, do you think it's necessary to do this refactoring to thoroughly address the mutual influence of jmx metrics between different nodes in a query runner? @tdcmeehan

nmahadevuni:
Thanks @hantangwangd. Yes I think that helps us get the real world setup in the query runner too. Will help a lot with metrics integration tests.

PR #24417, opened 2025-01-23
[Iceberg] Add table and session property for split size

### LINKED ISSUE #24419: Make iceberg table target split size configurable
We should have a way to set a table's target split size for the Iceberg connector in Presto. The current default is 128MB and in some cases generates too much overhead. We should be able to set the target split size in order to tune the performance of some or all queries.

We should be able to set the target split size either via table properties or session properties.

Iceberg connector.

## Possible Implementation

- Add session property to configure split size for all queries in a session (overrides table property)
- Add support for target split size in `ALTER TABLE ... SET PROPERTIES` statement

## Context

During table scan planning within the Iceberg connector, we generate splits based on the iceberg table's [`read.split.target-size`](https://iceberg.apache.org/docs/latest/configuration/) property. For while there was no easy way to set table properties. However #21495 introduced this feature, but still doesn't allow setting arbitrary properties unless we update the set of supported table properties in `IcebergAbstractMetadata#setTableProperties`. 

We should introduce support for setting the table property size to allow tuning this parameter on Iceberg tables.

ZacBlanco:
## Description

Adds a session property for target split size.

Closes #24419 

## Motivation and Context

Makes it easier to do performance debugging by setting the desired split size on a per-query basis.

## Impact

New configuration property.

## Test Plan

unit tests

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

Iceberg Connector Changes
* Add ``target_split_size`` session property
* Add ``read.split.target-size`` table property
```

### DIFF
--- a/presto-docs/src/main/sphinx/connector/iceberg.rst
+++ b/presto-docs/src/main/sphinx/connector/iceberg.rst
@@ -357,9 +357,9 @@ connector using a WITH clause:
 
 The following table properties are available, which are specific to the Presto Iceberg connector:
 
-=======================================   ===============================================================   ============
+=======================================   ===============================================================   =========================
 Property Name                             Description                                                       Default
-=======================================   ===============================================================   ============
+=======================================   ===============================================================   =========================
 ``format``                                 Optionally specifies the format of table data files,             ``PARQUET``
                                            either ``PARQUET`` or ``ORC``.
 
@@ -388,7 +388,11 @@ Property Name                             Description                           
 
 ``metrics_max_inferred_column``            Optionally specifies the maximum number of columns for which     ``100``
                                            metrics are collected.
-=======================================   ===============================================================   ============
+
+``read.split.target-size``                 The target size for an individual split when generating splits     ``134217728`` (128MB)
+                                           for a table scan. Generated splits may still be larger or
+                                           smaller than this value. Must be specified in bytes.
+=======================================   ===============================================================   =========================
 
 The table definition below specifies format ``ORC``, partitioning by columns ``c1`` and ``c2``,
 and a file system location of ``s3://test_bucket/test_schema/test_table``:
@@ -421,6 +425,9 @@ Property Name                                         Description
 ``iceberg.rows_for_metadata_optimization_threshold``  Overrides the behavior of the connector property
                                                       ``iceberg.rows-for-metadata-optimization-threshold`` in the current
                                                       session.
+``iceberg.target_split_size``                         Overrides the target split size for all tables in a query in bytes.
+                                                      Set to 0 to use the value in each Iceberg table's
+                                                      ``read.split.target-size`` property.
 ===================================================== ======================================================================
 
 Caching Support
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
@@ -201,6 +201,7 @@ import static org.apache.iceberg.SnapshotSummary.REMOVED_EQ_DELETES_PROP;
 import static org.apache.iceberg.SnapshotSummary.REMOVED_POS_DELETES_PROP;
 import static org.apache.iceberg.TableProperties.DELETE_ISOLATION_LEVEL;
 import static org.apache.iceberg.TableProperties.DELETE_ISOLATION_LEVEL_DEFAULT;
+import static org.apache.iceberg.TableProperties.SPLIT_SIZE;
 import static org.apache.iceberg.TableProperties.UPDATE_MODE;
 
 public abstract class IcebergAbstractMetadata
@@ -719,6 +720,7 @@ public abstract class IcebergAbstractMetadata
         properties.put(METADATA_PREVIOUS_VERSIONS_MAX, IcebergUtil.getMetadataPreviousVersionsMax(icebergTable));
         properties.put(METADATA_DELETE_AFTER_COMMIT, IcebergUtil.isMetadataDeleteAfterCommit(icebergTable));
         properties.put(METRICS_MAX_INFERRED_COLUMN, IcebergUtil.getMetricsMaxInferredColumn(icebergTable));
+        properties.put(SPLIT_SIZE, IcebergUtil.getSplitSize(icebergTable));
 
         SortOrder sortOrder = icebergTable.sortOrder();
         // TODO: Support sort column transforms (https://github.com/prestodb/presto/issues/24250)
@@ -1126,6 +1128,9 @@ public abstract class IcebergAbstractMetadata
             switch (entry.getKey()) {
                 case COMMIT_RETRIES:
                     updateProperties.set(TableProperties.COMMIT_NUM_RETRIES, String.valueOf(entry.getValue()));
+                    break;
+                case SPLIT_SIZE:
+                    updateProperties.set(TableProperties.SPLIT_SIZE, entry.getValue().toString());
                     break;
                 default:
                     throw new PrestoException(NOT_SUPPORTED, "Updating property " + entry.getKey() + " is not supported currently");
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergSplitSource.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergSplitSource.java
@@ -20,13 +20,13 @@ import com.facebook.presto.spi.ConnectorSplit;
 import com.facebook.presto.spi.ConnectorSplitSource;
 import com.facebook.presto.spi.SplitWeight;
 import com.facebook.presto.spi.connector.ConnectorPartitionHandle;
+import com.facebook.presto.spi.schedule.NodeSelectionStrategy;
 import com.google.common.collect.ImmutableList;
 import com.google.common.io.Closer;
 import org.apache.iceberg.FileScanTask;
 import org.apache.iceberg.PartitionSpec;
 import org.apache.iceberg.PartitionSpecParser;
 import org.apache.iceberg.TableScan;
-import org.apache.iceberg.io.CloseableIterable;
 import org.apache.iceberg.io.CloseableIterator;
 
 import java.io.IOException;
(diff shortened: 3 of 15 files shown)


steveburnett:
Thanks for the release note! Nit formatting suggestion. 

```
== RELEASE NOTES ==

Iceberg Connector Changes
* Add ``target_split_size session`` property. :pr:`24417`
```

Consider documenting the new session property, perhaps in 
https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/connector/iceberg.rst#session-properties .

steveburnett (approved):
LGTM! (docs) 

hantangwangd (commented):
Bring one thing for discussion. From a high level perspective, considering that a complex sql (like tpcds sqls) may involve multiple tables, is it a suitable way to set a unified session level target split size to override their own split size? Or should we respect each table's own split size (if exist) more than the session level uniform value?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
```diff
@@ -1007,6 +1009,9 @@ public void setTableProperties(ConnectorSession session, ConnectorTableHandle ta
                 case COMMIT_RETRIES:
                     updateProperties.set(TableProperties.COMMIT_NUM_RETRIES, String.valueOf(entry.getValue()));
                     break;
+                case SPLIT_SIZE:
+                    updateProperties.set(TableProperties.SPLIT_SIZE, entry.getValue().toString());
```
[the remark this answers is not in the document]
ZacBlanco: Good question. My preference is to let it error when a query is executed. It should be up to the iceberg library to determine when an error is thrown. In the future they could assign a special value to 0 and we would have to maintain our checks to keep in sync

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -388,7 +388,10 @@ Property Name                             Description
 
 ``metrics_max_inferred_column``            Optionally specifies the maximum number of columns for which     ``100``
                                            metrics are collected.
-=======================================   ===============================================================   ============
+
+``read.split.target-size``                 The target size for an individual split when generating splits     ``134217728`` (128MB)
```
yingsu00, line 392: It seems Presto convention does not use the prefixes from the Iceberg lib. E.g. "write.metadata.metrics.max-inferred-column-defaults" was just named "metrics_max_inferred_column" table property. While I do think the notion with prefixes is clearer, I think it's better to name it "split_target_size" for now. Please also add explanation here this correspond to `read.split.target-size` table property in Iceberg library. 

Later on, we can send a proposal to use full iceberg property names since it will affect the users.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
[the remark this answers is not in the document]
yingsu00: I think we should check it when setting the table property or session property

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergSplitSource.java
```diff
@@ -131,7 +132,7 @@ private ConnectorSplit toIcebergSplit(FileScanTask task)
                 PartitionSpecParser.toJson(spec),
                 partitionData.map(PartitionData::toJson),
                 getNodeSelectionStrategy(session),
-                SplitWeight.fromProportion(Math.min(Math.max((double) task.length() / tableScan.targetSplitSize(), minimumAssignedSplitWeight), 1.0)),
+                SplitWeight.fromProportion(Math.min(Math.max((double) task.length() / getTargetSplitSize(session, tableScan).toBytes(), minimumAssignedSplitWeight), 1.0)),
```
yingsu00, line 135: I don't think this would enforce the `iceberg.target_split_size` session property. This line only sets the weight of the split, but not the actual split size. Your split size is still `task.length(),` on line 128. To really enforce it and make it overwrite the table property , we need to adopt similar logic in HiveSplitSource.getNextBatch(). It may even be complex, because `iceberg.target_split_size`  could be larger than `read.split.target-size`. If we don't want that to happen, we will need to state it clearly in the document and do propery checking when setting these values.

yingsu00 (changes requested):

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -421,6 +424,9 @@ Property Name                                         Description
 ``iceberg.rows_for_metadata_optimization_threshold``  Overrides the behavior of the connector property
                                                       ``iceberg.rows-for-metadata-optimization-threshold`` in the current
                                                       session.
+``iceberg.target_split_size``                         Overrides the target split size for all tables in a query in bytes.
```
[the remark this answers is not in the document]
ZacBlanco: Iceberg and Hive function differently. Splits are not calculated in the same way. As such, the target size property does not refer to the absolute maximum size of the split. `FixedSizeSplitScanTaskIterator` only applies to files which don't support offsets. Parquet and ORC do support offsets, so they most likely won't use that class, but instead the `OffsetsWareSplitScanTaskIterator` from the iceberg library.

Either way, these are implementation details. If the property meant "max split size" it would be documented as such. This property does not represent the max split size and does not guarantee that splits will be below this size

presto-docs/src/main/sphinx/connector/iceberg.rst
ZacBlanco: @hantangwangd and I have discussed this in another PR and decided that moving forward for new properties we will use the iceberg property names. We will introduce backwards-compatible property names for the properties which were already introduced and slowly phase the old names out over the next few releases. I have filed this issue to address it: https://github.com/prestodb/presto/issues/24483

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergSplitSource.java
ZacBlanco: Whether or not this calculation is correct can be another discussion. I am just matching the previous behavior for the case when the user wants to use this session property. Previously, we would use the table scan's target split size property when calculating the split weight. Now instead, I am calling the `getTargetSplitSize` function which will get the session property if the user has set it, or defer to the table scan's split size if not. When deferring to the table scan's split size, the logic remains the same as before this PR.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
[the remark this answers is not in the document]
hantangwangd: FYI @ZacBlanco @yingsu00, the behavior of Spark on Iceberg is that, it do not check when setting the table property, but it will adjust the split size when scanning if the table's `read.split.adaptive-size.enabled` is true (the default value is true). And the adjustment will handle negative values properly. So it will not fail when executing query after setting the table's `read.split.metadata-target-size` to a negative value, unless explicitly set `read.split.adaptive-size.enabled` to false.

Should we be consistent with Spark's behavior?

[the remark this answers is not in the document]
hantangwangd: Supplement: from the discussions in [here](https://github.com/apache/iceberg/pull/12201#discussion_r1947861228), I guess that the Spark tends to throw an exception during the execution in this situation. Adjusting negative numbers to normal values is just an unexpected behavior.

presto-docs/src/main/sphinx/connector/iceberg.rst
yingsu00: I agree that using Iceberg names is clearer. But ideally we should make changes for #24483 before directly using the new name. When and who will be working on it? Do you think you can send PR for that sooner?

presto-docs/src/main/sphinx/connector/iceberg.rst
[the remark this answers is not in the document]
yingsu00: Ok. In this case, can we add some clarification here: "Unlike hive.max_split_size, this can be smaller or greater than the actual split size"

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergSplitSource.java
yingsu00: The problem is, in the documentation you clearly stated that the `iceberg.target_split_size` property "Overrides the target split size for all tables in a query in bytes.". The code is not the same as this statement. You added a new property and defined its behavior but the code does not implement it. This line `SplitWeight.fromProportion(Math.min(Math.max((double) task.length() / getTargetSplitSize(session, tableScan).toBytes(), minimumAssignedSplitWeight), 1.0)),
` only changes the weight of the split.

ZacBlanco: The split size is used in the IcebergSplitManager used as a parameter when passed to `TableScanUtil.splitFiles` and to the ChangelogSplitSource here:

 https://github.com/prestodb/presto/blob/04f8aaae893dd451b3a1af3c6b8ab96643d77f1a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergSplitManager.java#L121

and here:

https://github.com/prestodb/presto/blob/04f8aaae893dd451b3a1af3c6b8ab96643d77f1a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergSplitManager.java#L99

This should take the session property into account when calculating splits, no?

presto-docs/src/main/sphinx/connector/iceberg.rst
ZacBlanco: Sure, I can try to get that PR out this week. I can definitely have it in soon, but since most users won't see any changes until the 292 release, I am in the camp that it would be fine to not hold up this PR to align the old property names

yingsu00: Sounds good! It should be good as long as it's before the 292 release.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergSplitSource.java
yingsu00: Oh I see, you passed it to Iceberg library, so yes it should be correct then. But it is not quite straightforward because the logic is implemented at a different place (IcebergSplitManager) other than the place where the splits are created (IcebergSplitSource.toIcebergSplit(FileScanTask task)). It would be better to encapsulate the target size logic in side of IcebergSplitSource itself. Also, We don't need to compute the target size for every split in IcebergSplitSource.toIcebergSplit(FileScanTask task). How about this:

Since the only call site of IcebergSplitSource constructor is on line 118 in SplitManager, we can assign its fileScanTaskIterable member directly inside of IcebergSplitSource's constructor and remove the `CloseableIterable<FileScanTask> fileScanTaskIterable,` parameter from the constructor. We can make `targetSize` as a data member of `IcebergSplitSource` and assign its value while constructing the IcebergSplitSource:
```
IcebergSplitSource splitSource = new IcebergSplitSource(
                    session,
                    tableScan,
                    getTargetSplitSize(session, tableScan).toBytes(),
                    getMinimumAssignedSplitWeight(session),
                    getMetadataColumnConstraints(layoutHandle.getValidPredicate()));
```
In IcebergSplitSource cstr:
```
public IcebergSplitSource(
            ConnectorSession session,
            TableScan tableScan,
            DataSize targetSplitSize,
            double minimumAssignedSplitWeight,
            TupleDomain<IcebergColumnHandle> metadataColumnConstraints)
    {
        this.session = requireNonNull(session, "session is null");
        this.tableScan = requireNonNull(tableScan, "tableScan is null");
        this.targetSplitSize = targetSplitSize;
        this.fileScanTaskIterator = TableScanUtil.splitFiles(tableScan.planFiles(), targetSplitSize).iterator();
        this.minimumAssignedSplitWeight = minimumAssignedSplitWeight;
        this.metadataColumnConstraints = requireNonNull(metadataColumnConstraints, "metadataColumnConstraints is null");
        closer.register(fileScanTaskIterator);
    }
this.fileScanTaskIterator = fileScanTaskIterable.iterator();
```
Then this line becomes
```SplitWeight.fromProportion(Math.min(Math.max((double) task.length() / targetSplitSize, minimumAssignedSplitWeight), 1.0)),
                
```
I think this will be easier to understand.

steveburnett:

```

:pr:`12345`

```

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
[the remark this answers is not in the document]
ZacBlanco: I think it makes sense to match spark's behavior

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergSplitSource.java
ZacBlanco: I agree with this. The PR is updated with this approach.

presto-docs/src/main/sphinx/connector/iceberg.rst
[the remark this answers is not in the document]
ZacBlanco: I've added this to the description for the table property. The session property refers to the table property, so I think it should be clear enough

presto-docs/src/main/sphinx/connector/iceberg.rst
ZacBlanco: Here is a draft PR for introducing the deprecation of table property names: https://github.com/prestodb/presto/pull/24581

It is still WIP. Just needs some tests

yingsu00 (approved):

yingsu00 (commented):
@ZacBlanco @steveburnett Shall we also mention we're adding a Iceberg table property read.split.target-size?

presto-docs/src/main/sphinx/connector/iceberg.rst
yingsu00: > Here is a draft PR for introducing the deprecation of table property names: #24581
> 
> It is still WIP. Just needs some tests

Great! Ping me when it's ready.

hantangwangd (approved):

steveburnett:
> @ZacBlanco @steveburnett Shall we also mention we're adding a Iceberg table property read.split.target-size?

Thanks for the catch @yingsu00! Yes, doc should be added for all new properties introduced in a PR.

ZacBlanco:
@yingsu00 there is already a new entry in the doc for `read.split.target-size` in the table properties. I will add it to the release note though.

PR #24430, opened 2025-01-24
[native] Convert few more Velox operator names to Presto.

spershin:
## Description
In Prestissimo we convert some Velox operator names to Presto operator names to leverage the logic Presto has when processing query/operator stats.
Here we add few more conversions as we've noticed that the difference in naming prevents some tools from operating properly.

## Test Plan
Existing unit test.
Ran a query with the new binary to observe the feature that didn't work before now works.

```
== NO RELEASE NOTE ==
```


amitkdutta (approved):
Looks good. Thanks @spershin

PR #24449, opened 2025-01-28
Add health check to route queries to healthy cluster and router jmx counter updates

auden-woolfson:
## Description
Add coodinator health checks to presto router to ensure queries are sent to active/healthy clusters. Part of presto router forward fit. Also includes code implemented to patch CVEs.

```
== RELEASE NOTES ==

General Changes
* Add coordinator health checks to Presto router.
* Add counter JMX metrics to Presto router.
```


auden-woolfson:
Will fix CLA and squash commits upon review

saravanan19:
@Auden Woolfson - do we plan on including counter jmx metrics as part of this PR? is that intentional? If so, we need the PR title and description updated. If not, need a separate PR.

presto-router/src/test/java/com/facebook/presto/router/TestHealthChecks.java
```diff
@@ -0,0 +1,125 @@
...
+    {
+        for (TestingPrestoServer prestoServer : prestoServers) {
+            prestoServer.close();
+        }
+    }
+
+    @Test(enabled = false)
+    public void testHealthChecks()
+    {
+        prestoServers.get(0).stopResponding();
```
aaneja, line 109: Can you add a test where a server becomes unresponsive (i.e unhealthy), is removed from rotation, and then becomes responsive again and is added back to the rotation ?

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterManager.java
```diff
@@ -24,41 +25,108 @@
...
+        allClusters.forEach(uri -> {
+            log.info("Attaching cluster %s to the router", uri.getHost());
+            remoteClusterInfos.put(uri, remoteInfoFactory.createRemoteClusterInfo(discoveryURIs.get(uri)));
+            remoteQueryInfos.put(uri, remoteInfoFactory.createRemoteQueryInfo(discoveryURIs.get(uri)));
+            log.info("Successfully attached cluster %s to the router. Queries will be routed to cluster after successful health check", uri.getHost());
+        });
+    }
+
+    @PostConstruct
+    public void startConfigReloadTask()
```
aaneja, line 94: Can you add a test for this scenario - the file config gets updated, old servers are removed. New ones get added, existing ones stay as-is

aaneja (changes requested):
Overall approach looks good. Please squash and split into logically cohesive commits

presto-router/src/main/java/com/facebook/presto/router/cluster/RemoteState.java
```diff
@@ -67,10 +74,15 @@ public synchronized void asyncRefresh()
         if (nanosSince(lastWarningLogged.get()).toMillis() > 1_000 &&
                 sinceUpdate.toMillis() > 10_000 &&
                 future.get() != null) {
-            log.warn("Coordinator update request to %s has not returned in %s", remoteUri, sinceUpdate.toString(SECONDS));
+            log.warn("Coordinator update request to %s has not returned in %s", remoteUri.getHost(), sinceUpdate.toString(SECONDS));
             lastWarningLogged.set(System.nanoTime());
         }
 
+        if (nanosSince(lastHealthyResponseTime).toMillis() >= (secondsToUnhealthy * 1_000) && isHealthy) {
```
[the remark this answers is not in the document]
auden-woolfson: Sure, but the entire line becomes...

```
if (Duration.between(lastHealthyResponseTime, Instant.now()).compareTo(Duration.of(secondsToUnhealthy, ChronoUnit.SECONDS)) >= 1 && isHealthy) {
```

This is less concise and less readable in my opinion. I will default to you if you think this would be better, but I don't think it's necessary...

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterStatusTracker.java
```diff
@@ -78,63 +64,60 @@ public void startPollingQueryInfo()
...
-        remoteQueryInfos.keySet().removeAll(inactiveClusters);
-
-        allClusters.forEach(uri -> {
-            remoteClusterInfos.putIfAbsent(uri, remoteInfoFactory.createRemoteClusterInfo(uri));
-            remoteQueryInfos.putIfAbsent(uri, remoteInfoFactory.createRemoteQueryInfo(uri));
-        });
-
-        remoteClusterInfos.values().forEach(RemoteClusterInfo::asyncRefresh);
-        remoteQueryInfos.values().forEach(RemoteQueryInfo::asyncRefresh);
+        clusterManager.getRemoteClusterInfos().values().forEach(RemoteClusterInfo::asyncRefresh);
```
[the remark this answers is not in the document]
auden-woolfson: Make `ClusterManager` and internal class of `ClusterStatusTracker`?

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterManager.java
```diff
@@ -77,11 +145,20 @@ public Optional<URI> getDestination(RequestInfo requestInfo)
...
+                .filter(entry -> remoteClusterInfos.get(entry).isHealthy())
+                .collect(Collectors.toList());
+
+        if (healthyClusterURIs.isEmpty()) {
+            log.info("Healthy cluster not found!");
+            return Optional.empty();
         }
 
+        scheduler.setCandidates(healthyClusterURIs);
+        if (schedulerType == WEIGHTED_RANDOM_CHOICE || schedulerType == WEIGHTED_ROUND_ROBIN) {
```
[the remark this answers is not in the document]
auden-woolfson: @jp-sivaprasad, can you please add this to https://github.com/prestodb/presto/pull/24580?

presto-router/pom.xml
```diff
@@ -17,6 +17,11 @@
     </properties>
 
     <dependencies>
+        <dependency>
+            <groupId>com.facebook.presto</groupId>
```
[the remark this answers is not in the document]
auden-woolfson: We may be able to make this a test only dep for this PR, but there are other parts of router that we are forward fitting that will rely on this dependency in the main module. It might be beneficial to just leave this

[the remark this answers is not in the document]
aaneja: Not a fan of pulling in a huge dependency like presto-main into presto-router. What other PRs are bringing this in ? Let's see if we can avoid this by building good mocks for the Presto server. I think we may get by just having a few API endpoint's mocked

presto-router/src/main/java/com/facebook/presto/router/cluster/RemoteState.java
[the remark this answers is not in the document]
aaneja: Instead of `secondsToUnhealthy`, lets get an actual duration from the config (see comment on `RemoteStateConfig`). This will make it slightly easier to read.  
My issue is mostly with 1/ Forcing the user to specify actual seconds - with Duration they can use a config value like `750ms` e.g 2/Having to do time math which is better left to the std lib

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterStatusTracker.java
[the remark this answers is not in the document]
aaneja: The other way around. Make `ClusterStatusTracker` an internal class of `ClusterManager`

presto-router/pom.xml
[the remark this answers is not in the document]
auden-woolfson: I believe the authentication piece uses this. Other might as well. We can switch this for this PR but I can't guarantee that we will be able to leave it like this. We can cross that bridge when we get to it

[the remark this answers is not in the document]
auden-woolfson: Actually on second thought we might still need this. We are binding multiple classes (`ServerConfig`, `WebUiResource`, `PluginManagerConfig`) in the `RouterModule` from presto main.

auden-woolfson:
Ready for re-review @aaneja, also working on making the new test more efficient on another branch as suggested by @ZacBlanco.

presto-router/pom.xml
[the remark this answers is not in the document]
aaneja: Left comments in the RouterModule about these three classes. Let's see if we really need them.

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterManager.java
```diff
@@ -105,4 +218,126 @@ private void initializeServerWeights()
...
+    }
+
+    public void refreshHealthStatuses()
+    {
+        List<RemoteState> remoteStates = new ArrayList<>(remoteClusterInfos.values());
+        remoteStates.addAll(remoteQueryInfos.values());
+        remoteStates.forEach(RemoteState::asyncRefresh);
+    }
+
+    public ConcurrentHashMap<URI, RemoteClusterInfo> getRemoteClusterInfos()
```
aaneja, line 240: `getRemoteClusterInfos` and `getRemoteQueryInfos` can be made internal now (used only withing this class and `TestHealthChecks`). Also annotate these with `@VisibleForTesting`

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterManager.java
```diff
@@ -21,44 +22,150 @@
...
-    private final HashMap<String, HashMap<URI, Integer>> serverWeights = new HashMap<>();
+    private Map<String, GroupSpec> groups;
+    private List<SelectorRuleSpec> groupSelectors;
+    private SchedulerType schedulerType;
+    private Scheduler scheduler;
+    private HashMap<String, HashMap<URI, Integer>> serverWeights = new HashMap<>();
+    private HashMap<URI, URI> discoveryURIs = new HashMap<>();
+    public final RouterConfig routerConfig;
+    private final ScheduledExecutorService scheduledExecutorService;
+    private final AtomicLong lastConfigUpdate = new AtomicLong();
```
aaneja, line 77: Unused, can remove

presto-router/src/main/java/com/facebook/presto/router/cluster/RemoteStateConfig.java
```diff
@@ -0,0 +1,65 @@
...
+import com.facebook.airlift.configuration.Config;
+import com.facebook.airlift.configuration.ConfigDescription;
+
+import javax.validation.constraints.NotNull;
+
+import java.time.Duration;
+
+public class RemoteStateConfig
+{
+    private Duration clusterUnhealthyTimeout;
```
aaneja, line 25: Let's define same defaults for both of these here (and not in the classes using them)

aaneja (changes requested):

presto-router/src/test/java/com/facebook/presto/router/TestClusterManager.java
```diff
@@ -127,6 +136,37 @@ public void testQuery()
...
+
+        String originalConfigContent = new String(Files.readAllBytes(configFilePath));
+        String modifiedConfigContent = originalConfigContent.replaceAll("\"members\"\\s*:\\s*\\[.*?\\]", "\"members\": []");
+
+        FileOutputStream fos = new FileOutputStream(configFile, false);
+        fos.write(modifiedConfigContent.getBytes());
+        fos.flush();
+        fos.getFD().sync();
+        fos.close();
+        Thread.sleep(3000);
```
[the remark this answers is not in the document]
auden-woolfson: This is the maximum frequency at which the file watcher can be configured to poll for events in the config directory. Working on using a `CyclicBarrier` to sync these, but this requires a few significant structural changes that I will introduce in a future PR.

aaneja:
@auden-woolfson Please squash commits and ping me when it's ready for re-review

auden-woolfson:
This PR is ready for the next round of reviews

presto-router/src/test/java/com/facebook/presto/router/TestClusterManager.java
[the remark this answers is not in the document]
auden-woolfson: @aaneja, actually I was able to get this done and add it to this PR, so we are using a cyclic barrier now here

aaneja (changes requested):

steveburnett:
Thanks for the release note! A couple of nits:

```
== RELEASE NOTES ==

General Changes
* Add coordinator health checks to Presto router.
* Add counter JMX metrics to Presto router.
```

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterManager.java
```diff
@@ -13,81 +13,205 @@
...
+        ClusterManagerConfig config = currentConfig.get();
         Optional<String> target = matchGroup(requestInfo);
         if (!target.isPresent()) {
             return Optional.empty();
         }
 
-        checkArgument(groups.containsKey(target.get()));
-        GroupSpec groupSpec = groups.get(target.get());
-        scheduler.setCandidates(groupSpec.getMembers());
-        if (schedulerType == WEIGHTED_RANDOM_CHOICE) {
```
aaneja, line 81: The weights are no longer being set before making the routing decision ?

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterManager.java
```diff
@@ -13,81 +13,205 @@
...
     {
+        this.routerConfig = config;
         RouterSpec routerSpec = parseRouterConfig(config)
                 .orElseThrow(() -> new PrestoException(CONFIGURATION_INVALID, "Failed to load router config"));
+        Map<String, GroupSpec> groups = ImmutableMap.copyOf(routerSpec.getGroups().stream().collect(toMap(GroupSpec::getName, group -> group)));
+        List<SelectorRuleSpec> groupSelectors = ImmutableList.copyOf(routerSpec.getSelectors());
+        Scheduler scheduler = new SchedulerFactory(routerSpec.getSchedulerType()).create();
+        this.currentConfig = new AtomicReference<>(new ClusterManagerConfig(groups, groupSelectors, scheduler));
+        this.remoteInfoFactory = requireNonNull(remoteInfoFactory, "remoteInfoFactory is null");
+        this.initializeServerWeights();
```
aaneja, line 94: Instead of duplicating this call, just call `onConfigChangeDetection.apply()` after defining it ?

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterManager.java
```diff
@@ -13,81 +13,205 @@
...
+            List<URI> updatedAllClusters = getAllClusters();
 
-        this.groups = ImmutableMap.copyOf(routerSpec.getGroups().stream().collect(toMap(GroupSpec::getName, group -> group)));
-        this.groupSelectors = ImmutableList.copyOf(routerSpec.getSelectors());
-        this.schedulerType = routerSpec.getSchedulerType();
-        this.scheduler = new SchedulerFactory(routerSpec.getSchedulerType()).create();
+            updatedAllClusters.forEach(uri -> {
+                if (!remoteClusterInfos.containsKey(uri)) {
+                    log.info("Attaching cluster %s to the router", uri.getHost());
+                    remoteClusterInfos.put(uri, this.remoteInfoFactory.createRemoteClusterInfo(discoveryURIs.get(uri)));
```
aaneja, line 118: The creating and use of `discoveryURIs` is limited to just this `onConfigChangeDetection` method. There is no need to store it as a member of the ClusterManager class

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterManager.java
```diff
@@ -105,4 +229,159 @@ private void initializeServerWeights()
...
+                    builder.addAll(remoteQueryInfo.getQueryList().orElse(ImmutableList.of()).stream()
+                            .map(queryInfo -> ((ObjectNode) queryInfo).put("coordinatorUri", coordinator.toASCIIString()))
+                            .collect(toImmutableList())));
+            return builder.build();
+        }
+    }
+
+    private static class ClusterManagerConfig
+    {
+        private final Map<String, GroupSpec> groups;
```
aaneja, line 361: I think `serverWeights` should me a config as well

presto-router/src/test/java/com/facebook/presto/router/TestHealthChecks.java
```diff
@@ -0,0 +1,169 @@
...
+
+    @Test
+    public void testHealthChecks()
+            throws InterruptedException
+    {
+        TestingPrestoServer server0 = prestoServers.get(0);
+        TestingPrestoServer server1 = prestoServers.get(1);
+        TestingPrestoServer server2 = prestoServers.get(2);
+
+        List<URI> destinations = getDestinations(3);
```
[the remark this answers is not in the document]
aaneja: Can you add this comment ?

aaneja (changes requested):

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterManager.java
auden-woolfson: Just added this back

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterManager.java
```diff
@@ -21,88 +22,372 @@
...
+        private final ScheduledExecutorService queryInfoUpdateExecutor;
+
+        @Inject
+        public ClusterStatusTracker(ClusterManager clusterManager)
+        {
+            this.clusterManager = requireNonNull(clusterManager, "clusterManager is null");
+            this.queryInfoUpdateExecutor = newSingleThreadScheduledExecutor(threadsNamed("query-info-poller-%s"));
+        }
+
+        @PostConstruct
```
aaneja, line 257: I've read [mixed opinions](https://github.com/google/guice/issues/62) about how Guice will handle `PostConstruct` annotated methods. Can you just remove this annotation and simply call `startPollingQueryInfo()` from the constructor instead ? This removes any ambiguity around lifecycle management

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterManager.java
```diff
@@ -21,88 +22,372 @@
...
+            queryInfoUpdateExecutor.scheduleWithFixedDelay(() -> {
+                try {
+                    pollQueryInfos();
+                }
+                catch (Exception e) {
+                    log.error(e, "Error polling list of queries");
+                }
+            }, pollingInterval.toMillis(), pollingInterval.toMillis(), TimeUnit.MILLISECONDS);
+
+            pollQueryInfos();
```
aaneja, line 269: Move this to before L269, so that we do one round of `pollQueryInfos` _before_ we set up a periodic task ?

aaneja (changes requested):

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterManager.java
```diff
@@ -21,88 +22,372 @@
...
-    private final Scheduler scheduler;
-    private final HashMap<String, HashMap<URI, Integer>> serverWeights = new HashMap<>();
+    private final AtomicReference<ClusterManagerConfig> currentConfig = new AtomicReference<>();
+
+    public final RouterConfig routerConfig;
+    private final RemoteInfoFactory remoteInfoFactory;
+    private final Logger log = Logger.get(ClusterManager.class);
+
+    // Cluster status
+    private static Duration pollingInterval;
```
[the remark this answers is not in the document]
auden-woolfson: It's better to keep this static since ClusterStatusTracker uses it

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterManager.java
auden-woolfson: Adding a call to `pollQueryInfos` in the constructor

auden-woolfson:
Handled all suggestions @aaneja

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterManager.java
aaneja: Okay, so this can be removed then ?

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterManager.java
```diff
@@ -21,88 +22,376 @@
...
+                log.warn("Cluster [%s] is not available in remoteClusterInfos", entry);
+                return false;
+            }
+        }).collect(Collectors.toList());
+
+        if (healthyClusterURIs.isEmpty()) {
+            log.info("No healthy clusters found, enabled routing to unhealthy clusters");
+            healthyClusterURIs = groupSpec.getMembers();
         }
+        log.info("Available clusters: %s", healthyClusterURIs);
```
aaneja, line 199: This will spam the logs since we will log available cluster every request. Let's switch this to `log.debug`

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterManager.java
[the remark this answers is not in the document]
aaneja: Since this is used only in`ClusterStatusTracker`, lets move it there - 
```
public ClusterStatusTracker(ClusterManager clusterManager, RemoteStateConfig remoteStateConfig)
        {
            pollingInterval = remoteStateConfig.getPollingInterval();
...
```

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterManager.java
```diff
@@ -21,88 +22,376 @@
...
+                return remoteClusterInfos.get(entry).isHealthy();
+            }
+            else {
+                log.warn("Cluster [%s] is not available in remoteClusterInfos", entry);
+                return false;
+            }
+        }).collect(Collectors.toList());
+
+        if (healthyClusterURIs.isEmpty()) {
+            log.info("No healthy clusters found, enabled routing to unhealthy clusters");
```
aaneja, line 196: nit: `No healthy cluster found, will attempt to route using existing group spec`

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterManager.java
aaneja: Also, let's inline this method, it ends up being a single statement

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterManager.java
```diff
@@ -21,88 +22,376 @@
...
+        }
+    }
+
+    private static class ClusterManagerConfig
+    {
+        private final Map<String, GroupSpec> groups;
+        private final List<SelectorRuleSpec> groupSelectors;
+        private final Scheduler scheduler;
+        private final SchedulerType schedulerType;
+        private final Map<String, HashMap<URI, Integer>> serverWeights = new HashMap<>();
```
aaneja, line 350: nit : `Map<String, HashMap<URI, Integer>>` -> `Map<String, Map<URI, Integer>>`

aaneja (dismissed):
LGTM, thank you for fixing all the comments!

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterManager.java
auden-woolfson: Also had to change the variables in scheduler

aaneja (approved):

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterManager.java
```diff
@@ -21,88 +22,372 @@
...
-        this.groupSelectors = ImmutableList.copyOf(routerSpec.getSelectors());
-        this.schedulerType = routerSpec.getSchedulerType();
-        this.scheduler = new SchedulerFactory(routerSpec.getSchedulerType()).create();
+    @PostConstruct
+    public void startConfigReloadTaskFileWatcher()
+    {
+        CompletableFuture.supplyAsync(() -> {
+            try (WatchService watchService = FileSystems.getDefault().newWatchService()) {
+                File routerConfigFile = new File(routerConfig.getConfigFile());
+                log.info(String.format("Router config watch service monitoring %s", routerConfig.getConfigFile()));
```
ZacBlanco, line 129: don't need to do string format on the log statement
```suggestion
                log.info("Router config watch service monitoring %s", routerConfig.getConfigFile());
```

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterManager.java
```diff
@@ -21,88 +22,372 @@
...
+                        SensitivityWatchEventModifier.HIGH);
+                isWatchServiceStarted.set(true);
+                log.info("Successfully registered watch service for %s", parentDir);
 
-        this.initializeServerWeights();
+                while (true) {
+                    WatchKey key = watchService.take();
+                    log.info("Changes to router config directory detected: %s", routerConfigFile);
+                    for (WatchEvent<?> event : key.pollEvents()) {
+                        log.info("Event detected: %s, path: %s", event.kind().name(), event.context());
```
ZacBlanco, line 146: can we move this to debug level?

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterManager.java
```diff
@@ -21,88 +22,372 @@
...
+                        Path changed = (Path) event.context();
+                        if (changed.endsWith(routerConfigFile.getName())) {
+                            this.onConfigChangeDetection.apply();
+                        }
+                        else {
+                            log.info("Config change to %s ignored by ClusterManager (config file is %s)", event.context(), routerConfigFile.getName());
```
ZacBlanco, line 152: similarly, can we set this to debug?

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterManager.java
```diff
@@ -21,88 +22,372 @@
...
+                        }
+                    }
+                    key.reset();
+                }
+            }
+            catch (Exception e) {
+                log.error("Exception in file watcher loop while monitoring %s, %s", routerConfig.getConfigFile(), e);
```
ZacBlanco, line 159: If this future exits for some reason, should we stop the router? or attempt to restart it?

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterManager.java
```diff
@@ -21,88 +22,372 @@
...
+        checkArgument(config.getGroups().containsKey(target.get()));
+        GroupSpec groupSpec = config.getGroups().get(target.get());
+
+        List<URI> healthyClusterURIs = groupSpec.getMembers().stream().filter((entry) -> {
+            if (remoteClusterInfos.containsKey(entry)) {
+                return remoteClusterInfos.get(entry).isHealthy();
+            }
+            else {
+                log.warn("Cluster [%s] is not available in remoteClusterInfos", entry);
+                return false;
```
ZacBlanco, line 189: I don't think we need to do the `containsKey` check here. How about an Optional?
```suggestion
            Optional.ofNullable(remoteClusterInfos.get(entry)
                .map(RemoteClusterInfo::isHealthy)
                .orElse(false)
```

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterManager.java
```diff
@@ -21,88 +22,372 @@
...
+            }
+        }).collect(Collectors.toList());
+
+        if (healthyClusterURIs.isEmpty()) {
+            log.info("No healthy cluster found, will attempt to route using existing group spec");
```
ZacBlanco, line 194: this should either be removed or set to debug level

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterManager.java
```diff
@@ -21,88 +22,372 @@
...
+            healthyClusterURIs = groupSpec.getMembers();
         }
+        log.debug("Available clusters: %s", healthyClusterURIs);
```
ZacBlanco, line 197: This should be removed or debug

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterManager.java
```diff
@@ -21,88 +22,372 @@
...
+        {
+            queryInfoUpdateExecutor.scheduleWithFixedDelay(() -> {
+                try {
+                    clusterManager.getRemoteClusterInfos().values().forEach(RemoteClusterInfo::asyncRefresh);
+                    clusterManager.getRemoteQueryInfos().values().forEach(RemoteQueryInfo::asyncRefresh);
+                }
+                catch (Exception e) {
+                    log.error(e, "Error polling list of queries");
+                }
+            }, pollingInterval.toMillis(), pollingInterval.toMillis(), TimeUnit.MILLISECONDS);
```
ZacBlanco, line 280: nit: static import
```suggestion
            }, pollingInterval.toMillis(), pollingInterval.toMillis(),MILLISECONDS);
```

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterStatusResource.java
```diff
@@ -34,15 +34,15 @@
...
 public class ClusterStatusResource
 {
     private final String environment;
-    private final ClusterStatusTracker clusterStatusTracker;
+    private final ClusterManager.ClusterStatusTracker clusterStatusTracker;
     private final ClusterManager clusterManager;
 
     @Inject
-    public ClusterStatusResource(NodeInfo nodeInfo, ClusterStatusTracker clusterStatusTracker, ClusterManager clusterManager)
+    public ClusterStatusResource(NodeInfo nodeInfo, ClusterManager.ClusterStatusTracker clusterStatusTracker, ClusterManager clusterManager)
```
ZacBlanco, line 45: nit: can we static import ClusterStatusTracker

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterStatusResource.java
```diff
@@ -34,15 +34,15 @@
 import static java.util.Objects.requireNonNull;
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
 
-@Path("/")
+@Path("/v1")
```
ZacBlanco, line 37: Do we need to make this change to the `@Path`?

presto-router/src/main/java/com/facebook/presto/router/cluster/RemoteState.java
```diff
@@ -67,10 +77,18 @@ public synchronized void asyncRefresh()
         if (nanosSince(lastWarningLogged.get()).toMillis() > 1_000 &&
                 sinceUpdate.toMillis() > 10_000 &&
                 future.get() != null) {
-            log.warn("Coordinator update request to %s has not returned in %s", remoteUri, sinceUpdate.toString(SECONDS));
+            log.warn(
+                    "Coordinator update request to %s has not returned in %s",
+                    String.format("%s:%d", remoteUri.getHost(), remoteUri.getPort()),
+                    sinceUpdate.toString(SECONDS));
```
ZacBlanco, line 83: use log formatting
```suggestion
            log.warn(
                    "Coordinator update request to %s:%d has not returned in %s",
                    remoteUri.getHost(), remoteUri.getPort(), sinceUpdate.toString(SECONDS));
```

presto-router/src/main/java/com/facebook/presto/router/cluster/RemoteState.java
```diff
@@ -67,10 +77,18 @@ public synchronized void asyncRefresh()
...
             lastWarningLogged.set(System.nanoTime());
         }
 
+        if (java.time.Duration.between(lastHealthyResponseTime, Instant.now()).compareTo(clusterUnhealthyTimeout) >= 0 && isHealthy.get()) {
```
ZacBlanco, line 87: why fully-qualified usage? Can we import?

presto-router/src/main/java/com/facebook/presto/router/cluster/RemoteState.java
```diff
@@ -91,10 +109,19 @@ public void onSuccess(@Nullable FullJsonResponseHandler.JsonResponse<JsonNode> r
                             handleResponse(result.getValue());
                         }
                         if (result.getStatusCode() != OK.code()) {
-                            log.warn("Error fetching node state from %s returned status %d", remoteUri, result.getStatusCode());
-                            return;
+                            log.warn("Error fetching node state from %s returned status code %d", remoteUri, result.getStatusCode());
+                        }
+                        else {
+                            if (!isHealthy.get()) {
+                                log.info("%s:%d was unhealthy, and is now healthy", remoteUri.getHost(), remoteUri.getPort());
```
ZacBlanco, line 116: These feel like they should be at the debug level

presto-router/src/main/java/com/facebook/presto/router/cluster/RemoteState.java
```diff
@@ -91,10 +109,19 @@ public void onSuccess(@Nullable FullJsonResponseHandler.JsonResponse<JsonNode> r
...
+                                isHealthy.set(true);
+                            }
+                            lastHealthyResponseTime = Instant.now();
                         }
                     }
+                    else {
+                        log.info("RemoteState successful result was null");
```
ZacBlanco, line 123: This should either be removed or set to debug

presto-router/src/test/java/com/facebook/presto/router/BarrierClusterManager.java
```diff
@@ -0,0 +1,47 @@
...
+    public BarrierClusterManager(RouterConfig config, RemoteInfoFactory remoteInfoFactory, RemoteStateConfig remoteStateConfig, CyclicBarrier barrier)
+    {
+        super(config, remoteInfoFactory, remoteStateConfig);
+        this.barrier = barrier;
+
+        OnConfigChangeDetection parentOnConfigChangeDetection = this.onConfigChangeDetection;
+        this.onConfigChangeDetection = () -> {
+            try {
+                parentOnConfigChangeDetection.apply();
+                this.barrier.await(5, TimeUnit.SECONDS);
```
ZacBlanco, line 39:
```suggestion
               barrier.await(5, TimeUnit.SECONDS);
```

presto-router/src/test/java/com/facebook/presto/router/TestClusterManager.java
```diff
@@ -127,6 +141,42 @@ public void testQuery()
...
+        ClusterManager barrierClusterManager = new BarrierClusterManager(newRouterConfig, remoteInfoFactory, remoteStateConfig, barrier);
+        assertEquals(barrierClusterManager.getAllClusters().size(), 3);
+
+        while (!barrierClusterManager.getIsWatchServiceStarted()) {
+            Thread.sleep(10);
+        }
+
+        Path configFilePath = newConfig.toPath();
+        String originalConfigContent = new String(Files.readAllBytes(configFilePath));
+        String modifiedConfigContent = originalConfigContent.replaceAll("\"members\"\\s*:\\s*\\[.*?\\]", "\"members\": []");
```
ZacBlanco, line 163: Rather than do this replacing, can you just read it into a spec using a jsonCodec, and then write it out again via the same jsonCodec? it will be less error-prone

presto-router/src/test/java/com/facebook/presto/router/TestHealthChecks.java
```diff
@@ -0,0 +1,168 @@
...
+        List<String> serverURIs = prestoServers.stream()
+                .map(TestingPrestoServer::getBaseUrl)
+                .map(URI::toString)
+                .collect(Collectors.toList());
+
+        Map<String, List<String>> groups = ImmutableMap.of("group1", serverURIs);
+        TestingRouterConfig config = new TestingRouterConfig(groups, "group1", "ROUND_ROBIN");
+
+        FileOutputStream fileOutputStream = new FileOutputStream(configFile);
+        fileOutputStream.write(config.get().toString().getBytes(UTF_8));
```
ZacBlanco, line 84: Either use try-with-resources or just `Files.write`

presto-router/src/test/java/com/facebook/presto/router/TestingRouterConfig.java
```diff
@@ -0,0 +1,53 @@
...
+            group.put("members", membersArray);
+            groupsArray.add(group);
+        });
+
+        JSONArray selectorsArray = new JSONArray();
+        JSONObject selector = new JSONObject();
+        selector.put("targetGroup", targetGroup);
+        selectorsArray.add(selector);
+
+        config = new JSONObject();
```
ZacBlanco, line 43: Why are we creating custom `JSONObject`s? Doesn't the router config have a POJO? Then we can just serialize and deserialize that with airlift's jsonCodec

ZacBlanco (changes requested):

presto-router/pom.xml
```diff
@@ -162,22 +162,40 @@
...
+            <groupId>com.facebook.airlift</groupId>
+            <artifactId>stats</artifactId>
+        </dependency>
+
         <!-- test dependencies-->
         <dependency>
-            <groupId>org.testng</groupId>
-            <artifactId>testng</artifactId>
+            <groupId>com.googlecode.json-simple</groupId>
+            <artifactId>json-simple</artifactId>
```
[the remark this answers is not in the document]
auden-woolfson: Used in `TestingRouterConfig`

presto-router/src/main/java/com/facebook/presto/router/RouterResource.java
```diff
@@ -38,13 +40,13 @@
 import static javax.ws.rs.core.MediaType.TEXT_PLAIN_TYPE;
 import static javax.ws.rs.core.Response.Status.BAD_GATEWAY;
 
-@Path("/")
-@PermitAll
+@Path("/v1")
```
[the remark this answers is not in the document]
auden-woolfson: @saravanan19 what was the purpose of this change? I am going to go back to the old endpoints

presto-router/src/main/java/com/facebook/presto/router/cluster/RemoteState.java
auden-woolfson: Kind of confusing but we also use airlift duration in this file

presto-router/src/test/java/com/facebook/presto/router/TestClusterManager.java
auden-woolfson: Ok, I will need to add some setters to router spec

ZacBlanco: Do you need setters? You should be able to just create a new instance from the existing instance through the constructor

presto-router/src/test/java/com/facebook/presto/router/TestingRouterConfig.java
auden-woolfson: You are referring to RouterSpec?

ZacBlanco: I think so? That's the object type you're trying to create right?

ZacBlanco (changes requested):

presto-router/src/main/java/com/facebook/presto/router/cluster/RemoteStateConfig.java
```diff
@@ -0,0 +1,66 @@
...
+    public RemoteStateConfig setClusterUnhealthyTimeout(String clusterUnhealthyTimeout)
+    {
+        this.clusterUnhealthyTimeout = Duration.parse(clusterUnhealthyTimeout);
+        return this;
+    }
+
+    public void setClusterUnhealthyTimeout(Duration timeToUnhealthy)
+    {
+        this.clusterUnhealthyTimeout = timeToUnhealthy;
+    }
```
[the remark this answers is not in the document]
auden-woolfson: Isn't it necessary to have one that accepts a String so the config will work? I am getting this error...

1) Error: Could not coerce value 'PT4S' to Duration (property 'router.remote-state.cluster-unhealthy-timeout') in order to call [public void RemoteStateConfig.setClusterUnhealthyTimeout(Duration)]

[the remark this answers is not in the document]
ZacBlanco: The airlift:configuration module automatically parses strings into airlift's `Duration` type if the input argument to the `set` method is a Duration and return value from a `get` is a Duration. We should be using the airlift `Duration` type, and not the java Duration.

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterManager.java
ZacBlanco: I think you can inject a `LifeCycleManager` and call `lifecycleManager.stop()` in the case this loop ever exits

ZacBlanco: The other option, is to make sure that the loop has no exit conditions by catching all exceptions and only breaking on an interrupt

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterManager.java
```diff
@@ -21,88 +22,360 @@
...
+                log.info("Successfully registered watch service for %s", parentDir);
+
+                while (true) {
+                    WatchKey key = watchService.take();
+                    log.info("Changes to router config directory detected: %s", routerConfigFile);
+                    for (WatchEvent<?> event : key.pollEvents()) {
+                        log.debug("Event detected: %s, path: %s", event.kind().name(), event.context());
+                        Path changed = (Path) event.context();
+                        if (changed.endsWith(routerConfigFile.getName())) {
+                            onConfigChangeDetection();
```
ZacBlanco, line 146: If this ever throws an exception, the loop will exit. This could easily happen with a bad router config. Should we catch exceptions here?

presto-router/src/main/java/com/facebook/presto/router/cluster/ClusterStatusResource.java
```diff
@@ -51,15 +52,15 @@ public ClusterStatusResource(NodeInfo nodeInfo, ClusterStatusTracker clusterStat
 
     // The web UI depend on the following service endpoints.
     @GET
-    @Path("/v1/info")
+    @Path("v1/info")
```
ZacBlanco, line 55: Missing leading `/` intentional?

presto-router/src/main/java/com/facebook/presto/router/cluster/RemoteState.java
```diff
@@ -67,10 +77,17 @@ public synchronized void asyncRefresh()
...
                 sinceUpdate.toMillis() > 10_000 &&
                 future.get() != null) {
-            log.warn("Coordinator update request to %s has not returned in %s", remoteUri, sinceUpdate.toString(SECONDS));
+            log.warn(
+                    "Coordinator update request to %s:%d has not returned in %s",
+                    remoteUri.getHost(), remoteUri.getPort(), sinceUpdate.toString(SECONDS));
             lastWarningLogged.set(System.nanoTime());
         }
 
+        if (Duration.nanosSince(lastHealthyResponseTime.toEpochMilli()).toMillis() > clusterUnhealthyTimeout.toMillis() && isHealthy.get()) {
```
ZacBlanco, line 86: nit: static import on nanosSince

ZacBlanco (changes requested):

presto-router/src/main/java/com/facebook/presto/router/RouterResource.java
```diff
@@ -53,19 +55,20 @@ public RouterResource(ClusterManager clusterManager)
...
         URI coordinatorUri = clusterManager.getDestination(requestInfo).orElseThrow(() -> badRequest(BAD_GATEWAY, "No Presto cluster available"));
         URI statementUri = uriBuilderFrom(coordinatorUri).replacePath("/v1/statement").build();
+        successRedirectRequests.update(1);
         log.info("route query to %s", statementUri);
         return Response.temporaryRedirect(statementUri).build();
     }
 
     @GET
-    @Path("/")
+    @Path("v1")
```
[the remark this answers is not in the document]
auden-woolfson: This is the same path just refactored since the root no longer has v1

presto-router/src/main/java/com/facebook/presto/router/RouterResource.java
```diff
@@ -53,19 +55,20 @@ public RouterResource(ClusterManager clusterManager)
     }
 
     @POST
-    @Path("/v1/statement")
+    @Path("v1/statement")
```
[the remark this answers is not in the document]
auden-woolfson: I can add it back

presto-router/src/main/java/com/facebook/presto/router/RouterResource.java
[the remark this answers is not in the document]
ZacBlanco: We should keep all of the paths the same unless there is an intentional change to one of the endpoints

ZacBlanco (approved):
lgtm

PR #24481, opened 2025-02-03
[Iceberg] Add manifest file caching for HMS-based deployments

ZacBlanco:
## Description

Adds manifest file caching to the Iceberg connector for HMS-based deployments.

## Motivation and Context

In order to optimize and plan iceberg queries we call the `planFiles()` API multiple times throughout the query optimization lifecycle. Each time it requires reading and parsing metadata files which usually exist on an external filesystem such as S3. For large tables there could be hundreds of files. They usually range in a few kilobytes in size up to a few megabytes. When not cached in memory within Presto it can lead to significant E2E query latency degradation.

## Impact

- configuration properties for manifest file caching now affect the HIVE catalog type

## Test Plan

- tests added to verify caching is used in hive when the configuration flag is enabled/disabled

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

Iceberg Connector Changes
* Add manifest file caching for deployments which use the Hive metastore.
* Enable manifest caching by default.
```


jaystarshot (commented):
Sorry I may not have the correct context on this but is it possible to add some tests too?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergConfig.java
```diff
@@ -67,7 +67,7 @@ public class IcebergConfig
 
     private EnumSet<ColumnStatisticType> hiveStatisticsMergeFlags = EnumSet.noneOf(ColumnStatisticType.class);
     private String fileIOImpl = HadoopFileIO.class.getName();
-    private boolean manifestCachingEnabled;
+    private boolean manifestCachingEnabled = true;
```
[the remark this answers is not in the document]
ZacBlanco: Yes this is intentional. Performance is significantly worse without it disabled, and I don't think there are any known downsides to making this enabled by default other than an increased memory footprint

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergCommonModule.java
```diff
@@ -200,6 +202,21 @@ public StatisticsFileCache createStatisticsFileCache(IcebergConfig config, MBean
         return new StatisticsFileCache(delegate);
     }
 
+    @Singleton
+    @Provides
+    public ManifestFileCache createManifestFileCache(IcebergConfig config, MBeanExporter exporter)
+    {
+        Cache<ManifestFileCacheKey, ManifestFileCachedContent> delegate = CacheBuilder.newBuilder()
+                .maximumWeight(config.getManifestCachingEnabled() ? config.getMaxManifestCacheSize() : 0)
```
jaystarshot, line 210: If the caching is disabled i think we should not have any caching instead of adding this via 0 weight

presto-iceberg/src/main/java/com/facebook/presto/iceberg/HdfsCachedInputFile.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.List;
+
+import static java.util.Objects.requireNonNull;
+import static org.apache.iceberg.io.IOUtil.readRemaining;
+
+public class HdfsCachedInputFile
+        implements InputFile
+{
+    private static final Logger LOG = Logger.get(HdfsCachedInputFile.class);
+    private static final long BUFFER_CHUNK_SIZE = 2 * 1024 * 1024;
```
aaneja, line 33: Is the default here 2Mb for a specific reason ? Can we make this Integer.MAX_VALUE ?

aaneja (commented):
Can you post some metrics about cache hit ratios/eviction for a canonical read-heavy workload ? Maybe like partitioned/unpartitioned TPCDS ?

ZacBlanco:
> Can you post some metrics about cache hit ratios/eviction for a canonical read-heavy workload ? Maybe like partitioned/unpartitioned TPCDS ?

I have not tested on a partitioned table yet, but on our internal unpartitioned TPC-DS SF1k dataset executed in the "ds_power" configuration (1 query at a time, q1 through q99), the cache hit rate was 96.8%. There were a small number of evictions. When testing locally on an sf10 dataset generated from the `tpcds.sf10` schema the hit rate was 99.7%

presto-iceberg/src/main/java/com/facebook/presto/iceberg/HdfsCachedInputFile.java
ZacBlanco: I think `Integer.MAX_VALUE` is too large. The iceberg code does around 4MB I believe. There is a balance to strike in the chunk sizes. Since the byte buffer internally is going to store a contiguous array of bytes, you need that much memory available contiguously. When you break it into chunks, it's easier for the allocator to find smaller chunks and puts less pressure on the GC to evict when under load. I think going as high as 16MB would probably be OK. My gut says 2MB is probably fine but I am open to tweaking it

presto-iceberg/src/main/java/com/facebook/presto/iceberg/HdfsCachedInputFile.java
```diff
@@ -0,0 +1,120 @@
...
+    {
+        ManifestFileCachedContent cachedContent = cache.getIfPresent(cacheKey);
+        if (cachedContent != null) {
+            return ByteBufferInputStream.wrap(cachedContent.getData());
+        }
+
+        long fileLength = delegate.getLength();
+        if (fileLength <= cache.getMaxFileLength() && cache.isEnabled()) {
+            try {
+                ManifestFileCachedContent content = readFully(delegate, fileLength);
```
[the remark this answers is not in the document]
ZacBlanco: This is a good question. For the most part when dealing with manifests, the files are read fully. However, there are cases where it is not fully used. ex. when reading partition specs in avro format, you [only need to read the file metadata](https://github.com/apache/iceberg/blob/3e6da2e5437ffb3f643275927e5580cb9620256b/core/src/main/java/org/apache/iceberg/ManifestReader.java#L123)

However, in order to plan an entire query you need to read all of the (valid) manifest files fully. You won't really ever only need just the partition specs. The partition specs are going to be contained within one of those files anyways.

Additionally, when caching is enabled on catalogs other than HMS, this is [the same approach as in the Iceberg library](https://github.com/apache/iceberg/blob/3e6da2e5437ffb3f643275927e5580cb9620256b/core/src/main/java/org/apache/iceberg/io/ContentCache.java#L252)

presto-iceberg/src/main/java/com/facebook/presto/iceberg/ManifestFileCache.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    @Managed
+    @Nested
+    public DistributionStat getFileSizeDistribution()
+    {
+        return fileSizes;
+    }
+
+    public void recordFileSize(long size)
```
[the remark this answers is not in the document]
ZacBlanco: Ah good point. I dug deeper and found that the underlying `Distribution` objects use `synchronized` on the `add` function so this should be thread-safe already. See https://github.com/prestodb/airlift/blob/d497a42c7158a739cd743239dfa31b7d0f3dc44a/stats/src/main/java/com/facebook/airlift/stats/Distribution.java#L52-L58

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergCommonModule.java
```diff
@@ -200,6 +202,21 @@ public StatisticsFileCache createStatisticsFileCache(IcebergConfig config, MBean
         return new StatisticsFileCache(delegate);
     }
 
+    @Singleton
+    @Provides
+    public ManifestFileCache createManifestFileCache(IcebergConfig config, MBeanExporter exporter)
+    {
+        Cache<ManifestFileCacheKey, ManifestFileCachedContent> delegate = CacheBuilder.newBuilder()
```
aaneja, line 209: Should we just use Caffeine as the caching library since iceberg-core already brings it in ? It appears to have [better performance](https://github.com/ben-manes/caffeine/wiki/Benchmarks) and is [recommended by the Guava team](https://guava.dev/releases/snapshot/api/docs/com/google/common/cache/package-summary.html) too

ZacBlanco: I had the same thought too. However, most of the Presto codebase uses guava caches. Since caffeine and guava are different types, it would not be compatible with the current infrastructure such as the `CacheStatsMBean` object. Additionally, the way caches are implemented now use guava's `SimpleForwardingCache` which is not available in caffeine, so I would have to roll my own. Not a terrible amount of effort, but I think there's enough work there to push that effort into a separate PR

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergCommonModule.java
ZacBlanco: I have updated the PR to remove use of the 0 weight and bypass the cache entirely if it is not enabled.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/HdfsCachedInputFile.java
jaystarshot: Its better for this to come from a config

steveburnett:
Nit, suggested rephrase of release note to follow the [Order of changes](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines#order-of-changes) phrasing in the Release Notes Guidelines: 

```
== RELEASE NOTES ==

Iceberg Connector Changes
* Add manifest file caching for deployments which use the Hive metastore.
* Add enable by default for manifest file caching.
```

hantangwangd:
A little unsure about this. Please correct me if I'm wrong, should we just implement the method `Map<String, String> properties()` for `HdfsFileIO`, so that we can utilize the Iceberg lib's manifest file cache even when configuring with our native hive catalog? Or is there any other problems I didn't notice?

The reference code in Iceberg lib could be found [here](https://github.com/apache/iceberg/blob/3e6da2e5437ffb3f643275927e5580cb9620256b/core/src/main/java/org/apache/iceberg/ManifestFiles.java#L387-L405). So it seems that the following code in `HdfsFileIO` could utilize the manifest file cache:

```
    public Map<String, String> properties()
    {
        return IcebergUtil.loadCachingProperties(icebergConfig);
    }
```

aaneja:
LGTM % tests

presto-iceberg/src/main/java/com/facebook/presto/iceberg/HdfsFileIO.java
```diff
@@ -17,31 +17,63 @@
...
         this.environment = requireNonNull(environment, "environment is null");
         this.context = requireNonNull(context, "context is null");
+        this.manifestFileCache = requireNonNull(manifestFileCache, "manifestFileCache is null");
     }
 
     @Override
     public InputFile newInputFile(String path)
     {
-        return new HdfsInputFile(new Path(path), environment, context);
+        return new HdfsInputFile(new Path(path), environment, context, Optional.empty());
```
hantangwangd, line 49: nit: no need to do this change?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/HdfsCachedInputFile.java
```diff
@@ -0,0 +1,120 @@
...
+    public String location()
+    {
+        return delegate.location();
+    }
+
+    @Override
+    public boolean exists()
+    {
+        return false;
+    }
```
hantangwangd, line 90: Why directly return a `false` here?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/HdfsCachedInputFile.java
hantangwangd: Agree to set a slightly smaller value here.

hantangwangd (commented):
Thanks for the explanation. Lgtm, only a couple of nit and small question.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/HdfsCachedInputFile.java
ZacBlanco: fixed. Good catch

presto-iceberg/src/main/java/com/facebook/presto/iceberg/HdfsFileIO.java
ZacBlanco: reverted

hantangwangd (dismissed):
Thanks for the fix, lgtm!

presto-iceberg/src/main/java/com/facebook/presto/iceberg/HdfsCachedInputFile.java
ZacBlanco: I have updated this to be able to configure it from the catalog.properties file

aaneja (dismissed):

steveburnett (dismissed):
LGTM! (docs)

Pull branch, local doc build. Looks good, thanks!

presto-iceberg/src/test/java/com/facebook/presto/iceberg/hive/TestIcebergDistributedHive.java
```diff
@@ -66,6 +84,94 @@ public void testStatisticsFileCache()
...
+                .getCatalogMetadata(metadataSession.getTransactionId().get(), handle.get().getConnectorId());
+        Field delegate = ClassLoaderSafeConnectorMetadata.class.getDeclaredField("delegate");
+        delegate.setAccessible(true);
+        IcebergHiveMetadata metadata = (IcebergHiveMetadata) delegate.get(catalogMetadata.getMetadataFor(handle.get().getConnectorId()));
+        ManifestFileCache manifestFileCache = metadata.getManifestFileCache();
+        assertFalse(manifestFileCache.isEnabled());
+        CacheStats initial = manifestFileCache.stats();
+        assertQuerySucceeds(session, "SELECT count(*) from test_manifest_file_cache_disabled group by i");
+        CacheStats firstQuery = manifestFileCache.stats();
+        assertEquals(firstQuery.minus(initial).hitCount(), 0); //
```
hantangwangd, line 164:
```suggestion
        assertEquals(firstQuery.minus(initial).hitCount(), 0);
```

presto-iceberg/src/test/java/com/facebook/presto/iceberg/hive/TestIcebergDistributedHive.java
```diff
@@ -66,6 +84,94 @@ public void testStatisticsFileCache()
...
+        assertEquals(manifestFileCache.size(), 0);
+        assertQuerySucceeds(session, "SELECT count(*) from test_manifest_file_cache_disabled group by i");
+        CacheStats secondQuery = manifestFileCache.stats();
+        assertEquals(secondQuery.minus(firstQuery).hitCount(), 0);
+        assertEquals(manifestFileCache.size(), 0);
+
+        //drop table
+        assertQuerySucceeds(session, "DROP TABLE test_manifest_file_cache_disabled");
```
hantangwangd, line 172: nit: drop schema as well?

presto-iceberg/src/test/java/com/facebook/presto/iceberg/hive/TestIcebergDistributedHive.java
```diff
@@ -66,6 +84,94 @@ public void testStatisticsFileCache()
...
+        assertTrue(firstQuery.minus(initial).missCount() > 0);
+        assertTrue(manifestFileCache.size() > 0);
+        assertQuerySucceeds(session, "SELECT count(*) from test_manifest_file_cache group by i");
+        CacheStats secondQuery = manifestFileCache.stats();
+        assertEquals(secondQuery.minus(firstQuery).missCount(), 0);
+        assertTrue(secondQuery.minus(firstQuery).hitCount() > 0);
+        assertTrue(manifestFileCache.size() > 0);
+
+        //drop table
+        assertQuerySucceeds(session, "DROP TABLE test_manifest_file_cache");
```
hantangwangd, line 129: nit: drop schema as well?

hantangwangd (commented):
Some little nits, otherwise looks good to me.

hantangwangd (approved):

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergCommonModule.java
```diff
@@ -213,6 +215,26 @@ public StatisticsFileCache createStatisticsFileCache(IcebergConfig config, MBean
         return statisticsFileCache;
     }
 
+    @Singleton
+    @Provides
+    public ManifestFileCache createManifestFileCache(IcebergConfig config, MBeanExporter exporter)
+    {
+        CacheBuilder<ManifestFileCacheKey, ManifestFileCachedContent> delegate = CacheBuilder.newBuilder()
```
agrawalreetika, line 222: Two Questions here -
1. Did we consider using the Caffeine cache itself for Manifest file caching? Could Caffeine Window TinyLfu eviction be effective here?
2. For the cache key, we can just use the Manifest file path string, or are there issues with that?

agrawalreetika (commented):
Mostly LGTM apart from small nits and questions.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/HdfsFileIO.java
```diff
@@ -44,6 +49,33 @@ public InputFile newInputFile(String path)
...
+                inputFile;
+    }
+
+    @Override
+    public InputFile newInputFile(ManifestFile manifest)
+    {
+        checkArgument(
+                manifest.keyMetadata() == null,
+                "Cannot decrypt manifest: {} (use EncryptingFileIO)",
+                manifest.path());
```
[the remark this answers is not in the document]
ZacBlanco: Good catch. Fixed

presto-iceberg/src/main/java/com/facebook/presto/iceberg/HdfsCachedInputFile.java
```diff
@@ -0,0 +1,119 @@
...
+        long fileLength = delegate.getLength();
+        if (cache.isEnabled() && fileLength <= cache.getMaxFileLength()) {
+            try {
+                ManifestFileCachedContent content = readFully(delegate, fileLength, cache.getBufferChunkSize());
+                cache.put(cacheKey, content);
+                cache.recordFileSize(content.getLength());
+                return ByteBufferInputStream.wrap(content.getData());
+            }
+            catch (IOException e) {
+                LOG.warn("Failed to cache input file. Falling back to direct read.", e);
```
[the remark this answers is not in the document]
ZacBlanco: updated

presto-iceberg/src/main/java/com/facebook/presto/iceberg/HdfsCachedInputFile.java
```diff
@@ -0,0 +1,119 @@
...
+                int bytesToRead = (int) Math.min(chunkSize, totalBytesToRead);
+                byte[] buf = new byte[bytesToRead];
+                int bytesRead = readRemaining(stream, buf, 0, bytesToRead);
+                totalBytesToRead -= bytesRead;
+
+                if (bytesRead < bytesToRead) {
+                    throw new IOException(
+                            String.format(
+                                    "Failed to read %d bytes: %d bytes in stream",
+                                    fileLength, fileLength - totalBytesToRead));
```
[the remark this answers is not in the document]
ZacBlanco: updated

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergCommonModule.java
ZacBlanco: I opted to create a new type for the cache key even though it's just the file in case we need to modify it more easily in the future

As for caffeine, I spoke about this with Anant, but it's more complicated to use them due to a lack of utility classes and existing methods for Jmx metric support. It's a large enough effort that moving to Caffeine should be a separate PR. I have a draft up already for this. However, it still needs a little bit of work before it's ready to review #24608

agrawalreetika (approved):

hantangwangd (dismissed):

yingsu00 (dismissed):
Just one nit

presto-iceberg/src/main/java/com/facebook/presto/iceberg/HdfsFileIO.java
```diff
@@ -44,6 +49,33 @@ public InputFile newInputFile(String path)
         return new HdfsInputFile(new Path(path), environment, context);
     }
 
+    @Override
+    public InputFile newInputFile(String path, long length)
+    {
+        return new HdfsInputFile(new Path(path), environment, context, Optional.of(length));
+    }
+
+    protected InputFile newCachedInputFile(String path)
```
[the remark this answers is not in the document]
ZacBlanco: done

agrawalreetika (approved):

yingsu00 (approved):

hantangwangd (approved):

agrawalreetika (approved):

PR #24493, opened 2025-02-04
Upgrade cache Github action to v4

pramodsatya:
## Description
Github action `cache` in CI workflows is currently at `v2` and CI failures are randomly seen with the [following error](https://github.com/prestodb/presto/actions/runs/13141607984/job/36669864369?pr=24490):
```
Error: This request has been automatically failed because it uses a deprecated version of `actions/cache: v2`. Please update your workflow to use v3/v4 of actions/cache to avoid interruptions.
```
v2 of `actions/cache` appears to be deprecated, this PR upgrades `actions/cache` to v4 as [recommended](https://github.com/actions/cache).

## Motivation and Context
Fix flaky CI.

```
== NO RELEASE NOTE ==
```

PR #24512, opened 2025-02-07
Upgrade to actions/setup-java@v4

ZacBlanco:
## Description

Since #24493 was merged I was looking at CI and noticed that `setup-java` is also quite old. This PR upgrades to setup-java@v4 and removes the home-grown go-offline+cache solution in favor of the built-in solution. Internally they use actions/cache and it also simplifies the CI configs. The new cache key is the hash of all pom files in the project. 

The caching behavior is actually slightly different than before. Previously we would restore the cache using two cache keys: one with a hash of the pom files, and one without any hash, such as `linux-maven-2-`. This downside of this approach is that the cache grows over time because when you restore from a cache which doesn't used a hash of the POMs, and make a dependency update, the cache may be saved again that contains both old dependency and new dependencies.

When I compare the size of cache hits on our actions, I see that our current cache size [is quite large](https://github.com/prestodb/presto/actions/runs/13037238038/job/36370602879?pr=24456#step:4:20)!

```
Run actions/cache@v2
Received 251658240 of 900086244 (28.0%), 240.0 MBs/sec
Received 532676608 of 900086244 (59.2%), 254.0 MBs/sec
Received 805306368 of 900086244 (89.5%), 256.1 MBs/sec
Received 900086244 of 900086244 (100.0%), 209.3 MBs/sec
Cache Size: ~858 MB (900086244 B)
```

Compared to restoring fresh caches with [this PR](https://github.com/prestodb/presto/actions/runs/13193474967/job/36830812486?pr=24512#step:4:42):

```
Run actions/setup-java@v4
Installed distributions
Creating settings.xml with server-id: github
Writing to /home/runner/.m2/settings.xml
Received 209715200 of 426035899 (49.2%), 198.6 MBs/sec
Cache Size: ~406 MB (426035899 B)
```

Materially, it doesn't make a huge difference in time when cache restores take on the order of a few seconds, but it may save on some storage costs.

## Motivation and Context

Less complex CI pipelines and access to newer java versions

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/.github/workflows/arrow-flight-tests.yml
+++ b/.github/workflows/arrow-flight-tests.yml
@@ -46,28 +46,15 @@ jobs:
         with:
           show-progress: false
 
-      # Set up Java for the build environment
+      # Set up Java and dependencies for the build environment
       - uses: actions/setup-java@v4
         if: needs.changes.outputs.codechange == 'true'
         with:
           distribution: 'temurin'
           java-version: 8
-
-      # Cache Maven dependencies to speed up the build
-      - name: Cache local Maven repository
-        if: needs.changes.outputs.codechange == 'true'
-        id: cache-maven
-        uses: actions/cache@v4
-        with:
-          path: ~/.m2/repository
-          key: ${{ runner.os }}-maven-2-${{ hashFiles('**/pom.xml') }}
-          restore-keys: |
-            ${{ runner.os }}-maven-2-
-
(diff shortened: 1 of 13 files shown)


czentgr (approved):
Thanks, this looks good and simplified and more efficient!

yingsu00 (approved):

PR #24533, opened 2025-02-11
Improve boostrap log to expose the wrong classLoader

ericyuliu:
Summary: 
Improve bootstrap error message for incorrect classloader to include what classloader was found.

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/sql/gen/Bootstrap.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/gen/Bootstrap.java
@@ -44,8 +44,12 @@ public final class Bootstrap
     public static CallSite bootstrap(MethodHandles.Lookup callerLookup, String name, MethodType type, long bindingId)
     {
         ClassLoader classLoader = callerLookup.lookupClass().getClassLoader();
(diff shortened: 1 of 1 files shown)


rschlussel:
nit: capitalize the commit title and remove references to internal connectors.

jaystarshot:
Looks like you need a rebase to fix the maven checks

rschlussel:
You still need to fix the commit message body.  The current commit body includes references to an internal connector not in open source and isn't very comprehensible to someone without context. The body here doesn't actually need much because the change is pretty self explanatory. Something like the following should be sufficient: "Improve bootstrap error message for incorrect classloader to include what classloader was found."

ericyuliu:
> You still need to fix the commit message body. The current commit body includes references to an internal connector not in open source and isn't very comprehensible to someone without context. The body here doesn't actually need much because the change is pretty self explanatory. Something like the following should be sufficient: "Improve bootstrap error message for incorrect classloader to include what classloader was found."

Fixed, thanks!

rschlussel (approved):

rschlussel:
Thanks!

PR #24534, opened 2025-02-11
Add `Join PrestoDB on Slack` link on document pages

yhwang:
## Description
Adding a link in the document pages to encourage users to join the PrestoDB Slack when seeking help.

## Motivation and Context
Document pages are where users look for instructions/information when setting up a Presto cluster.
Adding the link to direct users to the Slack channels when encountering any issues would
improve the adoption process.

## Impact
No function impact

## Test Plan
Generate the doc website and check the new link

## Contributor checklist

- [x] Please make sure your submission complies with our [contributing guide](https://github.com/prestodb/presto/blob/master/CONTRIBUTING.md), in particular [code style](https://github.com/prestodb/presto/blob/master/CONTRIBUTING.md#code-style) and [commit standards](https://github.com/prestodb/presto/blob/master/CONTRIBUTING.md#commit-standards).
- [x] PR description addresses the issue accurately and concisely.  If the change is non-trivial, a GitHub Issue is referenced.
- [ ] Documented new properties (with its default value), SQL syntax, functions, or other functionality.
- [ ] If release notes are required, they follow the [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines).
- [ ] Adequate tests were added if applicable.
- [ ] CI passed.

## Release Notes
Please follow [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines) and fill in the release notes below.

```
== NO RELEASE NOTE ==
```

PR #24537, opened 2025-02-12
Fix GH Actions for doc-only PR

yhwang:
## Description
When there is no code change, some steps in the GA workflow are not able to be performed and cause failures. Put the `if condition` to check the code change in those steps. For non-required tests, use `paths-ignore` to skip the test directly.

## Motivation and Context
fix the GH action error when a PR contains only document changes.
example failures: 
- https://github.com/prestodb/presto/actions/runs/13273735011/job/37058943327?pr=24534
- https://github.com/prestodb/presto/actions/runs/13273734993/job/37063671179?pr=24534

## Impact
doc-only PR can't be merged and need this fix

## Test Plan
verify the doc-only PR after this PR is merged

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/.github/workflows/arrow-flight-tests.yml
+++ b/.github/workflows/arrow-flight-tests.yml
@@ -2,6 +2,8 @@ name: arrow flight tests
 
 on:
   pull_request:
+    paths-ignore:
+      - 'presto-docs/**'
 
 env:
   CONTINUOUS_INTEGRATION: true
@@ -12,22 +14,8 @@ env:
   RETRY: .github/bin/retry
 
 jobs:
-  changes:
-    runs-on: ubuntu-latest
-    permissions:
-      pull-requests: read
-    outputs:
-      codechange: ${{ steps.filter.outputs.codechange }}
-    steps:
-      - uses: dorny/paths-filter@v2
-        id: filter
-        with:
-          filters: |
-            codechange:
-               - '!presto-docs/**'
   test:
     runs-on: ubuntu-latest
-    needs: changes
     strategy:
       fail-fast: false
       matrix:
@@ -42,13 +30,11 @@ jobs:
     steps:
       # Checkout the code only if there are changes in the relevant files
       - uses: actions/checkout@v4
-        if: needs.changes.outputs.codechange == 'true'
         with:
           show-progress: false
 
       # Set up Java and dependencies for the build environment
       - uses: actions/setup-java@v4
-        if: needs.changes.outputs.codechange == 'true'
         with:
           distribution: 'temurin'
           java-version: 8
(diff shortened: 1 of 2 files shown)


.github/workflows/arrow-flight-tests.yml
```diff
@@ -54,6 +54,7 @@ jobs:
           java-version: 8
           cache: 'maven'
       - name: Download nodejs to maven cache
+        if: needs.changes.outputs.codechange == 'true'
```
czentgr, line 57: This seems convoluted from the beginning.
The steps are not performed if the filter finds changes in files except in presto-docs via the filter.

The same can be much more easily done by triggering the action on paths except presto-docs. Then the check for each step falls away and lines 19 - 27 are not needed.

One needs to add after the `pull_request:` event as part of the event
```
   paths-ignore:
      - 'presto-docs/**'
```
Check one of the prestissimo workflows where the `paths` is used to subset the workflow for changes in presto-native-execution.

This would make it more intuitive and clear what is happening:
See also https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore

czentgr (commented):
Thanks. General comment.

.github/workflows/arrow-flight-tests.yml
yhwang: I remember one of the issues is that some tests are labeled as required and if those tests are not performed, `the required tests need to pass` criteria will fail.

yhwang: You can check the GA actions for this PR. They were triggered, ran, and finished. For the required tests, they are defined in the repo settings. I don't have permission to see that.

yhwang: FYI, you can check this [PR](https://github.com/prestodb/presto/pull/24534), a doc-only PR. And those code-related tests failed.

yhwang: `presto-base-arrow-flight` is not one of the required ones. But it failed in the doc-only PR too. So, might as well also fix it here.

czentgr: For PR 24534 the workflow fails because the checkout is skipped so of course it can't use something that requires a checkout in a subsequent check. 
My point is that the arrow flight test workflow should not have triggered for a doc-only change in the first place. We should not run workflows that don't pertain to the changes made.

yhwang: That makes sense, but that won't be addressed in this PR, right? I guess those non-required tests can be skipped at the very beginning.

czentgr: This PR can address it by making it so that the workflow doesn't run for presto-doc changes. No point in excluding every single step of the workflow on presto-doc changes.

yhwang: updated the airflight workflow. it will be skipped if there are only doc changes in a PR.

czentgr (approved):
Thanks!
Nit: Please update the description to mention that the arrow workflow is excluded from runs for doc-only changes rather than adding the if check as described.

yhwang:
updated!

PR #24545, opened 2025-02-12
Add evb violation counter for exchange client and http server IO threadpool

kewang1024:
When prestissimo worker is busy on CPU, we observed delays in exchange connection.
Add evb violation for exchange client IO thread pool  and exchange server IO thread pool 
to give us better visibility.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/PrestoServer.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoServer.cpp
@@ -409,6 +409,12 @@ void PrestoServer::run() {
                            << exchangeHttpIoExecutor_->getName() << "' has "
                            << exchangeHttpIoExecutor_->numThreads()
                            << " threads.";
+  for (auto evb : exchangeHttpIoExecutor_->getAllEventBases()) {
+    evb->setMaxLatency(
+        std::chrono::milliseconds(systemConfig->exchangeIoEvbViolationThresholdMs()),
+        []() { RECORD_METRIC_VALUE(kCounterExchangeIoEvbViolation, 1); },
+        false);
+  }
 
   const auto numExchangeHttpClientCpuThreads = std::max<size_t>(
       systemConfig->exchangeHttpClientNumCpuThreadsHwMultiplier() *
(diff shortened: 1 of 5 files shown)


yuandagits (approved):

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -524,6 +530,12 @@ void PrestoServer::run() {
     PRESTO_STARTUP_LOG(INFO)
         << "HTTP Server CPU executor '" << httpSrvCpuExecutor_->getName()
         << "' has " << httpSrvCpuExecutor_->numThreads() << " threads.";
+    for (auto evb : httpSrvIoExecutor_->getAllEventBases()) {
+      evb->setMaxLatency(
+          std::chrono::milliseconds(systemConfig->httpSrvIoEvbViolationThresholdMs()),
+          []() { RECORD_METRIC_VALUE(kCounterHttpServerIoEvbViolation, 1); },
+          false);
+    }
```
spershin, line 538: What is this setMaxLatency() for?
For executing http callback?
So what we doing here is setting a latency for our callbacks and if it is breached the given functor will be executed?
Do we suspect some of our callbacks talking long time on http executor?

Will this setMaxLatency() fail the http call if limit breached?

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -409,6 +409,12 @@ void PrestoServer::run() {
                            << exchangeHttpIoExecutor_->getName() << "' has "
                            << exchangeHttpIoExecutor_->numThreads()
                            << " threads.";
+  for (auto evb : exchangeHttpIoExecutor_->getAllEventBases()) {
+    evb->setMaxLatency(
```
aditi-pandit, line 413: @kewang1024 : How do we turn this off ? Since we have a platform offering we would want to make this user (our client) configurable.

@majetideepak @czentgr

steveburnett: If this is user-configurable, documentation should be included in this PR.

PR #24551, opened 2025-02-13
[Do Not Review] WIP - Mixed case identifier support

agrawalreetika:
## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* ... 
* ... 

Hive Connector Changes
* ... 
* ... 
```

```
== NO RELEASE NOTE ==
```


steveburnett (approved):
LGTM! (docs)

presto-iceberg/src/test/java/com/facebook/presto/iceberg/procedure/TestSetTablePropertyProcedure.java
```diff
@@ -160,10 +166,22 @@ public void testInvalidSetTablePropertyProcedureCases()
 
     private Table loadTable(String tableName)
     {
+        tableName = normalizeIdentifier(tableName);
         Catalog catalog = CatalogUtil.loadCatalog(HadoopCatalog.class.getName(), ICEBERG_CATALOG, getProperties(), new Configuration());
         return catalog.loadTable(TableIdentifier.of(TEST_SCHEMA, tableName));
     }
 
+    protected String normalizeIdentifier(String name)
```
ScrapCodes, line 174: May be move this function to `AbstractTestQueryFramework`

agrawalreetika: Ohh yes, I missed this in the cleanup. Thanks, will update

presto-common/src/main/java/com/facebook/presto/common/QualifiedObjectName.java
```diff
@@ -61,10 +60,7 @@ public static QualifiedObjectName valueOf(String catalogName, String schemaName,
     @ThriftConstructor
     public QualifiedObjectName(String catalogName, String schemaName, String objectName)
     {
-        checkLowerCase(catalogName, "catalogName");
-        checkLowerCase(schemaName, "schemaName");
-        checkLowerCase(objectName, "objectName");
-        this.catalogName = catalogName;
+        this.catalogName = catalogName.toLowerCase(ENGLISH);
```
ScrapCodes, line 63: I am struggling to understand, why only catalogName is lower cased and why did you remove the lower case check? Isn't it better to have a check than do always lower case.

presto-docs/src/main/sphinx/connector/mysql.rst
```diff
@@ -66,6 +66,9 @@ Property Name                                      Description
 
 ``case-insensitive-name-matching.cache-ttl``       Duration for which remote dataset and table names will be
                                                    cached. Set to ``0ms`` to disable the cache.                         ``1m``
+
+``mixed-case-support``                             Enable mixed case identifier support for schema and table names      ``false``
+                                                   for the connector
```
ScrapCodes, line 71:
```suggestion
``mixed-case-support``                             Enable mixed case identifier support for schema and table names      ``false``
                                                   for the connector. When not enabled identifiers for schema and table names  are converted to lowercase.
```

ScrapCodes: Not sure about others, to me word `mixed case` is unheard of. It may be good to include, what do we imply by mixed case support.

ScrapCodes (changes requested):
Thanks for working on this @agrawalreetika. I have got some minor nits, some are for improving my own understandings.

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/BaseJdbcClient.java
```diff
@@ -626,7 +633,7 @@ protected String getTableSchemaName(ResultSet resultSet)
     protected String toRemoteSchemaName(JdbcIdentity identity, Connection connection, String schemaName)
     {
         requireNonNull(schemaName, "schemaName is null");
-        verify(CharMatcher.forPredicate(Character::isUpperCase).matchesNoneOf(schemaName), "Expected schema name from internal metadata to be lowercase: %s", schemaName);
+//        verify(CharMatcher.forPredicate(Character::isUpperCase).matchesNoneOf(schemaName), "Expected schema name from internal metadata to be lowercase: %s", schemaName);
```
BryanCutler, line 636: remove comment?

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/BaseJdbcClient.java
```diff
@@ -671,7 +678,7 @@ protected String toRemoteTableName(JdbcIdentity identity, Connection connection,
     {
         requireNonNull(remoteSchema, "remoteSchema is null");
         requireNonNull(tableName, "tableName is null");
-        verify(CharMatcher.forPredicate(Character::isUpperCase).matchesNoneOf(tableName), "Expected table name from internal metadata to be lowercase: %s", tableName);
+//        verify(CharMatcher.forPredicate(Character::isUpperCase).matchesNoneOf(tableName), "Expected table name from internal metadata to be lowercase: %s", tableName);
```
BryanCutler, line 681: same here

presto-docs/src/main/sphinx/connector/mysql.rst
```diff
@@ -66,6 +66,9 @@ Property Name                                      Description
 
 ``case-insensitive-name-matching.cache-ttl``       Duration for which remote dataset and table names will be
                                                    cached. Set to ``0ms`` to disable the cache.                         ``1m``
+
+``mixed-case-support``                             Enable mixed case identifier support for schema and table names      ``false``
```
BryanCutler, line 70: nit: add a period at the end of description and maybe the default value should be on the last line of the description.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/procedure/TestSetTablePropertyProcedure.java
```diff
@@ -160,6 +160,7 @@ public void testInvalidSetTablePropertyProcedureCases()
 
     private Table loadTable(String tableName)
     {
+        tableName = normalizeIdentifier(tableName, ICEBERG_CATALOG);
```
BryanCutler, line 163: Could all of these similar lines be in a base class?

BryanCutler (commented):
Looks good, just a few minor comments

presto-docs/src/main/sphinx/connector/mysql.rst
agrawalreetika: Sure, I am open for suggestion here?

presto-common/src/main/java/com/facebook/presto/common/QualifiedObjectName.java
agrawalreetika: `checkLowerCase` is removed from `schemaNama` & `objectName` since that's not necessarily is lowercase now for all connector. At the places QualifiedObjectName is used, we are normalizing is based on connector.

presto-common/src/main/java/com/facebook/presto/common/CatalogSchemaName.java
```diff
@@ -34,7 +34,7 @@ public final class CatalogSchemaName
     public CatalogSchemaName(String catalogName, String schemaName)
     {
         this.catalogName = requireNonNull(catalogName, "catalogName is null").toLowerCase(ENGLISH);
-        this.schemaName = requireNonNull(schemaName, "schemaName is null").toLowerCase(ENGLISH);
+        this.schemaName = requireNonNull(schemaName, "schemaName is null");
```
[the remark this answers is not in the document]
agrawalreetika: So here I kept it original then  all the places like accesscontrol & MetadataManager where `getSchemaName` is accessed, normalized it

presto-iceberg/src/test/java/com/facebook/presto/iceberg/procedure/TestSetTablePropertyProcedure.java
agrawalreetika: You mean loadTable() method?

steveburnett (approved):
LGTM! (docs)

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/BaseJdbcConfig.java
```diff
@@ -124,4 +126,18 @@ public BaseJdbcConfig setCaseInsensitiveNameMatchingCacheTtl(Duration caseInsens
...
+    {
+        return mixedCaseSupportEnabled;
+    }
+
+    @Config("case-sensitive-name-matching")
+    @ConfigDescription("Enable case-sensitive matching of schema, table names across the connector, " +
+            "When disabled, names are matched case-insensitively, using lowercase normalization.")
+    public BaseJdbcConfig setCaseSensitiveNameMatching(boolean mixedCaseSupportEnabled)
+    {
+        this.mixedCaseSupportEnabled = mixedCaseSupportEnabled;
```
ScrapCodes, line 140:
```suggestion
    public BaseJdbcConfig setCaseSensitiveNameMatching(boolean caseSensitiveNameMatchingEnabled)
    {
        this.caseSensitiveNameMatchingEnabled = caseSensitiveNameMatchingEnabled;
```

presto-base-jdbc/src/test/java/com/facebook/presto/plugin/jdbc/TestBaseJdbcConfig.java
```diff
@@ -49,6 +50,7 @@ public void testExplicitPropertyMappings()
                 .put("password-credential-name", "bar")
                 .put("case-insensitive-name-matching", "true")
```
ScrapCodes, line 51: Now there are two configs, one is existing `case-insensitive-name-matching` and second, `case-sensitive-name-matching` that we are trying to add. Do you think we can say `case-insensitive-name-matching` when set to false is same as `case-sensitive-name-matching` set to true?

presto-common/src/main/java/com/facebook/presto/common/QualifiedObjectName.java
ScrapCodes: I am not an expert in this area, but what about `catalogName` ? can we retain the check for it?

presto-docs/src/main/sphinx/connector/mysql.rst
```diff
@@ -66,6 +66,10 @@ Property Name                                      Description
 
 ``case-insensitive-name-matching.cache-ttl``       Duration for which remote dataset and table names will be
                                                    cached. Set to ``0ms`` to disable the cache.                         ``1m``
+
+``case-sensitive-name-matching``                   Enable mixed case identifier support for schema and table names      ``false``
```
ScrapCodes, line 70:
```suggestion
``case-sensitive-name-matching``                   Enable case sensitive identifier support for schema and table names      ``false``
```

ScrapCodes (changes requested):

presto-base-jdbc/src/test/java/com/facebook/presto/plugin/jdbc/TestBaseJdbcConfig.java
agrawalreetika: Not really. Existing `case-insensitive-name-matching` has multiple limitations. So it converts all the actual schema/tables to lower case on Presto side, gives user a way to query databases that are not lower case in actual data source(like MySQL).  But this will have issue in case, suppose MySQL has (Testdb & testdb) and on Presto config we have `case-insensitive-name-matching=true` then all the queries from that catalog in Presto will start failing with below error -
```
Query 20250423_055900_00003_xdtrw failed: Failed to find remote schema name: Multiple entries with same key: testdb=testdb and testdb=TestDb

```
So `case-insensitive-name-matching=true` shows users all the data sets/tables in lower case on Presto side, could work fine if actual data source strictly doesn't have same names in difference cases.

And on the other side, setting `case-insensitive-name-matching=false` will anyways wouldn't entertain any schema/tables which are not lower case in actual data source.

agrawalreetika: Also I feel now, even though the config names sounds appropriate but renaming currrent PR config to `case-sensitive-name-matching` could be little confusing since there is an existing similar flag there. 
I think we should be eventually able to deprecate that once we have this PR changes in. 

So I think it would be better to name the current PR configuration something different and then eventually we can deprecate exiting one. Let me know your thoughts around this?

ScrapCodes: Okay, thanks for clarifying. If all that can be achieved via `case-insensitive-name-matching=true` is a subset of the current enhancement, then definitely adding a deprecation note(including the recommended way) is a way to go. Regarding config name change, the deprecation warning along with suggestion of how to achieve the same using new flag is enough.

aaneja (commented):
Flush 1 - Finished looking at code. Looking at tests next

presto-main-base/src/main/java/com/facebook/presto/metadata/CatalogMetadata.java
```diff
@@ -110,11 +111,14 @@ public ConnectorTransactionHandle getTransactionHandleFor(ConnectorId connectorI
 
     public ConnectorId getConnectorId(Session session, QualifiedObjectName table)
     {
-        if (table.getSchemaName().equals(INFORMATION_SCHEMA_NAME)) {
+        if (table.getSchemaName().equalsIgnoreCase(INFORMATION_SCHEMA_NAME)) {
```
ZacBlanco, line 114: why ignore case now?

ZacBlanco (commented):
Overall, I think this looks good, but I have two concerns

1. I am paranoid about the overhead that `normalizeIdentifiers` could introduce. There is quite a few layers of functions calls to get into a plugin's ConnectorMetadata instance before hitting `normalizeIdentifiers`, including setting up the ContextClassLoader for each metadata call. If we have a large query, or have to call many metadata functions during query planning, We might see some small performance hit on the query analysis times. I don't really know how small the impact would be, but it would be nice to verify that we don't have any regressions.

2. There are a lot of locations where we call normalizeIdentifier in the current approach. However, the source of identifiers all comes from one place: the user query. It seems like with the current solution it is possible to call `normalizeIdentifier` many times for a single identifier. Do you think it would be possible to move the normalization of identifiers closer to the query analysis phase in order to not have to run it again? Theoretically, we should only need to call it once per identifier in the query, right?

agrawalreetika:
> Overall, I think this looks good, but I have two concerns
> 
> 1. I am paranoid about the overhead that `normalizeIdentifiers` could introduce. There is quite a few layers of functions calls to get into a plugin's ConnectorMetadata instance before hitting `normalizeIdentifiers`, including setting up the ContextClassLoader for each metadata call. If we have a large query, or have to call many metadata functions during query planning, We might see some small performance hit on the query analysis times. I don't really know how small the impact would be, but it would be nice to verify that we don't have any regressions.
> 2. There are a lot of locations where we call normalizeIdentifier in the current approach. However, the source of identifiers all comes from one place: the user query. It seems like with the current solution, it is possible to call `normalizeIdentifier` many times for a single identifier. Do you think it would be possible to move the normalization of identifiers closer to the query analysis phase in order to not have to run it again? Theoretically, we should only need to call it once per identifier in the query, right?

Thanks for your review @ZacBlanco 
1. I will try to take some benchmarks with and w/o `normalizeIdentifiers` and post the results.
2. In the current state, lowercase conversion is done for user query (SELECT, DML) as well as the metadata queries. so in the current approach `normalizeIdentifiers` is called even for metadata calls (for ex - `show tables` etc) as well, to preserve the current lowercase conversion when no specific casing is done at the connector level. I think there are some places where normalizeIdentifiers call can be reduced in ConnectorMetadata where QualifiedObjectName is already normalized, I will make the changes for the same.

aaneja:
> I will try to take some benchmarks with and w/o normalizeIdentifiers and post the results.

YMMV, but adding a [session RuntimeMetric](https://github.com/prestodb/presto/blob/47a708022aba003ebfda483ea3c940231fdbd282/presto-main-base/src/main/java/com/facebook/presto/metadata/MetadataManager.java#L383) should show up the impact pretty well, both in terms of count (times called) and wall clock time

presto-main-base/src/main/java/com/facebook/presto/metadata/MetadataManager.java
```diff
@@ -547,13 +552,14 @@ public List<QualifiedObjectName> listTables(Session session, QualifiedTablePrefi
...
         if (catalog.isPresent()) {
             CatalogMetadata catalogMetadata = catalog.get();
-
             for (ConnectorId connectorId : catalogMetadata.listConnectorIds()) {
                 ConnectorMetadata metadata = catalogMetadata.getMetadataFor(connectorId);
                 ConnectorSession connectorSession = session.toConnectorSession(connectorId);
                 metadata.listTables(connectorSession, prefix.getSchemaName()).stream()
                         .map(convertFromSchemaTableName(prefix.getCatalogName()))
-                        .filter(prefix::matches)
+                        .filter(name -> prefix.matches(new QualifiedObjectName(name.getCatalogName(),
```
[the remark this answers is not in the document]
agrawalreetika: `prefix` would have `catalog.schema` mapping on which listTables needs to be called. So I think that should be passed the original value to `listTables` API here.

agrawalreetika:
@ZacBlanco @aaneja, updated based on your review comments. Please take a look at your convenience.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/procedure/TestSetTablePropertyProcedure.java
BryanCutler: You could add a method to the base class like `loadTableAndNormalizeName` that just calls `normalizeIdentifier` then `loadTable` in the subclass. You would just have to update the existing calls to `loadTable`.

presto-main-base/src/main/java/com/facebook/presto/metadata/MetadataManager.java
```diff
@@ -1479,6 +1489,23 @@ public void addConstraint(Session session, TableHandle tableHandle, TableConstra
...
+    {
+        long startTime = System.nanoTime();
+        String normalizedString = identifier.toLowerCase(ENGLISH);
+        Optional<CatalogMetadata> catalogMetadata = getOptionalCatalogMetadata(session, transactionManager, catalogName);
+        if (catalogMetadata.isPresent()) {
+            ConnectorId connectorId = catalogMetadata.get().getConnectorId();
+            ConnectorMetadata metadata = catalogMetadata.get().getMetadataFor(connectorId);
+            normalizedString = metadata.normalizeIdentifier(session.toConnectorSession(connectorId), identifier);
+            session.getRuntimeStats().addMetricValue(GET_IDENTIFIER_NORMALIZATION_TIME_NANOS, NANO, System.nanoTime() - startTime);
+            return normalizedString;
```
BryanCutler, line 1503: you could remove this return and the above session.getRuntimeStats().addMetricValue, since you are just assigning `normalizedString`

BryanCutler: Also just curious how useful this metric might be - are there some normalization that could take considerable time?

presto-iceberg/src/test/java/com/facebook/presto/iceberg/procedure/TestRemoveOrphanFilesProcedureBase.java
```diff
@@ -49,6 +49,7 @@
 import java.util.Map;
 
 import static com.facebook.presto.SystemSessionProperties.LEGACY_TIMESTAMP;
+import static com.facebook.presto.iceberg.IcebergQueryRunner.ICEBERG_CATALOG;
```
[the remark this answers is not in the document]
agrawalreetika: This is unused, will remove it. Didn't get caught in checkstyle

presto-main-base/src/main/java/com/facebook/presto/metadata/MetadataManager.java
agrawalreetika: > You could add a method to the base class like loadTableAndNormalizeName that just calls normalizeIdentifier then loadTable in the subclass. You would just have to update the existing calls to loadTable.

Regarding `loadTable` method clean up in tests, I was thinking I could do that in different PR since that change is not related? Would it be Ok?

BryanCutler: It's not a big deal, it's only a few repeated lines and I thought it would a little cleaner but either way is fine.

BryanCutler, line 1503: I still don't see why you need to return here? It looks like you `session.getRuntimeStats().addMetricValue(GET_IDENTIFIER_NORMALIZATION_TIME_NANOS, NANO, System.nanoTime() - startTime);` is the same below, so just assign `normalizedString` and let it return at the end of the method

agrawalreetika: You are right, I overlooked it. We can remove `return normalizedString;` and line before this. I will make the changes. Thanks!

BryanCutler (approved):
LGTM

presto-base-jdbc/src/test/java/com/facebook/presto/plugin/jdbc/TestBaseJdbcConfig.java
ZacBlanco: I understand the differences you pointed out between the two config keys, but from a user experience perspective this is very confusing. I think we need to make the new property name a little different or deprecate+change the name of the old one because the way it stands currently the names are too similar and it will be too easy for users to configure the wrong thing

presto-parser/src/main/java/com/facebook/presto/sql/tree/QualifiedName.java
```diff
@@ -105,6 +106,11 @@ public String getSuffix()
         return Iterables.getLast(parts);
     }
 
+    public String getOriginalSuffix()
+    {
+        return Iterables.getLast(originalParts);
```
ZacBlanco, line 111: nit: static import

presto-parser/src/main/java/com/facebook/presto/sql/tree/QualifiedName.java
```diff
@@ -81,12 +81,13 @@ public String toString()
      */
     public Optional<QualifiedName> getPrefix()
     {
-        if (parts.size() == 1) {
+        if (originalParts.size() == 1) {
```
ZacBlanco, line 84: Should we consider getting rid of parts/originalParts altogether in this PR and just accept the parts as-is from the caller? We lowercase in the constructor of this class. I feel like we already moved to prevent lowercasing elsewhere

presto-product-tests/src/main/java/com/facebook/presto/tests/hive/TestHiveMixedCaseSupport.java
```diff
@@ -0,0 +1,218 @@
...
+        query("CREATE TABLE " + SCHEMA_NAME + ".testtable (name VARCHAR(50), ID INT)");
+        query("CREATE TABLE " + SCHEMA_NAME + ".TestTable1 (Name VARCHAR(50), id INT)");
+        query("CREATE TABLE " + SCHEMA_NAME + ".TESTTABLE2 (Name VARCHAR(50), ID INT)");
+        query("CREATE TABLE " + SCHEMA_NAME_UPPER + ".TESTTABLE4 (Name VARCHAR(50), ID INT)");
+        query("CREATE TABLE " + SCHEMA_NAME_UPPER + ".testtable02 (name VARCHAR(50), id INT, num DOUBLE)");
+        query("CREATE TABLE " + SCHEMA_NAME_UPPER + ".TESTTABLE3 (name VARCHAR(50), id INT, num DOUBLE)");
+
+        // Verify table existence and column properties
+        assertThat(query("SHOW TABLES FROM " + SCHEMA_NAME))
+                .containsOnly(row("testtable0"), row("testtable"), row("testtable1"), row("testtable2"));
```
ZacBlanco, line 119: I thought we would expect table names to come back with the same casing used in creation?

presto-product-tests/src/main/java/com/facebook/presto/tests/hive/TestHiveMixedCaseSupport.java
```diff
@@ -0,0 +1,218 @@
...
+    public void testCreateSchemasWithMixedCaseNames()
+    {
+        // Define schema names and their expected stored values in Hive
+        String schemaNameMixedSyllables = "HiveMixedCaseOn";
+        String schemaNameMixed = "HiveMixedCase";
+
+        String schemaNameLowerStored = "hivemixedcaseon";
+        String schemaNameMixedSyllablesStored = "hivemixedcaseon";
+        String schemaNameMixedStored = "hivemixedcase";
+        String schemaNameUpperStored = "hivemixedcaseon1";
```
ZacBlanco, line 73: Can't we just get this from `SCHEMA_NAME_UPPER.toLowerCase()`? Same with the rest. If we expect them all to be returned with lower case, just call the lowercase method instead during the assertion. It's clearer what the test is expecting

presto-product-tests/src/main/java/com/facebook/presto/tests/hive/TestHiveMixedCaseSupport.java
```diff
@@ -0,0 +1,218 @@
...
+        query("DROP SCHEMA IF EXISTS " + SCHEMA_NAME);
+        query("DROP SCHEMA IF EXISTS " + SCHEMA_NAME_UPPER);
+    }
+
+    /*
+     * Test cases for creating schemas with different naming conventions in Hive.
+     *
+     * This class includes tests for:
+     * 1. Creating a schema with a lowercase name.
+     * 2. Creating a schema with a mixed-case name that has the same syllables as an existing schema.
```
ZacBlanco, line 57: Why are we using the term syllables here? Just say it's the same name as the existing schema but mxed case

presto-product-tests/src/main/java/com/facebook/presto/tests/hive/TestHiveMixedCaseSupport.java
```diff
@@ -0,0 +1,218 @@
...
+     * 4. Creating a schema with an uppercase name.
+     *
+     * Each test ensures the schema is created successfully.
+     */
+    @Test(groups = {MIXED_CASE})
+    public void testCreateSchemasWithMixedCaseNames()
+    {
+        // Define schema names and their expected stored values in Hive
+        String schemaNameMixedSyllables = "HiveMixedCaseOn";
+        String schemaNameMixed = "HiveMixedCase";
```
ZacBlanco, line 68: Why is this and the above separate cases? They are basically the same test case

presto-product-tests/src/main/java/com/facebook/presto/tests/hive/TestHiveMixedCaseSupport.java
```diff
@@ -0,0 +1,218 @@
...
+        assertThat(query("DESCRIBE " + SCHEMA_NAME_UPPER + ".testtable02"))
+                .contains(row("name", "varchar(50)", "", ""), row("id", "integer", "", ""), row("num", "double", "", ""));
+    }
+
+    /*
+     * This test validates inserting data into tables with different case variations in schema and table names.
+     * It ensures that data is inserted and retrieved correctly regardless of case sensitivity.
+     */
+    @Test(groups = {MIXED_CASE})
+    public void testInsertDataWithMixedCaseNames()
```
ZacBlanco, line 142: Is this test necessary? We know none of our code is touching the datapath, so mixed case would not affect rows stored in the table

presto-product-tests/src/main/java/com/facebook/presto/tests/hive/TestHiveMixedCaseSupport.java
```diff
@@ -0,0 +1,218 @@
...
+        assertThat(query("SELECT * FROM " + SCHEMA_NAME_UPPER + ".TESTTABLE3"))
+                .containsOnly(row("emma", 112, 200.002), row("mark", 123, 300.003));
+    }
+
+    /*
+     * This test verifies selecting data from tables with different case variations in schema and table names.
+     * It ensures that queries return correct results regardless of case sensitivity.
+     */
+    @Test(groups = {MIXED_CASE})
+    public void testSelectDataWithMixedCaseNames()
```
ZacBlanco, line 166: This test isn't really any different than the above test. I also think it is not necessary to check on the data path.

presto-product-tests/src/main/java/com/facebook/presto/tests/hive/TestHiveMixedCaseSupport.java
```diff
@@ -0,0 +1,218 @@
...
+    public void testTableAlterWithMixedCaseNames()
+    {
+        // Add columns to tables with various schema and table name cases
+        query("ALTER TABLE " + SCHEMA_NAME + ".testtable ADD COLUMN num REAL");
+        query("ALTER TABLE " + SCHEMA_NAME_UPPER + ".testtable02 ADD COLUMN num1 REAL");
+        query("ALTER TABLE " + SCHEMA_NAME + ".TESTTABLE2 ADD COLUMN num01 REAL");
+        query("ALTER TABLE " + SCHEMA_NAME_UPPER + ".TESTTABLE02 ADD COLUMN num2 REAL");
+
+        // Verify the added columns
+        assertThat(query("DESCRIBE hivemixedcaseon.testtable"))
```
ZacBlanco, line 199: Why is this one not using a variable for the schema?

ZacBlanco (changes requested):

presto-product-tests/src/main/java/com/facebook/presto/tests/hive/TestHiveMixedCaseSupport.java
agrawalreetika: For hive, since there is no specific connector level normalization, so it would be default to lowecase.

presto-product-tests/src/main/java/com/facebook/presto/tests/hive/TestHiveMixedCaseSupport.java
agrawalreetika: Yes, SELECT looks the same as in the INSERT one. So I think we can remove this one altogether and just keep testInsertDataWithMixedCaseNames() one as it is. WDYT?

presto-product-tests/src/main/java/com/facebook/presto/tests/hive/TestHiveMixedCaseSupport.java
NivinCS: The first test checks the behaviour when a schema with the same name already exists but in a different case,
second one checks the behaviour when we use mixed case

presto-parser/src/main/java/com/facebook/presto/sql/tree/QualifiedName.java
agrawalreetika: So QualifiedName, maintains both parts -> which convert the supplied value to lower case & originalParts which keeps the original value name of same QualifiedName. I think it should be Ok, to preserve both as of now, this PR we have used originalParts from QualifiedName in the required place to keep the case original for Schema & Table names.

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueryFramework.java
```diff
@@ -644,4 +648,15 @@ public static void dropTableIfExists(QueryRunner queryRunner, String catalogName
     {
         queryRunner.execute(format("DROP TABLE IF EXISTS %s.%s.%s", catalogName, schemaName, tableName));
     }
+
+    protected String normalizeIdentifier(String name, String catalogName)
```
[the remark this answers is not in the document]
agrawalreetika: Currenlty yes, its only used for Iceberg tests since it had mixed-case used in the tests. But as I checked, this method seems genric, if needed we could use it with other connectors as well. Do you see any problem in keeping here?

presto-main-base/src/main/java/com/facebook/presto/metadata/MetadataUtil.java
```diff
@@ -60,35 +61,30 @@ private MetadataUtil() {}
...
-    public static SchemaTableName toSchemaTableName(QualifiedObjectName qualifiedObjectName)
-    {
-        return new SchemaTableName(qualifiedObjectName.getSchemaName(), qualifiedObjectName.getObjectName());
-    }
+//    public static SchemaTableName toSchemaTableName(QualifiedObjectName qualifiedObjectName, Metadata metadata, Session session)
+//    {
+//        String schemaName = metadata.normalizeIdentifier(session, qualifiedObjectName.getCatalogName(), qualifiedObjectName.getSchemaName());
+//        String tableName = metadata.normalizeIdentifier(session, qualifiedObjectName.getCatalogName(), qualifiedObjectName.getObjectName());
+//        return toSchemaTableName(schemaName, tableName);
+//    }
```
hantangwangd, line 87: nit: remove this?

presto-main-base/src/main/java/com/facebook/presto/sql/analyzer/MaterializedViewColumnMappingExtractor.java
```diff
@@ -110,7 +113,7 @@ public MaterializedViewColumnMappingExtractor(Analysis analysis, Session session
      */
     public Map<String, Map<SchemaTableName, String>> getMaterializedViewColumnMappings()
     {
-        return getColumnMappings(analysis, mappedBaseColumns);
+        return getColumnMappings(analysis, mappedBaseColumns, metadata, session);
```
hantangwangd, line 116: Seems parameter `metadata` and `session` turns out to be useless in the end, so should we remove them?

presto-main-base/src/main/java/com/facebook/presto/sql/analyzer/MaterializedViewColumnMappingExtractor.java
```diff
@@ -222,15 +225,15 @@ private static Map<String, TableColumn> getOriginalColumnsFromAnalysis(Analysis
                         field -> new TableColumn(toSchemaTableName(field.getOriginTable().get()), field.getOriginColumnName().get(), true)));
     }
 
-    private static Map<String, Map<SchemaTableName, String>> getColumnMappings(Analysis analysis, Map<TableColumn, Set<TableColumn>> columnMappings)
+    private static Map<String, Map<SchemaTableName, String>> getColumnMappings(Analysis analysis, Map<TableColumn, Set<TableColumn>> columnMappings, Metadata metadata, Session session)
```
hantangwangd, line 228: The same as above.

presto-main-base/src/main/java/com/facebook/presto/sql/analyzer/PredicateStitcher.java
```diff
@@ -60,11 +61,13 @@ public class PredicateStitcher
...
     private final Map<SchemaTableName, Expression> predicates;
     private final Session session;
+    private final Metadata metadata;
 
-    public PredicateStitcher(Session session, Map<SchemaTableName, Expression> predicates)
+    public PredicateStitcher(Session session, Map<SchemaTableName, Expression> predicates, Metadata metadata)
     {
         this.session = requireNonNull(session, "session is null");
         this.predicates = requireNonNull(predicates, "predicates is null");
+        this.metadata = metadata;
```
hantangwangd, line 70: nit: add `requireNonNull` check for metadata?

presto-product-tests/src/main/java/com/facebook/presto/tests/hive/TestHiveMixedCaseSupport.java
```diff
@@ -0,0 +1,190 @@
...
+    private static final String SCHEMA_NAME_UPPER = "HIVEMIXEDCASEON1";
+
+    /*
+     * Cleanup test case for removing schemas and tables created during mixed-case table tests.
+     *
+     * This test ensures that all test schemas and tables are properly dropped to maintain a clean test environment.
+     */
+    @Test(groups = {MIXED_CASE})
+    public void testCleanupMixedCaseTablesAndSchemas()
+    {
```
hantangwangd, line 38: I have a question, can we ensure that all test methods in `TestHiveMixedCaseSupport` run strictly in the defined order and non-concurrently?

presto-product-tests/src/main/java/com/facebook/presto/tests/hive/TestHiveMixedCaseSupport.java
```diff
@@ -0,0 +1,190 @@
...
+        // Create schemas
+        query("CREATE SCHEMA IF NOT EXISTS " + SCHEMA_NAME);
+        query("CREATE SCHEMA IF NOT EXISTS " + schemaNameMixedSyllables);
+        query("CREATE SCHEMA IF NOT EXISTS " + schemaNameMixed);
+        query("CREATE SCHEMA " + SCHEMA_NAME_UPPER);
+
+        // Verify schema creation
+        assertThat(query("SHOW SCHEMAS"))
+                .contains(row(SCHEMA_NAME.toLowerCase(ENGLISH)))
+                .contains(row(SCHEMA_NAME.toLowerCase(ENGLISH)))
```
hantangwangd, line 80: nit: duplicate code?

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueryFramework.java
```diff
@@ -644,4 +648,15 @@ public static void dropTableIfExists(QueryRunner queryRunner, String catalogName
...
+    {
+        Metadata metadata = getQueryRunner().getMetadata();
+        TransactionId txid = getQueryRunner().getTransactionManager().beginTransaction(false);
+        Session session = getSession().beginTransactionId(txid, getQueryRunner().getTransactionManager(), new AllowAllAccessControl());
+        CatalogManager catalogManager = getDistributedQueryRunner().getCoordinator().getCatalogManager();
+        ConnectorId connectorId = catalogManager.getCatalog(catalogName).get().getConnectorId();
+
+        return metadata.normalizeIdentifier(session, connectorId.getCatalogName(), name);
```
hantangwangd, line 660:
```suggestion
        String result = metadata.normalizeIdentifier(session, catalogName, name);
        getQueryRunner().getTransactionManager().asyncAbort(txid);
        return result;
```

As I understand, `catalog.getConnectorId().getCatalogName()` is always the same as `catalogName`, so do you think it makes sense to use `catalogName` here directly? 

Besides, is it better to abort the transaction explicitly? Since the transaction manager in query runner is a long lived instance through the whole test class scope, and I observed nearly one hundred additional un-closed transactions after this change when running `TestIcebergDistributedHive`.

agrawalreetika: Thanks for pointing it out. I have made the changes.

presto-product-tests/src/main/java/com/facebook/presto/tests/hive/TestHiveMixedCaseSupport.java
agrawalreetika: Good question! I looked into the tempto-configuration.yaml ([link](https://github.com/prestodb/presto/blob/master/presto-product-tests/src/main/resources/tempto-configuration.yaml)) and didn't find any specific configuration enforcing parallel or sequential execution.

Regarding test execution order: from what I understand, TestNG defaults to executing test methods alphabetically if no specific order is defined, and in this class, the method names appear to follow alphabetical order—possibly explaining the current behavior. I'll try renaming a few methods to confirm if the order changes in the CI logs.

If you're aware of any additional ways to check method ordering and non-concurrency, I'm happy to explore further. Let me know your thoughts!

aaneja: You can annotate the class with `@Test(singleThreaded = true)`

aaneja: Btw this is a teardown method, so lets decorate it with `@AfterClass(alwaysRun = true)` and not make it a `@Test` ?

aaneja (changes requested):

steveburnett (dismissed):
LGTM! (docs) 

presto-product-tests/src/main/java/com/facebook/presto/tests/hive/TestHiveMixedCaseSupport.java
hantangwangd: After checking the document of TestNG, I found that it do not guarantee the order of invocation for methods that belong in the same group. See [here](https://testng.org/#_dependencies_with_annotations), the original words are:

```
as stated before, the order of invocation for methods that belong in the same group is not guaranteed to be the same across test runs. 
```

So, the "alphabetical execution order" seems more likely to be just the result of the specific operating system and JVM happens to return the methods in that order, and TestNG itself does not explicitly shuffle or reorder them. Considering this, it seems that we cannot rely on this default invocation sequence. IMO, we can use the `dependsOnMethods` to explicitly define the invocation sequence, or use one test to calls the methods in the order we want, as suggested by @aaneja. What's your opinion?

hantangwangd (dismissed):
Thanks for the whole work, looks good to me!

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/BaseJdbcConfig.java
```diff
@@ -124,4 +136,27 @@ public BaseJdbcConfig setCaseInsensitiveNameMatchingCacheTtl(Duration caseInsens
...
+    public BaseJdbcConfig setCaseSensitiveNameMatching(boolean caseSensitiveNameMatchingEnabled)
+    {
+        this.caseSensitiveNameMatchingEnabled = caseSensitiveNameMatchingEnabled;
+        return this;
+    }
+
+    @PostConstruct
+    public void validateConfig()
+    {
+        if (isCaseInsensitiveNameMatching() && isCaseSensitiveNameMatching()) {
```
aaneja, line 157: Both being `false` is also invalid state ?

agrawalreetika: Not really, that would default to the current behaviour of Presto. Converting everything to lowe case

aaneja (approved):

presto-product-tests/src/main/java/com/facebook/presto/tests/hive/TestHiveMixedCaseSupport.java
ZacBlanco: ping. I also think we should test with mixed case schema and there to make sure that `INSERT` statements on cased when we don't quote them. e.g.

```SQL
-- schema names
INSERT INTO  "HiveMixedCaseOn".testtable VALUES ('amy', 112), ('mia', 123);
INSERT INTO  HiveMixedCaseOn.testtable VALUES ('amy', 112), ('mia', 123);
INSERT INTO  hivemixedcaseon.testtable VALUES ('amy', 112), ('mia', 123);
INSERT INTO  "hivemixedcaseon".testtable VALUES ('amy', 112), ('mia', 123);

-- table names
INSERT INTO  hivemixedcaseon.testtable VALUES ('amy', 112), ('mia', 123);
INSERT INTO  hivemixedcaseon.TestTable VALUES ('amy', 112), ('mia', 123);
INSERT INTO  hivemixedcaseon."testtable" VALUES ('amy', 112), ('mia', 123);
INSERT INTO  hivemixedcaseon."TestTable" VALUES ('amy', 112), ('mia', 123);
```

ZacBlanco (commented):
The code looks good for the most part. Just want to make sure our tests are comprehensive

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/BaseJdbcConfig.java
```diff
@@ -99,12 +105,18 @@ public BaseJdbcConfig setPasswordCredentialName(String passwordCredentialName)
         return this;
     }
 
+    @Deprecated
```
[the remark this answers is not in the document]
agrawalreetika: @ZacBlanco I had used `@Deprecated`  to indicate that in the future this may be removed. Also, currently Deprecated `case-insensitive-name-matching` is still present, and I think adding `@LegacyConfig` would fit in if we are changing just property names from earlier to now? And maybe once the `case-insensitive-name-matching` flag is removed, then we can add `@LegacyConfig`to this new property config.

ScrapCodes (approved):

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/BaseJdbcClient.java
```diff
@@ -673,9 +681,10 @@ protected String toRemoteTableName(JdbcIdentity identity, Connection connection,
     {
         requireNonNull(remoteSchema, "remoteSchema is null");
         requireNonNull(tableName, "tableName is null");
-        verify(CharMatcher.forPredicate(Character::isUpperCase).matchesNoneOf(tableName), "Expected table name from internal metadata to be lowercase: %s", tableName);
 
         if (caseInsensitiveNameMatching) {
+            log.warn("The 'case-insensitive-name-matching' config is deprecated and may lead to query issues. " +
+                    "Please use 'case-sensitive-name-matching=true' for proper case handling.");
```
ZacBlanco, line 687: At what frequency is this logged? Every query? I wonder if we could instead add this to the session's warning collector.

ZacBlanco: I think we can also more clearly steer users away from this config by more clearly warning them about the consequences of using this configuration

Consider this message instead:

> 'case-insensitive-name-matching' is deprecated. Use of this configuration value may lead to inconsistent behavior for schema, table, and column name matching. Please switch to using 'case-sensitive-name-matching' for proper case sensitivity behavior.

Also, in this block if `case-insensitive-name-matching` is true, wouldn't 'case-sensitive-name-matching=true` as the recommendation be the opposite of what the user is trying to configure?

ZacBlanco (approved):
LGTM, thanks for this big effort! I have one remaining request about the warnings, otherwise I am happy with this.

ZacBlanco (dismissed):

hantangwangd (dismissed):

hantangwangd (approved):

PR #24567, opened 2025-02-14
Remove synchronized access for PlanFragment.bytesForTaskSerialization()

NikhilCollooru:
## Description
Remove synchronized access for PlanFragment.bytesForTaskSerialization() to reduce contention

## Motivation and Context
During lock profiling when coordinator is running hot, we observed contention for PlanFragment.bytesForTaskSerialization(). This method does not need synchronized access. byte[] cachedSerialization is gated by " @GuardedBy("this")" so its thread safe. Also once the serialized plan is cached, making only one thread access it a time makes no sense. 
langThread run (100)

## Impact
None

## Test Plan
existing unit tests and verifier testing

## Contributor checklist

- [ ] CI passed.

## Release Notes
```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/sql/planner/PlanFragment.java
+++ b/presto-main/src/main/java/com/facebook/presto/sql/planner/PlanFragment.java
@@ -165,7 +165,7 @@ public class PlanFragment
 
     // Serialize this plan fragment with the provided codec, caching the results
     // This should be used when serializing the fragment to send to worker nodes.
-    public synchronized byte[] bytesForTaskSerialization(Codec<PlanFragment> codec)
+    public byte[] bytesForTaskSerialization(Codec<PlanFragment> codec)
     {
         requireNonNull(codec, "codec is null");
         if (cachedSerialization != null) {


presto-main/src/main/java/com/facebook/presto/sql/planner/PlanFragment.java
```diff
@@ -165,7 +165,7 @@ public Optional<String> getJsonRepresentation()
 
     // Serialize this plan fragment with the provided codec, caching the results
     // This should be used when serializing the fragment to send to worker nodes.
-    public synchronized byte[] bytesForTaskSerialization(Codec<PlanFragment> codec)
+    public byte[] bytesForTaskSerialization(Codec<PlanFragment> codec)
```
arhimondr, line 168: I wonder whether it is better to allow more than one serializer to run concurrently or having a single serializer do the work and let the other consumers wait for it?

NikhilCollooru: I dont have strong opinion here. But since this method is synchronized, just accessing the `cachedSerialization` bytes is restricted.. which is a problem. Like once the serialized bytes are ready/cached...access to it should be non-synchronized.

arhimondr: That is fair. Ideally multiple task management threads shouldn't be waiting for serialization. What do you think about serializing PlanFragment during planning and then simply read it when scheduling a task?

PR #24569, opened 2025-02-14
[native] Enable node pool type specification when reporting to the coordinator from a C++ worker 

jaystarshot:
## Description

Presto supports splitting workers into leaf and non leaf pools. This was added [here](https://github.com/prestodb/presto/commit/88c548544df80a8d97e60349e2195f42cdea6706).

For this the coordinator needs `worker-isolation-enabled=true` and workers can report their pool_type to the resource manager.

Adding the same configs in presto c++ so that the announcer can report the node type. 

With this (since support was already added in coordinator) one can deploy any combination of native/java workers in LEAF/INTERMEDIATE etc.

## Test Plan

Tested that setting works and if we set pool_type=LEAF in presto c++ the coordinator will only schedule leaf tasks there

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Enable node pool type specification when reporting to the coordinator from a C++ worker.

### DIFF
--- a/presto-native-execution/presto_cpp/main/Announcer.h
+++ b/presto-native-execution/presto_cpp/main/Announcer.h
@@ -31,6 +31,7 @@ class Announcer : public PeriodicServiceInventoryManager {
       const std::string& environment,
       const std::string& nodeId,
       const std::string& nodeLocation,
+      const std::string& nodePoolType,
       const bool sidecar,
       const std::vector<std::string>& connectorIds,
       const uint64_t maxFrequencyMs_,
--- a/presto-native-execution/presto_cpp/main/PrestoServer.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoServer.cpp
@@ -241,6 +241,7 @@ void PrestoServer::run() {
       address_ = fmt::format("[{}]", address_);
     }
     nodeLocation_ = nodeConfig->nodeLocation();
+    nodePoolType_ = systemConfig->poolType();
     prestoBuiltinFunctionPrefix_ = systemConfig->prestoDefaultNamespacePrefix();
   } catch (const velox::VeloxUserError& e) {
     PRESTO_STARTUP_LOG(ERROR) << "Failed to start server due to " << e.what();
@@ -576,6 +577,7 @@ void PrestoServer::run() {
           environment_,
           nodeId_,
           nodeLocation_,
+          nodePoolType_,
           systemConfig->prestoNativeSidecar(),
           catalogNames,
           systemConfig->announcementMaxFrequencyMs(),
--- a/presto-native-execution/presto_cpp/main/common/Configs.cpp
+++ b/presto-native-execution/presto_cpp/main/common/Configs.cpp
@@ -240,6 +240,7 @@ SystemConfig::SystemConfig() {
           BOOL_PROP(kEnableRuntimeMetricsCollection, false),
           BOOL_PROP(kPlanValidatorFailOnNestedLoopJoin, false),
           STR_PROP(kPrestoDefaultNamespacePrefix, "presto.default"),
+          STR_PROP(kPoolType, "DEFAULT"),
       };
 }
 
@@ -288,6 +289,17 @@ folly::Optional<std::string> SystemConfig::httpsClientCertAndKeyPath() const {
 
 std::string SystemConfig::prestoVersion() const {
   return requiredProperty(std::string(kPrestoVersion));
+}
+
+std::string SystemConfig::poolType() const {
+    static const std::unordered_set<std::string> kPoolTypes = {"LEAF", "INTERMEDIATE", "DEFAULT"};
+    static constexpr std::string_view kPoolTypeDefault = "DEFAULT";
+    auto value = optionalProperty<std::string>(kPoolType).value_or(std::string(kPoolTypeDefault));
+    VELOX_USER_CHECK(
+        kPoolTypes.count(value),
+        "{} must be one of 'LEAF', 'INTERMEDIATE', or 'DEFAULT'",
+        kPoolType);
+    return value;
 }
 
 bool SystemConfig::mutableConfig() const {
--- a/presto-native-execution/presto_cpp/main/common/Configs.h
+++ b/presto-native-execution/presto_cpp/main/common/Configs.h
@@ -659,6 +659,9 @@ class SystemConfig : public ConfigBase {
   static constexpr std::string_view kPrestoDefaultNamespacePrefix{
       "presto.default-namespace"};
 
+  // Specifies the type of worker pool
+  static constexpr std::string_view kPoolType{"pool-type"};
+
   SystemConfig();
 
   virtual ~SystemConfig() = default;
@@ -898,6 +901,8 @@ class SystemConfig : public ConfigBase {
 
   bool prestoNativeSidecar() const;
   std::string prestoDefaultNamespacePrefix() const;
+
+  std::string poolType() const;
 };
 
 /// Provides access to node properties defined in node.properties file.
--- a/presto-native-execution/presto_cpp/main/Announcer.cpp
+++ b/presto-native-execution/presto_cpp/main/Announcer.cpp
@@ -30,6 +30,7 @@ std::string announcementBody(
     const std::string& nodeVersion,
     const std::string& environment,
     const std::string& nodeLocation,
+    const std::string& nodePoolType,
     const bool sidecar,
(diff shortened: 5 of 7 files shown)

### CODE REVIEW

**Comment by steveburnett** [comment]

```

:pr:`12345`

```

**Comment by aditi-pandit** [comment]
@jaystarshot : Is the node pool type an optimization or it affects functionality as well ? So if a non-leaf worker received a TableScan would it throw error ?

**Comment by jaystarshot** [comment]
@aditi-pandit The coordinator won’t send a non leaf fragment. If it does i believe the execution should be normal since there is no blocking logic in the worker

**Comment by jaystarshot** [comment]
@majetideepak  can you please take a look

File: presto-native-execution/presto_cpp/main/PrestoServer.cpp [hunk 1]
```diff
@@ -241,6 +241,7 @@ void PrestoServer::run() {
       address_ = fmt::format("[{}]", address_);
     }
     nodeLocation_ = nodeConfig->nodeLocation();
+    nodePoolType_ = systemConfig->poolType();
```
aditi-pandit, line 244: What are the possible values for pool Type ? Might be good to validate the value.

presto-native-execution/presto_cpp/main/common/Configs.h
```diff
@@ -659,6 +659,9 @@ class SystemConfig : public ConfigBase {
   static constexpr std::string_view kPrestoDefaultNamespacePrefix{
       "presto.default-namespace"};
 
+  // Specifies the type of worker pool
+  static constexpr std::string_view kPoolType{"pool-type"};
```
aditi-pandit, line 663: Doesn't seem like there is perfect consistency in the naming of config, but we could consider a 2 part name.. announcement.pool-type ? There are other announcement config .

presto-native-execution/presto_cpp/main/common/Configs.cpp
```diff
@@ -290,6 +290,12 @@ std::string SystemConfig::prestoVersion() const {
   return requiredProperty(std::string(kPrestoVersion));
 }
 
+std::string SystemConfig::poolType() const {
```
aditi-pandit, line 293: Wanted to confirm that this is not a registered property... Else it has to be a part of registeredProps_. If you add it to registeredProps_ it might be simpler to initialize the default value as well.

presto-native-execution/presto_cpp/main/common/Configs.h
```diff
@@ -898,6 +901,8 @@ class SystemConfig : public ConfigBase {
 
   bool prestoNativeSidecar() const;
   std::string prestoDefaultNamespacePrefix() const;
+
+  std::string poolType() const;
```
aditi-pandit, line 905: Another question... This config could be a good fit for node.config instead of system config ? Is there a reason for not considering that option ?

aditi-pandit (commented):
@jaystarshot : Just had 2 questions else this PR should be fine.

presto-native-execution/presto_cpp/main/common/Configs.cpp
jaystarshot: What is the difference between registeredProps_ and configs?

presto-native-execution/presto_cpp/main/PrestoServer.cpp
jaystarshot: The accepted values are [here](https://github.com/prestodb/presto/blob/a766a3a1ecd5fa45426977bc05e707da5d1c138c/presto-spi/src/main/java/com/facebook/presto/spi/NodePoolType.java#L16) Sure I can add that

jaystarshot:
@aditi-pandit I am just using the same property name as presto java uses. Its defined in serverconfig [here](https://github.com/prestodb/presto/blob/a766a3a1ecd5fa45426977bc05e707da5d1c138c/presto-main/src/main/java/com/facebook/presto/server/ServerConfig.java#L203). The accepted values are

aditi-pandit:
> @aditi-pandit I am just using the same property name as presto java uses. Its defined in serverconfig [here](https://github.com/prestodb/presto/blob/a766a3a1ecd5fa45426977bc05e707da5d1c138c/presto-main/src/main/java/com/facebook/presto/server/ServerConfig.java#L203).

Thanks @jaystarshot for the explanation. I saw the linked PR later. Yeah, we'll have to stick with this name.

presto-native-execution/presto_cpp/main/common/Configs.cpp
aditi-pandit: Registered properties are validated on server startup https://github.com/prestodb/presto/blob/master/presto-native-execution/presto_cpp/main/common/Configs.cpp#L110

Might be better to add it to avoid warnings.

aditi-pandit:
@pdabre12 @pramodsatya : Can you'll take a look at the failures in [prestocpp-linux-build-and-unit-test / prestocpp-linux-presto-sidecar-tests (pull_request)](https://github.com/prestodb/presto/actions/runs/13440743390/job/37555161792?pr=24569)

steveburnett:
Suggest rephrasing of release note entry to follow the [Order of changes](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines#order-of-changes), and [Order of sections](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines#order-of-sections) grouping in the Release Notes Guidelines: 

```
== RELEASE NOTES ==

* Add node pool type specification in Presto C++ for reporting to the coordinator from a C++ worker. 
```

presto-native-execution/presto_cpp/main/common/Configs.cpp
```diff
@@ -290,6 +291,16 @@ std::string SystemConfig::prestoVersion() const {
...
+    static const std::unordered_set<std::string> validTypes = {"LEAF", "INTERMEDIATE", "DEFAULT"};
+    auto value = requiredProperty(std::string(kPoolType));
```
aditi-pandit, line 296: Should this be optionalProperty instead ? Configs which don't have pool-type will cause failures post this change. So this could be very disruptive. Could that be the reason for all the test failures ? I haven't looked at the details but the timeouts most likely could be because the workers didn't start up.

presto-native-execution/presto_cpp/main/common/Configs.cpp
```diff
@@ -290,6 +291,16 @@ std::string SystemConfig::prestoVersion() const {
   return requiredProperty(std::string(kPrestoVersion));
 }
 
+std::string SystemConfig::poolType() const {
+    static const std::unordered_set<std::string> validTypes = {"LEAF", "INTERMEDIATE", "DEFAULT"};
```
aditi-pandit, line 295: Nit rename variable to kPoolTypes. 

Constant variable names should be prefixed with 'k'

presto-native-execution/presto_cpp/main/Announcer.h
```diff
@@ -34,6 +34,7 @@ class Announcer : public PeriodicServiceInventoryManager {
       const bool sidecar,
       const std::vector<std::string>& connectorIds,
       const uint64_t maxFrequencyMs_,
+      const std::string& nodePoolType,
```
aditi-pandit, line 37: We could move this parameter higher up since there aren't any optional parameters in this method. After nodeLocation or sidecar ?

aditi-pandit (commented):
@jaystarshot : Have one major comment about the required vs optional. The rest are nits.

Another thing : Please change your commit to add a prefix of [native]

presto-native-execution/presto_cpp/main/common/Configs.cpp
jaystarshot: i thought the default will always be initialized when we put this in the registeredProps_ map i can try with optional

presto-native-execution/presto_cpp/main/common/Configs.cpp
```diff
@@ -290,6 +291,17 @@ std::string SystemConfig::prestoVersion() const {
...
```
aditi-pandit, line 295: Nit: Rename this set of constants with k Prefix as well. kPoolTypes maybe

aditi-pandit (commented):
@jaystarshot : Please change your commit title to have prefix [native] as well.

One minor nit otherwise.

aditi-pandit (approved):
Thanks for the iterations @jaystarshot

aditi-pandit (approved):
Thanks for the test change @jaystarshot

Issue #23150, opened 2024-07-09
Scalar functions throw exceptions even if wrapped in try()


kagamiori:
Some functions throw exceptions on invalid argument even if they are used inside try() that is supposed to suppress exceptions. Below is a list of these functions and the example queries:
| Functions | Example queries |
|-----------|------------------|
| contains | select try(contains(c0, c1)) from (values (array[array[1], array[null]], array[null])) t(c0, c1) |
| array_distinct | select try(array_distinct(c0)) from (values (array[array[1, null], array[1, null]])) t(c0) |
| array_remove | select try(array_remove(c0, c1)) from (values (array[array[1], array[null]], array[null])) t(c0, c1) |
| array_union | select try(array_union(c0, c1)) from (values (array[array[1], array[null]], array[array[null]])) t(c0, c1) |
| array_min/array_max | select try(array_max(c0)) from (values (array[array[1], array[null]], array[null])) t(c0, c1) |
| array_position | select try(array_position(c0, c1)) from (values (array[array[1], array[null]], array[null])) t(c0, c1) |
| transform_keys | select try(transform_keys(c0, (k, v) -> array[null])) from (values (map(array[1, 2], array[3, 4]))) t(c0) |

## Your Environment

* Presto version used: 0.288
* Storage (HDFS/S3/GCS..):
* Data source and connector used:
* Deployment (Cloud or On-prem):
* [Pastebin](https://pastebin.com/) link to the complete debug logs:

## Expected Behavior

I expect these queries to return NULL instead of throwing.

## Current Behavior

These queries throw exceptions.

## Steps to Reproduce

1.
2.
3.
4.

kagamiori:
cc @mbasmanova @amitkdutta

kagamiori:
I found that these functions do not always throw when input elements are arrays containing NULLs. It's likely because the function implementation compares the inner array sizes first and hence avoid comparing elements in the inner arrays. This, however, brings inconsistency as the implementation should be transparent to users.

| Functions | Example query |
|-----------|----------------|
| contains | select contains(c0, c1) from (values (array[array[1], array[1, null]], array[2])) t(c0, c1) |
| array_remove | select array_remove(c0, c1) from (values (array[array[1], array[1, null]], array[1])) t(c0, c1) |
| array_distinct | select array_distinct(c0) from (values (array[array[1], array[1, null]])) t(c0) |
[...]

PR #24570, opened 2025-02-15
Update the error type for ARRAY_NULL_ELEMENT_MSG

duxiao1212:
## Description
Update the error type for ARRAY_NULL_ELEMENT_MSG to INVALID_FUNCTION_ARGUMENT in order to suppress the error for try(map_keys_by_top_n_values ...) when the value is an array with null.

This issue is also a continuation of our previous discussion, which can be found here: https://github.com/prestodb/presto/issues/23150#issuecomment-2261057958

## Motivation and Context
Addressing the discrepancy with the try function related to map_keys_by_top_n_values. In Presto, we throw a NOT_SUPPORTED error (20250214_205756_02064_7wui9) 
image
Whereas in Prestissimo, the error is suppressed, and null is returned.
The reason the try block isn't catching these errors is that Presto is using an incorrect error code when throwing exceptions for those functions. Specifically, it is throwing NOT_SUPPORTED errors instead of INVALID_FUNCTION_ARGUMENT errors. The try code does not catch NOT_SUPPORTED errors, and the try function does not catch these errors as shown in [TryFunction.java from lines 166 to 170](https://github.com/prestodb/presto/blob/master/presto-main/src/main/java/com/facebook/presto/operator/scalar/TryFunction.java#L166-L170).

## Impact
Low impact

## Contributor checklist

- [] CI passed.

## Release Notes

```

== NO RELEASE NOTE ==

### DIFF
--- a/presto-common/src/main/java/com/facebook/presto/common/type/TypeUtils.java
+++ b/presto-common/src/main/java/com/facebook/presto/common/type/TypeUtils.java
@@ -13,7 +13,7 @@
  */
 package com.facebook.presto.common.type;
 
-import com.facebook.presto.common.NotSupportedException;
+import com.facebook.presto.common.InvalidFunctionArgumentException;
 import com.facebook.presto.common.block.Block;
 import com.facebook.presto.common.block.BlockBuilder;
 import io.airlift.slice.Slice;
@@ -189,7 +189,7 @@ public final class TypeUtils
     static void checkElementNotNull(boolean isNull, String errorMsg)
     {
         if (isNull) {
-            throw new NotSupportedException(errorMsg);
+            throw new InvalidFunctionArgumentException(errorMsg);
         }
     }
 
--- a/presto-main/src/test/java/com/facebook/presto/type/TestArrayOperators.java
+++ b/presto-main/src/test/java/com/facebook/presto/type/TestArrayOperators.java
@@ -710,13 +710,13 @@ public class TestArrayOperators
     @Test
     public void testArrayMinWithNullInFirstArrayIsCompared()
     {
-        assertInvalidFunction("ARRAY_MIN(ARRAY [ARRAY[1, NULL], ARRAY[1, 2]])", NOT_SUPPORTED);
+        assertInvalidFunction("ARRAY_MIN(ARRAY [ARRAY[1, NULL], ARRAY[1, 2]])", INVALID_FUNCTION_ARGUMENT);
     }
 
     @Test
     public void testArrayMinWithNullInSecondArrayIsCompared()
     {
-        assertInvalidFunction("ARRAY_MIN(ARRAY [ARRAY[1, 2], ARRAY[1, NULL]])", NOT_SUPPORTED);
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: presto-common/src/main/java/com/facebook/presto/common/type/TypeUtils.java [hunk 1]
```diff
@@ -189,7 +191,7 @@ public static boolean isFloatingPointNaN(Type type, Object value)
     static void checkElementNotNull(boolean isNull, String errorMsg)
     {
         if (isNull) {
-            throw new NotSupportedException(errorMsg);
+            throw ARRAY_NULL_ELEMENT_MSG.equals(errorMsg) ? new InvalidFunctionArgumentException(errorMsg) :  new NotSupportedException(errorMsg);
```
zacw7, line 194: In what scenario it throws not supported? Does it align with velox? Thanks

duxiao1212: Great question, I'm not sure how velox determines which error should be NotSupportedException, but I think we can fix it later if there's a discrepancy. What's your opinion, @kagamiori

rschlussel (changes requested):
This seems like a bit of a hacky solution. Can we instead have these checks throw INVALID_FUNCTION_ARGUMENT errors for all cases?  These checks are only used for different array and map functions. and even if there were other cases, INVALID_FUNCTION_ARGUMENT seems a reasonable enough error code.

presto-main/src/test/java/com/facebook/presto/type/TestArrayOperators.java
```diff
@@ -1145,7 +1145,7 @@ public void testSort()
         assertInvalidFunction(
                 "ARRAY_SORT(ARRAY[ARRAY[1], ARRAY[null]])",
                 INVALID_FUNCTION_ARGUMENT,
-                "Array contains elements not supported for comparison");
+                "ARRAY comparison not supported for arrays with null elements");
```
rschlussel, line 1148: why did the error message change here?

rschlussel (commented):
Changes look good. Can you fix the commit message to match these guidelines https://cbea.ms/git-commit/

presto-main/src/test/java/com/facebook/presto/type/TestArrayOperators.java
duxiao1212: The error message "Array contains elements not supported for comparison" is actually re-thrown in https://github.com/prestodb/presto/blob/4a0bbc9bbb22362a5c60ccf22b50c3a69058fd58/presto-main/src/main/java/com/facebook/presto/operator/scalar/ArraySortFunction.java#L81 ;

rschlussel: got it. normally i'd suggest catching the INVALID_FUNCTION_ARGUMENT there to keep the behavior the same, but the new error message seems better to me.

duxiao1212:
> Changes look good. Can you fix the commit message to match these guidelines https://cbea.ms/git-commit/

Updated, thanks.

Issue #24435, opened 2025-01-27
Upgrade to Hive 4.0.x


imjalpreet:
### Description
This issue proposes upgrading the Hive dependency in Presto to version 4.0.0 or 4.0.1 to leverage the latest improvements, security fixes, and features provided by Apache Hive.

### Background
The current Hive version used in Presto is 3.0.0 and may lack:

- **Bug fixes**: Addressing known issues in older versions.
- **Performance enhancements**: Optimizations introduced in Hive 4.x.
- **Security Updates**: Updated dependencies within Hive that address known vulnerabilities in older versions.
- **Compatibility**: Staying aligned with modern Hadoop ecosystems and tools.

### Challenges and Considerations
- **Backward Compatibility**: Hive 4.0.x may introduce breaking changes or deprecated APIs that could affect existing Presto integrations
- **Dependency Conflicts**: Analyze transitive dependencies for potential conflicts with Presto’s existing libraries.
- **Runtime Testing**: Perform runtime tests with different Hive metastore versions and configurations.

### Updates from the Hive Community

https://hive.apache.org/general/downloads/

- The Apache Hive Community has voted to declare the Hive 1.x, Hive 2.x and Hive 3.x versions as End of Life in 2024. 
- Hive 4.0.0 and Hive 4.0.1 released in 2024

imjalpreet:
Some observations based on my investigations till now:

**Applicable to both Hive 4.0.0 and Hive 4.0.1**

- I noticed that hive-dwrf (see [here](https://github.com/prestodb/presto-hive-dwrf/blob/b93e2e025b031fe21bce1c0a57be0983d55a2df4/pom.xml#L72)) still relies on a very old version of presto-hive-apache (v0.8). There have been breaking changes in the SerDe, Serializer, and Deserializer classes in Hive, which have made the current implementation of ORCSerde (in DWRF) incompatible with Hive 4.0.x. As a result, I am investigating the best approach to address this issue.
-  I encountered some breaking changes in the ThriftHiveMetastore.Client.
    -  Specifically, `delete_table_column_statistics` and `delete_partition_column_statistics` APIs require a new argument engine to be passed in the API call. This is because of Hive 4.0.x - Hive differentiates among column stats computed by different engines.

**Applicable only to Hive 4.0.0**
[...]

PR #24571, opened 2025-02-16
feat: Upgrade Hive to 4.0.1 and align timestamp behaviour with SQL standard

imjalpreet:
## Description

Upgrade to Hive 4.0.1 and align timestamp behaviour with SQL standard

Depends on https://github.com/prestodb/presto-hive-apache/pull/65, https://github.com/prestodb/tempto/pull/299, https://github.com/prestodb/orc-protobuf/pull/25, https://github.com/prestodb/presto-hive-dwrf/pull/13, https://github.com/prestodb/presto-hive-dwrf/pull/14 and https://github.com/prestodb/docker-images/pull/58

## Motivation and Context

#24435 and #27028

## Summary by Sourcery

Upgrade the Hive/ORC/Parquet stack and default timestamp semantics, wiring session-aware timezone handling through readers, writers, connectors, and tests.

New Features:
- Add support for writer and stripe timezones in ORC and Parquet metadata and readers, enabling standard-compliant timestamp handling.
- Expose connector session timestamp semantics to Hive, Cassandra, Iceberg, Hudi, and Parquet page sources and sinks for consistent behavior across connectors.

Bug Fixes:
- Align timestamp, date, and string statistics handling with upstream ORC/Parquet/Hive behavior to avoid incorrect predicate evaluation and metadata validation failures.
- Fix various tests and CI flows that depended on legacy timestamp semantics, incorrect timezones, or outdated expectations.

Enhancements:
- Upgrade Hive, ORC, DWRF, and Tempto dependencies and remove obsolete components such as hive-llap-common and PrestoAvroSerDe.
- Refine selective/batch readers and writers in ORC and Parquet to be timezone-aware and to honor microsecond-precision timestamps where applicable.
- Simplify S3 Select split schema handling by relying on LIST_COLUMNS/LIST_COLUMN_TYPES instead of Thrift DDL mangling.
- Improve materialized view and partition statistics handling by using session-aware partition value parsing and timestamp statistics computation.
- Adjust default function configuration to disable legacy timestamp semantics and update JSON/timestamp casts accordingly.

Build:
- Bump hive-apache, hive-dwrf, orc-protobuf, tempto, and related dependency versions and clean up transitive dependencies in Parquet, Iceberg, Pinot, and Hive function namespace modules.

CI:
- Relax product test workflow conditions to always run specific job groups when there are code changes, improving signal for storage-format and connector tests.

Tests:
- Update Hive, ORC, Parquet, Cassandra, Kafka, Iceberg, Druid, and expression engine tests to use the new timestamp semantics, UTC-based expectations, and session-aware timezone handling.
- Add Hive 4 insert-overwrite test coverage and adjust various smoke and integration tests to reflect upgraded Hive images and behavior.

Chores:
- Switch Docker-based Hive images to newer Hive 3/4 variants and adjust MinIO/HDFS wiring for tests and Presto-on-Spark environments.

## Sourcery Reviewer's Guide

https://github.com/prestodb/presto/pull/24571#issuecomment-4464222430

## Impact

Default value of `deprecated.legacy-timestamp` changed from `true` to `false`.

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Fix timestamp operations to match the SQL specification. The value of a TIMESTAMP type is not affected by the session time zone.
* Update default value of `deprecated.legacy-timestamp` to false.

Hive Connector Changes
* Upgrade to Hive 4.0.1
```


ScrapCodes:
Looking closely at the reason why tests are failing, beginning with presto-orc module.
Progress so far:
* Tests try to write content with hive and read by presto, especially in case of timestamps, it is incorrectly read by presto. Why exactly, I am still trying to understand - not clear yet.
The contents written by tests is correct because on reading with an external orc reader like my following python script, it gives correct result:
```py
from pyarrow import orc

table2 = orc.read_table('/tmp/3420396529049254202/data.orc')
print(table2)
```
```
python read_orc.py 
pyarrow.Table
test: timestamp[ns]
----
test: [[1970-01-01 00:00:00.001000000,1970-01-01 00:00:00.003000000,1970-01-01 00:00:00.005000000,1970-01-01 00:00:00.007000000,1970-01-01 00:00:00.011000000,...,1970-01-01 00:00:00.001000000,1970-01-01 00:00:00.003000000,1970-01-01 00:00:00.005000000,1970-01-01 00:00:00.007000000,1970-01-01 00:00:00.011000000]]

```

Whereas, record reader for ORC in presto reads it differently
Screenshot from 2025-04-28 17-35-06

* These values e.g. `21600001` when converted to timestamp values give a 6h difference. e.g. 

Screenshot from 2025-04-28 17-47-33
This is the reason for failure of these tests.

ScrapCodes:
Another interesting find:
Trying to read directly by hive version 3.0.0, orc file produced by presto without hive 4 changes. Produces incorrect results, whereas it produces correct results when reading a file with hive 4 changes.

```
prashant@prashant:/drive1/work/temp/apache-hive-3.0.0-bin$ HADOOP_HOME=../hadoop-3.3.6/ bin/hive --orcfiledump ../../orc-reader/data-without-hive4.orc 

Processing data file ../../orc-reader/data-without-hive4.orc [length: 931]
Structure for ../../orc-reader/data-without-hive4.orc
File Version: 0.12 with ORC_135
Rows: 30000
Compression: ZLIB
Compression size: 262144
Type: struct<test:timestamp>

Stripe Statistics:
  Stripe 1:
    Column 0: count: 30000 hasNull: false
    Column 1: count: 30000 hasNull: false min: 1969-12-31 16:00:00.001 max: 1969-12-31 16:00:00.017

File Statistics:
  Column 0: count: 30000 hasNull: false
  Column 1: count: 30000 hasNull: false min: 1969-12-31 16:00:00.001 max: 1969-12-31 16:00:00.017

Stripes:
  Stripe: offset: 3 data: 342 rows: 30000 tail: 78 index: 379
    Stream: column 0 section ROW_INDEX start: 3 length 17
    Stream: column 1 section ROW_INDEX start: 20 length 61
    Stream: column 1 section BLOOM_FILTER start: 81 length 175
    Stream: column 1 section BLOOM_FILTER_UTF8 start: 256 length 126
    Stream: column 1 section DATA start: 382 length 23
    Stream: column 1 section SECONDARY start: 405 length 319
    Encoding column 0: DIRECT
    Encoding column 1: DIRECT_V2

File length: 931 bytes
Padding length: 0 bytes
Padding ratio: 0%
________________________________________________________________________________________________________________________


```

ScrapCodes:
The test failures such as:
`TestOrcReader>AbstractTestOrcReader.testLongDirect:158->AbstractTestOrcReader.testRoundTripNumeric:325 expected [1970-01-01 00:00:00.001] but found [1970-01-01 06:00:00.001]`

Is in the way presto reads the data and writes the data. Somehow even before the data is interpreted as a timestamp type i.e it is still a long type, it has a timestamp adjusted to system timezone. Why it happens is not yet clear to me, when the data written by presto is read via an external ORC reader it has a 6h adjustment applied to it. A similar thing happens when presto reads the data written by hive.

imjalpreet:
This is dependent on multiple PRs. First PR in line: https://github.com/prestodb/presto-hive-apache/pull/65

imjalpreet:
Dependent PR still waiting for reviews: https://github.com/prestodb/presto-hive-apache/pull/65

imjalpreet:
The following dependent PRs under review:

Tempto: https://github.com/prestodb/tempto/pull/299
orc-protobuf (for dwrf): https://github.com/prestodb/orc-protobuf/pull/25
presto-dwrf: https://github.com/prestodb/presto-hive-dwrf/pull/13 (JDK Upgrade) followed by https://github.com/prestodb/presto-hive-dwrf/pull/14 (Hive Upgrade)

imjalpreet:
Cassandra Product Tests failing due to https://github.com/prestodb/tempto/pull/298

imjalpreet:
Need to debug why a couple of recently added native sidecar plugin unit tests are failing after the latest rebase

imjalpreet:
> Need to debug why a couple of recently added native sidecar plugin unit tests are failing after the latest rebase

```
TestNativeSidecarPlugin.testNativeExpressionOptimizer:828->getNowInstant:901->AbstractTestQueryFramework.computeActual:149 ? Runtime  Timezone cannot be null native.default.fail(Timezone cannot be null:VARCHAR) Top-level Expression: cast((native.default.fail(Timezone cannot be null:VARCHAR)) as TIMESTAMP WITH TIME ZONE)
```

Found a bug in Velox due to which a couple of sidecar tests were failing. If  `adjustTimestampToTimezone` is false, timezone returns a nullptr, and it's not handled properly in Velox.

https://github.com/facebookincubator/velox/blob/805db6b72fe326e7b2c97666fe4ba04a7e448568/velox/expression/CastExpr.cpp#L41-L49
+
https://github.com/facebookincubator/velox/blob/805db6b72fe326e7b2c97666fe4ba04a7e448568/velox/functions/prestosql/DateTimeFunctions.h#L1835-L1838

I will be raising a bug in Velox for this.

imjalpreet:
@sourcery-ai summary

imjalpreet:
@sourcery-ai guide

imjalpreet:
PR is waiting for a review.

presto-base-arrow-flight/src/test/java/com/facebook/plugin/arrow/TestArrowFlightQueries.java
```diff
@@ -131,11 +131,12 @@ public void testSelectTime()
     {
         MaterializedResult actualRow = computeActual("SELECT * from event WHERE id = 1");
         Session session = getSession();
+        TimeZoneKey timeZoneKey = session.getSqlFunctionProperties().isLegacyTimestamp() ? session.getTimeZoneKey() : TimeZoneKey.UTC_KEY;
         MaterializedResult expectedRow = resultBuilder(session, INTEGER, DATE, TIME, TIMESTAMP)
                 .row(1,
                         getDate("2004-12-31"),
-                        getTimeAtZone("23:59:59", session.getTimeZoneKey()),
-                        getDateTimeAtZone("2005-12-31 23:59:59", session.getTimeZoneKey()))
+                        getTimeAtZone("23:59:59", timeZoneKey),
```
nmahadevuni: Let's run the test with both legacy and UTK_KEY?

presto-cassandra/src/main/java/com/facebook/presto/cassandra/CassandraPageSink.java
```diff
@@ -156,9 +162,12 @@ else if (REAL.equals(type)) {
         else if (DATE.equals(type)) {
             values.add(toCassandraDate.apply(type.getLong(block, position)));
         }
-        else if (TIMESTAMP.equals(type)) {
+        else if (session.getSqlFunctionProperties().isLegacyTimestamp() && TIMESTAMP.equals(type)) {
```
nmahadevuni, line 190: Is this mutually exclusive? If its legacy, then it must be only TIMESTAMP without TIME ZONE?

presto-cassandra/src/test/java/com/facebook/presto/cassandra/TestCassandraIntegrationSmokeTest.java
```diff
@@ -66,6 +68,7 @@ public class TestCassandraIntegrationSmokeTest
     private static final String KEYSPACE = "smoke_test";
     private static final Session SESSION = createCassandraSession(KEYSPACE);
 
+    private static final ZonedDateTime TIMESTAMP_VALUE = ZonedDateTime.of(1970, 1, 1, 3, 4, 5, 0, ZoneId.of("UTC"));
     private static final Timestamp DATE_TIME_LOCAL = Timestamp.valueOf(LocalDateTime.of(1970, 1, 1, 3, 4, 5, 0));
     private static final LocalDateTime TIMESTAMP_LOCAL = LocalDateTime.of(1969, 12, 31, 23, 4, 5); // TODO #7122 should match DATE_TIME_LOCAL
```
nmahadevuni: This is unused. Can remove it.

nmahadevuni: Same for DATE_TIME_LOCAL

presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/decoders/plain/TimestampPlainValuesDecoder.java
```diff
@@ -48,7 +52,9 @@ public void readNext(long[] values, int offset, int length)
         int localBufferOffset = bufferOffset;
 
         while (offset < endOffset) {
-            values[offset++] = getTimestampMillis(localByteBuffer, localBufferOffset);
+            AtomicLong utcMillis = new AtomicLong(getTimestampMillis(localByteBuffer, localBufferOffset));
```
nmahadevuni: why do we need AtomicLong here?

presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/Int64TimeAndTimestampMicrosNestedBatchReader.java
nmahadevuni, line 70: Don't we need to apply the timezone conversion for this?

presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/Int64TimeAndTimestampMicrosNestedBatchReader.java
```diff
@@ -90,7 +91,7 @@ else if (definitionLevels[definitionLevelIndex] == maxDefinitionLevel - 1) {
     }
 
     @Override
-    protected ColumnChunk readNestedNoNull()
+    protected ColumnChunk readNestedNoNull(Optional<DateTimeZone> timezone)
```
nmahadevuni, line 96: Same comment as in above method.

presto-parquet/src/main/java/com/facebook/presto/parquet/reader/BinaryColumnReader.java
```diff
@@ -37,7 +40,7 @@ public BinaryColumnReader(RichColumnDescriptor descriptor)
     }
 
     @Override
-    protected void readValue(BlockBuilder blockBuilder, Type type)
+    protected void readValue(BlockBuilder blockBuilder, Type type, Optional<DateTimeZone> timezone)
```
nmahadevuni: We are making changes to all non-timestamp readers methods because we want to pass the DateTimeZone. Can we make an abstract class for Timestamp related class and derive all Timestamp readers from it to contain this to few related files?

presto-cassandra/src/main/java/com/facebook/presto/cassandra/CassandraPageSink.java
nmahadevuni: Please add a comment why its so for cassandra.

presto-hive/src/main/java/com/facebook/presto/hive/HiveUtil.java
```diff
@@ -598,21 +600,13 @@ private static void initializeDeserializer(Configuration configuration, Deserial
     {
         try {
             configuration = copy(configuration); // Some SerDes (e.g. Avro) modify passed configuration
-            deserializer.initialize(configuration, schema);
-            validate(deserializer);
+            ((AbstractSerDe) deserializer).initialize(configuration, schema, null);
```
nmahadevuni, line 603: why don't we need to validate the deserializer anymore?

presto-orc/src/main/java/com/facebook/presto/orc/metadata/OrcMetadataReader.java
```diff
@@ -339,10 +347,16 @@ static StringStatistics toStringStatistics(HiveWriterVersion hiveWriterVersion,
             return null;
         }
 
-        Slice maximum = stringStatistics.hasMaximum() ? maxStringTruncateToValidRange(byteStringToSlice(stringStatistics.getMaximumBytes()), hiveWriterVersion) : null;
-        Slice minimum = stringStatistics.hasMinimum() ? minStringTruncateToValidRange(byteStringToSlice(stringStatistics.getMinimumBytes()), hiveWriterVersion) : null;
+        Slice maximum = stringStatistics.hasUpperBound() ?
+                maxStringTruncateToValidRange(byteStringToSlice(stringStatistics.getUpperBoundBytes()), hiveWriterVersion) :
```
nmahadevuni, line 351: these changes look unrelated, why are they here?

nmahadevuni (commented):
Thanks @imjalpreet . Reviewed a part of the code. Can you add new or modify current tests to test the new timestamp behaviour as well as old?

presto-druid/src/test/java/com/facebook/presto/druid/TestDruidQueryGenerator.java
```diff
@@ -187,7 +184,7 @@ public void testTimestampLiteralPushdown()
                                 tableScan(planBuilder, druidTable, regionId, city, fare, datetime),
                                 getRowExpression("datetime = timestamp '2016-06-26 19:00:00.000'", defaultSessionHolder)),
                         ImmutableList.of("city", "datetime")),
-                "SELECT \"city\", \"datetime\" FROM \"realtimeOnly\" WHERE (\"datetime\" = TIMESTAMP '2016-06-26 06:00:00.000')");
+                "SELECT \"city\", \"datetime\" FROM \"realtimeOnly\" WHERE (\"datetime\" = TIMESTAMP '2016-06-26 19:00:00.000')");
         //test timestamp with timezone
```
nmahadevuni, line 188: The result of this test seems to be the same as the next test. Should we test with legacy timestamp for this one?

presto-elasticsearch/src/main/java/com/facebook/presto/elasticsearch/ElasticsearchQueryBuilder.java
```diff
@@ -161,10 +162,10 @@ private static FieldValue getValue(ConnectorSession session, Type type, Object v
         }
 
         if (type.equals(TIMESTAMP)) {
-            checkState(session.getSqlFunctionProperties().isLegacyTimestamp(), "New timestamp semantics not yet supported");
-
             String dateValue = Instant.ofEpochMilli((Long) value)
-                    .atZone(ZoneId.of(session.getSqlFunctionProperties().getTimeZoneKey().getId()))
+                    .atZone(session.getSqlFunctionProperties().isLegacyTimestamp() ?
```
nmahadevuni, line 168: Are there tests to verify with both legacy and new behaviour?

presto-hive/src/main/java/com/facebook/presto/hive/util/SerDeUtils.java
```diff
@@ -279,35 +278,27 @@ private static long formatDateAsLong(Object object, DateObjectInspector inspecto
...
-        long millisLocal = inspector.getPrimitiveJavaObject(object).getTime();
-        // Convert it to midnight in UTC
-        long millisUtc = DateTimeZone.getDefault().getMillisKeepLocal(DateTimeZone.UTC, millisLocal);
-        // Convert midnight UTC to days
-        return TimeUnit.MILLISECONDS.toDays(millisUtc);
+        return inspector.getPrimitiveJavaObject(object).toEpochDay();
     }
 
-    private static long formatTimestampAsLong(Object object, TimestampObjectInspector inspector, DateTimeZone hiveStorageTimeZone)
+    private static long formatTimestampAsLong(Object object, TimestampObjectInspector inspector, DateTimeZone hiveStorageTimeZone, boolean legacyTimestampEnabled)
```
nmahadevuni, line 288: hiveStorageTimeZone already has UTC or a different timezone, do we need to pass legacyTimestampEnabled too? just `hiveStorageTimeZone.convertLocalToUTC(hiveMillis, false) ` should be enough? which would be a no-op if timezone is already UTC?

presto-hive/src/main/java/com/facebook/presto/hive/HiveUtil.java
```diff
@@ -471,9 +471,11 @@ public static long parseHiveDate(String value)
         return TimeUnit.MILLISECONDS.toDays(millis);
     }
 
-    public static long parseHiveTimestamp(String value, DateTimeZone timeZone)
+    public static long parseHiveTimestamp(ConnectorSession session, String value, DateTimeZone timeZone)
     {
-        return HIVE_TIMESTAMP_PARSER.withZone(timeZone).parseMillis(value);
+        return session != null && session.getSqlFunctionProperties().isLegacyTimestamp() ?
```
nmahadevuni: pass just isLegacyTimestamp from the caller making the check there?

presto-hive-metastore/src/main/java/com/facebook/presto/hive/metastore/thrift/ThriftHiveMetastoreClient.java
```diff
@@ -248,7 +253,8 @@ public void setTableColumnStatistics(String databaseName, String tableName, List
     public void deleteTableColumnStatistics(String databaseName, String tableName, String columnName)
             throws TException
     {
-        client.delete_table_column_statistics(constructSchemaName(catalogName, databaseName), tableName, columnName);
+        // TODO: This is not backwards compatible
+        client.delete_table_column_statistics(constructSchemaName(catalogName, databaseName), tableName, columnName, "hive");
```
nmahadevuni: what's the change that requires hardcoded argument "hive" to be sent to the metastore? Can we create a static final member for this?

presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedSmokeTestBase.java
```diff
@@ -2137,12 +2137,10 @@ private void testBucketTransformsOnTimeForFormat(Session session, FileFormat for
                 "(time '07:31:55.425', 7)";
         assertUpdate(session, insertSql, 7);
 
-        assertQuery(session, "SELECT COUNT(*) FROM \"test_bucket_transform_on_time$partitions\"", "SELECT 4");
+        assertQuery(session, "SELECT COUNT(*) FROM \"test_bucket_transform_on_time$partitions\"", "SELECT 2");
```
nmahadevuni, line 2140: Why does this change lead to different number of partitions?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergFileWriterFactory.java
```diff
@@ -94,14 +97,16 @@ public IcebergFileWriterFactory(
...
+            HiveDwrfEncryptionProvider dwrfEncryptionProvider,
+            HiveClientConfig hiveClientConfig)
     {
         this.hdfsEnvironment = requireNonNull(hdfsEnvironment, "hdfsEnvironment is null");
         this.typeManager = requireNonNull(typeManager, "typeManager is null");
         this.readStats = requireNonNull(readStats, "readStats is null");
         this.nodeVersion = requireNonNull(nodeVersion, "nodeVersion is null");
         this.orcFileWriterConfig = requireNonNull(orcFileWriterConfig, "orcFileWriterConfig is null");
         this.dwrfEncryptionProvider = requireNonNull(dwrfEncryptionProvider, "DwrfEncryptionProvider is null").toDwrfEncryptionProvider();
+        this.writerTimezone = requireNonNull(hiveClientConfig, "hiveClientConfig is null").getDateTimeZone();
```
nmahadevuni, line 109: Don't we apply the same legacy vs new semantics for Iceberg?

presto-main-base/src/main/java/com/facebook/presto/util/JsonUtil.java
```diff
@@ -544,7 +545,7 @@ public void writeJsonValue(JsonGenerator jsonGenerator, Block block, int positio
             }
             else {
                 long value = TIMESTAMP.getLong(block, position);
-                jsonGenerator.writeString(printTimestampWithoutTimeZone(properties.getTimeZoneKey(), value));
+                jsonGenerator.writeString(printTimestampWithoutTimeZone(properties.isLegacyTimestamp() ? properties.getTimeZoneKey() : UTC_KEY, value));
```
nmahadevuni, line 548: We could do the same in JsonOperators.java instead of branching on isLegacyTimestamp.

imjalpreet:
Currently, working on the rebase with master to make Cassandra 4.x upgrade compatible with the timestamp semantics changes.

presto-cassandra/src/main/java/com/facebook/presto/cassandra/CassandraPageSink.java
imjalpreet: Yes, it's mutually exclusive. The implementation before this PR will be treated as legacy behaviour(not aligned with Cassandra Timestamp Semantics) and, for backwards compatibility, is mapped to timestamp, and the new implementation after this PR is aligned with Cassandra UTC semantics and is mapped to TIMESTAMP_WITH_TIME_ZONE.

I have added code comments for clarity

presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/decoders/plain/TimestampPlainValuesDecoder.java
imjalpreet: We actually don't need it and can be refactored. It was used only because lambda allows only effectively final variables, and, probably in some iteration of the PR, this change was made. I have refactored to remove the usage of lambda and updated the variable to a simple long.

presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/Int64TimeAndTimestampMicrosNestedBatchReader.java
imjalpreet: `Int64TimeAndTimestampMicrosValuesDecoder` handles timezone at construction time (via the withTimezone flag) rather than at readNext time, so timezone is not forwarded here.

https://github.com/prestodb/presto/blob/master/presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/decoders/plain/Int64TimeAndTimestampMicrosPlainValuesDecoder.java#L43-L49

presto-parquet/src/main/java/com/facebook/presto/parquet/batchreader/Int64TimeAndTimestampMicrosNestedBatchReader.java
imjalpreet: Same as above

presto-parquet/src/main/java/com/facebook/presto/parquet/reader/BinaryColumnReader.java
imjalpreet: Yes, that makes sense. I will refactor.

presto-hive/src/main/java/com/facebook/presto/hive/HiveUtil.java
imjalpreet: This is not required anymore as Hive 4.x changed `AbstractSerDe.initialize()` to throw `SerDeException` directly on configuration errors. The separate `getConfigurationErrors()` check from Hive 3.x was removed from the API. I have handled errors via try-catch now due to this change.

presto-orc/src/main/java/com/facebook/presto/orc/metadata/OrcMetadataReader.java
imjalpreet: No, these changes have come as part of the Hive 4.x upgrade. There are certain statistics changes.

imjalpreet: ORC changed how it handles StringStatistics (see [PR](https://github.com/apache/orc/pull/292/files))

.github/workflows/product-tests-specific-environment.yml
```diff
@@ -77,25 +77,25 @@ jobs:
           OVERRIDE_JDK_DIR: ${{ env.JAVA_HOME }}
         run: presto-product-tests/bin/run_on_docker.sh singlenode -g hdfs_no_impersonation,avro,mixed_case
       - name: Product Tests Specific 1.2
-        if: needs.changes.outputs.codechange == 'true'
+        if: needs.changes.outputs.codechange == 'true' && ${{ always() }}
```
tdcmeehan, line 80: Why do we need this change?

presto-cassandra/src/main/java/com/facebook/presto/cassandra/CassandraRecordCursor.java
```diff
@@ -16,27 +16,33 @@
...
+import static com.facebook.presto.common.type.DateTimeEncoding.packDateTimeWithZone;
 import static io.airlift.slice.Slices.utf8Slice;
 import static java.lang.Float.floatToRawIntBits;
+import static java.util.Objects.requireNonNull;
 
 public class CassandraRecordCursor
         implements RecordCursor
 {
     private final List<FullCassandraType> fullCassandraTypes;
+    private ConnectorSession session;
```
tdcmeehan:
```suggestion
    private final ConnectorSession session;
```

tdcmeehan (dismissed):
Just a couple of nits

.github/workflows/product-tests-specific-environment.yml
imjalpreet: @tdcmeehan It’s not related to this PR, so I can raise it in a separate PR if needed. 

The reason for this change is that this module includes seven product test suites, all of which are independent of each other. Currently, they run sequentially, and if one suite fails, the entire test run fails immediately, preventing the remaining suites from running.

I think it would be better to allow all the suites to run since they’re unrelated. Otherwise, if multiple suites have failures, we end up fixing one issue, waiting for the entire suite to run again, and only then discovering that another suite has also failed.

presto-hive/src/main/java/com/facebook/presto/hive/util/SerDeUtils.java
imjalpreet: No, we still need the legacy timestamp check for two reasons. First, it keeps the behavior consistent with other parts of Presto. Second, we cannot assume that users will configure hiveStorageTimezone to UTC. For backward compatibility, there is currently no restriction preventing the storage time zone from being set to a non-UTC value, even when legacy timestamps are disabled by default. In such cases, the timezone shift should be applied only when the legacy timestamp mode is enabled through the session property, otherwise, no timezone shift should be performed.

Also,  to keep backward compatibility, for certain file formats where we are reading the timestamp from the file footer, we allow users to configure a non-UTC timezone(even when legacy timestamp is disabled) for older files where the footer might not contain the timezone information.

presto-hive-metastore/src/main/java/com/facebook/presto/hive/metastore/thrift/ThriftHiveMetastoreClient.java
imjalpreet: The Hive 4.x Thrift API added a engine parameter to both delete_*_column_statistics calls, and "hive" is the engine name for the default Hive statistics engine. Earlier it used to be fixed as hive but now thrift API allows each engine to maintain its own version of column stats. For backwards compatibility, we are leaving it as hive.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedSmokeTestBase.java
imjalpreet: Iceberg's bucket transform hashes the internal numeric encoding of the value. For TIME, that encoding is microseconds since midnight. 

I debugged the test again, and below are my findings.

Before this PR (`legacy` mode): `TIME` values were encoded with the session timezone offset applied, so for eg time '01:02:03.123' in a session with timezone `America/Bahia_Banderas` (UTC-7, the test default) would have a different microsecond value than in UTC. That offset shifted the values before hashing, distributing them across all 4 buckets.

After this PR (non-legacy, SQL standard): `TIME` values are encoded as pure microseconds since midnight UTC, no session timezone adjustment. The same 7 time literals now hash with their true microsecond values, and it happens that all 7 fall into only 2 of the 4 buckets (0 and 2).

So the partition count drop from 4 to 2 is an expected consequence of fixing the TIME encoding, not a regression.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergFileWriterFactory.java
imjalpreet: This behaviour is consistent with Hive connector's com.facebook.presto.hive.parquet.ParquetFileWriterFactory.

presto-druid/src/test/java/com/facebook/presto/druid/TestDruidQueryGenerator.java
imjalpreet: Yes, but this one is for Timestamp, and the next one is for Timestamp with Timezone. They exercise different code branches in DruidPushdownUtils.getConstantAsString

presto-elasticsearch/src/main/java/com/facebook/presto/elasticsearch/ElasticsearchQueryBuilder.java
imjalpreet: There are timestamp tests executed via TestElasticsearchIntegrationSmokeTest (which extends AbstractTestIntegrationSmokeTest and AbstractTestQueryFramework). These will test the new default behaviour.

For legacy behaviour, we can add tests if needed via a follow-up PR.

imjalpreet: I will add a UT for this PR

nmahadevuni (approved):

tdcmeehan (approved):

imjalpreet:
Both test failures are unrelated.

- [arrow flight tests / arrowflight-native-e2e-tests (:presto-flight-shim) (pull_request)](https://github.com/prestodb/presto/actions/runs/29855233997/job/88914362427?pr=24571) is flaky and I have raised an issue here: https://github.com/prestodb/presto/issues/28205
- [codenotify / Label and notify (pull_request_target)](https://github.com/prestodb/presto/actions/runs/29855232411/job/88717789524?pr=24571) should be fixed after https://github.com/prestodb/presto/pull/28198 is merged

PR #24582, opened 2025-02-18
Change from DataSize to long to remove unnecessary data creation

shangm2:
## Description
1. This is the first step of a series of efforts to make the communication between coordinator and worker slimmer by using simple type instead of DataSize / Duration / DateTime.
2. Facebook internal test will fail due to this change but the corresponding fix is in facebook repo.

## Motivation and Context
1. From profiling, we see the creation and garbage collection of certain objects can be expensive
Screenshot 2025-02-17 at 16 47 30

4. These objects are designed to represent DataSize in a human readable way. However they are expensive to create and cause unnecessary allocation.

## Test Plan
1. passed verifier test: https://www.internalfb.com/intern/presto/verifier/results/?test_id=212504                                                     
3. passed cpp worker verifier: https://our.internmc.facebook.com/intern/presto/verifier/results/?test_id=212506
4. make sure UI does not break

Screenshot 2025-03-04 at 12 02 54

## Contributor checklist

- [ ] CI passed.

```
== RELEASE NOTES ==

General Changes
* Improve scheduling by using long instead of DataSize for critical path.
```


steveburnett:
Thanks for the release note! Nit suggestion rephrasing to consider, to follow the [Order of changes](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines#order-of-changes) in the Release Notes Guidelines: 

```
== RELEASE NOTES ==

General Changes
* Improve scheduling by using long instead of DataSize for critical path.
```

ZacBlanco:
I have a concern that this will break compatibility with some of the existing tooling around Presto. A good number of JSON blobs get exposed to downstream tools parse these JSON blobs. Is there any plan to maintain backwards compatibility or help migrate tools that will break due to this changes?

shangm2:
> I have a concern that this will break compatibility with some of the existing tooling around Presto. A good number of JSON blobs get exposed to downstream tools through the query event listener framework. Is there any plan to maintain backwards compatibility or help migrate tools that will break due to this change?

Hey @ZacBlanco  we certainly dont want to break anything. Could you point me to those downstream tools so I can talk to the team and come up with a plan? Thanks a lot.

ZacBlanco:
At first I was concerned about the event listener, however it seems to have a separate API for the objects that are returned. So that should be good. We do have a few closed-source tools which are used internally to grab information from the coordinator REST API. I think as long as we don't affect the output format of the JSONs at `/v1/query/{queryId}` and  `EXPLAIN ANALYZE` queries it should be fine as that's what most of our tools use. I see that in BasicQueryStats you've already made sure that the fields serialize to DataSize when converted to JSON. After taking a closer look I think this PR is ok in the current state

presto-main/src/main/java/com/facebook/presto/execution/QueryStats.java
```diff
@@ -726,43 +726,43 @@ public double getCumulativeTotalMemory()
     @JsonProperty
     public DataSize getUserMemoryReservation()
     {
-        return userMemoryReservation;
+        return succinctBytes(userMemoryReservationInBytes);
```
NikhilCollooru, line 729: Why not change the method signature to return bytes ? some dependencies outside of this codebase ?

succinctBytes(...) --> this will create a new DataSize object everytime its called. So i am wondering if this change will really help or not .

NikhilCollooru: Discussed offline. This commit alone will not give any wins, we need to change the return types to long as well. So Shang agreed to work on the remaining api conversion stuff in a new commit on top of this exisiting commit.

NikhilCollooru (approved):
LGTM

presto-spi/src/main/java/com/facebook/presto/spi/eventlistener/OperatorStatistics.java
ZacBlanco, line 1: I don't think we'll be able to support merging this PR in the current state if it changes the interfaces for the event listener. We need some kind of transition plan or feature flag since it changes the public API

presto-main/src/main/java/com/facebook/presto/server/BasicQueryStats.java
ZacBlanco, line 1: Similarly, this is part of the public API and there are tools which rely on the response format. We cannot just change it. I think the Presto UI relies on it too. Did you check if there were any errors in the UI?

ZacBlanco (changes requested):
I think the changes to the public APIs in the event listener SPI and BasicQueryStats are a cause of concern for me. Other than that I support this change. It looks fine other than that

presto-spi/src/main/java/com/facebook/presto/spi/eventlistener/OperatorStatistics.java
shangm2: Thanks a lot for catching this @ZacBlanco . Let me remove those public facing changes.

presto-main/src/main/java/com/facebook/presto/server/BasicQueryStats.java
shangm2: Thanks a lot for catching this @ZacBlanco . Let me remove these change as well. I didnot see any UI issue while running verifier but let me remove them just in case.
Screenshot 2025-02-27 at 10 35 15

ZacBlanco: It will likely appear in the Query details page because `BasicQueryStats` populates the `queryStats` object. When I checked out the PR and built it, then ran `SELECT count(*) FROM tpch.orders` The query detail page failed to load

image

shangm2: @ZacBlanco  I have removed changes within presto-spi which we can revisit on a transition plan later. I also updated the UI to make sure it can parse the new "long" date type.
Screenshot 2025-02-27 at 17 20 59

NikhilCollooru (approved):
Changes LGTM. I wonder if we have any tests for the front end UI changes. Just want to make sure we dont break anything.

ZacBlanco:
I think our front end UI tests are lacking, though @yhwang might want to take a look at the changes as well

shangm2:
@ami

> I think our automated front end UI tests are lacking, though @yhwang might want to take a look at the changes as well

@yhwang Feel free to take a look whenever you get a chance. Thanks!

shangm2:
> I think our automated front end UI tests are lacking, though @yhwang might want to take a look at the changes as well

@ZacBlanco any further comments? Thanks!

NikhilCollooru:
@ZacBlanco @yhwang  did you get a chance to review the code changes ?

ZacBlanco:
Sorry I have been preoccupied with some other things. I will try to review again later today or tomorrow

presto-ui/src/components/QueryDetail.jsx
```diff
@@ -1474,7 +1480,7 @@ export class QueryDetail extends React.Component {
                                             Input Data
                                         </td>
                                         <td className="info-text">
-                                            {query.queryStats.processedInputDataSize}
+                                            {formatDataSizeBytes(query.queryStats.processedInputDataSizeInBytes) + " Bytes"}
```
yhwang, line 1483: The old style only appends `B`. Do you want to use the old style?
Same as other fields in this file.

yhwang: thanks for the update. My fault that I didn't mention there is no space before the `B`.

yhwang: I found that you can directly use `formatDataSize`. it will append `B` for you.

shangm2: @yhwang  I have updated the code to use formatDataSize. Thanks for the insight.

yhwang: I haven't seen your latest update yet. Can you also apply these changes to `QueryOverview.jsx`? Thanks.

shangm2: @yhwang  My bad. I think I am committed the wrong version. Let me update both.

shangm2: @yhwang  I double checked. I make sure that formatDataSize is being used for memory related metrics for both queryoverview and querydetail. Feel free to take a look. Thanks a lot!

yhwang: nice!

yhwang (approved):
LGTM for the UI part

NikhilCollooru (approved):

ZacBlanco (commented):
Thanks again for all the updates @shangm2 . If we do end up changing public endpoints (see my comment), I would request that we also make sure to update the existing openapi spec under the `presto-openapi` module.

I think this change should be a topic to bring up at the upcoming TSC to see how folks feel about it and whether they are OK breaking the REST API compatibility. I will defer to the community to make the final decision, but I would err on the side that we should not break the API. If you can present some compelling performance results at the TSC meeting, it might also help convince the community that this change is worth breaking API compatibility.

How does that sound to you?

NikhilCollooru:
Merging this, since we got a go ahead regarding the UI changes.

amitkdutta (approved):
Looks good.

NikhilCollooru:
sorry ! I missed @ZacBlanco latest comment. We will address the comments in the next PR

NikhilCollooru:
> I think this change should be a topic to bring up at the upcoming TSC to see how folks feel about it and whether they are OK breaking the REST API compatibility. I will defer to the community to make the final decision, but I would err on the side that we should not break the API. If you can present some compelling performance results at the TSC meeting, it might also help convince the community that this change is worth breaking API compatibility.

We agree we will make sure to bring this up in the next TSC. Thanks for the review. We will also discuss internally regarding this change.

presto-main/src/main/java/com/facebook/presto/execution/QueryStats.java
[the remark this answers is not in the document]
shangm2: The reason we want to get rid of those heavy classes is to make the communication between coordinator and worker leaner, not just the internals of the coordinator. Let me put api changes under /v2 endpoint and let workers and ui use v2 but keep v1 for backward compatible.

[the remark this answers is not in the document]
ZacBlanco: Do we have any endpoints which are generally coordinator/worker only? E.g. I believe the `/v1/query` should never be accessed by the worker. But the `/v1/task` API would be. Can we limit breaking changes to APIs which are only coordinator-worker?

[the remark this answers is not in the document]
shangm2: right, you are absolutely correct. Let me create a new pr to remove all changes that could be exposed through public-facing endpoints. Thanks for the insights @ZacBlanco

Issue #24563, opened 2025-02-14
Inconsistent behavior in json functions


bikramSingh91:
While working on implementing JSON functions in Velox for Presto-CPP, I came across some inconsistencies that I wanted to highlight and get the community's opinion on whether it makes sense for us to fix them, either both in Presto-Java and Presto-CPP or just in Presto-CPP.

### 1. `json_extract` Produces Non-Normalized JSON as Output

We expect the JSON type to be normalized whenever it exists in a query. Since Presto does not store JSON as a type that can be written to files, this normalization is usually done when a VARCHAR column is converted to JSON using one of the JSON functions (typically `json_parse`). This is important because it can affect the equality of two JSON objects and provide inconsistent results if sometimes the JSONs are normalized and sometimes they are not.

For example, suppose the input string is:

```json
{ "key_2": 2, "key_3": 3, "key_1": 1 }
```

`json_parse` will normalize this to:

```sql
presto:default> select JSON_PARSE('{ "key_2": 2, "key_3": 3, "key_1": 1 }');
_col0
---------------------------------
{"key_1":1,"key_2":2,"key_3":3}
```

**NOTE**: Normalization puts the keys in ascending order and removes spaces.

Now, if there are two inputs:

```json
{ "key_2": 2, "key_3": 3, "key_1": 1 }
{ "key_1": 1, "key_2": 2, "key_3": 3 }
```

These should be considered equal JSONs when parsed and compared; otherwise, this can result in correctness issues. If we use `json_extract`, which returns a JSON object type, it would extract and return a non-normalized JSON:

```sql
presto:default> select JSON_EXTRACT('{ "key_2": 2, "key_3": 3, "key_1": 1 }', '$');
_col0
---------------------------------
{"key_2":2,"key_3":3,"key_1":1}
```

**NOTE**: Spaces are removed, but keys remain untouched.

The same issue occurs in `json_array_get`:

```sql
presto:default> select json_array_get('[{ "key_2": 2, "key_1": 1 }]', 0);
_col0
-----------------------
{"key_2":2,"key_1":1}
```

**Recommendation**: The expectation is that once a VARCHAR is converted to JSON, it should be in a normalized form. Therefore, we should always normalize the resultant JSON in all functions that take VARCHAR as prospective JSON input and output a JSON.

### 2. `json_extract` / `_scalar`, `json_array_get` Accepts Invalid JSON to Produce Valid Result

[...]

PR #24614, opened 2025-02-24
fix Update json_extract to Produce Canonicalized Output

duxiao1212:
## Description
Canonicalize json_extract output by replacing the copy operation with writing the JSON to output stream with sorted keys

## Motivation and Context
This relates to issue [#24563](https://github.com/prestodb/presto/issues/24563).
Ensuring JSON output canonicalization guarantees the accuracy of JSON comparisons and aligns the behavior with json_parse

## Impact
low impact

## Test Plan
SET SESSION canonicalized_json_extract = true

1. json_extract result should be canonicalized
image

2. If the JSON result is invalid, a PrestoException will be thrown
image

3. Benchmark test result
Enabling canonicalized JSON extract (isCanonicalizedJsonExtract = true) causes a 11.7% performance drop compared to disabling it (isCanonicalizedJsonExtract = false), specifically in complex test cases involving nested JSON objects in the JSON path.
Pasted Graphic

## Contributor checklist

- [x] CI passed.

## Release Notes
== RELEASE NOTES ==

Presto Changes
- Add session property ``canonicalized_json_extract``. :pr:`24614`

### DIFF
--- a/presto-common/src/main/java/com/facebook/presto/common/function/SqlFunctionProperties.java
+++ b/presto-common/src/main/java/com/facebook/presto/common/function/SqlFunctionProperties.java
@@ -37,6 +37,7 @@ public class SqlFunctionProperties
     private final boolean legacyJsonCast;
     private final Map<String, String> extraCredentials;
     private final boolean warnOnCommonNanPatterns;
+    private final boolean canonicalizedJsonExtract;
 
     private SqlFunctionProperties(
             boolean parseDecimalLiteralAsDouble,
@@ -50,7 +51,8 @@ public class SqlFunctionProperties
             boolean fieldNamesInJsonCastEnabled,
             boolean legacyJsonCast,
             Map<String, String> extraCredentials,
-            boolean warnOnCommonNanPatterns)
+            boolean warnOnCommonNanPatterns,
+            boolean canonicalizedJsonExtract)
     {
         this.parseDecimalLiteralAsDouble = parseDecimalLiteralAsDouble;
         this.legacyRowFieldOrdinalAccessEnabled = legacyRowFieldOrdinalAccessEnabled;
@@ -64,6 +66,7 @@ public class SqlFunctionProperties
         this.legacyJsonCast = legacyJsonCast;
         this.extraCredentials = requireNonNull(extraCredentials, "extraCredentials is null");
         this.warnOnCommonNanPatterns = warnOnCommonNanPatterns;
+        this.canonicalizedJsonExtract = canonicalizedJsonExtract;
     }
 
     public boolean isParseDecimalLiteralAsDouble()
@@ -127,6 +130,9 @@ public class SqlFunctionProperties
         return warnOnCommonNanPatterns;
     }
 
+    public boolean isCanonicalizedJsonExtract()
+    { return canonicalizedJsonExtract; }
+
     @Override
     public boolean equals(Object o)
     {
@@ -146,7 +152,8 @@ public class SqlFunctionProperties
                 Objects.equals(sessionLocale, that.sessionLocale) &&
                 Objects.equals(sessionUser, that.sessionUser) &&
                 Objects.equals(extraCredentials, that.extraCredentials) &&
-                Objects.equals(legacyJsonCast, that.legacyJsonCast);
+                Objects.equals(legacyJsonCast, that.legacyJsonCast) &&
+                Objects.equals(canonicalizedJsonExtract, that.legacyJsonCast);
     }
 
     @Override
@@ -154,7 +161,7 @@ public class SqlFunctionProperties
     {
         return Objects.hash(parseDecimalLiteralAsDouble, legacyRowFieldOrdinalAccessEnabled, timeZoneKey,
                 legacyTimestamp, legacyMapSubscript, sessionStartTime, sessionLocale, sessionUser,
-                extraCredentials, legacyJsonCast);
+                extraCredentials, legacyJsonCast, canonicalizedJsonExtract);
     }
 
     public static Builder builder()
@@ -176,6 +183,7 @@ public class SqlFunctionProperties
         private boolean legacyJsonCast;
         private Map<String, String> extraCredentials = emptyMap();
         private boolean warnOnCommonNanPatterns;
+        private boolean canonicalizedJsonExtract;
 
         private Builder() {}
 
@@ -251,6 +259,12 @@ public class SqlFunctionProperties
             return this;
         }
 
+        public Builder setCanonicalizedJsonExtract(boolean canonicalizedJsonExtract)
+        {
+            this.canonicalizedJsonExtract = canonicalizedJsonExtract;
+            return this;
+        }
+
         public SqlFunctionProperties build()
         {
             return new SqlFunctionProperties(
@@ -265,7 +279,8 @@ public class SqlFunctionProperties
                     fieldNamesInJsonCastEnabled,
                     legacyJsonCast,
                     extraCredentials,
-                    warnOnCommonNanPatterns);
+                    warnOnCommonNanPatterns,
+                    canonicalizedJsonExtract);
         }
     }
 }
--- a/presto-main/src/main/java/com/facebook/presto/operator/scalar/JsonExtract.java
+++ b/presto-main/src/main/java/com/facebook/presto/operator/scalar/JsonExtract.java
@@ -13,6 +13,8 @@
  */
 package com.facebook.presto.operator.scalar;
 
+import com.facebook.airlift.json.JsonObjectMapperProvider;
+import com.facebook.presto.common.function.SqlFunctionProperties;
 import com.facebook.presto.spi.PrestoException;
 import com.fasterxml.jackson.core.JsonFactory;
 import com.fasterxml.jackson.core.JsonGenerator;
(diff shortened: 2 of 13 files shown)


shangm2:
Maybe also add a verifier run?

presto-main/src/test/java/com/facebook/presto/operator/scalar/TestJsonExtractFunctions.java
```diff
@@ -87,6 +87,47 @@ public void testJsonExtract()
...
+
+        // Test with Array of JSON objects
+        json = "[{\"key_b\":\"v_b\",\"key_a\":\"v_a\"}, {\"key_2\": \"value_2\"}]";
+        path = "$[0]";
+        expected = "{\"key_a\":\"v_a\",\"key_b\":\"v_b\"}";
+        assertFunction(format("JSON_EXTRACT('%s', '%s')", json, path), JSON, expected);
+    }
+
+    @Test
+    public void testParseNullIfJsonInvalid()
```
rschlussel, line 113: test for the expected exception here instead of checking for null you can use assertInvalidFunction() once you change to throwing an invalid_function_argument error code.

presto-main/src/main/java/com/facebook/presto/operator/scalar/JsonExtract.java
```diff
@@ -302,11 +306,12 @@ public Slice extract(JsonParser jsonParser)
...
-
             DynamicSliceOutput dynamicSliceOutput = new DynamicSliceOutput(ESTIMATED_JSON_OUTPUT_SIZE);
-            try (JsonGenerator jsonGenerator = createJsonGenerator(JSON_FACTORY, dynamicSliceOutput)) {
-                jsonGenerator.copyCurrentStructure(jsonParser);
-            }
+            // Write the JSON to output stream with sorted keys
+            SORTED_MAPPER.writeValue((OutputStream) dynamicSliceOutput, SORTED_MAPPER.readValue(jsonParser, Object.class));
+            // nextToken() returns null if the input is parsed correctly,
+            // but will throw an exception if there are trailing characters.
+            jsonParser.nextToken();
```
rschlussel, line 314: catch the exception and throw a PrestoException with an invalid_function_argument error code. include the failing jsonInput in the error message.  Also, when there is no exception, test that the value is null (non-null and no exception means there's another valid json object after this one) and throw a PrestoException with invalid_function_argument for that case too.

rschlussel:
Also let's gate this change behind a configuration/session property.  You can look at usages of other properties in SqlFunctionProperties for examples

presto-main/src/main/java/com/facebook/presto/operator/scalar/JsonExtract.java
duxiao1212: Hi @rschlussel, the condition jsonParser.nextToken() == null is only suitable for json_parse func, but not for json_extract. For example, given the JSON {k1: {...}, k2: v2} and the path $.k1, in this case, the next token would essentially be "FIELD_NAME".
I believe that to validate the entire input, we would need to either implement logic to traverse the JSON structure and verify the validity of all elements or parse the entire input as a JSON object before performing the extraction. However, both of these approaches seem excessive, especially considering that Presto is currently in maintenance mode
Instead, I suggest fixing it to produce Canonicalized Output and throwing a parser error instead of returning null. What do you think?

rschlussel: got it. That makes sense.

I'm not sure what you mean "throwing a parser error instead of returning null". Do you mean a parsing error when there are exceptions in the nextToken call you added?  That sounds good to me.

presto-main/src/main/java/com/facebook/presto/operator/scalar/JsonExtract.java
```diff
@@ -121,13 +127,25 @@ public final class JsonExtract
...
+    public static <T> T extract(Slice jsonInput, JsonExtractor<T> jsonExtractor, SqlFunctionProperties properties)
     {
         requireNonNull(jsonInput, "jsonInput is null");
         try {
-            return jsonExtractor.extract(jsonInput.getInput());
+            return jsonExtractor.extract(jsonInput.getInput(), properties);
+        }
+        // Instead of suppressing PrestoException, enhance the error message to provide more context and information.
+        catch (PrestoException e) {
+            if (e.getMessage().endsWith("Invalid Json input: ")) {
```
duxiao1212, line 143: May not be necessary.

presto-main/src/test/java/com/facebook/presto/operator/scalar/TestJsonExtract.java
```diff
@@ -221,120 +257,120 @@ public void testObjectFieldJsonExtractor()
     @Test
     public void testFullScalarExtract()
```
rschlussel, line 258: we should run all these tests for both legacy and non-legacy

presto-common/src/main/java/com/facebook/presto/common/function/SqlFunctionProperties.java
```diff
@@ -37,6 +37,7 @@ public class SqlFunctionProperties
     private final boolean legacyJsonCast;
     private final Map<String, String> extraCredentials;
     private final boolean warnOnCommonNanPatterns;
+    private final boolean legacyJsonExtract;
```
rschlussel, line 40: might be more helpful for users to have the config property say what's different.  e.g. cannonicalizeJsonExtract() or something.

rschlussel (commented):
general approach looks good to me!

presto-main/src/main/java/com/facebook/presto/operator/scalar/JsonExtract.java
```diff
@@ -210,25 +228,31 @@ public ObjectFieldJsonExtractor(String fieldName, PrestoJsonExtractor<? extends
...
-                return processJsonArray(jsonParser);
+                return processJsonArray(jsonParser, properties);
             }
 
             throw new JsonParseException(jsonParser, "Expected a JSON object or array");
         }
 
-        public T processJsonObject(JsonParser jsonParser)
+        @Override
+        T extract(JsonParser jsonParser, SqlFunctionProperties properties)
```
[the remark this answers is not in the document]
duxiao1212: no need, updated.

presto-main/src/test/java/com/facebook/presto/type/TestRowOperators.java
```diff
@@ -498,47 +512,54 @@ public void testJsonToRow()
                 mapType(BIGINT, new ArrayType(RowType.from(ImmutableList.of(
                         RowType.field("name", VARCHAR),
                         RowType.field("age", VARCHAR))))),
-                asMap(ImmutableList.of(1L), asList(asList(asList("John", "30")))));
+                asMap(ImmutableList.of(1L), Collections.singletonList(Collections.singletonList(asList("John", "30")))));
```
rschlussel, line 515: why did this (and the other similar examples in this file) change?

presto-main/src/test/java/com/facebook/presto/sql/gen/TestExpressionCompiler.java
```diff
@@ -797,27 +796,27 @@ public void testCast()
             throws Exception
     {
         for (Boolean value : booleanValues) {
-            assertExecute(generateExpression("cast(%s as boolean)", value), BOOLEAN, value == null ? null : (value ? true : false));
+            assertExecute(generateExpression("cast(%s as boolean)", value), BOOLEAN, (value));
```
rschlussel, line 799: confirming: most of these changes were opportunistic fixes removing unneeded conditionals, is that right? would be nice to pull that out into a separate commit (can stay in this PR), just to make it easy to isolate the changes that are relevant to json_extract.

rschlussel (commented):
Can you run a verifier comparing this with Prestissimo and check if they now match?

presto-main/src/test/java/com/facebook/presto/type/TestRowOperators.java
duxiao1212: not related to this update, reverted

presto-main/src/main/java/com/facebook/presto/sql/analyzer/FunctionsConfig.java
```diff
@@ -47,6 +48,7 @@ public class FunctionsConfig
     private boolean warnOnPossibleNans;
     private boolean legacyCharToVarcharCoercion;
     private boolean legacyJsonCast = true;
+    private boolean canonicalizedJsonExtract = true;
```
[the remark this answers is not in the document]
rschlussel: I think we should because it brings parity with presto c++. do you have any performance numbers about the extra cost? You can create a benchmark and get numbers for both on and off (see https://github.com/prestodb/presto/blob/master/presto-main/src/test/java/com/facebook/presto/operator/scalar/BenchmarkJsonToArrayCast.java for an example) and if the regression isn't too significant we can enable by default.

duxiao1212:
> Can you run a verifier comparing this with Prestissimo and check if they now match?

duxiao1212:
> Can you run a verifier comparing this with Prestissimo and check if they now match?
@bikramSingh91, @kgpai, can you please confirm whether the JSON_EXTRACT function is currently available and functional in prestissimo?

presto-main/src/main/java/com/facebook/presto/sql/analyzer/FunctionsConfig.java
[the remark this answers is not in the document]
duxiao1212: @rschlussel, @amitkdutta, please check the benchmark presto-main/src/test/java/com/facebook/presto/operator/scalar/BenchmarkJsonExtract.java and the test result in the summary.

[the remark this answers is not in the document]
rschlussel: given the 11% regression, let's disable by default and we can enable it ourselves in our configuration.  Otherwise, this looks good to go.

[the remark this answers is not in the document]
duxiao1212: Hey @rschlussel, I've updated the default to false. Thanks for your thoughtful review! Can I get a stamp on this?

rschlussel (approved):

kgpai:
@duxiao1212 Yes json_extract is available in Prestissimo . Feel free to verify against it.

PR #24089, opened 2024-11-19
Add overflow detection to IntervalYearMonthOperators

tdcmeehan:
## Description
Part of #24087

## Motivation and Context
Fix incorrect data caused by overflow, and correct the type mapping in certain operators.

## Impact
Where previously some expressions could have resulted in an overflow and incorrect results, now they will fail with a `NUMERIC_VALUE_OUT_OF_RANGE` error.

## Test Plan
Unit tests

## Contributor checklist

- [ ] Please make sure your submission complies with our [development](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#development), [formatting](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#formatting), [commit message](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#commit-formatting-and-pull-requests), and [attribution guidelines](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#attribution).
- [ ] PR description addresses the issue accurately and concisely.  If the change is non-trivial, a GitHub Issue is referenced.
- [ ] Documented new properties (with its default value), SQL syntax, functions, or other functionality.
- [ ] If release notes are required, they follow the [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines).
- [ ] Adequate tests were added if applicable.
- [ ] CI passed.

## Release Notes
Please follow [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines) and fill in the release notes below.

```
== RELEASE NOTES ==

General Changes
* Fix addition, subtraction, multiplication and division of ``INTERVAL HOUR`` and ``INTERVAL YEAR`` values.  :pr:`24089`
```

Issue #24087, opened 2024-11-19
Interval values can overflow


tdcmeehan:
This produces the correct value:

```
presto> select interval '1' month * 2147483647;
    _col0    
-------------
 178956970-7 
(1 row)
```

The next higher number will overflow:

```
presto> select interval '1' month * 2147483648;
    _col0     
--------------
 -178956970-8 
(1 row)
```

## Your Environment
This is true for any functional environment of Presto.

## Expected Behavior
It should be an error to exceed the boundary of the interval type.

## Current Behavior
The interval will silently overflow and return incorrect results.

## Possible Solution
Modify the interval operators to do proper overflow checking, as is done for e.g. `INTEGER operators.

## Steps to Reproduce
See the above queries for a simple reproduction.

## Context
This was discovered while converting constant folding of expressions to use Velox for their evaluation.

asharan2buff:
Can I take this up?

tdcmeehan:
@asharan2buff sure.  I have opened #24089 to address the issue for `IntervalYearMonthOperators`, please take up [`IntervalDayTimeOperators`](https://github.com/prestodb/presto/blob/master/presto-main/src/main/java/com/facebook/presto/type/IntervalDayTimeOperators.java).

LiBrian415:
@asharan2buff Did you ever start working on this? If not, could I pick it up?

tdcmeehan:
@LiBrian415 please, go ahead.

PR #24617, opened 2025-02-24
Add overflow detection to IntervalYearMonthOperators

pramodsatya:
## Description
Part of #24087. Duplicate of https://github.com/prestodb/presto/pull/24089, continuing it here since Tim is OOO and this change is needed for Velox PR: https://github.com/facebookincubator/velox/pull/11612.

## Motivation and Context
Fix incorrect data caused by overflow, and correct the type mapping in certain operators.

## Impact
Where previously some expressions could have resulted in an overflow and incorrect results, now they will fail with a `NUMERIC_VALUE_OUT_OF_RANGE` error.

## Test Plan
Unit tests

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Fix addition, subtraction, multiplication and division of ``INTERVAL YEAR MONTH`` values.
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/type/IntervalYearMonthOperators.java
+++ b/presto-main/src/main/java/com/facebook/presto/type/IntervalYearMonthOperators.java
@@ -17,6 +17,7 @@ import com.facebook.presto.client.IntervalYearMonth;
 import com.facebook.presto.common.block.Block;
 import com.facebook.presto.common.type.AbstractIntType;
 import com.facebook.presto.common.type.StandardTypes;
+import com.facebook.presto.spi.PrestoException;
 import com.facebook.presto.spi.function.BlockIndex;
 import com.facebook.presto.spi.function.BlockPosition;
 import com.facebook.presto.spi.function.IsNull;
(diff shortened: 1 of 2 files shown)


steveburnett:
Thanks for the release note!

```

:pr:`12345`

```

pramodsatya:
Thanks @steveburnett, removed the manual PR link from release note. Could you please take another look?

steveburnett:
> Thanks @steveburnett, removed the manual PR link from release note. Could you please take another look?

LGTM! Thanks!

hantangwangd (approved):
Looks good to me, thanks!

pramodsatya:
@jaystarshot, @yingsu00, could you please help review this PR?

majetideepak:
@jaystarshot Can you please help with this review? Thanks!

presto-main/src/main/java/com/facebook/presto/type/IntervalYearMonthOperators.java
```diff
@@ -56,49 +58,81 @@ private IntervalYearMonthOperators()
...
     }
 
     @ScalarOperator(MULTIPLY)
     @SqlType(StandardTypes.INTERVAL_YEAR_TO_MONTH)
     public static long doubleMultiply(@SqlType(StandardTypes.DOUBLE) double left, @SqlType(StandardTypes.INTERVAL_YEAR_TO_MONTH) long right)
     {
-        return (long) (left * right);
+        long result = (long) (left * right);
+        if (result < Integer.MIN_VALUE || result > Integer.MAX_VALUE) {
+            throw new PrestoException(NUMERIC_VALUE_OUT_OF_RANGE, "Overflow multiplying double by interval year-month value: " + left + " * " + right);
```
jaystarshot, line 122: Why are we comparing the result with int_max here? The return type is long

hantangwangd: As I understand, like `IntegerType`, the type `IntervalYearMonthType` in presto specifies `long.class` as its `javaType`, but the actual primitive value is int. So the functions here need to declare `long` as their parameters and return types, but the actual values should be within the range of integers. This PR handles integer overflow scenarios by directly throwing expected exceptions rather than returning unexpected values after overflow.

pramodsatya: +1, thanks @hantangwangd.

jaystarshot (approved):

PR #24630, opened 2025-02-26
Add hive session property to disable stats based filter reordering

kewang1024:
```
== RELEASE NOTES ==
Hive Connector Changes
* Add session property ``hive.stats_based_filter_reorder_disabled`` for disabling reader stats based filter reordering  

```

### DIFF
--- a/presto-hive/src/main/java/com/facebook/presto/hive/HiveSessionProperties.java
+++ b/presto-hive/src/main/java/com/facebook/presto/hive/HiveSessionProperties.java
@@ -134,6 +134,8 @@ public final class HiveSessionProperties
     public static final String SKIP_EMPTY_FILES = "skip_empty_files";
     public static final String LEGACY_TIMESTAMP_BUCKETING = "legacy_timestamp_bucketing";
 
+    public static final String STATS_BASED_FILTER_REORDER_DISABLED = "stats_based_filter_reorder_disabled";
+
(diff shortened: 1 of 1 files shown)


presto-hive/src/main/java/com/facebook/presto/hive/HiveSessionProperties.java
```diff
@@ -134,6 +134,8 @@ public final class HiveSessionProperties
     public static final String SKIP_EMPTY_FILES = "skip_empty_files";
     public static final String LEGACY_TIMESTAMP_BUCKETING = "legacy_timestamp_bucketing";
 
+    public static final String STATS_BASED_FILTER_REORDER_DISABLED = "stats_based_filter_reorder_disabled";
```
xiaoxmeng, line 137: Do we want to add native prefix as the other native only session properties?

xiaoxmeng (dismissed):

presto-hive/src/main/java/com/facebook/presto/hive/HiveSessionProperties.java
kewang1024: Good point, but looks like the naming conversion only exists for system session property, connector specific ones we don't have.

Maybe we can add that in the follow up as long term solution

zacw7 (approved):

NikhilCollooru (approved):

PR #24644, opened 2025-02-27
[native] Add queuedTimeInNanos in task stats

kewang1024:
```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/TaskManager.cpp
+++ b/presto-native-execution/presto_cpp/main/TaskManager.cpp
@@ -591,6 +591,8 @@ std::unique_ptr<TaskInfo> TaskManager::createOrUpdateTaskImpl(
     execTask->start(maxDrivers, concurrentLifespans);
 
     prestoTask->taskStarted = true;
+    prestoTask->info.stats.queuedTimeInNanos =
+        velox::getCurrentTimeMs() - prestoTask->createTimeMs;
     resultRequests = std::move(prestoTask->resultRequests);
     statusRequest = prestoTask->statusRequest;
     infoRequest = prestoTask->infoRequest;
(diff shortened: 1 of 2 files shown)


presto-native-execution/presto_cpp/main/TaskManager.cpp
```diff
@@ -591,6 +591,8 @@ std::unique_ptr<TaskInfo> TaskManager::createOrUpdateTaskImpl(
     execTask->start(maxDrivers, concurrentLifespans);
 
     prestoTask->taskStarted = true;
+    prestoTask->info.stats.queuedTimeInNanos =
+        velox::getCurrentTimeMs() - prestoTask->createTimeMs;
```
spershin, line 595: So, we update Task's `queuedTimeInNanos` in `TaskManager::createOrUpdateTaskImpl()` upon start and we use difference between `now` and `prestoTask->createTimeMs`.
I think it will pretty much going to be ~0 all the time.

We need to understand the semantics of Task's `queuedTimeInNanos`.
What does it hold - the sum of all queued times of all its drivers in all its pipelines?

Worth checking in Java and also how we populate the `taskBlockedTimeNanos`.

kewang1024: I checked the java implementation, basically it's between creating a task and a task is started and I try to mimic its behavior: https://github.com/prestodb/presto/blob/f7103fe0c052460c38328f0708a1127aeccbf0c3/presto-main/src/main/java/com/facebook/presto/operator/TaskContext.java#L539

yuandagits (approved):

presto-native-execution/presto_cpp/main/TaskManager.cpp
spershin: Strange metric, honestly. In native worker it would always be close to zero.

spershin (approved):

PR #24647, opened 2025-02-27
Fix check_access_control_on_utilized_columns_only for CTE

rschlussel:
## Description
Previously we weren't associating the with query with the columns used from it in the main part of the query, so we wouldn't consider any columns as used.  When the usage within the cte or outside of it was just an identifier, we would have the utilized column collected anyway because the column collected in the main query kept the source table information. However, if there was an expression in between, we would lose that information, and we wouldn't check access control for the column that was used in an expression in the cte.

Example:

For the following query
```
with cte as (select  x as c1, z + 1 as c2 from t13) select c1, c2 from (select * from cte)
```

Previously we would only check access permissions on t13.x, but not t13.z.  With this change we will check column access on both 13.x and t13.z.

## Motivation and Context
Fix a bug

## Impact

Fixes a security hole where we were not checking access for some columns used in a query when the reference is within a with clause.

## Test Plan
unit test

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Fix a security bug when ``check_access_control_for_utlized_columns`` is true for queries that uses a ``WITH`` clause. Previously we would sometime not check permissions for certain columns that were used in the query.  Now we will always check permissions for all columns used in the query. There are some corner cases for CTEs with the same name where we may check more columns than are used or fall back to checking all columns referenced in the query. 
```


rschlussel:
I'm realizing i didn't test for the following cases, and not sure they will work properly

1. 2 with clauses and one refers to the other  e.g. with cte1 as (SELECT x as c1, z+1 as c2 from t13), cte2 as (SELECT c1+1 as a c2 as b) SELECT * from cte2.  Not sure what order the with clauses show up in the list, but they may actually need to be evaluated in reverse order
2. 2 ctes with the same name (e.g. nested/within different scopes).  I think to handle this we'd need a different way to identify the source table that takes into account the scope.  If it's too complex IMO it's okay that we don't handle this and just fall back to regular, but failing the checkState is probably not the clearest way to do that.  And ideally we'd add a query warning or something so people know that happened.

rschlussel:
I've updated this to handle multiple ctes that reference each other, and add more explicit handling for falling back to checking all columns when multiple ctes have the same name. Also added a warning for queries that fall back to checking all columns so that it's easier to see when it's happening.  Since this is a security vulnerability, i'd rather get the fix in quickly than invest in handling that corner case.

presto-main/src/main/java/com/facebook/presto/sql/analyzer/UtilizedColumnsAnalyzer.java
```diff
@@ -526,7 +526,10 @@ public void copyFieldIdsToExploreForWithQuery(WithQuery withQuery)
             List<RelationId> relationIds = fieldsToExplore.keySet().stream()
                     .filter(key -> key.getSourceNode() instanceof Table && ((Table) key.getSourceNode()).getName().equals(name))
                     .collect(toImmutableList());
-            if (relationIds.size() != 1) {
+            if (relationIds.isEmpty()) {
```
rschlussel, line 529: In what cases would the relationId would be empty?  Is that if the CTE doesn't actually get used in the query or are there other cases as well? Can you add tests?

presto-main/src/test/java/com/facebook/presto/sql/analyzer/TestUtilizedColumnsAnalyzer.java
```diff
@@ -519,6 +519,69 @@ public void testUDF()
...
+        assertUtilizedTableColumns(
+                "with cte1 as (select x + 1 as c1, y as c2, z + 1 as c3 from t13), cte2 as (with cte1 AS (select y +1 as c1 from t13) select * from cte1) SELECT cte1.c1, cte2.c1 from cte1 join cte2 on cte1.c1=cte2.c1",
+                ImmutableMap.of(QualifiedObjectName.valueOf("tpch.s1.t13"), ImmutableSet.of("x", "y")));
+    }
+
+    @Test
+    public void testMultipleCtesSameNameSameTable()
+    {
+        // TODO(kevintang2022): Fix visitWithQuery so that it looks up relation properly.
+        // Currently, it just looks the relations using the name (string) of the CTE
```
kevintang2022, line 553: Added some todo for followup. The name conflict is caused by just using name for the look up. We might need to lookup with more info or add a unique identifier to each column

kevintang2022 (approved):

presto-analyzer/src/main/java/com/facebook/presto/sql/analyzer/RelationId.java
```diff
@@ -53,6 +53,10 @@ public boolean isAnonymous()
         return sourceNode == null;
     }
 
+    public Node getSourceNode()
+    {
+        return sourceNode;
+    }
```
rschlussel, line 59: this is missing a newline

presto-main/src/test/java/com/facebook/presto/sql/analyzer/TestUtilizedColumnsAnalyzer.java
```diff
@@ -519,6 +519,69 @@ public void testUDF()
...
+                ImmutableMap.of(QualifiedObjectName.valueOf("tpch.s1.t13"), ImmutableSet.of("x", "y", "z")));
+    }
+
+    @Test
+    public void testMultipleCtesSameNameDifferentTable()
+    {
+        assertUtilizedTableColumns(
+                "with cte1 as (select x + 1 as c1 from t13), cte2 as (with cte1 as (select a + 1 as c2, b + 1 as c3, c as c4 from t1) select * from cte1) select cte1.c1 from cte1 join cte2 on true",
+                ImmutableMap.of(QualifiedObjectName.valueOf("tpch.s1.t13"), ImmutableSet.of("x"),
+                        QualifiedObjectName.valueOf("tpch.s1.t1"), ImmutableSet.of("a")));
```
rschlussel, line 565: why is this only 'a' and not 'a', 'b', 'c' or empty?  is it looking up column cte1.c1 by index and ending up with this? 

Maybe add some comments explaining what's going on

presto-main/src/test/java/com/facebook/presto/sql/analyzer/TestUtilizedColumnsAnalyzer.java
```diff
@@ -519,6 +519,69 @@ public void testUDF()
...
+                ImmutableMap.of(QualifiedObjectName.valueOf("tpch.s1.t13"), ImmutableSet.of("x", "y")));
+    }
+
+    @Test
+    public void testMultipleCtesDifferentNameDifferentTable()
+    {
+        assertUtilizedTableColumns(
+                "with cte1 as (select x + 1 as c1 from t13), cte2 as (with cte3 as (select a + 1 as c2, b + 1 as c3, c as c4 from t1) select * from cte3) select cte1.c1 from cte1 join cte2 on true",
+                ImmutableMap.of(QualifiedObjectName.valueOf("tpch.s1.t13"), ImmutableSet.of("x"),
+                        QualifiedObjectName.valueOf("tpch.s1.t1"), ImmutableSet.of()));
```
rschlussel, line 582: is this empty because cte2 isn't really used? maybe as a follow up we should have some logic that removes the map entry if the selected columns are empty.

presto-main/src/test/java/com/facebook/presto/sql/analyzer/TestUtilizedColumnsAnalyzer.java
```diff
@@ -519,6 +519,69 @@ public void testUDF()
...
+    {
+        // TODO(kevintang2022): Fix visitWithQuery so that it looks up relation properly.
+        // Currently, it just looks the relations using the name (string) of the CTE
+        assertUtilizedTableColumns(
+                "with cte1 as (select x + 1 as c1 from t13), cte2 as (with cte1 as (select x + 1 as c2, y + 1 as c3, z as c4 from t13) select * from cte1) select cte1.c1, cte2.c3 from cte1 join cte2 on true",
+                ImmutableMap.of(QualifiedObjectName.valueOf("tpch.s1.t13"), ImmutableSet.of("x", "y", "z")));
+    }
+
+    @Test
+    public void testMultipleCtesSameNameDifferentTable()
```
rschlussel, line 560: can you also add a test that uses columns from both tables in the query?  (and then presumably needs to fall back to checking all columns)?

presto-main/src/test/java/com/facebook/presto/sql/analyzer/TestUtilizedColumnsAnalyzer.java
kevintang2022: cte2 is not affecting the final results of the query. You can see that the only column is cte1.c1. Yes we can add a followup if this is not the expected behavior

rschlussel: just because at least for our internal permissions service, when we don't specify any columns, we check all columns.  Instead we could skip the check altogether in this case.

presto-main/src/test/java/com/facebook/presto/sql/analyzer/TestUtilizedColumnsAnalyzer.java
rschlussel: i think the part that's not as intuitive is that cte.c1 gets mapped to the first column in the other cte1 (even though it has a different name). I think a comment here explaining will be helpful (with the todo link like you have above to improve this)

presto-main/src/test/java/com/facebook/presto/sql/analyzer/TestUtilizedColumnsAnalyzer.java
kevintang2022: This is the same behavior for join queries like "select t1.a from t1 join t13 on true" regardless of wrapping in CTE

I will add a comment, but it seems like a seperate issue

rschlussel: yeah, agreed. it's a separate issue.

rschlussel (commented):
looks good, thanks! let's find someone else to review. (i can't approve because it's my PRO)

jaystarshot (approved):
Please add a Release note

PR #17367, opened 2022-02-28
adapt disable filesystem cache

gjhkael:
If disabled the filesystem cache, then presto can not construct the FileSystem.
 
[issue-17366](https://github.com/prestodb/presto/issues/17366)

Issue #17366, opened 2022-02-28
Presto can not work when disabled cache of hdfs or viewfs


gjhkael:
When i disabled the filesystem cache with the following conf, then the presto can not create FileSystem object, with the following exception:
```shell
<name>fs.viewfs.impl.disable.cache</name>
    <value>true</value>
    </property>
  <property>
    <name>fs.hdfs.impl.disable.cache</name>
    <value>true</value>
  </property>
```

```shell
com.facebook.presto.spi.PrestoException: HIVE_UNKNOWN_ERROR
	at com.facebook.presto.hive.BackgroundHiveSplitLoader$HiveSplitLoaderTask.process(BackgroundHiveSplitLoader.java:128)
	at com.facebook.presto.hive.util.ResumableTasks.safeProcessTask(ResumableTasks.java:47)
	at com.facebook.presto.hive.util.ResumableTasks.access$000(ResumableTasks.java:20)
	at com.facebook.presto.hive.util.ResumableTasks$1.run(ResumableTasks.java:35)
	at com.facebook.airlift.concurrent.BoundedExecutor.drainQueue(BoundedExecutor.java:78)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.IllegalStateException: undefined
	at com.google.common.base.Preconditions.checkState(Preconditions.java:491)
	at com.facebook.presto.hive.HdfsEnvironment.lambda$getFileSystem$0(HdfsEnvironment.java:73)
	at com.facebook.presto.hive.authentication.UserGroupInformationUtils.lambda$executeActionInDoAs$0(UserGroupInformationUtils.java:29)
	at java.security.AccessController.doPrivileged(Native Method)
	at javax.security.auth.Subject.doAs(Subject.java:360)
	at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1726)
	at com.facebook.presto.hive.authentication.UserGroupInformationUtils.executeActionInDoAs(UserGroupInformationUtils.java:27)
	at com.facebook.presto.hive.authentication.ImpersonatingHdfsAuthentication.doAs(ImpersonatingHdfsAuthentication.java:39)
[...]

PR #24650, opened 2025-02-28
auto-wrap non-ExtendedFileSystem instances

lutongzero:
## Description
- Automatically wrap non-`ExtendedFileSystem` instances with `HadoopExtendedFileSystem` adapter  

- ​**Fixes**​ #17366 (Filesystem compatibility exceptions)  
- ​**Resubmits**​ #17367

### DIFF
--- a/presto-hive/src/test/java/com/facebook/presto/hive/TestFileSystemCache.java
+++ b/presto-hive/src/test/java/com/facebook/presto/hive/TestFileSystemCache.java
@@ -61,8 +61,7 @@ public class TestFileSystemCache
                         new HiveHdfsConfiguration(new HdfsConfigurationInitializer(hiveClientConfig, new MetastoreClientConfig()), ImmutableSet.of(), hiveClientConfig),
                         new MetastoreClientConfig(),
                         new ImpersonatingHdfsAuthentication(new SimpleHadoopAuthentication()));
-        String filePath = Resources.getResource("test_file_system_cache/id.parquet").getPath();
-        Path path = new Path(filePath);
+        Path path = new Path(Files.createTempFile("id-", ".parquet").toUri());
 
         Configuration defaultConf = new Configuration(true);
         FileSystem fs1 = environment.getFileSystem("user", path, defaultConf);


lutongzero:
Hi @jaystarshot , could you help review this PR when you get a chance?

jaystarshot:
Is it possible to add a unit test?

jaystarshot:
Though looks good to me

presto-hive-metastore/src/main/java/com/facebook/presto/hive/HdfsEnvironment.java
```diff
@@ -70,7 +70,9 @@ public ExtendedFileSystem getFileSystem(String user, Path path, Configuration co
         return hdfsAuthentication.doAs(user, () -> {
             FileSystem fileSystem = path.getFileSystem(configuration);
             fileSystem.setVerifyChecksum(verifyChecksum);
-            checkState(fileSystem instanceof ExtendedFileSystem);
+            if (!(fileSystem instanceof ExtendedFileSystem)) {
```
ScrapCodes, line 73: Correct me if I am wrong, the very reason these checks exist is  - filesystem should be already wrapped at this point. The wrapping happens at ` HadoopExtendedFileSystemCache.initialize();` at line 35.

ScrapCodes: Just wondering what is causing the wrapping to be missed.

lutongzero: The provided code from Hadoop's FileSystem.get() method confirms the caching behavior:
```
public static FileSystem get(URI uri, Configuration conf) throws IOException {
    // ... [scheme handling logic] ...

    String disableCacheName = String.format("fs.%s.impl.disable.cache", scheme);
    if (conf.getBoolean(disableCacheName, false)) {
        return createFileSystem(uri, conf); // Bypasses cache
    }
    
    return CACHE.get(uri, conf); // Uses cache
}
```
When fs.%s.impl.disable.cache returns true, the caching mechanism will be bypassed. This occurs regardless of whether it's explicitly set through configuration or uses the default value.

lutongzero:
> Is it possible to add a unit test?

Hi @jaystarshot  I've added the unit tests as requested. Please review again when you have time. If there are any further issues, feel free to let me know! Thanks!

presto-hive/src/main/java/com/facebook/presto/hive/cache/HiveCachingHdfsConfiguration.java
```diff
@@ -71,7 +71,9 @@ public Configuration getConfiguration(HdfsContext context, URI uri)
                     currentConfig = hiveHdfsConfiguration.getConfiguration(context, factoryUri);
                 }
                 FileSystem fileSystem = (new Path(factoryUri)).getFileSystem(currentConfig);
-                checkState(fileSystem instanceof ExtendedFileSystem);
+                if (!(fileSystem instanceof ExtendedFileSystem)) {
+                    fileSystem = new HadoopExtendedFileSystem(fileSystem);
+                }
```
hantangwangd, line 76: Is this change necessary? As I see, the newly added test succeeds as well without it. If we really need this, can you also add a test which will fail without this.

presto-hive/src/test/java/com/facebook/presto/hive/TestFileSystemCache.java
```diff
@@ -50,6 +51,36 @@ public void testFileSystemCache()
...
+            throws IOException
+    {
+        HiveClientConfig hiveClientConfig = new HiveClientConfig();
+        HdfsEnvironment environment =
+                new HdfsEnvironment(
+                        new HiveHdfsConfiguration(new HdfsConfigurationInitializer(hiveClientConfig, new MetastoreClientConfig()), ImmutableSet.of(), hiveClientConfig),
+                        new MetastoreClientConfig(),
+                        new ImpersonatingHdfsAuthentication(new SimpleHadoopAuthentication()));
+        String filePath = Resources.getResource("test_file_system_cache/id.parquet").getPath();
+        Path path = new Path(filePath);
```
hantangwangd, line 65:
```suggestion
        Path path = new Path(Files.createTempFile("id-", ".parquet").toUri());
```

Could we simply use a temp file here?

presto-hive/src/test/java/com/facebook/presto/hive/TestFileSystemCache.java
```diff
@@ -50,6 +51,36 @@ public void testFileSystemCache()
...
+        Configuration defaultConf = new Configuration(true);
+        FileSystem fs1 = environment.getFileSystem("user", path, defaultConf);
+        FileSystem fs2 = environment.getFileSystem("user", path, defaultConf);
+        assertSame(fs1, fs2);
+
+        Configuration cacheDisableConf = new Configuration(true);
+        cacheDisableConf.set("fs.file.impl.disable.cache", "true");
+        FileSystem fs3 = environment.getFileSystem("user", path, cacheDisableConf);
+        FileSystem fs4 = environment.getFileSystem("user", path, cacheDisableConf);
+        assertNotSame(fs3, fs4);
```
hantangwangd, line 76: I think we can test `HDFS` and `S3` path as well through setting the following config:

```
cacheDisableConf.set("fs.hdfs.impl.disable.cache", "true");
```

and 

```
defaultConf.set("fs.s3.impl", PrestoS3FileSystem.class.getName());
......
cacheDisableConf.set("fs.s3.impl", PrestoS3FileSystem.class.getName());
cacheDisableConf.set("fs.s3.impl.disable.cache", "true");
```

hantangwangd (changes requested):
Thanks for this fix. The change overall looks good to me, just some little things.

PR #24655, opened 2025-02-28
[sidecar] Fix array_sort lambda function failure when sidecar is enabled and add e2e lambda function tests

pdabre12:
## Description
Introduced a new exception `SignatureMatchingException`. 
When identifying applicable functions, we capture all the `SemanticExceptions` in a list and rethrow an exception based on the number of exceptions captured.
In the case of one exception , we rethrow the `SemanticException`.
If there are >1 exceptions we throw the new exception `SignatureMatchingException` with a neatly formatted message showcasing all the encountered exceptions.

## Motivation and Context
Resolves: https://github.com/prestodb/presto/issues/24908

When the sidecar is enabled, function signatures are retrieved from the native execution engine. These signatures may differ from those supported by Presto Java. One such example is the `array_sort` function, which supports multiple lambda function signatures:
	•	`array_sort(array(T))`
	•	`array_sort(array(T), function(T, U))`
	•	`array_sort(array(T), function(T, T, bigint))`
We should be able to support all these function signatures.

For example:
`SELECT array_sort(quantities, (x, y) -> if (x < y, 1, if (x > y, -1, 0))) FROM orders_ex` which correctly identifies `array_sort(array(T), function(T, T, bigint))` as an applicable function , but will also try to match the function signature to` array_sort(array(T), function(T, U))` which 
would throw `Expected a lambda that takes 1 argument\\(s\\) but got 2`

Alternatively, 
`SELECT array_sort(quantities, x -> abs(x)) FROM orders_ex` which correctly identifies `array_sort(array(T), function(T, U))` as an applicable function , but will also try to match the function signature to` array_sort(array(T), function(T, T, bigint))` which 
would throw `Expected a lambda that takes 2 argument\\(s\\) but got 1`

Under the hood, the coordinator correctly determines the applicable lambda function signature, but during function identification, it also attempts to resolve the other, incompatible signature, resulting in an error.

The reason why Presto java does not run into the same issue is because it supports one lambda signature for array_sort :
`array_sort(array(T), function(T, T, int))`

## Impact
Users would be able to see all the exceptions thrown in the coordinator while identifying applicable functions.

## Test Plan
E2e tests included

## Contributor checklist

- [x] CI passed.

```
== NO RELEASE NOTE ==
```


pdabre12:
@ZacBlanco Can you please have a look?

presto-main/src/main/java/com/facebook/presto/metadata/FunctionSignatureMatcher.java
```diff
@@ -124,15 +126,29 @@ private Optional<Signature> matchFunction(Collection<? extends SqlFunction> cand
...
+                boundSignature.ifPresent(signature -> applicableFunctions.add(new ApplicableFunction(declaredSignature, signature, function.isCalledOnNullInput())));
             }
+            catch (SemanticException e) {
+                if (e.getCode().equals(INVALID_PARAMETER_USAGE)) {
+                    invalidParameterUsageException = Optional.of(e);
+                    continue;
+                }
+                throw e;
+            }
+        }
```
ZacBlanco, line 144: I'm wondering whether this solution is generic enough to apply to all cases. Will it only catch mis-uses of SQL functions which take varying types of lambdas as argument? What about other functions which accept generic arguments?

I can't think of a concrete case where we would see a situation other than what you described for the issue with `array_sort`, but this catch statement and checking of the error code, and then conditionally re-throwing doesn't sit right with me because it seems very specific to the error of binding arguments multiple times.

Is there a reason we can't just always `continue` on every exception, but add them all to a list and throw at the end if there are no applicable functions? I am not deeply familiar with how we decide a function is applicable, so I may not be understanding the full picture

Here's some pseudo-code to describe the approach I'm thinking of:


```
applicableFunctions = []
semanticExceptions = []
for signature in signatures:
    if signature.doesBind(actualParameters):
       applicableFunctions.add(signature)
    else:
        semanticExceptions.add(/* exception */)

if applicableFunctions.isEmpty() && semanticExceptions.size() = 1:
    throw semanticExceptions[0]
else if applicableFunctions.isEmpty() && semanticExceptions.isEmpty():
    /* handle this somehow? */
else if applicableFunctions.isEmpty() && semanticExceptions.size() > 1:
      /* find a good error to throw here */

return applicableFunctions

```

pdabre12: @ZacBlanco Thanks for the suggestion.
I made the necessary changes with one change.

I thought about this for a while and even I wasn't able to think of another concrete case where we would see this situation again, so I am assuming that we have one exception in most of the cases. If there's one exception, I am still throwing only that Exception back, but if there's multiple  we throws a `SignatureMatchingException`.

presto-analyzer/src/main/java/com/facebook/presto/sql/analyzer/SignatureMatchingException.java
```diff
@@ -0,0 +1,66 @@
...
+{
+    private final SemanticErrorCode code;
+
+    public SignatureMatchingException(
+            SemanticErrorCode code,
+            String format,
+            List<SemanticException> failedExceptions)
+    {
+        super(code, formatMessage(format, failedExceptions));
+        this.code = requireNonNull(code, "code is null");
```
ZacBlanco, line 33: do we even need to have a code parameter if it's always the same?

presto-analyzer/src/main/java/com/facebook/presto/sql/analyzer/SignatureMatchingException.java
```diff
@@ -0,0 +1,66 @@
...
+    {
+        return code;
+    }
+
+    /**
+      Decides which exception to throw based on the number of failed attempts.
+      If there's only one failure, it returns that failure directly.
+      If there are multiple, it returns the SignatureMatchingException.
+
+      @return The appropriate SemanticException to throw.
```
ZacBlanco, line 46: technically, this method doesn't return anything...consider `@throws`?

presto-native-sidecar-plugin/src/test/java/com/facebook/presto/sidecar/TestNativeSidecarPlugin.java
```diff
@@ -187,10 +187,20 @@ public void testWindowFunctions()
     }
 
     @Test
-    public void testArraySort()
-    {
-        assertQueryFails("SELECT array_sort(quantities, (x, y) -> if (x < y, 1, if (x > y, -1, 0))) FROM orders_ex",
-                "line 1:31: Expected a lambda that takes 1 argument\\(s\\) but got 2");
+    public void testLambdaFunctions()
```
ZacBlanco, line 190: I think we need to have a negative case where matching fails and we can see the expected exceptions. Ideally testing cases of both 1 and 2 exceptions. You might need to register some custom functions or something to do this

presto-analyzer/src/main/java/com/facebook/presto/sql/analyzer/SemanticException.java
```diff
@@ -61,6 +71,11 @@ public SemanticErrorCode getCode()
         return code;
     }
 
+    public String getFormattedMessage()
+    {
+        return getMessage();
+    }
+
```
ZacBlanco, line 78: is this method even necessary?

ZacBlanco (changes requested):
Thanks for the changes @pdabre12

presto-analyzer/src/main/java/com/facebook/presto/sql/analyzer/SemanticException.java
pdabre12: Removed.

presto-native-sidecar-plugin/src/test/java/com/facebook/presto/sidecar/TestNativeSidecarPlugin.java
pdabre12: Added a test case for >1 exceptions.
Added a tests for `array_sort` lambda signature with 3 args, it would throw exceptions for both the lambda signatures supported for `array_sort`.

presto-analyzer/src/main/java/com/facebook/presto/sql/analyzer/SignatureMatchingException.java
```diff
@@ -0,0 +1,56 @@
...
+     * If there's only one SemanticException, it throws that SemanticException directly.
+     * If there are multiple SemanticExceptions, it throws the SignatureMatchingException.
+     */
+    public static void decideAndThrow(List<SemanticException> failedExceptions)
+            throws SemanticException
+    {
+        if (failedExceptions.size() == 1) {
+            throw failedExceptions.get(0);
+        }
+        else {
```
pdabre12, line 43: I was thinking what could be a good base message for this.
Maybe `Exceptions encountered while identifying applicable functions: `

pdabre12:
@ZacBlanco Thanks for the review. 
Addressed your comments, PTAL.

ZacBlanco (changes requested):
A few more minor things. Tests cases look good now. Thanks for the updates!

pdabre12:
@ZacBlanco 
Thank you, addressed your comments.
Please have another look.

presto-analyzer/src/main/java/com/facebook/presto/sql/analyzer/SignatureMatchingException.java
```diff
@@ -0,0 +1,56 @@
...
+    {
+        super(code, formatMessage(format, failedExceptions));
+    }
+
+    /**
+     * Decides which exception to throw based on the number of failed attempts.
+     * If there's only one SemanticException, it throws that SemanticException directly.
+     * If there are multiple SemanticExceptions, it throws the SignatureMatchingException.
+     */
+    public static void decideAndThrow(List<SemanticException> failedExceptions)
```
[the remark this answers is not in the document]
pdabre12: Done.

presto-analyzer/src/main/java/com/facebook/presto/sql/analyzer/SignatureMatchingException.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.List;
+
+import static com.facebook.presto.sql.analyzer.SemanticErrorCode.EXCEPTIONS_WHEN_RESOLVING_FUNCTIONS;
+import static java.lang.String.format;
+
+public class SignatureMatchingException
+        extends SemanticException
+{
+    public SignatureMatchingException(
+            SemanticErrorCode code,
```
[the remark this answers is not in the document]
pdabre12: Done.

presto-analyzer/src/main/java/com/facebook/presto/sql/analyzer/SemanticException.java
```diff
@@ -44,10 +46,18 @@ public SemanticException(SemanticErrorCode code, Optional<NodeLocation> location
...
         this.code = requireNonNull(code, "code is null");
         this.location = requireNonNull(location, "location is null");
+        this.formattedMessage = requireNonNull(format, "format is null");
+        this.args = requireNonNull(args, "args is null");
+    }
+
+    @Override
+    public String getMessage()
+    {
+        return formatMessage(formattedMessage, location, args);
```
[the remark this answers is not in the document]
pdabre12: Done.

ZacBlanco (approved):

pdabre12:
@jaystarshot Can you please have a look?

aditi-pandit:
@pdabre12 : Just curious did array_sort with the non-Presto Java supported signature work previously in Prestissimo at all ?

aditi-pandit (approved):
Thanks @pdabre12

PR #24548, opened 2025-02-13
Revert event loop related changes

shangm2:
## Description
1. we see some issues at certain clusters where workers are not receiving any splits. Reverting the following two prs back to unblock others so I can take time to root cause it.
    a. #24412
    b. #24288 

## Motivation and Context
1. reverting event loop related changes to unblock others.

## Contributor checklist

- [ ] Please make sure your submission complies with our [contributing guide](https://github.com/prestodb/presto/blob/master/CONTRIBUTING.md), in particular [code style](https://github.com/prestodb/presto/blob/master/CONTRIBUTING.md#code-style) and [commit standards](https://github.com/prestodb/presto/blob/master/CONTRIBUTING.md#commit-standards).
- [ ] PR description addresses the issue accurately and concisely.  If the change is non-trivial, a GitHub Issue is referenced.
- [ ] Documented new properties (with its default value), SQL syntax, functions, or other functionality.
- [ ] If release notes are required, they follow the [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines).
- [ ] Adequate tests were added if applicable.
- [ ] CI passed.

## Release Notes
Please follow [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines) and fill in the release notes below.

```
== RELEASE NOTES ==

General Changes
* ... 
* ... 

Hive Connector Changes
* ... 
* ... 
```

If release note is NOT required, use:

```
== NO RELEASE NOTE ==
```

PR #24668, opened 2025-03-03
Put event loop behind a toggle

shangm2:
## Description
1. This pr proposed to bring back event loop implementation for HttpRemoteTask.java but have it behind a feature toggle. Huge thanks to @arhimondr on his comments here https://github.com/prestodb/presto/pull/24548#issuecomment-2671515802
 
2. This pr basically bring back the following two prs: #24288 and #24412 

## Motivation and Context
1. we saw a big number of tasks were created for certain queries which causes contention on the underlying synchronous queue
2. this pr proposed to use a event loop to run task's method so that multiple tasks can be assigned to the same event loop and a task's method will be executed on the same thread
4. some methods within http remote task does not need the "synchronized" keywords if we are certain they will be executed in the same thread.

## Test Plan
1. Running the verifier tests on two cluster with different config
2. Confirmed that the config can be used to enable/disable the event loop for clusters
 
Two different clusters have different config
Screenshot 2025-03-03 at 17 56 38

For the one with event loop enabled
Screenshot 2025-03-03 at 18 05 53

for the one without event loop
Screenshot 2025-03-03 at 18 07 16
And both pass the verifier runs:
https://www.internalfb.com/intern/presto/verifier/results/?test_id=212715
https://www.internalfb.com/intern/presto/verifier/results/?test_id=212714

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
General Changes
* Improve efficiency of coordinator when running a large number of tasks by adding a toggle-controlled event loop implementation for HttpRemoteTask.java.
```

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTaskFactory.java
+++ b/presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTaskFactory.java
@@ -31,6 +31,7 @@ import com.facebook.presto.execution.QueryManager;
 import com.facebook.presto.execution.QueryManagerConfig;
 import com.facebook.presto.execution.RemoteTask;
 import com.facebook.presto.execution.RemoteTaskFactory;
+import com.facebook.presto.execution.SafeEventLoopGroup;
 import com.facebook.presto.execution.SchedulerStatsTracker;
 import com.facebook.presto.execution.TaskId;
 import com.facebook.presto.execution.TaskInfo;
@@ -49,19 +50,24 @@ import com.facebook.presto.server.TaskUpdateRequest;
 import com.facebook.presto.spi.plan.PlanNodeId;
 import com.facebook.presto.sql.planner.PlanFragment;
 import com.google.common.collect.Multimap;
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import io.airlift.units.Duration;
+import io.netty.channel.EventLoop;
+import io.netty.util.concurrent.AbstractEventExecutorGroup;
 import org.weakref.jmx.Managed;
 import org.weakref.jmx.Nested;
 
 import javax.annotation.PreDestroy;
 import javax.inject.Inject;
 
+import java.util.Optional;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadPoolExecutor;
 
 import static com.facebook.airlift.concurrent.Threads.daemonThreadsNamed;
+import static com.facebook.presto.server.remotetask.HttpRemoteTaskWithEventLoop.createHttpRemoteTaskWithEventLoop;
 import static com.facebook.presto.server.thrift.ThriftCodecWrapper.wrapThriftCodec;
 import static java.lang.Math.toIntExact;
 import static java.util.Objects.requireNonNull;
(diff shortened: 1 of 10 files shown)


presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTaskFactory.java
```diff
@@ -222,44 +240,89 @@ public RemoteTask createRemoteTask(
...
-                maxTaskUpdateSizeInBytes,
-                metadataManager,
-                queryManager,
-                taskUpdateRequestSize,
-                taskUpdateSizeTrackingEnabled,
-                handleResolver,
-                connectorTypeSerdeManager,
-                schedulerStatsTracker);
+        RemoteTask remoteTask;
+        if (!this.isEventLoopEnabled) {
```
NikhilCollooru, line 244: no need of if and else.. just do 

```
if(isEventLoopEnabled) {
    return new HttpRemoteTask(...,eventloop,...)
}

return  new HttpRemoteTask(..the old way...)
```

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTaskFactory.java
```diff
@@ -186,6 +192,16 @@ else if (binaryTransportEnabled) {
         this.errorScheduledExecutor = newSingleThreadScheduledExecutor(daemonThreadsNamed("remote-task-error-delay-%s"));
         this.taskUpdateRequestSize = new DecayCounter(ExponentialDecay.oneMinute());
         this.taskUpdateSizeTrackingEnabled = taskConfig.isTaskUpdateSizeTrackingEnabled();
+        this.eventLoopGroup = new SafeEventLoopGroup(config.getRemoteTaskMaxCallbackThreads(),
```
NikhilCollooru, line 195: Will this spawn threads irrespective of whether eventloop is enabled or not ?

Wondering if we should initialize this eventLoopGroup only if isEventLoopEnabled==true.

steveburnett:
Thanks for the release note entry! Consider adding some more detail, maybe something like this?

```
General Changes
* Improve efficiency of coordinator when running a large number of tasks by adding a toggle-controlled event loop implementation for HttpRemoteTask.java.
```

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTaskFactory.java
shangm2: Good point.

presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTaskFactory.java
```diff
@@ -222,7 +241,50 @@ public RemoteTask createRemoteTask(
             TableWriteInfo tableWriteInfo,
             SchedulerStatsTracker schedulerStatsTracker)
     {
-        return new HttpRemoteTask(
+        if (!eventLoopGroup.isPresent()) {
```
NikhilCollooru, line 244: nit: can we do the other way. check if eventLoop is enabled in the if condition and return remoteTask accordinglt.

`if (!eventLoopGroup.isPresent()) { `     ---> `if (eventLoopGroup.isPresent()) { return eventLoopRemoteTask}`

shangm2: @NikhilCollooru  updated. Thanks for the comments!

NikhilCollooru (approved):
LGTM

Issue #24597, opened 2025-02-20
Support choosing compression codecs for exchange


yingsu00:
Compression might benefit queries with large amount of data bring shuffled, like TPCH Q9. However, 

1) Presto defaults to LZ4 and users don't have a way to specify other codecs. See https://github.com/prestodb/presto/blob/a766a3a1ecd5fa45426977bc05e707da5d1c138c/presto-main/src/main/java/com/facebook/presto/execution/buffer/PagesSerdeFactory.java#L66

2) Velox defaults to none and no compression is made at all. See PartitionedOutput.cpp
```
if (serde_->kind() == VectorSerde::Kind::kPresto) {
      serializer::presto::PrestoVectorSerde::PrestoOptions options;
      options.compressionKind =
          OutputBufferManager::getInstance().lock()->compressionKind();
      options.minCompressionRatio = PartitionedOutput::minCompressionRatio();
      current_->createStreamTree(rowType, rowsInCurrent_, &options);
    }
```
Here OutputBufferManager::getInstance() uses the default Option that doesn't compress. 
```
struct Options {
    common::CompressionKind compressionKind{
        common::CompressionKind::CompressionKind_NONE};
  };
```

3) THere are not enough tests for different compression codecs for exchange.

## Expected Behavior or Use Case
Create a config property "exchange.compression-codec" that chooses from an enum. This will replace the current Presto "exchange.compression-enabled". Introduce the same for Presto CPP and pass it to Velox.

## Presto Component, Service, or Connector
Exchange

## Possible Implementation
Just straightforward implementation

## Context
This is one part of the exchange optimization. When working on it I discovered the missing functionality.

xin-zhang2:
@yingsu00 

1. Velox has added support for shuffle compression codec in facebookincubator/velox#11914.
```
std::unique_ptr<VectorSerde::Options> getVectorSerdeOptions(
    const core::QueryConfig& queryConfig,
    VectorSerde::Kind kind) {
  std::unique_ptr<VectorSerde::Options> options =
      kind == VectorSerde::Kind::kPresto
      ? std::make_unique<serializer::presto::PrestoVectorSerde::PrestoOptions>()
      : std::make_unique<VectorSerde::Options>();
  options->compressionKind =
      common::stringToCompressionKind(queryConfig.shuffleCompressionKind());
  return options;
}
```
[...]

PR #24670, opened 2025-03-04
Add compression codec support for PageSerde

xin-zhang2:
## Description

The compression codec for PageSerde is currently hardcoded as LZ4. This PR adds supports for additional codecs, including ZSTD, SNAPPY, LZO, GZIP and ZLIB. 

## Motivation and Context

#24597 

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Replace `exchange.compression-enabled`,  `fragment-result-cache.block-encoding-compression-enabled`, `experimental.spill-compression-enabled` with `exchange.compression-codec`, `fragment-result-cache.block-encoding-compression-codec` to enable compression codec configurations. Supported codecs include GZIP, LZ4, LZO, SNAPPY, ZLIB and ZSTD.
```

### DIFF
--- a/presto-docs/src/main/sphinx/admin/properties.rst
+++ b/presto-docs/src/main/sphinx/admin/properties.rst
@@ -388,13 +388,14 @@ Limit for memory used for unspilling a single aggregation operator instance.
 
 The corresponding session property is :ref:`admin/properties-session:\`\`aggregation_operator_unspill_memory_limit\`\``. 
 
-``experimental.spill-compression-enabled``
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+``experimental.spill-compression-codec``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-* **Type:** ``boolean``
-* **Default value:** ``false``
+* **Type:** ``string``
+* **Allowed value:** ``NONE``, ``LZ4``
+* **Default value:** ``NONE``
 
-Enables data compression for pages spilled to disk.
+The data compression codec to be used for pages spilled to disk.
 
 ``experimental.spill-encryption-enabled``
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/presto-docs/src/main/sphinx/admin/properties.rst
+++ b/presto-docs/src/main/sphinx/admin/properties.rst
@@ -392,7 +392,7 @@ The corresponding session property is :ref:`admin/properties-session:\`\`aggrega
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 * **Type:** ``string``
-* **Allowed value:** ``NONE``, ``LZ4``
+* **Allowed value:** ``SNAPPY``, ``NONE``, ``GZIP``, ``LZ4``, ``LZO``,, ``ZLIB`` ``ZSTD``
 * **Default value:** ``NONE``
 
 The data compression codec to be used for pages spilled to disk.
--- a/presto-native-execution/presto_cpp/main/QueryContextManager.cpp
+++ b/presto-native-execution/presto_cpp/main/QueryContextManager.cpp
@@ -212,19 +212,17 @@ QueryContextManager::toVeloxConfigs(
       traceFragmentId = it.second;
     } else if (it.first == SessionProperties::kQueryTraceShardId) {
       traceShardId = it.second;
-    } else if (it.first == SessionProperties::kShuffleCompressionEnabled) {
-      if (it.second == "true") {
-        // NOTE: Presto java only support lz4 compression so configure the same
-        // compression kind on velox.
-        configs[core::QueryConfig::kShuffleCompressionKind] =
-            velox::common::compressionKindToString(
-                velox::common::CompressionKind_LZ4);
-      } else {
-        VELOX_USER_CHECK_EQ(it.second, "false");
-        configs[core::QueryConfig::kShuffleCompressionKind] =
-            velox::common::compressionKindToString(
-                velox::common::CompressionKind_NONE);
-      }
+    } else if (it.first == SessionProperties::kShuffleCompressionCodec) {
+      auto compression = it.second;
+      std::transform(
+          compression.begin(),
+          compression.end(),
+          compression.begin(),
+          ::tolower);
+      velox::common::CompressionKind compressionKind =
+          common::stringToCompressionKind(compression);
+      configs[core::QueryConfig::kShuffleCompressionKind] =
+          velox::common::compressionKindToString(compressionKind);
     } else {
       configs[sessionProperties_.toVeloxConfig(it.first)] = it.second;
       sessionProperties_.updateVeloxConfig(it.first, it.second);
--- a/presto-docs/src/main/sphinx/admin/spill.rst
+++ b/presto-docs/src/main/sphinx/admin/spill.rst
@@ -93,10 +93,10 @@ there is no need to use RAID for spill.
 Spill Compression
 -----------------
 
-When spill compression is enabled (``spill-compression-enabled`` property in
-:ref:`tuning-spilling`), spilled pages will be compressed using the same
-implementation as exchange compression when they are sufficiently compressible.
-Enabling this feature can reduce the amount of disk IO at the cost
+When :ref:`admin/properties:\`\`experimental.spill-compression-codec\`\`` is 
+configured, spilled pages are compressed using 
+the configured codec implementation when they are sufficiently compressible.
+This feature can reduce the amount of disk IO at the cost
 of extra CPU load to compress and decompress spilled pages.
 
 Spill Encryption
--- a/presto-jdbc/src/test/java/com/facebook/presto/jdbc/TestJdbcConnection.java
+++ b/presto-jdbc/src/test/java/com/facebook/presto/jdbc/TestJdbcConnection.java
@@ -302,7 +302,7 @@ public class TestJdbcConnection
         try (Connection connection = createConnection("sessionProperties=query_max_run_time:2d;max_failed_task_percentage:0.6")) {
             assertThat(listSession(connection))
                     .contains("join_distribution_type|AUTOMATIC|AUTOMATIC")
-                    .contains("exchange_compression|false|false")
+                    .contains("exchange_compression_codec|NONE|NONE")
                     .contains("query_max_run_time|2d|100.00d")
                     .contains("max_failed_task_percentage|0.6|0.3");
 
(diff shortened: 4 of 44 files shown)


steveburnett (changes requested):
Thanks for the doc! Added a direct link to the property so the reader doesn't have to scroll through the topic. See screenshot. 
Screenshot 2025-03-10 at 9 48 05 AM

steveburnett (dismissed):
LGTM! (docs) 

presto-main/src/main/java/com/facebook/presto/execution/buffer/GzipCompressor.java
```diff
@@ -0,0 +1,68 @@
...
+import io.airlift.compress.Compressor;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.Buffer;
+import java.nio.ByteBuffer;
+import java.util.zip.GZIPOutputStream;
+
+public class GzipCompressor
```
yingsu00, line 25: Actually gzip is a file format, not a compression algorithm. And this class is essentially the same as your DeflateCompressor. I think it's better to keep only one and name it ZlibCompressor.

presto-main/src/test/java/com/facebook/presto/spiller/TestFileSingleStreamSpiller.java
```diff
@@ -69,32 +70,32 @@ public void tearDown()
...
-        assertSpill(false, false);
+        assertSpill(CompressionCodec.NONE, false);
     }
 
     @Test
     public void testSpillCompression()
             throws Exception
     {
-        assertSpill(true, false);
+        assertSpill(CompressionCodec.LZ4, false);
```
yingsu00, line 80: Shall we test all codecs? But it's ok to do it in future PRs.

presto-native-execution/presto_cpp/main/QueryContextManager.cpp
```diff
@@ -198,19 +198,17 @@ QueryContextManager::toVeloxConfigs(
...
-                velox::common::CompressionKind_NONE);
-      }
+    } else if (it.first == SessionProperties::kShuffleCompressionCodec) {
+      auto compression = it.second;
+      std::transform(
+          compression.begin(),
+          compression.end(),
+          compression.begin(),
+          ::tolower);
+      velox::common::CompressionKind compressionKind =
```
yingsu00, line 208: Why do you do the string ->CompressionKind -> string conversion? Can we just use the lowercased string directly?

presto-main/src/main/java/com/facebook/presto/execution/buffer/DeflateCompressor.java
```diff
@@ -0,0 +1,83 @@
...
+
+import java.nio.Buffer;
+import java.nio.ByteBuffer;
+import java.util.OptionalInt;
+import java.util.zip.Deflater;
+
+import static java.util.Objects.requireNonNull;
+import static java.util.zip.Deflater.FULL_FLUSH;
+
+public class DeflateCompressor
```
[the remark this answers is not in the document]
yingsu00: In Velox, GZIP and ZLIB were separate. But they were both configured to use zlib in the writers

```
if (kind == velox::common::CompressionKind_ZLIB ||
      kind == velox::common::CompressionKind_GZIP) {
    options.format.zlib.windowBits = Compressor::DWRF_ORC_ZLIB_WINDOW_BITS;
    options.format.zlib.compressionLevel = zlibCompressionLevel;
  }
```
To avoid the mapping issue, let's just rename this as ZLib now.

presto-docs/src/main/sphinx/admin/properties.rst
```diff
@@ -388,13 +388,14 @@ Limit for memory used for unspilling a single aggregation operator instance.
 
 The corresponding session property is :ref:`admin/properties-session:\`\`aggregation_operator_unspill_memory_limit\`\``. 
 
-``experimental.spill-compression-enabled``
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+``experimental.spill-compression-codec``
```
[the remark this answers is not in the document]
xin-zhang2: @yingsu00 
Thanks for the review. I've made some adjustments to the commits. 
As we're changing the exchange compression property from a boolean to an enum,  the related field in PagesSerdeFactory needs to be updated accordingly. Since spill compression also relies on PagesSerdeFactory, I think it would be better to include them in a single commit. Would that be okay?

presto-main/src/main/java/com/facebook/presto/execution/buffer/AirliftCompressorAdapter.java
```diff
@@ -0,0 +1,56 @@
...
+
+import com.facebook.presto.spi.page.PageCompressor;
+import io.airlift.compress.Compressor;
+
+import java.nio.ByteBuffer;
+
+import static java.util.Objects.requireNonNull;
+
+public class AirliftCompressorAdapter
+        implements PageCompressor
```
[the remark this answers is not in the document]
xin-zhang2: presto-hive has a dependency on presto-main, but it is currently limited to the test scope. I’m not sure if we can broaden the dependency scope, so I duplicated the AirliftCompressorAdapter for now. Is it safe for us to remove the scope tag from the dependency?

presto-main/src/main/java/com/facebook/presto/execution/buffer/DeflateCompressor.java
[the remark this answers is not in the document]
xin-zhang2: Renamed it to ZLibCompressor.

presto-main/src/test/java/com/facebook/presto/spiller/TestFileSingleStreamSpiller.java
xin-zhang2: Added the all codecs tests for spill compression.

presto-native-execution/presto_cpp/main/QueryContextManager.cpp
xin-zhang2: I'm calling the stringToCompressionKind here just to validate the compression codec string.
```
CompressionKind stringToCompressionKind(const std::string& kind) {
  static const std::unordered_map<std::string, CompressionKind>
      stringToCompressionKindMap = {
          {"none", CompressionKind_NONE},
          {"zlib", CompressionKind_ZLIB},
          {"snappy", CompressionKind_SNAPPY},
          {"lzo", CompressionKind_LZO},
          {"zstd", CompressionKind_ZSTD},
          {"lz4", CompressionKind_LZ4},
          {"gzip", CompressionKind_GZIP}};
  auto iter = stringToCompressionKindMap.find(kind);
  if (iter != stringToCompressionKindMap.end()) {
    return iter->second;
  } else {
    VELOX_UNSUPPORTED("Not support compression kind {}", kind);
  }
}
```
Do you think it’s necessary to keep this check?

presto-main/src/main/java/com/facebook/presto/execution/buffer/GzipCompressor.java
xin-zhang2: I think we need keep both.
DeflateCompressor calls java.util.zip.Deflater.deflate(), which accepts a nowrap boolean flag to control compatibility with GZIP. When nowrap is set to true, the zlib header and checksum are omitted during compression.
```
    /**
     * Creates a new compressor using the specified compression level.
     * If 'nowrap' is true then the ZLIB header and checksum fields will
     * not be used in order to support the compression format used in
     * both GZIP and PKZIP.
     * @param level the compression level (0-9)
     * @param nowrap if true then use GZIP compatible compression
     */
    public Deflater(int level, boolean nowrap) {
        this.level = level;
        this.strategy = DEFAULT_STRATEGY;
        this.zsRef = new ZStreamRef(init(level, DEFAULT_STRATEGY, nowrap));
    }

```
This works well for compression and decompression within Presto Java, as we can set nowrap to true to make the output compatible with both zlib and gzip.
However, in Prestissmo, compression and decompression might occur on different sides. Since the C++ implementation includes the zlib or gzip headers and checksums, we must set nowrap to false for zlib compressor to ensure compatibility and add a separate compressor for gzip to properly include the gzip header and checksum.

presto-docs/src/main/sphinx/admin/properties.rst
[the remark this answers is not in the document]
yingsu00: We can have the changes in PagesSerdeFactory and the change in exchange in the first commit, then make the spilling change as the second commit

presto-main/src/main/java/com/facebook/presto/execution/buffer/AirliftCompressorAdapter.java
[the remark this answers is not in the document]
yingsu00: If it's only in test, then let's keep it duplicated.

presto-native-execution/presto_cpp/main/QueryContextManager.cpp
yingsu00: > I'm calling the stringToCompressionKind here just to validate the compression codec string.
> 
> ```
> CompressionKind stringToCompressionKind(const std::string& kind) {
>   static const std::unordered_map<std::string, CompressionKind>
>       stringToCompressionKindMap = {
>           {"none", CompressionKind_NONE},
>           {"zlib", CompressionKind_ZLIB},
>           {"snappy", CompressionKind_SNAPPY},
>           {"lzo", CompressionKind_LZO},
>           {"zstd", CompressionKind_ZSTD},
>           {"lz4", CompressionKind_LZ4},
>           {"gzip", CompressionKind_GZIP}};
>   auto iter = stringToCompressionKindMap.find(kind);
>   if (iter != stringToCompressionKindMap.end()) {
>     return iter->second;
>   } else {
>     VELOX_UNSUPPORTED("Not support compression kind {}", kind);
>   }
> }
> ```
> 
> Do you think it’s necessary to keep this check?

We can do the check but this two way conversion is very confusing. We can add a new function just to check if it's valid, and make stringToCompressionKind to call this new function. Then in your code call this new function too.

yingsu00:
@xin-zhang2 Could you please rebase?

presto-docs/src/main/sphinx/admin/properties.rst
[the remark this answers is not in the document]
xin-zhang2: I think the changes for PagesSerdeFactory, exchange and spilling need to be in one commit, as we are modifying the function signatures in PagesSerdeFactory, which are used by both exchange and spilling.

Currently, I’ve split this PR into three commits:
1. Refactor the compression codec field from a boolean to an enum in PagesSerdeFactory, along with the corresponding changes in exchange and spilling.
2. Add the implementation for various compression codecs support.
3. Address the review comments.

Does this seem reasonable?

presto-native-execution/presto_cpp/main/QueryContextManager.cpp
xin-zhang2: @yingsu00 
stringToCompressionKind is defined in velox, and I submit a PR https://github.com/facebookincubator/velox/pull/12886 for this.

xin-zhang2:
> @xin-zhang2 Could you please rebase?

Rebased.

yingsu00 (commented):
@xin-zhang2 Thanks for the update. I think separating it to two commits makes sense. Will you be able to squash the last one into the first two?

xin-zhang2:
> @xin-zhang2 Thanks for the update. I think separating it to two commits makes sense. Will you be able to squash the last one into the first two?

@yingsu00 Thanks for the suggestion. I have updated the commits.

yingsu00 (approved):
Thanks @xin-zhang2 Just one more thing: Could you please list the supported codecs in the release note? 
Also want to confirm that you have tested the compression works for native engine?

xin-zhang2:
@yingsu00 
The supported codecs have been added to the release note.

Yes, I have tested the compression with `exchange_compression_codec` session property for both C++ and Java engines.
In Presto C++, all codecs work correctly except for LZO. This is due to the lack of LZO support in folly codecs used by velox, and users will get an error like
```
Query 20250404_082337_00017_x7gk9 failed:  Not support lzo in folly
```
In Presto Java, all codecs work as expected.

PR #24673, opened 2025-03-04
Convert DateTime to long along the  critical path within coordinator

shangm2:
## Description
1. Similar to #24582 , we would like to continue replacing heavy data type with primitive data type.

## Motivation and Context
1. We would like to make coordinator leaner to save cpu time on objection creation and gc.

## Test Plan
1. passed java verifier test: https://www.internalfb.com/intern/presto/verifier/results/?test_id=213345
2. passed cpp worker verifier: https://www.internalfb.com/intern/presto/verifier/results/?test_id=213454
3. Will make sure UI does not break
Screenshot 2025-03-08 at 20 20 00
Screenshot 2025-03-08 at 20 20 19

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Improve scheduling by using long instead of DateTime for critical path.
```


presto-native-execution/presto_cpp/main/PrestoTask.cpp
```diff
@@ -313,9 +313,9 @@ void updatePipelineStats(
     protocol::PipelineStats& prestoPipelineStats) {
   prestoPipelineStats.inputPipeline = veloxPipelineStats.inputPipeline;
   prestoPipelineStats.outputPipeline = veloxPipelineStats.outputPipeline;
-  prestoPipelineStats.firstStartTime = prestoTaskStats.createTime;
-  prestoPipelineStats.lastStartTime = prestoTaskStats.endTime;
-  prestoPipelineStats.lastEndTime = prestoTaskStats.endTime;
+  prestoPipelineStats.firstStartTimeInMillis = prestoTaskStats.createTimeInMillis;
```
yuandagits, line 316: For my own learning, since there is a protocol change in the schema, do we need to ensure compatibility?

shangm2: Pls check presto_protocol_core.h change in the same commit.

presto-main/src/main/java/com/facebook/presto/operator/TaskStats.java
```diff
@@ -192,113 +318,123 @@ public TaskStats(
...
-        this.runningDrivers = runningDrivers;
-        checkArgument(runningPartitionedDrivers >= 0, "runningPartitionedDrivers is negative");
-        this.runningPartitionedDrivers = runningPartitionedDrivers;
-        checkArgument(runningPartitionedSplitsWeight >= 0, "runningPartitionedSplitsWeight must be non-negative");
-        this.runningPartitionedSplitsWeight = runningPartitionedSplitsWeight;
+        this(Optional.ofNullable(createTime).map(DateTime::getMillis).orElse(0L),
+                Optional.ofNullable(firstStartTime).map(DateTime::getMillis).orElse(0L),
+                Optional.ofNullable(lastStartTime).map(DateTime::getMillis).orElse(0L),
+                Optional.ofNullable(lastEndTime).map(DateTime::getMillis).orElse(0L),
+                Optional.ofNullable(endTime).map(DateTime::getMillis).orElse(0L),
```
ZacBlanco, line 325: In the interest of reducing object creation, we could just turn this into a utility method to avoid the `Optional` allocation throughout this PR.

```java
public static long getDateTimeMillis(@Nullable datetime) {
    if (datetime == null) {
         return 0L;
    }
    return datetime.getMillis();
}
```

ZacBlanco (changes requested):
Thanks for this PR. I have a few comments. Mostly pedantic issues

presto-main/src/main/java/com/facebook/presto/operator/TaskContext.java
```diff
@@ -94,9 +93,9 @@ public class TaskContext
     private final AtomicLong endFullGcCount = new AtomicLong(-1);
     private final AtomicLong endFullGcTimeNanos = new AtomicLong(-1);
 
-    private final AtomicReference<DateTime> executionStartTime = new AtomicReference<>();
-    private final AtomicReference<DateTime> lastExecutionStartTime = new AtomicReference<>();
-    private final AtomicReference<DateTime> executionEndTime = new AtomicReference<>();
+    private final AtomicReference<Long> executionStartTime = new AtomicReference<>();
+    private final AtomicReference<Long> lastExecutionStartTime = new AtomicReference<>();
+    private final AtomicReference<Long> executionEndTime = new AtomicReference<>();
```
NikhilCollooru, line 98: similar to previous comment. we can use `AtomicLong` here as well

presto-main/src/main/java/com/facebook/presto/operator/TaskContext.java
```diff
@@ -573,11 +572,11 @@ public TaskStats getTaskStats()
         boolean fullyBlocked = hasRunningPipelines && runningPipelinesFullyBlocked;
 
         return new TaskStats(
-                taskStateMachine.getCreatedTime(),
-                executionStartTime.get(),
-                lastExecutionStartTime.get(),
-                lastExecutionEndTime == 0 ? null : new DateTime(lastExecutionEndTime),
-                executionEndTime.get(),
+                taskStateMachine.getCreatedTimeInMillis(),
+                Optional.ofNullable(executionStartTime.get()).orElse(0L),
```
NikhilCollooru, line 576: we are already defining it above as `AtomicReference<Long> executionStartTime = new AtomicReference<>();`....will it be null ever ? do we need Optional.ofNullable ?

shangm2:
@ZacBlanco  Feel free to take another look. Thanks a lot.

ZacBlanco (approved):
Thanks @shangm2 for these changes. LGTM

presto-ui/src/components/QueryOverview.jsx
```diff
@@ -1338,7 +1338,7 @@ export default function QueryOverview({ data, show }: { data: QueryData, show: b
                                     Completion Time
                                 </td>
                                 <td className="info-text">
-                                    {data.queryStats.endTime ? formatShortDateTime(new Date(data.queryStats.endTime)) : ""}
+                                    {new Date(data.queryStats.endTime).getTime() !== 0 ? formatShortDateTime(new Date(data.queryStats.endTime)) : ""}
```
yhwang, line 1341: One question: this component requires backward compatibility. users may use this to open a JSON file with old format. Do you know the value is `null` or `undefined` before your change when there is no value for this field? If it's undefined, the new condition would also be `true`
```
> new Date(undefined).getTime() !== 0
true
```

shangm2: @yhwang  The old value was null and new Date(null).getTime in js should be returnning 0, right? I am not a fe person so I got the answer from chatgpt. Pls let me know if this is legitimate.

yhwang: as long as it is not `undefined`, then we are good!
FYI:
```
> new Date(null).getTime() !== 0
false
```

NikhilCollooru (approved):
LGTM

amitkdutta (approved):
Looks good. Thanks @shangm2

PR #24681, opened 2025-03-06
Reduce JDBC read throwing on invalid Timestamp

gggrace14:
## Description
To read JDBC Timestamp type, add catch block around ResultSet.getTimestamp(),
and retry with ResultSet.getObject(columnIndex, LocalDateTime.class)
and Timestamp.valueOf(LocalDateTime) to attempt to return the underlying value,
valid or not valid. In the case of MySQL, both DATETIME type and TIMESTAMP
type map to JDBC Timestamp type.

## Motivation and Context
Implementation of ResultSet.getTimestamp() in mysql-connector-java 8.0.28 does
validity check and throws if the time is not valid due to daylight saving. An example
error is "HOUR_OF_DAY: 2 -> 3" when hour 2 is skipped due to daylight saving.
Older version of mysql-connector-java 5.1.44.2 used in the repo today does not
throw in this case. This change retries once with ResultSet.getObject() without
validity check, which returns the same value as ResultSet.getTimestamp() in the
older version.

## Impact
Behavior of JDBC Timestamp type read could change. Can return a value
when it previously threw on an invalid value. However, for mysql-connector-java,
the returned value is the same as ResultSet.getTimestamp() in 5.1.44.2.

## Test Plan
To read DATETIME column with value '2024-03-10 02:30:00' from MySQL
through mysql-connector-java 8.0.28,
Before throws  "HOUR_OF_DAY: 2 -> 3" error;
After returns 1710064800.

```
== RELEASE NOTES ==

General Changes
*  Reduce JDBC read throwing on invalid Timestamp
```

### DIFF
--- a/presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/StandardReadMappings.java
+++ b/presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/StandardReadMappings.java
@@ -21,9 +21,11 @@ import com.google.common.base.CharMatcher;
 import org.joda.time.chrono.ISOChronology;
 
 import java.sql.ResultSet;
+import java.sql.SQLException;
 import java.sql.Time;
 import java.sql.Timestamp;
 import java.sql.Types;
+import java.time.LocalDateTime;
 import java.util.Optional;
 
 import static com.facebook.presto.common.type.BigintType.BIGINT;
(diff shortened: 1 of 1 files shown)


NikhilCollooru:
Will this be a change of behavior for some users of this ? or is it safe ?
changes look fine.

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/StandardReadMappings.java
```diff
@@ -156,13 +158,14 @@ public static ReadMapping timeReadMapping()
...
     {
         return longReadMapping(TIMESTAMP, (resultSet, columnIndex) -> {
-            /*
-             * TODO `resultSet.getTimestamp(columnIndex)` returns wrong value if JVM's zone had forward offset change and the local time
-             * corresponding to timestamp value being retrieved was not present (a 'gap'), this includes regular DST changes (e.g. Europe/Warsaw)
-             * and one-time policy changes (Asia/Kathmandu's shift by 15 minutes on January 1, 1986, 00:00:00).
-             * The problem can be averted by using `resultSet.getObject(columnIndex, LocalDateTime.class)` -- but this is not universally supported by JDBC drivers.
-             */
-            Timestamp timestamp = resultSet.getTimestamp(columnIndex);
+            Timestamp timestamp;
```
shangm2, line 161: How about the following to make it more concise:
`try {
    return resultSet.getTimestamp(columnIndex).getTime();
} catch (SQLException e) {
    return Timestamp.valueOf(resultSet.getObject(columnIndex, LocalDateTime.class)).getTime();
}`

gggrace14:
> Will this be a change of behavior for some users of this ? or is it safe ? changes look fine.

Yes, I think there's a behavior change. If ResultSet.getTimestamp() previously threw, now it can return a value. That's why I'd like to keep this change separate, and see what the OSS community says. It is a common JDBC path not just for MySQL.

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/StandardReadMappings.java
gggrace14: > How about the following to make it more concise: `try { return resultSet.getTimestamp(columnIndex).getTime(); } catch (SQLException e) { return Timestamp.valueOf(resultSet.getObject(columnIndex, LocalDateTime.class)).getTime(); }`

Sure, I can revise and make it short.

PR #24697, opened 2025-03-10
Add github action to prepare stable release 

unidevel:
## Description

The document of stable release process can be found in https://github.com/prestodb/presto/wiki/Stable-release-process-(every-2-months)

Prior to the coming release, there were 5 steps:

- [Step 1 - Create a release branch and increment the version in the master branch](https://github.com/prestodb/presto/wiki/Stable-release-process-(every-2-months)#step-1-create-a-release-branch-and-increment-the-version-in-the-master-branch)
- [Step 2 - Create release notes PR](https://github.com/prestodb/presto/wiki/Stable-release-process-(every-2-months))
- [Step 3: clean up the tag](https://github.com/prestodb/presto/wiki/Stable-release-process-(every-2-months)#step-3-clean-up-the-tag)
- [Step 4: Publish Maven and Docker artifacts](https://github.com/prestodb/presto/wiki/Stable-release-process-(every-2-months)#step-4-publish-maven-and-docker-artifacts)
- [Step 5: Publish docs website](https://github.com/prestodb/presto/wiki/Stable-release-process-(every-2-months)#step-5-publish-docs-website)

By the suggestion from @ethanyzhang, will merge the steps into 2 github actions:
### Action 1 - "Presto Stable Release - Prepare"

presto_release_preapre_workflow

- [Step 1 - Create a release branch and increment the version in the master branch](https://github.com/prestodb/presto/wiki/Stable-release-process-(every-2-months)#step-1-create-a-release-branch-and-increment-the-version-in-the-master-branch)
- [Step 2 - Create release notes PR](https://github.com/prestodb/presto/wiki/Stable-release-process-(every-2-months))

#### Parameters
presto_release_prepare_inputs

### Action 2 -"Presto Stable Release - Publish"

In the PR https://github.com/prestodb/presto/pull/24821

## Impact
Presto stable release

## Test Plan
- Tested with my env
  -  Prepare => https://github.com/unix280/presto/actions/runs/13843832419
- Test with new release 0.292 after merged

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```


.github/workflows/presto-release-prepare.yml
```diff
@@ -0,0 +1,121 @@
...
+
+      - name: Add git upstream
+        run: |
+          git remote add upstream ${{ github.server_url }}/${{ github.repository }}.git
+          git fetch upstream --tags
+          git remote -v
+
+      - name: Create release notes pull request
+        run: |
+          ./src/release/release-notes.sh ${{ github.repository_owner }} ${{ secrets.PRESTODB_CI_TOKEN }}
```
czentgr, line 121: Nit: please add a new line here and in the other new file.

.github/workflows/presto-release-prepare.yml
```diff
@@ -0,0 +1,121 @@
...
+        with:
+          java-version: '11'
+          distribution: 'temurin'
+
+      - name: Configure git
+        run: |
+          git config --global --add safe.directory ${{github.workspace}}
+          git config --global user.email "ci@lists.prestodb.io"
+          git config --global user.name "prestodb-ci"
+          git config pull.rebase false
```
czentgr, line 111: Do we need this? I suppose it is needed for the release notes script/jar and comes from previous instructions?
i guess it is a common git config that is done. I see it in the other step too.

.github/workflows/presto-release-publish.yml
```diff
@@ -0,0 +1,449 @@
...
+          git config --global user.name "prestodb-ci"
+          git config pull.rebase false
+
+      - name: Cherry-pick release notes
+        run: |
+          git cherry-pick ${{ env.RELEASE_NOTES_COMMIT }}
+
+      - name: Delete existing release tag
+        run: |
+          git push origin :${{ env.RELEASE_TAG }} || true
```
czentgr, line 88: Can we use
```
git push --delete origin ${{ env.RELEASE_TAG }} || true
```
to be a bit more reader friendly?

The || true will always result in success of the step. It is no problem if the deletion of the tag fails?

.github/workflows/presto-release-publish.yml
```diff
@@ -0,0 +1,449 @@
...
+            presto-cli/target/presto-cli-*-executable.jar
+            presto-docs/target/presto-docs-*.zip
+
+      - name: Set up GPG key
+        run: |
+          echo "${{ secrets.GPG_SECRET }}" > ${{ env.GPG_KEY_FILE }}
+          chmod 600 ${{ env.GPG_KEY_FILE }}
+          gpg --import --batch ${{ env.GPG_KEY_FILE }}
+          gpg --batch --yes --pinentry-mode loopback --passphrase "${{ secrets.GPG_PASSPHRASE }}" --sign ${{ env.GPG_KEY_FILE }}
+        env:
```
czentgr, line 189: Suggest to move the env prior to the run. It is more readable.
Same for the other occurrences.

.github/workflows/presto-release-publish.yml
```diff
@@ -0,0 +1,449 @@
...
+
+      - name: Release Maven Artifacts
+        if: ${{ github.event.inputs.publish_maven == 'true' }}
+        run: |
+          unset MAVEN_CONFIG
+          ./mvnw -s ${{ github.workspace }}/settings.xml -V -B -U -e -T1C deploy \
+            -Dgpg.passphrase="${{ secrets.GPG_PASSPHRASE }}" \
+            -Dmaven.wagon.http.retryHandler.count=8 \
+            -DskipTests \
+            -DstagingProfileId=28a0d8c4350ed \
```
czentgr, line 201: Should this be a secret passed in (looks like some kind of userid).

czentgr (commented):
Nice job automating this! A few questions.

.github/workflows/presto-release-prepare.yml
unidevel: Because each job has its own environment, so have to configure this for each job as we want to use the `prestodb-ci` as PR or commit owner

.github/workflows/presto-release-publish.yml
unidevel: Sure, let me update it to secrets.

unidevel: BTW, it is some kind of profile id for checking on the maven repo

.github/workflows/presto-release-publish.yml
unidevel: Add `|| true` to avoid error when the tag deleted by accident, for example, deleted before re-run the job. In this case, the error can be ignored

.github/workflows/presto-release-prepare.yml
```diff
@@ -0,0 +1,123 @@
...
+
+      - name: Push release branch
+        run: |
+          echo "In case this job failed, please delete the tag ${{ env.PRESTO_RELEASE_VERSION }} and the branch release-${{ env.PRESTO_RELEASE_VERSION }}, and re-run the job"
+          git checkout ${{ env.PRESTO_RELEASE_VERSION }}
+          git switch -c release-${{ env.PRESTO_RELEASE_VERSION }}
+          echo "Pushing release branch release-${{ env.PRESTO_RELEASE_VERSION }} and tag ${{ env.PRESTO_RELEASE_VERSION }}"
+          git log --pretty="format:%ce: %s" -3
+          git push origin release-${{ env.PRESTO_RELEASE_VERSION }} --tags
+          echo "Pusing master branch"
```
czentgr, line 81: Typo" Pusing".

.github/workflows/presto-release-prepare.yml
```diff
@@ -0,0 +1,123 @@
...
+          git log --pretty="format:%ce: %s" -5
+          git checkout master
```
czentgr, line 83: Why do we need to checkout the master branch and push it?
We pushed the new release branch above but why do we need to do this. It doesn't look like this changes anything.

.github/workflows/presto-release-prepare.yml
```diff
@@ -0,0 +1,123 @@
...
+      - name: Push release branch
+        run: |
+          echo "In case this job failed, please delete the tag ${{ env.PRESTO_RELEASE_VERSION }} and the branch release-${{ env.PRESTO_RELEASE_VERSION }}, and re-run the job"
+          git checkout ${{ env.PRESTO_RELEASE_VERSION }}
+          git switch -c release-${{ env.PRESTO_RELEASE_VERSION }}
+          echo "Pushing release branch release-${{ env.PRESTO_RELEASE_VERSION }} and tag ${{ env.PRESTO_RELEASE_VERSION }}"
+          git log --pretty="format:%ce: %s" -3
+          git push origin release-${{ env.PRESTO_RELEASE_VERSION }} --tags
+          echo "Pusing master branch"
+          git log --pretty="format:%ce: %s" -5
```
czentgr, line 82: Why this log when we just logged on line 79 from the same branch? Is this supposed to be logging from the master branch? Why is the number of commits 5 vs 3 earlier?

.github/workflows/presto-release-prepare.yml
unidevel: > Why do we need to checkout the master branch and push it? We pushed the new release branch above but why do we need to do this. It doesn't look like this changes anything.

Because mvn release:prepare will create 2 commits on master branch like below and also a tag for each release
```
2025-01-27 943868a918 (oss-release-bot): [maven-release-plugin] prepare for next development iteration
2025-01-27 3374f1b274 (oss-release-bot): [maven-release-plugin] prepare release 0.291  <tag: 0.291> 
```
Next we will create a release branch based on the tag for the next publish action

To make it easy to revert, I will first push the release branch and tag, then push the master branch.

In case anything failed during release, we just need to delete the branch and tag without reverting any commits on master branch.

.github/workflows/presto-release-prepare.yml
unidevel: Good catch, there is a bug here. First need checkout with depth at least  5 so that the logs can be shown correctly.

As in my last comments,  release:prepare  will create 2 commits, that's why I use 5 here to show the last 3 commits in master branch plus the 2 maven release commits

For release branch, `git log --pretty="format:%ce: %s" -3`, should show the last 4 commits(include 1 maven release commit)

.github/workflows/presto-release-prepare.yml
wanglinsong: My opinion is to stick with maven as much as possible, even it means to revert master branch.

.github/workflows/presto-release-prepare.yml
unidevel: Fixed

.github/workflows/presto-release-prepare.yml
unidevel: > My opinion is to stick with maven as much as possible, even it means to revert master branch.

@wanglinsong I think in most of the cases, it should be fine with both pushing order since they all do the same thing - pushing the release branch & tag, and commits to master branch. For the failed cases, let me explain it in detail.

With the order as you suggested ( push the master branch & tag, then the release branch):
- When failed to push to the master branch, just delete the tag and re-run the job
- When successfully pushed to master but failed to push the release tag, need reset `master` branch to remove the commits, then re-run the failed job
- When failed to push the release branch, need reset `master` branch to remove the commits, delete the tag and try again.

With the order I used( push the release branch & tag, then master branch)
- When push to release branch & tag failed, just delete the branch & tag(if exists), and re-run the job
- When push to master failed, just delete the release branch & tag(since they are newly created, no one is working on it)

Because reset the commits on master branch will have impact on the open PRs which has greater risk than delete the newly created branch and tag, that why I think it's better to keep the pushing order in my PR.

czentgr: Thanks for explaining. I agree that we wouldn't want to reset the master branch if we can avoid it.

Do we need to generally worry about race conditions if a committer merges a PR while the running. I suspect in that case we would fail the push to master because the master branch base is different and it can't push the maven commits.

.github/workflows/presto-release-prepare.yml
```diff
@@ -0,0 +1,136 @@
...
+
+      - name: Push release tag, branch and commits
+        run: |
+          echo "In case this job failed, please delete the tag ${{ env.PRESTO_RELEASE_VERSION }} and the branch release-${{ env.PRESTO_RELEASE_VERSION }}, and re-run the job"
+          git checkout ${{ env.PRESTO_RELEASE_VERSION }}
+          git switch -c release-${{ env.PRESTO_RELEASE_VERSION }}
+          echo "Pushing release branch release-${{ env.PRESTO_RELEASE_VERSION }} and tag ${{ env.PRESTO_RELEASE_VERSION }}"
+
+          echo "commits on release-${{ env.PRESTO_RELEASE_VERSION }} branch"
+          git ls -4
```
czentgr, line 88: Nit: for consistence use 5 here? Before and after this 5 is used.

czentgr (commented):
Only one nit.

.github/workflows/presto-release-prepare.yml
unidevel: @czentgr  The reason the number changes between 4 and 5 is that I want to keep the logs containing the last 3 commits that were not generated by the Maven release plugin

Take 0.291 for example, in the branch release-0.291 and tag 0.291, the command `git ls -4` will display the 4 commits as follows:

image

In the master branch, the comand `git ls -5` will display the last 5 commits, which has the same 3 commits in the release branch
image

But it's also fine to keep 5 commits in both situations. We only use the logs to verify that the release branch was created correctly.

.github/workflows/presto-release-prepare.yml
unidevel: > Do we need to generally worry about race conditions if a committer merges a PR while the running. I suspect in that case we would fail the push to master because the master branch base is different and it can't push the maven commits.

Because there is a special settings to let the user `prestodb-ci` by pass the PR rules, this user can push commits directly on master branch. Also about the race conditions, I think it follows the git pushing rules, the worse case is that some PRs merged during the action, which will cause it failed to push the commits to master branch. In this case, we can delete the branch and tag, then try the action again without any impact on all the open PRs.

unidevel:
Fixed the conditions for each job, seems `success()` not equal to `!failure() && !cancelled()` when job is skipped

czentgr:
> Fixed the conditions for each job, seems `success()` not equal to `!failure() && !cancelled()` when job is skipped

I was looking at the conditions earlier and they seemed ok. I don't recall the original check but it didn't work for skipped checks because then success() would be false when the previous step was skipped?

ethanyzhang:
@unidevel In today's meeting you mentioned only `prestodb-ci` can trigger the job, can we make it a required approval from the committer group? Not all committers can access the `prestodb-ci` user.

.github/workflows/presto-release-prepare.yml
wanglinsong: > > My opinion is to stick with maven as much as possible, even it means to revert master branch.
> 
> @wanglinsong I think in most of the cases, it should be fine with both pushing order since they all do the same thing - pushing the release branch & tag, and commits to master branch. For the failed cases, let me explain it in detail.
> 
> With the order as you suggested ( push the master branch & tag, then the release branch):
> 
> * When failed to push to the master branch, just delete the tag and re-run the job
> * When successfully pushed to master but failed to push the release tag, need reset `master` branch to remove the commits, then re-run the failed job
> * When failed to push the release branch, need reset `master` branch to remove the commits, delete the tag and try again.
> 
> With the order I used( push the release branch & tag, then master branch)
> 
> * When push to release branch & tag failed, just delete the branch & tag(if exists), and re-run the job
> * When push to master failed, just delete the release branch & tag(since they are newly created, no one is working on it)
> 
> Because reset the commits on master branch will have impact on the open PRs which has greater risk than delete the newly created branch and tag, that why I think it's better to keep the pushing order in my PR.

This is no reason of failing to push the release branch, which is a new branch, no any conflicts possible at all.

czentgr:
Yes, always() is always true and you need to use it if you want to force additional eval for a step because success() is otherwise used which fails if a step is skipped. success() is the default eval for each step unless one of the other functions is used.

czentgr (commented):
Thanks.
As for the presentation you had it helps to document what to do if aspects of the pipeline fail. Will this be added to the documentation in the wiki for prestodb?
Maybe add this to the description of the PR where to find that documentation in case someone looks in the PR?

@majetideepak If you have time, please take a look as well.

steveburnett:
> Thanks. As for the presentation you had it helps to document what to do if aspects of the pipeline fail. Will this be added to the documentation in the wiki for prestodb? Maybe add this to the description of the PR where to find that documentation in case someone looks in the PR?

Let me know if I can help with any documentation needs!

.github/workflows/presto-release-publish.yml
```diff
@@ -0,0 +1,444 @@
...
+        type: boolean
+        default: true
+        required: false
+      tag_image_as_latest:
+        description: 'Tag the image as latest'
+        type: boolean
+        default: true
+        required: false
+      dependency_image:
+        description: 'Dependency image(e.g., prestodb/presto-native-dependency:0.290-20241014120930-e1fc090)'
```
majetideepak, line 38: This seems to be out of date. We should probably keep `prestodb/presto-native-dependency:latest` tag as well.
CC: @wanglinsong

unidevel: @majetideepak I can update the description here, but I don't want to tag `prestodb/presto-native-dependency:latest` when publishing docker images for a release. 
Because the image may update during the release process, e.g. 
1. We cut the release for `0.292`, create the release-notes PR, advance the version in `master` branch to `0.293-SNAPSHOT`
2. About 1-2 weeks for the release-notes PR, developer continue working on the `master` branch
3. Someone pushed a new **new dependency image** to dockerhub for `0.293-SNAPSHOT` on `master` branch duing this period, and tag as `presto-native-dependency:latest`
4. The release-notes PR of 0.292 merged, and we start to publish the docker images for 0.292 - if we tag the `prestodb/presto-native-dependency:latest` here, it will override the new docker image for `0.293-SNAPSHOT`

That's why I only publish the dependency image(if not provided) but not tag it as `latest` version

unidevel:
> > Thanks. As for the presentation you had it helps to document what to do if aspects of the pipeline fail. Will this be added to the documentation in the wiki for prestodb? Maybe add this to the description of the PR where to find that documentation in case someone looks in the PR?
> 
> Let me know if I can help with any documentation needs!

@steveburnett I'll update the document to https://github.com/prestodb/presto/wiki/Stable-release-process-(every-2-months) , I'll let you know after it is updated, thanks

unidevel:
@steveburnett I added the document of this new release process to https://github.com/prestodb/presto/wiki/Stable-release-process-(every-2-months)#for-release-after-0291 under section **Release after 0.291**, please help to review, thanks.

.github/workflows/presto-release-prepare.yml
unidevel: > This is no reason of failing to push the release branch, which is a new branch, no any conflicts possible at all.

unidevel: > This is no reason of failing to push the release branch, which is a new branch, no any conflicts possible at all.±

Yes, but if the release needs to be reverted for some reason, the release branch and tag might be forgotten to be deleted.

steveburnett:
> @steveburnett I added the document of this new release process to https://github.com/prestodb/presto/wiki/Stable-release-process-(every-2-months)#for-release-after-0291 under section **Release after 0.291**, please help to review, thanks.

Looks good, thanks! I have some minor edits to suggest - punctuation and similar nits, nothing affecting the meaning or the sequence of events. Would you like me to tell you these nits, or would you like me to just edit the wiki page myself? I can work either way but I think the second option is faster and simpler.

unidevel:
> > @steveburnett I added the document of this new release process to https://github.com/prestodb/presto/wiki/Stable-release-process-(every-2-months)#for-release-after-0291 under section **Release after 0.291**, please help to review, thanks.
> 
> Looks good, thanks! I have some minor edits to suggest - punctuation and similar nits, nothing affecting the meaning or the sequence of events. Would you like me to tell you these nits, or would you like me to just edit the wiki page myself? I can work either way but I think the second option is faster and simpler.

Thanks, please feel free to edit the wiki.

.github/workflows/presto-release-publish.yml
```diff
@@ -0,0 +1,444 @@
...
+        type: boolean
+        default: true
+        required: false
+      tag_image_as_latest:
+        description: 'Tag the image as latest'
+        type: boolean
+        default: true
+        required: false
+      dependency_image:
+        description: 'Dependency image(e.g., prestodb/presto-native-dependency:latest)'
```
majetideepak, line 38: @unidevel The latest is not up to date. Use [0.292-20250204112033-cf8ba84](https://hub.docker.com/layers/prestodb/presto-native-dependency/0.292-20250204112033-cf8ba84/images/sha256-89868c4742469b04c9b13440eee0a1edee13a1dbf9a0640f556520a11b72cf80)

unidevel: @majetideepak I think this is fine, because this is only an example in description and it will not only be used for the release 0.292. Ideally we can use `prestodb/presto-native-dependency:latest` in future releases.

@wanglinsong I think we need either update or remove the `latest` tag for prestodb/presto-native-dependency docker images.

steveburnett:
> > > @steveburnett I added the document of this new release process to https://github.com/prestodb/presto/wiki/Stable-release-process-(every-2-months)#for-release-after-0291 under section **Release after 0.291**, please help to review, thanks.
> > 
> > 
> > Looks good, thanks! I have some minor edits to suggest - punctuation and similar nits, nothing affecting the meaning or the sequence of events. Would you like me to tell you these nits, or would you like me to just edit the wiki page myself? I can work either way but I think the second option is faster and simpler.
> 
> Thanks, please feel free to edit the wiki.

Done!

unidevel:
per discussion with @wanglinsong , will put the publish action to a new PR.

wanglinsong (approved):
LGTM

majetideepak:
@wanglinsong, @unidevel Is there any value to publicly releasing the dependency image?

unidevel:
> @wanglinsong, @unidevel Is there any value to publicly releasing the dependency image?

Yes, I believe there are values such as:

- Accelerating the build time for docker images
- Improving CI/CD jobs by making them faster
- Speeding up the setup of a local dev env with docker

czentgr (approved):

PR #24714, opened 2025-03-12
Dynamically compute node memory based on configured spark executor memory

pgupta2:
Summary:
PoS Memory configs are currently statically defined in session properties 
or config properties. These are not dynamically computed based on the 
configured spark's container memory. Adding the logic to dynamically 
compute different memory configs based on configured spark executor 
memory.

Differential Revision: D70345244

### DIFF
--- a/presto-spark-base/pom.xml
+++ b/presto-spark-base/pom.xml
@@ -110,6 +110,11 @@
         </dependency>
 
         <dependency>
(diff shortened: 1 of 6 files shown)


shrinidhijoshi (dismissed):
LGTM. Thanks for making these changes!

shrinidhijoshi (approved):

PR #24736, opened 2025-03-17
[native] Advance velox.

amitkdutta:
```
== NO RELEASE NOTE ==
```


xiaoxmeng (approved):

ZacBlanco:
Thanks for bringing this to my attention @aditi-pandit I investigated and found out we just got unlucky a few times with a flaky test. I was able to reproduce the failure locally and will work on a fix. After re-running one more time, all checks are green now

aditi-pandit:
> Thanks for bringing this to my attention @aditi-pandit I investigated and found out we just got unlucky a few times with a flaky test. I was able to reproduce the failure locally and will work on a fix. After re-running one more time, all checks are green now

Thanks @ZacBlanco. Please create an issue for it as well.

PR #24744, opened 2025-03-17
Add unified check logic for permissions

kevintang2022:
## Description

This change will allow for the functionality of skipping accessControl calls if a checkQueryIntegrity is successful. checkQueryIntegrity is where a unified permissions check will occur, and if it is successful, it will skip all other authentication and authorization checks like checkPermissions under QuerySessionSupplier.java, and checks like checkCanSelectFromColumns, etc.

## Motivation and Context

Currently, Presto authentication and authorization happens in different phases of the query execution. Authentication happens during dispatch in checkPermissions in QuerySessionSupplier.java, which is before the queuing phase. The authorization checks, like checkCanSelectFromColumns will happen after the queuing phase. By utilizing the result from the unified permissions check under checkQueryIntegrity, all the permission check can be done at once.

## Impact

- Gating this feature under isSkipPermissionsCheckEnabled configuration flag. If the flag is true, and the result of validateQueryToken is successful, then all other existing permissions checks get skipped

- Followup to this task is to determine the best place to have the validateQueryToken call. Currently, it will be in the same place as checkPermissions so that it can be skipped if the validateQueryToken call is successful. But it could be better for security purposes to move it after queuing phase in case a view definition changes during queuing phase, which can introduce a security loophole. This is because view definitions is retrieved after queuing. However, this can be addressed in a followup task. Currently, validateQueryToken will only be used for logging purposes, and its result will not be utilized.

## Test Plan

Use HiveQueryRunner with debugger enabled. Set breakpoints in QuerySessionSupplier to see coverage on validateQueryToken usage
## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Add synchronous shadow call to unified permission check, which can be used to skip existing permissions check if successful

### DIFF
--- a/presto-main-base/src/main/java/com/facebook/presto/dispatcher/DispatchManager.java
+++ b/presto-main-base/src/main/java/com/facebook/presto/dispatcher/DispatchManager.java
@@ -32,13 +32,11 @@ import com.facebook.presto.server.SessionPropertyDefaults;
 import com.facebook.presto.server.SessionSupplier;
 import com.facebook.presto.spi.PrestoException;
 import com.facebook.presto.spi.QueryId;
-import com.facebook.presto.spi.WarningCollector;
 import com.facebook.presto.spi.analyzer.AnalyzerOptions;
 import com.facebook.presto.spi.analyzer.QueryPreparerProvider;
 import com.facebook.presto.spi.resourceGroups.SelectionContext;
 import com.facebook.presto.spi.resourceGroups.SelectionCriteria;
 import com.facebook.presto.spi.security.AccessControl;
-import com.facebook.presto.spi.security.AccessControlContext;
 import com.facebook.presto.sql.analyzer.QueryPreparerProviderManager;
 import com.facebook.presto.transaction.TransactionManager;
 import com.google.common.util.concurrent.AbstractFuture;
@@ -96,7 +94,7 @@ public class DispatchManager
 
     /**
      * Dispatch Manager is used for the pre-queuing part of queries prior to the query execution phase.
-     *
+     * <p>
      * Dispatch Manager object is instantiated when the presto server is launched by server bootstrap time. It is a critical component in resource management section of the query.
      *
      * @param queryIdGenerator query ID generator for generating a new query ID when a query is created
@@ -183,7 +181,7 @@ public class DispatchManager
 
     /**
      * Create a query id
-     *
+     * <p>
      * This method is called when a {@code Query} object is created
      *
      * @return {@link QueryId}
@@ -272,20 +270,8 @@ public class DispatchManager
             }
 
             // decode session
-            sessionBuilder = sessionSupplier.createSessionBuilder(queryId, sessionContext, warningCollectorFactory);
+            sessionBuilder = sessionSupplier.createSessionBuilder(queryId, sessionContext, warningCollectorFactory, query);
 
-            AccessControlContext accessControlContext = new AccessControlContext(
-                    queryId,
-                    Optional.ofNullable(sessionContext.getClientInfo()),
-                    sessionContext.getClientTags(),
-                    Optional.ofNullable(sessionContext.getSource()),
-                    WarningCollector.NOOP,
-                    sessionContext.getRuntimeStats(),
-                    Optional.empty(),
-                    Optional.ofNullable(sessionContext.getCatalog()),
-                    Optional.ofNullable(sessionContext.getSchema()));
-
-            accessControl.checkQueryIntegrity(sessionContext.getIdentity(), accessControlContext, query);
             session = sessionBuilder.build();
 
             // prepare query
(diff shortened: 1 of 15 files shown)

### CODE REVIEW

File: presto-main-base/src/main/java/com/facebook/presto/server/QuerySessionSupplier.java [hunk 1]
```diff
@@ -156,4 +171,6 @@ private Identity authenticateIdentity(QueryId queryId, SessionContext context)
                 Optional.of(identity.getUserName()),
                 identity.getReasonForSelect())).orElseGet(context::getIdentity);
     }
+
+    public void logUnifiedPermissionsCheckFailure() {}
```
kevintang2022, line 175: Leave empty for now. It is difficult to add to warning collector since identity is authenticated before warning collector is set

presto-main-base/src/main/java/com/facebook/presto/server/QuerySessionSupplier.java
```diff
@@ -141,10 +143,23 @@ else if (context.getTimeZoneId() != null) {
...
+        try {
+            validateQueryToken(accessControl, queryId, context, query);
+
+            if (isSkipPermissionsCheckEnabled(accessControl)) {
+                return context.getIdentity();
+            }
+        }
+        catch (Exception e) {
+            logUnifiedPermissionsCheckFailure();
+        }
```
kevintang2022, line 157: This is the part where the call is made. Before, it was not wrapped in a try catch block. Since we just want it to be a shadow call for now, we need to catch the exception that will be thrown

presto-main-base/src/main/java/com/facebook/presto/server/QuerySessionSupplier.java
```diff
@@ -141,10 +143,23 @@ else if (context.getTimeZoneId() != null) {
...
     {
+        try {
+            validateQueryToken(accessControl, queryId, context, query);
+
+            if (isSkipPermissionsCheckEnabled(accessControl)) {
+                return context.getIdentity();
+            }
+        }
+        catch (Exception e) {
+            logUnifiedPermissionsCheckFailure();
```
rschlussel, line 156: we should catch/log the exception directly inside our implementation of validateQueryToken. A different accessControl might implement validateQueryToken and not want to catch exceptions.

Also, unifiedPermissionCheck is an internal implementation detail. It shouldn't leak out here.

presto-main-base/src/main/java/com/facebook/presto/server/QuerySessionSupplier.java
```diff
@@ -141,10 +143,23 @@ else if (context.getTimeZoneId() != null) {
         return sessionBuilder;
     }
 
-    private Identity authenticateIdentity(QueryId queryId, SessionContext context)
+    private Identity authenticateIdentity(QueryId queryId, SessionContext context, String query)
     {
+        try {
+            validateQueryToken(accessControl, queryId, context, query);
+
+            if (isSkipPermissionsCheckEnabled(accessControl)) {
```
rschlussel, line 151: what is the purpose of skipPermissionsCheckEnabled?  I think this is because we only want to skip if validateQueryToken was successful.  However, most of the permissions get checked elsewhere (e.g. checking column access), so not sure how useful this is.

presto-spi/src/main/java/com/facebook/presto/spi/security/AccessControl.java
```diff
@@ -322,4 +325,6 @@ default AuthorizedIdentity selectAuthorizedIdentity(Identity identity, AccessCon
      * @throws com.facebook.presto.spi.security.AccessDeniedException if not allowed
      */
     void checkCanAddConstraints(TransactionId transactionId, Identity identity, AccessControlContext context, QualifiedObjectName constraintName);
+
+    boolean isSkipPermissionsCheckEnabled();
```
rschlussel, line 329: have this default to false so accessControls it's not relevant to don't need to implement it.

presto-main-base/src/main/java/com/facebook/presto/dispatcher/DispatchManager.java
```diff
@@ -272,20 +270,8 @@ private <C> void createQueryInternal(QueryId queryId, String slug, int retryCoun
...
-                    Optional.ofNullable(sessionContext.getClientInfo()),
-                    sessionContext.getClientTags(),
-                    Optional.ofNullable(sessionContext.getSource()),
-                    WarningCollector.NOOP,
-                    sessionContext.getRuntimeStats(),
-                    Optional.empty(),
-                    Optional.ofNullable(sessionContext.getCatalog()),
-                    Optional.ofNullable(sessionContext.getSchema()));
-
-            accessControl.checkQueryIntegrity(sessionContext.getIdentity(), accessControlContext, query);
```
rschlussel, line 288: can you add the motivation for moving this check to the commit description?

steveburnett:
Please resolve the file conflicts.

PR #24754, opened 2025-03-19
Upgrading mysql jar in presto

namya28:
## Description
This PR is for upgrading the mysql jar version to the latest version 9.2.0.
This fixes CVE-2023-22102.

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

Security Changes

* Upgrade the MySQL to version 9.2.0 in response to `CVE-2023-22102 <https://github.com/advisories/GHSA-m6vm-37g8-gqvh>`_. :pr:`24754`

```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -82,7 +82,7 @@
         <dep.j2objc.version>3.0.0</dep.j2objc.version>
         <dep.avro.version>1.11.4</dep.avro.version>
         <dep.commons.compress.version>1.26.2</dep.commons.compress.version>
-        <dep.protobuf-java.version>3.25.5</dep.protobuf-java.version>
+        <dep.protobuf-java.version>4.29.0</dep.protobuf-java.version>
         <dep.netty.version>4.1.118.Final</dep.netty.version>
         <dep.snakeyaml.version>2.0</dep.snakeyaml.version>
         <dep.jetty.version>9.4.56.v20240826</dep.jetty.version>
@@ -1406,9 +1406,9 @@
             </dependency>
 
             <dependency>
-                <groupId>mysql</groupId>
-                <artifactId>mysql-connector-java</artifactId>
-                <version>5.1.48</version>
+                <groupId>com.mysql</groupId>
+                <artifactId>mysql-connector-j</artifactId>
+                <version>9.2.0</version>
             </dependency>
 
             <dependency>
@@ -2001,8 +2001,14 @@
 
             <dependency>
                 <groupId>com.facebook.presto</groupId>
-                <artifactId>testing-mysql-server-5</artifactId>
+                <artifactId>testing-mysql-server-8</artifactId>
                 <version>${dep.testing-mysql-server-5.version}</version>
+                <exclusions>
+                    <exclusion>
+                        <groupId>mysql</groupId>
+                        <artifactId>mysql-connector-java</artifactId>
+                    </exclusion>
+                </exclusions>
             </dependency>
 
             <dependency>
--- a/presto-mysql/src/main/java/com/facebook/presto/plugin/mysql/MySqlClient.java
+++ b/presto-mysql/src/main/java/com/facebook/presto/plugin/mysql/MySqlClient.java
@@ -29,8 +29,8 @@ import com.facebook.presto.spi.ConnectorTableMetadata;
 import com.facebook.presto.spi.PrestoException;
 import com.facebook.presto.spi.SchemaTableName;
 import com.google.common.collect.ImmutableSet;
+import com.mysql.cj.jdbc.JdbcStatement;
 import com.mysql.jdbc.Driver;
-import com.mysql.jdbc.Statement;
 
 import javax.inject.Inject;
 
@@ -39,6 +39,7 @@ import java.sql.DatabaseMetaData;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.sql.SQLSyntaxErrorException;
 import java.util.Collection;
 import java.util.Optional;
 import java.util.Properties;
@@ -53,14 +54,18 @@ import static com.facebook.presto.plugin.jdbc.JdbcErrorCode.JDBC_ERROR;
 import static com.facebook.presto.spi.StandardErrorCode.ALREADY_EXISTS;
 import static com.facebook.presto.spi.StandardErrorCode.NOT_SUPPORTED;
 import static com.google.common.util.concurrent.MoreExecutors.directExecutor;
-import static com.mysql.jdbc.SQLError.SQL_STATE_ER_TABLE_EXISTS_ERROR;
-import static com.mysql.jdbc.SQLError.SQL_STATE_SYNTAX_ERROR;
 import static java.lang.String.format;
 import static java.util.Locale.ENGLISH;
 
 public class MySqlClient
         extends BaseJdbcClient
 {
+    /**
+     * Error code corresponding to code thrown when a table already exists.
+     * The code is derived from the MySQL documentation.
+     * @see <a href="https://dev.mysql.com/doc/connector-j/en/connector-j-reference-error-sqlstates.html">MySQL documentation</a>
+     */
+    private static final String SQL_STATE_ER_TABLE_EXISTS_ERROR = "42S01";
     @Inject
     public MySqlClient(JdbcConnectorId connectorId, BaseJdbcConfig config, MySqlConfig mySqlConfig)
             throws SQLException
@@ -127,8 +132,8 @@ public class MySqlClient
             throws SQLException
     {
         PreparedStatement statement = connection.prepareStatement(sql);
-        if (statement.isWrapperFor(Statement.class)) {
-            statement.unwrap(Statement.class).enableStreamingResults();
+        if (statement.isWrapperFor(JdbcStatement.class)) {
+            statement.unwrap(JdbcStatement.class).enableStreamingResults();
         }
         return statement;
     }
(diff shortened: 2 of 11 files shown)


presto-mysql/src/main/java/com/facebook/presto/plugin/mysql/MySqlClientModule.java
```diff
@@ -40,12 +39,10 @@ protected void setup(Binder binder)
...
         try {
-            Driver driver = new Driver();
-            Properties urlProperties = driver.parseURL(connectionUrl, null);
-            checkArgument(urlProperties != null, "Invalid JDBC URL for MySQL connector");
-            checkArgument(driver.database(urlProperties) == null, "Database (catalog) must not be specified in JDBC URL for MySQL connector");
+            checkArgument(ConnectionUrlParser.isConnectionStringSupported(connectionUrl), "Invalid JDBC URL for MySQL connector");
+            checkArgument(StringUtils.isNullOrEmpty(ConnectionUrl.getConnectionUrlInstance(connectionUrl, null).getDatabase()), "Database (catalog) must not be specified in JDBC URL for MySQL connector");
         }
-        catch (SQLException e) {
+        catch (Exception e) {
```
agrawalreetika, line 45: Whats the specific exception here? Should we catch that here?

agrawalreetika (changes requested):

agrawalreetika:
Thanks for the changes, please squash your changes into 1 with the relevant commit message.

presto-product-tests/pom.xml
```diff
@@ -167,9 +167,15 @@
...
-            <artifactId>mysql-connector-java</artifactId>
+            <groupId>com.mysql</groupId>
+            <artifactId>mysql-connector-j</artifactId>
             <scope>runtime</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>com.google.protobuf</groupId>
+                    <artifactId>protobuf-java</artifactId>
+                </exclusion>
+            </exclusions>
```
ZacBlanco, line 178: Do we need to exclude this or can we just pull up the dependency version through the `dependencyManagement` section? Or is it due to duplicate class conflicts?

namya28: @ZacBlanco, Excluding this as it is resulting in upper bound dependencies error.
Screenshot 2025-03-24 at 1 09 29 PM

presto-mysql/src/main/java/com/facebook/presto/plugin/mysql/MySqlClientModule.java
namya28: Hi @agrawalreetika , thank you for your feedback. Based on my research , The code is building fine without an exception as well. Since it is optional and if required , would you recommend me adding an `IllegalArgumentException` instead of `Exception`?

agrawalreetika: Earlier code was throwing `SQLException` which is why it was added now in your current implementation I don't think try-catch is even needed

presto-product-tests/pom.xml
agrawalreetika: Can't we try to just upgrade the dependency to the required version?

namya28: @agrawalreetika , thank you for your feedback! I tried removing the exclusions and tried upgrading the protobuf-java version to 4.29.0 using 2 approaches : 
1. upgraded the version to 4.29.0 in root pom. It is resulting in error in presto-main
Screenshot 2025-03-25 at 2 57 37 PM
2. Kept the earlier version as is in the root pom (3.25.5) and tried upgrading the version to 4.29.0 using dependency management in the module pom (presto-function-namespace-managers) and it is resulting in the same error in presto-main . (same in the case of presto-product-tests module as well).
Screenshot 2025-03-25 at 3 05 20 PM
3. Also tried ignoring it under <ignoredResourcePatterns> and <ignoredClassPatterns> under duplicate-finder-maven-plugin. Resulting in the same error.

agrawalreetika: If with `4.29.0` version upgrade we are hitting duplicate class error then I think I am ok with excluding it from `<artifactId>mysql-connector-j</artifactId>` if the CI tests are passing.

@ZacBlanco Do you have any concern with exclusion? Or have any other approach to handle this?

agrawalreetika (approved):

pom.xml
```diff
@@ -1980,7 +1980,7 @@
 
             <dependency>
                 <groupId>com.facebook.presto</groupId>
-                <artifactId>testing-mysql-server-5</artifactId>
+                <artifactId>testing-mysql-server-8</artifactId>
                 <version>${dep.testing-mysql-server-5.version}</version>
```
ZacBlanco, line 1984: In order to prevent needing to add the exclusion of the mysql client for every use of the `testing-mysql-server-8` dependency, can you just add the exclusion here?

presto-mysql/src/main/java/com/facebook/presto/plugin/mysql/MySqlClient.java
```diff
@@ -53,14 +53,15 @@
...
-import static com.mysql.jdbc.SQLError.SQL_STATE_ER_TABLE_EXISTS_ERROR;
-import static com.mysql.jdbc.SQLError.SQL_STATE_SYNTAX_ERROR;
 import static java.lang.String.format;
 import static java.util.Locale.ENGLISH;
 
 public class MySqlClient
         extends BaseJdbcClient
 {
+    private static final String SQL_STATE_SYNTAX_ERROR = "42000";
+    private static final String SQL_STATE_ER_TABLE_EXISTS_ERROR = "42S01";
```
ZacBlanco, line 63: Are these not defined/available in the new mysql client? Where is this documented?

presto-product-tests/pom.xml
ZacBlanco: I am not a fan of excluding random dependencies. The conflicts aren't even on Java classes, so this could potentially lead to runtime issues if we're excluding core protobuf classes. In this case I would rather add exceptions to the duplicate finder for the `.proto` files

ZacBlanco (changes requested):
Do we have a PR to upgrade the testing-mysql-server? I would prefer if we updated that + released a new version for Presto first before this gets merged in order to reduce the number of exclusions

presto-mysql/src/main/java/com/facebook/presto/plugin/mysql/MySqlClient.java
namya28: These are sql error codes corresponding to the mysql error code. https://dev.mysql.com/doc/connector-j/en/connector-j-reference-error-sqlstates.html.

ZacBlanco: I think for the syntax error, we should instead catch the SQLSyntaxError exception from the client rather than catching and comparing the code. For the table exists error, there isn't a corresponding exception type, so we can keep the current approach. Though we should leave a comment with a link to the documentation of error codes for future developers to know where the code came from

namya28:
> Do we have a PR to upgrade the testing-mysql-server? I would prefer if we updated that + released a new version for Presto first before this gets merged in order to reduce the number of exclusions

There already is a release for testing mysql-server 8 for the version being used. https://repo.maven.apache.org/maven2/com/facebook/presto/testing-mysql-server-8/ .

presto-product-tests/pom.xml
namya28: Upgraded the `protobuf-java` version in the root pom and ignored the dependency `ratis-thirdparty-misc` under `duplicate-finder-maven-plugin`.

presto-mysql/src/main/java/com/facebook/presto/plugin/mysql/MySqlClient.java
namya28: Added the SQLSyntaxError exception as well. Had to keep the SQLException as well as I was getting the error _Unhandled exception: java.sql.SQLException_ .

namya28:
@ZacBlanco @agrawalreetika , I have addressed the review comments. Could you please help review once again whenever convenient.

agrawalreetika (dismissed):
LGTM. Please squash all your commits into one.

presto-mysql/src/main/java/com/facebook/presto/plugin/mysql/MySqlClient.java
```diff
@@ -53,14 +54,15 @@
...
-import static com.mysql.jdbc.SQLError.SQL_STATE_ER_TABLE_EXISTS_ERROR;
-import static com.mysql.jdbc.SQLError.SQL_STATE_SYNTAX_ERROR;
 import static java.lang.String.format;
 import static java.util.Locale.ENGLISH;
 
 public class MySqlClient
         extends BaseJdbcClient
 {
+    private static final String SQL_STATE_ER_TABLE_EXISTS_ERROR = "42S01";
+    //Mapping the MySql error numbers to JDBC SqlState Codes. See: https://dev.mysql.com/doc/connector-j/en/connector-j-reference-error-sqlstates.html.
```
ZacBlanco, line 64: Can you please move this comment above the variable. And use javadoc comment formatting.
```suggestion
    /** 
    * Error code corresponding to code thrown when a table already exists.  The code is derived from the MySQL documentation. See {@link https://dev.mysql.com/doc/connector-j/en/connector-j-reference-error-sqlstates.html}
    */
```

ZacBlanco (commented):
one minor nit

presto-mysql/src/main/java/com/facebook/presto/plugin/mysql/MySqlClient.java
namya28: @ZacBlanco , thank you for the feedback. I made the changes . However , I could not use `See {@link https:...} ` as I was encountering an error `Cannot resolve symbol 'https'` . Upon a little digging , I found that Javadoc's `{@link}` is not meant for external URLs, so using `See {@link https:...} `will cause an error. I went with using `@see <a href ..>`instead for linking to external URLs. Please do let me know if this is okay and if there's something you'd want me to change/edit further.

ZacBlanco: Thanks - sorry I forgot the link syntax for javadoc 😅 . This new version is much better

ZacBlanco (approved):
LGTM! Thanks for the upgrade!

jaystarshot (approved):

PR #24785, opened 2025-03-23
[native] Map kUnsupportedInputUncatchable to not supported error code.

amitkdutta:
## Description
Mapping kUnsupportedInputUncatchable with NOT_SUPPORTED

## Motivation and Context
Better error reporting from presto side.

```
== NO RELEASE NOTE ==
```


gggrace14:
Adding Wei to review. I happen to see comment like below in the Timestamp.cpp.
>     // Invalid argument means we hit a conversion not supported by
>     // external/date. Need to throw a RuntimeError so that try() statements do
>     // not suppress it.

kgpai (approved):

amitkdutta:
@gggrace14 Thanks for looking. You are right, it changes Presto categorization. It does not change any behavior of throwing.

PR #24786, opened 2025-03-23
Add union optimization

jainxrohit:
## Description

Added optimization to convert UNION to UNION ALL. This optimization is backed by the session configuration optimize_union_to_union_all.

## Motivation and Context
UNION specified without ALL or DISTINCT keyword is equivalent to UNION DISTINCT, which is computationally expensive.
Adding a new optimization to convert union to union all.

## Impact
This optimization can boost the performance to latency sensitive queries.

## Test Plan
Added few tests to verify union optimization.

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
*  Add ``optimize_union_to_union_all`` session property to optimize union with union all.
```

### DIFF
--- a/presto-main-base/src/main/java/com/facebook/presto/sql/rewrite/UnionOptimizationRewrite.java
+++ b/presto-main-base/src/main/java/com/facebook/presto/sql/rewrite/UnionOptimizationRewrite.java
@@ -20,31 +20,23 @@ import com.facebook.presto.spi.security.AccessControl;
 import com.facebook.presto.sql.analyzer.QueryExplainer;
 import com.facebook.presto.sql.parser.SqlParser;
 import com.facebook.presto.sql.tree.Expression;
+import com.facebook.presto.sql.tree.Node;
 import com.facebook.presto.sql.tree.NodeRef;
 import com.facebook.presto.sql.tree.Parameter;
 import com.facebook.presto.sql.tree.Statement;
-import com.google.common.collect.ImmutableList;
+import com.facebook.presto.sql.tree.Union;
 
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
-import static java.util.Objects.requireNonNull;
+import static com.facebook.presto.SystemSessionProperties.isOptimizeUnionToUnionAll;
 
-public final class StatementRewrite
+public class UnionOptimizationRewrite
+        implements StatementRewrite.Rewrite
 {
-    private static final List<Rewrite> REWRITES = ImmutableList.of(
-            new DescribeInputRewrite(),
-            new DescribeOutputRewrite(),
-            new ShowQueriesRewrite(),
-            new ShowStatsRewrite(),
-            new ExplainRewrite(),
-            new MaterializedViewOptimizationRewrite(),
-            new UnionOptimizationRewrite());
-
-    private StatementRewrite() {}
-
-    public static Statement rewrite(
+    @Override
+    public Statement rewrite(
             Session session,
             Metadata metadata,
             SqlParser parser,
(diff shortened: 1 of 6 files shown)


presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java
```diff
@@ -8008,6 +8009,91 @@ public void testEvaluateProjectOnValues()
...
+        Session optimizeUnionSession = Session.builder(getSession())
+                .setSystemProperty(OPTIMIZE_UNION_TO_UNION_ALL, "true")
+                .build();
+
+        @Language("SQL") String testQuery = "select max(rn) from (select row_number() over(partition by custkey) as rn from (select custkey from orders union all select custkey from customer))";
+        MaterializedResult materializedRows = computeActual(defaultSession, testQuery);
+        List<MaterializedRow> rows = materializedRows.getMaterializedRows();
+
+        testQuery = "select max(rn) from (select row_number() over(partition by custkey) as rn from (select custkey from orders union select custkey from customer))";
+        assertNotEquals(rows, computeActual(defaultSession, testQuery).getMaterializedRows());
```
kaikalur, line 8093: Will be good to test this = 1 when distinct is on and actual number when not.

kaikalur (approved):

kaikalur (approved):
LGTM

steveburnett:
Thanks for the release note! Suggestions for formatting, and to include the name of the (session?) property in the release note entry:

```
== RELEASE NOTES ==

General Changes
* Add ``optimize_union_to_union_all`` session property to optimize union with union all. 
```

Also, please add documentation for the new session property to https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/admin/properties-session.rst. 

(If there is a corresponding configuration property, please document it in https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/admin/properties.rst. )

presto-main-base/src/main/java/com/facebook/presto/sql/rewrite/UnionOptimizationRewrite.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    private static final class UnionRewriter
+            extends DefaultASTRewriter<Void>
+    {
+        @Override
+        protected Node visitUnion(Union node, Void context)
+        {
+            // Skip rewrite for union and union distinct cases
+            if (node.isDistinct().orElse(false)) {
```
NikhilCollooru, line 65: `node.isDistinct()` is true for both UNION and UNION ALL ?

we come here only if isOptimizeUnionToUnionAll is true. In such a case we should not skip for "UNION" right ?

presto-main-base/src/main/java/com/facebook/presto/sql/rewrite/UnionOptimizationRewrite.java
```diff
@@ -0,0 +1,72 @@
...
+import com.facebook.presto.sql.tree.Statement;
+import com.facebook.presto.sql.tree.Union;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+import static com.facebook.presto.SystemSessionProperties.isOptimizeUnionToUnionAll;
+
+public class UnionOptimizationRewrite
```
rschlussel, line 35: instead of the rewrite, why not check a session property in relation planner for how to handle union where distinct is empty.

rschlussel (commented):
This feature seems very sketchy.  I don't think we should do it. I also wouldn't call this an optimization. You're not improving how something executes. You're changing the semantics of UNION from UNION DISTINCT to UNION ALL, which doesn't follow SQL standard.  How would you even enable this without causing wrong results?

presto-main-base/src/main/java/com/facebook/presto/sql/rewrite/UnionOptimizationRewrite.java
jainxrohit: Its not skipping union, its only rewriting incase of union.

presto-main-base/src/main/java/com/facebook/presto/sql/rewrite/UnionOptimizationRewrite.java
jainxrohit: You are right about this being of rewrite than optimization. Are you suggesting to build this at the plan stage?
We are not planning to automatically leverage this, we are planning to make this option available to users.

jainxrohit:
Chatted with Rebecca, abandoning this PR, as this can lead to confusion while debugging.

PR #24789, opened 2025-03-24
Utilized column analyzer unnest fix map case

kevintang2022:
## Description

Currently, unnest cannot handle input expresssions that are the Map type. This is because 2 output expressions are created when unnest is used on a Map (one for key and one for value). This leads to index out of range errors because the current code only expects that Array types are passed into unnest.

Here is an example of building a mapping from output column to its corresponding input expression in the unnest.

unnest(m1, a1, m2, a2, m3,m4) t (m1k, m1v, a1v, m2k,m2v, a2v, m3k, m3v, m4k, m4v)
The correct field to expression index mapping should be 
{
 m1k: 0
m1v: 0
a1v: 1
m2k: 2
m2v: 2
a2v: 3
...
}

but it is currently

{
m1k: 0
m1v: 1
a1v: 2
m2k: 3
m2v: 4
...
}

## Motivation and Context

[T217212061](https://www.internalfb.com/intern/tasks/?t=217212061)
Currently, for utilized column analyzer, unnest will throw an index out of range exception when a map column is passed into the unnest function. This is important because if an index error is thrown, all columns in utilized tables will be checked for ACL.

## Impact

With this change, utilized column analyzer will check all expressions passed to unnest instead of checking all columns in the entire query. This will unblock cases like `unnest(m) t (c1, c2)`

## Test Plan

Unit tests in utilized column analyzer covers the cases mentioned, which were failing before.

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Fix index error when a map column is passed into an unnest function by using the column analyzer to correctly map key and value output fields back to correct input expression

### DIFF
--- a/presto-main-base/src/main/java/com/facebook/presto/sql/analyzer/UtilizedColumnsAnalyzer.java
+++ b/presto-main-base/src/main/java/com/facebook/presto/sql/analyzer/UtilizedColumnsAnalyzer.java
@@ -15,6 +15,7 @@ package com.facebook.presto.sql.analyzer;
 
 import com.facebook.airlift.log.Logger;
 import com.facebook.presto.common.QualifiedObjectName;
+import com.facebook.presto.common.type.MapType;
 import com.facebook.presto.spi.PrestoWarning;
 import com.facebook.presto.spi.WarningCollector;
 import com.facebook.presto.spi.analyzer.AccessControlInfo;
@@ -57,6 +58,7 @@ import com.google.common.collect.HashMultimap;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -339,9 +341,20 @@ public class UtilizedColumnsAnalyzer
         protected Void visitUnnest(Unnest unnest, Context context)
         {
             handleRelation(unnest, context);
-            // Prune fields that are not referenced in any outer scope
-            for (FieldId fieldId : context.getFieldIdsToExploreInRelation(unnest)) {
-                process(unnest.getExpressions().get(fieldId.getFieldIndex()), context);
+
+            List<Expression> unnestExpressions = unnest.getExpressions();
+            List<Expression> expandedExpressions = new ArrayList<>();
+            for (Expression expression : unnestExpressions) {
+                if (analysis.getType(expression) instanceof MapType) {
+                    // Map produces two output columns, so input expression gets added twice
+                    // in order for key and value columns to map back to the correct input expression
+                    expandedExpressions.add(expression);
+                }
+                expandedExpressions.add(expression);
+            }
+
+            for (FieldId field : context.getFieldIdsToExploreInRelation(unnest)) {
+                process(expandedExpressions.get(field.getFieldIndex()), context);
             }
 
             return null;
--- a/presto-main-base/src/test/java/com/facebook/presto/sql/analyzer/TestUtilizedColumnsAnalyzer.java
+++ b/presto-main-base/src/test/java/com/facebook/presto/sql/analyzer/TestUtilizedColumnsAnalyzer.java
@@ -282,6 +282,27 @@ public class TestUtilizedColumnsAnalyzer
     }
 
     @Test
+    public void testUnnestMap()
+    {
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: presto-main/src/test/java/com/facebook/presto/sql/analyzer/TestUtilizedColumnsAnalyzer.java [hunk 1]
```diff
@@ -272,15 +272,38 @@ public void testUnnest()
         assertUtilizedTableColumns("SELECT a, x, y FROM t7 CROSS JOIN UNNEST (c, d) AS t(x, y)",
                 ImmutableMap.of(QualifiedObjectName.valueOf("tpch.s1.t7"), ImmutableSet.of("a", "c", "d")));
 
-        // Unused unnest columns are pruned, but used ones are kept
+        // This test unnecessarily includes column c. This will be fixed once unnest for Map types are supported
```
kevintang2022, line 275: This is the type of cases that will experience some regression, but it should easy for the user to unblock themselves if they can just clean out the unused expressions passed into unnest

steveburnett:
Thanks for the release note entry! Suggest rephrasing to follow the [Order of Changes](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines#order-of-changes) in the Release Notes Guidelines. I adapted phrasing from your Description in the PR, but if you have a better way to phrase it, let me know.

```
== RELEASE NOTES ==

General Changes
* Fix index error when a map column is passed into an unnest function by using the column analyzer to check ACL on all input expressions for an unnest function.
```

presto-main/src/main/java/com/facebook/presto/sql/analyzer/UtilizedColumnsAnalyzer.java
```diff
@@ -339,9 +339,16 @@ protected Void visitUnion(Union union, Context context)
...
-            // Prune fields that are not referenced in any outer scope
-            for (FieldId fieldId : context.getFieldIdsToExploreInRelation(unnest)) {
-                process(unnest.getExpressions().get(fieldId.getFieldIndex()), context);
+
+            Set<FieldId> transformedFields = context.getFieldIdsToExploreInRelation(unnest);
+            List<Expression> unnestExpressions = unnest.getExpressions();
+
+            if (!transformedFields.isEmpty()) {
+                // TODO(kevintang2022): Update this fallback case to check if expressions
+                // inside the unnest are map type or array type
```
rschlussel, line 348: can we do this now by checking analysis.getType(expression)?

kevintang2022: Yes I can check for Map type

presto-main/src/main/java/com/facebook/presto/sql/analyzer/UtilizedColumnsAnalyzer.java
```diff
@@ -339,9 +341,20 @@ protected Void visitUnion(Union union, Context context)
...
             handleRelation(unnest, context);
-            // Prune fields that are not referenced in any outer scope
-            for (FieldId fieldId : context.getFieldIdsToExploreInRelation(unnest)) {
-                process(unnest.getExpressions().get(fieldId.getFieldIndex()), context);
+
+            List<Expression> unnestExpressions = unnest.getExpressions();
+            List<Expression> expandedExpressions = new ArrayList<>();
+            for (Expression expression : unnestExpressions) {
+                if (analysis.getType(expression) instanceof MapType) {
+                    // Map produces two output columns, so input expression gets added twice
```
rschlussel, line 349: technically it might be possible to do even better by adding keys and values expressions separately in case only one is used, but that seems complicated to do, so probably not worth it.

rschlussel (dismissed):
thanks for this fix!

presto-main/src/main/java/com/facebook/presto/sql/analyzer/UtilizedColumnsAnalyzer.java
kevintang2022: Yes I think the case you're referring to is 
Unnest(MAP(array_agg(a), array_agg(b))) t (c1, c2)

If only c2 is used, then we should only count b as utilized, but this change would count both a and b as utilized

rschlussel:
there's a merge conflict now. you'll need to rebase.

rschlussel (approved):

PR #24808, opened 2025-03-26
Fix flaky test: TestBrutalShutdown.testQueryRetryOnShutdown

### LINKED ISSUE #22125: Flaky TestBrutalShutdown.testQueryRetryOnShutdown
Another test that failed on a completely unrelated change (a config change to a different CI):

https://github.com/prestodb/presto/actions/runs/8203469130/job/22436277814?pr=22124

Life cycle stopping...
2024-03-08T13:29:55.2222787Z [INFO] 
2024-03-08T13:29:55.2223309Z [INFO] Results:
2024-03-08T13:29:55.2223782Z [INFO] 
2024-03-08T13:29:55.2224208Z [ERROR] Failures: 
2024-03-08T13:29:55.2228422Z [ERROR]   TestBrutalShutdown.testQueryRetryOnShutdown:83 expected [5] but found [2]
2024-03-08T13:29:55.2229497Z [INFO] 
2024-03-08T13:29:55.2231113Z [ERROR] Tests run: 3107, Failures: 1, Errors: 0, Skipped: 0
2024-03-08T13:29:55.2231792Z [INFO] 
2024-03-08T13:29:55.2268645Z [INFO] ----------------

Three things that might help here:

1. Track and report the state of unsuccessful queries when this method fails
2. Make the test single threaded
3. Change the setUp and shutdown methods to BeforeMethod and AfterMethod instead of BeforeClass and AfterClass

ZacBlanco:
## Description

Set error code `REMOTE_BUFFER_CLOSE_FAILED` as a retriable error.

(Hopefully) Fixes #22125 

## Motivation and Context

I set this test's invocationCount to 10 or 15 in order to run this on repeat and trigger the failure. I found that for all failures, it was because the query failed with `REMOTE_BUFFER_CLOSE_FAILED`. This caused the test to fail because it is not marked as a retriable error. By changing the respective error code to one that is retriable this test no longer seems to fail. I was not able to reproduce the failures on my local machine

This may have occurred due to changes in the way the task event loops are scheduled, but I am not 100% sure. Either way, I think this error code is capable of query retries, so the solution seemed suitable.

## Impact

- `StandardErrorCode#REMOTE_BUFFER_CLOSE_FAILED` is now a retriable query error.

## Test Plan

Existing tests

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

SPI Changes
* REMOTE_BUFFER_CLOSE_FAILED is now a retriable error

```


jaystarshot:
Don't think we should add **REMOTE_BUFFER_CLOSE_FAILED** as an retryble error just for unit tests.  We can add it if its indeed needed in prod too.
cc: @rschlussel

jaystarshot:
That being said i don't see many errors `REMOTE_BUFFER_CLOSE_FAILED` internally so its probably fine

ZacBlanco:
Another solution is to try and re-write the test assertion to assert that all queries which initially failed with retriable errors all eventually successfully pass - but it's possible you can get 0 queries with retriable errors, so it would defeat the purpose of the test in some runs. Since the query failure type is non-deterministic when the worker(s) shut down, which seems to be the main cause of flakiness.

I am open to other ideas

rschlussel (approved):
I think if this is an error code that you see after a forced shutdown, it makes sense to mark it as retriable.

jaystarshot (approved):

PR #24824, opened 2025-03-28
Arrow Flight tests should use open port for server

BryanCutler:
## Description
Arrow Flight tests should use a random free port instead of a fixed value to prevent test failure if the port is not available.

## Motivation and Context
Arrow Flight tests can fail if using a fixed port value that is not available.

## Test Plan
Updated Arrow Flight tests to first find an open port for the Flight server.

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-base-arrow-flight/src/test/java/com/facebook/plugin/arrow/ArrowFlightQueryRunner.java
+++ b/presto-base-arrow-flight/src/test/java/com/facebook/plugin/arrow/ArrowFlightQueryRunner.java
@@ -26,6 +26,8 @@ import org.apache.arrow.flight.Location;
 import org.apache.arrow.memory.RootAllocator;
 
 import java.io.File;
+import java.io.IOException;
+import java.net.ServerSocket;
 import java.net.URI;
 import java.util.Map;
 import java.util.Optional;
@@ -40,6 +42,14 @@ public class ArrowFlightQueryRunner
     private ArrowFlightQueryRunner()
     {
         throw new UnsupportedOperationException("This is a utility class and cannot be instantiated");
+    }
+
+    public static int findUnusedPort()
+            throws IOException
+    {
+        try (ServerSocket socket = new ServerSocket(0)) {
+            return socket.getLocalPort();
+        }
     }
 
     public static DistributedQueryRunner createQueryRunner(int flightServerPort) throws Exception
--- a/presto-base-arrow-flight/src/test/java/com/facebook/plugin/arrow/TestArrowFlightEchoQueries.java
+++ b/presto-base-arrow-flight/src/test/java/com/facebook/plugin/arrow/TestArrowFlightEchoQueries.java
@@ -113,11 +113,18 @@ public class TestArrowFlightEchoQueries
 {
     private static final Logger logger = Logger.get(TestArrowFlightEchoQueries.class);
     private static final CallOption CALL_OPTIONS = CallOptions.timeout(300, TimeUnit.SECONDS);
+    private final int serverPort;
     private RootAllocator allocator;
     private FlightServer server;
     private DistributedQueryRunner arrowFlightQueryRunner;
     private JsonCodec<TestingArrowFlightRequest> requestCodec;
     private JsonCodec<TestingArrowFlightResponse> responseCodec;
+
+    public TestArrowFlightEchoQueries()
+            throws IOException
+    {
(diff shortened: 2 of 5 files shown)


BryanCutler:
related https://github.ibm.com/lakehouse/tracker/issues/18340

presto-base-arrow-flight/src/test/java/com/facebook/plugin/arrow/ArrowFlightQueryRunner.java
```diff
@@ -115,8 +125,16 @@ public static void main(String[] args)
     {
         Logging.initialize();
 
+        int serverPort;
+        if (args.length >= 2) {
+            serverPort = Integer.valueOf(args[1]);
```
jaystarshot, line 130: Is this main function even used?

presto-base-arrow-flight/src/test/java/com/facebook/plugin/arrow/TestArrowFlightEchoQueries.java
```diff
@@ -149,21 +151,12 @@ public void close()
...
-    {
-        return Location.forGrpcTls("localhost", getServerPort());
-    }
-
     @Override
     protected QueryRunner createQueryRunner()
             throws Exception
     {
-        return ArrowFlightQueryRunner.createQueryRunner(9444);
+        serverPort = ArrowFlightQueryRunner.findUnusedPort();
```
jaystarshot, line 158: If you need serverport elsewhere can we move it to the constructor else we don't need to pass a random port

BryanCutler: Yes, `serverPort` is also needed in `setup()` when starting the server. I agree, this is a little awkward because `createQueryRunner` is called from the base class init() first. I'll try assigning the port in the constructor.

presto-base-arrow-flight/src/test/java/com/facebook/plugin/arrow/ArrowFlightQueryRunner.java
BryanCutler: I have only used it locally to run queries, which I thought is similar to [IcebergQueryRunner](https://github.com/prestodb/presto/blob/75dce5110fbc63e5167c9db89a1d64cf15556a55/presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergQueryRunner.java#L328) and others. Should I remove it?

BryanCutler:
Thanks @jaystarshot for reviewing, please take another look

presto-base-arrow-flight/src/test/java/com/facebook/plugin/arrow/ArrowFlightQueryRunner.java
jaystarshot: Yes better to remove

ZacBlanco: I think the `main` functions in query runners serve a good purpose. It's useful for quick debugging when iterating on features for a particular connector. It is also a good starting point for new developers who want to contribute to a connector as the configuration (or lack thereof) makes it faster to spin up a presto instance than using the main `PrestoServer` class. Additionally, many query runners already have one of these methods: Hive, BlackHole, Iceberg, Druid, etc. Not all of them include it, but many do. I don't think it needs to be removed

ZacBlanco: That being said, it might be prudent to split this into two commits. One for the open port fix, and one to add the main method

BryanCutler: Good point @ZacBlanco , I guess it's not really useful to pass in a port number as an arg or necessary to use a random port. I'll revert these changes.

jaystarshot: Okay if main() is useful

jaystarshot (approved):

aditi-pandit (approved):

PR #24829, opened 2025-03-30
[Iceberg]Support bucket transform for TimeType

hantangwangd:
## Description

In the Iceberg specification, columns of type TimeType support being specified bucket transforms. This PR support bucket transform for columns of type TimeType in presto.

## Motivation and Context

Support as many types as possible for partition transforms according to Iceberg specification.

## Impact

We can now use bucket transform on TimeType columns in Iceberg table.

## Test Plan

 - Newly added test case for testing the bucket transform set and used on TimeType columns of the Iceberg table.

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

Iceberg Connector Changes
* Support bucket transform for columns of type `TimeType` in Iceberg table 
```

### DIFF
--- a/presto-docs/src/main/sphinx/connector/iceberg.rst
+++ b/presto-docs/src/main/sphinx/connector/iceberg.rst
@@ -1224,7 +1224,8 @@ Transform Name        Source Types
 ``Identity``          ``boolean``, ``int``, ``bigint``, ``real``, ``double``, ``decimal``,
                       ``varchar``, ``varbinary``, ``date``, ``time``, ``timestamp``
 
-``Bucket``            ``int``, ``bigint``, ``decimal``, ``varchar``, ``varbinary``, ``date``
+``Bucket``            ``int``, ``bigint``, ``decimal``, ``varchar``, ``varbinary``, ``date``,
+                      ``time``
 
 ``Truncate``          ``int``, ``bigint``, ``decimal``, ``varchar``, ``varbinary``
 
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/PartitionTransforms.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/PartitionTransforms.java
@@ -42,6 +42,7 @@ import static com.facebook.presto.common.type.Decimals.isLongDecimal;
 import static com.facebook.presto.common.type.Decimals.isShortDecimal;
 import static com.facebook.presto.common.type.Decimals.readBigDecimal;
 import static com.facebook.presto.common.type.IntegerType.INTEGER;
+import static com.facebook.presto.common.type.TimeType.TIME;
 import static com.facebook.presto.common.type.TimestampType.TIMESTAMP;
 import static com.facebook.presto.common.type.TypeUtils.readNativeValue;
 import static com.facebook.presto.common.type.VarbinaryType.VARBINARY;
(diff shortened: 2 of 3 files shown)


steveburnett:
Thanks for the release note entry! Minor nits:

```
== RELEASE NOTES ==

Iceberg Connector Changes
* Add support for bucket transform for columns of type ``TimeType`` in Iceberg tables.
```

presto-docs/src/main/sphinx/connector/iceberg.rst
```diff
@@ -1224,7 +1224,8 @@ Transform Name        Source Types
 ``Identity``          ``boolean``, ``int``, ``bigint``, ``real``, ``double``, ``decimal``,
                       ``varchar``, ``varbinary``, ``date``, ``time``, ``timestamp``
 
-``Bucket``            ``int``, ``bigint``, ``decimal``, ``varchar``, ``varbinary``, ``date``
+``Bucket``            ``int``, ``bigint``, ``decimal``, ``varchar``, ``varbinary``, ``date``,
+                      ``time``
```
ZacBlanco, line 1228: is this intentionally put on a new line?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/PartitionTransforms.java
```diff
@@ -309,6 +316,16 @@ private static int bucketValueDate(Block block, int position, int count)
         return bucketValue(block, position, count, pos -> bucketHash(DATE.getLong(block, pos)));
     }
 
+    private static Block bucketTime(Block block, int count)
+    {
+        return bucketBlock(block, count, position -> bucketHash(MILLISECONDS.toMicros(TIME.getLong(block, position))));
```
ZacBlanco, line 321: Just to confirm my understanding, we need to do this `MILLISECONDS.toMicros` here because the bucketing needs to be compatible with the bucketing of other systems? Is ther anything else stopping us from just doing `bucketHash` directly on the `TIME.getLong(...)`?

ZacBlanco (commented):
I have one question

presto-iceberg/src/main/java/com/facebook/presto/iceberg/PartitionTransforms.java
hantangwangd: To be compatible with the bucketing of other systems is one of the reason, but the more important reason is to be compatible with Iceberg's file planning with filter. When we query the partitioned table with a filter on column of type TimeType as follows:

```
select * from test_bucket_transform_on_time where a = time '01:02:03.123';
```

We will actually first use the predicate `a = time '01:02:03.123'` to execute Iceberg's file planning. Iceberg will use this time value internally to calculate partition values, and then scan some specific partitions, if the calculated partition value is not compatible, we won't find the corresponding data. So we have to make sure that the calculation of partition value is exactly the same as Iceberg lib, that is, use the micro seconds value to calculate the partition value.

So if we doing bucketHash directly on the `TIME.getLong(...)`, we will find that the example query above get an empty result.

presto-docs/src/main/sphinx/connector/iceberg.rst
hantangwangd: Yes, otherwise we would need to increase the length of '======' above. I can modify it if you think it's more appropriate to increase the length, please let me know your thoughts.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/PartitionTransforms.java
ZacBlanco: This was roughly my thinking. Thanks for confirming

presto-docs/src/main/sphinx/connector/iceberg.rst
ZacBlanco: not necessary

ZacBlanco (approved):

PR #24830, opened 2025-03-31
[presto][ui] Display queued queries per subgroup RG

prashantgolash:
Summary: Adding num queued queries against subGroups RG to make debugging easier

Differential Revision: D72147612


yhwang (approved):
This is a good feature. I am also thinking about `numRunningQueries`. Would adding this also help with the debugging? However, at the same time, I don't want too much information added to the sub-groups list.

Let's do `numQueuedQueries` in this PR first.

prashantgolash:
> This is a good feature. I am also thinking about `numRunningQueries`. Would adding this also help with the debugging? However, at the same time, I don't want too much information added to the sub-groups list.
> 
> Let's do `numQueuedQueries` in this PR first.

I also added numRunningQueries

prashantgolash:
> > This is a good feature. I am also thinking about `numRunningQueries`. Would adding this also help with the debugging? However, at the same time, I don't want too much information added to the sub-groups list.
> > Let's do `numQueuedQueries` in this PR first.
> 
> I also added numRunningQueries

@yhwang can you please help approve when you get a chance. I also verified in the UI

yhwang (approved):
thanks for the update!

Issue #24828, opened 2025-03-30
Redis HBO Local Test Setup throws java.lang.NoSuchMethodException: com.facebook.presto.statistic.RedisProviderPlugin.<init>()


fafacao86:
## Describe the problem you faced
I was trying to turn on HBO on my local machine to test a few queries. I followed the documentation `Local Test Setup` section: https://prestodb.io/docs/current/plugin/redis-hbo-provider.html. In `presto-main/etc/config.properties`, I have added `../redis-hbo-provider/pom.xml,\` to `plugin.bundles`.
And also in `presto-main/etc/`, I have created the file `redis-provider.properties` with the sample properties.
And I added these two options in etc/config.properties too.
```
optimizer.use-history-based-plan-statistics=true
optimizer.track-history-based-plan-statistics=true
```

Then I ran `com.facebook.presto.server.PrestoServer.java` with the following JVM options
```
-ea -XX:+UseG1GC -XX:G1HeapRegionSize=32M -XX:+UseGCOverheadLimit -Djdk.attach.allowAttachSelf=true -XX:+ExplicitGCInvokesConcurrent -Dconfig=PATH_TO/presto-main/etc/config.properties -Dlog.levels-file=PATH_TO/presto-main/etc/log.properties -XX:+UnlockCommercialFeatures
```
I got the error stack:
```
2025-03-30T23:46:18.511+0800	WARN	main	com.facebook.presto.server.PluginManager	No service providers of type com.facebook.presto.spi.CoordinatorPlugin
2025-03-30T23:46:18.514+0800	ERROR	main	com.facebook.presto.server.PrestoServer	com.facebook.presto.spi.Plugin: Provider com.facebook.presto.statistic.RedisProviderPlugin could not be instantiated
java.util.ServiceConfigurationError: com.facebook.presto.spi.Plugin: Provider com.facebook.presto.statistic.RedisProviderPlugin could not be instantiated
	at java.util.ServiceLoader.fail(ServiceLoader.java:232)
	at java.util.ServiceLoader.access$100(ServiceLoader.java:185)
	at java.util.ServiceLoader$LazyIterator.nextService(ServiceLoader.java:384)
	at java.util.ServiceLoader$LazyIterator.next(ServiceLoader.java:404)
	at java.util.ServiceLoader$1.next(ServiceLoader.java:480)
	at com.google.common.collect.ImmutableList.copyOf(ImmutableList.java:278)
[...]

Issue #23070, opened 2024-06-25
Flaky test: TestNoisySumGaussianLongAggregation.testNoisySumGaussianLongClippingSomeNoiseScaleWithinSomeStd()


hantangwangd:
https://github.com/prestodb/presto/actions/runs/9654987275/job/26630141867?pr=23015

```
Error:  Failures: 
Error:    TestNoisySumGaussianLongAggregation.testNoisySumGaussianLongClippingSomeNoiseScaleWithinSomeStd:441 Test: Consistent results with reversed channels, Expected: 48.34359877440436, actual: 42.58083933484183

```

hantangwangd:
The absolute value of noise generated in `NoisyCountAggregationUtils.getNoise(state)` may exceed the value in judging conditions in the test case. It's related to probability. So the current judgement logic in `withinSomeStdDoubleAssertion` in this test case may not be very appropriate.

hantangwangd:
See again: https://github.com/prestodb/presto/actions/runs/16802762338/job/47587775331?pr=25697

PR #24835, opened 2025-04-01
Fix RedisProviderPlugin cannot be instantiated using reflection in Pl…

fafacao86:
## Description

Add back no-arg constructor in `RedisProviderPlugin`
Related issue: https://github.com/prestodb/presto/issues/24828
## Motivation and Context

When PrestoServer is bootstraping, it tries to load the RedisProviderPlugin using Reflection, which requires no-arg constructor, but the no-arg constructor in RedisProviderPlugin is somehow missing.
In `com.facebook.presto.server.PluginManager` the plugins will be instantiated by ImmutableList.copyOf using no-arg constructor reflection. The code for this is as follows: 
```
// com.facebook.presto.server.PluginManager#loadPlugin
private void loadPlugin(URLClassLoader pluginClassLoader, Class<?> clazz)
{
     ServiceLoader<?> serviceLoader = ServiceLoader.load(clazz, pluginClassLoader);
     List<?> plugins = ImmutableList.copyOf(serviceLoader);
     // ............
}
```

## Test Plan
I've tried to come up with a test, I think this is not necessary to test, and it would be a little hacky to test this.

### DIFF
--- a/redis-hbo-provider/src/main/java/com/facebook/presto/statistic/RedisProviderPlugin.java
+++ b/redis-hbo-provider/src/main/java/com/facebook/presto/statistic/RedisProviderPlugin.java
@@ -41,6 +41,11 @@ public class RedisProviderPlugin
         this(initializeConfigs(configPath));
     }
 
+    public RedisProviderPlugin()
+    {
+        this(RedisProviderConfig.REDIS_PROPERTIES_PATH);
+    }
+
     @VisibleForTesting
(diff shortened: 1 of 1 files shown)


fafacao86:
Encountered a flaky test: https://github.com/prestodb/presto/issues/23070

fafacao86:
> Please add a release note to the PR description

ok

PR #24841, opened 2025-04-01
Remove negative check on 'outputDataSizeInBytes' stat.

spershin:
## Description
Native worker can report really large outputDataSizeInBytes from some operators (like CrossJoin).
Coordinator then can have overflow in this stat and it become negative.
This fails the query and leaves it in a permanent running state.

We plan to fix this in the native worker, however, it is also useful to remove this check for some valid high memory cases.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-main-base/src/main/java/com/facebook/presto/execution/StageExecutionStats.java
+++ b/presto-main-base/src/main/java/com/facebook/presto/execution/StageExecutionStats.java
@@ -203,8 +203,10 @@ public class StageExecutionStats
 
         checkArgument(bufferedDataSizeInBytes >= 0, "bufferedDataSizeInBytes is negative");
         this.bufferedDataSizeInBytes = bufferedDataSizeInBytes;
-        checkArgument(outputDataSizeInBytes >= 0, "outputDataSizeInBytes is negative");
-        this.outputDataSizeInBytes = outputDataSizeInBytes;
+
+        // An overflow could have occurred on this stat - handle this gracefully.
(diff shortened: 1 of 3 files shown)


singcha (approved):
Thanks @spershin

presto-main-base/src/main/java/com/facebook/presto/execution/StageExecutionStats.java
```diff
@@ -203,8 +203,10 @@ public StageExecutionStats(
 
         checkArgument(bufferedDataSizeInBytes >= 0, "bufferedDataSizeInBytes is negative");
         this.bufferedDataSizeInBytes = bufferedDataSizeInBytes;
-        checkArgument(outputDataSizeInBytes >= 0, "outputDataSizeInBytes is negative");
-        this.outputDataSizeInBytes = outputDataSizeInBytes;
+
+        // An overflow could have occurred on this stat - handle this gracefully.
+        this.outputDataSizeInBytes = (outputDataSizeInBytes >= 0) ? outputDataSizeInBytes : Long.MAX_VALUE;
```
rschlussel, line 208: should we do this for all the argument checks?

spershin: @rschlussel 
Let's do it for only this 1st. Bytes are much bigger magnitude than, say, rows.
Later we can extend it for other bytes-related metrics.

rschlussel (approved):

PR #24854, opened 2025-04-02
[native] Allow options for wget and tar commands during setup Part:2

anandamideShakyan:
## Description
issue: 
The current setup scripts generate excessive output, the goal is to reduce the log output by implementing the following changes:

Avoid printing the files extracted by tar.
Silence the download progress from curl by using an appropriate curl option.
Suppress unnecessary CMake messages, logging only warnings instead of info or status messages.
Introduce a new environment variable to enable the extensive logging when needed.
These changes will help minimize log size while maintaining flexibility for detailed logging when required.

## Motivation and Context

The current setup scripts generate excessive log output, which can overwhelm users and make it difficult to identify important information. This excessive logging increases log file sizes unnecessarily, which can be problematic for storage and debugging. By implementing changes to reduce log output, we aim to streamline the logging process, improve readability, and maintain flexibility for detailed logging when required.

This part addresses since part 1 is already merged : [this](https://github.com/prestodb/presto/pull/24455#issuecomment-2768519161) comment from the reviewer.

### DIFF
--- a/presto-native-execution/scripts/setup-centos.sh
+++ b/presto-native-execution/scripts/setup-centos.sh
@@ -19,7 +19,6 @@ export CC=/opt/rh/gcc-toolset-12/root/bin/gcc
 export CXX=/opt/rh/gcc-toolset-12/root/bin/g++
 
 WGET_OPTIONS=${WGET_OPTIONS:-""}
-TAR_OPTIONS=${TAR_OPTIONS:-"-v"}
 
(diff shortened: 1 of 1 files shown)


czentgr (approved):

PingLiuPing (approved):
Thanks.

czentgr:
@anandamideShakyan Please rebase.
@majetideepak FYI.

PR #24856, opened 2025-04-02
[presto] Add state filter to the queryState resource

prashantgolash:
Summary:
It looks during recent S506379, we were trying to preempt all QUEUED queries and it was erroring out.

python client first get all the queries using `/v1/queryState` and then filter via `state=QUEUED`. Added a param filter to pass state as well which can then be used in `query.py` client lib (pass `state=QUEUED`).

While it may not always help if queued queries are way too high, but it can roughly save ~10-20%  (to filter out RUNNING/DISPATCHING/PLANNING etc) of queries with relatively safe change and may alleviate coordinator n/w pressure.

Differential Revision: D72342392

### DIFF
--- a/presto-main/src/main/java/com/facebook/presto/server/QueryStateInfoResource.java
+++ b/presto-main/src/main/java/com/facebook/presto/server/QueryStateInfoResource.java
@@ -14,6 +14,7 @@
 package com.facebook.presto.server;
 
 import com.facebook.presto.dispatcher.DispatchManager;
+import com.facebook.presto.execution.QueryState;
 import com.facebook.presto.execution.resourceGroups.ResourceGroupManager;
 import com.facebook.presto.metadata.InternalNode;
 import com.facebook.presto.metadata.InternalNodeManager;
@@ -42,8 +43,10 @@ import javax.ws.rs.core.Response;
 import javax.ws.rs.core.UriInfo;
 
 import java.net.URI;
+import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Locale;
 import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.OptionalInt;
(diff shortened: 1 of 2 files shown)


presto-main/src/main/java/com/facebook/presto/server/QueryStateInfoResource.java
```diff
@@ -106,8 +109,14 @@ public void getQueryStateInfos(
             List<BasicQueryInfo> queryInfos = dispatchManager.getQueries();
             Optional<Pattern> userPattern = isNullOrEmpty(user) ? Optional.empty() : Optional.of(Pattern.compile(user));
 
+            final QueryState expectedQueryState = stateFilter == null ? null : QueryState.valueOf(stateFilter.toUpperCase(Locale.ENGLISH));
```
rschlussel, line 112: maybe throw a user friendly exception if the stateFilter is not a valid queryState

rschlussel (commented):
Thanks for this change

presto-main/src/main/java/com/facebook/presto/server/QueryStateInfoResource.java
prashantgolash: > maybe throw a user friendly exception if the stateFilter is not a valid queryState
Done.

rschlussel (approved):

rschlussel:
Fix the checkstyle errors. once the CI is green should be good to go.

prashantgolash:
> Fix the checkstyle errors. once the CI is green should be good to go.

@rschlussel  i fixed check-style. Also the current failure doesn't look related to my changed. Appreciate any comments on how to move forward.

rschlussel (approved):

aditi-pandit:
HI @prashantgolash @rschlussel : Would be great for this change to have a release note for the new 0.293 release. Can you help me draft something ? If you can reply to this comment in the PR with the text, then I can help put in the requisite docs. 

Thanks in advance.

rschlussel:
> HI @prashantgolash @rschlussel : Would be great for this change to have a release note for the new 0.293 release. Can you help me draft something ? If you can reply to this comment in the PR with the text, then I can help put in the requisite docs.
> 
> Thanks in advance.

REST API Changes
* Add parameter to the queryState resource to enable filtering based on query state. For example, "/v1/queryState/state=QUEUED" will return all queued queries.

PR #24867, opened 2025-04-04
Fix error when describing a nonexistent table

elbinpallimalilibm:
## Description

This change fixes a null pointer exception when table columns were attempted to be listed for a nonexistent table

## Motivation and Context

If a statement like `DESCRIBE arrowdb2.information_schema.enabled_roles` was executed, where `enabled_roles` table does not exist in `information_schema` in the remote database, this would result in a null pointer exception. 

This error occurs in special scenarios where a table like `enabled_roles` exist in the Presto `information_schema` schema, so statement analyzer does not throw "table not found" error. In normal scenario, statement analyzer would throw "table not found" error and execution does not reach the point where table handle is retrieved from remote datasource.

## Test Plan

New unit test added to test this special scenario.

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-base-arrow-flight/src/test/java/com/facebook/plugin/arrow/TestArrowFlightQueries.java
+++ b/presto-base-arrow-flight/src/test/java/com/facebook/plugin/arrow/TestArrowFlightQueries.java
@@ -44,6 +44,7 @@ import static com.facebook.presto.common.type.TimestampType.TIMESTAMP;
 import static com.facebook.presto.common.type.VarcharType.VARCHAR;
 import static com.facebook.presto.common.type.VarcharType.createVarcharType;
 import static com.facebook.presto.testing.MaterializedResult.resultBuilder;
+import static com.facebook.presto.testing.assertions.Assert.assertEquals;
 import static java.lang.String.format;
 import static org.testng.Assert.assertTrue;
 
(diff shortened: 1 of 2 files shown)


presto-base-arrow-flight/src/test/java/com/facebook/plugin/arrow/TestArrowFlightQueries.java
```diff
@@ -137,6 +138,16 @@ public void testSelectTime()
...
     }
 
+    @Test
+    public void testDescribeUnknownTable()
+    {
+        MaterializedResult actualRows = computeActual("DESCRIBE information_schema.enabled_roles");
+        MaterializedResult expectedRows = resultBuilder(getSession(), VARCHAR, VARCHAR, VARCHAR, VARCHAR)
+                .row("role_name", "varchar", "", "")
+                .build();
+        assertEquals(actualRows, expectedRows, format("%s does not match %s", actualRows, expectedRows));
```
pramodsatya, line 148: nit: `assertEquals(actualRows, expectedRows)` should be fine?

pramodsatya (dismissed):
Thanks @elbinpallimalilibm.

presto-base-arrow-flight/src/test/java/com/facebook/plugin/arrow/TestArrowFlightQueries.java
elbinpallimalilibm: Updated, can you review?

jaystarshot:
If you don't want to throw NPE, then please throw a Presto exception in this case we shouldn't ignore and continue when ConnectorTableHandle is null

elbinpallimalilibm:
> If you don't want to throw NPE, then please throw a Presto exception in this case we shouldn't ignore and continue when ConnectorTableHandle is null

`DESCRIBE arrowdb2.information_schema.enabled_roles`, but we are expecting a response for this query right? Since information_schema is a Presto schema. Why should we throw an exception in this case?

jaystarshot:
Okay i would assume that this function shouldn’t be called in that case. Anyways looks like other connectors also have such code so I think its fine

jaystarshot (approved):

elbinpallimalilibm:
@jaystarshot Can you merge?

PR #24874, opened 2025-04-06
Reuse JoinType in IndexJoinNode

zacw7:
Reuse JoinType instead of creating IndexJoinNode's own. JoinType is already part of Prestissimo protocol. Adding IndexJoinNode with another JoinType would cause naming conflict.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-spi/src/main/java/com/facebook/presto/spi/plan/JoinType.java
+++ b/presto-spi/src/main/java/com/facebook/presto/spi/plan/JoinType.java
@@ -20,7 +20,8 @@ public enum JoinType
     INNER("InnerJoin"),
     LEFT("LeftJoin"),
     RIGHT("RightJoin"),
-    FULL("FullJoin");
+    FULL("FullJoin"),
+    SOURCE_OUTER("SourceOuter");
 
     private final String joinLabel;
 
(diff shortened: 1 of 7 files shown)


presto-spi/src/main/java/com/facebook/presto/spi/plan/JoinType.java
```diff
@@ -20,7 +20,8 @@ public enum JoinType
     INNER("InnerJoin"),
     LEFT("LeftJoin"),
     RIGHT("RightJoin"),
-    FULL("FullJoin");
+    FULL("FullJoin"),
+    SOURCE_OUTER("SourceOuter");
```
xiaoxmeng, line 24: What does SOURCE_OUTER mean? Left join?

xiaoxmeng (commented):
@zacw7 thanks for the change!

presto-spi/src/main/java/com/facebook/presto/spi/plan/JoinType.java
zacw7: I haven't got the chance to dig deep in the execution. Based on the code, if it's `SOURCE_OUTER`, Presto execution planner will create a LookupJoinOperator with JoinType PROBE_OUTER (the Probe is the outer side of the join). (https://github.com/prestodb/presto/blob/47a708022aba003ebfda483ea3c940231fdbd282/presto-main-base/src/main/java/com/facebook/presto/operator/LookupJoinOperators.java#L99-L110)

xiaoxmeng (approved):
@zacw7 thanks!

NikhilCollooru (approved):

Issue #24510, opened 2025-02-06
Add distinct if needed to the right side of a semijoin


kaikalur:
Looks like SetBuilder works best when you give it distinct elements. So when we see something like:

select ... from t1 
where id in (select id from t2)

we should distinct id in t2

PR #24884, opened 2025-04-08
Add distinct to the right side when LOJ + IS NULL is rewritten to Semijoin

tanjialiang:
## Description

issue https://github.com/prestodb/presto/issues/24510

The join might have a huge right side in cases of following optimization:
When left join has the 'is null' key filter on the right side, it is effectively making the query return rows from the left side where there is a no match on the right side. This currently is optimized by converting the section into a left semi join. But the current issue is the right side (key only) might have large amount of duplication that is completely unnecessary for evaluation of the join. This PR addresses it by adding a distinct aggregation operator before to optimize the performance

Selective Meta internal production queries showed 100x performance gain and 3x memory reduction
```
== RELEASE NOTE ==
General Changes
Add distinct on the right side optimization on optimizer rule ``LeftJoinNullFilterToSemiJoin``.
```

### DIFF
--- a/presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java
+++ b/presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java
@@ -7426,10 +7426,26 @@ public abstract class AbstractTestQueries
         // null in both sides
         sql = "with t1 as (select * from (values 1, 1, 2, 2, 3, 3, null) t(k)), t2 as (select * from (values 1, 1, 2, 2, null) t(k)) select t1.* from t1 left join t2 on t1.k=t2.k where t2.k is null";
         assertQuery(enableOptimization, sql, "values 3, 3, null");
+        sql = "with t1 as (select * from (values 1, 1, 2, 2, 3, 3, null) t(k)), t2 as (select * from (values 1, 1, 2, 2, null, null, null, null) t(k)) select t1.* from t1 left join t2 on t1.k=t2.k where t2.k is null";
+        assertQuery(enableOptimization, sql, "values 3, 3, null");
+        assertNotEquals(computeActual("EXPLAIN(TYPE DISTRIBUTED) " + sql).getOnlyValue().toString().indexOf("Aggregate"), -1);
 
         // null in right side
         sql = "with t1 as (select * from (values 1, 1, 2, 2, 3, 3) t(k)), t2 as (select * from (values 1, 1, 2, 2, null) t(k)) select t1.* from t1 left join t2 on t1.k=t2.k where t2.k is null";
         assertQuery(enableOptimization, sql, "values 3, 3");
+        sql = "with t1 as (select * from (values 1, 1, 2, 2, 3, 3) t(k)), t2 as (select * from (values 1, 1, 2, 2, null, null, null, null) t(k)) select t1.* from t1 left join t2 on t1.k=t2.k where t2.k is null";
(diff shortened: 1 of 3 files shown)


presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java
```diff
@@ -7426,10 +7426,14 @@ public void testLeftJoinNullFilterToSemiJoin()
         // null in both sides
         sql = "with t1 as (select * from (values 1, 1, 2, 2, 3, 3, null) t(k)), t2 as (select * from (values 1, 1, 2, 2, null) t(k)) select t1.* from t1 left join t2 on t1.k=t2.k where t2.k is null";
         assertQuery(enableOptimization, sql, "values 3, 3, null");
+        sql = "with t1 as (select * from (values 1, 1, 2, 2, 3, 3, null) t(k)), t2 as (select * from (values 1, 1, 2, 2, null, null, null, null) t(k)) select t1.* from t1 left join t2 on t1.k=t2.k where t2.k is null";
+        assertQuery(enableOptimization, sql, "values 3, 3, null");
 
         // null in right side
         sql = "with t1 as (select * from (values 1, 1, 2, 2, 3, 3) t(k)), t2 as (select * from (values 1, 1, 2, 2, null) t(k)) select t1.* from t1 left join t2 on t1.k=t2.k where t2.k is null";
         assertQuery(enableOptimization, sql, "values 3, 3");
+        sql = "with t1 as (select * from (values 1, 1, 2, 2, 3, 3) t(k)), t2 as (select * from (values 1, 1, 2, 2, null, null, null, null) t(k)) select t1.* from t1 left join t2 on t1.k=t2.k where t2.k is null";
```
kaikalur, line 7435: Also add a test that checks the plan (maybe using explain) to have a distinct with the optimization on (and absent without)

kaikalur (changes requested):

kaikalur:
Also this only partially addresses the issue linked because there could be other direct uses of semijoin that we don't address. This just improves on the original optimization.

kaikalur:
Add couple more tests:

1. already distinct on the right side (and make sure you don't get double distinct - not incorrect but just unncessary)
2. add group by like: .. (select custkey from orders group by custkey having count(1) > 5) and see if you get the correct plan

kaikalur (dismissed):
LGTM

kaikalur:
@jaystarshot - please take a look. Thanks!

jaystarshot (dismissed):
Can you please also add a release note?

jaystarshot:
This only applies an agg if the previous join is a Left join, isn't it better to apply this to all semi joins for a general case (as mentioned in the [issue](https://github.com/prestodb/presto/issues/24510)) ?

tanjialiang:
> This only applies an agg if the previous join is a Left join, isn't it better to apply this to all semi joins for a general case (as mentioned in the [issue](https://github.com/prestodb/presto/issues/24510)) ?

Hi @jaystarshot, thanks for your review. Yeah we are going to follow up on extracting this out as a more general case support as a followup. We will keep it updated in the linked issue.

steveburnett:
Thanks for the release note! Some formatting nits. 

```
== RELEASE NOTE ==

General Changes
Add distinct on the right side optimization on optimizer rule ``LeftJoinNullFilterToSemiJoin``.
```

kaikalur (approved):

tanjialiang:
Hi @jaystarshot could you help with another stamp after rebase? Thank you

Issue #24894, opened 2025-04-09
Set table qualified name for alter and drop operations at query events


evanvdia:
For ALTER and DROP queries, there are no input/output datasets. Since there are no input/output datasets, we cannot pass the source/destination table details to the query events.  Implement the changes for set the qualified table name for alter and drop operations at query events.

## Presto Component, Service, or Connector

presto-main, presto-spi, presto-eventlistener

PR #24899, opened 2025-04-10
Enrich Update information in QueryInfo 

evanvdia:
## Description
Add additional information to about DDL operations to QueryInfo

## Motivation and Context
Issue: https://github.com/prestodb/presto/issues/24894

updateType tracking in QueryInfo was added in https://github.com/prestodb/presto/commit/3715be85ef5f5886efceddcbc3fb89131d6f72fc

For lineage tracing, we would like to enhance this with additional information about the object(s) we are updating

## Impact
User's can track what tables/columns/schemas were operated on in query info and event listener data

## Test Plan
CI

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-cli/src/main/java/com/facebook/presto/cli/Query.java
+++ b/presto-cli/src/main/java/com/facebook/presto/cli/Query.java
@@ -160,7 +160,7 @@ public class Query
         // if running or finished
         if (client.isRunning() || (client.isFinished() && client.finalStatusInfo().getError() == null)) {
             QueryStatusInfo results = client.isRunning() ? client.currentStatusInfo() : client.finalStatusInfo();
-            if (results.getUpdateType() != null) {
+            if (results.getUpdateInfo() != null) {
                 renderUpdate(errorChannel, results);
             }
             else if (results.getColumns() == null) {
@@ -220,7 +220,8 @@ public class Query
 
     private void renderUpdate(PrintStream out, QueryStatusInfo results)
     {
-        String status = results.getUpdateType();
+        checkState(results.getUpdateInfo() != null, "Update info is null");
+        String status = results.getUpdateInfo().getUpdateType();
         if (results.getUpdateCount() != null) {
             long count = results.getUpdateCount();
             status += format(": %s row%s", count, (count != 1) ? "s" : "");
--- a/presto-hive/src/test/java/com/facebook/presto/hive/TestHiveIntegrationSmokeTest.java
+++ b/presto-hive/src/test/java/com/facebook/presto/hive/TestHiveIntegrationSmokeTest.java
@@ -5847,7 +5847,7 @@ public class TestHiveIntegrationSmokeTest
         assertQueryFails(
                 "CREATE MATERIALIZED VIEW test_customer_view AS SELECT name FROM test_customer_base",
                 format(
-                        ".* Materialized view '%s.%s.test_customer_view' already exists",
+                        ".* Destination materialized view '%s.%s.test_customer_view' already exists",
                         getSession().getCatalog().get(),
                         getSession().getSchema().get()));
         assertQuerySucceeds("CREATE MATERIALIZED VIEW IF NOT EXISTS test_customer_view AS SELECT name FROM test_customer_base");
--- a/presto-main-base/src/main/java/com/facebook/presto/execution/DDLDefinitionExecution.java
+++ b/presto-main-base/src/main/java/com/facebook/presto/execution/DDLDefinitionExecution.java
@@ -17,8 +17,13 @@ import com.facebook.presto.common.analyzer.PreparedQuery;
 import com.facebook.presto.common.resourceGroups.QueryType;
 import com.facebook.presto.metadata.Metadata;
 import com.facebook.presto.metadata.MetadataManager;
+import com.facebook.presto.spi.VariableAllocator;
 import com.facebook.presto.spi.WarningCollector;
+import com.facebook.presto.spi.analyzer.AnalyzerContext;
 import com.facebook.presto.spi.analyzer.AnalyzerProvider;
+import com.facebook.presto.spi.analyzer.QueryAnalysis;
+import com.facebook.presto.spi.analyzer.QueryAnalyzer;
+import com.facebook.presto.spi.plan.PlanNodeIdAllocator;
 import com.facebook.presto.spi.security.AccessControl;
 import com.facebook.presto.sql.analyzer.BuiltInQueryPreparer;
 import com.facebook.presto.sql.tree.Expression;
@@ -32,6 +37,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
+import static com.facebook.presto.util.AnalyzerUtil.getAnalyzerContext;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkState;
 import static java.util.Objects.requireNonNull;
@@ -102,6 +108,13 @@ public class DDLDefinitionExecution<T extends Statement>
             checkState(preparedQuery instanceof BuiltInQueryPreparer.BuiltInPreparedQuery, "Unsupported prepared query type: %s", preparedQuery.getClass().getSimpleName());
             BuiltInQueryPreparer.BuiltInPreparedQuery builtInQueryPreparer = (BuiltInQueryPreparer.BuiltInPreparedQuery) preparedQuery;
 
+            // Run thru analyzer to get info about the update
+            QueryAnalyzer queryAnalyzer = analyzerProvider.getQueryAnalyzer();
+            AnalyzerContext analyzerContext = getAnalyzerContext(queryAnalyzer,
+                    metadata.getMetadataResolver(stateMachine.getSession()), new PlanNodeIdAllocator(), new VariableAllocator(), stateMachine.getSession(), query);
+            QueryAnalysis analysis = queryAnalyzer.analyze(analyzerContext, preparedQuery);
+            stateMachine.setUpdateInfo(analysis.getUpdateInfo());
+
             return createDDLDefinitionExecution(builtInQueryPreparer.getStatement(), builtInQueryPreparer.getParameters(), stateMachine, slug, retryCount, query);
         }
 
@@ -117,7 +130,6 @@ public class DDLDefinitionExecution<T extends Statement>
             DDLDefinitionTask<T> task = (DDLDefinitionTask<T>) tasks.get(statement.getClass());
             checkArgument(task != null, "no task for statement: %s", statement.getClass().getSimpleName());
 
-            stateMachine.setUpdateType(task.getName());
             return new DDLDefinitionExecution<>(task, statement, slug, retryCount, transactionManager, metadata, accessControl, stateMachine, parameters, query);
         }
     }
--- a/presto-main-base/src/main/java/com/facebook/presto/execution/SessionDefinitionExecution.java
+++ b/presto-main-base/src/main/java/com/facebook/presto/execution/SessionDefinitionExecution.java
@@ -19,6 +19,7 @@ import com.facebook.presto.metadata.Metadata;
 import com.facebook.presto.metadata.MetadataManager;
 import com.facebook.presto.spi.WarningCollector;
 import com.facebook.presto.spi.analyzer.AnalyzerProvider;
+import com.facebook.presto.spi.analyzer.UpdateInfo;
 import com.facebook.presto.spi.security.AccessControl;
 import com.facebook.presto.sql.analyzer.BuiltInQueryPreparer;
 import com.facebook.presto.sql.tree.Expression;
@@ -116,7 +117,7 @@ public class SessionDefinitionExecution<T extends Statement>
             SessionTransactionControlTask<T> task = (SessionTransactionControlTask<T>) tasks.get(statement.getClass());
             checkArgument(task != null, "no task for statement: %s", statement.getClass().getSimpleName());
 
-            stateMachine.setUpdateType(task.getName());
+            stateMachine.setUpdateInfo(new UpdateInfo(task.getName(), ""));
             return new SessionDefinitionExecution<>(task, statement, slug, retryCount, transactionManager, metadata, accessControl, stateMachine, parameters, query);
         }
     }
--- a/presto-main-base/src/main/java/com/facebook/presto/sql/analyzer/StatementAnalyzer.java
+++ b/presto-main-base/src/main/java/com/facebook/presto/sql/analyzer/StatementAnalyzer.java
@@ -78,9 +78,11 @@ import com.facebook.presto.sql.tree.Call;
 import com.facebook.presto.sql.tree.Commit;
 import com.facebook.presto.sql.tree.CreateFunction;
 import com.facebook.presto.sql.tree.CreateMaterializedView;
+import com.facebook.presto.sql.tree.CreateRole;
 import com.facebook.presto.sql.tree.CreateSchema;
 import com.facebook.presto.sql.tree.CreateTable;
 import com.facebook.presto.sql.tree.CreateTableAsSelect;
+import com.facebook.presto.sql.tree.CreateType;
 import com.facebook.presto.sql.tree.CreateView;
 import com.facebook.presto.sql.tree.Cube;
 import com.facebook.presto.sql.tree.Deallocate;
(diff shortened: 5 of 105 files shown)


aaneja (changes requested):
The implementation you have is limited in scope to specific table Statement types. We also have an existing field in QueryInfo called `updateType` for tracing the type of update statement that was made.

IMO we should make changes to start tracing more information in a POJO that we attach during query analysis. I've made a (draft) PR for this - https://github.com/prestodb/presto/pull/24933

Please take a look, and feel free to productionize it

evanvdia:
@aaneja Will check and implement these changes.

presto-main-base/src/main/java/com/facebook/presto/execution/DDLDefinitionExecution.java
```diff
@@ -102,6 +108,13 @@ public DDLDefinitionExecution<?> createQueryExecution(
             checkState(preparedQuery instanceof BuiltInQueryPreparer.BuiltInPreparedQuery, "Unsupported prepared query type: %s", preparedQuery.getClass().getSimpleName());
             BuiltInQueryPreparer.BuiltInPreparedQuery builtInQueryPreparer = (BuiltInQueryPreparer.BuiltInPreparedQuery) preparedQuery;
 
+            // Run thru analyzer to get info about the update
+            QueryAnalyzer queryAnalyzer = analyzerProvider.getQueryAnalyzer();
+            AnalyzerContext analyzerContext = getAnalyzerContext(queryAnalyzer,
+                    metadata.getMetadataResolver(stateMachine.getSession()), new PlanNodeIdAllocator(), new VariableAllocator(), stateMachine.getSession(), query);
+            QueryAnalysis analysis = queryAnalyzer.analyze(analyzerContext, preparedQuery);
+            stateMachine.setUpdateInfo(analysis.getUpdateInfo());
```
aaneja, line 116: The core of the change is here - we now run DDL statements through the analyzer and use `analysis.getUpdateInfo()` to set the update info
This responsibility to set the correct/apt info about the DDL operation will now rest solely on `StatementAnalyzer`

presto-spi/src/main/java/com/facebook/presto/spi/analyzer/UpdateInfo.java
```diff
@@ -0,0 +1,79 @@
...
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.Objects;
+
+@ThriftStruct
+public class UpdateInfo
+{
+    private final String updateType;
+    private final String updateObject;
```
aaneja, line 29: DDL statements can stringify what is relevant to them to track and store it in this `updateObject`

presto-main-base/src/main/java/com/facebook/presto/sql/analyzer/StatementAnalyzer.java
```diff
@@ -941,35 +945,82 @@ protected Scope visitResetSession(ResetSession node, Optional<Scope> scope)
...
+        protected Scope visitCreateRole(CreateRole node, Optional<Scope> scope)
+        {
+            analysis.setUpdateInfo(new UpdateInfo("CREATE ROLE", node.getName().toString()));
+            return createAndAssignScope(node, scope);
+        }
+
+        @Override
+        protected Scope visitDropRole(DropRole node, Optional<Scope> scope)
+        {
+            analysis.setUpdateInfo(new UpdateInfo("CREATE ROLE", node.getName().toString()));
```
ZacBlanco: `"DROP ROLE"`?

presto-tests/src/main/java/com/facebook/presto/tests/TestingPrestoClient.java
```diff
@@ -110,12 +111,12 @@ private class MaterializedResultSession
 
         private final AtomicReference<List<Type>> types = new AtomicReference<>();
 
-        private final AtomicReference<Optional<String>> updateType = new AtomicReference<>(Optional.empty());
+        private final AtomicReference<Optional<UpdateInfo>> updateType = new AtomicReference<Optional<com.facebook.presto.spi.analyzer.UpdateInfo>>(Optional.empty());
```
ZacBlanco: Is there a reason we use the fully-qualified name here?

ZacBlanco (dismissed):
minor things. Also wondering if there were other ways we considered implementing this.

It might be worth creating an enum instead to represent all the DDL types? Another way to do this too would be creating a top-level interface such as `interface UpdateInfo` and let the implementations just define themselves as serializable json objects rather than having the explicit field that is serializable.

presto-main-base/src/main/java/com/facebook/presto/sql/analyzer/StatementAnalyzer.java
evanvdia: @ZacBlanco Done. I have moved the DDL types to enum.

presto-tests/src/main/java/com/facebook/presto/tests/TestingPrestoClient.java
evanvdia: Removed

presto-hive/src/test/java/com/facebook/presto/hive/TestHiveIntegrationSmokeTest.java
```diff
@@ -5847,7 +5847,7 @@ public void testCreateMaterializedView()
         assertQueryFails(
                 "CREATE MATERIALIZED VIEW test_customer_view AS SELECT name FROM test_customer_base",
                 format(
-                        ".* Materialized view '%s.%s.test_customer_view' already exists",
+                        ".* Destination materialized view '%s.%s.test_customer_view' already exists",
```
ZacBlanco, line 5850: Why is this updated?

presto-spi/src/main/java/com/facebook/presto/spi/analyzer/UpdateInfo.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @JsonCreator
+    @ThriftConstructor
+    public UpdateInfo(
+            @JsonProperty("updateType") String updateType,
```
ZacBlanco, line 43: We could use the enum type here instead of a string, otherwise the enum is a bit useless. I suggested the enum originally as an idea to make the update type less error prone so that you couldn't put any random string as the update type.

presto-hive/src/test/java/com/facebook/presto/hive/TestHiveIntegrationSmokeTest.java
evanvdia: @ZacBlanco The test case was failing due to a mismatch in the exception message.

presto-spi/src/main/java/com/facebook/presto/spi/analyzer/UpdateInfo.java
evanvdia: Moved to enum.

evanvdia:
Could you please look into this when you get a chance?"

ethanyzhang:
Hi @ZacBlanco would it be possible for you to take a look?

presto-spi/src/main/java/com/facebook/presto/spi/analyzer/UpdateType.java
```diff
@@ -0,0 +1,79 @@
...
+    ALTER_COLUMN_NOT_NULL("ALTER COLUMN NOT NULL"),
+    RENAME_VIEW("RENAME VIEW"),
+    DROP_VIEW("DROP VIEW"),
+    DROP_MATERIALIZED_VIEW("DROP MATERIALIZED VIEW"),
+    GRANT("GRANT"),
+    REVOKE("REVOKE"),
+    CALL("CALL"),
+    ADD_CONSTRAINT("ADD CONSTRAINT"),
+    UPDATE_TYPE("UPDATE TYPE"),
+    SET_SESSION("SET SESSION");
```
hantangwangd: It seems that we missed several types here, such as `USE`, `START TRANSACTION`, `COMMIT`, and so on. This will result in the `UpdateType` being null when building the `UpdateInfo` for these statements, leading to subsequent errors on the client side, such as in the CLI console.

presto-spi/src/main/java/com/facebook/presto/spi/analyzer/UpdateInfo.java
```diff
@@ -0,0 +1,79 @@
...
+    private final String updateObject;
+
+    @JsonCreator
+    @ThriftConstructor
+    public UpdateInfo(
+            @JsonProperty("updateType") UpdateType updateType,
+            @JsonProperty("updateObject") String updateObject)
+    {
+        this.updateType = updateType;
+        this.updateObject = updateObject;
```
hantangwangd: Should we do the `requireNonNull` check for `updateType` here?

presto-cli/src/main/java/com/facebook/presto/cli/Query.java
```diff
@@ -220,7 +220,7 @@ private void processInitialStatusUpdates(WarningsPrinter warningsPrinter)
 
     private void renderUpdate(PrintStream out, QueryStatusInfo results)
     {
-        String status = results.getUpdateType();
+        String status = results.getUpdateInfo().getUpdateType().getUpdateType();
```
hantangwangd: As described below, we will get NPE here for some kinds of statements such as `USE`, `START TRANSACTION` etc. So I guess we have to make sure that if `results.updateInfo` exists, it's `updateType` must not be null?

presto-main-base/src/main/java/com/facebook/presto/execution/SessionDefinitionExecution.java
```diff
@@ -116,7 +118,7 @@ private <T extends Statement> SessionDefinitionExecution<T> createSessionDefinit
             SessionTransactionControlTask<T> task = (SessionTransactionControlTask<T>) tasks.get(statement.getClass());
             checkArgument(task != null, "no task for statement: %s", statement.getClass().getSimpleName());
 
-            stateMachine.setUpdateType(task.getName());
+            stateMachine.setUpdateInfo(new UpdateInfo(UpdateType.getValue(task.getName()), ""));
```
hantangwangd: Here will get null update types for some kinds of session definition statements.

presto-cli/src/main/java/com/facebook/presto/cli/Query.java
evanvdia: Done

evanvdia: Added null check

presto-spi/src/main/java/com/facebook/presto/spi/analyzer/UpdateType.java
evanvdia: Done

presto-spi/src/main/java/com/facebook/presto/spi/analyzer/UpdateInfo.java
evanvdia: Done

presto-main-base/src/main/java/com/facebook/presto/execution/SessionDefinitionExecution.java
evanvdia: Done

tdcmeehan:
> Another way to do this too would be creating a top-level interface such as interface UpdateInfo and let the implementations just define themselves as serializable json objects rather than having the explicit field that is serializable.

@evanvdia I like this approach--have you investigated how hard it would be to simply tie the update type to each Statement?

evanvdia:
> > Another way to do this too would be creating a top-level interface such as interface UpdateInfo and let the implementations just define themselves as serializable json objects rather than having the explicit field that is serializable.
> 
> @evanvdia I like this approach--have you investigated how hard it would be to simply tie the update type to each Statement?

@tdcmeehan I haven’t tried this approach yet. Could you please elaborate a bit more? I didn’t fully understand it. What’s your perspective?

evanvdia:
> > > Another way to do this too would be creating a top-level interface such as interface UpdateInfo and let the implementations just define themselves as serializable json objects rather than having the explicit field that is serializable.
> > 
> > 
> > @evanvdia I like this approach--have you investigated how hard it would be to simply tie the update type to each Statement?
> 
> @tdcmeehan I haven’t tried this approach yet. Could you please elaborate a bit more? I didn’t fully understand it. What’s your perspective?

@tdcmeehan Implemented the suggested approach. Could you please review this.

tdcmeehan (changes requested):
With this latest round of updates, I'm not sure if it makes sense to retain the `UpdateType` enum--if each `Statement` is in charge of returning an object that describes what type of statement it is, plus some additional updates, then we should just look at subclasses of `Statement` rather than duplicate that in a `UpdateType` enum.  I would just have it return an object, which can be serialized, which just describes the type of statement plus some additional information.  We can also remove the `UpdateInfoProvider` method and just put an abstract method directly in `Statement`.

evanvdia:
> With this latest round of updates, I'm not sure if it makes sense to retain the `UpdateType` enum--if each `Statement` is in charge of returning an object that describes what type of statement it is, plus some additional updates, then we should just look at subclasses of `Statement` rather than duplicate that in a `UpdateType` enum. I would just have it return an object, which can be serialized, which just describes the type of statement plus some additional information. We can also remove the `UpdateInfoProvider` method and just put an abstract method directly in `Statement`.

@tdcmeehan As per the previous discussion https://github.com/prestodb/presto/pull/24899#discussion_r2127737143 with  @ZacBlanco  We moved the types to an enum as a way to make the updateType less error-prone. So, is it okay if I keep it as-is? and will proceed with implementing the other changes.

Do you think I should rename `UpdateType` to something more context-specific, or is it okay to keep the current name?

tdcmeehan:
I feel like in light of this approach, the text of the "type" isn't random anymore because it will be coming from the Statement classes themselves.

evanvdia:
@tdcmeehan i have implemented the changes. Could you please review.

evanvdia:
@tdcmeehan Could you please take a look into this?

presto-spi/src/main/java/com/facebook/presto/spi/analyzer/UpdateInfo.java
```diff
@@ -0,0 +1,81 @@
...
+import com.facebook.drift.annotations.ThriftStruct;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.Objects;
+
+import static java.util.Objects.requireNonNull;
+
+@ThriftStruct
+public class UpdateInfo
```
tdcmeehan, line 28: Please add some documentation that describes what type of values go into each field, preferably with an example.

tdcmeehan (commented):
LGTM % adding some documentation on `UpdateType`

presto-spi/src/main/java/com/facebook/presto/spi/analyzer/UpdateInfo.java
evanvdia: @tdcmeehan Done

tdcmeehan (approved):

evanvdia:
Two unit test cases failed with the error below. 

1.This unit test case is working fine locally
image

2.
image

Could you please help look into it?

evanvdia:
@hantangwangd Could you please take a look into this? https://github.com/prestodb/presto/pull/24899#issuecomment-3067600494

hantangwangd:
I think the second is an occasional failure, and the first one is a reported known flaky test which is unrelated to this change. Both can be handled by rerunning.

evanvdia:
> I think the second is an occasional failure, and the first one is a reported known flaky test which is unrelated to this change. Both can be handled by rerunning.

@hantangwangd Thanks for the update!

hantangwangd (approved):

ethanyzhang:
Hi @ZacBlanco can you recheck this and approve if all looks good to you? Thank you!

tdcmeehan:
I think we can follow lazy consensus and merge if there's no further objections.  I've dismissed @ZacBlanco and @aaneja's reviews, PTAL if you have any objections, otherwise I think we can merge tomorrow.

rschlussel:
@evanvdia @tdcmeehan this was a backwards incompatible change for the presto-client (changing updateType to updateInfo in the QueryResults).  It's causing older presto-jdbc clients to fail on update queries.

rschlussel:
Revert PR here: https://github.com/prestodb/presto/pull/25740

PR #24912, opened 2025-04-14
Reduce memory usage in reader by reset all underlying readers

chenyangfb:
## Description
Currently for reader with multiple underlying readers (e.g. dictionaryReader and directReader in LongSelectiveStreamReader), only the current reader get reset in startStripe(), the other reader won't get reset, leads to extra memory usage and OOM.

This PR avoid those extra memory usage by reset all underlying readers in startStripe(). This behavior is controlled by resetAllReaders in OrcReaderOptions, and it's disabled by default

## Impact
Reduce memory usage in reader.

## Test Plan
Tested with Spark workload.
Around 10% workload triggered the code path which reset all underlying readers, leading to less memory usage and OOM.

```
== RELEASE NOTES ==
General change
* Improve memory usage in reader with nested readers by resetting all nested readers.
```

### DIFF
--- a/presto-orc/src/main/java/com/facebook/presto/orc/reader/BatchStreamReaders.java
+++ b/presto-orc/src/main/java/com/facebook/presto/orc/reader/BatchStreamReaders.java
@@ -40,7 +40,7 @@ public final class BatchStreamReaders
             case INT:
             case LONG:
             case DATE:
-                return new LongBatchStreamReader(type, streamDescriptor, systemMemoryContext);
+                return new LongBatchStreamReader(type, streamDescriptor, systemMemoryContext, options.isResetAllReaders());
             case FLOAT:
                 return new FloatBatchStreamReader(type, streamDescriptor);
             case DOUBLE:
@@ -49,7 +49,7 @@ public final class BatchStreamReaders
             case STRING:
             case VARCHAR:
             case CHAR:
-                return new SliceBatchStreamReader(type, streamDescriptor, systemMemoryContext, options.getMaxSliceSize());
+                return new SliceBatchStreamReader(type, streamDescriptor, systemMemoryContext, options.getMaxSliceSize(), options.isResetAllReaders());
             case TIMESTAMP:
             case TIMESTAMP_MICROSECONDS:
                 boolean enableMicroPrecision = type == TIMESTAMP_MICROSECONDS;
--- a/presto-orc/src/main/java/com/facebook/presto/orc/reader/LongSelectiveStreamReader.java
+++ b/presto-orc/src/main/java/com/facebook/presto/orc/reader/LongSelectiveStreamReader.java
@@ -52,9 +52,10 @@ public class LongSelectiveStreamReader
             Optional<Type> outputType,
             OrcAggregatedMemoryContext systemMemoryContext,
             boolean isLowMemory,
-            long maxSliceSize)
+            long maxSliceSize,
+            boolean resetAllReaders)
     {
-        this.context = new SelectiveReaderContext(streamDescriptor, outputType, filter, systemMemoryContext, isLowMemory, maxSliceSize);
+        this.context = new SelectiveReaderContext(streamDescriptor, outputType, filter, systemMemoryContext, isLowMemory, maxSliceSize, resetAllReaders);
     }
 
     @Override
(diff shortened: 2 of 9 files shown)


presto-orc/src/main/java/com/facebook/presto/orc/reader/BatchStreamReaders.java
```diff
@@ -40,7 +40,7 @@ public static BatchStreamReader createStreamReader(Type type, StreamDescriptor s
             case INT:
             case LONG:
             case DATE:
-                return new LongBatchStreamReader(type, streamDescriptor, systemMemoryContext);
+                return new LongBatchStreamReader(type, streamDescriptor, systemMemoryContext, options.isResetAllReaders());
```
sdruzkin, line 43: Why are you changing batch reader? They are not used in Spark.

presto-orc/src/main/java/com/facebook/presto/orc/reader/LongSelectiveStreamReader.java
```diff
@@ -52,9 +52,10 @@ public LongSelectiveStreamReader(
             Optional<Type> outputType,
             OrcAggregatedMemoryContext systemMemoryContext,
             boolean isLowMemory,
-            long maxSliceSize)
+            long maxSliceSize,
+            boolean resetAllReaders)
```
sdruzkin, line 56: Don't introduce resetAllReaders, it does not make much sense.

presto-orc/src/main/java/com/facebook/presto/orc/reader/LongSelectiveStreamReader.java
```diff
@@ -73,12 +74,20 @@ public void startStripe(Stripe stripe)
                     directReader = new LongDirectSelectiveStreamReader(context);
                 }
                 currentReader = directReader;
+                if (dictionaryReader != null && context.isResetAllReaders()) {
```
sdruzkin, line 77: I think it would be better to close and nullify the other reader if it exists. Reader creation is not very expensive, and usually stripes stick with the same encoding.

sdruzkin:
Please add unit tests checking various combinations of stripe encodings.

presto-orc/src/main/java/com/facebook/presto/orc/reader/LongSelectiveStreamReader.java
chenyangfb: sure. will try that and run some test.

presto-orc/src/main/java/com/facebook/presto/orc/reader/LongSelectiveStreamReader.java
chenyangfb: The main purpose of resetAllReaders for performance comparison in prod.
I plan to remove this option if we decide to enable it by default later.
Let me know if this make sense

presto-orc/src/main/java/com/facebook/presto/orc/reader/BatchStreamReaders.java
chenyangfb: Right, those are not used in spark. but I think not reset all underlying readers is a generic bug, hence including them. Let me know if you prefer to keep the current behavior in this PR.

sdruzkin: Batch reader is rarely used in prod. I'd suggest to not touch it.

presto-orc/src/main/java/com/facebook/presto/orc/reader/LongSelectiveStreamReader.java
sdruzkin: Please remove it now to make it a default behavior.

presto-orc/src/main/java/com/facebook/presto/orc/reader/LongSelectiveStreamReader.java
```diff
@@ -73,12 +74,20 @@ public void startStripe(Stripe stripe)
...
+                }
                 break;
             case DICTIONARY:
                 if (dictionaryReader == null) {
                     dictionaryReader = new LongDictionarySelectiveStreamReader(context);
                 }
                 currentReader = dictionaryReader;
+                if (directReader != null && context.isResetAllReaders()) {
+                    directReader.startStripe(stripe);
+                    System.setProperty("RESET_LONG_READER", "RESET_LONG_READER");
```
sdruzkin, line 89: Don't set system properties in OSS, remove this from all classes. If you need it for testing use a build with local changes.

sdruzkin (commented):
Overall LGTM. Asks are to 1) remove the new flag and make it default behavior; 2) remove setting system properties.

steveburnett:
Thanks for the release note entry! Nits of formatting, so the automation can pick it up for the next release note PR. 

```
== RELEASE NOTES ==

General Changes
* Reduce memory usage in reader with multiple underlying readers by reset all underlying readers.
```

sdruzkin (approved):

PR #24915, opened 2025-04-14
Add shared-arbitrator.memory-pool-abort-capacity-limit into property registry

tanjialiang:
Summary: Adding the memory pool abort capacity limit to presto native property registry and default it to 10GB.

Differential Revision: D73001832


gggrace14 (approved):
LGTM

gggrace14 (approved):

PR #15683, opened 2021-02-04
Support named types in Type and TypeSignature

rongrong:
* Named type's type signature will be formatted as
  qualified_type_name:base_type_signature.
* Introduces TypeWithName, which delegate all execution aspects of the
  type to its base type.
* Make SingatureBinder, type coercion, type constraint, etc to work with
  TypeWithName.

Test plan - (Please fill in how you tested your changes)

```
== NO RELEASE NOTE ==
```

PR #24916, opened 2025-04-15
Add NativeExecutionTypeRewrite

zacw7:
Queries can fail on native worker due to following missing support in Velox:
1. Named types: Presto supports TypeWithName which Velox is not able to parse.
2. EnumType: Velox does not support EnumTypes as well as its companion function ENUM_KEY.

This rewrite addresses the above issues by resolving the type or function in coordinator for native execution:
1. Peel TypeWithName and only preserve the actual base type.
2. Rewrite `CAST(col AS EnumType<T>)` -> `CAST(col AS <T>)` (TODO: preserve the original type information for `typeof`).
3. Since enum can be treated as a map, rewrite `ENUM_KEY(EnumType<T>)` -> `ELEMENT_AT(MAP(<T>, VARCHAR))`.

```
== RELEASE NOTES ==

General Changes
* Add type rewrite support for native execution. This feature can be enabled by ``native-execution-type-rewrite-enabled`` configuration property and ``native_execution_type_rewrite_enabled`` session property. :pr:`24916`

```


arhimondr:
Some earlier discussion: https://github.com/prestodb/presto/pull/15683#pullrequestreview-584806691

There are two main limitations of this approach that I know of:

- For queries like `CREATE TABLE x LIKE y` the custom type names will be erased. For example if table `y` contains a `User` column, the `x` table will get created with a `ROW(firstName, lastName)` column
- Some errors can potentially be more obscure. For example a potential error message on a worker side may not include the custom type name, but instead use the expanded type


The second doesn't seem to be too big of a problem to me. Native execution already has differences in error codes and error messages. And if the feature is not widely used perhaps it is fine to ignore this problem.

For the first one (assuming we care enough) we may consider peeling the custom type information before sending the plan to Prestissimo (Velox) instead. I don't really have a strong opinion here.

CC: @rschlussel

steveburnett:
Please add documentation for these new native configuration and session properties to 
https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/presto_cpp/properties.rst
and 
https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/presto_cpp/properties-session.rst.

rschlussel (dismissed):
you'll need to rebase and push again to get the tests to run/pass.

zacw7:
> @HeidiHan0000 has imported this pull request. If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D73775243).

updated. thanks!

presto-docs/src/main/sphinx/presto_cpp/properties-session.rst
```diff
@@ -97,6 +97,19 @@ If set to ``true``, disables the optimization in expression evaluation to delay
 
 This should only be used for debugging purposes.
 
+``native_execution_type_rewrite_enabled``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``boolean``
+* **Default value:** ``false``
+
+If set to ``true``:
```
steveburnett, line 106: I like something like this more specific way to describe the state than the config property's version "When enabled". See comment in properties.rst with screenshot for more.

presto-docs/src/main/sphinx/presto_cpp/properties.rst
```diff
@@ -57,6 +57,18 @@ alphabetical order.
   This property is required when running Presto C++ workers because of
   underlying differences in behavior from Java workers.
 
+``native-execution-type-rewrite-enabled``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``boolean``
+* **Default value:** ``false``
+
+  When enabled:
```
steveburnett, line 66: I had several suggestions to make for both of these that were mostly the same comment repeated twice, so I decided to write them here and include a screenshot of a local build with my suggested revisions, so you can see the combined suggestions together. 

* Suggest making the two entries - configuration property and session property - the same, as much as possible, for consistency. 
* "When enabled:" could be more specific, like you wrote the session property condition as. 
* It doesn't seem that the items must be sequential, suggest they be an unordered list.
* I revised the text in "Custom type names will be peeled" a little, please make sure it's accurate and correct my suggested text if it's not. 
* Minor suggestions in the two expression examples. 

Screenshot of the configuration property entry with my suggestions: 
Screenshot 2025-04-28 at 2 17 08 PM

steveburnett (changes requested):
Thanks for the doc! I had several small suggestions that it seemed simpler to revise my local doc build and share a screenshot for you to consider. Let me know what you think!

presto-docs/src/main/sphinx/presto_cpp/properties-session.rst
```diff
@@ -97,6 +97,19 @@ If set to ``true``, disables the optimization in expression evaluation to delay
 
 This should only be used for debugging purposes.
 
+``native_execution_type_rewrite_enabled``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``boolean``
+* **Default value:** ``false``
+
+When enabled:
```
steveburnett, line 106:
```suggestion
When set to ``true``:
```
Sorry if I was unclear in my earlier review! I think a more explicit description is better for both the configuration and session property than "When enabled:".

presto-docs/src/main/sphinx/presto_cpp/properties-session.rst
```diff
@@ -97,6 +97,19 @@ If set to ``true``, disables the optimization in expression evaluation to delay
...
+- Custom type names are peeled in the coordinator. Only the actual base type is preserved.
```
steveburnett, line 107: The config property entry looks good in a local doc build, but the session property entry has these formatting problems (see screenshot). Could you make the relevant lines the same in both files? 

Screenshot 2025-04-28 at 2 33 46 PM

steveburnett (changes requested):
Thanks for the quick update! Just a couple of minor issues.

presto-docs/src/main/sphinx/presto_cpp/properties-session.rst
zacw7: Updated. Does the format look good now?

rschlussel (approved):

PR #24922, opened 2025-04-16
[native] Make PeriodicMemoryChecker member of PrestoServer

spershin:
## Description
We well need this in the upcoming changes to be able to fetch the current memory use from PeriodicMemoryChecker to determine if the worker is overloaded in terms of memory.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/PeriodicMemoryChecker.h
+++ b/presto-native-execution/presto_cpp/main/PeriodicMemoryChecker.h
@@ -76,9 +76,14 @@ class PeriodicMemoryChecker {
   /// Stops the 'PeriodicMemoryChecker'.
   void stop();
 
+  /// Returns the last known cached 'current' system memory usage in bytes.
+  int64_t cachedSystemUsedMemoryBytes() const {
+    return cachedSystemUsedMemoryBytes_;
+  }
+
  protected:
-  /// Returns current system memory usage. The returned value is used to compare
-  /// with 'Config::systemMemLimitBytes'.
+  /// Fetches and returns current system memory usage in bytes.
+  /// The returned value is used to compare with 'Config::systemMemLimitBytes'.
   virtual int64_t systemUsedMemoryBytes() = 0;
 
   /// Returns current bytes allocated by malloc. The returned value is used to
@@ -102,6 +107,7 @@ class PeriodicMemoryChecker {
   virtual void pushbackMemory();
 
   const Config config_;
+  std::atomic<int64_t> cachedSystemUsedMemoryBytes_{0};
 
  private:
   // Struct that stores the file names of the heap profiles dumped and the
@@ -137,4 +143,6 @@ class PeriodicMemoryChecker {
       std::greater<DumpFileInfo>>
       dumpFilesByHeapMemUsageMinPq_;
 };
+
+std::unique_ptr<PeriodicMemoryChecker> createMemoryChecker();
 } // namespace facebook::presto
--- a/presto-native-execution/presto_cpp/main/PrestoServer.h
+++ b/presto-native-execution/presto_cpp/main/PrestoServer.h
@@ -67,6 +67,7 @@ class Announcer;
 class SignalHandler;
 class TaskManager;
 class TaskResource;
+class PeriodicMemoryChecker;
 class PeriodicTaskManager;
 class SystemConfig;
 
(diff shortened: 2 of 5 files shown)


presto-native-execution/presto_cpp/main/PeriodicMemoryChecker.h
```diff
@@ -76,6 +76,11 @@ class PeriodicMemoryChecker {
   /// Stops the 'PeriodicMemoryChecker'.
   void stop();
 
+  /// Returns the last known 'current' system memory usage.
+  int64_t currentMemBytes() const {
```
tanjialiang, line 80: uint64_t to align with declared type

presto-native-execution/presto_cpp/main/PrestoServer.h
```diff
@@ -115,16 +116,14 @@ class PrestoServer {
   void enableAnnouncer(bool enable);
 
  protected:
+  virtual void createPeriodicMemoryChecker();
```
tanjialiang, line 119: Why not encapsulate the init and stop in
void startPeriodicMemoryChecker()
void stopPeriodicMemoryChecker()

presto-native-execution/presto_cpp/main/PeriodicMemoryChecker.h
tanjialiang: Why do you need this member and method? systemUsedMemoryBytes() has what's needed?

presto-native-execution/presto_cpp/main/PrestoServer.h
tanjialiang: Maybe we can make this an empty virtual? The implementing PrestoServer will fill it up with the needed checker, like other virtual functions.

presto-native-execution/presto_cpp/main/PeriodicMemoryChecker.h
spershin: > uint64_t to align with declared type

Original systemUsedMemoryBytes() returns int64_t.
I added atomic uint64_t.
I will sync all three types.

spershin: > Why do you need this member and method? systemUsedMemoryBytes() has what's needed?
Because systemUsedMemoryBytes() is actually retrieveing the number from the OS and it is not cheap.
This get method can be called by anyone at any time and simply returns the last stored number.

So we will have (already have):
1. Periodic task that fetches the number and stores it in the atomic int.
2. Any other thread can read from this atomic int.

presto-native-execution/presto_cpp/main/PrestoServer.h
spershin: > Why not encapsulate the init and stop in void startPeriodicMemoryChecker() void stopPeriodicMemoryChecker()

The idea is to have only one overridable method: create().
Then only the base class need to call start() and stop() for the checker if it exists.
No need to have two methods overridable.

Any class inheriting from Server simply needs to override one method of creating the checker.

spershin: > Maybe we can make this an empty virtual? The implementing PrestoServer will fill it up with the needed checker, like other virtual functions.

I was planning that, but the problem is backed support for a define into the project and creation of LinuxMemoryChecker beind the scenes. Check the code.

So can't do what you propose.

presto-native-execution/presto_cpp/main/PeriodicMemoryChecker.h
tanjialiang: I don't see this method used. Is it planned to be used in some followup use case? If not can we add it when it's actually needed?

presto-native-execution/presto_cpp/main/PrestoServer.h
tanjialiang: The macro based singleton checker pick is quite unintuitive. Since now we are storing the checker in server, can we remove the singleton and macro based checker pick. We might need to collaborate with oss community on this so this does not break their deployment.

presto-native-execution/presto_cpp/main/PeriodicMemoryChecker.h
spershin: Yes, it will be used in Meta's fbcode - it creates its own checker, which is a member of the class inherited from the PrestoServer.

presto-native-execution/presto_cpp/main/PrestoServer.h
spershin: > The macro based singleton checker pick is quite unintuitive. Since now we are storing the checker in server, can we remove the singleton (which is removed from this PR) and macro based checker pick. We might need to collaborate with oss community on this so this does not break their deployment.

I don't like macro either.
But changing this is outside of the scope of this diff.
I made the change as less intrusive as possible.
We can follow up removing the define and maybe make it based on config or something else at any time later.

tanjialiang (approved):
Thanks for the iterations. LGTM

PR #24927, opened 2025-04-16
Check Query integrity in Analyzer Util

kevintang2022:
## Description

An ACL check checkQueryIntegrity needs to be moved from DispatchManager to the Analyzer. This PR will pass down the raw query string as well as accessControl down to the Analyzer so that this ACL call can be performed.

## Motivation and Context

The checkQueryIntegrity call needs to be done after queuing to address a potential security issue with the ACL check. If the query uses views, then the view definition could be changed between dispatch and when view definitions are read in the analyzer. Making this change will reduce the time window of this security gap.

## Impact

Changes to createQueryExecution and various other interfaces so that accessControl and raw query string can be passed down to Analyzer and AnalyzerUtil.

## Test Plan

Using debugger in checkAccessPermissions in AnalyzerUtil, the query string and accessControlInfo are passed down properly.

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Improve ACL check by moving checkQueryIntegrity from Dispatch phase to Analyzer phase.

### DIFF
--- a/presto-main-base/src/main/java/com/facebook/presto/sql/analyzer/Analyzer.java
+++ b/presto-main-base/src/main/java/com/facebook/presto/sql/analyzer/Analyzer.java
@@ -57,6 +57,7 @@ public class Analyzer
     private final Map<NodeRef<Parameter>, Expression> parameterLookup;
     private final WarningCollector warningCollector;
     private final MetadataExtractor metadataExtractor;
+    private final String query;
 
     public Analyzer(
             Session session,
@@ -66,9 +67,10 @@ public class Analyzer
             Optional<QueryExplainer> queryExplainer,
             List<Expression> parameters,
             Map<NodeRef<Parameter>, Expression> parameterLookup,
-            WarningCollector warningCollector)
+            WarningCollector warningCollector,
+            String query)
     {
-        this(session, metadata, sqlParser, accessControl, queryExplainer, parameters, parameterLookup, warningCollector, Optional.empty());
+        this(session, metadata, sqlParser, accessControl, queryExplainer, parameters, parameterLookup, warningCollector, Optional.empty(), query);
     }
 
     public Analyzer(
@@ -80,7 +82,8 @@ public class Analyzer
             List<Expression> parameters,
             Map<NodeRef<Parameter>, Expression> parameterLookup,
             WarningCollector warningCollector,
-            Optional<ExecutorService> metadataExtractorExecutor)
+            Optional<ExecutorService> metadataExtractorExecutor,
+            String query)
     {
         this.session = requireNonNull(session, "session is null");
         this.metadata = requireNonNull(metadata, "metadata is null");
@@ -92,6 +95,7 @@ public class Analyzer
         this.warningCollector = requireNonNull(warningCollector, "warningCollector is null");
         requireNonNull(metadataExtractorExecutor, "metadataExtractorExecutor is null");
         this.metadataExtractor = new MetadataExtractor(session, metadata, metadataExtractorExecutor, sqlParser, warningCollector);
+        this.query = requireNonNull(query, "query is null");
     }
 
     public Analysis analyze(Statement statement)
@@ -103,13 +107,13 @@ public class Analyzer
     public Analysis analyze(Statement statement, boolean isDescribe)
     {
         Analysis analysis = analyzeSemantic(statement, isDescribe);
-        checkAccessPermissions(analysis.getAccessControlReferences());
+        checkAccessPermissions(analysis.getAccessControlReferences(), query);
         return analysis;
     }
 
     public Analysis analyzeSemantic(Statement statement, boolean isDescribe)
     {
-        Statement rewrittenStatement = StatementRewrite.rewrite(session, metadata, sqlParser, queryExplainer, statement, parameters, parameterLookup, accessControl, warningCollector);
+        Statement rewrittenStatement = StatementRewrite.rewrite(session, metadata, sqlParser, queryExplainer, statement, parameters, parameterLookup, accessControl, warningCollector, query);
         Analysis analysis = new Analysis(rewrittenStatement, parameterLookup, isDescribe);
 
         metadataExtractor.populateMetadataHandle(session, rewrittenStatement, analysis.getMetadataHandle());
--- a/presto-main-base/src/main/java/com/facebook/presto/util/AnalyzerUtil.java
+++ b/presto-main-base/src/main/java/com/facebook/presto/util/AnalyzerUtil.java
@@ -85,22 +85,37 @@ public class AnalyzerUtil
             MetadataResolver metadataResolver,
             PlanNodeIdAllocator idAllocator,
             VariableAllocator variableAllocator,
-            Session session)
+            Session session,
+            String query)
     {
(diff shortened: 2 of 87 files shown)

### CODE REVIEW

**Comment by steveburnett** [comment]
Thanks for the release note! A nit - release note entries should begin with one of the keywords in the [Order of changes](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines#order-of-changes) in the Release Note Guidelines: 
Fix, Improve, Add, Replace, Rename, Remove, Upgrade, Downgrade, Deprecate, Update. 

Maybe something like this? 

```
== RELEASE NOTES ==

General Changes
* Improve ACL check by moving checkQueryIntegrity from Dispatch phase to Analyzer phase. 
```

File: presto-main-base/src/main/java/com/facebook/presto/util/AnalyzerUtil.java [hunk 1]
```diff
@@ -95,8 +95,9 @@ public static AnalyzerContext getAnalyzerContext(
         return new AnalyzerContext(metadataResolver, idAllocator, variableAllocator);
     }
 
-    public static void checkAccessPermissions(AccessControlReferences accessControlReferences)
+    public static void checkAccessPermissions(AccessControlReferences accessControlReferences, AccessControl accessControl, AccessControlContext accessControlContext, Identity identity, String query)
```
rschlussel, line 98: Should we add the accessControl/accessControlContext/identity for the QueryIntegrity check to the AccessControlReferences instead (and maybe all wrapped up in an AccessControlInfo field which seems basically made for those things)?

presto-main/src/main/java/com/facebook/presto/dispatcher/LocalDispatchQueryFactory.java
```diff
@@ -167,7 +167,7 @@ public DispatchQuery createDispatchQuery(
                 throw new PrestoException(NOT_SUPPORTED, "Unsupported statement type: " + preparedQuery.getStatementClass().getSimpleName());
             }
 
-            return queryExecutionFactory.createQueryExecution(analyzerProviderManager.getAnalyzerProvider(getAnalyzerType(session)), preparedQuery, stateMachine, slug, retryCount, warningCollector, queryType);
+            return queryExecutionFactory.createQueryExecution(analyzerProviderManager.getAnalyzerProvider(getAnalyzerType(session)), preparedQuery, stateMachine, slug, retryCount, warningCollector, queryType, accessControl, query);
```
rschlussel, line 170: we shouldn't need to pass the accessControl from the dispatcher, the analyzer already has the accessControl

presto-main-base/src/main/java/com/facebook/presto/sql/analyzer/Analyzer.java
```diff
@@ -103,7 +104,7 @@ public Analysis analyze(Statement statement)
     public Analysis analyze(Statement statement, boolean isDescribe)
     {
         Analysis analysis = analyzeSemantic(statement, isDescribe);
-        checkAccessPermissions(analysis.getAccessControlReferences());
+        checkAccessPermissions(analysis.getAccessControlReferences(), accessControl, session.getAccessControlContext(), session.getIdentity(), getFormattedSql(statement, sqlParser, Optional.empty()));
```
rschlussel, line 107: one question about getFormattedSql, is whether the query text needs to exactly match what the user passed in (e.g. quotes around identifiers vs. not, line breaks, etc.) because this basically takes the ast and recreates sql from it, so it's standardized.

rschlussel: also, does this mean you aren't actually using the query string that's getting passed around everywhere? because in that case you can remove all those changes.

presto-main/src/main/java/com/facebook/presto/dispatcher/LocalDispatchQueryFactory.java
kevintang2022: The call actually is made in AnalyzerUtil, not Analyzer

AnalyzerUtil does not have accessControl

```
    public static void checkAccessPermissions(AccessControlReferences accessControlReferences, AccessControl accessControl, AccessControlContext accessControlContext, Identity identity, String query)
    {
        accessControl.checkQueryIntegrity(identity, accessControlContext, query);
        // Table checks
        checkAccessPermissionsForTable(accessControlReferences);
        // Table Column checks
        checkAccessPermissionsForColumns(accessControlReferences);
    }

```

presto-main-base/src/main/java/com/facebook/presto/util/AnalyzerUtil.java
kevintang2022: I'm not sure if this works because AccessControlReferences contains a Map of column and table references with AccessControlInfo for each. We wouldn't want to call checkQueryIntegrity for each table and column. 

Let me know if you think there's a way to access Identity and AccessControl using AccessControlReferences. Because it's possible that the Maps could be empty, so it does not seem like it would work in all cases

rschlussel: I meant add another field to accessControlReferences that contains the AccessControlInfo for the query integrity check.

presto-main/src/main/java/com/facebook/presto/dispatcher/LocalDispatchQueryFactory.java
rschlussel: but it gets called after analysis so we acn pass the info from the analyzer to checkAccessPermissions (e.g. via accessControlReferences as suggested above).

presto-main-base/src/main/java/com/facebook/presto/sql/analyzer/Analyzer.java
kevintang2022: In order to pass down raw query string, I would need to make a much larger refactor of a bunch of classes

I probably need to combine my changes with this PR too https://github.com/prestodb/presto/pull/24924

kevintang2022:
Another general question: should I put this change (moving the call from dispatch to analyzer) under a config flag?

rschlussel:
> Another general question: should I put this change (moving the call from dispatch to analyzer) under a config flag?

no, it doesn't need a config flag. it was only recently reintroduced.

presto-spi/src/main/java/com/facebook/presto/spi/analyzer/AccessControlReferences.java
```diff
@@ -28,9 +28,11 @@ public class AccessControlReferences
 {
     private final Map<AccessControlRole, Set<AccessControlInfoForTable>> tableReferences;
     private final Map<AccessControlInfo, Map<QualifiedObjectName, Set<Subfield>>> tableColumnAndSubfieldReferencesForAccessControl;
+    private final AccessControlInfo queryAccessControlInfo;
```
rschlussel, line 31: maybe make this non-final/settable. This class is already not threadsafe, the maps all get populated after this is constructed.  The advantage is then we get set this field in StatementAnalyzer when we call analyze(), instead of having to pass it through to analysis.  That should make the scope of changes much smaller (wouldn't need to make changes for everything that constructs an analysis object)

rschlussel (approved):

PR #24928, opened 2025-04-16
Fix TDigest::scale() to update sum member.

spershin:
## Description
TDigest::sum member variable is not really used in any method that could affect result of TDigest functions.
However, for complete correctness we should update it.

```
== NO RELEASE NOTE ==
```


skyelves (approved):
LGTM

amitkdutta (approved):
Thanks @spershin

PR #24929, opened 2025-04-16
[native] Fix blank HTTP Host header in requests

ZacBlanco:
## Description

This change properly sets the request destination address so that ensureHostHeader sets the correct value for the HTTP `Host` header.

## Motivation and Context

Because the request's destination address wasn't set inside of the RequestBuilder object, the Host was populated with a blank value. On older HTTP servers this is permissible, but on newer ones (such as the incoming Jetty 12 upgrade), this is a hard error and results in the request being rejected with a 400 error.

## Impact

N/A

## Test Plan

Existing tests

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/http/HttpClient.cpp
+++ b/presto-native-execution/presto_cpp/main/http/HttpClient.cpp
@@ -522,9 +522,11 @@ void HttpClient::sendRequest(std::shared_ptr<ResponseHandler> responseHandler) {
 }
 
 folly::SemiFuture<std::unique_ptr<HttpResponse>> HttpClient::sendRequest(
-    const proxygen::HTTPMessage& request,
+    proxygen::HTTPMessage& request,
     const std::string& body,
     int64_t delayMs) {
+  request.setDstAddress(this->address_);
+  request.ensureHostHeader();
   auto responseHandler = std::make_shared<ResponseHandler>(
       request,
       maxResponseAllocBytes_,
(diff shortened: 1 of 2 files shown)


presto-native-execution/presto_cpp/main/http/HttpClient.cpp
```diff
@@ -522,17 +522,19 @@ void HttpClient::sendRequest(std::shared_ptr<ResponseHandler> responseHandler) {
 }
 
 folly::SemiFuture<std::unique_ptr<HttpResponse>> HttpClient::sendRequest(
-    const proxygen::HTTPMessage& request,
+    proxygen::HTTPMessage& request,
     const std::string& body,
     int64_t delayMs) {
+  request.setDstAddress(this->address_);
+  request.ensureHostHeader();
+  VLOG(1) << "HTTP CLIENT REQUEST HEADERS" << request;
```
pramodsatya, line 530: nit: this log can be removed

pramodsatya (dismissed):
Thanks @ZacBlanco, LGTM.

presto-native-execution/presto_cpp/main/http/HttpClient.cpp
ZacBlanco: Thanks, removed

presto-native-execution/presto_cpp/main/http/HttpClient.h
```diff
@@ -244,7 +244,6 @@ class RequestBuilder {
   send(HttpClient* client, const std::string& body = "", int64_t delayMs = 0) {
     addJwtIfConfigured();
     header(proxygen::HTTP_HEADER_CONTENT_LENGTH, std::to_string(body.size()));
-    headers_.ensureHostHeader();
```
BryanCutler, line 247: I think `send` is used other places too, can this be left in?

ZacBlanco: The problem is this will _always_ set the `Host` header as an empty string because the HttpMessage doesn't have a dstAddress set. It isn't possible to get the dstAddress except for when the client actually sends the request

aditi-pandit (approved):
Thanks @ZacBlanco

PR #24949, opened 2025-04-21
[native] Track CPU & Memory overload in native worker.

spershin:
## Description
We introduce new set of config properties to setup cpu and memory use thresholds above which we will consider worker overloaded.
Default values are zero, meaning we are not tracking it.

## Motivation and Context
We want to spin up a system where Presto is aware of the workers' load and acts accordingly.

## Impact
The change is harmless - only tracking the overload (optionally).

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-docs/src/main/sphinx/presto_cpp/properties.rst
+++ b/presto-docs/src/main/sphinx/presto_cpp/properties.rst
@@ -578,6 +578,33 @@ The default value of 60 gb is calculated based on available machine memory of 64
 Specifies the amount of memory to shrink when the memory pushback is
 triggered. This only applies if ``system-mem-pushback-enabled`` is ``true``.
 
+``system-mem-pushback-abort-enabled``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``boolean``
+* **Default value:** ``false``
+
+If true, memory pushback will abort queries with the largest memory usage under
+low memory condition. This only applies if ``system-mem-pushback-enabled`` is ``true``.
+
+``worker-overloaded-threshold-mem-gb``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``integer``
+* **Default value:** ``0``
+
+Memory threshold in GB above which the worker is considered overloaded in terms of
+memory use. Ignored if zero.
+
+``worker-overloaded-threshold-cpu-pct``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``integer``
+* **Default value:** ``0``
+
+CPU threshold in % above which the worker is considered overloaded in terms of
+CPU use. Ignored if zero.
+
 Environment Variables As Values For Worker Properties
 -----------------------------------------------------
 
--- a/presto-native-execution/presto_cpp/main/PrestoServer.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoServer.cpp
@@ -1426,7 +1426,54 @@ void PrestoServer::populateMemAndCPUInfo() {
   });
   RECORD_METRIC_VALUE(kCounterNumQueryContexts, numContexts);
   cpuMon_.update();
+  checkOverload();
   **memoryInfo_.wlock() = std::move(memoryInfo);
+}
+
+void PrestoServer::checkOverload() {
+  auto systemConfig = SystemConfig::instance();
+
+  const auto overloadedThresholdMemBytes =
+      systemConfig->workerOverloadedThresholdMemGb() * 1024 * 1024 * 1024;
+  if (overloadedThresholdMemBytes > 0) {
+    const auto currentUsedMemoryBytes = (memoryChecker_ != nullptr)
+        ? memoryChecker_->cachedSystemUsedMemoryBytes()
+        : 0;
+    const bool isMemOverloaded =
+        (currentUsedMemoryBytes > overloadedThresholdMemBytes);
+    if (isMemOverloaded) {
+      LOG(WARNING) << "Server memory is overloaded. Currently used: "
+                   << velox::succinctBytes(currentUsedMemoryBytes)
+                   << ", threshold: "
+                   << velox::succinctBytes(overloadedThresholdMemBytes);
+    } else if (isMemOverloaded_) {
+      LOG(INFO) << "Server memory is no longer overloaded. Currently used: "
+                << velox::succinctBytes(currentUsedMemoryBytes)
+                << ", threshold: "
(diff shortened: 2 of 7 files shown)


presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -1426,9 +1426,56 @@ void PrestoServer::populateMemAndCPUInfo() {
...
+                   << velox::succinctBytes(currentUsedMemoryBytes)
+                   << ", threshold: "
+                   << velox::succinctBytes(overloadedThresholdMemBytes);
+    } else if (isMemOverloaded_) {
+      LOG(INFO) << "Server memory is no longer overloaded. Currently used: "
+                << velox::succinctBytes(currentUsedMemoryBytes)
+                << ", threshold: "
+                << velox::succinctBytes(overloadedThresholdMemBytes);
+    }
+    RECORD_METRIC_VALUE(kCounterOverloadedMem, isMemOverloaded ? 100 : 0);
```
amitkdutta, line 1455: Curious why is 0 and 100, but not a range of values in between. or why not just 0 and 1.

spershin: @amitkdutta 
Because when the counters backend starts averaging the signal over some time period, say 60 seconds the 1 will turn into 0 and we won't see any signal.
100 however won't have such issue and will be easy to reason about how much % of the period the system was overloaded if the number is below 100.

arhimondr (dismissed):

presto-native-execution/presto_cpp/main/PrestoServer.cpp
```diff
@@ -1426,9 +1426,56 @@ void PrestoServer::populateMemAndCPUInfo() {
   });
   RECORD_METRIC_VALUE(kCounterNumQueryContexts, numContexts);
   cpuMon_.update();
+  checkOverload();
   **memoryInfo_.wlock() = std::move(memoryInfo);
 }
 
+void PrestoServer::checkOverload() {
```
tanjialiang, line 1433: Maybe offloading all these in a separate class?

spershin: @tanjialiang 
I would rather keep this logic here.
Next we will need Server to tell TaskManager that we are overloaded or no longer, so it can act on the Task queue.
Doing this in some another side class seems wrong.
Spinning up the class to just decide if we are overloaded based on a couple of metrics seems too much.

If that somehow grows over time, then we can refactor.

aditi-pandit (dismissed):

presto-docs/src/main/sphinx/presto_cpp/properties.rst
```diff
@@ -578,6 +578,33 @@ The default value of 60 gb is calculated based on available machine memory of 64
 Specifies the amount of memory to shrink when the memory pushback is
 triggered. This only applies if ``system-mem-pushback-enabled`` is ``true``.
 
+``system-mem-pushback-abort-enabled``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``boolean``
+* **Default value:** ``false``
+
+If true, memory pushback will quickly abort queries with the most memory usage under
```
steveburnett, line 587:
```suggestion
If true, memory pushback will abort queries with the largest memory usage under
```
Suggest deleting "quickly" because, quickly in comparison to what? 
Suggest changing "most" to "largest".

steveburnett (changes requested):
Thanks for the doc! Just two very minor suggestions, let me know if they don't work for you.

presto-docs/src/main/sphinx/presto_cpp/properties.rst
spershin: @steveburnett 
No problem, I can update. Copied the text from the source code. Will update it in the source code as well.

presto-docs/src/main/sphinx/presto_cpp/properties.rst
```diff
@@ -578,6 +578,33 @@ The default value of 60 gb is calculated based on available machine memory of 64
...
+* **Default value:** ``false``
+
+If true, memory pushback will abort queries with the largest memory usage under
+low memory condition. This only applies if ``system-mem-pushback-enabled`` is ``true``.
+
+``worker-overloaded-threshold-mem-gb``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``integer``
+* **Default value:** ``0``
```
aditi-pandit, line 594: It seems a bit odd to use 0 for default threshold. Can we use something like -1 to indicate that it can be ignored ?

aditi-pandit (commented):
Thanks @spershin for this change. Had one question:

presto-docs/src/main/sphinx/presto_cpp/properties.rst
spershin: @aditi-pandit 
Thanks for reviewing this PR, seem like hard to find anyone brave enough these days to review a small change.

Using zeros for 'no op' is the practice we use already in the config.
See lines:
252 for kDriverCancelTasksWithStuckOperatorsThresholdMs
296 for kSystemMemLimitGb
365 for kAsyncCachePersistenceInterval
451 for kSharedArbitratorMaxMemoryArbitrationTime
532 for kSharedArbitratorGlobalArbitrationMemoryReclaimPct

It actually makes a certain semantic sense - threshold 0 for something that cannot be negative does not make sense, so it can be used as the 'magic number'.

hitarth (approved):
Looks good. Looking forward for passing this signal to TaskManager for better task queue management.

spershin:
> Looks good. Looking forward for passing this signal to TaskManager for better task queue management.

@hitarth 
Yes, we can certainly work on this in parallel. It might need a new endpoint, but maybe we can use existing ones already.
The first idea I would like to try out is to have protection mechanism in the worker itself.

The two can coexist together.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
tanjialiang: We are adding more and more stuff to PrestoServer now. It will be easier to read and maintain if we move these logic to a separate class maybe called SystemMonitor. We can start from this one and maybe move a couple of stuff from current PrestoServer there, as well.

presto-native-execution/presto_cpp/main/PrestoServer.cpp
tanjialiang: We are adding more and more stuff to PrestoServer now. It will be easier to read and maintain if we move these logic to a separate class maybe called SystemMonitor. We can start from this one and maybe move a couple of stuff from current PrestoServer there, as well.

PR #24960, opened 2025-04-23
Fix commit after statement fail in non-autocommit transaction

hantangwangd:
## Description

Following PR #23247 , this PR handle the session corruption cause by a failed statement in non-autocommit transaction as well. 

When executing `commit` in above scenario, it should fail while aborting the whole corrupted transaction successfully. Thus we can keep running following statements within the same session.

## Motivation and Context

Fix: #23246 

## Test Plan

 - Newly added test cases in `AbstractTestDistributedQueries` to simulate the scenarios described in #23246
 - Manually confirm the fix in CLI console

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
General Changes
* Ensure `COMMIT` statement aborts the non-autocommit transaction which was corrupted by failed statements
```


presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestDistributedQueries.java
```diff
@@ -280,6 +280,40 @@ public void testNonAutoCommitTransactionWithFailAndRollback()
...
+                            assertQueryFails(txnSession, "SELECT fail('forced failure')", "forced failure");
+
+                            // cannot execute any SQLs except `rollback` in current session
+                            assertQueryFails(txnSession, "select count(*) from test_non_autocommit_table_v2", "Current transaction is aborted, commands ignored until end of transaction block");
+                            assertQueryFails(txnSession, "show tables", "Current transaction is aborted, commands ignored until end of transaction block");
+                            assertQueryFails(txnSession, "insert into test_non_autocommit_table_v2 values(1, '1001')", "Current transaction is aborted, commands ignored until end of transaction block");
+                            assertQueryFails(txnSession, "create table test_table(a int, b varchar)", "Current transaction is aborted, commands ignored until end of transaction block");
+
+                            // `commit` statement fail, but abort the transaction successfully
+                            assertQueryFails(txnSession, "commit", "Current transaction has already been aborted");
```
ZacBlanco, line 310: Can we still call `rollback` here before or after this query? Can we add that to verify what the comment above states?

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestDistributedQueries.java
```diff
@@ -280,6 +280,40 @@ public void testNonAutoCommitTransactionWithFailAndRollback()
...
+                        });
+
+        assertQuery("select count(*) from test_non_autocommit_table_v2", "values(0)");
```
ZacBlanco, line 313: Just so I understand what this PR fixes: Before this change, would this query have failed due to the previous transaction failure?

ZacBlanco (commented):
Changes look good. I have a question about the tests

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestDistributedQueries.java
hantangwangd: After this statement, the transaction is completely aborted. So, if we calling `ROLLBACK` again, the behavior should be the same as when no transaction was started at all, and an error message will be reported as follows:

```
Query xxx failed: No transaction in progress
```

In CLI console, it indeed behavior as above. However, currently in the testing framework, since the flag `clearTransactionId` cannot be returned to the client's session, executing rollback again will report the following error message:

```
Expected exception message 'Unknown transaction ID: cddea4f9-4dd9-437f-bab5-0cc17c76dee5. Possibly expired? ...
```

Now I have added all these scenarios to the test case. I thinks this failure message can demonstrate that the transaction has been aborted by the server.

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestDistributedQueries.java
hantangwangd: Yes, before this change, the query would always fail after a transaction is corrupted by previous failure. However, the test framework can't accurately mimic the real world behavior, so in the test case `transaction(...).execute(...)` itself will finally fail because of the corrupted transaction if we didn't abort it properly.

I have refactored the test case and added some comments to clarify the expected behaviors of subsequent statements. Please take a look when available, thanks.

Issue #24964, opened 2025-04-23
Productionize native-sidecar feature


amitkdutta:
We are working to run native-side feature with worker as sidecar option. This can be enabled when [native-sidecar plugin](https://github.com/prestodb/presto/tree/master/presto-native-sidecar-plugin) is in plugins directory and following two configs need to be set:

1. In coordinator config properties: `"coordinator-sidecar-enabled": "true"`
2. In worker config properties: `"native-sidecar": "true"`

While deploying it in Meta internal systems, we are finding improvement opportunities which will be listed here. To start with we are able to build and package the plugin. Now we saw 2 issues.

1. coordinator always loads the plugin even if `"coordinator-sidecar-enabled": "false"`, this makes coordinator crash
```
025-04-23T12:55:49.520-0700	ERROR	main	com.facebook.presto.server.PrestoServer	System session property provider is already registered for property provider : native-worker
java.lang.IllegalArgumentException: System session property provider is already registered for property provider : native-worker
	at com.facebook.presto.metadata.SessionPropertyManager.loadSessionPropertyProvider(SessionPropertyManager.java:147)
	at com.facebook.presto.metadata.SessionPropertyManager.loadSessionPropertyProviders(SessionPropertyManager.java:154)
	at com.facebook.presto.server.PrestoServer.run(PrestoServer.java:193)
```
Since this plugin will always be included in the build, we need to conditionally load it given native worker may not use this plugin at runtime, or we disable it with config for java workers at runtime. We won't be able to control the build scenario as build will unconditionally include the plugins.
2. When the plugin runs, we failed to run any query when coordinator and worker are in https mode.
```
com.facebook.presto.spi.PrestoException: Failed to get session properties from sidecar.
	at com.facebook.presto.sidecar.sessionpropertyproviders.NativeSystemSessionPropertyProvider.fetchSessionProperties(NativeSystemSessionPropertyProvider.java:105)
	at com.google.common.base.Suppliers$ExpiringMemoizingSupplier.get(Suppliers.java:261)
	at com.facebook.presto.sidecar.sessionpropertyproviders.NativeSystemSessionPropertyProvider.getSessionProperties(NativeSystemSessionPropertyProvider.java:89)
	at com.facebook.presto.metadata.SessionPropertyManager.lambda$getWorkerSessionProperties$0(SessionPropertyManager.java:216)
	at java.base/java.util.stream.ReferencePipeline$7$1.accept(ReferencePipeline.java:271)
[...]

PR #24965, opened 2025-04-23
[sidecar] Prevent coordinator crashing in non-sidecar enabled clusters by making sidecar plugin functionalities config driven

pdabre12:
## Description
Prevent coordinator crashing in non-sidecar enabled clusters by making sidecar plugin functionalities config driven

## Motivation and Context
Solves part of #24964 

## Impact
Users would be able to include the ` presto-native-sidecar-plugin` plugin in the build even for non-sidecar enabled clusters.

## Test Plan
Unit tests

## Contributor checklist

- [x] CI passed.

## Release Notes
```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-main-base/src/main/java/com/facebook/presto/metadata/FunctionAndTypeManager.java
+++ b/presto-main-base/src/main/java/com/facebook/presto/metadata/FunctionAndTypeManager.java
@@ -418,13 +418,6 @@ public class FunctionAndTypeManager
         servingTypeManagerParametricTypesSupplier.set(this::getServingTypeManagerParametricTypes);
     }
 
-    public void loadTypeManagers()
-    {
-        for (String typeManagerName : typeManagerFactories.keySet()) {
-            loadTypeManager(typeManagerName);
-        }
-    }
-
     public void addTypeManagerFactory(TypeManagerFactory factory)
     {
         if (typeManagerFactories.putIfAbsent(factory.getName(), factory) != null) {
--- a/presto-main-base/src/main/java/com/facebook/presto/metadata/SessionPropertyManager.java
+++ b/presto-main-base/src/main/java/com/facebook/presto/metadata/SessionPropertyManager.java
@@ -15,6 +15,7 @@ package com.facebook.presto.metadata;
 
 import com.facebook.airlift.json.JsonCodec;
 import com.facebook.airlift.json.JsonCodecFactory;
+import com.facebook.airlift.log.Logger;
 import com.facebook.presto.Session;
 import com.facebook.presto.SystemSessionProperties;
 import com.facebook.presto.common.block.BlockBuilder;
@@ -52,6 +53,8 @@ import com.google.common.collect.Maps;
 import javax.annotation.Nullable;
 import javax.inject.Inject;
 
+import java.io.File;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -59,14 +62,18 @@ import java.util.Optional;
 import java.util.TreeMap;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Supplier;
 
 import static com.facebook.presto.common.type.TypeUtils.writeNativeValue;
 import static com.facebook.presto.spi.StandardErrorCode.INVALID_SESSION_PROPERTY;
 import static com.facebook.presto.sql.planner.ExpressionInterpreter.evaluateConstantExpression;
+import static com.facebook.presto.util.PropertiesUtil.loadProperties;
 import static com.google.common.base.MoreObjects.firstNonNull;
 import static com.google.common.base.Preconditions.checkState;
+import static com.google.common.base.Strings.isNullOrEmpty;
 import static com.google.common.collect.ImmutableList.toImmutableList;
+import static com.google.common.io.Files.getNameWithoutExtension;
 import static java.lang.String.format;
 import static java.util.Objects.requireNonNull;
 import static java.util.concurrent.TimeUnit.HOURS;
@@ -74,6 +81,9 @@ import static java.util.concurrent.TimeUnit.HOURS;
 public final class SessionPropertyManager
 {
     private static final JsonCodecFactory JSON_CODEC_FACTORY = new JsonCodecFactory();
+    private static final Logger log = Logger.get(SessionPropertyManager.class);
+    private static final String SESSION_PROPERTY_PROVIDER_NAME = "session-property-provider.name";
+
     private final ConcurrentMap<String, PropertyMetadata<?>> systemSessionProperties = new ConcurrentHashMap<>();
     private final ConcurrentMap<ConnectorId, Map<String, PropertyMetadata<?>>> connectorSessionProperties = new ConcurrentHashMap<>();
     private final Map<String, WorkerSessionPropertyProvider> workerSessionPropertyProviders;
@@ -81,28 +91,33 @@ public final class SessionPropertyManager
     private final Supplier<Map<String, PropertyMetadata<?>>> memoizedWorkerSessionProperties;
     private final Optional<NodeManager> nodeManager;
     private final Optional<TypeManager> functionAndTypeManager;
+    private final File configDir;
+    private final AtomicBoolean sessionPropertyProvidersLoading = new AtomicBoolean();
 
     @Inject
     public SessionPropertyManager(
             SystemSessionProperties systemSessionProperties,
             Map<String, WorkerSessionPropertyProvider> workerSessionPropertyProviders,
             FunctionAndTypeManager functionAndTypeManager,
-            NodeManager nodeManager)
+            NodeManager nodeManager,
+            SessionPropertyProviderConfig config)
     {
-        this(systemSessionProperties.getSessionProperties(), workerSessionPropertyProviders, Optional.ofNullable(functionAndTypeManager), Optional.ofNullable(nodeManager));
+        this(systemSessionProperties.getSessionProperties(), workerSessionPropertyProviders, Optional.ofNullable(functionAndTypeManager), Optional.ofNullable(nodeManager), config);
     }
 
     public SessionPropertyManager(
             List<PropertyMetadata<?>> sessionProperties,
             Map<String, WorkerSessionPropertyProvider> workerSessionPropertyProviders,
             Optional<TypeManager> functionAndTypeManager,
-            Optional<NodeManager> nodeManager)
+            Optional<NodeManager> nodeManager,
+            SessionPropertyProviderConfig config)
     {
         this.nodeManager = requireNonNull(nodeManager, "nodeManager is null");
         this.functionAndTypeManager = requireNonNull(functionAndTypeManager, "functionAndTypeManager is null");
         this.memoizedWorkerSessionProperties = Suppliers.memoizeWithExpiration(this::getWorkerSessionProperties,
                 1, HOURS);
         this.workerSessionPropertyProviders = new ConcurrentHashMap<>(workerSessionPropertyProviders);
+        this.configDir = requireNonNull(config, "config is null").getSessionPropertyProvidersConfigurationDir();
         addSystemSessionProperties(sessionProperties);
     }
 
@@ -135,24 +150,42 @@ public final class SessionPropertyManager
                                 javaFeaturesConfig,
                                 nodeSpillConfig)),
                 Optional.empty(),
-                Optional.empty());
+                Optional.empty(),
+                new SessionPropertyProviderConfig());
     }
 
-    public void loadSessionPropertyProvider(String sessionPropertyProviderName, Optional<TypeManager> typeManager, Optional<NodeManager> nodeManager)
+    public void loadSessionPropertyProviders()
+            throws Exception
     {
+        if (!sessionPropertyProvidersLoading.compareAndSet(false, true)) {
+            return;
+        }
+
+        for (File file : listFiles(configDir)) {
+            if (file.isFile() && file.getName().endsWith(".properties")) {
+                String sessionPropertyProviderName = getNameWithoutExtension(file.getName());
+                Map<String, String> properties = loadProperties(file);
+                checkState(!isNullOrEmpty(properties.get(SESSION_PROPERTY_PROVIDER_NAME)),
+                        "Session property manager configuration %s does not contain %s",
+                        file.getAbsoluteFile(),
+                        SESSION_PROPERTY_PROVIDER_NAME);
+                properties = new HashMap<>(properties);
+                properties.remove(SESSION_PROPERTY_PROVIDER_NAME);
+                loadSessionPropertyProvider(sessionPropertyProviderName, properties, functionAndTypeManager, nodeManager);
+            }
+        }
+    }
+
+    public void loadSessionPropertyProvider(String sessionPropertyProviderName, Map<String, String> properties, Optional<TypeManager> typeManager, Optional<NodeManager> nodeManager)
+    {
+        log.info("-- Loading %s session property provider --", sessionPropertyProviderName);
         WorkerSessionPropertyProviderFactory factory = workerSessionPropertyProviderFactories.get(sessionPropertyProviderName);
         checkState(factory != null, "No factory for session property provider : " + sessionPropertyProviderName);
         WorkerSessionPropertyProvider sessionPropertyProvider = factory.create(new SessionPropertyContext(typeManager, nodeManager));
         if (workerSessionPropertyProviders.putIfAbsent(sessionPropertyProviderName, sessionPropertyProvider) != null) {
             throw new IllegalArgumentException("System session property provider is already registered for property provider : " + sessionPropertyProviderName);
         }
-    }
-
-    public void loadSessionPropertyProviders()
-    {
-        for (String sessionPropertyProviderName : workerSessionPropertyProviderFactories.keySet()) {
-            loadSessionPropertyProvider(sessionPropertyProviderName, functionAndTypeManager, nodeManager);
-        }
+        log.info("-- Added session property provider [%s] --", sessionPropertyProviderName);
     }
 
     public void addSessionPropertyProviderFactory(WorkerSessionPropertyProviderFactory factory)
@@ -222,6 +255,17 @@ public final class SessionPropertyManager
             workerSessionProperties.put(sessionProperty.getName(), sessionProperty);
         });
         return workerSessionProperties;
+    }
+
+    private static List<File> listFiles(File dir)
+    {
+        if (dir != null && dir.isDirectory()) {
+            File[] files = dir.listFiles();
+            if (files != null) {
+                return ImmutableList.copyOf(files);
+            }
+        }
+        return ImmutableList.of();
     }
 
     public List<SessionPropertyValue> getAllSessionProperties(Session session, Map<String, ConnectorId> catalogs)
--- a/presto-native-sidecar-plugin/src/test/java/com/facebook/presto/sidecar/TestNativeSidecarPlugin.java
+++ b/presto-native-sidecar-plugin/src/test/java/com/facebook/presto/sidecar/TestNativeSidecarPlugin.java
@@ -40,7 +40,6 @@ import static org.testng.Assert.assertEquals;
 import static org.testng.Assert.assertFalse;
 import static org.testng.Assert.fail;
 
-@Test(singleThreaded = true)
 public class TestNativeSidecarPlugin
         extends AbstractTestQueryFramework
 {
--- a/presto-main-base/src/main/java/com/facebook/presto/metadata/SessionPropertyProviderConfig.java
+++ b/presto-main-base/src/main/java/com/facebook/presto/metadata/SessionPropertyProviderConfig.java
@@ -1,0 +1,38 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.metadata;
+
+import com.facebook.airlift.configuration.Config;
+
+import javax.validation.constraints.NotNull;
+
+import java.io.File;
+
+public class SessionPropertyProviderConfig
+{
+    private File sessionPropertyProvidersConfigurationDir = new File("etc/session-property-providers/");
+
+    @NotNull
+    public File getSessionPropertyProvidersConfigurationDir()
+    {
+        return sessionPropertyProvidersConfigurationDir;
+    }
+
+    @Config("session-property-provider.config-dir")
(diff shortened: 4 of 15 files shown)


pdabre12:
Logs from running PrestoServer locally:

```
2025-04-23T15:51:11.214-0700	INFO	main	com.facebook.presto.server.PluginManager	Registering connector hudi
2025-04-23T15:51:11.216-0700	INFO	main	com.facebook.presto.server.PluginManager	-- Finished loading plugin ../presto-hudi/pom.xml --
2025-04-23T15:51:11.216-0700	INFO	main	com.facebook.presto.server.PluginManager	-- Skipping loading disabled plugin ../presto-native-sidecar-plugin/pom.xml --
2025-04-23T15:51:11.227-0700	INFO	main	com.facebook.presto.metadata.StaticCatalogStore	-- Loading catalog properties etc/catalog/blackhole.properties --
```

rschlussel:
The general model for all plugins has been that you should be able to load plugins without having any side effects if they aren't enabled. Also, as far as cluster configuration, it's inconvenient to have to disable the plugin in addition to setting use side car to false.  instead, i think it would be better for the NativeSessionpropertyProviders to have different names (currently the coordinator based one and the sidecar based one are both called native-worker), and loadSystemSessionProperty providers only loads the providers that are relevant for the configuration (similar to how AccessControlManager.loadSystemAccessControl(), but with a different/slightly more complicated combination of configs.

pdabre12:
@rschlussel Thanks for the response.
I think it might be much simpler to have annotations on the sidecar specific functionalities as shown in this PR.
We can mark the other functionalities aka. NativeTypeManager, NativeFunctionNamespaceManager as dependent on `coordinator-sidecar-enabled` as well.
From a user's POV, all they would need to do even if the plugin was loaded is set `coordinator-sidecar-enabled` to false, to not use the sidecar functionalities. What do you think?

rschlussel (approved):
This seems like a fine option

1. is it possible to add a test that everything starts up properly now if the plugin is loaded but the config is not enabled?  
2. For some sidecar features it might be nice to have an additional flag that enables them (e.g. you want to disable sidecar constant folding at first, but still use the function registry).  Would this kind of approach still work if something needs two configs to be enabled?

pdabre12:
@rschlussel 
Added the unit test.
The logic is to install the `NativeSidecarPlugin` on both the java and c++ clusters but not set the coordinator-sidecar-enabled(default false). If we are still able to run queries , means both the clusters were up without any issues. Will this work?

> For some sidecar features it might be nice to have an additional flag that enables them (e.g. you want to disable sidecar constant folding at first, but still use the function registry). Would this kind of approach still work if something needs two configs to be enabled?

For supporting multiple configs, added a new annotation which handles that, still could not find out how to unit test this though. Maybe we can let this change pass, create an issue and handle the unit test at a later time?

pdabre12:
We only need the annotation for `NativeTypeManager` and the `NativeSessionPropertyProvider`, the other functionalities aka `NativeFunctionNamespaceManager` (using the StaticFunctionNamespaceStore) and `NativePlanChecker` are loaded through a configuration file and if these configuration files aren't present, we will not load them.

presto-common/src/main/java/com/facebook/presto/common/ConditionalOnConfigs.java
```diff
@@ -0,0 +1,26 @@
...
+package com.facebook.presto.common;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface ConditionalOnConfigs
```
ZacBlanco, line 23: Why do we need this variant?

presto-main-base/src/main/java/com/facebook/presto/metadata/FunctionAndTypeManager.java
```diff
@@ -420,8 +427,11 @@ public void loadTypeManager(String typeManagerName)
 
     public void loadTypeManagers()
     {
-        for (String typeManagerName : typeManagerFactories.keySet()) {
-            loadTypeManager(typeManagerName);
+        Map<String, String> configMap = ImmutableMap.of("coordinator-sidecar-enabled", String.valueOf(serverConfig.isCoordinatorSidecarEnabled()));
```
ZacBlanco, line 430: Do we have a static constant anywhere that we can reference rather than repeating this string?

presto-main-base/src/main/java/com/facebook/presto/metadata/SessionPropertyManager.java
```diff
@@ -150,8 +157,13 @@ public void loadSessionPropertyProvider(String sessionPropertyProviderName, Opti
 
     public void loadSessionPropertyProviders()
     {
-        for (String sessionPropertyProviderName : workerSessionPropertyProviderFactories.keySet()) {
-            loadSessionPropertyProvider(sessionPropertyProviderName, functionAndTypeManager, nodeManager);
+        Map<String, String> configMap = ImmutableMap.of("coordinator-sidecar-enabled", String.valueOf(serverConfig.isCoordinatorSidecarEnabled()));
```
ZacBlanco, line 160: I'm wondering if there's a better way to load the map of properties so that this code is more generic. It seems tailored specifically for the sidecar. If any other plugin needed to make use of ConditionalOnConfig, the property would need to get added here which is not ideal.

ZacBlanco (changes requested):

presto-main-base/src/main/java/com/facebook/presto/metadata/FunctionAndTypeManager.java
pdabre12: No, but can make one.

presto-common/src/main/java/com/facebook/presto/common/ConditionalOnConfigs.java
pdabre12: It handles a list of `@ConditionalOnConfig`.
The idea is to have the ability to have additional config flags on the sidecar functionalities to conditionally load individual functionalities.
As in let's say on the `NativeFunctionNamespaceManager`, we have two `@ConditionalOnConfig` configs `coordinator-sidecar-enabled` and `load-function-registry`.

presto-main-base/src/main/java/com/facebook/presto/metadata/SessionPropertyManager.java
pdabre12: It seems like the best option.
We could have the factories provide the required properties, but we would need to push the config values down to the factories, which doesn't seem to be possible.

pdabre12: It might be possible to have just one generic configMap with the properties, because if the properties aren't present in the provider class we just assume it to be true

pdabre12: @ZacBlanco 
What if we get a list of all the config properties using `injector.getInstance(ConfigurationFactory.class).getProperties()`; in `PrestoServer.java` and pass it down to `PluginManager.java`.
While loading the plugin, check the annotations to see if all the properties are satisfied, if yes load it.

Alternatively , if we don't want to use annotations, we can add a `getRequiredConfigs()` method in both `Plugin.java `and `CoordinatorPlugin.java` which would return a map of required properties to load this plugin.
Any plugin can override this method, returns its own set of configs, which can be validated before loading the plugin, making this validation generic for all plugins.
What do you think of this approach?

pdabre12: However, for the second approach, loading individual components of plugins like NativeSidecarPlugin might be tricky. For that use case, the annotations approach would be easier to handle.

amitkdutta:
@pdabre12 @ZacBlanco @rschlussel Folks will be really helpful if we can merge this one. Without this change, we can't include the plugin in build and can't test out any of the good changes coming with it. CC: @kewang1024

presto-main-base/src/main/java/com/facebook/presto/metadata/SessionPropertyManager.java
rschlussel: I think the second approach should be okay.  If you have the sidecar configured, as long as it's possible to *set* the particular component (e.g. constant folding thing) to the non-sidecar version without issue, it should be okay to have it loaded as long as there's no side effect. I think just making sure different implementations of the component (e.g. java builtin vs. sidecar) are registered different names should generally be enough.

pdabre12: How to control the loading of functionalities  of the sidecar? Like loading the function registry without the constant folding thing.
For eg: If you do not want to use `NativeTypeManager` and you load the `NativeSidecarPlugin` with `coordinator-sidecar-enabled` true, we load the `NativeTypeManager` which is the default type manager now, in this case, we want to skip loading the native type manager itself.

rschlussel: i think it's okay if it's loaded as long as it's not configured as the one to use and loading it in itself won't break anything.

rschlussel: and yes getRequiredConfigs() approach.

pdabre12: With the way the current code is, we can only have one between the `NativeWorkerSessionPropertyProvider` or pull the session properties from the sidecar. If we enable the sidecar, ideally we should not have the `NativeWorkerSessionPropertyProvider` loaded up. In the case we do, there will be a duplication of the session properties even if the session property providers are under different names. 
Maybe we can choose a mix of both the approaches, have `@ConditionalOnConfig` annotations for the inidvidual functionalities and have `getRequiredConfigs`() at the Plugin  level.

However, I suggest we let this one in , this change should skip the loading of the plugin if the necessary configs aren't satisfied. We can add the support for loading the individual functionalities of a plugin as a follow-up, depending on some more discussion.

presto-native-sidecar-plugin/src/test/java/com/facebook/presto/sidecar/TestNativeSidecarPlugin.java
```diff
@@ -69,7 +69,11 @@ protected QueryRunner createQueryRunner()
     protected QueryRunner createExpectedQueryRunner()
             throws Exception
     {
-        return PrestoNativeQueryRunnerUtils.createJavaQueryRunner();
+        DistributedQueryRunner queryRunner = (DistributedQueryRunner) PrestoNativeQueryRunnerUtils.createJavaQueryRunner();
```
pdabre12, line 72: @rschlussel This should test the case where the plugin is included but is not loaded because the configs weren't satisfied. What do you think?

presto-spi/src/main/java/com/facebook/presto/spi/Plugin.java
```diff
@@ -153,4 +155,9 @@ default Iterable<ClientRequestFilterFactory> getClientRequestFilterFactories()
     {
         return emptyList();
     }
+
+    default Map<String, String> getRequiredConfigs()
```
ZacBlanco, line 159: Can we add javadoc for this method?

presto-spi/src/main/java/com/facebook/presto/spi/CoordinatorPlugin.java
```diff
@@ -52,4 +55,9 @@ default Iterable<TypeManagerFactory> getTypeManagerFactories()
     {
         return emptyList();
     }
+
+    default Map<String, String> getRequiredConfigs()
```
ZacBlanco, line 59: Can we add javadoc for this?

presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/NativeSidecarPlugin.java
```diff
@@ -51,6 +54,12 @@ public Iterable<FunctionNamespaceManagerFactory> getFunctionNamespaceManagerFact
         return ImmutableList.of(new NativeFunctionNamespaceManagerFactory());
     }
 
+    @Override
+    public Map<String, String> getRequiredConfigs()
+    {
+        return ImmutableMap.of("coordinator-sidecar-enabled", "true");
```
ZacBlanco, line 60: Let's make `"coordinator-sidecar-enabled"` a constant somewhere?

presto-main-base/src/main/java/com/facebook/presto/server/PluginManager.java
```diff
@@ -384,6 +394,11 @@ public void installPlugin(Plugin plugin)
 
     public void installCoordinatorPlugin(CoordinatorPlugin plugin)
     {
+        if (!loadPluginIfConfigured(plugin.getRequiredConfigs())) {
+            log.info("Skipped loading %s because some conditions weren't satisfied", plugin.getClass().getName());
```
ZacBlanco, line 398: Maybe let's log inside the `loadPluginIfConfigured` instead? Then we could log the configuration property and value which was not satisfied

presto-main-base/src/main/java/com/facebook/presto/server/PluginManager.java
```diff
@@ -267,6 +272,11 @@ else if (plugin instanceof CoordinatorPlugin) {
 
     public void installPlugin(Plugin plugin)
     {
+        if (!loadPluginIfConfigured(plugin.getRequiredConfigs())) {
+            log.info("Skipped loading %s because some conditions weren't satisfied", plugin.getClass().getName());
```
ZacBlanco, line 276: Can remove if we put this inside `loadPluginIfConfigured`

ZacBlanco (changes requested):
Thanks for this improvement @pdabre12 . I like this approach. Just a few comments

presto-native-sidecar-plugin/src/test/java/com/facebook/presto/sidecar/TestNativeSidecarPlugin.java
rschlussel: the test seems okay, but would be nice to make it more explicit (single test that tests it).

presto-main-base/src/main/java/com/facebook/presto/server/PluginManager.java
```diff
@@ -506,4 +521,20 @@ private void processPlugins(Artifact artifact, ClassLoader classLoader, String s
             writePluginServices(plugins, artifact.getFile(), servicesFile);
         }
     }
+
+    private boolean loadPluginIfConfigured(Map<String, String> expectedProperties)
```
rschlussel, line 525: I found the name confusing since this doesn't do the loading. Can we rename it? Maybe something like `pluginIsEnabled`.

rschlussel (commented):
the approach looks good. Thanks!

pdabre12:
@rschlussel @ZacBlanco 
Thanks for the review.
Addressed your comments, please have another look.

presto-native-sidecar-plugin/src/test/java/com/facebook/presto/sidecar/TestNativeSidecarPlugin.java
pdabre12: Added an explicit test.

rschlussel (approved):

ZacBlanco (dismissed):

presto-native-sidecar-plugin/src/test/java/com/facebook/presto/sidecar/TestNativeSidecarPluginWithCoordinatorSidecarDisabled.java
```diff
@@ -0,0 +1,75 @@
...
+        DistributedQueryRunner queryRunner = (DistributedQueryRunner) PrestoNativeQueryRunnerUtils.createJavaQueryRunner();
+        // Installing the native sidecar plugin on a Java cluster, does not load the plugin functionalities because
+        // coordinator-sidecar-enabled is set to false on Java clusters.
+        queryRunner.installCoordinatorPlugin(new NativeSidecarPlugin());
+        return queryRunner;
+    }
+
+    @Test
+    public void testBasicQueries()
+    {
```
aditi-pandit, line 62: Would any of these queries be affected if the side-car plugin functionality was loaded ? It would be good to have a test which works when the side-car is enabled but does not otherwise.

aditi-pandit (commented):
Thanks @pdabre12 for this code. Have a quick question about testing.

presto-native-sidecar-plugin/src/test/java/com/facebook/presto/sidecar/TestNativeSidecarPluginWithCoordinatorSidecarDisabled.java
pdabre12: Added a test case.

aditi-pandit (dismissed):
Thanks @pdabre12

pdabre12:
@ZacBlanco @rschlussel Can one of you please approve again?

ZacBlanco (dismissed):

presto-spi/src/main/java/com/facebook/presto/spi/Plugin.java
```diff
@@ -153,4 +155,15 @@ default Iterable<ClientRequestFilterFactory> getClientRequestFilterFactories()
...
         return emptyList();
     }
+
+    /**
+     * This method is used to declare the minimal set of configuration entries that must be present
+     * for the plugin to be eligible for loading.
+     *
+     * @return a map of required configuration keys and expected values for the plugin.
+     **/
+    default Map<String, String> getRequiredConfigs()
```
tdcmeehan, line 165: Right now, this is just used to binary enable/disable the plugin.  Instead of making it freeform string, and then doing a deep comparison to ensure both keys and values match, why not introduce a new type, say, `PluginConfiguration`, that has just one field, `enabled`?

As a Plugin author, this method seems confusing, because it's only used to enable or disable the plugin, yet the contract appears much wider ranging, accepting an arbitrary string to string Map.

tdcmeehan (changes requested):

amitkdutta:
Folks, it will be really helpful if we can merge this PR quickly. We have new C++ specific functions, that does not exist in java world and native sidecar is the best way to make those discoverable to the coordinator. Our customers are adding C++ functions, and as a work around they are adding dummy java functions that just throws in execution. With the sidecar integration, they can just add the function in worker.

presto-spi/src/main/java/com/facebook/presto/spi/Plugin.java
pdabre12: I changed the approach altogether.

pdabre12:
Updated the approach:
All functionality of the sidecar plugin is now controlled via a configuration file. If the config file is present, the corresponding functionality will be loaded; otherwise, it will not be. This ensures that the build will not fail even if the plugin is included in clusters where sidecar support is disabled.

The type manager loading process remains straightforward and is similar to the function namespace manager loading.

In java clusters and non-sidecar enabled clusters, all worker providers will still be bound as before. However, when sidecar is enabled, `NativeWorkerSessionPropertyProvider ` is now conditionally loaded based on the config. Specifically, when `coordinator-sidecar-enabled` is set to true, the `NativeWorkerSessionPropertyProvider` will not be bound, and all `native-worker` specific properties will be provided via the sidecar.

presto-spi/src/main/java/com/facebook/presto/spi/Plugin.java
pdabre12: Instead of asking the Plugin author to provide the required set of configs, we would use config files to load individual functionalities of the sidecar. What do you think?

amitkdutta:
@pdabre12 Thanks for iterating on it. Just to clarify a bit more, there is no build failure. In https://github.com/prestodb/presto/issues/24964, we discussed that

1. coordinator always loads the plugin even if `"coordinator-sidecar-enabled": "false"`, this makes coordinator crash
```
025-04-23T12:55:49.520-0700	ERROR	main	com.facebook.presto.server.PrestoServer	System session property provider is already registered for property provider : native-worker
java.lang.IllegalArgumentException: System session property provider is already registered for property provider : native-worker
	at com.facebook.presto.metadata.SessionPropertyManager.loadSessionPropertyProvider(SessionPropertyManager.java:147)
	at com.facebook.presto.metadata.SessionPropertyManager.loadSessionPropertyProviders(SessionPropertyManager.java:154)
	at com.facebook.presto.server.PrestoServer.run(PrestoServer.java:193)
```

Coordinator crashes at runtime, not in build time. We want to control it with  `"coordinator-sidecar-enabled"` property (for example in java clusters even if plugin directory contains this plugin, the property will be set to `false` and won't make coordinator crash at runtime.

pdabre12:
@amitkdutta Yes, my bad, its a runtime issue.
We will control it using `coordinator-sidecar-enabled` as well as using the individual functionalities config files.

To clarify, let's say if users want to use the `NativeFunctionNamespaceManager`, users need to set up the function namespace under `presto-main/etc/function-namespace/`, if not the `NativeFunctionNamespaceManager` won't be used.
With this approach, we can include the `presto-native-sidecar-plugin` in the build directory, but not use its functionalities unless explicitly configured using config files.

```
025-04-23T12:55:49.520-0700	ERROR	main	com.facebook.presto.server.PrestoServer	System session property provider is already registered for property provider : native-worker
java.lang.IllegalArgumentException: System session property provider is already registered for property provider : native-worker
	at com.facebook.presto.metadata.SessionPropertyManager.loadSessionPropertyProvider(SessionPropertyManager.java:147)
	at com.facebook.presto.metadata.SessionPropertyManager.loadSessionPropertyProviders(SessionPropertyManager.java:154)
	at com.facebook.presto.server.PrestoServer.run(PrestoServer.java:193)
```
	
If the `NativeSystemSessionPropertyProvider` which pulls the session properties from the sidecar isn't configured, the coordinator won't attempt to load it and the above coordinator crash would not happen.

presto-native-sidecar-plugin/src/test/java/com/facebook/presto/sidecar/TestNativeSidecarPlugin.java
```diff
@@ -39,7 +39,6 @@
 import static org.testng.Assert.assertFalse;
 import static org.testng.Assert.fail;
 
-@Test(singleThreaded = true)
```
rschlussel, line 42: why did you remove this?

rschlussel (commented):
approach seems good. At some point when all the functionality is complete/well tested we should change the default. if sidecar is enabled use the sidecar, just to make configuration easier. Because this does make configuration more complex.

rschlussel:
actually, this seems like it might become really burdensome for people to configure to use the sidecar vs. just a single config. would it be possible to make the default be to load the sidecar plugin when sidecar is enabled even if there is no e.g. session properties config set.

rschlussel:
because previously you just needed to set one config (coordinator-sidecar-enabled).  And now if you want to use the sidecar you need to add new config files for all the different features of the sidecar to say that you want to use the sidecar for it (function manager, session properties manager).  I worry it makes it harder to get started.

presto-native-sidecar-plugin/src/test/java/com/facebook/presto/sidecar/TestNativeSidecarPlugin.java
pdabre12: I added this change in. Maybe at the time being I was checking some tests that needed to be executed concurrently, but never pushed those upstream. I don't think there shouldn't be any issue in making this change.

tdcmeehan (approved):

PR #24973, opened 2025-04-24
Fix ANALYZE when Hive partition has non-canonical value

denodo-research-labs:
## Description
Extracted from https://github.com/trinodb/trino/pull/15995

In Hive it may well happen that a partition value is written by the writer process as a string,
e.g. : `month=02`, even though the column is registered in Hive as an integer.

When updating the table or when doing `ANALYZE`, the output in Presto of the statistics computation though for the partition
from the example above will be though `2`, ending in the the following error: `All computed statistics must be used`.

## Motivation and Context

While performing ANALYZE on the following partitioned dataset:
`store_sales/d_year=2025/d_month=01/d_day=10/d_hour=00`

the following exception occurs:

```
com.google.common.base.VerifyException: All computed statistics must be used
at com.google.common.base.Verify.verify(Verify.java:126)
at com.facebook.presto.hive.HiveMetadata.finishStatisticsCollection(HiveMetadata.java:1552)
at com.facebook.presto.spi.connector.classloader.ClassLoaderSafeConnectorMetadata.finishStatisticsCollection(ClassLoaderSafeConnectorMetadata.java:210)
at com.facebook.presto.metadata.MetadataManager.finishStatisticsCollection(MetadataManager.java:761)
```

This PR addresses the above mentioned issue by parsing the partition values to Presto values in order to avoid ignoring computed statistics.

## Test Plan
Added test method `testAnalyzePartitionedTableWithNonCanonicalValues`

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

Hive Connector Changes
* Fix incorrectly ignoring computed table statistics in `ANALYZE`

### DIFF
--- a/presto-hive/src/main/java/com/facebook/presto/hive/HiveMetadata.java
+++ b/presto-hive/src/main/java/com/facebook/presto/hive/HiveMetadata.java
@@ -16,7 +16,9 @@ package com.facebook.presto.hive;
 import com.facebook.airlift.json.JsonCodec;
 import com.facebook.airlift.json.smile.SmileCodec;
 import com.facebook.presto.common.CatalogSchemaName;
+import com.facebook.presto.common.Page;
 import com.facebook.presto.common.Subfield;
+import com.facebook.presto.common.block.Block;
 import com.facebook.presto.common.predicate.NullableValue;
 import com.facebook.presto.common.predicate.TupleDomain;
 import com.facebook.presto.common.type.ArrayType;
@@ -233,7 +235,6 @@ import static com.facebook.presto.hive.HiveSessionProperties.isStatisticsEnabled
 import static com.facebook.presto.hive.HiveSessionProperties.isUsePageFileForHiveUnsupportedType;
 import static com.facebook.presto.hive.HiveSessionProperties.shouldCreateEmptyBucketFilesForTemporaryTable;
 import static com.facebook.presto.hive.HiveStatisticsUtil.createPartitionStatistics;
-import static com.facebook.presto.hive.HiveStatisticsUtil.getColumnStatistics;
 import static com.facebook.presto.hive.HiveStorageFormat.AVRO;
 import static com.facebook.presto.hive.HiveStorageFormat.DWRF;
 import static com.facebook.presto.hive.HiveStorageFormat.ORC;
@@ -281,6 +282,7 @@ import static com.facebook.presto.hive.HiveUtil.encodeMaterializedViewData;
 import static com.facebook.presto.hive.HiveUtil.encodeViewData;
 import static com.facebook.presto.hive.HiveUtil.getPartitionKeyColumnHandles;
 import static com.facebook.presto.hive.HiveUtil.hiveColumnHandles;
+import static com.facebook.presto.hive.HiveUtil.parsePartitionValue;
 import static com.facebook.presto.hive.HiveUtil.translateHiveUnsupportedTypeForTemporaryTable;
 import static com.facebook.presto.hive.HiveUtil.translateHiveUnsupportedTypesForTemporaryTable;
 import static com.facebook.presto.hive.HiveUtil.verifyPartitionTypeSupported;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: presto-hive/src/test/java/com/facebook/presto/hive/TestHiveAnalyze.java [hunk 1]
```diff
@@ -0,0 +1,100 @@
...
+
+        String tableLocation = (String) computeActual(format("SELECT DISTINCT regexp_replace(\"$path\", '/[^/]*/[^/]*$', '') FROM %s", tableName)).getOnlyValue();
+
+        String externalTableName = "external_" + tableName;
+        assertUpdate(format(
+                "CREATE TABLE %s (a_varchar varchar, month integer) WITH (partitioned_by = ARRAY['month'], external_location='%s')", externalTableName, tableLocation));
+
+        assertUpdate(format("CALL system.sync_partition_metadata('%s', '%s', 'ADD')", SCHEMA, externalTableName));
+        assertQuery(format("SELECT * FROM \"%s$partitions\"", externalTableName), "SELECT * FROM VALUES 1, 2, 3");
+        assertUpdate(format("ANALYZE %s", externalTableName), 4);
```
hantangwangd, line 78: Maybe we should also consider the scenario when we inserting `VALUES ('A', '1'), ('B', '01'), ('C', '001'), ('D', '03')` into the first table. It appears that there are still some problems in this scenario.

hantangwangd (commented):
Thanks for this fix. Found an issue still pending, please LMK if I got anything wrong.

hantangwangd:
Yes, I think your point about this not being a common scenario makes sense. But theoretically such scenarios do exist. So, do you think it makes sense to throw an exception with more explicit message rather than `All computed statistics must be used` in such a scenario to indicate that this is a known unsupported case rather than an unexpected issue or bug?

presto-hive/src/test/java/com/facebook/presto/hive/TestHiveAnalyze.java
hantangwangd: Yes, I think your point about this not being a common scenario makes sense. But theoretically such scenarios do exist. So, do you think it makes sense to throw an exception with more explicit message rather than `All computed statistics must be used` in such a scenario to indicate that this is a known unsupported case rather than an unexpected issue or bug?

presto-hive/src/main/java/com/facebook/presto/hive/HiveMetadata.java
```diff
@@ -1575,11 +1595,44 @@ public void finishStatisticsCollection(ConnectorSession session, ConnectorTableH
                     partitionStatistics.put(partitionValues, createPartitionStatistics(session, columnTypes, collectedStatistics, timeZone));
                 }
             }
-            verify(usedComputedStatistics == computedStatistics.size(), "All computed statistics must be used");
+            verify(usedComputedStatistics == computedStatistics.size(),
+                    "There are multiple variants of the same partition, e.g. p=1, p=01, p=001. All partitions must follow the same key=value representation");
```
hantangwangd:
```suggestion
            verify(usedComputedStatistics == computedStatistics.size(),
                    usedComputedStatistics > computedStatistics.size() ?
                            "There are multiple variants of the same partition, e.g. p=1, p=01, p=001. All partitions must follow the same key=value representation" :
                            "All computed statistics must be used");
```

Do you think it makes sense to throw the new exception message only if when `usedComputedStatistics > computedStatistics.size()`? Since we're not entirely sure whether there are other scenarios that could cause partition value mismatches, so IMO maybe it's better to still throw the original exception message in other cases. What's your opinion?

denodo-research-labs: Sounds good. Do you agree with the content of the new error message?

hantangwangd: It looks fine to me.

hantangwangd (approved):

PR #16412, opened 2021-07-13
Supports pushing down part of the filters when all filters cannot be pushed down

397090770:
The corresponding issue is [#16408](https://github.com/prestodb/presto/issues/16408)
```
== RELEASE NOTES ==

JDBC Changes
* Support partial pushdown of JDBC filters.

PR #24992, opened 2025-04-28
Changes to support Join Pushdown for JDBC Connectors

Haritha-Koloth:
## Description
This PR implements connector-side support for join pushdown in JDBC connectors, leveraging the foundational changes introduced in [24115](https://github.com/prestodb/presto/pull/24115). The changes enable JDBC connectors to perform join pushdown by rebuilding the query to execute joins directly on the remote datasource.

## Motivation and Context
Fixes [23152](https://github.com/prestodb/presto/issues/23152)

## Impact
This update includes the connector-side changes required to enable join pushdown in JDBC connectors, building on the foundational SPI and core changes introduced earlier. 

Detailed RFC - https://github.com/prestodb/rfcs/pull/32

## Test Plan
Unit tests added for newly added logic

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* ... 
* ... 

Hive Connector Changes
* ... 
* ... 
```

```
== NO RELEASE NOTE ==
```


presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/QueryBuilder.java
```diff
@@ -198,7 +211,109 @@ else if (typeAndValue.getType() instanceof CharType) {
...
+            accumulator.addAll(additionalPredicate.get().getBoundConstantValues().stream()
+                    //see https://github.com/Ahana-Inc/prestodb/pull/144#discussion_r895353628 for details.
+                    .map(constantExpression -> new TypeAndValue(constantExpression.getType(), constantExpression.getValue()))
+                    .collect(ImmutableList.toImmutableList()));
+        }
+        if (!clauses.isEmpty()) {
+            sql.append(" WHERE ");
+            sql.append(String.join(" AND ", clauses));
+        }
+        sql.append(String.format("/* %s : %s */", session.getUser(), session.getQueryId()));
```
aaneja, line 262: Please remove - we are leaking presto internal data here

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/QueryBuilder.java
```diff
@@ -198,7 +211,109 @@ else if (typeAndValue.getType() instanceof CharType) {
...
+                statement.setString(i + 1, ((Slice) typeAndValue.getValue()).toStringUtf8());
+            }
+            else if (typeAndValue.getType() instanceof CharType) {
+                statement.setString(i + 1, ((Slice) typeAndValue.getValue()).toStringUtf8());
+            }
+            else {
+                throw new UnsupportedOperationException("Can't handle type: " + typeAndValue.getType());
+            }
+        }
+        log.debug("Query is Pushed down");
```
aaneja, line 315: Please remove these logs

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/QueryBuilder.java
```diff
@@ -198,7 +211,109 @@ else if (typeAndValue.getType() instanceof CharType) {
...
+        log.debug("Actual Query :: ", sql.toString());
+        return statement;
+    }
+
+    private boolean checkIfJoinQuery(Optional<List<ConnectorTableHandle>> joinPushdownTables)
+    {
+        return (joinPushdownTables != null && joinPushdownTables.isPresent() && !joinPushdownTables.get().isEmpty());
+    }
+
+    private PreparedStatement buildJoinSql(
```
aaneja, line 224: Instead of completely forking the buildSql code (and repeating its parts), please refactor by creating smaller methods that take in a input of`StringBuilder sql` and append SQL that is needed to query the joined tables using the `FROM table1, table2,.., tableX` format. This will make writing tests for buildSQL much easier too

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/QueryBuilder.java
```diff
@@ -271,29 +386,40 @@ private String toPredicate(String columnName, Domain domain, Type type, List<Typ
...
+            disjuncts.add(getColumnIdentifier(columnName, columnHandle, joinPushdownEnabled) + " IS NULL");
         }
 
         return "(" + Joiner.on(" OR ").join(disjuncts) + ")";
     }
 
-    private String toPredicate(String columnName, String operator, Object value, Type type, List<TypeAndValue> accumulator)
+    private String getColumnIdentifier(String columnName, JdbcColumnHandle columnHandle, boolean joinPushdownEnabled)
+    {
+        Optional<String> tableAlias = columnHandle.getTableAlias();
```
aaneja, line 410: Please remove the `joinPushdownEnabled` parameter. If a tableAlias is set, let's use it

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/JdbcTableHandle.java
```diff
@@ -36,19 +38,29 @@ public final class JdbcTableHandle
     private final String schemaName;
     private final String tableName;
 
+//    joinTables holds all the table handles for the same connector which are participating for the join push down
+//    this is used later in query builder to rebuild the join query
+
+    private Optional<List<ConnectorTableHandle>> joinTables;
```
aaneja, line 44: Use a `List<ConnectorTableHandle>> joinTables`. An empty list is equivalent to an `Optional.empty()` ?

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/optimization/JdbcJoinPushdown.java
```diff
@@ -0,0 +1,132 @@
...
+        {
+            TableHandle tableHandle = node.getTable();
+            if (tableHandle.getConnectorHandle() instanceof JoinTableSet) {
+                JoinTableSet tableHandles = (JoinTableSet) tableHandle.getConnectorHandle();
+                ImmutableList<JoinTableInfo> newJoinTableInfos = ImmutableList.copyOf(tableHandles.getInnerJoinTableInfos());
+
+                List<ConnectorTableHandle> newConnectorTableHandles = new ArrayList<>();
+                Map<VariableReferenceExpression, ColumnHandle> groupAssignments = new HashMap<>();
+                final String aliasPrefix = "T";
+                int tableCount = 0;
```
aaneja, line 99: Let's rename this `aliasTableCounter`

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/optimization/JdbcJoinPushdown.java
```diff
@@ -0,0 +1,132 @@
...
+        public PlanNode visitTableScan(TableScanNode node, RewriteContext<Void> context)
+        {
+            TableHandle tableHandle = node.getTable();
+            if (tableHandle.getConnectorHandle() instanceof JoinTableSet) {
+                JoinTableSet tableHandles = (JoinTableSet) tableHandle.getConnectorHandle();
+                ImmutableList<JoinTableInfo> newJoinTableInfos = ImmutableList.copyOf(tableHandles.getInnerJoinTableInfos());
+
+                List<ConnectorTableHandle> newConnectorTableHandles = new ArrayList<>();
+                Map<VariableReferenceExpression, ColumnHandle> groupAssignments = new HashMap<>();
+                final String aliasPrefix = "T";
```
aaneja, line 98: Can you add a comment here -
`Generate aliases for each table being joined to avoid any ambuigious column name references. These aliases are used in QueryBuilder#buildSql`

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/optimization/JdbcJoinPushdown.java
```diff
@@ -0,0 +1,132 @@
...
+                int tableCount = 0;
+
+                // Create new table handles and update column handles
+                for (JoinTableInfo tableMapping : newJoinTableInfos) {
+                    JdbcTableHandle handle = (JdbcTableHandle) tableMapping.getTableHandle();
+                    JdbcTableHandle newHandle = new JdbcTableHandle(handle.getConnectorId(), handle.getSchemaTableName(), handle.getCatalogName(),
+                            handle.getSchemaName(), handle.getTableName(), handle.getJoinTables(), Optional.of(aliasPrefix + (++tableCount)));
+                    newConnectorTableHandles.add(newHandle);
+
+                    // Update column handles using Streams
```
aaneja, line 108: I couldn't understand this comment

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/optimization/JdbcJoinPushdown.java
```diff
@@ -0,0 +1,132 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+import static com.facebook.presto.spi.ConnectorPlanRewriter.rewriteWith;
+import static java.util.Objects.requireNonNull;
+
+/**
+ * JdbcJoinPushdown Optimizer:
+ * This Optimizer converts ConnectorHandleSet with various TableHandles to JdbcTableHandle which has list of all the JdbcTableHandles that can be pushed down.
```
aaneja, line 46:
```suggestion
* This Optimizer converts a {@link JoinTableSet} to a {@link JdbcTableHandle} with it's {@link JdbcTableHandle#joinTables} set to the
 * those from {@link JoinTableSet#innerJoinTableInfos}
```

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/optimization/JdbcComputePushdown.java
```diff
@@ -80,69 +96,138 @@ public PlanNode optimize(
...
 
-            JdbcTableLayoutHandle oldTableLayoutHandle = (JdbcTableLayoutHandle) oldTableHandle.getLayout().get();
+            List<String> sqlBodies = mergeSqlBodies(translatedExpressions);
+            List<ConstantExpression> variableBindings = mergeVariableBindings(translatedExpressions);
+            translated = Optional.of(new JdbcExpression(format("%s", Joiner.on(" AND ").join(sqlBodies)), variableBindings));
+            TableScanNode newTableScanNode = createNewTableScanNode(oldTableScanNode, oldTableHandle, oldTableLayoutHandle, translated);
+
+            RowExpression remainingPredicates = logicalRowExpressions.combineConjuncts(remainingExpressions);
+            /// !!! For now, we only create a new FilterNode with the remaining predicates
+            // As per https://github.com/prestodb/presto/pull/16864, this is not sufficient and we should re-use the existing filter predicate for correctness
```
aaneja, line 199: @rschlussel @NikhilCollooru 
Do you recall what was the scenario where a partial pushdown of a filter (while retaining the non-pushed down part in Presto) was causing correctness issues mentioned in https://github.com/prestodb/presto/pull/16864 ?

We want to reintroduce the changes from https://github.com/prestodb/presto/pull/16412, and we'd like to know if we need to put these behind a feature-flag

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/JdbcMetadata.java
```diff
@@ -267,4 +280,24 @@ public TableStatistics getTableStatistics(ConnectorSession session, ConnectorTab
...
     }
+
+    @Override
+    public boolean isPushdownSupportedForFilter(ConnectorSession session,
+            ConnectorTableHandle tableHandle,
+            RowExpression filter,
+            Map<VariableReferenceExpression, ColumnHandle> symbolToColumnHandleMap)
+    {
+        if (tableHandle instanceof JoinTableSet) {
+            // We now have all relevant pieces to build a JdbcJoinPredicateToSqlTranslator to test if the passed in filter will translate to SQL
```
aaneja, line 291: I think this comment can be removed

aaneja (changes requested):
Please refactor QueryBuilder to smaller methods that can be unit tested better. This will remove the need for large tests like `TestJdbcQueryBuilderJoinPushdown` IMO

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/QueryBuilder.java
Haritha-Koloth: @aaneja This code was copied from the existing buildSql() function and not introduced specifically for handling pushdown. We are not sure of the impact of removing from the existing code flow.

presto-base-jdbc/src/test/java/com/facebook/presto/plugin/jdbc/TestJoinQueriesWithPushDown.java
```diff
@@ -0,0 +1,2382 @@
...
+        Session pushdownEnabled = getSession(true);
+        Session disabled = getSession(false);
+        assertQueryWithSameQueryRunner(pushdownEnabled, "SELECT o.orderkey, c.name, o.totalprice, c.custkey " +
+                "FROM orders o " +
+                "JOIN customer c ON o.custkey = c.custkey " +
+                "WHERE o.totalprice < c.acctbal / 2", disabled);
+    }
+
+    @Test
+    public void testRowFieldAccessorInJoin()
```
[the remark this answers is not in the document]
Haritha-Koloth: @aaneja We observed performance degradation in several of these tests due to bottlenecks associated with the H2 database. Therefore, we're unable to include the suggested class with `optimizer_inner_join_pushdown_enabled` enabled in the base test class `AbstractTestJoinQueries`. Is that okay, or do you think there’s a need for additional tests?

presto-base-jdbc/src/test/java/com/facebook/presto/plugin/jdbc/TestJdbcQueryBuilderJoinPushdownExtended.java
```diff
@@ -0,0 +1,547 @@
...
+            ImmutableSet<Integer> expectedIntegerCol1Values = ImmutableSet.of(0, 1, 2, 3, 4, 5, 6, 7, 8, 9);
+            ImmutableSet<Integer> expectedIntegerCol2Values = ImmutableSet.of(0, 1, 2, 3, 4, 5, 6, 7, 8, 9);
+            ImmutableSet<Integer> expectedIntegerCol3Values = ImmutableSet.of(0, 1, 2, 3, 4, 5, 6, 7, 8, 9);
+
+            // We are Asserting the preparedStatement and also the resultSet.
+            String expectedPreparedStatement = "SELECT \"table_alias_1\".\"integer_col_1\", \"table_alias_2\".\"integer_col_2\", \"table_alias_3\".\"integer_col_3\" " +
+                    "FROM \"test_schema\".\"test_table_3\" \"table_alias_3\", \"test_schema\".\"test_table_2\" \"table_alias_2\", \"test_schema\".\"test_table_1\" \"table_alias_1\" " +
+                    "WHERE \"table_alias_1\".\"integer_col_1\" = \"table_alias_2\".\"integer_col_2\" AND \"table_alias_2\".\"integer_col_2\" = \"table_alias_3\".\"integer_col_3\"";
+
+            assertTrue(preparedStatement.toString().contains(expectedPreparedStatement),
```
aaneja, line 333: While I appreciate the thoroughness of this and other tests in this class, the core of what we want to test is how QueryBuilder is now able to -
1. Use aliases while building the SELECT list
2. Can build a FROM clause that represents a CSV of join tables

These tests can be added as additional tests in `TestJdbcQueryBuilder`.
The rest of what you're testing here - that a query with aliases, FROM clause and join related WHERE clauses in the body is only testing if H2 can support this syntax. This is something we already know it does, so we don't need to _acutally execute the query_ on H2 to prove this

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/QueryBuilder.java
```diff
@@ -103,24 +107,49 @@ public PreparedStatement buildSql(
             String catalog,
             String schema,
             String table,
+            List<ConnectorTableHandle> joinPushdownTables,
             List<JdbcColumnHandle> columns,
             TupleDomain<ColumnHandle> tupleDomain,
             Optional<JdbcExpression> additionalPredicate)
             throws SQLException
     {
+        boolean joinPushdownEnabled = checkIfJoinQuery(joinPushdownTables);
```
aaneja, line 116: I am going to push up a commit to your branch that further refactors this class, making it more readable, please take a look

aaneja (changes requested):

presto-base-jdbc/src/test/java/com/facebook/presto/plugin/jdbc/optimization/TestJdbcJoinPushdown.java
```diff
@@ -0,0 +1,280 @@
...
+    {
+        return new JdbcColumnHandle(CONNECTOR_ID, name, new JdbcTypeHandle(Types.BIGINT, "integer", 10, 0), BIGINT, false, Optional.empty(), Optional.empty());
+    }
+
+    private static JdbcColumnHandle booleanJdbcColumnHandle(String name)
+    {
+        return new JdbcColumnHandle(CONNECTOR_ID, name, new JdbcTypeHandle(Types.BOOLEAN, "boolean", 1, 0), BOOLEAN, false, Optional.empty(), Optional.empty());
+    }
+
+    private TableScanNode combinedJdbcTableScan(Type type, String... columnNames)
```
aaneja, line 152: Can you add a Javadoc for this - 
```
Gets a TableScanNode that represent the output of {@code GroupInnerJoinsByConnectorRuleSet} for a JDBC source - 
TableScanNode
   - Pushed down predicate - `('c1' + 'c2') - 'c2')`
   - (T1,T2,T3)
```

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/optimization/JdbcJoinPushdown.java
```diff
@@ -0,0 +1,135 @@
...
+
+        @Override
+        public PlanNode visitTableScan(TableScanNode node, RewriteContext<Void> context)
+        {
+            TableHandle tableHandle = node.getTable();
+            if (tableHandle.getConnectorHandle() instanceof JoinTableSet) {
+                JoinTableSet tableHandles = (JoinTableSet) tableHandle.getConnectorHandle();
+                ImmutableList<JoinTableInfo> newJoinTableInfos = ImmutableList.copyOf(tableHandles.getInnerJoinTableInfos());
+
+                List<ConnectorTableHandle> newConnectorTableHandles = new ArrayList<>();
```
aaneja, line 97: Use immutable list builders

presto-base-jdbc/src/test/java/com/facebook/presto/plugin/jdbc/optimization/TestJdbcJoinPushdown.java
```diff
@@ -0,0 +1,280 @@
...
+
+    @Test
+    public void testJdbcJoinPushdownWithJoins()
+    {
+        PlanNode original = combinedJdbcTableScan(BIGINT, "c1", "c2");
+        ConnectorSession session = new TestingConnectorSession(ImmutableList.of());
+        PlanNode actual = this.jdbcJoinPushdown.optimize(original, session, null, ID_ALLOCATOR);
+
+        final String aliasPrefix = "T";
+        List<ConnectorTableHandle> tableHandles = new ArrayList<>();
```
aaneja, line 100: Let's rename this to be more descriptive `expectedJoinTableHandles`, please use a ImmutableList builder for this

presto-base-jdbc/src/test/java/com/facebook/presto/plugin/jdbc/optimization/TestJdbcJoinPushdown.java
```diff
@@ -0,0 +1,280 @@
...
+
+        @Override
+        public MatchResult detailMatches(PlanNode node, StatsProvider stats, Session session, Metadata metadata, SymbolAliases symbolAliases)
+        {
+            checkState(shapeMatches(node), "Plan testing framework error: shapeMatches returned false in detailMatches in %s", this.getClass().getName());
+
+            TableScanNode tableScanNode = (TableScanNode) node;
+            JdbcTableLayoutHandle layoutHandle = (JdbcTableLayoutHandle) tableScanNode.getTable().getLayout().get();
+
+            if (jdbcTableLayoutHandle.getTable().equals(layoutHandle.getTable())
```
aaneja, line 265: Add a comment mentioning the reason why we can't just do  `jdbcTableLayoutHandle.equals(layoutHandle)` and do a custom equals comparison instead

presto-base-jdbc/src/test/java/com/facebook/presto/plugin/jdbc/optimization/TestJdbcJoinPushdown.java
```diff
@@ -0,0 +1,280 @@
...
+
+        return PLAN_BUILDER.tableScan(
+                tableHandle,
+                outputVariables,
+                assignments);
+    }
+
+    private TableScanNode jdbcTableScan(String schema, String table, Type type, boolean isJoinPushdown, String... columnNames)
+    {
+        JdbcTableHandle jdbcTableHandle = null;
```
aaneja, line 184: null assignments here are redundant

presto-base-jdbc/src/test/java/com/facebook/presto/plugin/jdbc/optimization/TestJdbcJoinPushdown.java
```diff
@@ -0,0 +1,280 @@
...
+        ConnectorTableHandle connectorTableHandle = null;
+        if (isJoinPushdown) {
```
aaneja, line 186: `isJoinPushdown` is always false in current usage. You can delete the parameter as well as the if-true block amd just retain the else part

presto-base-jdbc/src/test/java/com/facebook/presto/plugin/jdbc/optimization/TestJdbcJoinPushdown.java
```diff
@@ -0,0 +1,280 @@
...
+        public MatchResult detailMatches(PlanNode node, StatsProvider stats, Session session, Metadata metadata, SymbolAliases symbolAliases)
+        {
+            checkState(shapeMatches(node), "Plan testing framework error: shapeMatches returned false in detailMatches in %s", this.getClass().getName());
+
+            TableScanNode tableScanNode = (TableScanNode) node;
+            JdbcTableLayoutHandle layoutHandle = (JdbcTableLayoutHandle) tableScanNode.getTable().getLayout().get();
+
+            if (jdbcTableLayoutHandle.getTable().equals(layoutHandle.getTable())
+                    && jdbcTableLayoutHandle.getTupleDomain().equals(layoutHandle.getTupleDomain())
+                    && ((!jdbcTableLayoutHandle.getAdditionalPredicate().isPresent() && !layoutHandle.getAdditionalPredicate().isPresent())
```
aaneja, line 267: [additionalPredicate](https://github.com/prestodb/presto/blob/70f98518ac8080a3406b461d5ac534f19738ddc5/presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/JdbcTableLayoutHandle.java#L102) direct comparison should work fine ?

aaneja (changes requested):
Just some nits + a question on tests, otherwise LGTM

presto-base-jdbc/src/test/java/com/facebook/presto/plugin/jdbc/optimization/TestJdbcJoinPushdown.java
Haritha-Koloth: @aaneja Actually the simple comparison also seems to be working fine now. My guess is that this test was added during the initial development while there was dynamic alias generation which was breaking the comparison. I have updated the code to use direct object comparison.

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/optimization/JdbcComputePushdown.java
Haritha-Koloth: Hi @rschlussel , Could you share some specific query examples where this issue was observed? We tried a few queries on our end but couldn’t reproduce it.

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/QueryBuilder.java
```diff
@@ -106,17 +107,44 @@ public PreparedStatement buildSql(
...
             List<JdbcColumnHandle> columns,
             Map<String, String> columnExpressions,
             TupleDomain<ColumnHandle> tupleDomain,
             Optional<JdbcExpression> additionalPredicate)
             throws SQLException
     {
         StringBuilder sql = new StringBuilder();
+        buildSelectClause(columns, sql, columnExpressions);
 
+        if (!joinPushdownTables.isEmpty()) {
```
aaneja, line 129: One problem with this approach of passing join tables separately has been that if 
1. The connector decides to make a `JdbcTableHandle` with an alias (say as a default behavior)
2. There is no join pushdown, just a single table query

The `else` branch of the code is used, which does _not_ add the alias to the `FROM` clause
This would be okay, except `getColumnIdentifier(JdbcColumnHandle columnHandle)` will add the alias in the `WHERE` clause generated, and this makes the query inconsistent

An example we saw internally was this query - 
`SELECT "EMPLOYEE_ID", "EMPLOYEE_NAME", "DEPARTMENT_ID" FROM "TEST_SCHEMA"."T_EMPLOYEES" WHERE ("table_alias_99864094a938424a836d86a98c837879"."EMPLOYEE_NAME" = ?)`

IMO, the correct fix for this is to modify the `buildSql` method to either pass in the table alias, or remove the need to pass the `catalog`, `schema` and `table` separately

Haritha-Koloth: @aaneja Pushed the fix. Could you please review?

steveburnett:
Thanks for the release note entry! Formatting nit, section change, and minor rephrasing suggestion. 

```
== RELEASE NOTES ==

JDBC Changes
* Add ``catalog.partial-predicate-push-down`` configuration property and ``partial_predicate_push_down`` session property. Applicable only to JDBC connectors.
When set to ``true`` these properties enable partial predicate pushdown to the underlying data source. Enable this property for join pushdown to be effective with JDBC connectors.
```

Haritha-Koloth:
> Thanks for the release note entry! Formatting nit, section change, and minor rephrasing suggestion.
> 
> ```
> == RELEASE NOTES ==
> 
> JDBC Changes
> * Add ``catalog.partial-predicate-push-down`` configuration property and ``partial_predicate_push_down`` session property. Applicable only to JDBC connectors.
> When set to ``true`` these properties enable partial predicate pushdown to the underlying data source. Enable this property for join pushdown to be effective with JDBC connectors.
> ```

Thanks for the review @steveburnett . I've updated the release notes.

I am seeing this error though, Should we use `Connector Changes` instead of `JDBC Changes` ?

steveburnett:
@Haritha-Koloth , sorry about that! Looking at the failing test, I think it's looking for 
`JDBC Driver Changes`
so try that. 

I'll update the [Order of sections](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines#order-of-sections) in the Release Notes Guidelines from `JDBC` to `JDBC Driver`.

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/optimization/JdbcComputePushdown.java
```diff
@@ -80,69 +96,143 @@ public PlanNode optimize(
...
+            if (translated.isPresent()) {
+                return createNewTableScanNode(oldTableScanNode, oldTableHandle, oldTableLayoutHandle, translated);
             }
+            if (JdbcSessionProperties.isPartialPredicatePushDownEnabled(this.session)) {
+/*
+                Copied from the PR to enhance pushing part filters down to the tablehandle
+                - https://github.com/prestodb/presto/pull/16412/files.
+                This will find out which parts can be pushed down and able to translate by connector,
+                and the remaining part will keep in a new FilterNode.
+             */
```
aaneja: nit: Don't use block comments

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/optimization/JdbcComputePushdown.java
```diff
@@ -80,69 +96,143 @@ public PlanNode optimize(
...
+
+                List<String> sqlBodies = mergeSqlBodies(translatedExpressions);
+                List<ConstantExpression> variableBindings = mergeVariableBindings(translatedExpressions);
+                translated = Optional.of(new JdbcExpression(format("%s", Joiner.on(" AND ").join(sqlBodies)), variableBindings));
+                TableScanNode newTableScanNode = createNewTableScanNode(oldTableScanNode, oldTableHandle, oldTableLayoutHandle, translated);
+
+                RowExpression remainingPredicates = logicalRowExpressions.combineConjuncts(remainingExpressions);
+                /// !!! For now, we only create a new FilterNode with the remaining predicates
+                // As per https://github.com/prestodb/presto/pull/16864, this is not sufficient and we should re-use the existing filter predicate for correctness
+                // We need to find a test case for jdbc pushdown that fails with this implementation
```
aaneja: Remove this comment now that we have a `isPartialPredicatePushDownEnabled` to gate this partial pushdown on ?

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/QueryBuilder.java
aaneja: This does not fix the issue. The only caller for this method is on [L270](https://github.com/Haritha-Koloth/presto/blob/b987da99844a3ebcc35db00a62fb589de12e1f7b/presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/QueryBuilder.java#L270) and that always passes the alias as `Optional.empty()`

presto-mysql/src/main/java/com/facebook/presto/plugin/mysql/MySqlClient.java
```diff
@@ -235,6 +236,7 @@ public PreparedStatement buildSql(ConnectorSession session, Connection connectio
                 split.getCatalogName(),
                 split.getSchemaName(),
                 split.getTableName(),
+                ImmutableList.of(),
```
aaneja: This would mean that MySQL connector does not get join pushdown ?

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/QueryBuilder.java
Haritha-Koloth: Currently, no existing connector requires this behavior, so the method(with alias as a param) is not in use. However, it is available for future connectors that may need it. We introduced a new method and invoked it from the existing one using `Optional.empty()` to ensure the current code flow remains unaffected.

presto-mysql/src/main/java/com/facebook/presto/plugin/mysql/MySqlClient.java
Haritha-Koloth: yes, this was a coding error. I have fixed it and pushed the change

steveburnett:
* Please add documentation for these new properties (``catalog.partial-predicate-push-down`` configuration property and ``partial_predicate_push_down`` session property) in 
https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/admin/properties-session.rst#jdbc-properties and https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/admin/properties.rst. 

* After you add the doc, you can revise the release note entry to link to the entries - see [Formatting](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines#formatting) in the Release Notes Guidelines for how to format the link. 

* The sentence 
  ```
  When set to ``true`` these properties enable partial predicate pushdown to the underlying data source. Enable this property for join pushdown to be effective with JDBC connectors.
  ```
  in the current release note entry can be included in the documentation and it can be removed from the release note entry.

Haritha-Koloth:
> * Please add documentation for these new properties (`catalog.partial-predicate-push-down` configuration property and `partial_predicate_push_down` session property) in
>   https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/admin/properties-session.rst#jdbc-properties and https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/admin/properties.rst.
> * After you add the doc, you can revise the release note entry to link to the entries - see [Formatting](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines#formatting) in the Release Notes Guidelines for how to format the link.
> * The sentence
>   ```
>   When set to ``true`` these properties enable partial predicate pushdown to the underlying data source. Enable this property for join pushdown to be effective with JDBC connectors.
>   ```
>   
>   
>       
>         
>       
>   
>         
>       
>   
>       
>     
>   in the current release note entry can be included in the documentation and it can be removed from the release note entry.

Thanks for the review @steveburnett . Clarifying the file where these properties should go in - Since these are supposed to be configured at connector level, ie in the corresponding catalog properties file, should we add them in all relevant connector.rst files?

steveburnett:
> Thanks for the review @steveburnett . Clarifying the file where these properties should go in - Since these are supposed to be configured at connector level, ie in the corresponding catalog properties file, should we add them in all relevant connector.rst files?

Thanks for catching that! Yes, you are correct.

PR #24996, opened 2025-04-28
Added support for geometry type in mysql connector

Mariamalmesfer:
Co-authored-by: pratyakssarma <pratyaks13@gmail.com>h13@gmail.com>

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* ... 
* ... 

Hive Connector Changes
* ... 
* ... 
```

```
== NO RELEASE NOTE ==
```


steveburnett:
Do the Type Mapping tables in https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/connector/mysql.rst#type-mapping need to be updated with this added type?

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/QueryBuilder.java
```diff
@@ -202,6 +197,24 @@ else if (typeAndValue.getType() instanceof CharType) {
         return statement;
     }
```
infvg, line 199: to keep backwards compatibility, maybe you could add a function with the old parameters that calls this with ImmutableMap.of() as a default value eg
```suggestion
     public PreparedStatement buildSql(
            JdbcClient client,
            ConnectorSession session,
            Connection connection,
            String catalog,
            String schema,
            String table,
            List<JdbcColumnHandle> columns,
            TupleDomain<ColumnHandle> tupleDomain,
            Optional<JdbcExpression> additionalPredicate)
            throws SQLException
    {
    return buildSql(client, session, connection, catalog, schema, table, columns, ImmutableMap.of(), tupleDomain, additionalPredicate);
    }
```

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/BaseJdbcClient.java
```diff
@@ -313,6 +313,7 @@ public PreparedStatement buildSql(ConnectorSession session, Connection connectio
                 split.getSchemaName(),
                 split.getTableName(),
                 columnHandles,
+                ImmutableMap.of(),
```
infvg, line 316: if the below is implemented this can be removed

presto-base-jdbc/src/test/java/com/facebook/presto/plugin/jdbc/TestJdbcQueryBuilder.java
```diff
@@ -201,7 +201,7 @@ public void testNormalBuildSql()
                 .build());
 
         Connection connection = database.getConnection();
-        try (PreparedStatement preparedStatement = new QueryBuilder("\"").buildSql(jdbcClient, session, connection, "", "", "test_table", columns, tupleDomain, Optional.empty());
+        try (PreparedStatement preparedStatement = new QueryBuilder("\"").buildSql(jdbcClient, session, connection, "", "", "test_table", columns, ImmutableMap.of(), tupleDomain, Optional.empty());
```
infvg, line 204:
```suggestion
        try (PreparedStatement preparedStatement = new QueryBuilder("\"").buildSql(jdbcClient, session, connection, "", "", "test_table", columns, tupleDomain, Optional.empty());
```

presto-mysql/src/main/java/com/facebook/presto/plugin/mysql/MySqlClient.java
```diff
@@ -173,6 +191,10 @@ protected String toSqlType(Type type)
             // In order to preserve microsecond information for TIMESTAMP_MICROSECONDS
             return "datetime(6)";
         }
+        if (type.getTypeSignature().getBase().equalsIgnoreCase("Geometry")) {
```
infvg, line 194:
```suggestion
        if (GEOMETRY.equals(type)) {
```
Use `com.facebook.presto.geospatial.type.GEOMETRY` 
You'll need to add presto-main-base as a dependency

presto-mysql/src/main/java/com/facebook/presto/plugin/mysql/MySqlClient.java
```diff
@@ -196,6 +218,71 @@ protected String toSqlType(Type type)
...
+    public PreparedStatement buildSql(ConnectorSession session, Connection connection, JdbcSplit split, List<JdbcColumnHandle> columnHandles) throws SQLException
+    {
+        Map<String, String> supposedColumnExps = new HashMap<>();
+        for (JdbcColumnHandle columnHandle : columnHandles) {
+            JdbcTypeHandle typeHandle = columnHandle.getJdbcTypeHandle();
+            if (typeHandle.getJdbcTypeName().equalsIgnoreCase("geometry")) {
+                String columnName = columnHandle.getColumnName();
+                supposedColumnExps.put(columnName, "ST_AsBinary(" + columnName + ")");
+            }
+        }
```
infvg, line 231:
```suggestion
        columnHandles.forEach(handle -> {
            if (handle.getJdbcTypeHandle().getJdbcTypeName().equals("geometry")) {
                String columnName = handle.getColumnName();
                supposedColumnExps.put(columnName, "ST_AsBinary(" + columnName + ")");
            }
        });
```

presto-mysql/src/main/java/com/facebook/presto/plugin/mysql/MySqlClient.java
```diff
@@ -196,6 +218,71 @@ protected String toSqlType(Type type)
...
+                session,
+                connection,
+                split.getCatalogName(),
+                split.getSchemaName(),
+                split.getTableName(),
+                columnHandles,
+                builder.build(),
+                split.getTupleDomain(),
+                split.getAdditionalPredicate());
+    }
```
infvg, line 248:
```suggestion
        ImmutableMap<String, String> columnExpressions = ImmutableMap.<String, String>builder().putAll(supposedColumnExps).build();

        return new QueryBuilder(identifierQuote).buildSql(
                this,
                session,
                connection,
                split.getCatalogName(),
                split.getSchemaName(),
                split.getTableName(),
                columnHandles,
                columnExpressions,
                split.getTupleDomain(),
                split.getAdditionalPredicate());
```

presto-mysql/src/main/java/com/facebook/presto/plugin/mysql/MySqlClient.java
```diff
@@ -196,6 +218,71 @@ protected String toSqlType(Type type)
...
+
+    @Override
+    public Optional<ReadMapping> toPrestoType(ConnectorSession session, JdbcTypeHandle typeHandle)
+    {
+        String typeName = typeHandle.getJdbcTypeName();
+
+        if (typeName.equalsIgnoreCase("geometry")) {
```
infvg, line 255: Maybe  check what the typeHandle int value is for geometry and use it with java.sql.Types.equals()

presto-postgresql/pom.xml
```diff
@@ -27,6 +27,7 @@
             <artifactId>configuration</artifactId>
         </dependency>
 
+
```
infvg, line 30: nit: remove this

infvg (commented):
Nice work LGTM logic is great just some minor code style suggestions

Mariamalmesfer:
> Do the Type Mapping tables in https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/connector/mysql.rst#type-mapping need to be updated with this added type?

Yes, I added the required change. The MySQL to PrestoDB type mapping table is updated to include GEOMETRY as VARBINARY

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/QueryBuilder.java
Mariamalmesfer: Yes, makes sense. I added it

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/BaseJdbcClient.java
Mariamalmesfer: Applied

steveburnett (commented):
LGTM! (docs)

Pull branch, local doc build, looks good. 

This updates the table in [MySQL to PrestoDB type mapping](https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/connector/mysql.rst#mysql-to-prestodb-type-mapping), but should a similar line be added to the table in [PrestoDB to MySQL type mapping](https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/connector/mysql.rst#prestodb-to-mysql-type-mapping)?

Mariamalmesfer:
> LGTM! (docs)
> 
> Pull branch, local doc build, looks good.
> 
> This updates the table in [MySQL to PrestoDB type mapping](https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/connector/mysql.rst#mysql-to-prestodb-type-mapping), but should a similar line be added to the table in [PrestoDB to MySQL type mapping](https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/connector/mysql.rst#prestodb-to-mysql-type-mapping)?

Right, this change only adds read support for GEOMETRY. Since we’re reading it as VARBINARY, it’s not supported for writing. That’s why I only updated the MySQL to PrestoDB type mapping table

steveburnett:
> yes I only updated the MySQL to PrestoDB type mapping table. the change only adds read support for GEOMETRY. Since we’re reading it as VARBINARY, it’s not supported for writing

Thanks for the explanation! Makes sense.

steveburnett (dismissed):
LGTM! (docs)

presto-mysql/src/main/java/com/facebook/presto/plugin/mysql/MySqlClient.java
Mariamalmesfer: I just checked the returned int type and it is -2 which corresponds to Types.BINARY so we still need to check for the type name to make sure it is a geometry type.

presto-mysql/src/main/java/com/facebook/presto/plugin/mysql/MySqlClient.java
Mariamalmesfer: Applied

presto-mysql/src/main/java/com/facebook/presto/plugin/mysql/MySqlClient.java
```diff
@@ -196,6 +219,73 @@ protected String toSqlType(Type type)
...
+    }
+
+    @Override
+    public Optional<ReadMapping> toPrestoType(ConnectorSession session, JdbcTypeHandle typeHandle)
+    {
+        String typeName = typeHandle.getJdbcTypeName();
+//
+//        if (typeName.equalsIgnoreCase("geometry")) {
+//            return Optional.of(geometryReadMapping());
+//        }
```
BryanCutler, line 254: This shouldn't be commented, right?

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/QueryBuilder.java
```diff
@@ -201,6 +197,38 @@ else if (typeAndValue.getType() instanceof CharType) {
...
+            String table,
+            List<JdbcColumnHandle> columns,
+            TupleDomain<ColumnHandle> tupleDomain,
+            Optional<JdbcExpression> additionalPredicate)
+            throws SQLException
+    {
+        return buildSql(client, session, connection, catalog, schema, table, columns, ImmutableMap.of(), tupleDomain, additionalPredicate);
+    }
+
+    private String addColumnExpression(List<JdbcColumnHandle> columns, Map<String, String> columnExpressions)
```
BryanCutler, line 215: just a suggestion - the name is a bit misleading because even there are no `columnExpressions` then it will still return a list of `columnAlias`, so you could just rename to `addColumns` or similar

BryanCutler (changes requested):

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/QueryBuilder.java
Mariamalmesfer: your right , I renamed it to addColumns

presto-mysql/src/main/java/com/facebook/presto/plugin/mysql/MySqlClient.java
Mariamalmesfer: right, fixed it

BryanCutler (commented):
Code looks good, but I tried to run your test locally and it doesn't pass for me. It might just be my environment, but could you confirm it's being run in CI?

Mariamalmesfer:
> Code looks good, but I tried to run your test locally and it doesn't pass for me. It might just be my environment, but could you confirm it's being run in CI?

You're right, the test failed because I missed equalsIgnoreCase when switching to forEach here: https://github.com/prestodb/presto/pull/24996#discussion_r2075812967. It's fixed now and the test passes. thanks for catching that

infvg (approved):

presto-docs/src/main/sphinx/connector/mysql.rst
```diff
@@ -166,6 +166,8 @@ The connector maps MySQL types to the corresponding PrestoDB types:
     - ``DATETIME``
   * - ``TIMESTAMP(n)``
     - ``TIMESTAMP``
+  * - ``GEOMETRY``
+    - ``VARBINARY``
```
pratyakshsharma, line 170: We are mapping it to VARCHAR. please correct this.

presto-mysql/pom.xml
```diff
@@ -65,6 +65,28 @@
...
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>com.facebook.presto</groupId>
+            <artifactId>presto-geospatial-toolkit</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>com.facebook.airlift</groupId>
+                    <artifactId>json</artifactId>
```
pratyakshsharma, line 85: the original commit has one more exclusion. Please see if we missed that somehow.

presto-mysql/src/main/java/com/facebook/presto/plugin/mysql/MySqlClient.java
```diff
@@ -173,6 +192,10 @@ protected String toSqlType(Type type)
             // In order to preserve microsecond information for TIMESTAMP_MICROSECONDS
             return "datetime(6)";
         }
+        if (GEOMETRY.equals(type.getTypeSignature().getBase())) {
+            return "geometry";
```
pratyakshsharma, line 196: Did we test reading and writing geometry types via presto? 
`toSqlType` gets invoked in the write flow while `toPrestoType` gets invoked in the read flow. Please correct me if I am wrong but the only test case written as part of this PR tries to write directly on mysql and later read the values via presto. So we are not testing write flow anywhere

pratyakshsharma (changes requested):
Thank you for raising this PR. Please confirm if we have tested write flow as well for geometry type via presto. Else the changes in toSqlType should ideally be removed.

presto-docs/src/main/sphinx/connector/mysql.rst
Mariamalmesfer: Oh, I thought since we were reading it from MySQL as WKB, it should be listed as VARBINARY. But you're right we're mapping it to VARCHAR well update the docs accordingly.

presto-mysql/pom.xml
Mariamalmesfer: double checked the original commit, and the json exclusion is the only one here nothing seems to be missing

presto-mysql/src/main/java/com/facebook/presto/plugin/mysql/MySqlClient.java
Mariamalmesfer: You ate right  the current test writes directly to MySQL and only verifies reading via Presto. Since the original PR doesn’t support writing, I’ll go ahead and remove the toSqlType() override for GEOMETRY.

presto-mysql/src/main/java/com/facebook/presto/plugin/mysql/MySqlClient.java
```diff
@@ -196,6 +214,68 @@ protected String toSqlType(Type type)
...
+
+    private static Slice stGeomFromBinary(Slice input)
+    {
+        requireNonNull(input, "input is null");
+        OGCGeometry geometry;
+        try {
+            geometry = fromBinary(input.toByteBuffer().slice());
+        }
+        catch (IllegalArgumentException | IndexOutOfBoundsException e) {
+            throw new PrestoException(INVALID_FUNCTION_ARGUMENT, "Invalid WKB", e);
```
BryanCutler, line 273: nit: please spell out WKB so the error message is more obvious

BryanCutler (dismissed):
one minor nit, otherwise looks good. Tests pass for me locally now too.

pratyakshsharma (commented):
Thank you for addressing previous comments. Please update the commit message to "Add support for geometry type in mysql connector". Changes look good otherwise.

Mariamalmesfer:
> Thank you for addressing previous comments. Please update the commit message to "Add support for geometry type in mysql connector". Changes look good otherwise.

Done

pratyakshsharma (approved):

ZacBlanco (changes requested):
Thanks for these changes @Mariamalmesfer . I have some feedback

presto-docs/src/main/sphinx/connector/mysql.rst
```diff
@@ -166,6 +166,8 @@ The connector maps MySQL types to the corresponding PrestoDB types:
     - ``DATETIME``
   * - ``TIMESTAMP(n)``
     - ``TIMESTAMP``
+  * - ``GEOMETRY``
+    - ``VARCHAR``
```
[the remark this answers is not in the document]
Mariamalmesfer: This PR only supports reading GEOMETRY, not writing. So I only added it to the MySQL to PrestoDB table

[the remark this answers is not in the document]
ZacBlanco: Thanks- that's what I assumed, but I wanted to check

presto-mysql/src/main/java/com/facebook/presto/plugin/mysql/MySqlClient.java
```diff
@@ -196,6 +214,68 @@ protected String toSqlType(Type type)
...
+
+    private static Slice stGeomFromBinary(Slice input)
+    {
+        requireNonNull(input, "input is null");
+        OGCGeometry geometry;
+        try {
+            geometry = fromBinary(input.toByteBuffer().slice());
+        }
+        catch (IllegalArgumentException | IndexOutOfBoundsException e) {
+            throw new PrestoException(INVALID_FUNCTION_ARGUMENT, "Invalid Well-Known Binary (WKB)", e);
```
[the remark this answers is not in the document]
Mariamalmesfer: WKB is a way to store shapes like points and lines in binary it throws an exception when the binary is invalid I’ve only added testMysqlGeometry so far, but I’ll add a separate test that reads invalid geometry to make sure this part is covered

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/QueryBuilder.java
```diff
@@ -201,6 +197,38 @@ else if (typeAndValue.getType() instanceof CharType) {
...
+            String table,
+            List<JdbcColumnHandle> columns,
+            TupleDomain<ColumnHandle> tupleDomain,
+            Optional<JdbcExpression> additionalPredicate)
+            throws SQLException
+    {
+        return buildSql(client, session, connection, catalog, schema, table, columns, ImmutableMap.of(), tupleDomain, additionalPredicate);
+    }
+
+    private String addColumns(List<JdbcColumnHandle> columns, Map<String, String> columnExpressions)
```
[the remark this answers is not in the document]
Mariamalmesfer: It’s needed to support custom expressions per column. For GEOMETRY, we apply ST_AsBinary(column) during reads to convert the value properly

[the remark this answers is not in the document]
ZacBlanco: Thanks, makes sense 👍

presto-base-jdbc/src/main/java/com/facebook/presto/plugin/jdbc/QueryBuilder.java
```diff
@@ -201,6 +197,38 @@ else if (typeAndValue.getType() instanceof CharType) {
...
+    {
+        if (columns.isEmpty()) {
+            return "null";
+        }
+
+        return columns.stream()
+                .map(jdbcColumnHandle -> {
+                    String columnAlias = quote(jdbcColumnHandle.getColumnName());
+                    String expression = columnExpressions.get(jdbcColumnHandle.getColumnName());
```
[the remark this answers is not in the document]
ZacBlanco: Actually looking at this again, we may not need to quote the entire expression, but we should be quoting any variables in it. That means inside of `MySqlClient`'s `buildSql` we need to quote `columnName` inside the `ST_AsBinary` expression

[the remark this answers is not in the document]
Mariamalmesfer: I added the quote in buildSql within the MySqlClient class. I also updated the code here in addColumns  can you take a look and let me know if it's better to keep it this way or revert to the old version?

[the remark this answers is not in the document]
Mariamalmesfer: @ZacBlanco 
tried quoting the column names inside ST_AsBinary using quote(), but it caused the testMysqlGeometry test to fail due to SQL syntax in the generated query (e \\\g\` )l well  switch to wrapping the column name manually with backticks instead (`columnName``) to avoid this issue. is this sounds good?

presto-mysql/src/test/java/com/facebook/presto/plugin/mysql/TestMySqlIntegrationSmokeTest.java
```diff
@@ -191,6 +192,40 @@ public void testMysqlTimestamp()
...
+
+    @Test
+    public void testInvalidGeometry()
+            throws SQLException
+    {
+        execute("CREATE TABLE tpch.test_invalid_wkb (g GEOMETRY)");
+
+        execute("INSERT INTO tpch.test_invalid_wkb VALUES (ST_GeomFromText('POINT(1 2)'))");
+
+        assertThatThrownBy(() -> execute("UPDATE tpch.test_invalid_wkb SET g = X'0001020304'"))
```
ZacBlanco, line 223: This test doesn't actually check our error path for the message from the read mapping, it's just testing the error message thrown by MySQL.

presto-mysql/src/main/java/com/facebook/presto/plugin/mysql/MySqlClient.java
```diff
@@ -196,6 +219,66 @@ protected String toSqlType(Type type)
...
+        if (typeName.equalsIgnoreCase(GEOMETRY)) {
+            return Optional.of(geometryReadMapping());
+        }
+
+        return super.toPrestoType(session, typeHandle);
+    }
+
+    private ReadMapping geometryReadMapping()
+    {
+        return sliceReadMapping(VARCHAR,
```
ZacBlanco, line 259: Let's also write a unit test for this which supplied different valid and invalid `resultSet`s to the read mapping function and verify that it is properly converted. We can create a new file `TestMySqlClient.java` for it

presto-mysql/src/main/java/com/facebook/presto/plugin/mysql/MySqlClient.java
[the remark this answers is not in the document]
ZacBlanco: Thanks for the explanation. I think since we are expecting a binary format returned, this error message is ok.

ZacBlanco (changes requested):

presto-mysql/src/test/java/com/facebook/presto/plugin/mysql/TestMySqlIntegrationSmokeTest.java
Mariamalmesfer: The new test now goes through our geometryReadMapping() logic and checks both valid and invalid WKB here https://github.com/prestodb/presto/pull/24996/files#diff-3876bab41331853047766911e32d61a58d521504873c9d4d3560fd347a76c150R35

presto-mysql/src/test/java/com/facebook/presto/plugin/mysql/TestMySqlClient.java
```diff
@@ -0,0 +1,58 @@
...
+
+import java.nio.ByteBuffer;
+
+import static com.facebook.presto.geospatial.GeoFunctions.stGeomFromBinary;
+import static com.facebook.presto.plugin.mysql.MySqlClient.getAsText;
+import static io.airlift.slice.Slices.wrappedBuffer;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.AssertionsForClassTypes.assertThatThrownBy;
+
+@Test
```
ZacBlanco, line 31: This annotation isn't necessary

presto-mysql/src/test/java/com/facebook/presto/plugin/mysql/TestMySqlClient.java
```diff
@@ -0,0 +1,58 @@
...
+{
+    @Test
+    public void testValidGeometryReadMapping()
+    {
+        OGCGeometry geometry = new OGCPoint(new Point(1.0, 2.0), null);
+        ByteBuffer buffer = geometry.asBinary();
+        byte[] wkb = new byte[buffer.remaining()];
+        buffer.get(wkb);
+
+        Slice value = getAsText(stGeomFromBinary(wrappedBuffer(wkb)));
```
ZacBlanco, line 42: You don't need to do the extra steps of creating the `wkb` byte array. Just do `Slices.wrappedBuffer(buffer)`

presto-mysql/src/test/java/com/facebook/presto/plugin/mysql/TestMySqlClient.java
```diff
@@ -0,0 +1,58 @@
...
+    {
+        byte[] invalidWkb = new byte[] {0x00, 0x01, 0x02};
+
+        assertThatThrownBy(() -> {
+            Slice result = getAsText(stGeomFromBinary(wrappedBuffer(invalidWkb)));
+        })
+                .isInstanceOf(PrestoException.class)
+                .hasMessageContaining("Invalid WKB");
+    }
+}
```
ZacBlanco, line 58: I would also like to add a test which actually uses the `readMapping` function too.

I was able to get the following code working. You should add a failing example as well.

```java
@Test
    public void testReadMapping()
            throws SQLException
    {
        SliceReadFunction fn = (SliceReadFunction) geometryReadMapping().getReadFunction();
        OGCGeometry geometry = new OGCPoint(new Point(1.0, 2.0), null);
        ByteBuffer buffer = geometry.asBinary();
        Slice value = fn.readSlice(new MockResultSet(buffer.array()), 1);
        assertThat(value.toStringUtf8()).isEqualTo("POINT (1 2)");
    }

    private static class MockResultSet
            extends SimpleResultSet
    {
        byte[] bytes;

        public MockResultSet(byte[] bytes)
        {
            this.bytes = bytes;
        }

        @Override
        public byte[] getBytes(int columnIndex)
        {
            return bytes;
        }
    }
```

You will also need to add a dependency on `com.h2database:h2` in the test scope inside of the mysql pom.xml due to use of the `SimpleResultSet` class

ZacBlanco (changes requested):

presto-mysql/src/test/java/com/facebook/presto/plugin/mysql/TestMySqlClient.java
Mariamalmesfer: Added the test with both cases and the dependency

presto-mysql/pom.xml
```diff
@@ -65,6 +65,27 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>com.facebook.presto</groupId>
+            <artifactId>presto-geospatial-toolkit</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>com.h2database</groupId>
+            <artifactId>h2</artifactId>
```
ZacBlanco, line 86: This needs to be in the test scope and moved into the test dependencies section below

ZacBlanco (changes requested):
I also just realized you are using assertion functions from the `assertj` library. We use the assertion functions from `org.testng`. Please replace your usages of assertj with our standard assertion functions.

Mariamalmesfer:
> I also just realized you are using assertion functions from the `assertj` library. We use the assertion functions from `org.testng`. Please replace your usages of assertj with our standard assertion functions.

Thanks for pointing that out. Done  Ihave updated the assertions functions to use `org.testng.Assert`.

presto-mysql/src/test/java/com/facebook/presto/plugin/mysql/TestMySqlClient.java
```diff
@@ -0,0 +1,93 @@
...
+    @Test
+    public void testInvalidGeometryReadMapping()
+    {
+        byte[] invalidWkb = new byte[]{0x00, 0x01, 0x02};
+        try {
+            getAsText(stGeomFromBinary(Slices.wrappedBuffer(invalidWkb)));
+        }
+        catch (PrestoException e) {
+            assertEquals(e.getMessage(), "Invalid WKB");
+        }
```
ZacBlanco, line 56: This test will not fail properly if it doesn't throw at all. TestNG does have an `assertThrows` method, but it doesn't let us test the message IIRC.

If we expect it to throw, put a `fail("stGeomFromBinary should throw")` on the 2nd line of the `try` block.

presto-mysql/src/test/java/com/facebook/presto/plugin/mysql/TestMySqlClient.java
```diff
@@ -0,0 +1,93 @@
...
+        Slice value = fn.readSlice(new MockResultSet(buffer.array()), 1);
+        assertEquals(value.toStringUtf8(), "POINT (1 2)");
+
+        byte[] invalid = new byte[] {0x01, 0x02, 0x03};
+        try {
+            fn.readSlice(new MockResultSet(invalid), 1);
+        }
+        catch (PrestoException e) {
+            assertEquals(e.getMessage(), "Invalid Well-Known Binary (WKB)");
+        }
```
ZacBlanco, line 74: Same issue, the test will pass if this doesn't throw. Needs a `fail()` in the try block.

presto-mysql/src/main/java/com/facebook/presto/plugin/mysql/MySqlClient.java
```diff
@@ -196,6 +217,66 @@ protected String toSqlType(Type type)
         return super.toSqlType(type);
     }
 
+    @Override
+    public PreparedStatement buildSql(ConnectorSession session, Connection connection, JdbcSplit split, List<JdbcColumnHandle> columnHandles) throws SQLException
+    {
+        Map<String, String> columnExpressions = columnHandles.stream()
+                .filter(handle -> handle.getJdbcTypeHandle().getJdbcTypeName().equalsIgnoreCase(GEOMETRY))
+                .map(handle -> handle.getColumnName())
```
ZacBlanco, line 225:
```suggestion
                .map(JdbcColumnHandle::getColumnName)
```

ZacBlanco (changes requested):

ZacBlanco (approved):
LGTM

PR #25001, opened 2025-04-29
Overload util function addPartitionAndSerializeNode

emilysun201309:
Summary:
Provide function overloading for the util function addPartitionAndSerializeNode

The new function signature: 
addPartitionAndSerializeNode(
    uint32_t numPartitions,
    const std::vector<std::string>& partitionKeys,
    const std::optional<std::vector<std::string>>& sortExpr,
    memory::MemoryPool* pool)

This takes a input of partition key names, and sort params as an expression. 
Contains the logic of translating partitionKeys -> TypedExprPtr,
 sortExpr -> sortKeys and sortOrders
The new interface will be used by PySpark-Velox shuffle.

Differential Revision:
D73696009

Privacy Context Container: L1170431


steveburnett:
Please add a release note, or NO RELEASE NOTE, as appropriate.

PR #25003, opened 2025-04-29
Pass autocommit context to connector on starting a transaction

hantangwangd:
## Description

Currently, we only have a connector-level method `Connector.isSingleStatementWritesOnly` to determine whether the entire connector supports single-statement writes exclusively, that is, whether it only permits execution within auto commit transactions.

However, some connectors have the capability not to be single-statement writes only, but certain statements (such as some non-rollbackable DDL statements) should not be placed in non-autocommit transactions. Therefore, the transaction's autocommit context needs to be passed to the connectors, allowing them to determine whether individual statements are permitted in non-autocommit transactions.

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
@@ -216,6 +216,7 @@ public abstract class IcebergAbstractMetadata
     protected Transaction transaction;
     protected final StatisticsFileCache statisticsFileCache;
     protected final IcebergTableProperties tableProperties;
+    protected final boolean autoCommitContext;
 
     private final StandardFunctionResolution functionResolution;
     private final ConcurrentMap<SchemaTableName, Table> icebergTables = new ConcurrentHashMap<>();
@@ -228,7 +229,8 @@ public abstract class IcebergAbstractMetadata
             NodeVersion nodeVersion,
             FilterStatsCalculatorService filterStatsCalculatorService,
             StatisticsFileCache statisticsFileCache,
-            IcebergTableProperties tableProperties)
+            IcebergTableProperties tableProperties,
+            boolean autoCommitContext)
     {
         this.typeManager = requireNonNull(typeManager, "typeManager is null");
         this.commitTaskCodec = requireNonNull(commitTaskCodec, "commitTaskCodec is null");
@@ -238,6 +240,7 @@ public abstract class IcebergAbstractMetadata
         this.filterStatsCalculatorService = requireNonNull(filterStatsCalculatorService, "filterStatsCalculatorService is null");
         this.statisticsFileCache = requireNonNull(statisticsFileCache, "statisticsFileCache is null");
         this.tableProperties = requireNonNull(tableProperties, "tableProperties is null");
+        this.autoCommitContext = autoCommitContext;
     }
 
     protected final Table getIcebergTable(ConnectorSession session, SchemaTableName schemaTableName)
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergConnector.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergConnector.java
@@ -184,12 +184,12 @@ public class IcebergConnector
     }
 
     @Override
-    public ConnectorTransactionHandle beginTransaction(IsolationLevel isolationLevel, boolean readOnly)
+    public ConnectorTransactionHandle beginTransaction(IsolationLevel isolationLevel, boolean autoCommitContext, boolean readOnly)
     {
         checkConnectorSupports(SERIALIZABLE, isolationLevel);
         ConnectorTransactionHandle transaction = new HiveTransactionHandle();
         try (ThreadContextClassLoader ignored = new ThreadContextClassLoader(getClass().getClassLoader())) {
-            transactionManager.put(transaction, metadataFactory.create());
+            transactionManager.put(transaction, metadataFactory.create(autoCommitContext));
         }
         return transaction;
     }
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergHiveMetadata.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergHiveMetadata.java
@@ -173,9 +173,11 @@ public class IcebergHiveMetadata
             IcebergHiveTableOperationsConfig hiveTableOeprationsConfig,
             StatisticsFileCache statisticsFileCache,
             ManifestFileCache manifestFileCache,
-            IcebergTableProperties tableProperties)
+            IcebergTableProperties tableProperties,
+            boolean autoCommitContext)
     {
-        super(typeManager, functionResolution, rowExpressionService, commitTaskCodec, nodeVersion, filterStatsCalculatorService, statisticsFileCache, tableProperties);
+        super(typeManager, functionResolution, rowExpressionService, commitTaskCodec, nodeVersion,
+                filterStatsCalculatorService, statisticsFileCache, tableProperties, autoCommitContext);
         this.metastore = requireNonNull(metastore, "metastore is null");
         this.hdfsEnvironment = requireNonNull(hdfsEnvironment, "hdfsEnvironment is null");
         this.hiveTableOeprationsConfig = requireNonNull(hiveTableOeprationsConfig, "hiveTableOperationsConfig is null");
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergHiveMetadataFactory.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergHiveMetadataFactory.java
@@ -75,6 +75,11 @@ public class IcebergHiveMetadataFactory
 
     public ConnectorMetadata create()
     {
+        return create(true);
+    }
+
+    public ConnectorMetadata create(boolean autoCommitContext)
+    {
         return new IcebergHiveMetadata(
                 metastore,
                 hdfsEnvironment,
@@ -87,6 +92,7 @@ public class IcebergHiveMetadataFactory
                 operationsConfig,
                 statisticsFileCache,
                 manifestFileCache,
-                tableProperties);
+                tableProperties,
+                autoCommitContext);
     }
 }
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergMetadataFactory.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergMetadataFactory.java
@@ -18,4 +18,6 @@ import com.facebook.presto.spi.connector.ConnectorMetadata;
 public interface IcebergMetadataFactory
 {
     ConnectorMetadata create();
+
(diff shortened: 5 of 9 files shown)


tdcmeehan:
I am a bit confused about the intent of this PR.  Could you help to provide some examples on a scenario in Iceberg where it is helpful to add this flag?  I note that the `isSingleStatementWritesOnly` flag is still true in this PR for the Iceberg connector.

I also don't see where the `autoCommitContext` is actually used in this PR.  What changes once this PR is merged?

hantangwangd:
@Tdcmeehan, thanks for the review and comment. I think the core purpose of this PR is to support connectors that allow multi-write transaction for a subset of SQL statements. Currently, for many connectors (like Iceberg), multi-write statement transaction support is typically limited to some certain DML operations, with little to no support for DDL statements. It's not an all-or-nothing proposition, so a connector-level `isSingleStatementWritesOnly` flag appears insufficient for this purpose. We need a way to let each connector determine - based on the current transaction's `autoCommitContext` flag - whether a given statement should be permitted execution within the current transaction.

Take Iceberg for example, with this `autoCommitContext` flag passed in, it can mark it's connector level flag `isSingleStatementWritesOnly` as false, and explicitly check and throw an exception when execute DDL statements like `DROP TABLE` in a transaction with flag `autoCommitContext=false`.

> I also don't see where the autoCommitContext is actually used in this PR. What changes once this PR is merged?

This PR do not really open the multi-write statement transaction for Iceberg. After this is merged, I plan to support the actual multi-statements write transaction for Iceberg in one or several follow-up PRs. What's your opinion?

tdcmeehan:
@hantangwangd can you include the Iceberg changes into this PR as a separate commit?

hantangwangd:
> can you include the Iceberg changes into this PR as a separate commit?

Sure, I'll add the change later. My thought is, in the beginning, we can allow DQL and some certain DML to run together within a transaction with `autoCommitContext=false`, under the snapshot isolation. Then, we can progressively expand the range of statements permitted in multi-write statement transactions based on the specific characteristics of each Iceberg operation. Do you think this makes sense? @tdcmeehan

hantangwangd:
Hi @tdcmeehan, I've added the code changes to support multi-statement writes transactions for Iceberg, and added the relevant design choices to the PR description. Please take a look when convenient. Any feedback would be greatly appreciated!

hantangwangd:
@sourcery-ai review

tdcmeehan:
Basic question: while some systems don't seem to distinguish snapshot from repeatable read isolation, they are separate enum values in Presto `IsolationLevel`.  I believe, for systems that distinguish then, the difference is:

1) In snapshot isolation, reads from this transaction always read the data as of the start of the transaction
2) In repeatable read isolation, reads from this transaction always read the the data as of the first read.  So this leaves open the possibility of reading writes from another transaction, so long as the first read happens after that transaction is committed.

Aren't we still using snapshot isolation then, since all scans are planned as of the sequence number of the table at the start of the transaction?

hantangwangd:
@tdcmeehan Thanks for the feedback. Actually, our implementation follows exactly the behavior of repeatable read. For each table involved in a transaction, we maintain the snapshot of when it was first loaded into the transaction context, rather than taking a global catalog snapshot or some kind of timestamp at the very beginning of the transaction. 

And in Iceberg's PR #[6984](https://github.com/apache/iceberg/pull/6948), it adopted the same implementation behavior to support its snapshot isolation level. According to their design, the behavior of reading the table data as it's first loaded, combined with some kind of validations, could support snapshot or even serializable isolation levels by adjusting the sequence in which transactions take effect.

hantangwangd:
> IcebergTransactionContext.rollback currently only clears local state – it should also abort or clean up any open Iceberg Transaction to avoid leaking resources or locks on failure paths.

Iceberg lib's table transactions do not inherently have the concept of rollback, they either commit successfully or are discarded at all. Since there are no additional resources that need to be released, Iceberg does not even provide rollback-related interfaces for its table transaction. In Presto, iceberg transactions can be considered as java objects which have the same lifecycle as `IcebergAbstractMetadata` and would be garbage-collected once the transaction ends.

hantangwangd:
@sourcery-ai resolve

steveburnett:
Please resolve the file conflicts to move this PR closer to merging.

hantangwangd:
@steveburnett thank you for the reminder. Done! Please take a look when you get a chance.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/transaction/IcebergTransactionContext.java
```diff
@@ -0,0 +1,301 @@
...
+                .orElseGet(() -> initiallyReadTable(schemaTableName)
+                        .orElseGet(() -> {
+                            Table loadTable = rawIcebergTableLoader.apply(schemaTableName);
+                            initiallyReadTables.computeIfAbsent(schemaTableName, ignored -> loadTable);
+                            return loadTable;
+                        }));
+        return new TransactionalTable(schemaTableName, table, opsFromTable(table));
+    }
+
+    public void setCallback(Runnable callback)
```
tdcmeehan: Might be good to add a precondition that this is called only once.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergAbstractMetadata.java
```diff
@@ -263,12 +263,10 @@ public abstract class IcebergAbstractMetadata
     protected final RowExpressionService rowExpressionService;
     protected final FilterStatsCalculatorService filterStatsCalculatorService;
     protected Optional<IcebergProcedureContext> procedureContext = Optional.empty();
-    protected Transaction transaction;
+    protected IcebergTransactionContext transactionContext;
```
tdcmeehan:
```suggestion
    protected final IcebergTransactionContext transactionContext;
```

presto-iceberg/src/main/java/com/facebook/presto/iceberg/transaction/IcebergTransactionContext.java
```diff
@@ -0,0 +1,301 @@
...
+import static java.util.Objects.requireNonNull;
+import static org.apache.iceberg.IcebergLibUtils.getScanContext;
+
+public class IcebergTransactionContext
+{
+    private final IsolationLevel isolationLevel;
+    private final boolean autoCommitContext;
+    private final Map<SchemaTableName, Transaction> txByTable;
+    private final Map<SchemaTableName, Table> initiallyReadTables;
+    private Optional<Runnable> callbacksOnCommit = Optional.empty();
```
tdcmeehan:
```suggestion
    private final AtomicReference<Runnable> callbacksOnCommit = Optional.empty();
```

presto-iceberg/src/main/java/com/facebook/presto/iceberg/transaction/IcebergTransactionContext.java
```diff
@@ -0,0 +1,301 @@
...
+    public Optional<Table> initiallyReadTable(SchemaTableName tableName)
+    {
+        if (initiallyReadTables.containsKey(tableName)) {
+            return Optional.ofNullable(initiallyReadTables.get(tableName));
+        }
+
+        return Optional.empty();
+    }
+
+    public void setCreateTableTransaction(SchemaTableName tableName, Transaction transaction)
```
tdcmeehan:
```suggestion
    public void registerTransaction(SchemaTableName tableName, Transaction transaction)
```

presto-iceberg/src/main/java/com/facebook/presto/iceberg/transaction/IcebergTransactionContext.java
tdcmeehan:
```suggestion
    public void registerCallback(Runnable callback)
```

presto-iceberg/src/main/java/com/facebook/presto/iceberg/transaction/IcebergTransactionContext.java
```diff
@@ -0,0 +1,301 @@
...
+                            Table loadTable = rawIcebergTableLoader.apply(schemaTableName);
+                            initiallyReadTables.computeIfAbsent(schemaTableName, ignored -> loadTable);
+                            return loadTable;
+                        }));
+        return new TransactionalTable(schemaTableName, table, opsFromTable(table));
+    }
+
+    public void setCallback(Runnable callback)
+    {
+        this.callbacksOnCommit = Optional.of(callback);
```
tdcmeehan: Should we just add a sanity check that this is only ever called once?

tdcmeehan (commented):
Looks really great.  I only have some nits.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/transaction/IcebergTransactionContext.java
hantangwangd: `callbacksOnCommit` isn't a final field–it need to be set in the below `registerCallback` method.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/transaction/IcebergTransactionContext.java
hantangwangd: Sure, I've added a precondition check to make sure that it's called only once.

hantangwangd:
@tdcmeehan thanks a lot for your review. I've addressed the comments, please take another look when you get a chance.

steveburnett:
> @steveburnett thank you for the reminder. Done! Please take a look when you get a chance.

Just getting back to this, apologies for the delay. Almost all the tests have passed, barring one that was canceled. 

Please resolve the file conflicts that have developed over the time, and then we can find how the tests are doing now!

steveburnett:
Just a ping that this PR has file conflicts needing resolution, but otherwise seems close to mergeability.

hantangwangd:
@steveburnett thanks for your reminder. I've rebased and resolved the conflicts, please take a look when you get a chance.

steveburnett:
> @steveburnett thanks for your reminder. I've rebased and resolved the conflicts, please take a look when you get a chance.

And there's another file conflict now, in IcebergAbstractMetadata.java. When you have time, please resolve this new conflict to move the PR towards mergeability.

hantangwangd:
Thanks @steveburnett. I've rebased and resolved the conflicts. Please take a look when you get a chance.

steveburnett:
> Thanks @steveburnett. I've rebased and resolved the conflicts. Please take a look when you get a chance.

Looks good! I see no file conflict, and all checks have passed.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/transaction/IcebergTransactionContext.java
hantangwangd: @tdcmeehan sorry, just noticed this now — I initially misread your suggestion as `final Optional<Runnable>`. As you recommended, using `final AtomicReference<Runnable>` is undoubtedly a better choice, and I've already fixed it. Please take a look when you have a moment. Thanks a lot!

hantangwangd:
My pleasure! Thank you so much for the thorough review. @tdcmeehan

PR #25012, opened 2025-04-29
Separate native IndexJoinRewriter from Java path

zacw7:
## Description
Index Join has a slightly different representation in the Velox query plan compared to Presto Java, as well as its execution model. It is hard to make changes to the optimizer without affecting the Java path. Therefore, this change breaks the IndexJoinRewriter into two to decouple native query planning for index join from Java.

Currently, there is one main difference:

- `IndexJoinNode` doesn't support in-line filtering on Java execution, so if the original `Join` node has a filter, it is needed to add a `FilterNode` to the query plan, while it is not necessary for Velox as the filter can be expressed as `joinConditions` in Velox `IndexLookupJoinNode`.

- In addition, if the join type is `LEFT`/`RIGHT` while a filter is present, the join is ineligible for Index Join optimization for Java, while it is still eligible for native execution.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-main-base/src/main/java/com/facebook/presto/sql/planner/optimizations/IndexJoinOptimizer.java
+++ b/presto-main-base/src/main/java/com/facebook/presto/sql/planner/optimizations/IndexJoinOptimizer.java
@@ -93,21 +93,26 @@ public class IndexJoinOptimizer
         requireNonNull(variableAllocator, "variableAllocator is null");
         requireNonNull(idAllocator, "idAllocator is null");
 
-        Rewriter rewriter = new Rewriter(idAllocator, metadata, session);
-
+        IndexJoinRewriter rewriter;
+        if (SystemSessionProperties.isNativeExecutionEnabled(session)) {
+            rewriter = new NativeRewriter(idAllocator, metadata, session);
+        }
+        else {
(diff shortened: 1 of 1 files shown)


presto-main-base/src/main/java/com/facebook/presto/sql/planner/optimizations/IndexJoinOptimizer.java
```diff
@@ -119,6 +124,54 @@ public boolean isPlanChanged()
...
+        {
+            checkArgument(probeVariables.size() == indexVariables.size());
+            ImmutableList.Builder<IndexJoinNode.EquiJoinClause> builder = ImmutableList.builder();
+            for (int i = 0; i < probeVariables.size(); i++) {
+                builder.add(new IndexJoinNode.EquiJoinClause(probeVariables.get(i), indexVariables.get(i)));
+            }
+            return builder.build();
+        }
+
+        protected static PlanNode createIndexJoinWithExpectedOutputs(
```
xiaoxmeng, line 139: Is this only used by native join? If it is, move to native join rewrite class?

xiaoxmeng (approved):
@zacw7 LGTM. Thanks!

NikhilCollooru (approved):

PR #21725, opened 2024-01-18
[native] Disable partial aggregation over system table scan

arhimondr:
## Description

Avoid pushing down aggregations to a coordinator scheduled fragment (system table scan) that runs in Java.

Partial aggregation output is not binary compatible between Presto Java and Prestissimo. 

## Motivation and Context

In https://github.com/prestodb/presto/pull/21285 an extra exchange was introduced to minimize the scope of operations run on coordinator. However past `AddExchanges` partial aggregation can be pushed down.

## Test Plan

Integration tests provided

## Contributor checklist

- [x] Please make sure your submission complies with our [development](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#development), [formatting](https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines#formatting), [commit message](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#commit-formatting-and-pull-requests), and [attribution guidelines](https://github.com/prestodb/presto/wiki/Review-and-Commit-guidelines#attribution).
- [x] PR description addresses the issue accurately and concisely.  If the change is non-trivial, a GitHub Issue is referenced.
- [x] Documented new properties (with its default value), SQL syntax, functions, or other functionality.
- [x] If release notes are required, they follow the [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines).
- [x] Adequate tests were added if applicable.
- [x] CI passed.

## Release Notes
Please follow [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines) and fill in the release notes below.

```
== NO RELEASE NOTE ==
```

PR #25041, opened 2025-05-05
Fix the partial aggregation pushdown for system table for native execution

kewang1024:
Follow up on https://github.com/prestodb/presto/pull/21725

`containsSystemTableScan` will limit all the partial aggregation generation and pushdown 
whenever we have a system table for native execution as leaf node.

When we have an aggregation on mix of empty grouping set and non-empty grouping set,
It would be forced to split into a partial aggregation and final aggregation on repartitioned 
exchange. The partial aggregation could have been pushed down under the repartitioned 
exchange but is not pushed down, due to the above constraint, leading to an invalid plan.

Fix the behavior by more precise check `directlyOnSystemTableScan` for partial aggregatio
pushdown.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeGeneralQueries.java
+++ b/presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeGeneralQueries.java
@@ -25,6 +25,8 @@ import com.facebook.presto.spi.QueryId;
 import com.facebook.presto.spi.SchemaTableName;
 import com.facebook.presto.spi.plan.SortNode;
 import com.facebook.presto.sql.analyzer.FeaturesConfig;
+import com.facebook.presto.sql.planner.Symbol;
+import com.facebook.presto.sql.planner.assertions.PlanMatchPattern;
 import com.facebook.presto.sql.planner.plan.ExchangeNode;
 import com.facebook.presto.testing.MaterializedResult;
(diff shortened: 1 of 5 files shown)


presto-native-execution/src/test/java/com/facebook/presto/nativeworker/AbstractTestNativeGeneralQueries.java
```diff
@@ -1520,89 +1521,95 @@ public void testSystemTables()
...
+                                                    PARTIAL,
+                                                    PlanMatchPattern.groupingSet(
+                                                            ImmutableList.of(ImmutableList.of("REGION_KEY"), ImmutableList.of()),
+                                                            ImmutableMap.of(),
+                                                            "groupid",
+                                                            ImmutableMap.of("regionkey$gid", expression("REGION_KEY")),
+                                                            exchange(REMOTE_STREAMING, GATHER,
+                                                                    tableScan(partitionsTableName, ImmutableMap.of("REGION_KEY", "regionkey"))))))))));
+
+//            String aggregation = format("SELECT count(*), sum(regionkey) FROM \"%s\"", partitionsTableName);
```
arhimondr, line 1550: note: remove this before merging

arhimondr (dismissed):

arhimondr (approved):

amitkdutta (approved):

PR #25051, opened 2025-05-05
[native] Export proper Task create time

spershin:
## Description
We used to put execution start time into Task creation time. Fix it.
Also remove unnecessary Task state string from a method.

```
== NO RELEASE NOTE ==
```


amitkdutta (approved):
Thanks @spershin

PR #25053, opened 2025-05-06
Refactor test framework to return `clearTransactionId` flag to client

hantangwangd:
## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* ... 
* ... 

Hive Connector Changes
* ... 
* ... 
```

```
== NO RELEASE NOTE ==
```


ZacBlanco (commented):
I think this change makes sense, but I need to spend more time thinking about if adding another flag to the client is a good way to go about it. I will keep this on the back of my mind and try to get back to this early next week

hantangwangd:
> High level question--can't the client infer that it is "clearTransactionId" by the absence of the transaction id?

@tdcmeehan Thanks for the feedback. As I understand, currently the continuity state of a multi-statements transaction in Presto is mainly maintained by the client side. The server does not return the current transaction ID in the response to every request. Instead, it only returns `startTransactionId` or `clearTransactionId` in response to `START TRANSACTION`, `COMMIT` or `ROLLBACK` statements. The client then uses these flags to determine whether it should start or end or still in a transaction, and includes the current `transactionId` when sending requests to server.

Therefore, in the testing framework, maintaining the transaction state based on `startTransactionId` and `clearTransactionId` flags returned by the server seems to be a reasonable approach to me. What's your opinion? Are there any potential alternatives we could consider?

tdcmeehan:
@hantangwangd got it.  I see that we're already sending back this information via headers, and I'm wondering if it's possible to use those existing headers without modifying the QueryResults class?

hantangwangd:
>  I see that we're already sending back this information via headers, and I'm wondering if it's possible to use those existing headers without modifying the QueryResults class?

@tdcmeehan sure, I'll figure out if we can make this work.

hantangwangd:
Hi @tdcmeehan, you are correct. We can directly retrieve the `clearTransactionId` and `startTransactionId` flags from the `StatementClient` (which is used by `AbstractTestingPrestoClient` for query execution) within the testing framework, as they are already sent back via headers.

I've modified the code to get these two flags without through `QueryResults`, thus avoiding the need to modify the `QueryResults` class. The handling logic of `clearTransactionId` and `startTransactionId` is now consistent with the existing logic for `setSessionProperties` and `resetSessionProperties`. Please take a look when available, thanks a lot!

tdcmeehan (approved):

PR #25061, opened 2025-05-06
[native] Separate Task start logic into a method.

spershin:
## Description
Separating the start of the Task into a new method.
We will use it to start queued Tasks later.

```
== NO RELEASE NOTE ==
```


aditi-pandit (approved):
Thanks @spershin

amitkdutta (approved):
Thanks @spershin

PR #25062, opened 2025-05-07
[Iceberg]Explicitly close the transactions opened manually in tests

hantangwangd:
## Description

The transaction manager in QueryRunner is a long lived instance through the whole test class scope. This PR explicitly abort the manually opened transactions in test methods, otherwise, the transaction manager will accumulate many stale un-closed transactions. For example, before this change, we will observe 47 un-closed transactions when running `TestIcebergDistributedHive`; after this change, the number is 0.

## Motivation and Context

Prevent the shared transaction manager of the whole test class from containing stale transactions that are not properly closed.

## Impact

N/A

## Test Plan

 - Make sure the change do not affect any existing CI test

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedTestBase.java
+++ b/presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedTestBase.java
@@ -1076,12 +1076,15 @@ public abstract class IcebergDistributedTestBase
         String tableName = snapshot.map(snap -> format("%s@%d", name, snap)).orElse(name);
         String qualifiedName = format("%s.%s.%s", getSession().getCatalog().get(), getSession().getSchema().get(), tableName);
         TableHandle handle = resolver.getTableHandle(QualifiedObjectName.valueOf(qualifiedName)).get();
-        return metadata.getTableStatistics(metadataSession,
+        TableStatistics tableStatistics = metadata.getTableStatistics(metadataSession,
                 handle,
                 new ArrayList<>(columns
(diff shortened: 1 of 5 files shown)


ZacBlanco (commented):
Thanks for this improvement! Instead of writing this for a bunch of test cases, do you think we can instead add an `@AfterMethod`-annotated cleanup after each tests which gets the list of open transactions, and then closes each of them there?

It would be a more reliable way to protect against this when future tests are added.

Also, is there a downside to leaving the open transactions on the tests? Just some additional memory utilization maybe?

hantangwangd:
Hi @ZacBlanco, thanks for the feedback. Yes, I believe the main downside of leaving the open transactions on the tests is that it constitutes a form of memory leak, especially considering that Presto transactions are somewhat non-lightweight concept. If not properly closed, these transactions will accumulates as the number of test cases in the class grows and could impact other test methods within the same class—for example, causing execution timeouts due to memory pressure.

> do you think we can instead add an @AfterMethod-annotated cleanup after each tests which gets the list of open transactions, and then closes each of them there?

Given that the QueryRunner's transaction manager is shared across the entire test class, and test methods may execute concurrently, I think we cannot simply clean up all active transactions in the transaction manager after one test method completes. Doing so would cause problems for other concurrently running test methods—unless the entire test class is explicitly marked with `@Test(singleThreaded = true)`. What's your opinion on this?

ZacBlanco:
Thanks for the explanation. Your approach makes sense

ZacBlanco (approved):

PR #24716, opened 2025-03-12
[native] Update prometheus metrics asynchronously

jaystarshot:
## Description

Make metric updation async to rule out any slowness in the main thread. This is to rule out any slowness due to extensive metric  use as we had seen such issue for histograms before.

## Contributor checklist

- [ ] Please make sure your submission complies with our [contributing guide](https://github.com/prestodb/presto/blob/master/CONTRIBUTING.md), in particular [code style](https://github.com/prestodb/presto/blob/master/CONTRIBUTING.md#code-style) and [commit standards](https://github.com/prestodb/presto/blob/master/CONTRIBUTING.md#commit-standards).
- [ ] PR description addresses the issue accurately and concisely.  If the change is non-trivial, a GitHub Issue is referenced.
- [ ] Documented new properties (with its default value), SQL syntax, functions, or other functionality.
- [ ] If release notes are required, they follow the [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines).
- [ ] Adequate tests were added if applicable.
- [ ] CI passed.

## Release Notes
Please follow [release notes guidelines](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines) and fill in the release notes below.

```
== NO RELEASE NOTE ==
```

PR #25065, opened 2025-05-07
[native] Add runtime-histogram-metrics-collection-enabled config

xin-zhang2:
## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* ... 
* ... 

Hive Connector Changes
* ... 
* ... 
```

```
== NO RELEASE NOTE ==
```


presto-docs/src/main/sphinx/presto_cpp/features.rst
```diff
@@ -256,3 +256,10 @@ Users can enable collection of worker level metrics by setting the property:
 
   When enabled and Presto C++ workers interact with the S3 filesystem, additional runtime metrics are collected.
   For a detailed list of these metrics, see `S3 FileSystem <https://facebookincubator.github.io/velox/monitoring/metrics.html#s3-filesystem>`_.
+
+``runtime-histogram-metrics-collection-enabled``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+* **Type:** ``boolean``
+* **Default value:** ``false``
+
+  Enable histogram metrics collection. This property is only valid when ``runtime-metrics-collection-enabled`` is set to true.
```
steveburnett, line 265:
```suggestion
  Enable histogram metrics collection. This property is only valid when ``runtime-metrics-collection-enabled`` is set to ``true``.
```

presto-docs/src/main/sphinx/presto_cpp/properties.rst
```diff
@@ -121,6 +121,14 @@ The configuration properties of Presto C++ workers are described here, in alphab
 
   Enables collection of worker level metrics.
 
+``runtime-histogram-metrics-collection-enabled``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+* **Type:** ``boolean``
+* **Default value:** ``false``
+
+  Enables collection of worker level histogram metrics. This property is only valid when
+  ``runtime-metrics-collection-enabled`` is set to true.
```
steveburnett, line 130:
```suggestion
  ``runtime-metrics-collection-enabled`` is set to ``true``.
```

steveburnett (changes requested):
Thanks for the doc! Just two formatting nits.

steveburnett (dismissed):
LGTM! (docs)

Pull updated branch, new local doc build. Thanks!

presto-native-execution/presto_cpp/main/runtime-metrics/PrometheusStatsReporter.cpp
```diff
@@ -99,6 +103,9 @@ void PrometheusStatsReporter::registerHistogramMetricExportType(
     int64_t min,
     int64_t max,
     const std::vector<int32_t>& pcts) const {
+  if (!enableHistogramMetricCollection) {
```
aditi-pandit, line 106: Do you really want to disable registering the metrics as well ? We could only disable adding metric values to keep this simple for the future.

presto-native-execution/presto_cpp/main/common/Configs.h
```diff
@@ -403,6 +403,9 @@ class SystemConfig : public ConfigBase {
   static constexpr std::string_view kEnableRuntimeMetricsCollection{
       "runtime-metrics-collection-enabled"};
 
+  static constexpr std::string_view kEnableRuntimeHistogramMetricsCollection{
```
aditi-pandit, line 406: Should we explicitly use prometheus in the same since this config is only for Prometheus stats ?

aditi-pandit (commented):
Thanks @xin-zhang2. Have few comments.

presto-native-execution/presto_cpp/main/runtime-metrics/PrometheusStatsReporter.cpp
xin-zhang2: Thanks for the review.
I’m inclined to disable both. If we only disable adding metric values, those metrics would still appear in the API output, which could be confusing — they’d show up in the response but wouldn’t actually be collected.
Do you think it would make more sense to only disable the metric collection?

presto-native-execution/presto_cpp/main/common/Configs.h
xin-zhang2: This config is only for PrometheusStatsReporter and it makes sense to use prometheus in the name. 
My thinking is that since PrometheusStatsReporter is the only implementation of StatReporter in Presto when the metrics collection is enabled, we can avoid referencing prometheus in Configs.h that is in the common directory.

aditi-pandit: @xin-zhang2 : Meta has their own implementation of metrics collection.

Though you have a valid point about the name in Configs.h. 

Lets keep it as is.

presto-native-execution/presto_cpp/main/runtime-metrics/PrometheusStatsReporter.cpp
majetideepak: We should disable registering as well. We should add a VLOG(1) that this histogram metric is disabled.

presto-docs/src/main/sphinx/presto_cpp/features.rst
```diff
@@ -256,3 +256,10 @@ Users can enable collection of worker level metrics by setting the property:
 
   When enabled and Presto C++ workers interact with the S3 filesystem, additional runtime metrics are collected.
   For a detailed list of these metrics, see `S3 FileSystem <https://facebookincubator.github.io/velox/monitoring/metrics.html#s3-filesystem>`_.
+
+``runtime-histogram-metrics-collection-enabled``
```
majetideepak, line 260: The issue is specific to prometheus. Meta's implementation does not have this issue.
Let's name this `prometheus-histogram-metrics-collection-enabled`. Move the documentation close to prometheus.

presto-docs/src/main/sphinx/presto_cpp/properties.rst
```diff
@@ -121,6 +121,14 @@ The configuration properties of Presto C++ workers are described here, in alphab
 
   Enables collection of worker level metrics.
 
+``runtime-histogram-metrics-collection-enabled``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
majetideepak, line 125: Do we not document `prometheus` here?

presto-docs/src/main/sphinx/presto_cpp/features.rst
```diff
@@ -256,3 +256,10 @@ Users can enable collection of worker level metrics by setting the property:
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+* **Type:** ``boolean``
+* **Default value:** ``false``
+
+  Enable histogram metrics collection. This property is only valid when ``runtime-metrics-collection-enabled`` is set to ``true``.
```
majetideepak, line 265: Clarify that there can be a significant performance degradation when enabled.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/PrestoNativeQueryRunnerUtils.java
```diff
@@ -571,6 +571,8 @@ public static Optional<BiFunction<Integer, URI, Process>> getExternalWorkerLaunc
                         if (enableRuntimeMetricsCollection) {
                             configProperties = format("%s%n" +
                                     "runtime-metrics-collection-enabled=true%n", configProperties);
+                            configProperties = format("%s%n" +
```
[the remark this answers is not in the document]
xin-zhang2: My thinking is that these are for non-production environemnt, so I kept the behavior unchanged from before we introduced the new config. 
Do you think it would be better to disable it? `enableRuntimeMetricsCollection` is now passed as a function parameter, and we could add another parameter to the function for the 
histogram metrics.

presto-native-execution/README.md
```diff
@@ -108,7 +108,12 @@ follow these steps:
 
 *CMake flags:* `PRESTO_STATS_REPORTER_TYPE=PROMETHEUS`
 
-*Runtime configuration:* `runtime-metrics-collection-enabled=true`
+*Runtime configuration:* 
+
+```
+runtime-metrics-collection-enabled=true
+runtime-histogram-metrics-collection-enabled=true
```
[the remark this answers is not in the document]
xin-zhang2: This doc is meant to show how to enable the metrics collection when building Prestissimo from source, so I guess it make sense to set it to true here.

presto-native-execution/presto_cpp/main/runtime-metrics/PrometheusStatsReporter.cpp
majetideepak: I am not sure if we want to enable/disable this at the session level. The downside of always enabling is that histograms will always be returned empty, and it might be confusing in certain settings. This fix is temporary. We need to fix the root cause of the performance degradation with histograms.

xin-zhang2:
Made some changes based on the review comments.
1. Renamed `runtime-histogram-metrics-collection-enabled` to `prometheus-histogram-metrics-collection-enabled`, and moved it to a new class `PrometheusConfig` under the `runtime-metrics` directory to ensure the `common` directory is clean. The `PrometheusConfig` also reads from `config.properties`.
2. Added a new parameter `enablePrometheusHistogramMetricsCollection` to `createNativeQueryRunner` function and set it to true only when `enableRuntimeMetricsCollection` is also true.
3. Removed the new config from `presto-native-execution/etc/config.properties`, so the Prometheus histogram metrics are disabled by default when runing `PrestoMain.cpp`.

Currently the registration and collection are both disabled. Not sure if we need a session property enable toggling it at runtime.

https://github.com/prestodb/presto/issues/23338 mentioned that a metric updated in the query path would degrade the performance. So I suspect we run into the same issue because kMetricTableScanBatchProcessTimeMs is updated in getOutput() of TableScan (https://github.com/facebookincubator/velox/pull/12759/files#diff-b04232f155cc15faca93b69cf34dd80e29b534a4e54dd774bf8bdbedbfc66d6bR286). 
If this is the root cause, it should be resolved by https://github.com/prestodb/presto/pull/24716.

@majetideepak @aditi-pandit

presto-native-execution/presto_cpp/main/runtime-metrics/PrometheusStatsReporter.cpp
```diff
@@ -99,6 +106,10 @@ void PrometheusStatsReporter::registerHistogramMetricExportType(
     int64_t min,
     int64_t max,
     const std::vector<int32_t>& pcts) const {
+  if (!enablePrometheusHistogramMetricCollection) {
+    VLOG(1) << "Prometheus histogram metrics collection is disabled";
```
aditi-pandit, line 110: Please add the name of the disabled metric in the log message.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/PrestoNativeQueryRunnerUtils.java
```diff
@@ -92,7 +92,7 @@ public class PrestoNativeQueryRunnerUtils
 
     private PrestoNativeQueryRunnerUtils() {}
 
-    public static QueryRunner createQueryRunner(boolean addStorageFormatToPath, boolean isCoordinatorSidecarEnabled, boolean enableRuntimeMetricsCollection, boolean enableSsdCache)
+    public static QueryRunner createQueryRunner(boolean addStorageFormatToPath, boolean isCoordinatorSidecarEnabled, boolean enableRuntimeMetricsCollection, boolean enablePrometheusHistogramMetricsCollection, boolean enableSsdCache)
```
aditi-pandit, line 95: Please wait for https://github.com/prestodb/presto/pull/25120 that adds builder style usage for QueryRunner constructor and update the methods accordingly.

presto-native-execution/presto_cpp/main/runtime-metrics/PrometheusStatsReporter.cpp
aditi-pandit: It will be good to investigate the performance degradation with histograms and be committed to fixing it. If we are not able to fix it, then it would be better to expose the histogram metrics to the user as empty and turn them on/off for debugging.

presto-native-execution/presto_cpp/main/runtime-metrics/PrometheusStatsReporter.cpp
xin-zhang2: The config will disable all histogram metrics, so it might not be necessary to list each of them individually in the log.
Currently there are 47 histogram metrics defined across presto_cpp and velox.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/PrestoNativeQueryRunnerUtils.java
aditi-pandit: @xin-zhang2 : That PR is merged. Please update this code.

xin-zhang2:
Close as the issue has been fixed by https://github.com/prestodb/presto/pull/24716.

PR #25075, opened 2025-05-08
Fix release note action with unescaped characters

ZacBlanco:
This PR fixes an error in the action when the PR had the description below

```
== NO RELEASE NOTE ==
```


czentgr (approved):
I had @mohsaka run into this in a PR which you might have seen.

The release notes section is for a test - but are these relevant but unrelated release notes? Maybe add a comment about this in the description that they are for test explicitly - it says test description but lets make sure it is understood this affects also the release notes?

ZacBlanco:
I will update the PR description to remove the test text and move the example to the original issue. I just had it here to verify the new logic would not fail

PR #25076, opened 2025-05-08
Use built-in functions for constant folding in ValuesStatsRule

pramodsatya:
## Description
When the optimizer computes plan stats for the rewritten plan, the rule `ValuesStatsRule` attempts to evaluate variables to constants in function [getVariableValues](https://github.com/prestodb/presto/blob/38f9ae6698d6cba4f513b8f1bb686e77043c074c/presto-main-base/src/main/java/com/facebook/presto/cost/ValuesStatsRule.java#L75). C++ functions should not be used to constant fold in `ValuesStatsRule` when native function namespace manager is used in sidecar until constant folding is supported by sidecar. Enable use of built-in functions for constant folding in default `RowExpressionOptimizer` and use built-in functions to evaluate constants in `ValuesStatsRule`.

## Motivation and Context
When the native CTE tests are run with the sidecar enabled, the testcase [`testPersistentCteForVarbinaryType`](https://github.com/prestodb/presto/blob/38f9ae6698d6cba4f513b8f1bb686e77043c074c/presto-hive/src/test/java/com/facebook/presto/hive/TestCteExecution.java#L277) fails with the following error:
```
java.lang.RuntimeException: RowExpression interpreter returned an unresolved expression
```
This is because the plan optimizer attempts to [evaluate CallExpressions](https://github.com/prestodb/presto/blob/38f9ae6698d6cba4f513b8f1bb686e77043c074c/presto-main-base/src/main/java/com/facebook/presto/sql/planner/RowExpressionInterpreter.java#L134C26-L134C55) in the query plan, such as `from_base64('Qm9iIEpvaG5zb24=')`, to ConstantExpressions when calculating plan stats in [`ValuesStatsRule`](https://github.com/prestodb/presto/blob/38f9ae6698d6cba4f513b8f1bb686e77043c074c/presto-main-base/src/main/java/com/facebook/presto/cost/ValuesStatsRule.java#L85). This evaluation will fail when native function implementations are used by the sidecar, since the function implementation type will be `CPP` and these functions will [not be evaluated](https://github.com/prestodb/presto/blob/38f9ae6698d6cba4f513b8f1bb686e77043c074c/presto-main-base/src/main/java/com/facebook/presto/sql/planner/RowExpressionInterpreter.java#L280) in the coordinator. `ValuesStatsRule` should use built-in functions for evaluating constant expressions until the sidecar supports expression evaluation with native engine (WIP: https://github.com/prestodb/presto/pull/24126).

## Test Plan
Please refer to [commit](https://github.com/prestodb/presto/pull/24717/commits/dc547bde61da9c0a7a452c13f9f6c0d6bc924a85), comprehensive e2e tests for function analysis with native sidecar will be added in https://github.com/prestodb/presto/pull/24717.

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-main-base/src/main/java/com/facebook/presto/cost/ValuesStatsRule.java
+++ b/presto-main-base/src/main/java/com/facebook/presto/cost/ValuesStatsRule.java
@@ -19,7 +19,9 @@ import com.facebook.presto.cost.ComposableStatsCalculator.Rule;
 import com.facebook.presto.matching.Pattern;
 import com.facebook.presto.metadata.Metadata;
 import com.facebook.presto.spi.plan.ValuesNode;
+import com.facebook.presto.spi.relation.ExpressionOptimizer;
 import com.facebook.presto.spi.relation.VariableReferenceExpression;
+import com.facebook.presto.sql.expressions.ExpressionOptimizerManager;
 import com.facebook.presto.sql.planner.TypeProvider;
 import com.facebook.presto.sql.planner.iterative.Lookup;
 
@@ -32,8 +34,8 @@ import java.util.stream.IntStream;
 
 import static com.facebook.presto.common.type.UnknownType.UNKNOWN;
 import static com.facebook.presto.cost.StatsUtil.toStatsRepresentation;
+import static com.facebook.presto.spi.relation.ExpressionOptimizer.Level.EVALUATED;
 import static com.facebook.presto.spi.statistics.SourceInfo.ConfidenceLevel.FACT;
-import static com.facebook.presto.sql.planner.RowExpressionInterpreter.evaluateConstantRowExpression;
 import static com.facebook.presto.sql.planner.plan.Patterns.values;
 import static com.google.common.collect.ImmutableList.toImmutableList;
 import static java.util.stream.Collectors.toList;
@@ -44,10 +46,12 @@ public class ValuesStatsRule
     private static final Pattern<ValuesNode> PATTERN = values();
 
     private final Metadata metadata;
+    private final ExpressionOptimizerManager expressionOptimizerManager;
 
-    public ValuesStatsRule(Metadata metadata)
+    public ValuesStatsRule(Metadata metadata, ExpressionOptimizerManager expressionOptimizerManager)
     {
         this.metadata = metadata;
+        this.expressionOptimizerManager = expressionOptimizerManager;
     }
 
     @Override
@@ -80,9 +84,10 @@ public class ValuesStatsRule
                     .mapToObj(rowId -> null)
                     .collect(toList());
         }
+        ExpressionOptimizer expressionOptimizer = expressionOptimizerManager.getExpressionOptimizer(session.toConnectorSession());
         return valuesNode.getRows().stream()
                 .map(row -> row.get(symbolId))
-                .map(rowExpression -> evaluateConstantRowExpression(rowExpression, metadata.getFunctionAndTypeManager(), session.toConnectorSession()))
+                .map(rowExpression -> expressionOptimizer.optimize(rowExpression, EVALUATED, session.toConnectorSession(), true))
                 .collect(toList());
     }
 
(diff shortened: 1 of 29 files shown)


pramodsatya:
> @pramodsatya : Thanks for debugging this.
> 
> ReplaceConstantVariableReferencesWithConstants and ValuesStatsRule are 2 independent rules imo. So I don't think we should use the config of one for the other.
> 
> ValuesStatsRule should be disabled for native execution since its relying on evaluating the constant values at the co-ordinator, when we should actually use the worker (or side-car when it supports it) . But disabling now might be too disruptive.
> 
> @rschlussel @amitkdutta : Would be great to hear your thoughts on this.

Thanks for pointing it out @aditi-pandit, I had used `rewrite_expression_with_constant_expression` property here because it appeared closest to the check required in `ValuesStatsRule`. Disabled `ValuesStatsRule` when native execution is enabled for now, and will update the check based on further feedback from the team.

presto-main-base/src/main/java/com/facebook/presto/cost/ValuesStatsRule.java
```diff
@@ -63,10 +64,12 @@ public Optional<PlanNodeStatsEstimate> calculate(ValuesNode node, StatsProvider
         statsBuilder.setOutputRowCount(node.getRows().size())
                 .setConfidence(FACT);
 
-        for (int variableId = 0; variableId < node.getOutputVariables().size(); ++variableId) {
-            VariableReferenceExpression variable = node.getOutputVariables().get(variableId);
-            List<Object> symbolValues = getVariableValues(node, variableId, session, variable.getType());
-            statsBuilder.addVariableStatistics(variable, buildVariableStatistics(symbolValues, session, variable.getType()));
+        if (!isNativeExecutionEnabled(session)) {
```
rschlussel, line 67: this is disabling for all nativeExecution, not just sidecar

rschlussel (commented):
I thought currently even if you are using the sidecar, constant evaluation and constant folding will happen in Java with the "java equivalent" functions. Is that not the case?

pramodsatya:
> I thought currently even if you are using the sidecar, constant evaluation and constant folding will happen in Java with the "java equivalent" functions. Is that not the case?

@rschlussel, yes that is mostly right. However, `ValuesStatsRule` expects constant expressions to be evaluated and typically this would happen with java function implementations. But when native function namespace manager is used via the sidecar, the function implementation type will be `CPP` and not `JAVA`, so constant expressions will not be evaluated on coordinator.

presto-main-base/src/main/java/com/facebook/presto/cost/ValuesStatsRule.java
pramodsatya: This check is just a placeholder, will revert it once the right check/fix is figured out.

tdcmeehan:
So it seems this issue is because we swap out the default function namespace to whatever is listed by the sidecar as being registered in Velox.  Without #24126, you can't constant fold at all because no function is eligible to be evaluated in the Presto coordinator (they're C++ functions and there's no mechanism to evaluate C++ functions yet).

We could probably tactically add a hacky `ExpressionOptimizer` that just swaps back the default namespace to the Java namespace, then evaluates them as usual.  This would preserve the old behavior of constant folding, while still validating overall function compatibility.  We would delete it once we complete the native expression optimizer (I don't think it would make sense to use function validation without native constant folding, once both are mature).

rschlussel:
> So it seems this issue is because we swap out the default function namespace to whatever is listed by the sidecar as being registered in Velox. Without #24126, you can't constant fold at all because no function is eligible to be evaluated in the Presto coordinator (they're C++ functions and there's no mechanism to evaluate C++ functions yet).
> 
> We could probably tactically add a hacky `ExpressionOptimizer` that just swaps back the default namespace to the Java namespace, then evaluates them as usual. This would preserve the old behavior of constant folding, while still validating overall function compatibility. We would delete it once we complete the native expression optimizer (I don't think it would make sense to use function validation without native constant folding, once both are mature).
> 
> We could also just wait for native constant folding to land.

I'd vote to add the temporary solution to support constant folding in the coordinator for now. We'd like to be able to enable sidecar in production, but no constant folding will be a deal breaker. We'd also want to do more testing for sidecar constant folding before enabling it once it lands and have some stretch of time where we can disable it without removing other sidecar features, because constant folding has the possibility of correctness bugs, so it's a riskier change to incorporate.

rschlussel:
fyi @amitkdutta

aditi-pandit:
The option to temporarily use a hacked ExpressionOptimizer that goes back to the Java namespace is definitely useful. It will help us also to turn of/off the constant folding logic without much disruption.

Given all the redesign requested for constant folding, it will take several cycles to get it right I feel.

pramodsatya:
Thanks for the feedback, will modify the ExpressionOptimizer to use Java namespace for function eval.

PR #25095, opened 2025-05-12
[native] Add Builder classes for custom PlanNodes

kevinwilfong:
## Description
In https://github.com/facebookincubator/velox/pull/13093 we added Builder classes for the PlanNode classes in Velox. This change adds Builder classes for the custom PlanNodes Presto Native provides.

The high level motivation is to make it easier and safer to rewrite an existing Velox Plan if needed, which requires reconstructing all of the PlanNodes because they are immutable. For more details so the description of the original PR in Velox.

```
== NO RELEASE NOTE ==
```


aditi-pandit (approved):
Thanks @kevinwilfong

PR #25103, opened 2025-05-13
[native] Update Velox

czentgr:
## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/.github/workflows/prestocpp-linux-adapters-build.yml
+++ b/.github/workflows/prestocpp-linux-adapters-build.yml
@@ -15,7 +15,7 @@ jobs:
   prestocpp-linux-adapters-build:
     runs-on: ubuntu-22.04
     container:
-      image: prestodb/presto-native-dependency:0.292-20250204112033-cf8ba84
+      image: ghcr.io/czentgr/prestissimo-deps:centos9
     env:
       CCACHE_DIR: "${{ github.workspace }}/ccache"
     steps:
--- a/.github/workflows/prestocpp-linux-build-and-unit-test.yml
+++ b/.github/workflows/prestocpp-linux-build-and-unit-test.yml
@@ -17,7 +17,7 @@ jobs:
   prestocpp-linux-build-for-test:
     runs-on: ubuntu-22.04
     container:
-      image: prestodb/presto-native-dependency:0.292-20250204112033-cf8ba84
+      image: ghcr.io/czentgr/prestissimo-deps:centos9
     env:
       CCACHE_DIR: "${{ github.workspace }}/ccache"
     steps:
@@ -94,7 +94,7 @@ jobs:
     needs: prestocpp-linux-build-for-test
     runs-on: ubuntu-22.04
     container:
-      image: prestodb/presto-native-dependency:0.292-20250204112033-cf8ba84
+      image: ghcr.io/czentgr/prestissimo-deps:centos9
     env:
       MAVEN_OPTS: "-Xmx4G -XX:+ExitOnOutOfMemoryError"
       MAVEN_FAST_INSTALL: "-B -V --quiet -T 1C -DskipTests -Dair.check.skip-all -Dmaven.javadoc.skip=true"
@@ -173,7 +173,7 @@ jobs:
       matrix:
         storage-format: [ "PARQUET", "DWRF" ]
     container:
-      image: prestodb/presto-native-dependency:0.292-20250204112033-cf8ba84
+      image: ghcr.io/czentgr/prestissimo-deps:centos9
     env:
       MAVEN_OPTS: "-Xmx4G -XX:+ExitOnOutOfMemoryError"
       MAVEN_FAST_INSTALL: "-B -V --quiet -T 1C -DskipTests -Dair.check.skip-all -Dmaven.javadoc.skip=true"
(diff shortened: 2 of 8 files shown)


czentgr:
The test failures for the Window test appear to done by changes in Velox prior to the FBOS upgrade.
See https://github.com/prestodb/presto/pull/25085

czentgr:
@majetideepak I investigated the macOS 13 issue. The folly build during the setup scrips seems to be failing the ` FOLLY_SSE_PREREQ(4, 2)` check to set `FOLLY_F14_CRC_INTRINSIC_AVAILABLE` to 1 (see https://github.com/facebook/folly/blob/d17bf897cb5bbf8f07b122a614e8cffdc38edcde/folly/container/detail/F14IntrinsicsAvailability.h#L51). This results in the symbol `folly::f14::detail::F14LinkCheck<(folly::f14::detail::F14IntrinsicsMode)1>::check()` to be created (the `1` is for `F14IntrinsicsMode::Simd`).

Meanwhile, when we build presto-native-execution, the folly header used will create the symbol `folly::f14::detail::F14LinkCheck<(folly::f14::detail::F14IntrinsicsMode)2>::check()` that is then later missing from the libfolly library. Therefore, ` FOLLY_SSE_PREREQ(4, 2)` must have been true when prestissimo was built. So we have an instance where the same header on the same machine results in two different results even though on the surface everything should be the same with regards to the CPU flags..

This is macOS 13 on x86 and it has AVX support so we don't set the `-msse4.2` compiler flag. I checked and you can see we set the compiler flags consistently for the setup and build. sse is never set from our scripts.

I've run a pre_FB_OS upgrade build. In that one folly DOES create the `folly::f14::detail::F14LinkCheck<(folly::f14::detail::F14IntrinsicsMode)2>::check()`
See https://github.com/prestodb/presto/actions/runs/15086222159/job/42409067932?pr=25138 (nm output at the end of the setup looking at libfolly.a).

It is not clear to me what has changed. The code around the F14IntrinsicsMode hasn't changed or at least it isn't obvious to me.
But it seems clear that when Folly looks at the SSE support in the Portability.h
```
#ifndef FOLLY_SSE
#if defined(__SSE4_2__)
#define FOLLY_SSE 4
#define FOLLY_SSE_MINOR 2
#elif defined(__SSE4_1__)
#define FOLLY_SSE 4
#define FOLLY_SSE_MINOR 1
#elif defined(__SSE4__)
#define FOLLY_SSE 4
#define FOLLY_SSE_MINOR 0
#elif defined(__SSE3__)
```

does not satisfy the check to pick the correct mode.
```
#if FOLLY_SSE_PREREQ(4, 2) || FOLLY_ARM_FEATURE_CRC32
#define FOLLY_F14_CRC_INTRINSIC_AVAILABLE 1
#else
#define FOLLY_F14_CRC_INTRINSIC_AVAILABLE 0
#endif
```
See: https://github.com/facebook/folly/blob/d17bf897cb5bbf8f07b122a614e8cffdc38edcde/folly/container/detail/F14IntrinsicsAvailability.h#L42 and following that set the mode based on flags.

But later in the build process it suddenly does? I'm very confused. I'm thinking that sse4.2. might be enabled later by one of the submodules (e.g. CPR?) so that the `if defined(__SSE4_2__)` becomes true whereas earlier during `install_folly` it was not.

I was playing around as to how to enable sse4.2 in the prestissimo build for folly to get past it. Hence the additional commits.

majetideepak:
Thanks for the investigation so far @czentgr. Very helpful.
I will follow up on a fix for macOS.

majetideepak:
@aditi-pandit one of the test output changed

```
Error:  Failures: 
Error:    TestWindowQueries>AbstractTestWindowQueries.testWindowImplicitCoercion:97->AbstractTestQueryFramework.assertQueryOrdered:246->AbstractTestQueryFramework.assertQueryOrdered:251 For query: 
 SELECT orderkey, 1e0 / row_number() OVER (ORDER BY orderkey) FROM orders LIMIT 2
 actual column types:
 [bigint, double]
expected column types:
[bigint, double]
: Lists differ at element [0]: [1, 1.0] != [2, 0.5] expected [[1, 1.0]] but found [[2, 0.5]]
```

Is it due to https://github.com/facebookincubator/velox/pull/11440?

majetideepak:
This work will be continued here https://github.com/prestodb/presto/pull/25146

aditi-pandit:
> @aditi-pandit one of the test output changed here https://github.com/prestodb/presto/actions/runs/15091938311/job/42422719751?pr=25103
> 
> ```
> Error:  Failures: 
> Error:    TestWindowQueries>AbstractTestWindowQueries.testWindowImplicitCoercion:97->AbstractTestQueryFramework.assertQueryOrdered:246->AbstractTestQueryFramework.assertQueryOrdered:251 For query: 
>  SELECT orderkey, 1e0 / row_number() OVER (ORDER BY orderkey) FROM orders LIMIT 2
>  actual column types:
>  [bigint, double]
> expected column types:
> [bigint, double]
> : Lists differ at element [0]: [1, 1.0] != [2, 0.5] expected [[1, 1.0]] but found [[2, 0.5]]
> ```
> 
> Is it due to [facebookincubator/velox#11440](https://github.com/facebookincubator/velox/pull/11440)?

@majetideepak : That's a good catch. Would be as I had changed the order of the output rows. We should change the test query to enforce an order of the output results. Do you want to do that in https://github.com/prestodb/presto/pull/25146 ?

So change the test query to SELECT orderkey, 1e0 / row_number() OVER (ORDER BY orderkey) FROM orders ORDER BY orderkey LIMIT 2

majetideepak:
Replaced by https://github.com/prestodb/presto/pull/25146

PR #25112, opened 2025-05-14
Avoid unnecessary object creation when call replaceChildren for CteConsumerNode

hantangwangd:
## Description

`CteConsumerNode` doesn't have a source node and it's method `replaceChildren` doesn't change any fields, so there is no need to create a new instance in this method.

## Motivation and Context

Avoid unnecessary object creation.

## Impact

N/A

## Test Plan

N/A

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-spi/src/main/java/com/facebook/presto/spi/plan/CteConsumerNode.java
+++ b/presto-spi/src/main/java/com/facebook/presto/spi/plan/CteConsumerNode.java
@@ -77,7 +77,7 @@ public final class CteConsumerNode
     {
         // this function expects a new instance
         checkArgument(newChildren.size() == 0, "expected newChildren to contain 0 node");
-        return new CteConsumerNode(getSourceLocation(), getId(), getStatsEquivalentPlanNode(), originalOutputVariables, cteId, originalSource);
+        return this;
     }
 
     @Override


presto-spi/src/main/java/com/facebook/presto/spi/plan/CteConsumerNode.java
```diff
@@ -77,7 +77,7 @@ public PlanNode replaceChildren(List<PlanNode> newChildren)
     {
         // this function expects a new instance
```
ZacBlanco, line 78: Curious why we have this comment then?

hantangwangd: Not very sure, but it seems to me that the `CteConsumerNode` plays a role similar to `RemoteSourceNode` or `TableScanNode`, and the `replaceChildren` function doesn't modify any of its fields, so there is no need to create a new object here. Maybe we should hear @jaystarshot's opinion on this.

hantangwangd:
Hi @jaystarshot, can you help review and share your opinion on this change? @ZacBlanco and I noticed the following comment in `replaceChildren`:

```
// this function expects a new instance
```

Is it intentional to create a new object every time? Or can we avoid creating new instance since no fields are modified?

presto-spi/src/main/java/com/facebook/presto/spi/plan/CteConsumerNode.java
jaystarshot: Don't have a strong opinion, if someone is calling replace children all other nodes create a new instance so thats why. Also shouldn't be that different since the last value will be gc ?

hantangwangd: @jaystarshot Thanks for sharing the details. Given this context, I think we can follow the same pattern as `TableScanNode` and `RemoteSourceNode` by returning `this` directly rather than creating a new object. If this object is used somewhere, which means it's reachable from GC roots, it won't be gc, so IMO there's no need to worry about the garbage collection.

I have removed the comment as well, please take a look when convenient, thanks! @jaystarshot @ZacBlanco

ZacBlanco (approved):

PR #25142, opened 2025-05-19
Extend native type rewrite for cast, function call, and simple case expression

HeidiHan0000:
This PR extends https://github.com/prestodb/presto/pull/24916 by adding support for additional cases in the NativeExecutionTypeRewrite class. Assuming enum types appear as either Cast or DereferenceExpression expressions within NativeExecutionTypeRewrite, implemented logic for rewriting these expressions into their base literal expressions.

Specifically the updates from https://github.com/prestodb/presto/pull/24916 are:
- visitCast - added support for if the expression is a custom type (cast custom type to base type)
- visitFunctionCall - added support when the argument is a custom type (and not wrapped in Cast)
   - allow rewrites for other functions other than enum_key
- visitCase - added support for rewriting cast and dereference expression arguments for visitCase

### DIFF
--- a/presto-main-base/src/main/java/com/facebook/presto/sql/rewrite/NativeExecutionTypeRewrite.java
+++ b/presto-main-base/src/main/java/com/facebook/presto/sql/rewrite/NativeExecutionTypeRewrite.java
@@ -30,6 +30,7 @@ import com.facebook.presto.sql.analyzer.SemanticException;
 import com.facebook.presto.sql.parser.SqlParser;
 import com.facebook.presto.sql.tree.ArrayConstructor;
 import com.facebook.presto.sql.tree.Cast;
+import com.facebook.presto.sql.tree.DereferenceExpression;
 import com.facebook.presto.sql.tree.Expression;
 import com.facebook.presto.sql.tree.FunctionCall;
 import com.facebook.presto.sql.tree.LongLiteral;
@@ -37,10 +38,13 @@ import com.facebook.presto.sql.tree.Node;
 import com.facebook.presto.sql.tree.NodeRef;
 import com.facebook.presto.sql.tree.Parameter;
 import com.facebook.presto.sql.tree.QualifiedName;
+import com.facebook.presto.sql.tree.SimpleCaseExpression;
 import com.facebook.presto.sql.tree.Statement;
 import com.facebook.presto.sql.tree.StringLiteral;
+import com.facebook.presto.sql.tree.WhenClause;
 import com.google.common.collect.ImmutableList;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -121,28 +125,73 @@ final class NativeExecutionTypeRewrite
             catch (IllegalArgumentException | UnknownTypeException e) {
                 throw new SemanticException(TYPE_MISMATCH, node, "Unknown type: " + node.getType());
             }
-            return node;
+
+            Expression expression = node.getExpression();
+            expression = peelIfDereferenceExpression(expression);
+            return new Cast(node.getLocation(), expression, node.getType(), node.isSafe(), node.isTypeOnly());
         }
 
         @Override
         protected Node visitFunctionCall(FunctionCall node, Void context)
         {
-            if (isValidEnumKeyFunctionCall(node)) {
-                Cast argument = (Cast) node.getArguments().get(0);
-                Type argumentType = functionAndTypeResolver.getType(parseTypeSignature(argument.getType()));
-                if (argumentType instanceof TypeWithName) {
-                    // Peel user defined type name.
-                    argumentType = ((TypeWithName) argumentType).getType();
+            QualifiedName functionName = node.getName();
+            List<Expression> arguments = node.getArguments();
+            if (node.getArguments().size() == 1) {
+                Expression argument = arguments.get(0);
+                if (isValidEnumKeyFunctionCall(node)) {
+                    functionName = QualifiedName.of(FUNCTION_ELEMENT_AT);
+                    Type argumentType;
+                    if (argument instanceof Cast) {
+                        argumentType = functionAndTypeResolver.getType(parseTypeSignature(((Cast) argument).getType()));
+                    }
+                    else if (argument instanceof DereferenceExpression) {
+                        argumentType = functionAndTypeResolver.getType(parseTypeSignature(((DereferenceExpression) argument).getBase().toString()));
+                    }
+                    else {
+                        // ENUM_KEY is only supported with Cast or DereferenceExpression for now, return node without rewriting
+                        return node;
+                    }
+                    argument = rewriteIfCastOrDereferenceExpression(argument);
+                    if (argumentType instanceof TypeWithName) {
+                        // Peel user defined type name.
+                        argumentType = ((TypeWithName) argumentType).getType();
+                        if (argumentType instanceof EnumType) {
+                            arguments = ImmutableList.of(convertEnumTypeToMapExpression(argumentType), argument);
+                        }
+                    }
                 }
-                if (argumentType instanceof EnumType) {
-                    // Convert enum_key to element_at.
-                    List<Expression> arguments = ImmutableList.of(convertEnumTypeToMapExpression(argumentType), argument.getExpression());
-                    return node.getLocation().isPresent()
-                            ? new FunctionCall(node.getLocation().get(), QualifiedName.of(FUNCTION_ELEMENT_AT), node.getWindow(), node.getFilter(), node.getOrderBy(), node.isDistinct(), node.isIgnoreNulls(), arguments)
-                            : new FunctionCall(QualifiedName.of(FUNCTION_ELEMENT_AT), node.getWindow(), node.getFilter(), node.getOrderBy(), node.isDistinct(), node.isIgnoreNulls(), arguments);
+                else {
+                    arguments.set(0, rewriteIfCastOrDereferenceExpression(argument));
                 }
             }
-            return super.visitFunctionCall(node, context);
+            return node.getLocation().isPresent()
+                ? new FunctionCall(node.getLocation().get(), functionName, node.getWindow(), node.getFilter(), node.getOrderBy(), node.isDistinct(), node.isIgnoreNulls(), arguments)
+                : new FunctionCall(functionName, node.getWindow(), node.getFilter(), node.getOrderBy(), node.isDistinct(), node.isIgnoreNulls(), arguments);
+        }
+
+        @Override
+        protected Node visitSimpleCaseExpression(SimpleCaseExpression node, Void context)
+        {
+            // SimpleCaseExpression has 3 parts: operand, whenClauses, and defaultValue.
+            Expression operand = node.getOperand();
+            List<WhenClause> whenClauses = node.getWhenClauses();
+            Optional<Expression> defaultValue = node.getDefaultValue();
+
+            // Rewrite each component
+            operand = rewriteIfCastOrDereferenceExpression(node.getOperand());
+            List<WhenClause> newWhenClauses = new ArrayList<>();
+            for (WhenClause when : whenClauses) {
+                Expression whenOperand = rewriteIfCastOrDereferenceExpression(when.getOperand());
+                Expression result = rewriteIfCastOrDereferenceExpression(when.getResult());
+                newWhenClauses.add(new WhenClause(whenOperand, result));
+            }
+            if (defaultValue.isPresent()) {
+                defaultValue = Optional.of(rewriteIfCastOrDereferenceExpression(defaultValue.get()));
+            }
+
+            return node.getLocation().isPresent()
+                ? new SimpleCaseExpression(node.getLocation().get(), operand, newWhenClauses, defaultValue)
+                : new SimpleCaseExpression(operand, newWhenClauses, defaultValue);
         }
 
         @Override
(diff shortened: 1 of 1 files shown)


presto-main-base/src/main/java/com/facebook/presto/sql/rewrite/NativeExecutionTypeRewrite.java
```diff
@@ -183,5 +231,56 @@ private Expression convertEnumTypeToMapExpression(Type type)
...
+                    Type argumentType = functionAndTypeResolver.getType(parseTypeSignature(arg.getBase().toString()));
+                    if (argumentType instanceof TypeWithName) {
+                        argumentType = ((TypeWithName) argumentType).getType();
+                        if (argumentType instanceof EnumType) {
+                            return convertEnumTypeToLiteral(arg, argumentType);
+                        }
+                    }
+                }
+                catch (IllegalArgumentException | UnknownTypeException e) {
+                    // return the original expression if rewrite fails
```
zacw7, line 279: If the comment is a sentence, it should be capitalized and end with a  period `.`
```suggestion
                    // Returns the original expression if rewrite fails.
```

zacw7 (approved):
LGTM. Thanks for adding this.

presto-main-base/src/main/java/com/facebook/presto/sql/rewrite/NativeExecutionTypeRewrite.java
HeidiHan0000: Changed!

zacw7 (approved):

presto-main-base/src/main/java/com/facebook/presto/sql/rewrite/NativeExecutionTypeRewrite.java
```diff
@@ -121,28 +125,74 @@ protected Node visitCast(Cast node, Void context)
...
+                    }
                 }
-                if (argumentType instanceof EnumType) {
-                    // Convert enum_key to element_at.
-                    List<Expression> arguments = ImmutableList.of(convertEnumTypeToMapExpression(argumentType), argument.getExpression());
-                    return node.getLocation().isPresent()
-                            ? new FunctionCall(node.getLocation().get(), QualifiedName.of(FUNCTION_ELEMENT_AT), node.getWindow(), node.getFilter(), node.getOrderBy(), node.isDistinct(), node.isIgnoreNulls(), arguments)
-                            : new FunctionCall(QualifiedName.of(FUNCTION_ELEMENT_AT), node.getWindow(), node.getFilter(), node.getOrderBy(), node.isDistinct(), node.isIgnoreNulls(), arguments);
+                else {
+                    arguments.set(0, rewriteIfCastOrDereferenceExpression(argument));
```
rschlussel, line 165: arguments will be immutable.  Need to make a new list.

presto-main-base/src/main/java/com/facebook/presto/sql/rewrite/NativeExecutionTypeRewrite.java
```diff
@@ -121,28 +125,74 @@ protected Node visitCast(Cast node, Void context)
...
+                    }
+                    else if (argument instanceof DereferenceExpression) {
+                        argumentType = functionAndTypeResolver.getType(parseTypeSignature(((DereferenceExpression) argument).getBase().toString()));
+                    }
+                    else {
+                        // ENUM_KEY is only supported with Cast or DereferenceExpression for now.
+                        // Return node without rewriting.
+                        return node;
+                    }
+                    argument = rewriteIfCastOrDereferenceExpression(argument);
```
rschlussel, line 155: instead of this custom rewriteIfCastOrdereferenceExpression, why don't we visit all the argument s

presto-main-base/src/main/java/com/facebook/presto/sql/rewrite/NativeExecutionTypeRewrite.java
```diff
@@ -121,28 +125,74 @@ protected Node visitCast(Cast node, Void context)
...
         {
-            if (isValidEnumKeyFunctionCall(node)) {
-                Cast argument = (Cast) node.getArguments().get(0);
-                Type argumentType = functionAndTypeResolver.getType(parseTypeSignature(argument.getType()));
-                if (argumentType instanceof TypeWithName) {
-                    // Peel user defined type name.
-                    argumentType = ((TypeWithName) argumentType).getType();
+            QualifiedName functionName = node.getName();
+            List<Expression> arguments = node.getArguments();
+            if (node.getArguments().size() == 1) {
```
rschlussel, line 139: why do we specifically check for 1 argument? why not visit all the arguments and rewrite as needed?

presto-main-base/src/main/java/com/facebook/presto/sql/rewrite/NativeExecutionTypeRewrite.java
```diff
@@ -121,28 +125,74 @@ protected Node visitCast(Cast node, Void context)
...
+                    functionName = QualifiedName.of(FUNCTION_ELEMENT_AT);
+                    Type argumentType;
+                    if (argument instanceof Cast) {
+                        argumentType = functionAndTypeResolver.getType(parseTypeSignature(((Cast) argument).getType()));
+                    }
+                    else if (argument instanceof DereferenceExpression) {
+                        argumentType = functionAndTypeResolver.getType(parseTypeSignature(((DereferenceExpression) argument).getBase().toString()));
+                    }
+                    else {
+                        // ENUM_KEY is only supported with Cast or DereferenceExpression for now.
```
rschlussel, line 151: why?

presto-main-base/src/main/java/com/facebook/presto/sql/rewrite/NativeExecutionTypeRewrite.java
```diff
@@ -183,5 +232,56 @@ private Expression convertEnumTypeToMapExpression(Type type)
...
+
+        private Expression rewriteIfCastOrDereferenceExpression(Expression argument)
+        {
+            if (argument instanceof Cast) {
+                argument = (Expression) visitCast((Cast) argument, null);
+            }
+            return peelIfDereferenceExpression(argument);
+        }
+
+        private Expression peelIfDereferenceExpression(Expression argument)
```
rschlussel, line 266: Are there any downsides to unwrapping the dereference expression? Can we just have this happen always? i.e. have  a visitDereferenceExpression() that handles it. (or have visitExpression call to an ExpressionTreeRewriter that handles all the expression rewrites including Dereference, Cast, SimpleCaseExpression, etc. that way we'll automatically recurse into e.g. cast contained in an argument to some other custom function no matter how many layers deep.)

presto-main-base/src/main/java/com/facebook/presto/sql/rewrite/NativeExecutionTypeRewrite.java
```diff
@@ -183,5 +232,56 @@ private Expression convertEnumTypeToMapExpression(Type type)
...
+                    DereferenceExpression arg = (DereferenceExpression) argument;
+                    Type argumentType = functionAndTypeResolver.getType(parseTypeSignature(arg.getBase().toString()));
+                    if (argumentType instanceof TypeWithName) {
+                        argumentType = ((TypeWithName) argumentType).getType();
+                        if (argumentType instanceof EnumType) {
+                            return convertEnumTypeToLiteral(arg, argumentType);
+                        }
+                    }
+                }
+                catch (IllegalArgumentException | UnknownTypeException e) {
```
rschlussel, line 279: in what cases does the rewrite fail.  Could we write the code so it doesn't happen instead of relying on exception handling, which could mask bugs.

presto-main-base/src/main/java/com/facebook/presto/sql/rewrite/NativeExecutionTypeRewrite.java
```diff
@@ -121,28 +125,74 @@ protected Node visitCast(Cast node, Void context)
...
                 }
             }
-            return super.visitFunctionCall(node, context);
+            return node.getLocation().isPresent()
+                ? new FunctionCall(node.getLocation().get(), functionName, node.getWindow(), node.getFilter(), node.getOrderBy(), node.isDistinct(), node.isIgnoreNulls(), arguments)
+                : new FunctionCall(functionName, node.getWindow(), node.getFilter(), node.getOrderBy(), node.isDistinct(), node.isIgnoreNulls(), arguments);
+        }
+
+        @Override
+        protected Node visitSimpleCaseExpression(SimpleCaseExpression node, Void context)
```
rschlussel, line 174: i think if you use an ExpressionTreeRewriter for rewriting expressions generally, you won't need to add custom logic for handling this or other expressions that might have arguments that have enum functions.  And it will work better because it can be nested as many layers deep as someone cares to write.

rschlussel (commented):
this needs tests too

presto-main-base/src/main/java/com/facebook/presto/sql/rewrite/NativeExecutionTypeRewrite.java
HeidiHan0000: I don't know if there is a reason why, but based on all the cases I've encountered, custom types were either a cast expression or dereference expression

rschlussel: That's most common, but not sure it's guaranteed. The rewriter should be able to support all query shapes. If we leave an enum_key function in, queries will fail during execution.

tdcmeehan:
Noob question, why do we rewrite the expression rather than the plan fragment?

zacw7:
> this needs tests too

agree. but it won't be trivial to set up a mock type manager which provides custom types.

presto-main-base/src/main/java/com/facebook/presto/sql/rewrite/NativeExecutionTypeRewrite.java
HeidiHan0000: I found that it can fail while getting the type (functionAndTypeResolver.getType) for other dereference expressions, for example for a named field in a row or getting some values from a CTE, as it is not recognized as a valid type. I'm not sure what else we can do if this throws other than return the original node. Would logging help?

zacw7:
> Noob question, why do we rewrite the expression tree rather than the plan fragment?

I think either way works?

zacw7 (commented):
There are 3 commits in the PR. Please either squash them or break the PR into 3. Thanks!

HeidiHan0000:
@rschlussel I added some tests, there are some cases that need to be handled (marked with TODO):

> Are there any downsides to unwrapping the dereference expression? Can we just have this happen always? 

With all the dereference expressions rewritten into their base types, operations which were not allowed between different enum types are now not throwing. Do you have any suggestions around this? I was thinking maybe something can be done to differentiate between base type and base type which has been rewritten from the enum type, and also in rewriteComparisonExpression. I will take these TODOs on a subsequent PR.

rschlussel:
> @rschlussel I added some tests, there are some cases that need to be handled (marked with TODO):
> 
> > Are there any downsides to unwrapping the dereference expression? Can we just have this happen always?
> 
> With all the dereference expressions rewritten into their base types, operations which were not allowed between different enum types are now not throwing. Do you have any suggestions around this? I was thinking maybe something can be done to differentiate between base type and base type which has been rewritten from the enum type, and also in rewriteComparisonExpression. I will take these TODOs on a subsequent PR.

can you give some examples?

presto-tests/src/test/java/com/facebook/presto/tests/TestEnumsWithNativeRewrite.java
```diff
@@ -0,0 +1,169 @@
...
+            queryRunner.createCatalog("tpch", "tpch");
+            return queryRunner;
+        }
+        catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Test
+    public void testEnumLiterals()
```
rschlussel: instead of overriding all the tests, can you break them up in TestEnums so that the unsupported stuff are in separate tests. Then we can just override the unsupported ones, but ensure that we have the same coverage otherwise..

rschlussel (commented):
for dealing with all the things that are unsupported, I think the best thing would actually be to move the rewrite from a sql rewrite to a plan rewrite. That way it will take place after type resolution is done.  You can look at SimplifyRowExpressions for an example optimizer that applies to all expression in the plan tree.

presto-tests/src/test/java/com/facebook/presto/tests/TestEnumsWithNativeRewrite.java
HeidiHan0000: Done!

HeidiHan0000:
@rschlussel 
>can you give some examples?

These are cases like comparing between 2 different enum types or between an enum value and a literal value which should throw. Since the type is rewritten into their base types, the query will get executed as if comparing between 2 base types and not throw.

> for dealing with all the things that are unsupported, I think the best thing would actually be to move the rewrite from a sql rewrite to a plan rewrite. That way it will take place after type resolution is done. You can look at SimplifyRowExpressions for an example optimizer that applies to all expression in the plan tree.

Could you give more details on where this rewrite can go so it will happen after the type resolution? I'm assuming this type resolution is the ExpressionAnalyzer where it will check if the operator is valid, is this correct? Following through with a query, it seems to go through SimplifyRowExpressions even before NativeExecutionTypeRewrite.. What does SimplifyRowExpressions do? 

Or if the type rewrite can be called from PlanBuilder.rewrite, would that work?

rschlussel:
It would get added to the list of plan optimizers in PlanOptimizers.java, which gets called after we finish analysis and build the initial query plan.

HeidiHan0000:
@rschlussel Got it, I will look into PlanOptimizers. Would it be okay if that happens on a separate PR? It seems like a major change and in this PR we've already gone through some different approaches, wanted to keep the history of them.

rschlussel (approved):

PR #25171, opened 2025-05-21
Extend filtered aggregation optimizer to support only masked partial aggregation cases

feilong-liu:
## Description
The current optimizer MergePartialAggregationsWithFilter only supports case where there are both aggregations with and without filter on the same filter. For example,
`select sum(totalprice), sum(totalprice) filter (where orderpriority='2-HIGH') from orders group by orderdate`
will work, but 
`select sum(totalprice) filter (where orderstatus='F'), sum(totalprice) filter (where orderpriority='2-HIGH') from orders group by orderdate`
will not work.
In this change, I extend it to work for the second case too.

## Motivation and Context
Extend existing optimizer to a broader range of queries. 

## Impact
Extend existing optimizer to a broader range of queries.

## Test Plan
Add unit tests

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Extend  MergePartialAggregationsWithFilter to work for queries where all aggregations have mask
```

### DIFF
--- a/presto-main-base/src/main/java/com/facebook/presto/sql/planner/optimizations/MergePartialAggregationsWithFilter.java
+++ b/presto-main-base/src/main/java/com/facebook/presto/sql/planner/optimizations/MergePartialAggregationsWithFilter.java
@@ -36,12 +36,16 @@ import com.google.common.collect.ImmutableMap;
 
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import static com.facebook.presto.SystemSessionProperties.isMergeAggregationsWithAndWithoutFilter;
+import static com.facebook.presto.expressions.LogicalRowExpressions.or;
 import static com.facebook.presto.spi.StandardErrorCode.GENERIC_INTERNAL_ERROR;
 import static com.facebook.presto.spi.plan.AggregationNode.Step.FINAL;
 import static com.facebook.presto.spi.plan.AggregationNode.Step.PARTIAL;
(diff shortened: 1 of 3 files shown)


presto-main-base/src/main/java/com/facebook/presto/sql/planner/optimizations/MergePartialAggregationsWithFilter.java
```diff
@@ -218,17 +228,32 @@ else if (node.getStep().equals(FINAL)) {
...
+                            .putAll(aggregationsAdded.entrySet().stream().collect(toImmutableMap(Map.Entry::getValue, Map.Entry::getKey))).build();
+            Map<AggregationNode.Aggregation, VariableReferenceExpression> aggregationsToMergeOutputCombined = node.getAggregations().entrySet().stream()
+                    .filter(x -> x.getValue().getMask().isPresent() && aggregationsWithoutMaskToOutputCombined.containsKey(removeFilterAndMask(x.getValue())))
+                    .collect(toImmutableMap(Map.Entry::getValue, Map.Entry::getKey));
+
+            context.get().getNewAggregationOutput().addAll(variableForNewAggregationNoMask);
+            context.get().getPartialResultToMask().putAll(aggregationsToMergeOutputCombined.entrySet().stream()
+                    .collect(toImmutableMap(Map.Entry::getValue, x -> x.getKey().getMask().get())));
+            context.get().getPartialOutputMapping().putAll(aggregationsToMergeOutputCombined.entrySet().stream()
+                    .collect(toImmutableMap(Map.Entry::getValue, x -> aggregationsWithoutMaskToOutputCombined.get(removeFilterAndMask(x.getKey())))));
```
kaikalur, line 256: Can we actually add the OR of the filters in this case?

So SUM(IF(p1, x)), SUM(IF(p2, x)) 

will have

SUM(IF(p1 or p2, x)) 

in the partial agg.

feilong-liu: Done

kaikalur (approved):

jaystarshot (approved):
I think the default case when one mask, one normal is always beneficial but this new case might not be beneficial for high cardinality cases so we should try to make the first one on by default later

kaikalur:
> I think the default case when one mask, one normal is always beneficial but this new case might not be beneficial for high cardinality cases so we should try to make the first one on by default later

But I hope adding the disjunction mitigates that? We will run some benchmarks and see.

yhwang:
Hi @feilong-liu , I am working on the release notes for this PR. Can you revise the PR note and use more user-understandable wording?

Issue #24422, opened 2025-01-23
[native] Using `setupNativeSidecarPlugin` in tests outside `presto-native-sidecar-plugin`


pramodsatya:
Helper function `setupNativeSidecarPlugin` in `presto-native-sidecar-plugin/src/test/java/com/facebook/presto/sidecar/NativeSidecarPluginQueryRunnerUtils.java` enables user to install the native sidecar plugin on the `QueryRunner` used for tests. This function is currently only used in the module `presto-native-sidecar-plugin`; attempting to use this function in a different module (`presto-native-tests` in this case) results in the following error:
```
java.lang.NoClassDefFoundError: com/facebook/presto/sidecar/NativeSidecarPlugin
	at com.facebook.presto.sidecar.NativeSidecarPluginQueryRunnerUtils.setupNativeSidecarPlugin(NativeSidecarPluginQueryRunnerUtils.java:24)
	at com.facebook.presto.nativetests.TestRepartitionQueriesWithSmallPages.createQueryRunner(TestRepartitionQueriesWithSmallPages.java:38)
...
```
 
Dependency in module's `pom.xml` (`presto-native-tests/pom.xml` in this case):
```
        <dependency>
            <groupId>com.facebook.presto</groupId>
            <artifactId>presto-native-sidecar-plugin</artifactId>
            <version>${project.version}</version>
            <scope>test</scope>
            <type>test-jar</type>
        </dependency>
```

This error came up while trying to add the native sidecar coordinator plugin to native query runner for running tests in `presto-native-tests`. 
Having a dependency on `presto-native-sidecar-plugin` with two different scopes, `test` (to access `NativeSidecarPluginQueryRunnerUtils.setupNativeSidecarPlugin`) and `compile` (to access `NativeSidecarPlugin`), appears to resolve the error. However this change throws an error with `maven-checks` Github action:
```
presto % ./mvnw install -B -V -T 1C -DskipTests -Dmaven.javadoc.skip=true --no-transfer-progress -P ci -pl '!presto-test-coverage,!:presto-docs'
...
...
[INFO] --- maven-dependency-plugin:3.1.1:analyze-only (default) @ presto-native-tests ---
[WARNING] Unused declared dependencies found:
[WARNING]    com.facebook.presto:presto-native-sidecar-plugin:jar:0.291-SNAPSHOT:compile
...
...
[...]

PR #25174, opened 2025-05-21
[native] Enable sidecar in presto-native-tests module

pdabre12:
## Description
Enable sidecar in presto-native-tests module.

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

* Add sidecar in presto-native-tests module 
```

### DIFF
--- a/presto-native-tests/pom.xml
+++ b/presto-native-tests/pom.xml
@@ -54,6 +54,26 @@
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
         </dependency>
+
+        <dependency>
+            <groupId>com.facebook.presto</groupId>
+            <artifactId>presto-common</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>com.facebook.presto</groupId>
+            <artifactId>presto-native-sidecar-plugin</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>com.facebook.presto</groupId>
+            <artifactId>presto-native-sidecar-plugin</artifactId>
+            <scope>test</scope>
+            <type>test-jar</type>
+            <version>${project.version}</version>
+        </dependency>
     </dependencies>
 
     <build>
--- a/presto-native-tests/src/test/java/com/facebook/presto/nativetests/TestDistributedEngineOnlyQueries.java
+++ b/presto-native-tests/src/test/java/com/facebook/presto/nativetests/TestDistributedEngineOnlyQueries.java
@@ -27,22 +27,34 @@ import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
 import java.util.Objects;
 
+import static com.facebook.presto.sidecar.NativeSidecarPluginQueryRunnerUtils.setupNativeSidecarPlugin;
 import static com.google.common.base.Preconditions.checkState;
+import static java.lang.Boolean.parseBoolean;
 
 public class TestDistributedEngineOnlyQueries
         extends AbstractTestEngineOnlyQueries
 {
-    private static final String timeTypeUnsupportedError = ".*Failed to parse type \\[time.*";
+    private static final String timeTypeUnsupportedErrorWithoutSidecar = ".*Failed to parse type \\[time.*";
+    private static final String timeTypeUnsupportedErrorWithSidecar = "^Unknown type time.*";
+    private String timeTypeUnsupportedError = timeTypeUnsupportedErrorWithoutSidecar;
 
-    @Parameters("storageFormat")
+    @Parameters({"storageFormat", "sidecarEnabled"})
     @Override
-    protected QueryRunner createQueryRunner() throws Exception
+    protected QueryRunner createQueryRunner()
+            throws Exception
     {
-        return PrestoNativeQueryRunnerUtils.nativeHiveQueryRunnerBuilder()
+        boolean sidecar = parseBoolean(System.getProperty("sidecarEnabled"));
+        QueryRunner queryRunner = PrestoNativeQueryRunnerUtils.nativeHiveQueryRunnerBuilder()
                 .setStorageFormat(System.getProperty("storageFormat"))
                 .setAddStorageFormatToPath(true)
                 .setUseThrift(true)
+                .setCoordinatorSidecarEnabled(sidecar)
                 .build();
+        if (sidecar) {
+            timeTypeUnsupportedError = timeTypeUnsupportedErrorWithSidecar;
+            setupNativeSidecarPlugin(queryRunner);
+        }
+        return queryRunner;
     }
 
     @Parameters("storageFormat")
--- a/presto-native-tests/src/test/java/com/facebook/presto/nativetests/TestOrderByQueries.java
+++ b/presto-native-tests/src/test/java/com/facebook/presto/nativetests/TestOrderByQueries.java
@@ -20,18 +20,27 @@ import com.facebook.presto.tests.AbstractTestOrderByQueries;
 import org.testng.annotations.Parameters;
 import org.testng.annotations.Test;
 
+import static com.facebook.presto.sidecar.NativeSidecarPluginQueryRunnerUtils.setupNativeSidecarPlugin;
+import static java.lang.Boolean.parseBoolean;
+
 public class TestOrderByQueries
         extends AbstractTestOrderByQueries
 {
-    @Parameters("storageFormat")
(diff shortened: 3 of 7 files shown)


pramodsatya:
Should we run these tests with both sidecar enabled and disabled? Maybe we can add an additional parameter for enabling sidecar? cc: @aditi-pandit

pdabre12:
@aditi-pandit Can you please have a look?

presto-native-tests/pom.xml
```diff
@@ -54,6 +54,26 @@
...
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>com.facebook.presto</groupId>
+            <artifactId>presto-native-sidecar-plugin</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
```
aditi-pandit, line 70: The 2 dependencies are for the same artifactId, can they be combined ?

presto-native-tests/src/test/java/com/facebook/presto/nativetests/TestOrderByQueries.java
```diff
@@ -20,13 +20,27 @@
...
+
 public class TestOrderByQueries
         extends AbstractTestOrderByQueries
 {
-    @Parameters("storageFormat")
+    @Parameters({"storageFormat", "sidecarEnabled"})
     @Override
     protected QueryRunner createQueryRunner() throws Exception
     {
+        if (parseBoolean(System.getProperty("sidecarEnabled"))) {
```
aditi-pandit: Something like the code below should work... Same comment for other tests as well.

```
QueryRunner queryRunner = PrestoNativeQueryRunnerUtils.nativeHiveQueryRunnerBuilder()
                .setStorageFormat(System.getProperty("storageFormat"))
                .setAddStorageFormatToPath(true)
                .setUseThrift(true);
                .setCoordinatorSidecarEnabled(sidecar)
                .build();
if (sidecar) {
    setupNativeSidecarPlugin(queryRunner);
}

return queryRunner;
```

aditi-pandit (commented):
Thanks @pdabre12. Have a minor comment.,

presto-native-tests/src/test/java/com/facebook/presto/nativetests/TestOrderByQueries.java
pdabre12: Done, PTAL.

presto-native-tests/pom.xml
pdabre12: It's required as mentioned here: https://github.com/prestodb/presto/issues/24422

steveburnett:
Thanks for the release note entry! A couple of nits to follow the phrasing in the [Order of changes](https://github.com/prestodb/presto/wiki/Release-Notes-Guidelines#order-of-changes) in the Release Notes Guidelines. 

```
== RELEASE NOTES ==

* Add sidecar in presto-native-tests module 
```

pdabre12:
@steveburnett Thanks for the suggestion.

presto-native-tests/src/test/java/com/facebook/presto/nativetests/TestDistributedEngineOnlyQueries.java
```diff
@@ -27,22 +27,34 @@
...
 
+import static com.facebook.presto.sidecar.NativeSidecarPluginQueryRunnerUtils.setupNativeSidecarPlugin;
 import static com.google.common.base.Preconditions.checkState;
+import static java.lang.Boolean.parseBoolean;
 
 public class TestDistributedEngineOnlyQueries
         extends AbstractTestEngineOnlyQueries
 {
-    private static final String timeTypeUnsupportedError = ".*Failed to parse type \\[time.*";
+    private static final String timeTypeUnsupportedErrorWithSidecarDisabled = ".*Failed to parse type \\[time.*";
```
aditi-pandit: Rename : The variables can be timeTypeUnsupportedErrorWith(out)Sidecar.

tdcmeehan (dismissed):

presto-native-tests/src/test/java/com/facebook/presto/nativetests/TestWindowQueries.java
```diff
@@ -236,4 +252,29 @@ public void testTypes()
                         "(INTERVAL '2' month, ARRAY[INTERVAL '1' month, INTERVAL '2' month]), " +
                         "(INTERVAL '5' year, ARRAY[INTERVAL '5' year])");
     }
+
+    // Todo: Refactor this test case when support for varchar(N) is added in native execution.
```
aditi-pandit, line 257: @tdcmeehan : Just want to confirm that sidecar treating varchar(n) compatible with unlimited length varchar here is intentional.

tdcmeehan: If you mean why doesn't the plan checker fail to validate this mismatch, it's because it goes through the same plan fragment rewrite in C++.  It shows up in other ways though, primarily in constant folding (parts of the plan which get optimized return unbounded varchar, vs. other parts don't get optimized and retain the bounded varchar).

aditi-pandit: There is explicit code here https://github.com/prestodb/presto/blob/master/presto-native-sidecar-plugin/src/main/java/com/facebook/presto/sidecar/typemanager/NativeTypeManager.java#L134. At the time I felt it is to continue processing of varchar(n) as varchar in functions (to avoid regressing native queries running so far). 

This test case got me thinking for a bit... The difference in the types comes from what is inferred as the type of 'foo'...Right now both sides use the same constant folding... so the result type is varchar(3) as well. The plan fragments will be the same but Velox execution will ignore this length now. So the result obtained is varchar(3) which is why expected is constructed so. But with side-car both sides return varchar(3)...Or is it something else ?

presto-native-tests/src/test/java/com/facebook/presto/nativetests/TestDistributedEngineOnlyQueries.java
pdabre12: Done.

presto-native-tests/src/test/java/com/facebook/presto/nativetests/TestWindowQueries.java
pdabre12: > But with side-car both sides return varchar(3)...Or is it something else ?

With sidecar enabled the native side will return unbounded varchar and the java side will return varchar(3).

pramodsatya: Would it be better to just modify the testcase in base class `AbstractTestWindowQueries` instead of overriding it here:
```
            MaterializedResult expected = resultBuilder(getSession(), createUnboundedVarcharType(), BIGINT)
                    .row("foo", 1L)
                    .build();
```

pdabre12: @pramodsatya I'm guessing that would cause the native execution without sidecar and the java query runner to fail because it will return a `varchar(3)`.

aditi-pandit: @pdabre12 : OIC...I see now that you have only assertQuerySucceeds with native side-car so there is a difference in types.

This test is more understandable if its written like

```
@Language("SQL") String sql = "SELECT *, rank() OVER (PARTITION BY x)\n" +
                "FROM (SELECT 'foo' x)\n" +
                "GROUP BY 1";

        MaterializedResult actual = computeActual(sql);
        MaterializedResult expected;
        if (parseBoolean(System.getProperty("sidecarEnabled"))) {
            expected = resultBuilder(getSession(), createUnboundedVarcharType(), BIGINT)
                    .row("foo", 1L)
                    .build();
        }
        else {
            expected = resultBuilder(getSession(), createVarcharType(3), BIGINT)
                    .row("foo", 1L)
                    .build();

        }
        assertEquals(actual, expected);

```

pdabre12: Done, PTAL.

aditi-pandit (approved):
@pdabre12 : Thanks for the revisions.

pdabre12:
@tdcmeehan Can you please approve again?

tdcmeehan (approved):

PR #25210, opened 2025-05-27
Add session property for passing query client timeout value

pgupta2:
Summary:
Currently, client timeout value is passed via query.client.timeout
config property which applies to all queries. Adding corresponding
session property which can be used to specify timeout values at
query level, if needed.

Differential Revision: D75480566

### DIFF
--- a/presto-docs/src/main/sphinx/admin/properties-session.rst
+++ b/presto-docs/src/main/sphinx/admin/properties-session.rst
@@ -406,3 +406,16 @@ Setting a duration controls how long to cache data.
 
 The value represents the max background fetch threads for refreshing metadata.
 
+Query Manager Properties
+------------------------
+
+``query_client_timeout``
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``Duration``
+* **Default value:** ``5m``
+
+This property can be used to configure how long a query runs without contact
+from the client application, such as the CLI, before it's abandoned.
+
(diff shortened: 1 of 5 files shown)


steveburnett:
Please document the new session property at https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/admin/properties-session.rst.

feilong-liu (dismissed):

presto-docs/src/main/sphinx/admin/properties-session.rst
```diff
@@ -93,6 +93,17 @@ be executed within a single node.
...
+``query_client_timeout``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* **Type:** ``Duration``
+* **Default value:** ``5m``
+
+This property can be used to configure how long a query runs without contact
+from the client application, such as the CLI, before its abandoned.
+
+The corresponding configuration property is :ref:`admin/properties:\`\`query.client.timeout\`\``.
```
steveburnett, line 105: This link doesn't work in a local build, see screenshot. I don't find a problem with the formatting of the link in the rst file, I believe it's because the destination - the configuration property - is also not documented. 

Would you add documentation for the configuration property in https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/admin/properties.rst? 

Screenshot 2025-05-28 at 9 55 19 AM

steveburnett (changes requested):
Thanks for the doc! One nit, and one request.

PR #25212, opened 2025-05-28
Refactor createProjectIfStatement

aaneja:
to generate `processBatch_XX` methods and invoke them instead of one giant block of `project_` invocations

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* ... 
* ... 

Hive Connector Changes
* ... 
* ... 
```

```
== NO RELEASE NOTE ==
```


presto-main-base/src/main/java/com/facebook/presto/sql/gen/CursorProcessorCompiler.java
```diff
@@ -255,31 +256,76 @@ private static IfStatement createProjectIfStatement(
...
+            Parameter properties = arg("properties", SqlFunctionProperties.class);
+            Parameter cursor = arg("cursor", RecordCursor.class);
+            Parameter pageBuilder = arg("pageBuilder", PageBuilder.class);
+            Parameter batchIndex = arg("batchIndex", int.class);
+
+            MethodDefinition batchMethod = classDefinition.declareMethod(
+                    a(PRIVATE),
+                    "processBatch_" + batchNumber,
+                    type(void.class),
+                    properties, cursor, pageBuilder, batchIndex);
```
imsayari404, line 299: Hey @aaneja , processBatch_X methods declare a batchIndex parameter, and it's passed during the call, but I couldn't find where it's being used inside the generateProjectBatchMethods loop. Is there a specific reason for including it, or is it perhaps redundant in this implementation?

aaneja: It's called here -
https://github.com/aaneja/presto2/blob/3eee6c2ced53942e68f8eef4b1b1a7f8b544df58/presto-main-base/src/main/java/com/facebook/presto/sql/gen/CursorProcessorCompiler.java#L265-L277 :

```java
 ifStatement.ifTrue()
                    .append(method.getThis())
                    .getVariable(properties)
                    .getVariable(cursor)
                    .getVariable(pageBuilder)
                    .push(batchNumber)
                    .invokeVirtual(classDefinition.getType(),
                            "processBatch_" + batchNumber,
                            type(void.class),
                            type(SqlFunctionProperties.class),
                            type(RecordCursor.class),
                            type(PageBuilder.class),
                            type(int.class));
```
Its the `.push(batchNumber)`

presto-main-base/src/main/java/com/facebook/presto/sql/gen/CursorProcessorCompiler.java
```diff
@@ -78,6 +78,7 @@
 public class CursorProcessorCompiler
         implements BodyCompiler
 {
+    private static final int PROJECT_LIST_BATCH_SIZE = 1000;
```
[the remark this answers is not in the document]
imsayari404: Got it, thanks for the clarification @aaneja

presto-verifier/src/test/java/com/facebook/presto/verifier/framework/TestDataVerification.java
```diff
@@ -427,7 +426,7 @@ public void testSelectNonStorableStructuredColumns()
     }
 
     @Test
-    public void testChecksumQueryCompilerError()
+    public void testLargeTableSelectCompiles()
```
aaneja: Maybe lets call this `testLargeTableSelectStarCompiles`

presto-main-base/src/test/java/com/facebook/presto/sql/gen/TestCursorProcessorCompiler.java
```diff
@@ -169,6 +176,145 @@ public void testCompilerWithCSE()
...
+     * the JVM constant pool size limit remains a constraint when projections contain many unique constants.
+     *
+     * This test creates projections with random constants generated via Java code, which fills up
+     * the constant pool and should still cause compilation failures even with our batching approach.
+     * This clearly shows the scope of what we're solving (MethodTooLarge) vs. what remains a JVM constraint.
+     */
+    @Test
+    public void testConstantPoolLimitStillConstrainsLargeProjections()
+    {
+        System.out.println("Testing constant pool limits with " + CONSTANT_POOL_STRESS_PROJECTION_COUNT + " projections containing random constants...");
```
aaneja: Don't use `System.out.println`, use `log.info` if you must have a message here. I don't think this or the ones on L265,266 are needed

presto-main-base/src/test/java/com/facebook/presto/sql/gen/TestCursorProcessorCompiler.java
```diff
@@ -169,6 +176,145 @@ public void testCompilerWithCSE()
...
+
+    /**
+     * Test to verify that we can handle more projections than the original failing case
+     * when they don't stress the constant pool (i.e., reuse field references).
+     */
+    @Test
+    public void testBatchingHandlesLargeProjectionCountWithFieldReferences()
+    {
+        int projectionCount = 6000;
+        System.out.println("Testing batching with " + projectionCount + " field reference projections...");
```
aaneja: Same comments on log messages as for `testConstantPoolLimitStillConstrainsLargeProjections`

presto-main-base/src/main/java/com/facebook/presto/sql/gen/CursorProcessorCompiler.java
```diff
@@ -78,6 +82,10 @@
...
+    private static final int DYNAMIC_INVOCATION_THRESHOLD = 500;
+    // New constant for projection warning threshold
```
aaneja: Remove the comment, it does not add value

presto-main-base/src/main/java/com/facebook/presto/sql/gen/CursorProcessorCompiler.java
```diff
@@ -162,6 +171,22 @@ public void generateMethods(SqlFunctionProperties sqlFunctionProperties, ClassDe
         constructorBody.ret();
     }
 
+    /**
+     * Check if the number of projections exceeds the warning threshold and log a warning.
+     * This helps customers understand when they're approaching JVM limits.
+     *
+     * @param projectionCount Number of projections in the query
+     */
+    private void checkHighProjectionCount(int projectionCount)
```
aaneja: This can be inlined, since we have only use it in `generateMethods`

presto-main-base/src/main/java/com/facebook/presto/sql/gen/CursorProcessorCompiler.java
```diff
@@ -162,6 +171,22 @@ public void generateMethods(SqlFunctionProperties sqlFunctionProperties, ClassDe
...
+    {
+        if (projectionCount >= HIGH_PROJECTION_WARNING_THRESHOLD) {
+            log.warn("Query contains %d projections, which exceeds the recommended threshold of %d. " +
```
aaneja: These warnings are not visible to the user. Use the `session.getWarningCollector().add()` method to add a PrestoWarning. This can also be moved to the call site in [LocalExecutionPlanner](https://github.com/prestodb/presto/blob/ddba4bafbd8fc1161e9cce2cf95d39b7c33031ed/presto-main-base/src/main/java/com/facebook/presto/sql/planner/LocalExecutionPlanner.java#L1517-L1523) where the session is more readily available

presto-main-base/src/main/java/com/facebook/presto/sql/gen/CursorProcessorCompiler.java
```diff
@@ -248,38 +271,105 @@ private static IfStatement createProjectIfStatement(
...
-
-            // project(block..., blockBuilder)gen
-            ifStatement.ifTrue()
-                    .invokeVirtual(classDefinition.getType(),
-                            "project_" + projectionIndex,
-                            type(void.class),
-                            type(SqlFunctionProperties.class),
-                            type(RecordCursor.class),
-                            type(BlockBuilder.class));
+        if (projections <= DYNAMIC_INVOCATION_THRESHOLD) {
```
aaneja: Can you explain what you're trying to acheive here ?

imsayari404: it's meant to choose the batching approach (generate and call processBatch_X methods) when projections ≤ 500, since that keeps the generated bytecode lighter and avoids dynamic calls.
For larger projection counts (>500), we switch to dynamic invocation using invokeMethod() in a loop to avoid hitting JVM bytecode limits like MethodTooLargeException.

presto-verifier/src/test/java/com/facebook/presto/verifier/framework/TestDataVerification.java
imsayari404: Sure

presto-main-base/src/main/java/com/facebook/presto/sql/gen/CursorProcessorCompiler.java
imsayari404: sure

presto-main-base/src/main/java/com/facebook/presto/sql/gen/CursorProcessorCompiler.java
imsayari404: sure

presto-main-base/src/main/java/com/facebook/presto/operator/project/CursorProcessor.java
```diff
@@ -19,6 +19,7 @@
 import com.facebook.presto.spi.RecordCursor;
 
 public interface CursorProcessor
+        extends CursorProcessors
```
hantangwangd: A little question: do we really need a new parent interface `CursorProcessors`? Can't we just add the new logic into `CursorProcessor`?

presto-main-base/src/main/java/com/facebook/presto/operator/project/CursorProcessors.java
```diff
@@ -0,0 +1,53 @@
...
+            void.class,
+            SqlFunctionProperties.class,
+            RecordCursor.class,
+            BlockBuilder.class);
+    CursorProcessorOutput process(SqlFunctionProperties properties, DriverYieldSignal yieldSignal, RecordCursor cursor, PageBuilder pageBuilder);
+
+    default void invokeMethod(int projectionIndex, SqlFunctionProperties properties, RecordCursor cursor, PageBuilder pageBuilder)
+    {
+        final String methodName = "project_" + projectionIndex;
+        System.out.println("Method to invoke: " + methodName);
```
hantangwangd: nit: avoid to use `System.out.println`

presto-main-base/src/main/java/com/facebook/presto/operator/project/CursorProcessors.java
```diff
@@ -0,0 +1,53 @@
...
+public interface CursorProcessors
+{
+    MethodType METHOD_TYPE = MethodType.methodType(
+            void.class,
+            SqlFunctionProperties.class,
+            RecordCursor.class,
+            BlockBuilder.class);
+    CursorProcessorOutput process(SqlFunctionProperties properties, DriverYieldSignal yieldSignal, RecordCursor cursor, PageBuilder pageBuilder);
+
+    default void invokeMethod(int projectionIndex, SqlFunctionProperties properties, RecordCursor cursor, PageBuilder pageBuilder)
```
hantangwangd: Do you think it makes sense to use a more explicit and meaningful name, such as `doProjection`? The name `invokeMethod` sounds like a general-purpose dynamic method invocation.

presto-main-base/src/main/java/com/facebook/presto/operator/project/CursorProcessors.java
```diff
@@ -0,0 +1,53 @@
...
+        System.out.println("blockBuilder to invoke: " + blockBuilder);
+        try {
+            lookup.findVirtual(this.getClass(),
+                            methodName,
+                            METHOD_TYPE)
+                    .bindTo(this)
+                    .invokeExact(properties, cursor, blockBuilder);
+        }
+        catch (Throwable e) {
+            e.printStackTrace();
```
hantangwangd: nit: avoid to use `e.printStackTrace()`

presto-main-base/src/main/java/com/facebook/presto/sql/gen/CursorProcessorCompiler.java
```diff
@@ -248,38 +259,105 @@ private static IfStatement createProjectIfStatement(
...
+            generateProjectBatchMethods(classDefinition, projections);
+
+            int batchCount = (projections + PROJECT_LIST_BATCH_SIZE - 1) / PROJECT_LIST_BATCH_SIZE;
```
hantangwangd:
```suggestion
            int batchCount = (projections + PROJECT_LIST_BATCH_SIZE - 1) / PROJECT_LIST_BATCH_SIZE;
            generateProjectBatchMethods(classDefinition, projections, batchCount);
```
Is it better to calculate `batchCount` once and pass it as a parameter to avoid another calculation in `generateProjectBatchMethods`?

presto-main-base/src/main/java/com/facebook/presto/sql/gen/CursorProcessorCompiler.java
```diff
@@ -248,38 +259,105 @@ private static IfStatement createProjectIfStatement(
...
+            ifStatement.ifTrue().append(forLoop);
         }
         return ifStatement;
     }
 
+    private static void generateProjectBatchMethods(
+            ClassDefinition classDefinition,
+            int projections)
+    {
+        int batchCount = (projections + PROJECT_LIST_BATCH_SIZE - 1) / PROJECT_LIST_BATCH_SIZE;
```
hantangwangd: As I see, the `batchCount` is always 1 based on the static constant we defined above. It seems a little weird, is this intentional?

presto-main-base/src/main/java/com/facebook/presto/sql/gen/CursorProcessorCompiler.java
aaneja: Let just do a single implementation. I am okay with the for-loop based way too. I don't think it has a performance penalty. There is no point in having a threshold IMO

presto-main-base/src/main/java/com/facebook/presto/sql/gen/CursorProcessorCompiler.java
aaneja: This was a quick poc, we should pass it as a param as suggested

presto-main-base/src/main/java/com/facebook/presto/sql/gen/CursorProcessorCompiler.java
aaneja: It would also remove the need for using the batches too, making the code easier to gork and explain. It will help if you can paste/share the decompiled source with the for-loop

presto-main-base/src/main/java/com/facebook/presto/operator/project/CursorProcessor.java
```diff
@@ -13,12 +13,41 @@
...
+    {
+        final String methodName = "project_" + projectionIndex;
+        MethodHandles.Lookup lookup = MethodHandles.lookup();
+        BlockBuilder blockBuilder = pageBuilder.getBlockBuilder(projectionIndex);
+        try {
+            lookup.findVirtual(this.getClass(),
+                            methodName,
+                            METHOD_TYPE)
+                    .bindTo(this)
+                    .invokeExact(properties, cursor, blockBuilder);
```
hantangwangd: I'm a little concerned about the performance of always using `lookup.findVirtual` directly for dynamic projection method invocations. There may be a significant performance drawback compared to directly `invokevirtual` the corresponding methods. Especially considering the `project_x` method is almost certain to be heavily invoked. Have we performed any kind of performance testing on this?

presto-main-base/src/main/java/com/facebook/presto/sql/gen/CursorProcessorCompiler.java
imsayari404: @aaneja  Thanks for the feedback! I've implemented the single for-loop approach as suggested.

presto-main-base/src/main/java/com/facebook/presto/operator/project/CursorProcessor.java
imsayari404: @hantangwangd Thank you for the performance feedback! You're right about the concern with lookup.findVirtual overhead. I've addressed this by implementing MethodHandle caching to eliminate the expensive reflection lookups on subsequent calls.
Performance testing hasn’t been done yet, but I plan to follow up on that too.

presto-main-base/src/main/java/com/facebook/presto/sql/gen/CursorProcessorCompiler.java
aaneja: Since you measure and there is a performance overhead, lets go back the original batching approach (generate and call processBatch_X methods)

imsayari404: @aaneja I've added the change for the original batching approach

hantangwangd (commented):
Thanks for the change, the batching approach looks good to me. Could you please squash the commits? It will make the code easier to understand for following readers.

imsayari404:
> Thanks for the change, the batching approach looks good to me. Could you please squash the commits? It will make the code easier to understand for following readers.

@hantangwangd Sure, I’ve squashed the commits. Let me know if there’s anything else you’d like me to address. Thanks again for the review!

PR #25233, opened 2025-06-01
Support native non-equal lookup join planning

zacw7:
This change adds an extractor to traverse the Join plan and get lookup variables in different PlanNode, which enables index look join without any equal join condition for native execution.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-main-base/src/main/java/com/facebook/presto/sql/planner/optimizations/HashGenerationOptimizer.java
+++ b/presto-main-base/src/main/java/com/facebook/presto/sql/planner/optimizations/HashGenerationOptimizer.java
@@ -69,6 +69,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.function.Function;
 
+import static com.facebook.presto.SystemSessionProperties.isNativeExecutionEnabled;
 import static com.facebook.presto.SystemSessionProperties.skipHashGenerationForJoinWithTableScanInput;
 import static com.facebook.presto.common.type.BigintType.BIGINT;
 import static com.facebook.presto.spi.plan.JoinType.INNER;
@@ -472,6 +473,13 @@ public class HashGenerationOptimizer
         @Override
         public PlanWithProperties visitIndexJoin(IndexJoinNode node, HashComputationSet parentPreference)
         {
+            if (isNativeExecutionEnabled(session)) {
+                PlanWithProperties left = planAndEnforce(node.getProbeSource(), new HashComputationSet(), true, new HashComputationSet());
+                PlanWithProperties right = planAndEnforce(node.getIndexSource(), new HashComputationSet(), true, new HashComputationSet());
+                verify(left.getHashVariables().isEmpty(), "probe side of the index join should not include hash variables");
+                verify(right.getHashVariables().isEmpty(), "build side of the index join should not include hash variables");
+                return new PlanWithProperties(replaceChildren(node, ImmutableList.of(left.getNode(), right.getNode())), ImmutableMap.of());
+            }
             List<IndexJoinNode.EquiJoinClause> clauses = node.getCriteria();
 
             // join does not pass through preferred hash variables since they take more memory and since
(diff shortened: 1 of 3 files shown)


presto-main-base/src/main/java/com/facebook/presto/sql/planner/optimizations/HashGenerationOptimizer.java
```diff
@@ -472,6 +473,13 @@ public PlanWithProperties visitSpatialJoin(SpatialJoinNode node, HashComputation
         @Override
         public PlanWithProperties visitIndexJoin(IndexJoinNode node, HashComputationSet parentPreference)
         {
+            if (isNativeExecutionEnabled(session)) {
+                PlanWithProperties left = planAndEnforce(node.getProbeSource(), new HashComputationSet(), true, new HashComputationSet());
+                PlanWithProperties right = planAndEnforce(node.getIndexSource(), new HashComputationSet(), true, new HashComputationSet());
+                verify(left.getHashVariables().isEmpty(), "probe side of the index join should not include hash variables");
```
xiaoxmeng, line 479: What's hash variable? Thanks!

presto-main-base/src/main/java/com/facebook/presto/sql/planner/optimizations/IndexJoinOptimizer.java
```diff
@@ -310,125 +315,193 @@ else if (leftIndexCandidate.isPresent()) {
...
-                Optional<PlanNode> leftIndexCandidate = IndexSourceRewriter.rewriteWithIndex(
+            Set<VariableReferenceExpression> leftLookupVariables = leftExtractorContext.getLookupVariables();
+            Set<VariableReferenceExpression> rightLookupVariables = rightExtractorContext.getLookupVariables();
+
+            // Extract non-equal join keys.
+            if (node.getFilter().isPresent()) {
+                LookupVariableExtractor.Context commonExtractorContext = new LookupVariableExtractor.Context(new HashSet<>());
+                LookupVariableExtractor.extractFromFilter(node.getFilter().get(), commonExtractorContext);
+                if (commonExtractorContext.isEligible()) {
+                    leftLookupVariables.addAll(commonExtractorContext.getLookupVariables());
```
xiaoxmeng, line 367: We add the lookup variables to both right and left lookup variables?

zacw7: Yes, because there is no easy to tell key in join filter is from which side, so we treat it as shared and add to both side's lookup variables and check which side can be resolved as index.

presto-main-base/src/main/java/com/facebook/presto/sql/planner/optimizations/HashGenerationOptimizer.java
zacw7: Can't say I fully understand the logic here. Those variables seem to be used for hash channels: https://github.com/prestodb/presto/blob/ddba4bafbd8fc1161e9cce2cf95d39b7c33031ed/presto-main-base/src/main/java/com/facebook/presto/sql/planner/LocalExecutionPlanner.java#L3535-L3543

NikhilCollooru (approved):

aditi-pandit:
@zacw7, @xiaoxmeng : This broke native tpch tests, we started getting different results. https://github.com/prestodb/presto/actions/workflows/prestocpp-linux-build-and-unit-test.yml?query=branch%3Amaster

@amitkdutta : Lets revert this PR. This needs further investigation.

Thanks @majetideepak for catching this PR.

amitkdutta:
@zacw7 Please take a look.

zacw7:
Should be fixed in https://github.com/prestodb/presto/pull/25282.

zacw7:
Thanks for reporting this issue.

Reverting: https://github.com/prestodb/presto/pull/25289

majetideepak:
Looks like the presto-cpp tests did not run since the presto-native-execution folder was not touched by this PR.
CC: @czentgr

PR #25242, opened 2025-06-03
Add support to provide thrift codec for connector specific fields

shangm2:
## Description
TODO: Clean up the json field https://github.com/prestodb/presto/issues/25671

To keep the trunk cleaner and also move this pr forward, [we had a offline discussion and agree to update this pr with the following items](https://prestodb.slack.com/archives/C091L9VBAFR/p1752721296470699?thread_ts=1752715197.618069&cid=C091L9VBAFR):

1. Keep ConnectorSplit and ConnectorTransactionHandle  in ConnectorCodecProvider.  
2. Remove Hive implementation, in particular Hive*Split and Hive*TransactionHandle  removing mixed JSON/Thrift support
3. Make ExecutionWriterTarget a Union and add support in `ConnectorCodecProvider` for related handles .  
4. Deprecate MetadataUpdate.

The original description:
1.  Add thrift support for split and transaction handle
2. but only activated if the feature toggle is on and a proper connector specific codec is provided
3. **We will be using byte array for serialization for now and will iterate on the _interface definition within SPI_ to avoid unnecessary allocations for better performance.**
4. Instructions on how to use the thrift .idl file can be found in the [rfc](https://github.com/prestodb/rfcs/pull/38/). 
Screenshot 2025-07-16 at 07 58 01
5. cpp side changes: #25595 

## Motivation and Context
https://github.com/prestodb/rfcs/pull/38

## Test Plan
1. passed verifier run

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Improve efficiency of coordinator by supporting thrift codec for connector-specific data.
```


presto-hive/src/main/java/com/facebook/presto/hive/HiveSplitCodec.java
```diff
@@ -0,0 +1,63 @@
...
+import com.facebook.drift.protocol.TMemoryBuffer;
+import com.facebook.presto.spi.ConnectorSpecificCodec;
+import com.facebook.presto.spi.ConnectorSplit;
+
+import static java.util.Objects.requireNonNull;
+
+public class HiveSplitCodec
+        implements ConnectorSpecificCodec<ConnectorSplit>
+{
+    private final ThriftCodec<HiveSplit> thriftCodec;
```
shangm2: HiveSplitCodec is a general codec which converts between hive split and a byte array.  Here, using the thrift codec is just one implementation and it could be other format as well.

presto-hive/src/main/java/com/facebook/presto/hive/HiveTransactionHandleCodec.java
```diff
@@ -0,0 +1,63 @@
...
+import com.facebook.drift.protocol.TMemoryBuffer;
+import com.facebook.presto.spi.ConnectorSpecificCodec;
+import com.facebook.presto.spi.connector.ConnectorTransactionHandle;
+
+import static java.util.Objects.requireNonNull;
+
+public class HiveTransactionHandleCodec
+        implements ConnectorSpecificCodec<ConnectorTransactionHandle>
+{
+    private final ThriftCodec<HiveTransactionHandle> thriftCodec;
```
shangm2: Similar to hive split codec above. This is a general codec for HiveTransactionHandle

presto-main-base/src/main/java/com/facebook/presto/connector/ConnectorManager.java
```diff
@@ -317,10 +321,12 @@ private synchronized void addConnectorInternal(MaterializedConnector connector)
                 .ifPresent(metadataUpdaterProvider -> connectorMetadataUpdaterManager.addMetadataUpdaterProvider(connectorId, metadataUpdaterProvider));
 
         connector.getConnectorTypeSerdeProvider()
-                        .ifPresent(
-                                connectorTypeSerdeProvider ->
+                .ifPresent(
```
shangm2: The existing connectorTypeSerdeProvider can be merged into new getConnectorSpecificCodecProvider because connectorTypeSerdeProvider currently is only being used to retrieve [connectorMetadataUpdateHandleSerde](https://github.com/prestodb/presto/blob/master/presto-spi/src/main/java/com/facebook/presto/spi/connector/ConnectorTypeSerdeProvider.java#L19)

presto-main-base/src/main/java/com/facebook/presto/server/thrift/SplitThriftCodec.java
```diff
@@ -0,0 +1,334 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+import static com.facebook.drift.annotations.ThriftField.Requiredness.NONE;
+import static com.facebook.presto.server.thrift.CustomCodecUtils.writeSingleJsonField;
+import static java.lang.String.format;
+import static java.util.Objects.requireNonNull;
+
+public class SplitThriftCodec
```
shangm2: Rename this file from "SplitCodec" to "SplitThriftCodec" because it is thrift-specific

presto-main-base/src/main/java/com/facebook/presto/server/thrift/SplitThriftCodec.java
```diff
@@ -0,0 +1,334 @@
...
+
+    private static ThriftType createSplitMetadata(ThriftCatalog catalog)
+    {
+        List<ThriftFieldMetadata> fieldMetadata = new ArrayList<>();
+
+        try {
+            fieldMetadata.add(new ThriftFieldMetadata(
+                    (short) 1,
+                    false, false, NONE, ImmutableMap.of(),
+                    new DefaultThriftTypeReference(catalog.getThriftType(ConnectorId.class)),
```
shangm2: For some fields within Split, like connector id, split context, and life span, they have already been annotated with drift annotation and is being serde as binary, not a string. We can retrieve their metadata from the catalog directly.

presto-main-base/src/main/java/com/facebook/presto/server/thrift/MetadataUpdatesCodec.java
```diff
@@ -47,7 +47,7 @@ public MetadataUpdatesCodec(JsonCodec<MetadataUpdates> jsonCodec, ThriftCatalog
     }
 
     @CodecThriftType
-    public static ThriftType getThriftType()
+    public static ThriftType getThriftType(ThriftCatalog catalog)
```
shangm2: This change allows us to retrieve metadata for annotated java class like https://github.com/prestodb/presto/pull/25242/files#r2127079231

presto-main-base/src/main/java/com/facebook/presto/server/thrift/SplitThriftCodec.java
```diff
@@ -0,0 +1,334 @@
...
+                    ImmutableList.of(),
+                    Optional.empty(),
+                    Optional.of(new ThriftMethodInjection(Split.class.getMethod("getTransactionHandle"), ImmutableList.of())),
+                    Optional.empty(),
+                    Optional.empty()));
+
+            fieldMetadata.add(new ThriftFieldMetadata(
+                    (short) 3,
+                    false, false, NONE, ImmutableMap.of(),
+                    new DefaultThriftTypeReference(ThriftType.BINARY),
```
shangm2: connectorSplit is connector-specific and its codec should convert it to a binary array.

presto-main-base/src/main/java/com/facebook/presto/server/thrift/SplitThriftCodec.java
```diff
@@ -0,0 +1,334 @@
...
+        return ThriftType.struct(new ThriftStructMetadata(
+                "Split",
+                ImmutableMap.of(),
+                Split.class, null,
+                ThriftStructMetadata.MetadataType.STRUCT,
+                Optional.empty(), ImmutableList.of(), fieldMetadata, Optional.empty(), ImmutableList.of()));
+    }
+
+    @CodecThriftType
+    public static ThriftType getThriftType(ThriftCatalog catalog)
```
shangm2: Need the catalog to retrieve metadata for annotated fields instead of manually constructing them.

presto-main-base/src/main/java/com/facebook/presto/server/thrift/SplitThriftCodec.java
```diff
@@ -0,0 +1,334 @@
...
+        SplitContext splitContext = null;
+
+        protocol.readStructBegin();
+        while (true) {
+            TField field = protocol.readFieldBegin();
+            if (field.getType() == TType.STOP || jsonSplit != null) {
+                break;
+            }
+            switch (field.getId()) {
+                case SPLIT_DATA_FIELD_ID:
```
shangm2: A special field id (in this case 1111) to tell the reader, this thrift payload is a json string wrapper for the whole split.

presto-main-base/src/main/java/com/facebook/presto/server/thrift/SplitThriftCodec.java
```diff
@@ -0,0 +1,334 @@
...
+                    break;
+                case 3:
+                    ConnectorSpecificCodec<ConnectorSplit> splitCodec = connectorSpecificCodecManager.getConnectorSplitCodec(connectorId);
+                    connectorSplit = splitCodec.deserialize(protocol.readBinary().array());
+                    break;
+                case 4:
+                    lifespan = readLifespan(protocol);
+                    break;
+                case 5:
+                    splitContext = readSplitContext(protocol);
```
shangm2: I tried to inject thrift codec for connectorid, splitcontext and life span so I can use them to deserialize the data like P1832135842. but it did not work. The task update request is not being properly deserialized.

presto-main-base/src/main/java/com/facebook/presto/server/thrift/SplitThriftCodec.java
```diff
@@ -0,0 +1,334 @@
...
+        catch (TException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Override
+    public void write(Split split, TProtocolWriter protocol)
+            throws Exception
+    {
+        if (connectorSpecificCodecManager.isConnectorSpecificCodecAvailable(split.getConnectorId())) {
```
shangm2: So far, we only have connector-specific codec for "prism". For internal connectors like tpch/tpcds/..., json will be used. As future work, we can decide if we want to fail here when thrift feature toggle is enabled but no connector-specific codec is provided.

presto-spi/src/main/java/com/facebook/presto/spi/ConnectorSpecificCodec.java
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.spi;
+
+public interface ConnectorSpecificCodec<T>
```
shangm2: A interface for all connector-specific codecs. It converts between connector-specific data and byte array.

presto-hive/src/main/java/com/facebook/presto/hive/HiveSplitCodec.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Override
+    public byte[] serialize(ConnectorSplit split)
+    {
+        HiveSplit hiveSplit = (HiveSplit) split;
+        TMemoryBuffer transport = new TMemoryBuffer(1024);
+        TBinaryProtocol protocol = new TBinaryProtocol(transport);
+        try {
+            thriftCodec.write(hiveSplit, protocol);
+            return transport.getBytes();
```
shangm2: Serialize to bytes

presto-hive/src/main/java/com/facebook/presto/hive/HiveTransactionHandleCodec.java
```diff
@@ -0,0 +1,63 @@
...
+    public HiveTransactionHandleCodec(ThriftCodecManager thriftCodecManager)
+    {
+        this.thriftCodec = requireNonNull(thriftCodecManager, "thriftCodecManager is null").getCodec(HiveTransactionHandle.class);
+    }
+
+    @Override
+    public byte[] serialize(ConnectorTransactionHandle transactionHandle)
+    {
+        HiveTransactionHandle hiveTransactionHandle = (HiveTransactionHandle) transactionHandle;
+        TMemoryBuffer transport = new TMemoryBuffer(1024);
```
arhimondr: nit: consider a helper utility method (e.g.: `<T> T fromThrift(byte[] bytes, Class<T> clazz)` and similar `toThirft ...`

presto-main-base/src/main/java/com/facebook/presto/connector/ConnectorSpecificCodecManager.java
```diff
@@ -0,0 +1,63 @@
...
+    {
+        requireNonNull(connectorId, "connectorId is null");
+        return Optional.ofNullable(connectorSpecificCodecProviders.get(connectorId)).map(ConnectorSpecificCodecProvider::getConnectorTransactionHandleCodec)
+                .orElseThrow(() -> new IllegalArgumentException("Can not find transactionHandle codec for connector: " + connectorId.getCatalogName()));
+    }
+
+    public boolean isConnectorSpecificCodecAvailable(ConnectorId connectorId)
+    {
+        requireNonNull(connectorId, "connectorId is null");
+        return connectorSpecificCodecProviders.containsKey(connectorId);
```
arhimondr: Can a certain connector provide a codec for only a subset of entities (for example only for Split but not for TransactionHandle)?

presto-spi/src/main/java/com/facebook/presto/spi/SplitContext.java
```diff
@@ -22,6 +25,7 @@
 import static java.util.Objects.requireNonNull;
 
 // TODO: Use builder pattern for SplitContext if we are to add optional field
+@ThriftStruct
```
arhimondr, line 28: Hmm, thrift annotations are already in the SPI ... Any thoughts on whether we should consider annotating the `HiveSplit` or other objects and go the handle resolves way?

arhimondr: CC: @tdcmeehan

presto-main-base/src/main/java/com/facebook/presto/server/thrift/TableWriteInfoCodec.java
```diff
@@ -47,7 +47,7 @@ public TableWriteInfoCodec(JsonCodec<TableWriteInfo> jsonCodec, ThriftCatalog th
     }
 
     @CodecThriftType
-    public static ThriftType getThriftType()
+    public static ThriftType getThriftType(ThriftCatalog catalog)
```
arhimondr: Why catalog is needed?

presto-main-base/src/main/java/com/facebook/presto/server/thrift/SplitThriftCodec.java
```diff
@@ -0,0 +1,334 @@
...
+        this.jsonCodec = requireNonNull(jsonCodec, "jsonCodec is null");
+        this.connectorSpecificCodecManager = requireNonNull(connectorSpecificCodecManager, "connectorSpecificCodecManager is null");
+    }
+
+    private static ThriftType createSplitMetadata(ThriftCatalog catalog)
+    {
+        List<ThriftFieldMetadata> fieldMetadata = new ArrayList<>();
+
+        try {
+            fieldMetadata.add(new ThriftFieldMetadata(
```
arhimondr: Writing the code manually is not ideal. It can easily go out of sync if `Split` is ever modified. Is it possible to specify a custom codec for the `ConnectorSplit` interface specifically?

presto-spi/src/main/java/com/facebook/presto/spi/connector/ConnectorSpecificCodecProvider.java
```diff
@@ -0,0 +1,24 @@
...
+
+import com.facebook.presto.spi.ConnectorSpecificCodec;
+import com.facebook.presto.spi.ConnectorSplit;
+
+public interface ConnectorSpecificCodecProvider
+{
+    ConnectorSpecificCodec<ConnectorSplit> getConnectorSplitCodec();
+
+    ConnectorSpecificCodec<ConnectorTransactionHandle> getConnectorTransactionHandleCodec();
+}
```
shangm2: .

presto-main-base/src/main/java/com/facebook/presto/connector/ConnectorSpecificCodecManager.java
shangm2: No. When a concrete connector provides its specific codec, it needs to provide implementations for both ConnectorSplit and ConnectorTransactionHandle https://github.com/prestodb/presto/pull/25242/files#r2136467778

presto-main-base/src/main/java/com/facebook/presto/server/PluginManagerUtil.java
```diff
@@ -66,6 +66,7 @@ public class PluginManagerUtil
             .add("com.facebook.drift.annotations.")
             .add("com.facebook.drift.TException")
             .add("com.facebook.drift.TApplicationException")
+            .add("com.facebook.drift.protocol.")
```
shangm2: this is to cross the boundary of different class loaders for TProtocolRead and TProtocolWriter.

presto-main-base/src/main/java/com/facebook/presto/server/thrift/TableWriteInfoCodec.java
shangm2: After updating the code, it is not needed now.

presto-main-base/src/main/java/com/facebook/presto/server/thrift/SplitThriftCodec.java
shangm2: Agree.

presto-spi/src/main/java/com/facebook/presto/spi/ConnectorSpecificThriftCodec.java
```diff
@@ -0,0 +1,24 @@
...
+
+import com.facebook.drift.protocol.TProtocolReader;
+import com.facebook.drift.protocol.TProtocolWriter;
+
+public interface ConnectorSpecificThriftCodec<T>
+{
+    void serialize(T value, TProtocolWriter writer);
+
+    T deserialize(TProtocolReader reader);
+}
```
tdcmeehan: I thought was discussed in the RFC was serialization-protocol agnostic: http://github.com/prestodb/rfcs/pull/38/files#diff-016f190cf5aeada0422f3c35f29caaca6c3ee388e562efd895013f1df97dd32bR143-R150

Why do we need to introduce the Thrift protocol in the SPI?  Why not simply make the interface to/from bytes?

arhimondr: We want to avoid extra allocations in the serialization. Maybe we can use a more generic interface (not make it Thrift specific)

tdcmeehan: Agreed that we can use an interface that avoids the extra allocations without coupling the design for connector specific serialization to Thrift.  Perhaps Input/Output streams?

presto-hive/src/main/java/com/facebook/presto/hive/HiveConnector.java
```diff
@@ -96,7 +97,8 @@ public HiveConnector(
             ConnectorPlanOptimizerProvider planOptimizerProvider,
             ConnectorMetadataUpdaterProvider metadataUpdaterProvider,
             ConnectorTypeSerdeProvider connectorTypeSerdeProvider,
-            ClassLoader classLoader)
+            ClassLoader classLoader,
+            ConnectorSpecificThriftCodecProvider connectorSpecificThriftCodecProvider)
```
arhimondr: nit: `ConnectorThriftCodecProvider`

presto-hive/src/main/java/com/facebook/presto/hive/HiveClientModule.java
```diff
@@ -179,6 +180,7 @@ public void configure(Binder binder)
         binder.bind(ConnectorPlanOptimizerProvider.class).to(HivePlanOptimizerProvider.class).in(Scopes.SINGLETON);
         binder.bind(ConnectorMetadataUpdaterProvider.class).to(HiveMetadataUpdaterProvider.class).in(Scopes.SINGLETON);
         binder.bind(ConnectorTypeSerdeProvider.class).to(HiveConnectorTypeSerdeProvider.class).in(Scopes.SINGLETON);
+        binder.bind(ConnectorSpecificThriftCodecProvider.class).to(HiveConnectorThriftCodecProvider.class).in(Scopes.SINGLETON);
```
arhimondr: nit: `HiveThriftCodecProvider`

presto-hive/src/main/java/com/facebook/presto/hive/HiveConnectorThriftCodecProvider.java
```diff
@@ -0,0 +1,48 @@
...
+    private final ThriftCodecManager thriftCodecManager;
+
+    @Inject
+    public HiveConnectorThriftCodecProvider(ThriftCodecManager thriftCodecManager)
+    {
+        this.thriftCodecManager = requireNonNull(thriftCodecManager, "thriftCodecManager is null");
+    }
+
+    @Override
+    public ConnectorSpecificThriftCodec<ConnectorSplit> getConnectorSplitCodec()
```
arhimondr: nit: `ConnectorThriftCodec`

presto-hive/src/main/java/com/facebook/presto/hive/HiveConnectorThriftCodecProvider.java
```diff
@@ -0,0 +1,48 @@
...
+    {
+        return new HiveSplitThriftCodec(thriftCodecManager);
+    }
+
+    @Override
+    public ConnectorSpecificThriftCodec<ConnectorTransactionHandle> getConnectorTransactionHandleCodec()
```
arhimondr: Should this be `Optional<ConnectoThriftCodec<...>>`? In other words do we want to allow connectors provide Thrift based serialization for only a subset of entities?

presto-hive/pom.xml
```diff
@@ -28,6 +28,12 @@
             <artifactId>drift-codec</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>com.facebook.drift</groupId>
+            <artifactId>drift-protocol</artifactId>
```
arhimondr: Consider moving `TProtocol` (and dependencies) to `drift-api`

presto-main-base/src/main/java/com/facebook/presto/connector/ConnectorManager.java
```diff
@@ -123,6 +124,7 @@ public class ConnectorManager
     private final FilterStatsCalculator filterStatsCalculator;
     private final BlockEncodingSerde blockEncodingSerde;
     private final ConnectorSystemConfig connectorSystemConfig;
+    private final ConnectorSpecificThriftCodecManager connectorSpecificThriftCodecManager;
```
arhimondr: nit: `ConnectorThriftCodecManager`

presto-main-base/src/main/java/com/facebook/presto/server/thrift/AbstractTypedThriftCodec.java
```diff
@@ -0,0 +1,197 @@
...
+                    ImmutableList.of(),
+                    Optional.empty(),
+                    Optional.of(new ThriftMethodInjection(AbstractTypedThriftCodec.class.getDeclaredMethod("getTypeField"), ImmutableList.of())),
+                    Optional.empty(),
+                    Optional.empty()));
+            fields.add(new ThriftFieldMetadata(
+                    THRIFT_FIELD_ID,
+                    false, false, Requiredness.NONE, ImmutableMap.of(),
+                    new DefaultThriftTypeReference(ThriftType.struct(new ThriftStructMetadata(
+                            baseClass.getSimpleName(),
```
shangm2: .

presto-main-base/src/main/java/com/facebook/presto/server/thrift/ConnectorSplitThriftCodec.java
```diff
@@ -0,0 +1,69 @@
...
+public class ConnectorSplitThriftCodec
+        extends AbstractTypedThriftCodec<ConnectorSplit>
+{
+    private static final ThriftType thriftType = createThriftType(ConnectorSplit.class);
+    private final ConnectorSpecificThriftCodecManager connectorSpecificThriftCodecManager;
+
+    @Inject
+    public ConnectorSplitThriftCodec(HandleResolver handleResolver, ConnectorSpecificThriftCodecManager connectorSpecificThriftCodecManager, JsonCodec<ConnectorSplit> jsonCodec)
+    {
+        super(ConnectorSplit.class,
```
shangm2, line 40: .

presto-main-base/src/main/java/com/facebook/presto/server/thrift/AbstractTypedThriftCodec.java
```diff
@@ -0,0 +1,197 @@
...
+                    Optional.of(new ThriftMethodInjection(AbstractTypedThriftCodec.class.getDeclaredMethod("getThriftField"), ImmutableList.of())),
+                    Optional.empty(),
+                    Optional.empty()));
+        }
+        catch (NoSuchMethodException e) {
+            throw new IllegalArgumentException("Failed to create ThriftFieldMetadata", e);
+        }
+
+        return ThriftType.struct(new ThriftStructMetadata(
+                baseClass.getSimpleName() + "Wrapper",
```
shangm2: .

presto-main-base/src/main/java/com/facebook/presto/connector/ConnectorManager.java
```diff
@@ -317,10 +321,12 @@ private synchronized void addConnectorInternal(MaterializedConnector connector)
                 .ifPresent(metadataUpdaterProvider -> connectorMetadataUpdaterManager.addMetadataUpdaterProvider(connectorId, metadataUpdaterProvider));
 
         connector.getConnectorTypeSerdeProvider()
-                        .ifPresent(
-                                connectorTypeSerdeProvider ->
+                .ifPresent(
+                        connectorTypeSerdeProvider ->
                                 connectorTypeSerdeManager.addConnectorTypeSerdeProvider(connectorId, connectorTypeSerdeProvider));
 
+        connector.getConnectorCodecProvider().ifPresent(connectorCodecProvider -> connectorThriftCodecManager.addConnectorThriftCodecProvider(connectorId, connectorCodecProvider));
```
shangm2: .

presto-hive/src/main/java/com/facebook/presto/hive/HiveTableLayoutHandle.java
```diff
@@ -268,7 +268,7 @@ public Object getIdentifier(Optional<ConnectorSplit> split, PlanCanonicalization
         if (split.isPresent() && (split.get() instanceof HiveSplit) && domainPredicate.getColumnDomains().isPresent()) {
             HiveSplit hiveSplit = (HiveSplit) split.get();
             Set<Subfield> subfields = hiveSplit.getRedundantColumnDomains().stream()
-                    .map(column -> new Subfield(((HiveColumnHandle) column).getName()))
+                    .map(column -> new Subfield(column.getName()))
```
arhimondr: Like the refactoring (replacing with HiveColumnHandle). Thank you for making the codebase better.

presto-main-base/src/main/java/com/facebook/presto/server/thrift/ConnectorSplitThriftCodec.java
```diff
@@ -0,0 +1,81 @@
...
+    public ThriftType getType()
+    {
+        return THRIFT_TYPE;
+    }
+
+    @Override
+    public ConnectorSplit readConcreteValue(String connectorId, TProtocolReader reader)
+            throws Exception
+    {
+        byte[] bytes = reader.readBinary().array();
```
arhimondr: `.array()` may not always work. `ByteBuffer` may have an offset and a length set to be other than `0` and `array.length`. You need to consider this or at least add an explicit assert. 

Unfortunately with the current API of `codec.deserialize(bytes)` you may need to copy if the offset is not `0` or the length is not `array().length`. A better API could be something like `byte[] array, int offset, int length`, but we keep keep it as is for now as we are planning on changing it anyways to optimize allocations.

presto-main-base/src/main/java/com/facebook/presto/server/thrift/TransactionHandleThriftCodec.java
```diff
@@ -0,0 +1,80 @@
...
+    public ThriftType getType()
+    {
+        return THRIFT_TYPE;
+    }
+
+    @Override
+    public ConnectorTransactionHandle readConcreteValue(String connectorId, TProtocolReader reader)
+            throws Exception
+    {
+        byte[] bytes = reader.readBinary().array();
```
arhimondr: ditto

arhimondr (approved):
Very well done. LGTM % comment

arhimondr:
@shangm2 could you please also make sure the `TestPrestoNative*Thrift` test suites have Thrift protocol fully enabled, there are no Thrift relates stack traces in the log and Thrift version of a test runs as fast as JSON. Some of the code paths (e.g.: taskDelete) are optional from correctness perspective, but may significantly impact latency if not implemented.

presto-spi/src/main/java/com/facebook/presto/spi/ConnectorThriftCodec.java
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.facebook.presto.spi;
+
+import com.facebook.presto.spi.api.Experimental;
+
+@Experimental
+public interface ConnectorThriftCodec<T>
```
tdcmeehan: Does it need to be called a `ThriftCodec`--can it be called `Codec`?

tdcmeehan:
Can you add instructions on how the protocol is generated for Hive?  Are there already relevant instructions on how the C++ side?

shangm2:
> Can you add instructions on how the protocol is generated for Hive? Are there already relevant instructions on how the C++ side?

Just add the command for idl generation in the description above. We will also update the rfc and readme file with it. 
And yes, we will update the rfc/readme with other instructions for cpp side as well.

presto-spi/src/main/java/com/facebook/presto/spi/ConnectorThriftCodec.java
shangm2: Make sense. I will change it within the next pr about pooled byte buffer which will update this interface anyway.

tdcmeehan: Let's do it now, we don't want to commit anything to SPI that needs to be rewritten.

tdcmeehan:
> > Can you add instructions on how the protocol is generated for Hive? Are there already relevant instructions on how the C++ side?
> 
> Just add the command for idl generation in the description above. We will also update the rfc and readme file with it. And yes, we will update the rfc/readme with other instructions for cpp side as well.

Thanks.  I noticed that we have a `createSyntheticMetadata` method which generates wrapper classes, how do those end up in the IDL?

shangm2:
> > > Can you add instructions on how the protocol is generated for Hive? Are there already relevant instructions on how the C++ side?
> > 
> > 
> > Just add the command for idl generation in the description above. We will also update the rfc and readme file with it. And yes, we will update the rfc/readme with other instructions for cpp side as well.
> 
> Thanks. I noticed that we have a `createSyntheticMetadata` method which generates wrapper classes, how do those end up in the IDL?

createSyntheticMetadata is to create wrappers which wraps json data string within a thrift payload. The reason we want to do this is some data fields are not on the critical path (e.g. datasize is small ) but requires a lot of effort to fully migrate to thrift. Then we wrap it for now, like 

```
struct MetadataUpdatesWrapper {
  1: string metadataUpdates;
}

struct TableWriteInfoWrapper {
  1: string tableWriteInfo;
}
```
      
We need to create some synthetic metadata to pass [a check](https://github.com/prestodb/airlift/blob/master/drift/drift-codec/src/main/java/com/facebook/drift/codec/internal/compiler/ThriftCodecByteCodeGenerator.java#L1021) within drift codec, a.k.a making sure this method returns true for those wrappers.

tdcmeehan:
Why do we need synthetic wrappers for these fields, why not directly store them as strings?

shangm2:
> Why do we need synthetic wrappers for these fields, why not directly store them as strings?

They are strings. But since these fields are fields of classes which are being thrift serded, drift codec requires codecs/metadata  exist for all fields and so we need to create those metadata to meet the requirement.

presto-spi/src/main/java/com/facebook/presto/spi/ConnectorThriftCodec.java
shangm2: I have updated the name. FYI, we do have Experimental annotation there.

tdcmeehan (changes requested):
1. The Hive specific structures should not be generated in the same spec as the Presto structures.
2. We should not generate `*Wrapper` structs just go work around a Drift framework issue.  The Thrift spec generated should be coherent on its own.  It's not likely that this will ever be followed up on to cleaned up, so it needs to happen now, otherwise we'll just be accumulating technical debt.  See `TaskUpdateRequest`:

```thrift
struct TaskUpdateRequest {
  1: SessionRepresentation session;
  2: map<string, string> extraCredentials;
  3: optional binary fragment;
  4: list<TaskSource> sources;
  5: OutputBuffers outputIds;
  6: optional TableWriteInfoWrapper tableWriteInfo;
}
```

Notice `fragment`, the rest should follow this convention.
3. I don't think we should focus only on "critical path" fields.  All connector-specific fields should be able to be either 100% serialized and deserialized in Thrift or some other custom binary format, or 100% serialized and deserialized in JSON.  If we planned to do it at a later time, let's just do it now, because if we don't finish this work we've just complicated our communication protocol and made it way more complicated.  I don't think we should just implement it half-way and wait and see, that's how we accumulate technical debt, and you are free to try this on a copy of the code without committing this to mainline.

shangm2:
> 1. The Hive specific structures should not be generated in the same spec as the Presto structures.
> 2. We should not generate `*Wrapper` structs just go work around a Drift framework issue.  The Thrift spec generated should be coherent on its own.  It's not likely that this will ever be followed up on to cleaned up, so it needs to happen now, otherwise we'll just be accumulating technical debt.  See `TaskUpdateRequest`:
> 
> ```thrift
> struct TaskUpdateRequest {
>   1: SessionRepresentation session;
>   2: map<string, string> extraCredentials;
>   3: optional binary fragment;
>   4: list<TaskSource> sources;
>   5: OutputBuffers outputIds;
>   6: optional TableWriteInfoWrapper tableWriteInfo;
> }
> ```
> 
> Notice `fragment`, the rest should follow this convention. 3. I don't think we should focus only on "critical path" fields. All connector-specific fields should be able to be either 100% serialized and deserialized in Thrift or some other custom binary format, or 100% serialized and deserialized in JSON. If we planned to do it at a later time, let's just do it now, because if we don't finish this work we've just complicated our communication protocol. I don't think we should just implement it half-way and wait and see, that's how we accumulate technical debt, and you are free to try this on a copy of the code without committing this to mainline.

I can update the 1st one. It was there to show how to generate idl file for certain classes but I disagree with 2 and 3. 

For 2, the reason it says it is binary because in the class definition, fragment is a byte array and we are [manually](https://github.com/prestodb/presto/blob/144c18c4a5496a73d9b1f385e342b3fe9d874341/presto-main/src/main/java/com/facebook/presto/server/remotetask/HttpRemoteTask.java#L879) serializing it to binary first without a codec and then add it to the request. This is a messy way to do it. Wrappers are not a work around. They are custom codec which is a feature of drift . For any custom codec, drift requires us to provide metadata with a name in it so that it can know how to use it. And the reason we need custom codec for those fields is because we want to use a custom way for its serde. But please let me know if you have any suggestion for a better naming scheme than *wrapper.

For 3, I dont believe we should consider them as tech debt. These classes are intentionally selected to be serded with a custom way to avoid ineffective efforts.

tdcmeehan:
> For any custom codec, drift requires us to provide metadata with a name in it so that it can know how to use it. 
> ...
> And the reason we need custom codec for those fields is because we want to use a custom way for its serde.

Can you help me understand this better?  There doesn't seem to be a principled reason to add Wrapper structs at all and you're mentioning that this is only to work around a limitation in the forked Thrift client we're using--that sounds like technical debt.  We already serialize and deserialize things like DateTimes, Durations, etc, without the need for a wrapper, and those are serialized as primitive values, what's the difference?

I'm not necessarily advocating for the exact same way we're serializing `fragment`, but I am saying the end output in the IDL should show as `binary` and we shouldn't generate them to work around issues in our Drift framework. 

> For 3, if we finish all classes in one go, this pr will take forever but those extra classes wont bring any benefit. I dont believe we should consider them as tech debt. These classes are intentionally selected to be serded with a custom way to avoid ineffective efforts.

It doesn't need to only have benefit, it also needs to not add cost to the project.  Presto as a platform continuing to serialize as JSON many of the connector-specific fields leaves the serialization and deserialization halfway done, which will be very confusing and complicated for the worker.  That being said, as I mentioned above, I think it's fine if the JSON serialization is done under the hood within _the Hive connector_, that way it's a Connector-level decision.

shangm2:
> > For any custom codec, drift requires us to provide metadata with a name in it so that it can know how to use it.
> > ...
> > And the reason we need custom codec for those fields is because we want to use a custom way for its serde.
> 
> Can you help me understand this better? There doesn't seem to be a principled reason to add Wrapper structs at all and you're mentioning that this is only to work around a limitation in the forked Thrift client we're using--that sounds like technical debt. We already serialize and deserialize things like DateTimes, Durations, etc, without the need for a wrapper, and those are serialized as primitive values, what's the difference?
> 
> I'm not necessarily advocating for the exact same way we're serializing `fragment`, but I am saying the end output in the IDL should show as `binary` and we shouldn't generate them to work around issues in our Drift framework.
> 
> > For 3, if we finish all classes in one go, this pr will take forever but those extra classes wont bring any benefit. I dont believe we should consider them as tech debt. These classes are intentionally selected to be serded with a custom way to avoid ineffective efforts.
> 
> It doesn't need to only have benefit, it also needs to not add cost to the project. Presto as a platform continuing to serialize as JSON many of the connector-specific fields leaves the serialization and deserialization halfway done, which will be very confusing and complicated for the worker. That being said, as I mentioned above, I think it's fine if the JSON serialization is done under the hood within _the Hive connector_, that way it's a Connector-level decision.

Good question for sure.
For the internal codec like datetime etc, they need a [static method](https://github.com/prestodb/drift/blob/5b295fd5452acf3a2f65d7f101b4df7e935446fc/drift-codec-utils/src/main/java/com/facebook/drift/codec/utils/DataSizeToBytesThriftCodec.java#L70) with correct annotation.  This could work for simple java types but for complicated fields, we wont have the json codec until creating a instance. 

I totally agree with not adding cost to the project. For those connector-specific fields with json serde, we don't recommend fully converting it to thrift for now so we believe they are not tech debt. It might be confusing to have a json field in a thrift payload but they are all connector-specific and the serde framework this pr adds does give the freedom to use the appropriate serde for those fields in different scenarios.

tdcmeehan:
> This could work for simple java types but for complicated fields, we wont have the json codec until creating a instance. 

Can you explain this a bit more?  I see as an example `TableWriteInfoCodec` has a `JsonCodec` wired into its constructor, isn't this what's needed?  Couldn't we do something along the lines of what is found in [UuidToLeachSalzBinaryEncodingThriftCodec](https://github.com/prestodb/drift/blob/33f2fcd7df6bcaa98859a97e7698996f86246e23/drift-codec-utils/src/main/java/com/facebook/drift/codec/utils/UuidToLeachSalzBinaryEncodingThriftCodec.java#L41)--its Thrift type is `binary`, and what we do specially is we convert the `ByteBuffer` contents and feed it into a `JsonCodec`, as found in e.g. `TableWriteInfoCodec`.  Yes you would need static methods with `@FromThrift` and `@ToThrift`, but everything should be in that binary payload and easily convertible to the `JsonCodec`.

> I totally agree with not adding cost to the project. For those connector-specific fields with json serde, we don't recommend fully converting it to thrift for now so we believe they are not tech debt. It might be confusing to have a json field in a thrift payload but they are all connector-specific and the serde framework this pr adds does give the freedom to use the appropriate serde for those fields in different scenarios.

Maybe I'm misunderstanding.  From what I see in the code, I only see splits and transaction handles actually participating in the serialization.  Other things like table write handles do not, and they are converted into JSON using the old method.  I don't think it's OK to leave the framework to omit some connector-specific fields--consider the connector author in C++ needing to JSON serialize some fields, Thrift serialize others.  I'm OK to wait for the better solution, which is that all connector specific fields may be serialized by the connector, rather than to have to potentially migrate at a later point in time after production dependencies on this halfway functionality have already been added.

tdcmeehan:
But once again, I think as long as it's connector-dictated and not Presto engine-dictated, that's fine.  If we can add the rest of the fields to `ConnectorCodecProvider`, but provide default implementations which serialize and deserialize as JSON, that's... messy, but fine.

shangm2:
> But once again, I think as long as it's connector-dictated and not Presto engine-dictated, that's fine. If we can add the rest of the fields to `ConnectorCodecProvider`, but provide default implementations which serialize and deserialize as JSON, that's... messy, but fine.

The custom codec is being registered to thrift codec manager and during serde, thrift codec will automatically discover them including the thrift serde for split, transacation handle and all their fields recursively. If you are fine with adding custom serde in ConnectorCodecProvider manually, why not just registering the custom codec for the specific connector and let drift recursively find them.

shangm2:
> > This could work for simple java types but for complicated fields, we wont have the json codec until creating a instance.
> 
> Can you explain this a bit more? I see as an example `TableWriteInfoCodec` has a `JsonCodec` wired into its constructor, isn't this what's needed? Couldn't we do something along the lines of what is found in [UuidToLeachSalzBinaryEncodingThriftCodec](https://github.com/prestodb/drift/blob/33f2fcd7df6bcaa98859a97e7698996f86246e23/drift-codec-utils/src/main/java/com/facebook/drift/codec/utils/UuidToLeachSalzBinaryEncodingThriftCodec.java#L41)--its Thrift type is `binary`, and what we do specially is we convert the `ByteBuffer` contents and feed it into a `JsonCodec`, as found in e.g. `TableWriteInfoCodec`. Yes you would need static methods with `@FromThrift` and `@ToThrift`, but everything should be in that binary payload and easily convertible to the `JsonCodec`.
> 
> > I totally agree with not adding cost to the project. For those connector-specific fields with json serde, we don't recommend fully converting it to thrift for now so we believe they are not tech debt. It might be confusing to have a json field in a thrift payload but they are all connector-specific and the serde framework this pr adds does give the freedom to use the appropriate serde for those fields in different scenarios.
> 
> Maybe I'm misunderstanding. From what I see in the code, I only see splits and transaction handles actually participating in the serialization. Other things like table write handles do not, and they are converted into JSON using the old method. I don't think it's OK to leave the framework to omit some connector-specific fields--consider the connector author in C++ needing to JSON serialize some fields, Thrift serialize others. I'm OK to wait for the better solution, which is that all connector specific fields may be serialized by the connector, rather than to have to potentially migrate at a later point in time after production dependencies on this halfway functionality have already been added.

The json codec is not available in a static method. But let me give the static method another try.

shangm2:
If it is fine to you, I can remove all hive related changes and only have thrift annotation for presto-main and presto-main-base.

presto-native-execution/src/test/java/com/facebook/presto/nativeworker/PrestoNativeQueryRunnerUtils.java
```diff
@@ -176,8 +177,8 @@ public HiveQueryRunnerBuilder setFailOnNestedLoopJoin(boolean failOnNestedLoopJo
 
         public HiveQueryRunnerBuilder setUseThrift(boolean useThrift)
         {
-            this.extraProperties
-                    .put("experimental.internal-communication.thrift-transport-enabled", String.valueOf(useThrift));
+            this.extraProperties.put("experimental.internal-communication.thrift-transport-enabled", String.valueOf(useThrift));
```
shangm2, line 180: .

presto-native-execution/presto_cpp/presto_protocol/java-to-struct-json.py
```diff
@@ -41,7 +41,7 @@
 language = {
     "cpp": {
         "TypeMap": {
-            r"([ ,<])(ColumnHandle|PlanNode|RowExpression|ConnectorMetadataUpdateHandle|ConnectorDeleteTableHandle)([ ,>])": r"\1std::shared_ptr<\2>\3",
+            r"([ ,<])(ColumnHandle|PlanNode|RowExpression|ConnectorDeleteTableHandle)([ ,>])": r"\1std::shared_ptr<\2>\3",
```
shangm2: .

shangm2:
Hey @tdcmeehan  Feel free to give this pr another review. Thanks.

presto-main-base/src/main/java/com/facebook/presto/connector/ConnectorThriftCodecManager.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    public void addConnectorThriftCodecProvider(ConnectorId connectorId, ConnectorCodecProvider connectorCodecProvider)
+    {
+        requireNonNull(connectorId, "connectorId is null");
+        requireNonNull(connectorCodecProvider, "connectorThriftCodecProvider is null");
+        connectorThriftCodecProviders.put(connectorId.getCatalogName(), connectorCodecProvider);
+    }
+
+    public Optional<ConnectorCodec<ConnectorSplit>> getConnectorSplitThriftCodec(String connectorId)
```
tdcmeehan: `getConnectorSplitCodec`

presto-main-base/src/main/java/com/facebook/presto/connector/ConnectorThriftCodecManager.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+
+import static com.facebook.presto.operator.ExchangeOperator.REMOTE_CONNECTOR_ID;
+import static java.util.Objects.requireNonNull;
+
+public class ConnectorThriftCodecManager
+{
+    private final Map<String, ConnectorCodecProvider> connectorThriftCodecProviders = new ConcurrentHashMap<>();
```
tdcmeehan: `connectorCodecProviders`

presto-main-base/src/main/java/com/facebook/presto/connector/ConnectorThriftCodecManager.java
```diff
@@ -0,0 +1,64 @@
...
+        connectorThriftCodecProviders.put(connectorId.getCatalogName(), connectorCodecProvider);
+    }
+
+    public Optional<ConnectorCodec<ConnectorSplit>> getConnectorSplitThriftCodec(String connectorId)
+    {
+        requireNonNull(connectorId, "connectorId is null");
+        return Optional.ofNullable(connectorThriftCodecProviders.get(connectorId)).flatMap(ConnectorCodecProvider::getConnectorSplitCodec);
+    }
+
+    public Optional<ConnectorCodec<ConnectorTransactionHandle>> getConnectorTransactionHandleThriftCodec(String connectorId)
```
tdcmeehan: `getConnectorTransactionHandleCodec`

presto-main-base/src/main/java/com/facebook/presto/connector/ConnectorThriftCodecManager.java
```diff
@@ -0,0 +1,64 @@
...
+import javax.inject.Inject;
+
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+
+import static com.facebook.presto.operator.ExchangeOperator.REMOTE_CONNECTOR_ID;
+import static java.util.Objects.requireNonNull;
+
+public class ConnectorThriftCodecManager
```
tdcmeehan: `ConnectorCodecManager`

presto-spi/src/main/java/com/facebook/presto/spi/connector/ConnectorCodecProvider.java
```diff
@@ -0,0 +1,62 @@
...
+import com.facebook.presto.spi.ConnectorTableHandle;
+import com.facebook.presto.spi.ConnectorTableLayoutHandle;
+
+import java.util.Optional;
+
+public interface ConnectorCodecProvider
+{
+    default Optional<ConnectorCodec<ConnectorSplit>> getConnectorSplitCodec()
+    {
+        throw new UnsupportedOperationException("This method is not implemented yet.");
```
tdcmeehan: Because these are optional, shouldn't the default be to return `Optional.empty()`?

presto-main-base/src/main/java/com/facebook/presto/server/thrift/AbstractTypedThriftCodec.java
```diff
@@ -0,0 +1,215 @@
...
+                    TYPE_FIELD_ID,
+                    false, false, Requiredness.OPTIONAL, ImmutableMap.of(),
+                    new DefaultThriftTypeReference(ThriftType.STRING),
+                    TYPE_VALUE,
+                    FieldKind.THRIFT_FIELD,
+                    ImmutableList.of(),
+                    Optional.empty(),
+                    Optional.of(new ThriftMethodInjection(AbstractTypedThriftCodec.class.getDeclaredMethod("getTypeField"), ImmutableList.of())),
+                    Optional.empty(),
+                    Optional.empty()));
```
tdcmeehan: Can you help me understand how this works, in particular the use of reflection to access this class's `getTypeField`, etc methods?

shangm2: This createThriftType method is to generate metadata for java interfaces like, say, ConnectorSplit so we can register it with thrift codec manager and also generate idl for it.  but drift [does have a requirement here](https://github.com/airlift/drift/blob/master/drift-codec/src/main/java/io/airlift/drift/codec/metadata/ThriftFieldMetadata.java#L99) for metadata that "A thrift field must have an injection or extraction point". so to generate for java interfaces,  we have to provide those dummy methods to meet this requirement. But in reality this method is only needed during serde for injecting data into an object or getting data out of an object. In runtime, the thrift codec of an interface will be delegated to the codec of its concrete class.  

Another way to meet the requirement would be adding some dummy getter and setter method in the interface itself and then annotate them with drift. But this will pollute the interface design. But do let me know if you have other suggestions.

presto-spi/src/main/java/com/facebook/presto/spi/connector/ConnectorCodecProvider.java
shangm2: ah, good point.

presto-main-base/src/main/java/com/facebook/presto/server/thrift/AbstractTypedThriftCodec.java
tdcmeehan: Can you add a detailed comment that this is a hack to work around a limitation in Drift, including a code reference to the Drift source so that this is obvious?

shangm2: Absolutely.  Feel free to take another look.

presto-main-base/src/main/java/com/facebook/presto/server/thrift/AbstractTypedThriftCodec.java
```diff
@@ -0,0 +1,221 @@
...
+            writer.writeFieldBegin(new TField(CUSTOM_SERIALIZED_VALUE, TType.STRING, CUSTOM_FIELD_ID));
+            writeConcreteValue(connectorId, value, writer);
+            writer.writeFieldEnd();
+        }
+        else {
+            // If thrift codec is not available for this connector, fall back to its json
+            writer.writeFieldBegin(new TField(JSON_VALUE, TType.STRING, JSON_FIELD_ID));
+            writer.writeString(jsonCodec.toJson(value));
+            writer.writeFieldEnd();
+        }
```
tdcmeehan, line 203: Can it be written in such a way that we always use the binary field, regardless of whether a custom codec is supplied or not?  Under the hood, writing out a string and a binary value are almost the same thing.


In `ConnectorCodecManager`, we can make all of the getters for each of the codec non-optional.  When a custom codec is not supplied by the plugin, Presto may supply a `DefaultJsonConnectoCodec` which takes in a `JsonCodec` and we do simple conversion from String to bytes.

presto-spi/src/main/java/com/facebook/presto/spi/connector/ConnectorCodecProvider.java
```diff
@@ -0,0 +1,62 @@
...
+import com.facebook.presto.spi.ConnectorDeleteTableHandle;
+import com.facebook.presto.spi.ConnectorInsertTableHandle;
+import com.facebook.presto.spi.ConnectorOutputTableHandle;
+import com.facebook.presto.spi.ConnectorSplit;
+import com.facebook.presto.spi.ConnectorTableHandle;
+import com.facebook.presto.spi.ConnectorTableLayoutHandle;
+
+import java.util.Optional;
+
+public interface ConnectorCodecProvider
```
tdcmeehan, line 26: I don't think there's a use case for selectively enabling codecs for only some fields.  If that's the case, should this be changed to supply all fields, instead of making each one optional?  In which case, `Connector#getConnectorCodecProvider` would return `Optional<ConnectorCodecProvider>`, indicating just binary yes/no on whether or not a custom codec has been provided.

shangm2: We actually have some internal connectors for which we don't intend to thrift-ize all fields.

tdcmeehan: Why don't they Thrift-ize all fields?

presto-main-base/src/main/java/com/facebook/presto/server/thrift/AbstractTypedThriftCodec.java
shangm2: There is some limitation on the cpp side where fromThrift/toThrift methods are being used for serde and[ the presence of the value](https://github.com/prestodb/presto/pull/25595/files#r2240886103) either json or thrift is indicating if the specific field is json or thrift. It is easier to reason the data than, say, using the presence of connector id to indicate whether the data is json or thrift.

presto-spi/src/main/java/com/facebook/presto/spi/connector/ConnectorCodecProvider.java
shangm2: They are ongoing development so the apis are still changing. Once we turn on the feature toggle for thrift, it will be turned on for all connectors so we need to be backward compatible with json.

presto-main-base/src/main/java/com/facebook/presto/server/thrift/AbstractTypedThriftCodec.java
tdcmeehan: Let me see if I understand correctly.  In the coordinator side, it's a connector decision on how to serialize the data.  It can either use the default behavior, which uses Java reflection and Jackson to generate JSON, or it can use a new codec.

On the C++ side, there is no such flexibility, and it only presumes that the data must be serialized as Thrift or as JSON.  It knows it must be one or the other based on which field was populated.

If this is correct, can't the decision on whether or not to interpret the binary blob as a JSON string or as a Thrift structure be toggled based on the implementation of `ConnectorProtocolTemplate`?

How do you folks generate your protocol, since apparently there is a custom connector that you are using?  How do your custom connector data structures end up getting properly serialized/deserialized alongside the built-ins from the OSS repo?

shangm2: "If this is correct, can't the decision on whether or not to interpret the binary blob as a JSON string or as a Thrift structure be toggled based on the implementation of ConnectorProtocolTemplate?"

I dont quite follow. Could you give me details? The thing is the coordinator need to tell the worker if a field is a json when thrift has been turned on. I understand you prefer not mixing thrift and json for oss hive. But for meta internal connector, we would like to keep this backward compatibility so that we dont have to implement thrift for all connectors before we can turn on this feature, similar to https://github.com/prestodb/presto/pull/25242#discussion_r2240935095.

"How do you folks generate your protocol, since apparently there is a custom connector that you are using? How do your custom connector data structures end up getting properly serialized/deserialized alongside the built-ins from the OSS repo?"

For custom connector serde in cpp, you will need to implement the correct serde methods for the type, https://github.com/prestodb/presto/pull/25595/files?show-deleted-files=true&show-viewed-files=true&file-filters%5B%5D=#r2241193543

tdcmeehan: Let me try to explain a bit better.

> The thing is the coordinator need to tell the worker if a field is a json when thrift has been turned on.

If you think about this, it's not true even now.  If you think about how the coordinator serializes things and how the worker deserializes them, it's purely from convention.  They reference the same code to Java POJOs and, by convention, use the exact same code to generate serializers and deserializers that follow the object structure of the POJOs.

You would need something embedded in the payload, like a boolean switch indicating whether or not we're sending JSON or a binary blob, if this differed on a per-query basis.  Then the worker would have no way of knowing if this particular message would send over JSON or Thrift.  

But actually for each connector, while the cluster is up, it will always be consistently either binary or JSON.  So we can use a convention-based approach: the connector itself determines if it should expect Thrift or JSON, instead of us determining this in the protocol, by convention, simply because that's how the connector is written.  Does that make sense?

shangm2: Yeah, this make sense. Workers should know what they can expect from the data, json or thrift. But does not worker get that info from http header? And coordinator is deciding that info for all connectors once the cluster is up running. But yes, normally once the data type is determined for a cluster, it does not change very often. For this migration, however,  things are more complicated in the sense for some internal connector where only json codec is used, we still want to turn on thrift for them since we can not do different header per split or per connector. And we need to wrap json data within thrift payload for some special cases. From the worker's perspective, the header will be thrift but some payload can be json. So in this case, when coordinator is saying the payload is thrift in the header, how can the worker know there might be special cases? Things get more complicated when the migration is ongoing and the special cases are changing over time. Pls do let me know if you have any suggestions.

presto-main-base/src/main/java/com/facebook/presto/server/thrift/AbstractTypedThriftCodec.java
```diff
@@ -0,0 +1,221 @@
...
+            throws Exception
+    {
+        if (value == null) {
+            return;
+        }
+        String connectorId = nameResolver.apply(value);
+        requireNonNull(connectorId, "connectorId is null");
+
+        writer.writeStructBegin(new TStruct(baseClass.getSimpleName()));
+        if (isThriftCodecAvailable(connectorId)) {
```
shangm2, line 189: .

presto-main-base/src/main/java/com/facebook/presto/server/thrift/AbstractTypedThriftCodec.java
```diff
@@ -0,0 +1,221 @@
...
+                    FieldKind.THRIFT_FIELD,
+                    ImmutableList.of(),
+                    Optional.empty(),
+                    // Drift requires at least one of the three arguments below, so we provide a dummy method here as a workaround.
+                    // https://github.com/airlift/drift/blob/master/drift-codec/src/main/java/io/airlift/drift/codec/metadata/ThriftFieldMetadata.java#L99
+                    Optional.of(new ThriftMethodInjection(AbstractTypedThriftCodec.class.getDeclaredMethod("getCustomField"), ImmutableList.of())),
+                    Optional.empty(),
+                    Optional.empty()));
+            fields.add(new ThriftFieldMetadata(
+                    JSON_FIELD_ID,
```
shangm2, line 106: .

tdcmeehan (approved):

tdcmeehan:
Appreciate the follow through, thanks @shangm2!

PR #25246, opened 2025-06-04
[native] fix arrow flight setup in docker dependency image

zoltan:
## Description
currently when building the native dependency image the native setup adapters does not actually rebuild arrow with flight support, so one cannot readily enable flight support when building the runtime image, because that would fail.

## Motivation and Context
I'd like to be able to enable flight when building the runtime image, but can't, because the dependency image has the wrong arrow build.

## Release Notes

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/scripts/dockerfiles/centos-dependency.dockerfile
+++ b/presto-native-execution/scripts/dockerfiles/centos-dependency.dockerfile
@@ -22,7 +22,7 @@ COPY velox/scripts /velox/scripts
 RUN bash -c "mkdir build && \
     (cd build && ../scripts/setup-centos.sh && \
                  ../velox/scripts/setup-centos9.sh install_adapters && \
-                 ../scripts/setup-adapters.sh && \
+                 PROMPT_ALWAYS_RESPOND="Y" ../scripts/setup-adapters.sh && \
                  source ../velox/scripts/setup-centos9.sh && \
                  install_clang15 && \
                  install_cuda 12.8) && \


majetideepak (approved):

zoltan:
rebased to latest

@majetideepak @czentgr can we get this in?

czentgr (approved):

czentgr:
@zoltan Let's wait for the required tests to pass.

aditi-pandit:
@majetideepak, @czentgr : Do you think always prompting "Y" in setup scripts can slow the build ? Maybe we could have a subsequent step that only calls the arrow flight specific build method in the setup script.

zoltan:
yes, it does minimally, as it reinstalls some stuff that is already there. I didn't think that would be a concern?

if it is, I see two options
* modify the script itself that it always reinstalls arrow
* we add a new variable (ARROW_FLIGHT_PROMPT) default it to N, and override it to Y in our docker build
let me know if any of these would be preferred and I'll modify the PR.

majetideepak:
@aditi-pandit this is inside a docker image build. We always start from a clean image and only arrow is repeated twice.
Edit: This should not cause any slowdown of the image build.

aditi-pandit (approved):

zoltan:
none of the test failures seem related. is there a way for me to retrigger the failed ones?

majetideepak:
@zoltan can you rebase? The failing test has been disabled on master.

zoltan:
I'll push it tonight.

PR #25248, opened 2025-06-04
Optimize partition value evaluation in filter pushdown

feilong-liu:
## Description
Constraint is used to validate whether a hive partition matches the filters specified in query during filter pushdown.
https://github.com/prestodb/presto/blob/c08c6ef45ac9f9cbf2efc4f8d6e8594183d6de99/presto-hive/src/main/java/com/facebook/presto/hive/HivePartitionManager.java#L417
Here `constraint.predicate().get().test(partition.getKeys())` is called for one partition, and this function is called for all candidate partitions. However, it may not be necessary to call this function for all partitions.
For example, let's say we have a table with two partition keys, `ds` and `ts`, where `ds` is the date and `ts` is the hour of the day. When the query is querying data in the past 180 days, there can be 180*24=4320 candidate partitions. However, if the `predicate` in `constraint` only contains `ds` but not `ts`, we actually only need to call this function 180 times. It can lead to big difference especially when the predicate is expensive, for example large expressions or include complex string/json operations.
In the motivating example, we see query planning time decrease from [3 minutes](https://www.internalfb.com/intern/presto/query/?query_id=20250601_163732_00007_4yudm#runtime_stats) to [10 seconds](https://www.internalfb.com/intern/presto/query/?query_id=20250601_163615_00004_4yudm#runtime_stats) if we reduce the number of such function calls.
In this PR, I added a function to constraint to return the column handles which are used in predicate, and populate it during filter pushdown, so as to enable the above optimization.

## Motivation and Context
Reduce query planning time.

## Impact
Reduce query planning time.

## Test Plan
Test with production queries end to end

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Add a function to Constraint to return the input arguments for the predicate
```

### DIFF
--- a/presto-hive/src/main/java/com/facebook/presto/hive/HivePartitionManager.java
+++ b/presto-hive/src/main/java/com/facebook/presto/hive/HivePartitionManager.java
@@ -36,6 +36,9 @@ import com.facebook.presto.spi.SchemaTableName;
 import com.facebook.presto.spi.TableNotFoundException;
 import com.google.common.base.Predicates;
 import com.google.common.base.VerifyException;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
@@ -69,8 +72,10 @@ import static com.facebook.presto.hive.HiveColumnHandle.BUCKET_COLUMN_NAME;
 import static com.facebook.presto.hive.HiveErrorCode.HIVE_EXCEEDED_PARTITION_LIMIT;
 import static com.facebook.presto.hive.HiveSessionProperties.getMaxBucketsForGroupedExecution;
 import static com.facebook.presto.hive.HiveSessionProperties.getMinBucketCountToNotIgnoreTableBucketing;
+import static com.facebook.presto.hive.HiveSessionProperties.getOptimizeParsingOfPartitionValuesThreshold;
 import static com.facebook.presto.hive.HiveSessionProperties.isLegacyTimestampBucketing;
 import static com.facebook.presto.hive.HiveSessionProperties.isOfflineDataDebugModeEnabled;
+import static com.facebook.presto.hive.HiveSessionProperties.isOptimizeParsingOfPartitionValues;
 import static com.facebook.presto.hive.HiveSessionProperties.isParallelParsingOfPartitionValuesEnabled;
 import static com.facebook.presto.hive.HiveSessionProperties.shouldIgnoreTableBucketing;
 import static com.facebook.presto.hive.HiveUtil.getPartitionKeyColumnHandles;
@@ -86,6 +91,7 @@ import static com.facebook.presto.spi.Constraint.alwaysTrue;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Predicates.not;
 import static com.google.common.collect.ImmutableList.toImmutableList;
+import static com.google.common.collect.ImmutableMap.toImmutableMap;
 import static com.google.common.util.concurrent.MoreExecutors.directExecutor;
 import static com.google.common.util.concurrent.MoreExecutors.listeningDecorator;
 import static java.lang.String.format;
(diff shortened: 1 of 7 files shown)


jaystarshot (commented):
Where are the predicateInputs set and used?

feilong-liu:
> Where are the predicateInputs set and used?

I was thinking to split changes to spi change and the rest, updated the code to include the logic which uses this information too.

presto-hive/src/main/java/com/facebook/presto/hive/HivePartitionManager.java
```diff
@@ -197,14 +199,67 @@ private List<HivePartition> getPartitionListFromPartitionNames(
...
+                            NullableValue value = partition.getKeys().get(column);
+                            Domain allowedDomain = domains.get(column);
+                            if (allowedDomain != null && !allowedDomain.includesNullableValue(value.getValue())) {
+                                return false;
+                            }
+                        }
+                        return true;
+                    }).collect(toImmutableList());
+
+            if (!constraint.predicate().isPresent()) {
```
feilong-liu, line 228: No predicate exists, directly return all partitions

presto-hive/src/main/java/com/facebook/presto/hive/HivePartitionManager.java
```diff
@@ -197,14 +199,67 @@ private List<HivePartition> getPartitionListFromPartitionNames(
...
+                return partitionList;
+            }
+            Optional<List<ColumnHandle>> predicateInputs = constraint.getPredicateInputs();
+            if (!predicateInputs.isPresent()) {
```
feilong-liu, line 232: predicate input not set, cannot apply the optimizations, evaluate for all partitions

presto-hive/src/main/java/com/facebook/presto/hive/HivePartitionManager.java
```diff
@@ -197,14 +199,67 @@ private List<HivePartition> getPartitionListFromPartitionNames(
...
+                return partitionList.stream().filter(x -> constraint.predicate().get().test(x.getKeys())).collect(toList());
+            }
+            List<ColumnHandle> usedPartitionColumnsInPredicate = partitionColumns.stream().filter(x -> predicateInputs.get().contains(x)).collect(toList());
+            if (usedPartitionColumnsInPredicate.size() == partitionColumns.size()) {
```
feilong-liu, line 236: All partition columns are used, cannot apply the optimizations, evaluate for all partitions

presto-hive/src/main/java/com/facebook/presto/hive/HivePartitionManager.java
```diff
@@ -197,14 +199,67 @@ private List<HivePartition> getPartitionListFromPartitionNames(
...
+                return partitionList.stream().filter(x -> constraint.predicate().get().test(x.getKeys())).collect(toList());
+            }
+
+            ImmutableList.Builder<HivePartition> resultBuilder = ImmutableList.builder();
+            Map<List<NullableValue>, Boolean> cacheResult = new HashMap<>();
```
feilong-liu: Cache the result for predicates

jaystarshot:
LGTM , Is it easy to add a unit test in TestHivePartitionManager?

presto-hive/src/main/java/com/facebook/presto/hive/HivePartitionManager.java
jaystarshot: Can this cache blow up for very large distinct partition values ? maybe a guava cache is better?

feilong-liu: Thanks for the suggestion, just updated with guava cache

feilong-liu:
> LGTM , Is it easy to add a unit test in TestHivePartitionManager?

Sure, added a new test

jaystarshot (approved):

feilong-liu:
@jaystarshot Can you review it again? I just updated the TestHiveClientConfig to fix the failed test.

steveburnett:
Thanks for the release note entry! Is there somewhere in the documentation that it could link to?

PR #25250, opened 2025-06-05
Add support for symlink files in quick stats

### LINKED ISSUE #25199: Make QuickStats work with Symlink files
Quick stats [does file listing](https://github.com/prestodb/presto/blob/941e4282bef012292b7beca6db74297375a73e07/presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java#L348-L361) and examines the [file's format](https://github.com/prestodb/presto/blob/f02046f22ae01af401ece030f28329cac78524e3/presto-hive/src/main/java/com/facebook/presto/hive/statistics/ParquetQuickStatsBuilder.java#L312-L315) to figure out if a quick stats provider can build quick stats for it
This fails for Symlink files

Symlink files should be followed to load up the actual list of files
This is similar to how we decipher actual paths in [StoragePartitionLoader](https://github.com/prestodb/presto/blob/f02046f22ae01af401ece030f28329cac78524e3/presto-hive/src/main/java/com/facebook/presto/hive/StoragePartitionLoader.java#L584-L609)

presto-hive

## Possible Implementation
Follow the symlink files

## Context
For lakehouses that use Delta, and use the [Symlink mechanism to query these files](https://docs.delta.io/0.7.0/presto-integration.html), we won't see any benefit from turning on quick stats

xieandrew:
## Description

Closes #25199.

## Motivation and Context
Makes quick stats work for Hive tables defined with a symlink file containing a list of paths where parquet files are stored. Currently, quick stats fails for symlink files because it expects only regular files in the directory, thus failing for a symlink file (which contains a listing of regular files) This changes reads the symlink file and follow the paths in it to then get the parquet data file paths for quick stats. 

## Test Plan
Testing was done an external hive table with the following storage format in .prestoSchema
```
"storageFormat" : {
  "serDe" : "org.apache.hadoop.hive.ql.io.parquet.serde.ParquetHiveSerDe",
  "inputFormat" : "org.apache.hadoop.hive.ql.io.SymlinkTextInputFormat",
  "outputFormat" : "org.apache.hadoop.hive.ql.io.HiveIgnoreKeyTextOutputFormat"
},
```
The `externalLocation` contains a `manifest` file with the list of parquet paths to follow.

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== RELEASE NOTES ==

Hive Connector Changes
* Add support for symlink files in :ref:`connector/hive:Quick Stats`.
```

### DIFF
--- a/presto-hive/src/main/java/com/facebook/presto/hive/HiveUtil.java
+++ b/presto-hive/src/main/java/com/facebook/presto/hive/HiveUtil.java
@@ -30,6 +30,7 @@ import com.facebook.presto.hadoop.TextLineLengthLimitExceededException;
 import com.facebook.presto.hive.avro.PrestoAvroSerDe;
 import com.facebook.presto.hive.filesystem.ExtendedFileSystem;
 import com.facebook.presto.hive.metastore.Column;
+import com.facebook.presto.hive.metastore.Partition;
 import com.facebook.presto.hive.metastore.Storage;
 import com.facebook.presto.hive.metastore.Table;
 import com.facebook.presto.hive.pagefile.PageInputFormat;
@@ -92,10 +93,12 @@ import org.joda.time.format.ISODateTimeFormat;
 
 import javax.annotation.Nullable;
 
+import java.io.BufferedReader;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.io.UncheckedIOException;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Field;
@@ -171,6 +174,7 @@ import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.google.common.collect.Iterables.filter;
 import static com.google.common.collect.Lists.newArrayList;
 import static com.google.common.collect.Lists.transform;
+import static com.google.common.io.CharStreams.readLines;
 import static java.lang.Byte.parseByte;
 import static java.lang.Double.parseDouble;
 import static java.lang.Float.floatToRawIntBits;
@@ -1343,9 +1347,25 @@ public final class HiveUtil
         return directoryContextProperties.build();
     }
 
+    public static List<Path> readSymlinkPaths(ExtendedFileSystem fileSystem, Iterator<HiveFileInfo> manifestFileInfos)
+            throws IOException
+    {
+        ImmutableList.Builder<Path> targets = ImmutableList.builder();
+        while (manifestFileInfos.hasNext()) {
+            HiveFileInfo symlink = manifestFileInfos.next();
+
+            try (BufferedReader reader = new BufferedReader(new InputStreamReader(fileSystem.open(new Path(symlink.getPath())), UTF_8))) {
+                readLines(reader).stream()
+                        .map(Path::new)
+                        .forEach(targets::add);
+            }
+        }
+        return targets.build();
+    }
+
     public static List<HiveFileInfo> getTargetPathsHiveFileInfos(
             Path path,
-            HivePartitionMetadata partition,
+            Optional<Partition> partition,
             Path targetParent,
             List<Path> currentTargetPaths,
             HiveDirectoryContext hiveDirectoryContext,
@@ -1402,7 +1422,7 @@ public final class HiveUtil
     }
 
     private static Map<String, HiveFileInfo> getTargetParentHiveFileInfoMap(
-            HivePartitionMetadata partition,
+            Optional<Partition> partition,
             Path targetParent,
             HiveDirectoryContext hiveDirectoryContext,
             ExtendedFileSystem targetFilesystem,
@@ -1411,7 +1431,7 @@ public final class HiveUtil
             NamenodeStats namenodeStats)
     {
         Map<String, HiveFileInfo> targetParentHiveFileInfos = new HashMap<>();
-        Iterator<HiveFileInfo> hiveFileInfoIterator = directoryLister.list(targetFilesystem, table, targetParent, partition.getPartition(), namenodeStats, hiveDirectoryContext);
+        Iterator<HiveFileInfo> hiveFileInfoIterator = directoryLister.list(targetFilesystem, table, targetParent, partition, namenodeStats, hiveDirectoryContext);
 
         // We will use the path without the scheme and authority since the manifest file may contain entries both with and without them
         hiveFileInfoIterator.forEachRemaining(hiveFileInfo -> targetParentHiveFileInfos.put(
--- a/presto-hive/src/main/java/com/facebook/presto/hive/StoragePartitionLoader.java
+++ b/presto-hive/src/main/java/com/facebook/presto/hive/StoragePartitionLoader.java
@@ -29,7 +29,6 @@ import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterators;
 import com.google.common.collect.ListMultimap;
-import com.google.common.io.CharStreams;
 import com.google.common.util.concurrent.ListenableFuture;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.Path;
@@ -41,10 +40,7 @@ import org.apache.hadoop.mapred.InputSplit;
 import org.apache.hadoop.mapred.JobConf;
 import org.apache.hadoop.mapred.TextInputFormat;
 
-import java.io.BufferedReader;
 import java.io.IOException;
-import java.io.InputStreamReader;
-import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.Deque;
@@ -78,6 +74,7 @@ import static com.facebook.presto.hive.HiveUtil.getHeaderCount;
 import static com.facebook.presto.hive.HiveUtil.getInputFormat;
 import static com.facebook.presto.hive.HiveUtil.getTargetPathsHiveFileInfos;
 import static com.facebook.presto.hive.HiveUtil.isHudiParquetInputFormat;
+import static com.facebook.presto.hive.HiveUtil.readSymlinkPaths;
 import static com.facebook.presto.hive.HiveUtil.shouldUseFileSplitsFromInputFormat;
 import static com.facebook.presto.hive.HiveWriterFactory.getBucketNumber;
 import static com.facebook.presto.hive.NestedDirectoryPolicy.FAIL;
@@ -260,7 +257,7 @@ public class StoragePartitionLoader
 
         List<HiveFileInfo> targetPathsHiveFileInfos = getTargetPathsHiveFileInfos(
                 path,
-                partition,
+                partition.getPartition(),
                 targetParent,
                 currentTargetPaths,
                 hiveDirectoryContext,
@@ -730,7 +727,6 @@ public class StoragePartitionLoader
     private List<Path> getTargetPathsFromSymlink(ExtendedFileSystem fileSystem, Path symlinkDir, Optional<Partition> partition)
     {
         try {
-            List<Path> targets = new ArrayList<>();
             HiveDirectoryContext hiveDirectoryContext = new HiveDirectoryContext(
                     IGNORED,
                     isUseListDirectoryCache(session),
@@ -738,16 +734,8 @@ public class StoragePartitionLoader
                     hdfsContext.getIdentity(),
                     buildDirectoryContextProperties(session),
                     session.getRuntimeStats());
-            List<HiveFileInfo> manifestFileInfos = ImmutableList.copyOf(directoryLister.list(fileSystem, table, symlinkDir, partition, namenodeStats, hiveDirectoryContext));
-
-            for (HiveFileInfo symlink : manifestFileInfos) {
-                try (BufferedReader reader = new BufferedReader(new InputStreamReader(fileSystem.open(new Path(symlink.getPath())), StandardCharsets.UTF_8))) {
-                    CharStreams.readLines(reader).stream()
-                            .map(Path::new)
-                            .forEach(targets::add);
-                }
-            }
-            return targets;
+            Iterator<HiveFileInfo> manifestFileInfos = directoryLister.list(fileSystem, table, symlinkDir, partition, namenodeStats, hiveDirectoryContext);
+            return readSymlinkPaths(fileSystem, manifestFileInfos);
         }
         catch (IOException e) {
             throw new PrestoException(HIVE_BAD_DATA, "Error parsing symlinks from: " + symlinkDir, e);
--- a/presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java
+++ b/presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java
@@ -32,8 +32,10 @@ import com.facebook.presto.hive.metastore.ExtendedHiveMetastore;
 import com.facebook.presto.hive.metastore.MetastoreContext;
 import com.facebook.presto.hive.metastore.Partition;
 import com.facebook.presto.hive.metastore.PartitionStatistics;
+import com.facebook.presto.hive.metastore.StorageFormat;
 import com.facebook.presto.hive.metastore.Table;
 import com.facebook.presto.spi.ConnectorSession;
+import com.facebook.presto.spi.PrestoException;
 import com.facebook.presto.spi.SchemaTableName;
 import com.google.common.base.Stopwatch;
 import com.google.common.cache.Cache;
@@ -41,6 +43,8 @@ import com.google.common.cache.CacheBuilder;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.hive.ql.io.SymlinkTextInputFormat;
+import org.apache.hadoop.mapred.InputFormat;
 import org.weakref.jmx.Managed;
 import org.weakref.jmx.Nested;
 
@@ -61,8 +65,10 @@ import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.stream.Collectors;
 
 import static com.facebook.airlift.concurrent.Threads.daemonThreadsNamed;
+import static com.facebook.presto.hive.HiveErrorCode.HIVE_BAD_DATA;
 import static com.facebook.presto.hive.HivePartition.UNPARTITIONED_ID;
 import static com.facebook.presto.hive.HiveSessionProperties.getQuickStatsBackgroundBuildTimeout;
 import static com.facebook.presto.hive.HiveSessionProperties.getQuickStatsInlineBuildTimeout;
@@ -70,6 +76,9 @@ import static com.facebook.presto.hive.HiveSessionProperties.isQuickStatsEnabled
 import static com.facebook.presto.hive.HiveSessionProperties.isSkipEmptyFilesEnabled;
 import static com.facebook.presto.hive.HiveSessionProperties.isUseListDirectoryCache;
 import static com.facebook.presto.hive.HiveUtil.buildDirectoryContextProperties;
+import static com.facebook.presto.hive.HiveUtil.getInputFormat;
+import static com.facebook.presto.hive.HiveUtil.getTargetPathsHiveFileInfos;
+import static com.facebook.presto.hive.HiveUtil.readSymlinkPaths;
 import static com.facebook.presto.hive.NestedDirectoryPolicy.IGNORED;
 import static com.facebook.presto.hive.NestedDirectoryPolicy.RECURSE;
 import static com.facebook.presto.hive.metastore.PartitionStatistics.empty;
@@ -323,15 +332,18 @@ public class QuickStatsProvider
         Table resolvedTable = metastore.getTable(metastoreContext, table.getSchemaName(), table.getTableName()).get();
         Optional<Partition> partition;
         Path path;
+        StorageFormat storageFormat;
         if (UNPARTITIONED_ID.getPartitionName().equals(partitionId)) {
             partition = Optional.empty();
             path = new Path(resolvedTable.getStorage().getLocation());
+            storageFormat = resolvedTable.getStorage().getStorageFormat();
         }
         else {
             partition = metastore.getPartitionsByNames(metastoreContext, table.getSchemaName(), table.getTableName(),
                     ImmutableList.of(new PartitionNameWithVersion(partitionId, Optional.empty()))).get(partitionId);
             checkState(partition.isPresent(), "getPartitionsByNames returned no partitions for partition with name [%s]", partitionId);
             path = new Path(partition.get().getStorage().getLocation());
+            storageFormat = partition.get().getStorage().getStorageFormat();
         }
 
         HdfsContext hdfsContext = new HdfsContext(session, table.getSchemaName(), table.getTableName(), partitionId, false);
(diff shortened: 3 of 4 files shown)


presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java
```diff
@@ -372,6 +392,40 @@ private PartitionStatistics buildQuickStats(String partitionKey, String partitio
         return partitionStatistics;
     }
 
+    private Iterator<HiveFileInfo> getSymlinkTargetsFileList(ExtendedFileSystem fs, Table resolvedTable, Optional<Partition> partition, HiveDirectoryContext hiveDirectoryContext, Iterator<HiveFileInfo> symlinkFileList)
```
ZacBlanco: Can we try refactoring some of the code from `StoragePartitionLoader` so that the logic for resolving symlinks isn't duplicated?

xieandrew: Did this refactor by adding a static method in HiveUtil

presto-hive/src/test/java/com/facebook/presto/hive/statistics/TestQuickStatsProvider.java
```diff
@@ -396,4 +426,97 @@ public void quickStatsBuildTimeIsBounded()
...
+    {
+        java.nio.file.Path testTempDir = Files.createTempDirectory("test");
+        java.nio.file.Path symlinkFileDir = testTempDir.resolve("symlink");
+        java.nio.file.Path tableDir = testTempDir.resolve("table");
+        Files.createDirectory(symlinkFileDir);
+        Files.createDirectory(tableDir);
+
+        // Copy a parquet file from resources to the test/table temp dir
+        String fileName = "data.parquet";
+        URL resourceUrl = TestParquetQuickStatsBuilder.class.getClassLoader().getResource("quick_stats/tpch_orders_100_rows/20230707_033200_00024_6saa4_991b13d4-29c3-49d0-b25b-0d14181e2cf4");
```
ZacBlanco: You can use `Resources.getResource` from guava here instead

presto-hive/src/test/java/com/facebook/presto/hive/statistics/TestQuickStatsProvider.java
```diff
@@ -396,4 +426,97 @@ public void quickStatsBuildTimeIsBounded()
...
+        assertNotNull(resourceUrl);
+
+        java.nio.file.Path targetFilePath = tableDir.resolve(fileName);
+
+        try (InputStream in = resourceUrl.openStream()) {
+            Files.copy(in, targetFilePath, StandardCopyOption.REPLACE_EXISTING);
```
ZacBlanco: nit: static import on `REPLACE_EXISTING`

presto-hive/src/test/java/com/facebook/presto/hive/statistics/TestQuickStatsProvider.java
```diff
@@ -396,4 +426,97 @@ public void quickStatsBuildTimeIsBounded()
...
+        }
+
+        // Create the symlink manifest pointing to data.parquet
+        java.nio.file.Path manifestFilePath = Files.createFile(symlinkFileDir.resolve("manifest"));
+        try (BufferedWriter writer = Files.newBufferedWriter(manifestFilePath, StandardOpenOption.CREATE, StandardOpenOption.TRUNCATE_EXISTING)) {
```
ZacBlanco: nit: static imports on `CREATE` and `TRUNCATE_EXISTING`

presto-hive/src/test/java/com/facebook/presto/hive/statistics/TestQuickStatsProvider.java
```diff
@@ -396,4 +426,97 @@ public void quickStatsBuildTimeIsBounded()
...
+                ImmutableMap.of(),
+                Optional.empty(),
+                Optional.empty());
+
+        metastoreMock.createTable(metastoreContext, symlinkTable, new PrincipalPrivileges(ImmutableMultimap.of(), ImmutableMultimap.of()), ImmutableList.of());
+
+        DirectoryLister directoryLister = new HadoopDirectoryLister();
+
+        QuickStatsBuilder quickStatsBuilder = (session1, metastore, table, metastoreContext, partitionId, files) ->
+                new PartitionQuickStats(HivePartition.UNPARTITIONED_ID.getPartitionName(), ImmutableList.of(mockIntegerColumnStats), ImmutableList.copyOf(files).size());
```
ZacBlanco: nit: static import `UNPARTITIONED_ID`

presto-hive/src/test/java/com/facebook/presto/hive/statistics/TestQuickStatsProvider.java
xieandrew: Made this change

presto-hive/src/test/java/com/facebook/presto/hive/statistics/TestQuickStatsProvider.java
xieandrew: Added static imports

presto-hive/src/test/java/com/facebook/presto/hive/statistics/TestQuickStatsProvider.java
xieandrew: Added static import

presto-hive/src/test/java/com/facebook/presto/hive/statistics/TestQuickStatsProvider.java
xieandrew: Added static import

presto-hive/src/main/java/com/facebook/presto/hive/HiveUtil.java
```diff
@@ -1311,4 +1318,20 @@ public static Map<String, String> buildDirectoryContextProperties(ConnectorSessi
...
     }
+
+    public static List<Path> readSymlinkPaths(ExtendedFileSystem fileSystem, Iterator<HiveFileInfo> manifestFileInfos)
+            throws IOException
+    {
+        List<Path> targets = new ArrayList<>();
+        while (manifestFileInfos.hasNext()) {
+            HiveFileInfo symlink = manifestFileInfos.next();
+
+            try (BufferedReader reader = new BufferedReader(new InputStreamReader(fileSystem.open(new Path(symlink.getPath())), StandardCharsets.UTF_8))) {
```
ZacBlanco: nit: static import for UTF_8

presto-hive/src/main/java/com/facebook/presto/hive/StoragePartitionLoader.java
```diff
@@ -584,24 +580,15 @@ private List<InternalHiveSplit> getVirtuallyBucketedSplits(Path path, ExtendedFi
...
-            for (HiveFileInfo symlink : manifestFileInfos) {
-                try (BufferedReader reader = new BufferedReader(new InputStreamReader(fileSystem.open(new Path(symlink.getPath())), StandardCharsets.UTF_8))) {
-                    CharStreams.readLines(reader).stream()
-                            .map(Path::new)
-                            .forEach(targets::add);
-                }
-            }
-            return targets;
+            Iterator<HiveFileInfo> manifestFileInfos = directoryLister.list(fileSystem, table, symlinkDir, partition, namenodeStats, hiveDirectoryContext);
+            return HiveUtil.readSymlinkPaths(fileSystem, manifestFileInfos);
```
ZacBlanco: nit: static import `readSymlinkPaths`

ZacBlanco (dismissed):
two nits, otherwise LGTM

presto-hive/src/main/java/com/facebook/presto/hive/HiveUtil.java
xieandrew: Added the import

presto-hive/src/main/java/com/facebook/presto/hive/StoragePartitionLoader.java
xieandrew: Fixed, added the import

xieandrew:
Converted some other static method calls to static imports for consistency, not sure what the standards are around which classes/methods to not static import.

ZacBlanco:
Generally, use static imports when possible according to our CONTRIBUTING.md file.

There are occasionally cases where the function name is ambiguous. For those, I will occasionally use a non-static import, but it is quite rare

presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java
```diff
@@ -372,6 +389,32 @@ private PartitionStatistics buildQuickStats(String partitionKey, String partitio
         return partitionStatistics;
     }
 
+    public Iterator<HiveFileInfo> getSymlinkTargetsFileList(ExtendedFileSystem fs, Table resolvedTable, Optional<Partition> partition, HiveDirectoryContext hiveDirectoryContext, Iterator<HiveFileInfo> symlinkFileList)
+    {
+        try {
+            List<Path> targetPaths = HiveUtil.readSymlinkPaths(fs, symlinkFileList);
```
agrawalreetika: static import here?

presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java
```diff
@@ -372,6 +389,32 @@ private PartitionStatistics buildQuickStats(String partitionKey, String partitio
...
+                    }
+                }
+            }
+
+            return targetFileInfos.iterator();
+        }
+        catch (IOException e) {
+            throw new PrestoException(HIVE_BAD_DATA, "Error parsing symlinks", e);
+        }
+    }
```
agrawalreetika: I have a question, why are we doing listing again with `targetPath.getParent()` ? We already got `targetPaths` which will have data files list in the given manifest path. And I think for non-symlink tables [this](https://github.com/prestodb/presto/pull/25250/files#diff-b1bea4b5765a84b2cd8d42d6ab864bcfd01308b54aa8284bad5d1c6b6e3997c6R359) is representing data files list?

Also Data files in manifest file can be pointed to different location like -
`manifestfile1 -> { /d1/data1, /d2/data2 } `
So I am little confused what are we trying to do [here](https://github.com/prestodb/presto/pull/25250/files#diff-b1bea4b5765a84b2cd8d42d6ab864bcfd01308b54aa8284bad5d1c6b6e3997c6R399) with parent of data file?

agrawalreetika (changes requested):

steveburnett:
Thanks for the release note entry! Consider adding a link to the documentation, maybe to https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/connector/hive.rst#quick-stats. Here's the formatting for that, please edit it if you want a different doc link. 

```
== RELEASE NOTES ==

Hive Connector Changes
* Add support for symlink files in :ref:`connector/hive:Quick Stats`.
```

presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java
xieandrew: I couldn't find a way to get instances of `HiveFileInfo` from a direct file path without listing from the parent directory after searching all usages of it. So the workaround is to list from the parent and find the matching path. `Iterator<HIveFileInfo>` is needed to replace the non-symlink version.

presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java
xieandrew: Changed to a static import

presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java
```diff
@@ -372,6 +389,32 @@ private PartitionStatistics buildQuickStats(String partitionKey, String partitio
...
+                    HiveFileInfo targetFileInfo = targetParentDirFileList.next();
+                    // Find the matching target file from the parent directory file list
+                    if (targetFileInfo.getPath().equals(targetPath.toString())) {
+                        targetFileInfos.add(targetFileInfo);
+                        break;
+                    }
+                }
+            }
+
+            return targetFileInfos.iterator();
```
imjalpreet: @xieandrew we can reduce the number of list calls. Coincidentally, I’ve been working on optimizing the reading of symlink tables, and there’s some overlap with this method. You can check out this PR: https://github.com/prestodb/presto/pull/25277, you might be able to reuse parts of it or can help in optimising this method.

imjalpreet: Additionally, since we’re using DirectoryLister, if file status caching is enabled, we risk getting stale listings of parent directories. Given that Symlink Tables are also used in Delta Lake’s integration with Presto, it’s important to ensure we always get the most up-to-date listing to avoid missing any target files. This is also addressed in the PR I shared and could potentially be reused. Unless, we are fine with skipping some files as this is for Quick Stats.

xieandrew: Ideally we should reuse the same code for this logic. I can wait until your PR is merged to use your optimized solution but I will need a util method that just returns an `Iterator<HiveFileInfo>`.

presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java
```diff
@@ -372,6 +389,32 @@ private PartitionStatistics buildQuickStats(String partitionKey, String partitio
         return partitionStatistics;
     }
 
+    public Iterator<HiveFileInfo> getSymlinkTargetsFileList(ExtendedFileSystem fs, Table resolvedTable, Optional<Partition> partition, HiveDirectoryContext hiveDirectoryContext, Iterator<HiveFileInfo> symlinkFileList)
```
aaneja: nit: Line length > 180. Consider putting params on their own lines

presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java
aaneja: +1 @imjalpreet can you refactor your PR to have a utility method to build and return a list/iterator of files

presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java
aaneja: Also please change from `public` -> `static` , and add a `@VisibleForTesting`

presto-hive/src/test/java/com/facebook/presto/hive/statistics/TestQuickStatsProvider.java
```diff
@@ -396,4 +432,97 @@ public void quickStatsBuildTimeIsBounded()
...
+
+        assertEquals(symlinkTargetFilesList.size(), 1);
+        assertEquals(symlinkTargetFilesList.get(0).getPath(), "file:" + targetFilePath);
+        assertEquals(symlinkTargetFilesList.get(0).getParent(), "file:" + tableDir);
+
+        // Test entire getQuickStats and ensure file count matches fileList size in buildQuickStats
+        PartitionStatistics quickStats = quickStatsProvider.getQuickStats(SESSION, table, metastoreContext, UNPARTITIONED_ID.getPartitionName());
+
+        assertTrue(quickStats.getBasicStatistics().getFileCount().isPresent());
+        assertEquals(quickStats.getBasicStatistics().getFileCount().getAsLong(), 1L);
```
aaneja: Let's make this test a little more robust -

1. Add a symlink file in the resource dir, so you don't have to make one every time
2. Make the symlink file point to multiple other parquet files, preferably across other directories
3. Change the mocked `QuickStatsBuilder` so that you can assert on the actual `files` passed to the builder rather than just the total count

xieandrew: Symlink files only work with absolute directories, so I believe I would still need to write to the file.

presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java
xieandrew: Made these requested changes

presto-hive/src/test/java/com/facebook/presto/hive/statistics/TestQuickStatsProvider.java
xieandrew: Made changes 2 and 3 to the test

aaneja: You should be able to get absolute directory paths for a resource using `getResource(xxx).toString()`, [see usage in this test](https://github.com/prestodb/presto/blob/636a75da1c846f8de2635d537e56bb1fc5c7f254/presto-hive/src/test/java/com/facebook/presto/hive/statistics/TestParquetQuickStatsBuilder.java#L272)

presto-hive/src/test/java/com/facebook/presto/hive/statistics/TestQuickStatsProvider.java
```diff
@@ -396,4 +433,126 @@ public void quickStatsBuildTimeIsBounded()
...
+
+        // Test that the input format is correct
+        InputFormat<?, ?> inputFormat = getInputFormat(
+                hdfsEnvironment.getConfiguration(hdfsContext, symlinkTablePath),
+                resolvedTable.getStorage().getStorageFormat().getInputFormat(),
+                false);
+
+        assertTrue(inputFormat instanceof SymlinkTextInputFormat);
+
+        // Test following symlink file paths leads to the parquet file
```
aaneja: This seems redundant since we already assert that the files passed in to the `QuickStatsBuilder` were the ones used in the symlink setup

presto-hive/src/test/java/com/facebook/presto/hive/statistics/TestQuickStatsProvider.java
```diff
@@ -396,4 +433,126 @@ public void quickStatsBuildTimeIsBounded()
...
+
+        QuickStatsBuilder quickStatsBuilder = (session1, metastore, table, metastoreContext, partitionId, files) -> {
+            List<HiveFileInfo> fileInfoList = ImmutableList.copyOf(files);
+            assertEquals(fileInfoList.size(), 2);
+            assertEquals(fileInfoList.get(0).getPath(), "file:" + targetFilePath1);
+            assertEquals(fileInfoList.get(1).getPath(), "file:" + targetFilePath2);
+            return new PartitionQuickStats(UNPARTITIONED_ID.getPartitionName(), ImmutableList.of(mockIntegerColumnStats), fileInfoList.size());
+        };
+
+        NamenodeStats namenodeStats = new NamenodeStats();
```
aaneja: nit: This can be inlined

presto-hive/src/test/java/com/facebook/presto/hive/statistics/TestQuickStatsProvider.java
xieandrew: What absolute directories do I include in the static manifest file I put in resources? Currently I generate this with `Files.createTempDirectory` and write to the file. Is there a temporary directory that works in all environments?

presto-hive/src/test/java/com/facebook/presto/hive/statistics/TestQuickStatsProvider.java
xieandrew: Changed this to be inlined

presto-hive/src/test/java/com/facebook/presto/hive/statistics/TestQuickStatsProvider.java
xieandrew: Alright, removed this section of the test.

presto-hive/src/test/java/com/facebook/presto/hive/statistics/TestQuickStatsProvider.java
aaneja: I'd assumed we could use relative paths in the Symlink files, but looking at the [Hive test for symlink files](https://github.com/apache/hive/blob/master/ql/src/test/org/apache/hadoop/hive/ql/io/TestSymlinkTextInputFormat.java) looks like this is not supported, at least not explicitly

You can use the absolute paths for the below resources that you copy right now -
```
URL resourceUrl1 = getResource("quick_stats/tpcds_store_sales_sf_point_01/20230706_110621_00007_4uxkh_10e94cd0-1f67-4440-afd0-75cd328ea570");
        URL resourceUrl2 = getResource("quick_stats/tpcds_store_sales_sf_point_01/20230706_110621_00007_4uxkh_12b3ec73-4952-4df7-9987-2beb20cd5953");
```

I am fine with the small extra copying we do though

presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java
xieandrew: I rebased on master and utilized the optimized `HiveUtil.getTargetPathsHiveFileInfos` method. Should I only use this if the optimized symlink reader is enabled or can I just use this method as a replacement?

presto-hive/src/test/java/com/facebook/presto/hive/statistics/TestQuickStatsProvider.java
xieandrew: So I don't need to change the current copying logic, is that correct?

presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java
xieandrew: @agrawalreetika this has been refactored to use optimizations from another PR

presto-hive/src/main/java/com/facebook/presto/hive/HiveUtil.java
```diff
@@ -1343,9 +1348,25 @@ public static Map<String, String> buildDirectoryContextProperties(ConnectorSessi
...
+        while (manifestFileInfos.hasNext()) {
+            HiveFileInfo symlink = manifestFileInfos.next();
+
+            try (BufferedReader reader = new BufferedReader(new InputStreamReader(fileSystem.open(new Path(symlink.getPath())), UTF_8))) {
+                readLines(reader).stream()
+                        .map(Path::new)
+                        .forEach(targets::add);
+            }
+        }
+        return targets;
```
tdcmeehan:
```suggestion
        ImmutableList.Builder<Path> targets = ImmutableList.builder<>();
        while (manifestFileInfos.hasNext()) {
            HiveFileInfo symlink = manifestFileInfos.next();

            try (BufferedReader reader = new BufferedReader(new InputStreamReader(fileSystem.open(new Path(symlink.getPath())), UTF_8))) {
                readLines(reader).stream()
                        .map(Path::new)
                        .forEach(targets::add);
            }
        }
        return targets.build();
```

presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java
```diff
@@ -347,6 +360,37 @@ private PartitionStatistics buildQuickStats(String partitionKey, String partitio
...
+                            entry.getValue(),
+                            hiveDirectoryContext,
+                            fs,
+                            directoryLister,
+                            resolvedTable,
+                            nameNodeStats,
+                            session));
+                }
+
+                fileList = targetFileInfoList.iterator();
```
tdcmeehan:
```suggestion
                ImmutableList.Builder<HiveFileInfo> targetFileInfoList = ImmutableList.builder<>();

                for (Map.Entry<Path, List<Path>> entry : parentToTargets.entrySet()) {
                    targetFileInfoList.addAll(getTargetPathsHiveFileInfos(
                            path,
                            partition,
                            entry.getKey(),
                            entry.getValue(),
                            hiveDirectoryContext,
                            fs,
                            directoryLister,
                            resolvedTable,
                            nameNodeStats,
                            session));
                }

                fileList = targetFileInfoList.build().iterator();
```

tdcmeehan (changes requested):
Some small nits--[we prefer immutable collections where possible](https://github.com/prestodb/presto/blob/master/CONTRIBUTING.md#code-style).

presto-hive/src/main/java/com/facebook/presto/hive/HiveUtil.java
xieandrew: Got it, changed this to ImmutableList.

presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java
xieandrew: Made this change

tdcmeehan (approved):

presto-hive/src/main/java/com/facebook/presto/hive/statistics/QuickStatsProvider.java
imjalpreet: @xieandrew Sorry, missed this question. 

For quick stats, I think we can always default to using the optimized reader since it’s a fresh implementation. In my case, I introduced a configuration flag to allow a transition period for users who might not want to switch immediately.

PR #25262, opened 2025-06-06
optimize APPROX_DISTINCT operations on constant values

hdikeman:
## Description
`APPROX_DISTINCT` operations on a constant value (e.g. `APPROX_DISTINCT(IF(..., const))`) are more expensive than and functionally equivalent to `ARBITRARY(IF(..., 1, 0))`
Adding an optimizer rule to replace any `APPROX_DISTINCT` operations on constant conditionals with equivalent calls to `ARBITRARY`

## Motivation and Context
Some autogenerated queries use this pattern, which is inefficient and causes OOM errors for complex queries

## Impact
Queries which use this APPROX_DISTINCT pattern will consume less memory

## Test Plan
Adding test coverage, E2E and unit tests

Also did some E2E testing manually to make sure the substitution was occurring:
```
presto:tiny> explain select approx_distinct(if(orderkey = orderkey, 'const')) from orders;
                                                                                                                    Query Plan                                                               >
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------->
 - Output[PlanNodeId 8][_col0] => [approx_distinct:bigint]                                                                                                                                   >
         _col0 := approx_distinct (1:16)                                                                                                                                                     >
     - Aggregate(FINAL)[PlanNodeId 3] => [approx_distinct:bigint]                                                                                                                            >
             approx_distinct := "presto.default.arbitrary"((arbitrary)) (1:16)                                                                                                               >
         - LocalExchange[PlanNodeId 285][SINGLE] () => [arbitrary:bigint]                                                                                                                    >
             - RemoteStreamingExchange[PlanNodeId 291][GATHER - COLUMNAR] => [arbitrary:bigint]                                                                                              >
                 - Aggregate(PARTIAL)[PlanNodeId 289] => [arbitrary:bigint]                                                                                                                  >
                         arbitrary := "presto.default.arbitrary"((expr_4)) (1:16)                                                                                                            >
                     - ScanProject[PlanNodeId 0,1][table = TableHandle {connectorId='tpch', connectorHandle='orders:sf0.01', layout='Optional[orders:sf0.01]'}, projectLocality = LOCAL] => [>
                             Estimates: {source: CostBasedSourceInfo, rows: 15,000 (131.84kB), cpu: 135,000.00, memory: 0.00, network: 0.00}/{source: CostBasedSourceInfo, rows: 15,000 (131.>
                             expr_4 := IF((orderkey) = (orderkey), BIGINT'1', BIGINT'0') (1:72)                                                                                              >
                             orderkey := tpch:orderkey (1:71)                                                                                                                                >
                             tpch:orderstatus                                                                                                                                                >
                                 :: [["F"], ["O"], ["P"]] 
```

All unit tests were run on latest revision. Also, a verifier run was performed:
```
2025-06-23T16:28:11.319-0700    INFO    main    com.facebook.presto.verifier.framework.VerificationManager      Progress: 4637 succeeded, 2510 skipped, 1 resolved, 13 failed, 210 resubmitted, 100.00% done                                
[...]
2025-06-23 16:28:23 INFO: [verifier.py:1484] Imported verifier logs. Success rate: 99.87%. Correctness rate: 99.94%.
```
failed queries were due to load

## Contributor checklist

- [x] CI passed.

## Release Notes

```
== NO RELEASE NOTE ==
```


presto-main-base/src/main/java/com/facebook/presto/SystemSessionProperties.java
```diff
@@ -1904,6 +1905,11 @@ public SystemSessionProperties(
                         "Enable adding an exchange below partial aggregation over a GroupId node to improve partial aggregation performance",
                         featuresConfig.getAddExchangeBelowPartialAggregationOverGroupId(),
                         false),
+                booleanProperty(
+                        OPTIMIZE_CONSTANT_APPROXIMATE_DISTINCT,
+                        "Optimize out APPROX_DISTINCT operations over constant values",
+                        featuresConfig.isOptimizeConstantApproxDistinct(),
+                        false),
```
kaikalur, line 1912: Maybe we can default this to true. I

kaikalur:
on thinking more - I think we can simply replace this to 1 - we don't even need arbitrary. I was just concerned about messing with aggs but I think we have good enough optimizations now to figure out the constant 1 and hoist it.

kaikalur:
> on thinking more - I think we can simply replace this to 1 - we don't even need arbitrary. I was just concerned about messing with aggs but I think we have good enough optimizations now to figure out the constant 1 and hoist it.

Actually approx_distinct(null) is 0 but any other constant should be 1

kaikalur:
Also look at:

#25301

that's the most general way to do this. And actually easier!

presto-main-base/src/main/java/com/facebook/presto/SystemSessionProperties.java
hdikeman: I think this is "hidden" not the default value. Most of the other optimizer system props were also hidden so I went that way too

I did default the property itself to true in FeaturesConfig.java with `private boolean optimizeConstantApproxDistinct = true;`

hdikeman:
> > on thinking more - I think we can simply replace this to 1 - we don't even need arbitrary. I was just concerned about messing with aggs but I think we have good enough optimizations now to figure out the constant 1 and hoist it.

Gotcha. Wonder if we could defer that or open a separate task. Might be finicky to try to eliminate the aggregation entirely

> Actually approx_distinct(null) is 0 but any other constant should be 1

The way I got around this @kaikalur was by just ignoring NULLs and only replacing true constants (I check `!isNull` prior to firing the optimizer on a particular aggregation)

I could have handled it but have another task for optimization of functions with NULL inputs where could handle it more generally

By the way, thanks for the review! Sorry I'm just seeing this, I dumped this PR into draft mode and forgot about it while I worked out a couple bugs

kaikalur:
> I could have handled it but have another task for optimization of functions with NULL inputs where could handle it more generally

This non-trivial. Too many subtle semantics. So your two choices:

* Just complete this PR to handle approx_distinct for all null and non-null constants only
* Fix the linked issue to extend RowExpressionInterpter to handle all these single value aggs like min/max/approx_distinct/count(distinct)/arbitarary explicitly for constant inputs.

hdikeman:
@kaikalur and I talked offline

For the case which motivated this change (`APPROX_DISTINCT(IF(expression, constant))`) the RowExpressionEvaluator will not work, since the conditional predicate may contain non-constant variable references. But for cases of `aggregation(constant)` generally, that is a better approach

So I will remove the handling for the `APPROX_DISTINCT(constant)` case and leave the if-then replacement in. The handling for aggregations with pure constant inputs will be handled in a follow-up task. Will get an amendment shortly

kaikalur:
> @kaikalur and I talked offline
> 
> For the case which motivated this change (`APPROX_DISTINCT(expression, constant)`) the RowExpressionEvaluator will not work, since the conditional predicate may contain non-constant variable references. But for cases of `aggregation(constant)` generally, that is a better approach
> 
> So I will remove the handling for the `APPROX_DISTINCT(constant)` case and leave the if-then replacement in. The handling for aggregations with pure constant inputs will be handled in a follow-up task. Will get an amendment shortly

Yeah - let's make sure to run this rule after constant pull up happens

presto-main-base/src/main/java/com/facebook/presto/sql/planner/iterative/rule/ReplaceConditionalApproxDistinct.java
```diff
@@ -0,0 +1,187 @@
...
+import static com.facebook.presto.sql.planner.plan.Patterns.source;
+import static com.facebook.presto.sql.relational.Expressions.constant;
+import static com.facebook.presto.sql.relational.Expressions.isNull;
+import static java.util.Objects.requireNonNull;
+
+/**
+ * elimination of approx distinct on conditional constant values.
+ * <p>
+ * the following cases are covered:
+ *     - APPROX_DISTINCT(IF(COND, CONST)) -> ARBITRARY(IF(COND, 1, 0))
```
kaikalur: Let's be more precise - that CONST can also be null in which case it will be 0

presto-main-base/src/main/java/com/facebook/presto/sql/planner/iterative/rule/ReplaceConditionalApproxDistinct.java
```diff
@@ -0,0 +1,187 @@
...
+                expr = variableAllocator.newVariable("expr", BIGINT);
+                assignments.put((VariableReferenceExpression) expr, new SpecialFormExpression(
+                        ifcondition.getSourceLocation(),
+                        IF,
+                        BIGINT,
+                        ImmutableList.of(
+                            ifcondition.getArguments().get(0),
+                            constant(1L, BIGINT),
+                            constant(0L, BIGINT))));
+            }
```
kaikalur: You need to each one if it's NullConst, then 0 else 1 - in both cases.

kaikalur (changes requested):
Also write some e2e test that test result

presto-main-base/src/main/java/com/facebook/presto/sql/planner/iterative/rule/ReplaceConditionalApproxDistinct.java
hdikeman: I don't handle that case at the moment, right now I leave the NULL case as-is

I was wondering whether we could pull that value out of the aggregation (in the separate change), since `APPROX_DISTINCT(IF(condition, NULL))` will always return 0 even if the condition contains variable references. Not sure if there are other gotchas though which I'm not aware of

presto-main-base/src/main/java/com/facebook/presto/sql/planner/iterative/rule/ReplaceConditionalApproxDistinct.java
hdikeman: Right now I don't replace that case, I filter out the null cases in that conditional check

presto-main-base/src/main/java/com/facebook/presto/sql/planner/iterative/rule/ReplaceConditionalApproxDistinct.java
hdikeman: I just added support for this and fixed the comment, closing this

presto-main-base/src/main/java/com/facebook/presto/sql/planner/iterative/rule/ReplaceConditionalApproxDistinct.java
hdikeman: Started handling all IF cases, so resolving this

hdikeman:
Looks like ARBITRARY does not exist with all SQL interpreters
```
Caused by: org.h2.jdbc.JdbcSQLSyntaxErrorException: Function "ARBITRARY" not found; SQL statement:
select arbitrary(if(orderkey = 1, 1, 1)) from orders [90022-199]
```

I need to add a check to the `enabled` function for this

steveburnett:
If this PR doesn't need a release note, please add 

```
== NO RELEASE NOTE ==
```

Thanks!

presto-main-base/src/main/java/com/facebook/presto/sql/planner/iterative/rule/ReplaceConditionalApproxDistinct.java
```diff
@@ -0,0 +1,209 @@
...
+import static com.facebook.presto.sql.relational.Expressions.isNull;
+import static java.util.Objects.requireNonNull;
+
+/**
+ * elimination of approx distinct on conditional constant values.
+ * <p>
+ * depending on the inner conditional, the expression is converted
+ * to its equivalent arbitrary() expression.
+ *     - approx_distinct(if(..., non-null)) -> arbitrary(if(..., 1, 0))
+ *     - approx_distinct(if(..., non-null, non-null)) -> arbitrary(1)
```
feilong-liu: I am not sure about this part. The result can be either 1 or 2, when the two non-null arguments are different.

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java
```diff
@@ -6803,6 +6804,38 @@ public void testMergeDuplicateAggregations()
...
 
+    @Test
+    public void testReplaceConditionalApproxDistinct()
+    {
+        Session session = Session.builder(getSession())
+                .setSystemProperty(OPTIMIZE_CONDITIONAL_APPROXIMATE_DISTINCT, "true")
+                .build();
+
+        // if(..., true, true) -> if(..., 1, 1)
+        assertQuery(session, "select approx_distinct(if(orderkey = orderkey, 'const', 'const')) from orders",
```
feilong-liu: Maybe add some test cases where the condition is not always true and depends on input

presto-main-base/src/test/java/com/facebook/presto/sql/planner/assertions/AggregationFunctionMatcher.java
```diff
@@ -117,8 +117,8 @@ private static boolean verifyAggregationOrderBy(OrderingScheme orderingScheme, O
     private static boolean isEquivalent(Optional<Expression> expression, Optional<RowExpression> rowExpression)
     {
         // Function's argument provided by FunctionCallProvider is SymbolReference that already resolved from symbolAliases.
-        if (rowExpression.isPresent() && expression.isPresent()) {
-            checkArgument(rowExpression.get() instanceof VariableReferenceExpression, "can only process variableReference");
+        if (rowExpression.isPresent() && expression.isPresent() && !(expression.get() instanceof AnySymbolReference)) {
```
feilong-liu, line 120: What is this change? Why we need it?

presto-main-base/src/main/java/com/facebook/presto/SystemSessionProperties.java
```diff
@@ -320,6 +320,7 @@ public final class SystemSessionProperties
     public static final String REWRITE_EXPRESSION_WITH_CONSTANT_EXPRESSION = "rewrite_expression_with_constant_expression";
     public static final String PRINT_ESTIMATED_STATS_FROM_CACHE = "print_estimated_stats_from_cache";
     public static final String REMOVE_CROSS_JOIN_WITH_CONSTANT_SINGLE_ROW_INPUT = "remove_cross_join_with_constant_single_row_input";
+    public static final String OPTIMIZE_CONDITIONAL_APPROXIMATE_DISTINCT = "optimize_conditional_approximate_distinct";
```
feilong-liu: can you rename it somehow to reflect the constant arguments part?

presto-spi/src/main/java/com/facebook/presto/spi/function/StandardFunctionResolution.java
```diff
@@ -68,6 +68,14 @@ public interface StandardFunctionResolution
...
     FunctionHandle countFunction(Type valueType);
 
+    boolean supportsArbitraryFunction();
+
+    boolean isArbitraryFunction(FunctionHandle functionHandle);
+
+    FunctionHandle arbitraryFunction();
+
+    FunctionHandle arbitraryFunction(Type valueType);
+
```
feilong-liu: This is a change in SPI which in general we want to minimize the the change, we may not need it.
Can you check how we are doing it here https://github.com/prestodb/presto/blob/00a30bd74c717690e2742fcfb7ecb965d015de97/presto-main-base/src/main/java/com/facebook/presto/sql/planner/iterative/rule/CombineApproxPercentileFunctions.java#L203 ?

presto-main-base/src/main/java/com/facebook/presto/sql/planner/iterative/rule/ReplaceConditionalApproxDistinct.java
```diff
@@ -0,0 +1,209 @@
...
+    {
+        requireNonNull(functionAndTypeManager, "functionManager is null");
+        this.functionResolution = new FunctionResolution(functionAndTypeManager.getFunctionAndTypeResolver());
+    }
+
+    @Override
+    public boolean isEnabled(Session session)
+    {
+        return isOptimizeConditionalApproxDistinctEnabled(session) &&
+                functionResolution.supportsArbitraryFunction();
```
feilong-liu: Arbitrary function should always be supported

presto-main-base/src/main/java/com/facebook/presto/sql/planner/iterative/rule/ReplaceConditionalApproxDistinct.java
```diff
@@ -0,0 +1,209 @@
...
+
+        if (!changed) {
+            return Result.empty();
+        }
+
+        ProjectNode modifiedProject = new ProjectNode(
+                project.getSourceLocation(),
+                project.getId(),
+                project.getSource(),
+                assignments.putAll(inputs).build(),
```
feilong-liu: Why adding a new project node? I see that this inputs is not modified above?

presto-main-base/src/main/java/com/facebook/presto/sql/planner/iterative/rule/ReplaceConditionalApproxDistinct.java
```diff
@@ -0,0 +1,209 @@
...
+                project.getLocality());
+
+        return Result.ofPlanNode(new AggregationNode(
+                parent.getSourceLocation(),
+                parent.getId(),
```
feilong-liu: whenever we create a new node, generate a new plan node id with the id allocator

presto-main-base/src/main/java/com/facebook/presto/sql/planner/iterative/rule/ReplaceConditionalApproxDistinct.java
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    private long convertConstant(ConstantExpression expr)
+    {
+        return isNull(expr) ? 0L : 1L;
+    }
+
+    private RowExpression replaceIfExpression(SpecialFormExpression ifcondition)
+    {
+        ConstantExpression truethen = (ConstantExpression) ifcondition.getArguments().get(1);
```
feilong-liu: use camelCase

presto-main-base/src/main/java/com/facebook/presto/sql/planner/iterative/rule/ReplaceConditionalApproxDistinct.java
```diff
@@ -0,0 +1,209 @@
...
+        ConstantExpression falsethen = (ConstantExpression) ifcondition.getArguments().get(2);
+        RowExpression replace;
+
+        if (isNull(truethen) ^ isNull(falsethen)) {
```
feilong-liu: I personally find ^ to be hard to understand, suggest to use other explicit conditions check here

presto-main-base/src/main/java/com/facebook/presto/sql/planner/iterative/rule/ReplaceConditionalApproxDistinct.java
```diff
@@ -0,0 +1,209 @@
...
+                        (VariableReferenceExpression) aggregation.getArguments().get(0));
+                expr = variableAllocator.newVariable("expr", BIGINT);
+                assignments.put((VariableReferenceExpression) expr, replaceIfExpression(ifcondition));
+            }
+            else {
+                continue;
+            }
+
+            changed = true;
+            aggregations.put(variable, new AggregationNode.Aggregation(
```
feilong-liu: Looks like that you are not putting the other non qualified aggregations back?

presto-main-base/src/main/java/com/facebook/presto/sql/planner/iterative/rule/ReplaceConditionalApproxDistinct.java
```diff
@@ -0,0 +1,209 @@
...
+                    new CallExpression(
+                            aggregation.getCall().getSourceLocation(),
+                            "arbitrary",
```
feilong-liu: Perhaps have a static variable for it

presto-main-base/src/test/java/com/facebook/presto/sql/planner/iterative/rule/TestReplaceConditionalApproxDistinct.java
```diff
@@ -0,0 +1,165 @@
...
+import static com.facebook.presto.common.type.DoubleType.DOUBLE;
+import static com.facebook.presto.common.type.VarcharType.VARCHAR;
+import static com.facebook.presto.spi.plan.AggregationNode.Step.SINGLE;
+import static com.facebook.presto.sql.planner.assertions.PlanMatchPattern.aggregation;
+import static com.facebook.presto.sql.planner.assertions.PlanMatchPattern.expression;
+import static com.facebook.presto.sql.planner.assertions.PlanMatchPattern.functionCall;
+import static com.facebook.presto.sql.planner.assertions.PlanMatchPattern.project;
+import static com.facebook.presto.sql.planner.assertions.PlanMatchPattern.values;
+
+public class TestReplaceConditionalApproxDistinct
```
feilong-liu, line 35: Add some tests with multiple aggregations

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java
feilong-liu: perhaps a better way is to compare the results when session is disable vs. enabled

presto-main-base/src/main/java/com/facebook/presto/sql/planner/iterative/rule/ReplaceConditionalApproxDistinct.java
```diff
@@ -0,0 +1,209 @@
...
+    @Override
+    public Result apply(AggregationNode parent, Captures captures, Context context)
+    {
+        VariableAllocator variableAllocator = context.getVariableAllocator();
+        ProjectNode project = captures.get(SOURCE);
+        Assignments inputs = project.getAssignments();
+        boolean changed = false;
+        Assignments.Builder assignments = Assignments.builder();
+
+        Map<VariableReferenceExpression, AggregationNode.Aggregation> aggregations = new LinkedHashMap<>(parent.getAggregations());
```
feilong-liu: In general we prefer immutable collections from guava, perhaps use Immutable map builder here.
Here is the code style guideline: https://github.com/prestodb/presto/wiki/Presto-Development-Guidelines

Also we "When appropriate use Java 8 Stream API" https://github.com/prestodb/presto/blob/master/CONTRIBUTING.md#code-style

presto-main-base/src/main/java/com/facebook/presto/sql/planner/iterative/rule/ReplaceConditionalApproxDistinct.java
```diff
@@ -0,0 +1,209 @@
...
+
+            changed = true;
+            aggregations.put(variable, new AggregationNode.Aggregation(
+                    new CallExpression(
+                            aggregation.getCall().getSourceLocation(),
+                            "arbitrary",
+                            functionResolution.arbitraryFunction(BIGINT),
+                            BIGINT,
+                            ImmutableList.of(expr)),
+                    aggregation.getFilter(),
```
feilong-liu, line 132: Worth to add some tests with filter on aggregations

presto-main-base/src/test/java/com/facebook/presto/sql/planner/assertions/AggregationFunctionMatcher.java
hdikeman: It was a testing issue, if you mark an aggregation input as AnySymbol() it still checks the type

So I bypassed that check if the caller fed in a symbol wildcard (as above)

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java
hdikeman: In practice I have found this to be difficult, since either (1) the condition is predictable but gets optimized out or (2) varies depending on the testcase

Do you have any recommendations for potential if conditions I might be able to use here? I went through quite a few iterations until i landed on this

presto-spi/src/main/java/com/facebook/presto/spi/function/StandardFunctionResolution.java
hdikeman: Hmm so just invoke `getDisplayName().equals("arbitrary")` directly?

To me, the function resolver seems the right place for this rather than making a one-off private function

presto-main-base/src/main/java/com/facebook/presto/sql/planner/iterative/rule/ReplaceConditionalApproxDistinct.java
hdikeman: Hmm that wasn't my experience, e.g. when I created a test query invoking arbitrary directly, the local runner SQL parser could not resolve it, so I decided better to be safe than sorry and included this check

I can remove it

presto-main-base/src/main/java/com/facebook/presto/sql/planner/iterative/rule/ReplaceConditionalApproxDistinct.java
hdikeman: If you see this line where I do the if-condition substitution
```
assignments.put((VariableReferenceExpression) expr, replaceIfExpression(ifcondition));
```
assignments is a builder. The line you referenced here is where I insert all the pre-existing inputs in addition to whichever I had to add. Do actually have to add new assignments for the new conditional I create

presto-main-base/src/main/java/com/facebook/presto/sql/planner/iterative/rule/ReplaceConditionalApproxDistinct.java
hdikeman: Sure I can make it explicit I'm testing for difference there

presto-main-base/src/main/java/com/facebook/presto/sql/planner/iterative/rule/ReplaceConditionalApproxDistinct.java
hdikeman: I don't recognize the term "non qualified", could you clarify what that means?

I do initialize the aggregation builder with the entire list of existing aggregations:
```
aggregations = new LinkedHashMap<>(parent.getAggregations());
```
Not sure if that's what you mean

presto-main-base/src/main/java/com/facebook/presto/sql/planner/iterative/rule/ReplaceConditionalApproxDistinct.java
hdikeman: This for loop is a bit complicated for the stream API perhaps

I can switch to an immutable map builder instead

hdikeman:
We talked about this comment (https://github.com/prestodb/presto/pull/25262#discussion_r2151082921) more offline:

>I am not sure about this part. The result can be either 1 or 2, when the two non-null arguments are different.

Turns out this indeterminism is also an issue even for the "typical" case of `approx_distinct(if(..., const))`

i.e. the function `arbitrary(if(..., 1, 0))` is ALSO indeterminate in the case of a condition which may be either true or false for different rows: arbitrary will return 0 or 1 at random basically. What we really want is `arbitrary(if(..., 1, NULL))`

However, this has the added complication that `approx_distinct(NULL) = 0` whereas `arbitrary(NULL) = NULL`. So we need to add another project on top to convert those NULLs to zeros appropriately

Something else to consider: we could switch from ARBITRARY to MAX and handle this as `max(if(..., 1, 0))`, but it would halfway defeat the point of optimizing this case at all. Maybe would still represent an improvement though (if my assumption is correct and max is likely to chew up less memory than approx_distinct)

presto-spi/src/main/java/com/facebook/presto/spi/function/StandardFunctionResolution.java
tdcmeehan: While in general we want to minimize SPI changes, this one is harmless because arbitrary has long been a part of the canon of Presto functions.

presto-tests/src/main/java/com/facebook/presto/tests/AbstractTestQueries.java
hdikeman: I added some as we discussed offline

presto-main-base/src/main/java/com/facebook/presto/SystemSessionProperties.java
```diff
@@ -320,6 +320,7 @@ public final class SystemSessionProperties
     public static final String REWRITE_EXPRESSION_WITH_CONSTANT_EXPRESSION = "rewrite_expression_with_constant_expression";
     public static final String PRINT_ESTIMATED_STATS_FROM_CACHE = "print_estimated_stats_from_cache";
     public static final String REMOVE_CROSS_JOIN_WITH_CONSTANT_SINGLE_ROW_INPUT = "remove_cross_join_with_constant_single_row_input";
+    public static final String OPTIMIZE_CONDITIONAL_CONST_APPROXIMATE_DISTINCT = "optimize_conditional_const_approximate_distinct";
```
feilong-liu: const -> constant
In general, we try to avoid abbreviation.

presto-main-base/src/main/java/com/facebook/presto/sql/planner/iterative/rule/ReplaceConditionalApproxDistinct.java
```diff
@@ -0,0 +1,229 @@
...
+                    ifCondition.getSourceLocation(),
+                    IF,
+                    BIGINT,
+                    ImmutableList.of(
+                        ifCondition.getArguments().get(0),
+                        convertConstant(trueThen),
+                        convertConstant(falseThen)));
+        }
+        else {
+            // if(..., null, null)
```
feilong-liu, line 204: add a checkState to assert both arguments are NULL

presto-main-base/src/main/java/com/facebook/presto/sql/planner/iterative/rule/ReplaceConditionalApproxDistinct.java
```diff
@@ -0,0 +1,229 @@
...
+        if (!changed) {
+            return Result.empty();
+        }
+
+        ProjectNode child = new ProjectNode(
+                project.getSourceLocation(),
+                context.getIdAllocator().getNextId(),
+                project.getSource(),
+                inputs.putAll(project.getAssignments()).build(),
+                project.getLocality());
```
feilong-liu, line 154: This child is the same as the project node? Why creating a new instance?

hdikeman: It has some additional inputs, if you look above:
```
Assignments.Builder inputs = Assignments.builder();
[...]
     inputs.put(expression, replaceIfExpression(replaced));
```

The Assignments are immutable so I had to create an all-new project node (I assume this is by design). I just wait until this point to add the old assignments back in

feilong-liu (approved):

hdikeman:
I don't know how to resolve Sreeni's change request. He's out on vacation. I was told the easiest way around this is to close and reopen the PR. New one is here: https://github.com/prestodb/presto/pull/25428

PR #25271, opened 2025-06-09
[native] Add remaining from spill header to http shuffle API

tanjialiang:
Summary: To support PartitionedOutput spilling, we need an additional flag telling the receiving end if the remaining buffer will be from spill, for the receiving end to make proper load decisions.

Differential Revision: D76286914


spershin (approved):
Looks good.

PR #25280, opened 2025-06-10
[Iceberg] Add support for $deleted and $delete_file_path metadata columns

### LINKED ISSUE #24733: Support more Iceberg Metadata columns
https://prestodb.io/docs/current/connector/iceberg.html#extra-hidden-metadata-columns
So far Presto only supports querying the $path and $data_sequence_number metadata columns. We will need to add more columns. For more see https://iceberg.apache.org/javadoc/1.8.1/org/apache/iceberg/MetadataColumns.html. 
The first several we should support

- [DELETE_FILE_PATH](https://iceberg.apache.org/javadoc/1.8.1/org/apache/iceberg/MetadataColumns.html#DELETE_FILE_PATH)
- [IS_DELETED](https://iceberg.apache.org/javadoc/1.8.1/org/apache/iceberg/MetadataColumns.html#IS_DELETED)

## Context
These columns are important for debugging

xieandrew:
## Description
Adds support for additional Iceberg metadata as hidden columns. The "$deleted" column shows whether each row is deleted, instead of a filtering out deleted rows. The "$delete_file_path" column shows the path of the delete file corresponding to a deleted row, or NULL if the row was not deleted.

Will need some feedback to confirm that this logic is what is intended for these columns and the naming of these columns ($is_deleted or $deleted)?

## Motivation and Context
Closes #24733.

## Impact
Exposes important debugging information not previously available.

## Test Plan
Added tests in IcebergDistributedTestBase. Equality delete files have not been tested yet.

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

Iceberg Connector Changes
* Add support for ``$deleted`` metadata column
* Add support for ``$delete_file_path`` metadata column
```

### DIFF
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergUpdateablePageSource.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergUpdateablePageSource.java
@@ -15,9 +15,12 @@ package com.facebook.presto.iceberg;
 
 import com.facebook.presto.common.Page;
 import com.facebook.presto.common.block.Block;
+import com.facebook.presto.common.block.BlockBuilder;
 import com.facebook.presto.common.block.ColumnarRow;
 import com.facebook.presto.common.block.RowBlock;
+import com.facebook.presto.common.block.RunLengthEncodedBlock;
 import com.facebook.presto.hive.HivePartitionKey;
+import com.facebook.presto.iceberg.delete.DeleteFilter;
 import com.facebook.presto.iceberg.delete.IcebergDeletePageSink;
 import com.facebook.presto.iceberg.delete.RowPredicate;
 import com.facebook.presto.spi.ConnectorPageSource;
@@ -25,8 +28,10 @@ import com.facebook.presto.spi.PrestoException;
 import com.facebook.presto.spi.UpdatablePageSource;
 import com.google.common.collect.ImmutableList;
 import io.airlift.slice.Slice;
+import io.airlift.slice.Slices;
 import org.apache.iceberg.Schema;
 import org.apache.iceberg.types.Types;
+import org.apache.iceberg.util.Pair;
 
 import java.io.IOException;
 import java.io.UncheckedIOException;
@@ -34,14 +39,19 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Consumer;
+import java.util.function.Predicate;
 import java.util.function.Supplier;
+import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
 import static com.facebook.presto.common.block.ColumnarRow.toColumnarRow;
+import static com.facebook.presto.common.type.BooleanType.BOOLEAN;
+import static com.facebook.presto.common.type.VarcharType.VARCHAR;
 import static com.facebook.presto.iceberg.IcebergErrorCode.ICEBERG_BAD_DATA;
 import static com.facebook.presto.iceberg.IcebergErrorCode.ICEBERG_MISSING_COLUMN;
 import static com.google.common.base.Throwables.throwIfInstanceOf;
@@ -64,12 +74,14 @@ public class IcebergUpdateablePageSource
     private final Supplier<IcebergDeletePageSink> deleteSinkSupplier;
     private IcebergDeletePageSink positionDeleteSink;
     private final Supplier<Optional<RowPredicate>> deletePredicate;
+    private final Supplier<List<DeleteFilter>> deleteFilters;
 
     private final List<IcebergColumnHandle> columns;
     /**
      * Columns actually updated in the query
      */
     private final List<IcebergColumnHandle> updatedColumns;
+    private final List<IcebergColumnHandle> delegateColumns;
     private final Schema tableSchema;
     private final Supplier<IcebergPageSink> updatedRowPageSinkSupplier;
     private IcebergPageSink updatedRowPageSink;
@@ -83,6 +95,8 @@ public class IcebergUpdateablePageSource
     // Maps the Iceberg field ids of modified columns to their indexes in the updatedColumns columnValueAndRowIdChannels array
     private final Map<ColumnIdentity, Integer> columnIdentityToUpdatedColumnIndex = new HashMap<>();
     private final int[] outputColumnToDelegateMapping;
+    private final int isDeletedColumnId;
+    private final int deleteFilePathColumnId;
 
     public IcebergUpdateablePageSource(
             Schema tableSchema,
@@ -95,6 +109,7 @@ public class IcebergUpdateablePageSource
             List<IcebergColumnHandle> delegateColumns,
             Supplier<IcebergDeletePageSink> deleteSinkSupplier,
             Supplier<Optional<RowPredicate>> deletePredicate,
+            Supplier<List<DeleteFilter>> deleteFilters,
             Supplier<IcebergPageSink> updatedRowPageSinkSupplier,
             // the columns that this page source is supposed to update
             List<IcebergColumnHandle> updatedColumns,
@@ -104,9 +119,11 @@ public class IcebergUpdateablePageSource
         this.tableSchema = requireNonNull(tableSchema, "tableSchema is null");
         this.columns = requireNonNull(outputColumns, "columns is null");
         this.delegate = requireNonNull(delegate, "delegate is null");
+        this.delegateColumns = requireNonNull(delegateColumns, "delegateColumns is null");
         // information for deletes
         this.deleteSinkSupplier = deleteSinkSupplier;
         this.deletePredicate = requireNonNull(deletePredicate, "deletePredicate is null");
+        this.deleteFilters = requireNonNull(deleteFilters, "deleteFilters is null");
         // information for updates
         this.updatedRowPageSinkSupplier = requireNonNull(updatedRowPageSinkSupplier, "updatedRowPageSinkSupplier is null");
         this.updatedColumns = requireNonNull(updatedColumns, "updatedColumns is null");
@@ -145,6 +162,8 @@ public class IcebergUpdateablePageSource
                 outputColumnToDelegateMapping[i] = columnToIndex.get(outputColumns.get(i).getColumnIdentity());
             }
         }
+        this.isDeletedColumnId = getDelegateColumnId(IcebergColumnHandle::isDeletedColumn);
+        this.deleteFilePathColumnId = getDelegateColumnId(IcebergColumnHandle::isDeleteFilePathColumn);
     }
 
     @Override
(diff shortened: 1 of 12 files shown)


steveburnett:
Thanks for the release note entry! Minor formatting nit suggestion. 

```
== RELEASE NOTES ==

Iceberg Connector Changes
* Add support for ``$deleted`` metadata column
* Add support for ``$delete_file_path`` metadata column
```

steveburnett:
Should documentation be added for these newly supported metadata columns, as we have others documented in https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/connector/iceberg.rst#extra-hidden-metadata-columns ?

xieandrew:
Thanks, I've edited the release notes with the formatting changes. And yes, I believe docs should be added for these metadata columns. I added draft documentation for this pending a decision on the naming of these columns.

steveburnett:
Thanks for the doc! 

Now that the doc exists in the branch, I was able to test and verify doc links for the release note entry in the local doc build. 

```
== RELEASE NOTES ==

Iceberg Connector Changes
* Add support for :ref:`connector/iceberg:\`\`$deleted\`\` column`.
* Add support for :ref:`connector/iceberg:\`\`$delete_file_path\`\` column`.
```

presto-iceberg/src/main/java/com/facebook/presto/iceberg/delete/RowPredicate.java
```diff
@@ -43,4 +47,31 @@ default Page filterPage(Page page)
...
+
+        Block block;
+        if (allSameValues) {
+            block = RunLengthEncodedBlock.create(BOOLEAN, firstValue, positionCount);
+        }
+        else {
+            block = blockBuilder.build();
+        }
+
+        return page.replaceColumn(deletedDelegateColumnId, block);
```
hantangwangd: Can we refactor this code to completely avoid building and writing to the `blockBuilder` when all the rows have the same `deleted` value?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergUpdateablePageSource.java
```diff
@@ -191,7 +212,22 @@ public Page getNextPage()
...
+                        // Instead of filtering rows, we mark whether the row is deleted in the $deleted column
+                        dataPage = deleteFilterPredicate.get().markDeleted(dataPage, getDelegateColumnId(IcebergColumnHandle::isDeletedColumn));
+                    }
+                    else {
+                        Block allFalseBlock = RunLengthEncodedBlock.create(BOOLEAN, false, dataPage.getPositionCount());
+                        dataPage = dataPage.replaceColumn(getDelegateColumnId(IcebergColumnHandle::isDeletedColumn), allFalseBlock);
+                    }
+                }
+                if (hasDeleteFilePathColumn) {
+                    dataPage = markDeleteFilePath(dataPage, getDelegateColumnId(IcebergColumnHandle::isDeleteFilePathColumn));
```
hantangwangd: Can we refactor method `getDelegateColumnId` as well as the fields `hasIsDeletedColumn` and `hasDeleteFilePathColumn`, so we can invoke `getDelegateColumnId` only in constructor rather than call it for handling each data page?

hantangwangd (commented):
Thanks for this work. Just completed seeing the ``$deleted`` metadata column part, and left some comments.  Will see the ``$delete_file_path`` part later.

One more question we need to consider. Since querying with these two columns will actually somewhat change the behavior, so as far as I can see, when delete as join rewrite is enabled, the results are not correct. So we either need to fix it, or disable the delete as join optimization when these two fields are present.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/delete/RowPredicate.java
xieandrew: Wouldn't this require an extra loop to first get the deleted values (stored as an array) in the current loop and then an extra write to the BlockBuilder? Or can I use the constructor of ByteArrayBlock to directly pass in the values.

xieandrew: Made this change as you requested, let me know if that is what you had in mind.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergUpdateablePageSource.java
xieandrew: Sure, I moved the call to getDelegateColumnId to the constructor and converted the fields to ints with a value of -1 if the columns aren't in the query.

xieandrew:
Yeah this doesn't work for equality deletes because of IcebergEqualityDeleteAsJoin. Do you have an opinion on whether we should support these columns with the join or just disable the optimization?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/delete/RowPredicate.java
```diff
@@ -43,4 +47,36 @@ default Page filterPage(Page page)
...
+        if (allSameValues) {
+            block = RunLengthEncodedBlock.create(BOOLEAN, firstValue, positionCount);
+        }
+        else {
+            BlockBuilder blockBuilder = BOOLEAN.createFixedSizeBlockBuilder(positionCount);
+            for (int position = 0; position < positionCount; position++) {
+                BOOLEAN.writeBoolean(blockBuilder, deletedValues[position]);
+            }
+            block = blockBuilder.build();
+        }
```
hantangwangd:
```suggestion
        boolean allSameValues = true;
        boolean firstValue = !test(page, 0);
        BlockBuilder blockBuilder = null;
        for (int position = 1; position < positionCount; position++) {
            boolean deleted = !test(page, position);
            if (allSameValues && deleted != firstValue) {
                blockBuilder = BOOLEAN.createFixedSizeBlockBuilder(positionCount);
                for (int idx = 0; idx < position; idx++) {
                    BOOLEAN.writeBoolean(blockBuilder, firstValue);
                }
                BOOLEAN.writeBoolean(blockBuilder, deleted);
                allSameValues = false;
            }
            else if (!allSameValues) {
                BOOLEAN.writeBoolean(blockBuilder, deleted);
            }
        }

        Block block;
        if (blockBuilder != null) {
            block = blockBuilder.build();
        }
        else {
            block = RunLengthEncodedBlock.create(BOOLEAN, firstValue, positionCount);
        }
```

Do you think this code makes sense? The key point is, the boolean `blockBuilder` is only constructed and starts writing data when the first different `deleted` value is encountered.

hantangwangd (commented):
IMO, currently we can simply add the following code into `IcebergEqualityDeleteAsJoin.visitTableScan` to skip the optimization if metadata columns `$deleted` or `$delete_file_path` exists.

```
if (node.getAssignments().containsValue(IS_DELETED_COLUMN_HANDLE) || node.getAssignments().containsValue(DELETE_FILE_PATH_COLUMN_HANDLE)) {
    // Skip this optimization if metadata columns `$deleted` or `$delete_file_path` exists
    return node;
}
```

What do you think?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/delete/RowPredicate.java
xieandrew: Yes this code makes sense, thanks for the optimization suggestion.

xieandrew:
That sounds good, I've added the code to disable the optimization if the columns exists and verified that equality deletes now show up.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergUpdateablePageSource.java
```diff
@@ -311,6 +357,71 @@ private Page setUpdateRowIdBlock(Page page)
...
+
+        // Build the varchar block with the deleted file path or null if the row isn't deleted
+        for (int position = 0; position < positionCount; position++) {
+            Optional<String> deleteFilePath = getDeleteFilePath(page, position, filterPredicates);
+            if (allSameValues && !Objects.equals(firstValue, deleteFilePath.orElse(null))) {
+                allSameValues = false;
+            }
+            if (firstValue == null && deleteFilePath.isPresent()) {
+                firstValue = deleteFilePath.get();
+            }
```
hantangwangd: This logic might have an edge case when all rows contain identical non-null delete file path.

Is it suitable to use the similar logic as `RowPredicate.markDeleted` here?

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergUpdateablePageSource.java
```diff
@@ -311,6 +357,71 @@ private Page setUpdateRowIdBlock(Page page)
...
+    {
+        int targetColumnId = 0;
+        for (int i = 0; i < columns.size(); i++) {
+            if (columnPredicate.test(columns.get(i))) {
+                targetColumnId = i;
+                break;
+            }
+        }
+        return outputColumnToDelegateMapping[targetColumnId];
+    }
```
hantangwangd: Can we refactor this method to return -1 when the target column is not found? In this way, we can avoid the `if...else` logic in the constructor above.

hantangwangd (commented):
Thanks for the fix, overall looks good. Some little things for the handle logic of `$delete_file_path`.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergUpdateablePageSource.java
xieandrew: Sure, I changed the logic to be similar to markDeleted. I also added a check if positionCount is 0, let me know if this is not needed.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/IcebergUpdateablePageSource.java
xieandrew: Sure, made this refactor.

presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedTestBase.java
```diff
@@ -2016,6 +2016,40 @@ public void testHiddenColumns()
...
+        assertEquals(computeActual("SELECT * FROM test_delete_file_path").getRowCount(), 1);
+        assertEquals(computeActual("SELECT \"$delete_file_path\" FROM test_delete_file_path").getRowCount(), 2);
+
+        assertUpdate("DELETE FROM test_delete_file_path WHERE regionkey=0", 1);
+        computeActual("SELECT \"$delete_file_path\" FROM test_delete_file_path").getMaterializedRows().forEach(row -> {
+            assertEquals(row.getFieldCount(), 1);
+            assertNotNull(row.getField(0));
+        });
+    }
+
```
hantangwangd, line 2082:
```suggestion
    @Test(dataProvider = "equalityDeleteOptions")
    public void testEqualityDeletesWithDeletedHiddenColumn(String fileFormat, boolean joinRewriteEnabled)
            throws Exception
    {
        Session session = deleteAsJoinEnabled(joinRewriteEnabled);
        String tableName = "test_v2_row_delete_" + randomTableSuffix();
        assertUpdate("CREATE TABLE " + tableName + "(id int, data varchar) WITH (\"write.format.default\" = '" + fileFormat + "')");
        assertUpdate("INSERT INTO " + tableName + " VALUES (1, 'a')", 1);

        Table icebergTable = updateTable(tableName);
        writeEqualityDeleteToNationTable(icebergTable, ImmutableMap.of("id", 1));

        assertUpdate("INSERT INTO " + tableName + " VALUES (1, 'b'), (2, 'a'), (3, 'a')", 3);

        assertQuery(session, "SELECT * FROM " + tableName, "VALUES (1, 'b'), (2, 'a'), (3, 'a')");

        assertQuery(session, "SELECT \"$deleted\", * FROM " + tableName,
                "VALUES (true, 1, 'a'), (false, 1, 'b'), (false, 2, 'a'), (false, 3, 'a')");
    }
```

Let's add a test case like this to verify that the query including `$deleted` or `$delete_file_path` works correctly when equality-delete files exist and `delete as join` is enabled.

xieandrew: Thanks for this test case, I've applied it now.

hantangwangd (approved):

PR #25281, opened 2025-06-10
[native] Fix incorrect use of TestNG parameters in native-tests

pramodsatya:
## Description
TestNG parameters for `storageFormat` and `sidecarEnabled` were used incorrectly in `presto-native-tests` and were not obtained from arguments to functions. Fixes this issue by adding a `NativeTestUtils` class to obtain these parameters from Jvm options when specified, and use default values of `storageFormat=PARQUET` and `sidecarEnabled=true` otherwise. `NativeTestUtils` also consolidates repeated code from test methods like `createTables` into a utility function.
Also enables hive config `allow-drop-tables` in java hive query runner so test tables can be dropped and the same data directory can be used for repeated runs of native tests locally. 

## Motivation and Context
Fix incorrect usage of TestNG parameters as pointed out in code review [here](https://github.com/prestodb/presto/pull/24809#discussion_r2066975353). Make parameters `storageFormat` and `sidecarEnabled` optional in `presto-native-tests` with default values of `PARQUET` and `true`.

```
== NO RELEASE NOTE ==
```


presto-native-tests/src/test/java/com/facebook/presto/nativetests/TestRepartitionQueriesWithSmallPages.java
```diff
@@ -13,59 +13,44 @@
...
     @Override
     protected QueryRunner createQueryRunner() throws Exception
     {
-        boolean sidecar = parseBoolean(System.getProperty("sidecarEnabled"));
-        QueryRunner queryRunner = PrestoNativeQueryRunnerUtils.nativeHiveQueryRunnerBuilder()
-                .setStorageFormat(System.getProperty("storageFormat"))
-                .setAddStorageFormatToPath(true)
-                .setUseThrift(true)
-                .setExtraProperties(
+        return NativeTestsUtils.createNativeQueryRunner(storageFormat, sidecarEnabled,
```
aditi-pandit: I feel we needn't use the NativeTestsUtils method here... Passing the half built builder is quite awkward.

aditi-pandit (commented):
Thanks @pramodsatya

presto-native-tests/src/test/java/com/facebook/presto/nativetests/TestRepartitionQueriesWithSmallPages.java
pramodsatya: Thanks for pointing it out, this was a bit hacky and reverted it as suggested.

aditi-pandit (approved):
Thanks @pramodsatya

PR #25282, opened 2025-06-10
Fall back to original plan if index join rewrite fails

zacw7:
```
== NO RELEASE NOTE ==
```


presto-main-base/src/main/java/com/facebook/presto/sql/planner/optimizations/IndexJoinOptimizer.java
```diff
@@ -101,6 +101,10 @@ public PlanOptimizerResult optimize(
         IndexJoinRewriter rewriter;
         if (SystemSessionProperties.isNativeExecutionEnabled(session)) {
             rewriter = new NativeJoinRewriter(idAllocator, metadata, session);
+            if (!rewriter.isPlanChanged()) {
```
rschlussel: you're calling this before you try rewriting, so rewriter.isPlanChanged() will always return false.

zacw7: right... 😅

presto-main-base/src/main/java/com/facebook/presto/sql/planner/optimizations/IndexJoinOptimizer.java
```diff
@@ -106,6 +106,10 @@ public PlanOptimizerResult optimize(
             rewriter = new DefaultJoinRewriter(idAllocator, metadata, session);
         }
         PlanNode rewrittenPlan = SimplePlanRewriter.rewriteWith(rewriter, plan, null);
+        if (SystemSessionProperties.isNativeExecutionEnabled(session) && !rewriter.isPlanChanged()) {
```
rschlussel: can you explain the bug this is fixing?  what does the rewrittenPlan look like if rewriter.isPlanChanged is false? Shouldn't it be the same as the original plan?

zacw7: IndexJoin Java only supports equal join condition (=). We extended this optimizer to support non-equal join condition (BETWEEN, CONTAINS), which is expressed as an additional FilterNode on top of the Join. In order to make the conditions visible when considering turning the Join into IndexJoin, this optimizer pushes the FilterNode into the JoinNode hence plan changes. But we should keep the two nodes separate if index join doesn't happen eventually.

rschlussel: Got it. Can you add a comment (or if it's possible to change the optimizer, but I see it could be tricky).  It's unusual behavior for an unchanged plan to actually be different. And also a test to cover this case.

zacw7: Thanks @rschlussel . I've updated `planChanged` to `rewriteSucceeded` which self-explains and makes more sense here. The test has to be in presto-facebook - the only connector triggers this lives there.

rschlussel (approved):

PR #25284, opened 2025-06-11
Make rowid optional in DELETE query plan

ghelmling:
Summary: Currently, the $row_id column is automatically inserted into the output variables for DELETE queries by QueryPlanner.  If a connector does not actually use $row_id to implement DELETE, then we should not require it.  This makes $row_id optional.  If the Optional is empty, then we don't need to project the output variable.

Differential Revision: D76325048


presto-spi/src/main/java/com/facebook/presto/spi/connector/ConnectorMetadata.java
```diff
@@ -524,16 +524,16 @@ default Optional<ConnectorOutputMetadata> finishInsert(ConnectorSession session,
     }
 
     /**
-     * Get the column handle that will generate row IDs for the delete operation.
+     * Get the column handle that will generate row IDs for the delete operation, if this connector requires row IDs to support delete.
      * These IDs will be passed to the {@code deleteRows()} method of the
      * {@link com.facebook.presto.spi.UpdatablePageSource} that created them.
      */
-    default ColumnHandle getDeleteRowIdColumnHandle(ConnectorSession session, ConnectorTableHandle tableHandle)
+    default Optional<ColumnHandle> getDeleteRowIdColumnHandle(ConnectorSession session, ConnectorTableHandle tableHandle)
```
hantangwangd: One little question: should we consider backward compatibility for potential externally-implemented connectors?

ghelmling: Do you have a preference for how you would want to support this instead?

I suppose another option would be to add a guard method to the interface that could disable the behavior:
`boolean requiresDeleteRowIdColumnHandle()`

That seems kind of cumbersome to have 2 methods for both the update and delete rowId handling.  I thought that the Optional usage would encapsulate this better.

Or we could add a system session property that QueryPlanner checks before injecting the rowId.  That would avoid the ConnectorMetadata API change.  But that doesn't seem quite as clean, as whether or not the rowId column is required is more a function/requirement of the connector implementation.  So you could theoretically be using different connectors with different requirements.

I'm open to any other suggestions.

hantangwangd: If we need to ensure backward compatibility for this Connector SPI change, the modification should not affect any potentially existing external connectors that may have already implemented the modified SPI methods `getDeleteRowIdColumnHandle` or `getUpdateRowIdColumnHandle`.

IMO, we can refer to a previous SPI method modification that took backward compatibility into account, see [here](https://github.com/prestodb/presto/blob/master/presto-spi/src/main/java/com/facebook/presto/spi/connector/ConnectorMetadata.java#L130-L166). In this approach, we retain the original SPI methods but mark them as @Deprecated, while adding new SPI methods with default implementations that internally call the legacy methods. Meanwhile, the engine's internal implementation logic now exclusively uses new SPI methods. This ensures that the SPI changes remain backward-compatible, leaving all existing connectors unaffected.

One minor issue with this approach is that the methods need to maintain distinct signatures, which may require using different method names for the new ones. The code would roughly look like this:

```
    @Deprecated
    default ColumnHandle getUpdateRowIdColumnHandle(ConnectorSession session, ConnectorTableHandle tableHandle, List<ColumnHandle> updatedColumns)
    {
        throw new PrestoException(NOT_SUPPORTED, "This connector does not support updates");
    }

    default Optional<ColumnHandle> getUpdateRowIdColumn(ConnectorSession session, ConnectorTableHandle tableHandle, List<ColumnHandle> updatedColumns)
    {
        ColumnHandle columnHandle = getUpdateRowIdColumnHandle(session, tableHandle, updatedColumns);
        return Optional.ofNullable(columnHandle);
    }
```

In this way, all existing connectors (especially external-implemented connectors which we can not modify directly) would not be affected by this SPI interface change. And we can remove these deprecated SPI methods in some major version upgrade that breaks backward compatibility.

What's your opinion about this?

ghelmling: Thanks for the reply.  Yeah, that approach makes sense.  I can make that update.

ghelmling:
@hantangwangd I updated the PR with the strategy discussed above to avoid breaking external connectors:
* existing `getDeleteRowIdColumnHandle` and `getUpdateRowIdColumnHandle` are now deprecated, but left in place to avoid breaking external connectors
* new methods added for `getDeleteRowIdColumn` and `getUpdateRowIdColumn` returning the `Optional` types.

All in-tree connectors and Metadata APIs are updated to the new methods.

ghelmling:
@hantangwangd would you mind taking another look?  The FB internal failing test is due to a corresponding internal code change needed.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/PartitionTransforms.java
```diff
@@ -61,8 +61,8 @@ public final class PartitionTransforms
     private static final Pattern TRUNCATE_PATTERN = Pattern.compile("truncate\\[(\\d+)]");
     private static final DateTimeField YEAR_UTC = getInstanceUTC().year();
     private static final DateTimeField MONTH_OF_YEAR_UTC = getInstanceUTC().monthOfYear();
-    public static final int MILLISECONDS_PER_HOUR = 60 * 60 * 1000;
-    public static final int MILLISECONDS_PER_DAY = MILLISECONDS_PER_HOUR * 24;
+    public static final long MILLISECONDS_PER_HOUR = 60 * 60 * 1000;
+    public static final long MILLISECONDS_PER_DAY = MILLISECONDS_PER_HOUR * 24;
```
hantangwangd: Is this change related?

presto-main-base/src/main/java/com/facebook/presto/sql/planner/QueryPlanner.java
```diff
@@ -268,11 +266,16 @@ public DeleteNode plan(Delete node)
...
-        VariableReferenceExpression rowIdVariable = variableAllocator.newVariable(getSourceLocation(node), "$rowId", rowIdField.getType());
-        outputVariablesBuilder.add(rowIdVariable);
-        columns.put(rowIdVariable, rowIdHandle);
-        fields.add(rowIdField);
+        Optional<ColumnHandle> rowIdHandle = metadata.getDeleteRowIdColumn(session, handle);
+        Optional<Field> rowIdField = Optional.empty();
+        if (rowIdHandle.isPresent()) {
+            Type rowIdType = metadata.getColumnMetadata(session, handle, rowIdHandle.get()).getType();
+            rowIdField = Optional.of(Field.newUnqualified(node.getLocation(), Optional.empty(), rowIdType));
+            VariableReferenceExpression rowIdVariable = variableAllocator.newVariable(getSourceLocation(node), "$rowId", rowIdField.get().getType());
```
hantangwangd:
```suggestion
            VariableReferenceExpression rowIdVariable = variableAllocator.newVariable(getSourceLocation(node), "$rowId", rowIdType);
```

nit: we can directly use `rowIdType`

presto-main-base/src/main/java/com/facebook/presto/sql/planner/QueryPlanner.java
```diff
@@ -338,11 +341,16 @@ public UpdateNode plan(Update node)
...
-        VariableReferenceExpression rowIdVariable = variableAllocator.newVariable(getSourceLocation(node), "$rowId", rowIdField.getType());
-        outputVariablesBuilder.add(rowIdVariable);
-        columns.put(rowIdVariable, rowIdHandle);
-        fields.add(rowIdField);
+        Optional<ColumnHandle> rowIdHandle = metadata.getUpdateRowIdColumn(session, handle, updatedColumns);
+        Optional<Field> rowIdField = Optional.empty();
+        if (rowIdHandle.isPresent()) {
+            Type rowIdType = metadata.getColumnMetadata(session, handle, rowIdHandle.get()).getType();
+            rowIdField = Optional.of(Field.newUnqualified(node.getLocation(), Optional.empty(), rowIdType));
+            VariableReferenceExpression rowIdVariable = variableAllocator.newVariable(getSourceLocation(node), "$rowId", rowIdField.get().getType());
```
hantangwangd:
```suggestion
            VariableReferenceExpression rowIdVariable = variableAllocator.newVariable(getSourceLocation(node), "$rowId", rowIdType);
```

presto-main-base/src/main/java/com/facebook/presto/type/TimestampOperators.java
```diff
@@ -58,8 +58,8 @@
 
 public final class TimestampOperators
 {
-    public static final int MILLISECONDS_PER_HOUR = 60 * 60 * 1000;
-    public static final int MILLISECONDS_PER_DAY = MILLISECONDS_PER_HOUR * 24;
+    public static final long MILLISECONDS_PER_HOUR = 60 * 60 * 1000;
+    public static final long MILLISECONDS_PER_DAY = MILLISECONDS_PER_HOUR * 24;
```
hantangwangd: nit: the same as above, is this change related?

hantangwangd (approved):
Thanks for the refactoring, I have verified it on Iceberg V1 tables that only support metadata deletion. Looks good to me, just some little nits.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/PartitionTransforms.java
ghelmling: No, not directly.  But one of the build checks was failing without it, with something like NoSuchMethodError for floorDiv(long, int).  I think it must have been on Java 8 because floorDiv(long, int) looks like it's been around since Java 9.

I can try changing it back and see if it recurs.

gggrace14 (approved):

PR #25289, opened 2025-06-11
Revert "Support native non-equal lookup join planning"

zacw7:
This reverts commit df71d054

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-main-base/src/main/java/com/facebook/presto/sql/planner/optimizations/HashGenerationOptimizer.java
+++ b/presto-main-base/src/main/java/com/facebook/presto/sql/planner/optimizations/HashGenerationOptimizer.java
@@ -69,7 +69,6 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.function.Function;
 
-import static com.facebook.presto.SystemSessionProperties.isNativeExecutionEnabled;
 import static com.facebook.presto.SystemSessionProperties.skipHashGenerationForJoinWithTableScanInput;
 import static com.facebook.presto.common.type.BigintType.BIGINT;
 import static com.facebook.presto.spi.plan.JoinType.INNER;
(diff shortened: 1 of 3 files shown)


rschlussel:
Can you add some context about why you are reverting?

rschlussel (approved):

xiaoxmeng (approved):

aditi-pandit:
Thanks @zacw7 for the quick turnaround... Lets just give some time for the native tests run. Will approve after.

majetideepak:
@aditi-pandit I doubt the native tests will run since the code does not touch the presto-native-execution folder. That is also why the issue was not caught in the first place when the PR got merged.

majetideepak:
The only way we can verify is by temporarily adding the root path to .github/workflows/prestocpp-linux-build-and-unit-test.yml

rschlussel:
I just removed the cpp-linux-build from required tests. it wasn't kicking off anything, just stuck and blocking merge.

rschlussel:
Also, wondering if https://github.com/prestodb/presto/pull/25282 fixed the issue for you or if we still need the revert. It fixed our failing tpch q2 test internally.

aditi-pandit:
> Also, wondering if #25282 fixed the issue for you or if we still need the revert. It fixed our failing tpch q2 test internally.

https://github.com/prestodb/presto/actions/runs/15586494196/job/43894795701 is the post merge run for those tests for #25282 ... They are still failing.

aditi-pandit:
@majetideepak : Lets get this PR merged in. If the tests continue to fail in the post merge run, then there is some other change affecting as well, (though unlikely since the tpch suite has a very small subset of stable operators). In any case we would definitely need this one removed to even get to the lower level issue.

aditi-pandit:
@zacw7 @rschlussel : The tests have stabilized after this revert, so this was indeed problematic. Do you'll have any followup for it ? We should ensure all TPC-H and TPC-DS queries remain stable after it.

rschlussel:
yeah, i think making the cpp tests always required will be the best thing to prevent this kind issue from occurring again.  We have some internal follow ups as well for improving our testing.

rschlussel:
https://github.com/prestodb/presto/pull/25299 to trigger cpp tests on all prs.

aditi-pandit:
> #25299 to trigger cpp tests on all prs.

Thanks @rschlussel

PR #25293, opened 2025-06-11
[Iceberg] Fix querying data_sequence_number with equality deletes

### LINKED ISSUE #24629: 【Iceberg equality scan】fix duplicate key issue when querying metadata column "$data_sequence_number" from an iceberg table within equality deletes
## Your Environment

* Presto version used: 0.291
* Storage (HDFS/S3/GCS..): local file
* Data source and connector used: IcebergConnecntor
* Deployment (Cloud or On-prem): run IcebergQueryRunner on local env
* [Pastebin](https://pastebin.com/) link to the complete debug logs:

## Expected Behavior

Select the hidden column "$data_sequence_number" successfully 

## Current Behavior

Query failed, error messages as follow:

Query 20250226_023830_00016_bkm44 failed: Duplicate key $data_sequence_number

## Possible Solution

When update the new TableScanNode, check whether need to add the extra metadata column to the assignments and outputs.

## Steps to Reproduce

1. Create an iceberg table with equality deletes by flink cdc
2. Configure the IcebergQueryRunner, the properties are:

Image

3. Running the IcebergQueryRunner and query with the sql:
select "$data_sequence_number", * from tbl_cdc_with_equality;
or
select "$data_sequence_number" from tbl_cdc_with_equality;
4. Query failed, and receive the error

## Screenshots (if appropriate)

Image

xieandrew:
## Description
Fixes the error "Duplicate key $data_sequence_number" when trying to query "$data_sequence_number" on an Iceberg table that has equality deletes. Because Presto implicitly adds a data_sequence_number column in IcebergEqualityDeleteAsJoin already, when this column is explicitly added to the query the column is duplicated and the error occurs.

## Motivation and Context
Closes #24629.

## Impact
Allows querying "$data_sequence_number" in tables with equality deletes.

## Test Plan
Since Presto does not generate equality deletes (only position deletes), use Apache Flink to create the table following the steps outlined in the original issue comments. Then use `call iceberg.system.register_table('schemaName', 'tableName', 'metadataDir')` to import the table to Presto and run the query `select "$data_sequence_number", * from sample;`

This query fails before this change and successfully returns rows after.

A unit test was added IcebergDistributedTestBase.

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

Iceberg Connector Changes
* Fix error querying ``$data_sequence_number`` metadata column for table with equality deletes.
```

### DIFF
--- a/presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergEqualityDeleteAsJoin.java
+++ b/presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergEqualityDeleteAsJoin.java
@@ -266,9 +266,10 @@ public class IcebergEqualityDeleteAsJoin
                     new SpecialFormExpression(SpecialFormExpression.Form.IS_NULL, BooleanType.BOOLEAN,
                             new SpecialFormExpression(SpecialFormExpression.Form.COALESCE, BigintType.BIGINT, deleteVersionColumns)));
 
+            boolean hasExplicitDataSequenceNumberCol = node.getAssignments().containsValue(DATA_SEQUENCE_NUMBER_COLUMN_HANDLE);
             Assignments.Builder assignmentsBuilder = Assignments.builder();
             filter.getOutputVariables().stream()
-                    .filter(variableReferenceExpression -> !variableReferenceExpression.getName().startsWith(DATA_SEQUENCE_NUMBER_COLUMN_HANDLE.getName()))
+                    .filter(variableReferenceExpression -> hasExplicitDataSequenceNumberCol || !variableReferenceExpression.getName().startsWith(DATA_SEQUENCE_NUMBER_COLUMN_HANDLE.getName()))
                     .forEach(variableReferenceExpression -> assignmentsBuilder.put(variableReferenceExpression, variableReferenceExpression));
             return new ProjectNode(Optional.empty(), idAllocator.getNextId(), filter, assignmentsBuilder.build(), ProjectNode.Locality.LOCAL);
         }
@@ -368,12 +369,12 @@ public class IcebergEqualityDeleteAsJoin
 
             VariableReferenceExpression dataSequenceNumberVariableReference = toVariableReference(DATA_SEQUENCE_NUMBER_COLUMN_HANDLE);
             ImmutableMap.Builder<VariableReferenceExpression, ColumnHandle> assignmentsBuilder = ImmutableMap.<VariableReferenceExpression, ColumnHandle>builder()
-                    .put(dataSequenceNumberVariableReference, DATA_SEQUENCE_NUMBER_COLUMN_HANDLE)
                     .putAll(unselectedAssignments)
                     .putAll(node.getAssignments());
             ImmutableList.Builder<VariableReferenceExpression> outputsBuilder = ImmutableList.builder();
             outputsBuilder.addAll(node.getOutputVariables());
-            if (!node.getAssignments().containsKey(dataSequenceNumberVariableReference)) {
+            if (!node.getAssignments().containsValue(DATA_SEQUENCE_NUMBER_COLUMN_HANDLE)) {
+                assignmentsBuilder.put(dataSequenceNumberVariableReference, DATA_SEQUENCE_NUMBER_COLUMN_HANDLE);
                 outputsBuilder.add(dataSequenceNumberVariableReference);
             }
             outputsBuilder.addAll(unselectedAssignments.keySet());
--- a/presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedTestBase.java
+++ b/presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedTestBase.java
@@ -1423,6 +1423,36 @@ public abstract class IcebergDistributedTestBase
         assertQuery(session, "SELECT * FROM " + tableName, "VALUES (1, '1001', NULL, NULL), (3, '1003', NULL, NULL), (6, '1004', 1, NULL), (6, '1006', 2, 'th002')");
     }
 
+    @Test(dataProvider = "equalityDeleteOptions")
+    public void testEqualityDeletesWithDataSequenceNumber(String fileFormat, boolean joinRewriteEnabled)
+            throws Exception
+    {
+        Session session = deleteAsJoinEnabled(joinRewriteEnabled);
+        String tableName = "test_v2_row_delete_" + randomTableSuffix();
+        String tableName2 = "test_v2_row_delete_2_" + randomTableSuffix();
+        assertUpdate("CREATE TABLE " + tableName + "(id int, data varchar) WITH (\"write.format.default\" = '" + fileFormat + "')");
+        assertUpdate("INSERT INTO " + tableName + " VALUES (1, 'a')", 1);
+
+        assertUpdate("CREATE TABLE " + tableName2 + "(id int, data varchar) WITH (\"write.format.default\" = '" + fileFormat + "')");
+        assertUpdate("INSERT INTO " + tableName2 + " VALUES (1, 'a')", 1);
+
+        Table icebergTable = updateTable(tableName);
+        writeEqualityDeleteToNationTable(icebergTable, ImmutableMap.of("id", 1));
+
+        Table icebergTable2 = updateTable(tableName2);
+        writeEqualityDeleteToNationTable(icebergTable2, ImmutableMap.of("id", 1));
+
+        assertUpdate("INSERT INTO " + tableName + " VALUES (1, 'b'), (2, 'a'), (3, 'a')", 3);
+        assertUpdate("INSERT INTO " + tableName2 + " VALUES (1, 'b'), (2, 'a'), (3, 'a')", 3);
+
+        assertQuery(session, "SELECT * FROM " + tableName, "VALUES (1, 'b'), (2, 'a'), (3, 'a')");
+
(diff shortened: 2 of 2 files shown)


presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedTestBase.java
```diff
@@ -1423,6 +1423,25 @@ public void testEqualityDeletesWithHiddenPartitionsEvolution(String fileFormat,
...
+        assertUpdate("INSERT INTO " + tableName + " VALUES (1, 'a')", 1);
+
+        Table icebergTable = updateTable(tableName);
+        writeEqualityDeleteToNationTable(icebergTable, ImmutableMap.of("id", 1));
+
+        assertUpdate("INSERT INTO " + tableName + " VALUES (1, 'b'), (2, 'a'), (3, 'a')", 3);
+
+        assertQuery(session, "SELECT * FROM " + tableName, "VALUES (1, 'b'), (2, 'a'), (3, 'a')");
+
+        assertQuery(session, "SELECT \"$data_sequence_number\", * FROM " + tableName, "VALUES (3, 1, 'b'), (3, 2, 'a'), (3, 3, 'a')");
```
hantangwangd: I think we should also consider the scenario as follows:

```
        String tableName2 = "test_v2_row_delete_2_" + randomTableSuffix();
        assertUpdate("CREATE TABLE " + tableName + "(id int, data varchar) WITH (\"write.format.default\" = '" + fileFormat + "')");
        assertUpdate("INSERT INTO " + tableName + " VALUES (1, 'a')", 1);

        assertUpdate("CREATE TABLE " + tableName2 + "(id int, data varchar) WITH (\"write.format.default\" = '" + fileFormat + "')");
        assertUpdate("INSERT INTO " + tableName2 + " VALUES (1, 'a')", 1);

        writeEqualityDeleteToNationTable(icebergTable, ImmutableMap.of("id", 1));

        Table icebergTable2 = updateTable(tableName2);
        writeEqualityDeleteToNationTable(icebergTable2, ImmutableMap.of("id", 1));

        assertUpdate("INSERT INTO " + tableName + " VALUES (1, 'b'), (2, 'a'), (3, 'a')", 3);
        assertUpdate("INSERT INTO " + tableName2 + " VALUES (1, 'b'), (2, 'a'), (3, 'a')", 3);

        ......

        assertQuery(session, "SELECT a.\"$data_sequence_number\", b.\"$data_sequence_number\" from " + tableName + " as a, " + tableName2 + " as b where a.id = b.id",
                "VALUES (3, 3), (3, 3), (3, 3)");
```

presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergEqualityDeleteAsJoin.java
```diff
@@ -368,12 +374,14 @@ private TableScanNode createNewRoot(TableScanNode node, IcebergTableHandle icebe
...
             ImmutableMap.Builder<VariableReferenceExpression, ColumnHandle> assignmentsBuilder = ImmutableMap.<VariableReferenceExpression, ColumnHandle>builder()
-                    .put(dataSequenceNumberVariableReference, DATA_SEQUENCE_NUMBER_COLUMN_HANDLE)
                     .putAll(unselectedAssignments)
                     .putAll(node.getAssignments());
             ImmutableList.Builder<VariableReferenceExpression> outputsBuilder = ImmutableList.builder();
             outputsBuilder.addAll(node.getOutputVariables());
-            if (!node.getAssignments().containsKey(dataSequenceNumberVariableReference)) {
+            // toVariableReference will append a suffix, so we use a non-suffixed instance to check if the assignment already exists
+            VariableReferenceExpression existingDataSequenceNumberVariableReference = new VariableReferenceExpression(Optional.empty(), DATA_SEQUENCE_NUMBER_COLUMN_HANDLE.getName(), DATA_SEQUENCE_NUMBER_COLUMN_HANDLE.getType());
+            if (!node.getAssignments().containsKey(existingDataSequenceNumberVariableReference)) {
```
hantangwangd: Seems this is not the case in some scenarios, the variables in a table scan node's assignments could append a suffix as well (see the example in the comment below), so I think it would be more safer to check that `assignments` contains a value of `DATA_SEQUENCE_NUMBER_COLUMN_HANDLE`.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergEqualityDeleteAsJoin.java
```diff
@@ -266,9 +271,10 @@ public PlanNode visitTableScan(TableScanNode node, RewriteContext<Void> context)
                     new SpecialFormExpression(SpecialFormExpression.Form.IS_NULL, BooleanType.BOOLEAN,
                             new SpecialFormExpression(SpecialFormExpression.Form.COALESCE, BigintType.BIGINT, deleteVersionColumns)));
 
+            boolean hasExplicitDataSequenceNumberCol = node.getAssignments().containsKey(new VariableReferenceExpression(Optional.empty(), DATA_SEQUENCE_NUMBER_COLUMN_HANDLE.getName(), DATA_SEQUENCE_NUMBER_COLUMN_HANDLE.getType()));
```
hantangwangd: The same as below, the variables in a table scan node's assignments could append a suffix as well, seems we should check that assignments contains a value of `DATA_SEQUENCE_NUMBER_COLUMN_HANDLE` to get the flag here.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergEqualityDeleteAsJoin.java
```diff
@@ -199,7 +199,12 @@ public PlanNode visitTableScan(TableScanNode node, RewriteContext<Void> context)
                     .getAssignments()
                     .entrySet()
                     .stream()
-                    .collect(Collectors.toMap(assignment -> ((IcebergColumnHandle) (assignment.getValue())).getId(), Map.Entry::getKey));
+                    .collect(Collectors.toMap(assignment -> ((IcebergColumnHandle) (assignment.getValue())).getId(), Map.Entry::getKey, (vre1, vre2) -> {
+                        if (!vre1.getSourceLocation().isPresent()) {
+                            return vre1;
+                        }
+                        return vre2;
+                    }));
```
hantangwangd: nit: do we need this change?

steveburnett:
Thanks for the release note! Formatting nit:

```
== RELEASE NOTES ==

Iceberg Connector Changes
* Fix error querying ``$data_sequence_number`` metadata column for table with equality deletes.
```

presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergEqualityDeleteAsJoin.java
xieandrew: Yes, without this function to handle key duplicates we get the original error "Duplicate key $data_sequence_number" because both data_sequence_number columns will have the same int id (`IcebergColumnHandle#getId`). I wasn't sure how do decide which `VariableReferenceExpression` should be the key, so I just decided to use one that is implicitly added by the optimizer (one without a source location). Do you know if there is a different way I should decide?

presto-iceberg/src/test/java/com/facebook/presto/iceberg/IcebergDistributedTestBase.java
xieandrew: Thanks for providing this test case, this actually caught a failure that was fixed by changing assignments to check for the value as stated in your other review comments. I've added this test.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergEqualityDeleteAsJoin.java
xieandrew: Made the change to check for the value

presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergEqualityDeleteAsJoin.java
xieandrew: Made the change to check for the value

xieandrew:
Thanks, I've edited the release note with the correct format.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergEqualityDeleteAsJoin.java
hantangwangd: After this fix, I think we shouldn't meet the situation where multiple `data_sequence_number` columns with the same id existing here which need to be handled by this function. 

IMO, this function feels like a somewhat hacky way. It simply masks the underlying issue and might lead to more subtle problems later, for example, something like variable mismatches in a large plan tree. So if duplicate keys problem indeed occur in certain scenarios, I would rather to let it throw an exception so we can address the root cause. What's your opinion? Please let me know if I've misunderstood anything.

hantangwangd (commented):
Thanks for the fix, overall looks good to me.  Just one thing for discussing.

presto-iceberg/src/main/java/com/facebook/presto/iceberg/optimizer/IcebergEqualityDeleteAsJoin.java
xieandrew: I was mistaken, we don't actually need this function once the other fixes were applied. I've removed it in the latest push.

PR #25338, opened 2025-06-16
[native] Fix Task status reporting for not started Tasks

spershin:
## Description
If Task is not started we don't report any splits we might have received.
That makes coordinator send all splits to such Task overloading the corresponding worker and split-starving other workers.

We change the behavior and now return splits stats even when the Task is not started yet.

A new unit test added and also been testing this code along with the worker pushback (Task Queuing).

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/PrestoTask.cpp
+++ b/presto-native-execution/presto_cpp/main/PrestoTask.cpp
@@ -74,17 +74,19 @@ PrestoTaskState toPrestoTaskState(exec::TaskState state) {
   return PrestoTaskState::kAborted;
 }
 
-protocol::TaskState toProtocolTaskState(exec::TaskState state) {
+protocol::TaskState toProtocolTaskState(PrestoTaskState state) {
   switch (state) {
-    case exec::TaskState::kRunning:
+    case PrestoTaskState::kRunning:
       return protocol::TaskState::RUNNING;
-    case exec::TaskState::kFinished:
+    case PrestoTaskState::kFinished:
       return protocol::TaskState::FINISHED;
-    case exec::TaskState::kCanceled:
+    case PrestoTaskState::kCanceled:
       return protocol::TaskState::CANCELED;
-    case exec::TaskState::kFailed:
+    case PrestoTaskState::kFailed:
       return protocol::TaskState::FAILED;
-    case exec::TaskState::kAborted:
+    case PrestoTaskState::kPlanned:
+      return protocol::TaskState::PLANNED;
+    case PrestoTaskState::kAborted:
       [[fallthrough]];
     default:
       return protocol::TaskState::ABORTED;
@@ -557,14 +559,6 @@ void PrestoTask::recordProcessCpuTime() {
 }
 
 protocol::TaskStatus PrestoTask::updateStatusLocked() {
-  if (!taskStarted && (error == nullptr)) {
-    protocol::TaskStatus ret = info.taskStatus;
-    if (ret.state != protocol::TaskState::ABORTED) {
-      ret.state = protocol::TaskState::PLANNED;
-    }
-    return ret;
-  }
-
   // Error occurs when creating task or even before task is created. Set error
   // and return immediately
   if (error != nullptr) {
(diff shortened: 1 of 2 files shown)


presto-native-execution/presto_cpp/main/PrestoTask.cpp
```diff
@@ -557,14 +559,6 @@ void PrestoTask::recordProcessCpuTime() {
 }
 
 protocol::TaskStatus PrestoTask::updateStatusLocked() {
-  if (!taskStarted && (error == nullptr)) {
-    protocol::TaskStatus ret = info.taskStatus;
-    if (ret.state != protocol::TaskState::ABORTED) {
-      ret.state = protocol::TaskState::PLANNED;
-    }
```
tanjialiang, line 564: The return state behavior also changed. Is this expected that we never check status and return PLANNED?

presto-native-execution/presto_cpp/main/PrestoTask.cpp
```diff
@@ -575,11 +569,15 @@ protocol::TaskStatus PrestoTask::updateStatusLocked() {
     recordProcessCpuTime();
     return info.taskStatus;
   }
-  VELOX_CHECK_NOT_NULL(task, "task is null when updating status");
+
+  // We can be here before the fragment plan is received and exec task created.
+  if (task == nullptr) {
+    return info.taskStatus;
```
tanjialiang, line 576: Shall we perform a check on taskStarted here?
VELOX_CHECK(!taskStarted);

tanjialiang (commented):
Thanks, some comments

presto-native-execution/presto_cpp/main/PrestoTask.cpp
spershin: The PLANNED logic is in another member function PrestoTask::taskState().

presto-native-execution/presto_cpp/main/PrestoTask.cpp
spershin: No, we should not.
It could be possible the task can be finished and the task cleared from it (not 100% sure on that logic though).
There is no business of this function to check on this.
We simply return some stats from Velox task if we have it.

tanjialiang: Your original code has this logic binding such that if task is null then taskStarted is guaranteed to be false. And the original code puts a check here to check the state of presto task so a similar state check should also be performed to make sure there's not check regression. Checking can help surfacing issues early, facilitating debugging.

spershin: @tanjialiang 
Added.

tanjialiang (approved):

presto-native-execution/presto_cpp/main/tests/PrestoTaskTest.cpp
```diff
@@ -76,3 +111,60 @@ TEST_F(PrestoTaskTest, basic) {
...
+  const facebook::velox::core::PlanFragment planFragment;
+
+  // No exec task yet (no fragment plan), so in planned state.
+  auto status = prestoTask.updateStatus();
+  EXPECT_EQ(status.state, protocol::TaskState::PLANNED);
+  EXPECT_EQ(status.queuedPartitionedDrivers, 0);
+  EXPECT_EQ(status.runningPartitionedDrivers, 0);
+
+  // Create exec task, but not started yet, so still in planned state.
+  prestoTask.task = createExecTask(taskId, prestoTask);
```
aditi-pandit, line 129: Can we add a split at this point, when the state is PLANNED, and check the status of the drivers ? My thinking was that you were trying to fix this situation as well.

aditi-pandit (commented):
@spershin : Thanks for your code. Had a comment for your test.

presto-native-execution/presto_cpp/main/tests/PrestoTaskTest.cpp
spershin: Add split after the task is created, but before we started?
Good point, will add it in the next PR, want to merge this sooner.

PR #25339, opened 2025-06-16
[native] Add 'hasStartedTasks' to QueryCtxCacheValue.

spershin:
## Description
Adding 'hasStartedTasks' member to QueryCtxCacheValue.
The member would be used in Task Queueing to determine if we should queue a new Task or not.
It would also be set to true whenever a Task gets started (next PR).

Added simple unit test.

```
== NO RELEASE NOTE ==
```

### DIFF
--- a/presto-native-execution/presto_cpp/main/QueryContextManager.h
+++ b/presto-native-execution/presto_cpp/main/QueryContextManager.h
@@ -28,7 +28,11 @@ class QueryContextCache {
  public:
   using QueryCtxWeakPtr = std::weak_ptr<velox::core::QueryCtx>;
   using QueryIdList = std::list<protocol::QueryId>;
-  using QueryCtxCacheValue = std::pair<QueryCtxWeakPtr, QueryIdList::iterator>;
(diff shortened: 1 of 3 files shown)


presto-native-execution/presto_cpp/main/QueryContextManager.h
```diff
@@ -59,22 +63,37 @@ class QueryContextCache {
...
+    return false;
+  }
+
+  void setHasStartedTasks(const protocol::QueryId& queryId) {
+    auto iter = queryCtxs_.find(queryId);
+    if (iter != queryCtxs_.end()) {
+      iter->second.hasStartedTasks = true;
+    }
+  }
+
```
tanjialiang, line 91: Do they need to be thread-safe?

spershin: @tanjialiang 
No, we always access the class via folly::synchronized.
To be more precise, QueryContextCache is private in QueryContextManager and wrapped in folly::synchronized.
That's why no member of QueryContextCache needs any synchronization.

PR #25349, opened 2025-06-18
[native] Do not queue Tasks if the Query is already running

spershin:
## Description
This PR modifies Task Queue o group Tasks by their Query to ensure we either run the whole Query or get is queued on a single worker.

```
== NO RELEASE NOTE ==
```


amitkdutta (approved):
Looks good. Thanks @spershin

amitkdutta:
One thing to note - this decision of query is running or not is local to a worker. Part of the query can be queued by another worker. It's better to design this in coordinator than worker.

PR #25351, opened 2025-06-18
Enhance ShowColumns

sumi-mathew:
## Description

Added “Scale” and “Precision” column to get the respective scale of the decimal value and precision of numerical values.Additionally "Length" column is introduced to get the length of Char and Varchar fields.

All the above higlighted columns will be fetched for the below query output:

show columns from  "catalog"."schema"."table";

Existing feature :For this query presto returns only Column | Type | Extra | Comment

New feature: Henceforth, presto will return-> Column | Type | Extra | Comment | Precision | Scale | Length

## Contributor checklist

- [ ] CI passed.

## Release Notes

```
== RELEASE NOTES ==

General Changes
* Add  Scale and Precision column to get the respective scale of the decimal value and precision of numerical values.Additionally  Length  column is introduced to get the length of Char and Varchar fields

### CODE REVIEW

**Comment by steveburnett** [comment]
Could https://github.com/prestodb/presto/blob/master/presto-docs/src/main/sphinx/sql/show-columns.rst be updated? It seems as though even current functionality is missing, let alone the new functionality added in this PR.

File: presto-hive/src/test/java/com/facebook/presto/hive/hudi/TestHudiIntegration.java [hunk 1]
```diff
@@ -151,10 +152,36 @@ public void testQueryOnUnavailablePartition()
...
+        String rows = Streams.concat(regularRows, partitionRows)
+                .collect(Collectors.joining(", "));
+
         return "SELECT * FROM VALUES " + rows;
     }
+
+    private static String getColumnSize(String type)
+    {
+        switch (type.toLowerCase(ENGLISH)) {
+            case "bigint":
```
pratyakshsharma, line 178: I think we have only added the data types which are getting used in this class currently. Let us have an elaborate method here for future reference and to avoid any confusions.

sumi-mathew: i've addressed the comments and committed the changes.

pratyakshsharma:
Everything seems to be failing.

sumi-mathew:
> Everything seems to be failing.

@pratyakshsharma  Could you please review this PR.

sumi-mathew:
> Bring one question for discussing. Should we call the new column `precision_radix` rather than `precision`? I know the corresponding column in our current `information_schema.columns` was named `precision` as well, but that seems incorrect. The values list in the `precision` column are actually precision radix values.

Thanks for flagging this! I've updated the column name to precision_radix to more accurately reflect its contents.

sumi-mathew:
> I noticed that the currently existing table `system.jdbc.types` contains fields `precision` and `num_prec_radix`, while table `system.jdbc.columns` contains fields `column_size` and `num_prec_radix`. The values for both `precision` and `column_size` fields are obtained by calling `ColumnJdbcTable.columnSize(Type type)`. As I see, the `ColumnJdbcTable.columnSize` method correctly computes the `precision` value for numerical column types, aligning with the corresponding `precision_radix` value.
> 
> So, if we keep the addition of `precision` field in table `information_schema.columns` and `SHOW COLUMNS`, we should better call `ColumnJdbcTable.columnSize` or reuse its numerical precision calculation logic. What's your opinion?
> 
> Besides, do you think it makes sense to add both `precision` and `precision_radix` for table `information_schema.columns` and `SHOW COLUMNS`?

Thank you for the detailed observation.

The main goal of this PR is to help external users understand the capacity of numeric fields when importing data into Presto — particularly how many decimal digits a column can store. This is especially useful for types like DECIMAL(p, s), where understanding precision and scale is critical for ensuring compatibility and preventing data truncation or rounding errors.

Let me know if you'd prefer to include precision_radix for completeness or standards alignment — otherwise, I'd recommend keeping it simple for now

steveburnett (dismissed):
LGTM! (docs)

hantangwangd:
I don't have a strong preference about including the `precision_radix` column right now -- MySQL doesn't seem to have this column as well. I'm okay if the `precision` column shows the right values. That is, for example, for a column of type `decimal(18, 2)`, the precision value we get should be 18 rather than 10.

sumi-mathew:
> I don't have a strong preference about including the `precision_radix` column right now -- MySQL doesn't seem to have this column as well. I'm okay if the `precision` column shows the right values. That is, for example, for a column of type `decimal(18, 2)`, the precision value we get should be 18 rather than 10.

@hantangwangd  Thanks for the feedback! I've made the code changes .

Please take a look and let me know if the changes look good — I’ll proceed to update the test cases accordingly.

hantangwangd:
@sumi-mathew Thanks for the fix, looks good to me.

yingsu00 (commented):
@sumi-mathew Why not just output `decimal(p, s)` as the type? Precision and Scale only applies to decimals and don't mean anything for other types. I think that's what Trino does as well.

sumi-mathew:
> Why not just output `decimal(p, s)` as the type? Precision and Scale only applies to decimals and don't mean anything for other types. I think that's what Trino does as well.

@yingsu00   Already supported in Presto: DECIMAL(p, s) types are rendered with precision and scale embedded in the type signature, similar to Trino. 

Screenshot 2025-08-12 at 2 31 40 PM

I've outlined the goal of this PR in the following comment: https://github.com/prestodb/presto/pull/25351#issuecomment-3166609871.

yingsu00 (commented):
What was the reason why we had to add the precision and scale columns for SHOW COLUMNS? The SQL standard does not define a SHOW COLUMNS statement at all, so there’s no requirement for it to match the output of INFORMATION_SCHEMA.COLUMNS. Since the current output of SHOW COLUMNS already contains precision and scale, why do we need to add extra columns?

sumi-mathew:
> What was the reason why we had to add the precision and scale columns for SHOW COLUMNS? The SQL standard does not define a SHOW COLUMNS statement at all, so there’s no requirement for it to match the output of INFORMATION_SCHEMA.COLUMNS. Since the current output of SHOW COLUMNS already contains precision and scale, why do we need to add extra columns?

@aaneja Could you please help me respond to this comment, or share your opinion on it?

aaneja:
Its just syntactic sugar. It was implemented internally, and our customers preferred it, so we thought to open source the change. 
IMO having NULLs for types that do not have a precision or scale is fine

Since its not SQL standard, its up to us to decide if having these as separate columns makes it easier to gork what the column types are. To me this does it make it slightly easier to parse

yingsu00:
THanks @aaneja for explaining. 
@sumi-mathew If our customers need this, why didn't WxD query information_schema.columns to get the info? I suppose this is some UI WxD provides and whether to use information_schema.columns or show columns is a decision of WxD, not our customers?
It's not too big problem here but just curious.

sumi-mathew:
@yingsu00   External customers are using the Presto JDBC driver to connect to WXD and are also using some console APIs to retrieve datasets. In this case they are not using the WXD UI

PR #25359, opened 2025-06-18
Add output size calculation to BinarySortableSerializer

emilysun201309:
Summary: `BinarySortableSerializer::serializedSizeInBytes` does a pass over the input data to calculate the output size in bytes. This is used to pre-determine the output buffer size to allocate.

Differential Revision: D76008301

```

== RELEASE NOTES ==
* Add BinarySortableSerializer::serializedSizeInBytes method that returns the serialized byte size of a given input row at 'rowId'. This allows us to pre-allocated the exact output buffer size needed for serialization, avoiding wasted memory space. 
```


steveburnett:
Thanks for the release note! Consider if you can add in the release note entry, describing what this means for a Presto user.

xiaoxmeng (dismissed):

